PR #3, opened 2017-05-27
Added ability to save tensors

### LINKED ISSUE #2: Need API for saving to file.
Currently we only have a load function. But after training the network it would be great if we can save the generated tensor to a file.

@soumith I think we can reuse a lot of code from this repo https://github.com/MattVitelli/GRUV

SeanNaren:
Ported the code from [here](https://github.com/soumith/lua---audio) to save tensors! Should also close #2, let me know if any feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Quick Usage
 ```python
 import torchaudio
 sound, sample_rate = torchaudio.load('foo.mp3')
+torchaudio.save('foo_save.mp3', sound, sample_rate) # saves tensor to file
 ```
 
 API Reference
@@ -46,6 +47,16 @@ loads an audio file into a Tensor
 audio.load(
 	string,  # path to file
 	out=None, # optionally pass output Tensor (any CPU Tensor type)
+)
+```
+
+torchaudio.save
+```
+saves a tensor into an audio file. The extension of the given path is used as the saving format.
+audio.save(
+    string, # path to file
+	tensor, # NSamples x NChannels 2D tensor
+	number, # sample_rate of the audio to be saved as
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ torchaudio.save
 ```
 saves a tensor into an audio file. The extension of the given path is used as the saving format.
 audio.save(
-    string, # path to file
+	string, # path to file
 	tensor, # NSamples x NChannels 2D tensor
 	number, # sample_rate of the audio to be saved as
 )
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -1,3 +1,5 @@
+import os
+
 import torch
 
 from cffi import FFI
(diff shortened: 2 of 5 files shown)


torchaudio/__init__.py
```diff
@@ -30,3 +32,27 @@ def load(filename, out=None):
...
+        func = th_sox.libthsox_Byte_write_audio_file
+    elif isinstance(src, torch.CharTensor):
+        func = th_sox.libthsox_Char_write_audio_file
+    elif isinstance(src, torch.ShortTensor):
+        func = th_sox.libthsox_Short_write_audio_file
+    elif isinstance(src, torch.IntTensor):
+        func = th_sox.libthsox_Int_write_audio_file
+    elif isinstance(src, torch.LongTensor):
+        func = th_sox.libthsox_Long_write_audio_file
+    func(bytes(filepath), src, extension.replace('.', ''), sample_rate)
```
apaszke: why not just slice the first character?

torchaudio/__init__.py
```diff
@@ -30,3 +32,27 @@ def load(filename, out=None):
...
+    elif isinstance(src, torch.ByteTensor):
+        func = th_sox.libthsox_Byte_write_audio_file
+    elif isinstance(src, torch.CharTensor):
+        func = th_sox.libthsox_Char_write_audio_file
+    elif isinstance(src, torch.ShortTensor):
+        func = th_sox.libthsox_Short_write_audio_file
+    elif isinstance(src, torch.IntTensor):
+        func = th_sox.libthsox_Int_write_audio_file
+    elif isinstance(src, torch.LongTensor):
+        func = th_sox.libthsox_Long_write_audio_file
```
apaszke: How about
```python
if not torch.is_tensor(src):
    raise ...
if not src.__module__ == 'torch':
    raise ...
typename = type(src).__name__.replace('Tensor', '')
func = getattr(th_sox, 'libthsox_{}_write_audio_file'.format(typename))
```

torchaudio/__init__.py
```diff
@@ -30,3 +32,27 @@ def load(filename, out=None):
     func(bytes(filename), out, sample_rate_p)
     sample_rate = sample_rate_p[0]
     return out, sample_rate
+
+
+def save(filepath, src, sample_rate):
+    assert torch.is_tensor(src)
+    assert not src.is_cuda
```
apaszke: It's better to use ifs and raise proper errors. You can use assertions to check invariants, not arguments

SeanNaren: Would it make sense for me to change this in the load function as well?

apaszke: Yeah, sure

SeanNaren: Did it anyway :)

SeanNaren:
Thanks for the feedback @apaszke, made a few changes if you could review again!

README.md
```diff
@@ -49,3 +50,13 @@ audio.load(
 )
 ```
 
+torchaudio.save
+```
+saves a tensor into an audio file. The extension of the given path is used as the saving format.
+audio.save(
+    string, # path to file
+	tensor, # NSamples x NChannels 2D tensor
+	number, # sample_rate of the audio to be saved as
```
apaszke: it seems that you mixed up tabs and spaces here

PR #4, opened 2017-06-12
bytes function requires type

dhpollack:
I had the following error after installation.  This patch fixed the problem.

>>> import torchaudio
>>> torchaudio.load("ffea344c-e8a6-4697-a952-2580c4c15e48.mp3")
  File "/home/david/miniconda3/envs/hu/lib/python3.6/site-packages/torchaudio-0.1-py3.6-linux-x86_64.egg/torchaudio/__init__.py", line 26, in load
    func(bytes(filename), out, sample_rate_p)
TypeError: string argument without an encoding


apaszke:
Thanks!

PR #9, opened 2017-07-24
cffi dependency

taras-sereda:
ImportError: No module named 'cffi'. Necessary for both mac and linux installations


apaszke:
Except that you should never install anything from pip with superuser permissions (it can execute any code).

PR #10, opened 2017-07-31
add gitignore

dhpollack:
This repo has no gitignore, which causes all the compiled files to get tracked by git if one builds from source.


soumith:
thanks!

PR #11, opened 2017-08-12
basic transforms and two datasets

dhpollack:
I created some basic transforms and datasets similar to the torchvision module.

transforms: scale, trim/pad, downmix to mono

datasets: [vctk dataset](http://datashare.is.ed.ac.uk/handle/10283/2651), [yesno dataset](http://www.openslr.org/1/)


soumith:
looks great! thanks.

PR #13, opened 2017-08-19
update save/load to prevent bad inputs

dhpollack:
added a bunch of checks to the load and save functions so that sox doesn't get a bad input and throw a strange error or crash.

load:
* check if file exists
* option to normalize tensor

save:
* convert 1d inputs from (L, ) to (L, 1)
* fail if inputs are not (L, ), (L, 1) or (L, 2)
* convert sample rate to type int if it's simply an integer as type float
* fail if sample rate is not an integer
* fail if directory does not exist
* automatically unnormalize files [-1, 1] -> 16-bit depth 

created unit tests for all these things.


soumith:
thanks a lot for this pr (and your comments on others). I've been out of commission for the last couple of weeks. I'll be back Monday and will review/merge.

soumith:
looks excellent! thanks.

PR #15, opened 2017-09-03
add MEL spectrograms transform and fixed a few tests.

dhpollack:
* Added an MEL spectrogram transform with librosa.  If librosa is not installed this fails gracefully.
* added two convenience permutation transformations.
* Removed the old tests in test/test.py and redid the relevant ones in the unit tests.
* created unit tests for the MEL spectrogram transform and the permutation transforms.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -1,11 +1,12 @@
 import unittest
+import torch
 import torchaudio
 import math
 import os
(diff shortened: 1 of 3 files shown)


soumith:
good stuff!

PR #17, opened 2017-10-08
add get_info, enhance load, remove normalization option to load

greaber:
I enhanced the load function to allow specifying an offset and number of desired frames, so that you can load just part of a long file.  I also added a get_info convenience function since you often will need to know the length of the audio before reading it with an offset.  I also made an incompatible API change by removing the `normalization` option from load and just making the `normalization=True` behavior the default.  I think this is more intuitive since the terminology "normalization" is a little weird for a simple rescaling, and other libraries like librosa just return audio in [-1,1], and it is simple for the user to rescale it to another interval if desired.

Further enhancements could try to make the functions work as similarly to possible to librosa, for instance one discrepancy currently I believe is that librosa will produce 1d tensors for mono (and also supports a "mono" option that monoizes the audio).

However, one bad thing about librosa is it requires you to specify offset and amount of audio to read in seconds, which could lead to rounding issues.  I think it is better to specify this info in frames.  (By the way, I'm not sure if "frame" is the best terminology.  What do you call the thing that a CD has 44,100 of per second?  I am saying "frames".  You could also say "samples", but arguably it has twice that many samples because of stereo.)


greaber:
One issue I just realized with my PR is that currently the library claims to support non-floating point tensors in some form, but I guess removing the option to not do normalization will break that.  On the other hand, do these non-floating point tensor versions of the load function even work correctly now?  They are not tested, and it wouldn't surprise me if no one uses them.  I don't really understand how these generic tensor macros work, but currently the 32 bit integers returned by sox are cast to "real" in `libthsox_(read_audio)`, and I wonder if this will really work for non-floating point (especially less than 32 bits).  Possibly the non-floating point versions should just be removed unless someone is willing to think a bit about how they should work.

Also, it would be logical to move normalization to C and just do it in the initial loop copying from the sox buffer to the tensor, although it might not give a noticeable speedup since the normalization  operation should already be pretty fast.

greaber:
It looks like there are some conflicts now that would have to be resolved before this pull request could be merged, but what in general is holding it up?  Loading just a small piece of an audio file is much faster with this patch than with librosa or (even slower) loading the whole file with sox and then just taking what you need.

soumith:
See https://github.com/pytorch/audio/blob/master/torchaudio/torch_sox.cpp and https://github.com/pytorch/audio/blob/master/torchaudio/torch_sox.h , they are the only C++ files in the repo now.

soumith:
@dhpollack has rewrote and merged in #59

PR #19, opened 2017-10-10
Quick and dirty fix for "bytes" (in torchaudio.save) not available in python2

raff:
Since pytorch is python2 compatible, torchaudio should also be python2 compatible.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import torch
 import torchaudio
 import torchaudio.transforms as transforms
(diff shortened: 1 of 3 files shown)


dhpollack:
This is something I didn't catch and the fix works.  

Could you add the __future__ print_function import to the test/test_transforms.py file?  As long as we are making this python2 compatible, we should go ahead and add that as well.

raff:
Done! torchaudio/transforms.py was also missing the import.

dhpollack:
great, I'll merge this

PR #28, opened 2018-01-23
Add __repr__ for transforms

vishwakftw:
Just like how `__repr__` exists for `transform` objects in `torchvision`, I have added them for `transform` objects here.

Tests are also included.


soumith:
awesome, thanks!

PR #34, opened 2018-03-12
Rewrote C code with C++ extensions

goldsborough:
Rewrote the C backend with our new C++ extensions. Much less config, much less dependencies (-cffi), much less code, much more happy.

Test plan:
- Ran existing tests,
- Wrote additional test to make sure that loading -> writing -> loading is an identity operation,
- Loaded an mp3 and wav file, and re-saved them. Listened to them and verified subjectively they're the same.

@soumith @colesbury


goldsborough:
Will add more tests. Still WIP.

torchaudio/torch_sox.cpp
```diff
@@ -0,0 +1,117 @@
...
+        "Error reading audio file: empty file or read failed in sox_read");
+  }
+
+  output.resize_({samples_read / number_of_channels, number_of_channels});
+  output = output.contiguous();
+
+  AT_DISPATCH_ALL_TYPES(output.type(), "read_audio_buffer", [&] {
+    auto* data = output.data<scalar_t>();
+    std::copy(buffer.begin(), buffer.begin() + samples_read, data);
+  });
```
ezyang: It doesn't look like you ever `sox_close(fd)` here

torchaudio/torch_sox.cpp
```diff
@@ -0,0 +1,117 @@
...
+    const std::string& extension,
+    int sample_rate) {
+  if (sox_init() != SOX_SUCCESS) {
+    throw std::runtime_error("Error initializing sox library");
+  }
+
+  if (!tensor.is_contiguous()) {
+    throw std::runtime_error(
+        "Error writing audio file: input tensor must be contiguous");
+  }
```
ezyang: This is missing a test that the input tensor is 2D. (The original code was missing this too, but it didn't use `numel` to compute length, so you wouldn't ever read out of bounds. This code can easily read out of bounds if you have a multi-dim tensor.)

torchaudio/torch_sox.cpp
```diff
@@ -0,0 +1,117 @@
...
+      /*encoding=*/nullptr,
+      extension.c_str(),
+      /*filetype=*/nullptr,
+      /*oob=*/nullptr);
+
+  if (fd == nullptr) {
+    throw std::runtime_error(
+        "Error writing audio file: could not open file for writing");
+  }
+
```
ezyang: I'm not sure what the merit of deleting the `read_audio` and `write_audio` helper functions was; they seem like a reasonable code organization.

torchaudio/torch_sox.cpp
```diff
@@ -0,0 +1,117 @@
...
+  std::vector<sox_sample_t> buffer(tensor.numel());
```
ezyang: This code is a bit perplexing. `sox_sample_t` is a 32-bit integer. So why don't you just convert the tensor into an `IntTensor` and then write samples out straight from the internal tensor buffer?

torchaudio/torch_sox.cpp
```diff
@@ -0,0 +1,117 @@
...
+  AT_DISPATCH_ALL_TYPES(tensor.type(), "write_audio_buffer", [&] {
+    auto* data = tensor.data<scalar_t>();
+    std::copy(data, data + tensor.numel(), buffer.begin());
+  });
+
+  const auto samples_written = sox_write(fd, buffer.data(), buffer.size());
+
+  // Free buffer and sox structures (before possibly throwing).
+  sox_close(fd);
+  sox_quit();
```
ezyang: Another quit here

torchaudio/torch_sox.cpp
goldsborough: We sometimes pass `FloatTensor` to this code, so we need to convert to integers first

PR #45, opened 2018-05-04
Bring back cpp_extension rewrite of torchaudio

ezyang:

PR #46, opened 2018-05-08
Revert "Bring back cpp_extension rewrite of torchaudio"

soumith:
Reverts pytorch/audio#45

Still has segfaults. I can provide repro for you guys tomorrow.


soumith:
I did a fresh install of anaconda3, and used system ubuntu 14.04 compiler, with pytorch 0.4 binary from anaconda. test.py still segfaults.

PR #55, opened 2018-07-17
Update torch audio for new stft signature

SsnL:
The `stft` signature is being changed in https://github.com/pytorch/pytorch/pull/9497 (landing rn) to be consistent with `librosa`. This patch updates `transforms. SPECTROGRAM` so the behavior still matches the doc. (actually behavior changed because doc says `pad` pads on both size but the previous behavior only pads at the end). Note that the new `stft` support richer options, so cc @dhpollack on whether to update `SPECTROGRAM`.

An identical patch is also landing to facebook internal `torchaudio`.


soumith:
merged. now that 0.4.1 is released.

PR #60, opened 2018-09-11
allow usage of SoX effects and more

dhpollack:
This commit contains the following:

1) Preprocessing with SoX effects  
2) Changed default options of load and save functions  
3) Create legacy load and save functions to mimic previous defaults  
4) Allow sox_signalinfo_t and sox_encodinginfo_t types to be used when loading or saving  
5) Changed info function to return objects of the above types  
6) Added a bunch of tests for the new functions (although not an exhaustive amount to find all edge cases)  
7) Added documentation for the new functions  
8) Bumped the version number to 0.2  

I have not tested this on Windows or Mac.  Importantly, there is some OSX specific code related to `sox_open_memstream_write`, which I wrote a workaround for but was not able to test.

Some feedback on the clarity of using an effects chain would be appreciated.


greaber:
Cool, thanks.  I didn't look at everything, but here are a few nitpicks about loading and saving, which are the parts most of interest to me.

On loading:

Using `num_frames` of 0 to load everything after the offset seems weird to me.  I think it would be better to either use -1 like before or perhaps None.  Especially now that zero-size tensors are properly supported.

I think no normalization should be the default (or actually my preference was to just remove normalization from the load function altogether).

On saving:

I'm a little confused by the precision and channels_first arguments.  Say I want to save a standard pcm 16 format wav.  Will the defaults do that for me?  When would I set channels_first to False?  Isn't there some distinction between integer and floating point audio?  Also, does the function assume that the audio is floating point with max 1 and min 0 or what exactly?  Also, a standard kind of convention that many saving functions use that seems useful is that by default it determines the format based on the extension of the filename, but you can override that if you need to use some weird extension.  Does this do this?

Note that I didn't look at the code to try to understand the behavior of saving.  But I think it should be clear from the documentation.

General remark:

Given that the currently available functions in the python ecosystem for loading and saving audio all have some problems (in particular, librosa discourages using its saving functions, and its loading function is slow, while pysoundfile doesn't allow loading segments, and its saving of ogg files is buggy), it would be great to have loading and saving functions that work independently of pytorch.  (And in general, it seems like a lot of the stuff in the pytorchaudio library could just use numpy given that it doesn't exploit the gpu or do any fancy neural net type stuff.)

dhpollack:
Thanks for the feedback @greaber 

I actually had the `num_frames` as -1 to load the entire audio file.  However, it does not make any sense to load 0 frames of an audio, because it literally would not load any audio.  Also in SoX effects, 0 is often the parameter used to load an entire file.  I wouldn't be opposed to using -1 and in fact, the code currently just checks to see if num_frames is greater than 0 so using -1 would also work.  Internally, the `num_frames` number is multiplied by the number of channels so I figure 0 * channels was better than -1 * channels.  

The channels_first option puts the audio into a tensor that is num_channels x length rather than length x num_channels as previously done.  The main reason for the option is for backwards compatibility, so the legacy functions can load and save audio using the new C++ functions.  Also, internally audio is stored in formats that are buffers like [c1s1, c2s1, c1s2, c2s2,...] so internally you create a length x num_channels tensor and then with this update transpose the dimensions to get num_channels x length.  For RNNs or traditional audio processing, it could be useful to have length first rather than channels first.  

"No normalization" would return a `LongTensor` which is 31-bit depth, which is what the library does now.  Normalization makes the numbers between [-1, 1].  Again, this has a lot to do with how SoX works internally.  Data manipulated as 32-bit signed integers, so I didn't want to mask this so much that one would be forced to use the [-1, 1].  Admittedly, I don't think many people will use audio as 32-bit signed ints, but it doesn't seem necessary to not allow that option, when that's how SoX works internally.

Finally, I looked into numpy version of all these and it actually seems fairly straightforward to add the ability to do this.  Numpy data-types look like they are build into pybind11 but it wasn't a priority at the moment.  Also, currently, the library only allows loading into CPU tensors, but I believe with the new ATen system, one could directly load into a GPU tensor.  Again, something I am going to investigate further, but not a priority at the moment.

Thanks for the feedback!  I'll definitely look at the documentation and make sure that I make things a bit more clear, especially with regards to the `channels_first` option and normalization.  I do think one problem with the term "normalization" is that in audio processing "[normalization](https://en.wikipedia.org/wiki/Audio_normalization)" means "scaling" in deep-learning terminology.  Specifically, it has nothing to do with normalizing the standard deviation to 1.  It is confusing, but I feel like since this is an audio library, I would stick with the audio terminology.  I'll try to make this more clear in the documentation.

dhpollack:
@soumith, I'm feeling pretty good about these changes.  Can you test it on a mac to make sure they work?

torchaudio/__init__.py
```diff
@@ -18,29 +14,48 @@ def check_input(src):
         raise TypeError('Expected a CPU based tensor, got %s' % type(src))
 
 
-def load(filepath, out=None, normalization=None, num_frames=-1, offset=0):
+def load(filepath,
+         out=None,
+         normalization=True,
+         channels_first=True,
```
Rikorose: This is not documented below. 

Also having `True` as default parameter breaks some transforms:

- [Example in MEL2()](https://github.com/dhpollack/audio/blob/sox_effects_test/torchaudio/transforms.py#L323): LC2CL() could be removed
- [DownmixMono()](https://github.com/dhpollack/audio/blob/sox_effects_test/torchaudio/transforms.py#L124) should probably also get a channels_first parameter that defaults to the same value

test/test_dataloader.py
```diff
@@ -0,0 +1,47 @@
...
+    def __getitem__(self, index):
+        fn = self.data[index]
+        self.E.set_input_file(fn)
+        x, sr = self.E.sox_build_flow_effects()
+        return x
+
+    def __len__(self):
+        return len(self.data)
+
+class Test_LoadSave(unittest.TestCase):
```
Rikorose: [Minor] Wrong name?

torchaudio/sox_effects.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Returns: SoxEffect(object)
+      - ename (str), name of effect
+      - eopts (list[str]), list of effect options
+    """
+    return _torch_sox.SoxEffect()
+
+
+class SoxEffectsChain(object):
+    """SoX effects chain class.
```
Rikorose: An example in the documentation would be nice. Also probably mention at this point that initialize_sox() needs to be called before, if I understand correctly.

dhpollack: That's a good idea.  I'll add that.

test/test_dataloader.py
dhpollack: I don't see what you mean here.  Could you be super explicit?  I'm sure it's just that I've overlooked at this code so many times before that I can't see which name is wrong.

torchaudio/__init__.py
dhpollack: This sent me down a bit of a rabbit hole.  I changed the MEL2 example and added `channels_first` to both `DownmixMono` and `PadTrim`.  The rabbit hole is that I changed the documentation to the new pytorch documentation style.  Once I get everything in order I'll post the commit here with the changes

test/test_dataloader.py
dhpollack: `Test_LoadSave` -> `Test_DataLoader` ... got it.

dhpollack:
@Rikorose here are the changes you suggested and an update to the docs to the new pytorch theme.  No rush to look over this.  Happy holidays.

soumith:
Thanks @dhpollack , and sorry for the delay. Tested fully now and works well on OSX!

PR #67, opened 2018-11-09
Update vctk.py

dhruvramani:
The number of audios and targets don't match in the VCTK dataset, hence an IndexError occurs. The best option is to ignore the sample, which can be done by calling the __getitem__ function or returning 0 and checking it later

### DIFF
--- a/torchaudio/datasets/vctk.py
+++ b/torchaudio/datasets/vctk.py
@@ -118,20 +118,23 @@ class VCTK(data.Dataset):
         Returns:
             tuple: (image, target) where target is index of the target class.
         """
-        if self.cached_pt != index // self.chunk_size:
-            self.cached_pt = int(index // self.chunk_size)
-            self.data, self.labels = torch.load(os.path.join(
-                self.root, self.processed_folder, "vctk_{:04d}.pt".format(self.cached_pt)))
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/vctk.py
```diff
@@ -118,20 +118,23 @@ def __getitem__(self, index):
...
+
+            if self.transform is not None:
+                audio = self.transform(audio)
+
+            if self.target_transform is not None:
+                target = self.target_transform(target)
+
+            return audio, target
+        except IndexError :
+            return self.__getitem__()
```
soumith: i'm not sure I understand this change. How does `self.__getitem__()` work at all?

dhruvramani: `return self.__getitem__()` skips the current sample and returns the next one. For our usage, we returned an integer value and were checking in the training loop. However, I thought this would generalize better.

Mistobaan: I don't think this is the right solution. The correctness of the index should be checked instead. maybe close?

vincentqb:
It's not clear what issue this is trying to solve. I'll close this PR, but please feel free to open an issue instead so that we can discuss an appropriate solution if still needed.

PR #74, opened 2019-02-03
stringify filename on load for compatibility with pathlib

jph00:
`pathlib` is recommended for path handling nowadays. Simply using `str()`  makes it compatible with pathlib, without introducing a dependency (although it's been part of stdlib for quite a while now).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@ __pycache__/
 *.py[cod]
 *$py.class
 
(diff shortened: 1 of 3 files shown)


jph00:
Sorry added additional PR to this one rather than creating a new one - hope that's OK!

soumith:
thank you!

PR #76, opened 2019-02-05
Bug fix: ff top_db is None then can not check `top_db>0`

jph00:


soumith:
this was sent and merged in https://github.com/pytorch/audio/pull/75

PR #78, opened 2019-02-05
Rename classes in line with PyTorch standards. Remove redundent slow librosa-based `MEL`. Add missing docstring params.

jph00:


soumith:
thanks @jph00 !

PR #82, opened 2019-02-11
Added torch requirement to setup.py

ccoulombe:
Added `torch` as requirement to `torchaudio`. 
Does not solve when building from source but when user
```
pip install torchaudio
```


soumith:
thanks!

PR #83, opened 2019-02-11
Librosa consistency

PCerles:
This is a start, appreciate comments and any advice on achieving consistency with torch codebase.
Main modifications:
- switched the argument orders of Spectrogram, MelSpectrogram so that n_fft is the default way to construct as opposed to window size
- made normalization after torch.stft in Spectrogram a parametrized option
- changed SpectrogramToDB to the librosa implementation, which is more numerically stable
- fixed a bug in MelScale where the matrix multiply would fail if the input tensor was on a different device than the filterbank matrix
- changed default n_mels in MelScale, MelSpectrogram to align with librosa's
- added MFCC function that calculates the MFCC features for an input audio file. This uses a torch DCT implementation from another repo. Am unsure of whether it makes sense to use that repo as a library or just credit and use the function, and whether or not to make DCT into a transform class. Probably depends on how much need there is for the other types of DCT. 
- added tests for consistency against librosa for new features, which also provides visibility into which default librosa options are changed for using this library. All tests pass on my machine, for CPU and Cuda. 
Future modifications:
- adding in non-HTK mel frequencies https://librosa.github.io/librosa/_modules/librosa/core/time_frequency.html#mel_frequencies

Appreciate any comments. This is my first PR to a torch repo so open to all feedback. Thanks!
-Phil

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -86,22 +86,6 @@ class Tester(unittest.TestCase):
         repr_test = transforms.LC2CL()
         self.assertTrue(repr_test.__repr__())
 
-    def test_mel(self):
-
-        audio = self.sig.clone()
-        audio = transforms.Scale()(audio)
-        self.assertTrue(audio.dim() == 2)
-        result = transforms.MEL()(audio)
-        self.assertTrue(result.dim() == 3)
-        result = transforms.BLC2CBL()(result)
-        self.assertTrue(result.dim() == 3)
-
-        repr_test = transforms.MEL()
-        self.assertTrue(repr_test.__repr__())
-
-        repr_test = transforms.BLC2CBL()
-        self.assertTrue(repr_test.__repr__())
-
     def test_compose(self):
 
         audio_orig = self.sig.clone()
@@ -155,7 +139,7 @@ class Tester(unittest.TestCase):
         audio_orig = self.sig.clone()  # (16000, 1)
         audio_scaled = transforms.Scale()(audio_orig)  # (16000, 1)
         audio_scaled = transforms.LC2CL()(audio_scaled)  # (1, 16000)
-        mel_transform = transforms.MEL2()
+        mel_transform = transforms.MelSpectrogram()
         # check defaults
         spectrogram_torch = mel_transform(audio_scaled)  # (1, 319, 40)
         self.assertTrue(spectrogram_torch.dim() == 3)
@@ -166,7 +150,7 @@ class Tester(unittest.TestCase):
         self.assertTrue(mel_transform.fm.fb.sum(1).le(1.).all())
         self.assertTrue(mel_transform.fm.fb.sum(1).ge(0.).all())
         # check options
-        mel_transform2 = transforms.MEL2(window=torch.hamming_window, pad=10, ws=500, hop=125, n_fft=800, n_mels=50)
+        mel_transform2 = transforms.MelSpectrogram(window=torch.hamming_window, pad=10, ws=500, hop=125, n_fft=800, n_mels=50)
         spectrogram2_torch = mel_transform2(audio_scaled)  # (1, 506, 50)
         self.assertTrue(spectrogram2_torch.dim() == 3)
         self.assertTrue(spectrogram2_torch.le(0.).all())
@@ -183,7 +167,7 @@ class Tester(unittest.TestCase):
         self.assertTrue(spectrogram_stereo.ge(mel_transform.top_db).all())
         self.assertEqual(spectrogram_stereo.size(-1), mel_transform.n_mels)
         # check filterbank matrix creation
-        fb_matrix_transform = transforms.F2M(n_mels=100, sr=16000, f_max=None, f_min=0., n_stft=400)
+        fb_matrix_transform = transforms.MelScale(n_mels=100, sr=16000, f_max=None, f_min=0., n_stft=400)
         self.assertTrue(fb_matrix_transform.fb.sum(1).le(1.).all())
         self.assertTrue(fb_matrix_transform.fb.sum(1).ge(0.).all())
         self.assertEqual(fb_matrix_transform.fb.size(), (400, 100))
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -136,15 +136,17 @@ class Tester(unittest.TestCase):
         self.assertTrue(repr_test.__repr__())
 
     def test_mel2(self):
+        top_db = 80.
+        s2db = transforms.SpectrogramToDB("power", top_db)
+
         audio_orig = self.sig.clone()  # (16000, 1)
         audio_scaled = transforms.Scale()(audio_orig)  # (16000, 1)
         audio_scaled = transforms.LC2CL()(audio_scaled)  # (1, 16000)
         mel_transform = transforms.MelSpectrogram()
         # check defaults
-        spectrogram_torch = mel_transform(audio_scaled)  # (1, 319, 40)
+        spectrogram_torch = s2db(mel_transform(audio_scaled))  # (1, 319, 40)
         self.assertTrue(spectrogram_torch.dim() == 3)
-        self.assertTrue(spectrogram_torch.le(0.).all())
-        self.assertTrue(spectrogram_torch.ge(mel_transform.top_db).all())
+        self.assertTrue(spectrogram_torch.ge(spectrogram_torch.max() - top_db).all())
         self.assertEqual(spectrogram_torch.size(-1), mel_transform.n_mels)
         # check correctness of filterbank conversion matrix
         self.assertTrue(mel_transform.fm.fb.sum(1).le(1.).all())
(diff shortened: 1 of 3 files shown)


dhpollack:
@PCerles 

1) tests in test_transforms.test_mel2 fail because it's expecting the output in the db scale.  You can either update the tests to put the output into the db scale or you can update them to expect a non-db scale spectrogram.  

2) your test consistency should be in the test_transforms.py file and not test.py.

3) similarly, you should write a test for MFCC in test_transforms 

4) you should run flake8 on your code.  There are a bunch of things there that need to be fixed.  

I'm still looking into the dct function.

PCerles:
Thanks!

1. Got it, somehow missed that test file. Does this mean you're okay with modifying MelSpectrogram to not return db-scaled output? Will modify tests if so

2. Got it

3. Should this test include consistency against librosa tests? 

4. Got it.

PCerles:
Addressed comments in above commits.

torchaudio/transforms.py
```diff
@@ -471,3 +518,36 @@ def __call__(self, x_mu):
...
+    Discrete Cosine Transform, Type II (a.k.a. the DCT)
+    For the meaning of the parameter `norm`, see:
+    https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.fftpack.dct.html
+    :param x: the input signal
+    :param norm: the normalization, None or 'ortho'
+    :return: the DCT-II of the signal over the last dimension
+    """
+    x_shape = x.shape
+    N = x_shape[-1]
+    x = x.contiguous().view(-1, N)
```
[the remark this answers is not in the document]
dhpollack: Also, dct should be in a transform class and not a standalone function.  It can be a method of the MFCC class if you like, but it doesn't make sense to have it as a separate function.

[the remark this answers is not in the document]
PCerles: Thanks, this looks good to me. I added it as a method of the MFCC class. I think eventually it makes sense to have it come from an outside "scipy.fftpack" torch equivalent but that doesn't really exist right now.

test/test_transforms.py
```diff
@@ -136,42 +136,157 @@ def test_mu_law_companding(self):
...
+        norm_check[:, :, 0] *= np.sqrt(n_mels) * 2
+        norm_check[:, :, 1:] *= np.sqrt(n_mels / 2) * 2
+
+        self.assertTrue(torch_mfcc_norm_none.allclose(norm_check))
+
+    def test_librosa_consistency(self):
+        try:
+            import librosa
+            import scipy
+        except:
```
dhpollack: One minor change then I'll merge this.  I'm getting a flake warning here.  Change this line to `except ImportError:` and that should solve the problem.

torchaudio/transforms.py
```diff
@@ -284,22 +291,102 @@ class SpectogramToDB(object):
...
+        self.amin = torch.tensor([1e-10])
+        self.ref_value = torch.tensor([1.0])
 
     def __call__(self, spec):
 
-        spec_db = self.multiplier * torch.log10(spec / spec.max())  # power -> dB
+        # numerically stable implementation from librosa
+        # https://librosa.github.io/librosa/_modules/librosa/core/spectrum.html
+        spec_db = self.multiplier * torch.log10(torch.max(self.amin.to(spec.device), spec))
+        spec_db -= self.multiplier * torch.log10(torch.max(self.amin.to(spec.device), self.ref_value.to(spec.device)))
```
f0k: Since `self.amin` and `self.ref_value` are scalars, it would be much cheaper to treat them as Python floats. The current implementation will result in two host-to-gpu transfers and two CUDA kernel calls to compute what's actually a constant known at construction time.

torchaudio/transforms.py
```diff
@@ -284,22 +291,102 @@ class SpectogramToDB(object):
...
+        Args:
+            sig (Tensor): Tensor of audio of size (channels [c], samples [n])
+
+        Returns:
+            spec_mel_db (Tensor): channels x hops x n_mels (c, l, m), where channels
+                is unchanged, hops is the number of hops, and n_mels is the
+                number of mel bins.
+
+        """
+        mel_spect_db = self.s2db(self.MelSpectrogram(sig))
```
f0k: This is dangerous. Any text book or wikipedia will tell that MFCCs are computed by a DCT on **logarithmic** mel spectrograms. This should be clearly documented, and possibly made an option.
(To be fair, I just checked librosa, and they also compute MFCCs from db-scaled mel spectrograms. I'm really surprised.)

torchaudio/transforms.py
```diff
@@ -284,22 +291,102 @@ class SpectogramToDB(object):
     """
     def __init__(self, stype="power", top_db=None):
         self.stype = stype
-        if top_db is not None and top_db > 0:
-            top_db = -top_db
```
f0k: You're now assuming `top_db` is positive, as in librosa. The docstring needs to change accordingly, and you will want to throw an error for negative values like librosa.

torchaudio/transforms.py
```diff
@@ -284,22 +291,102 @@ class SpectogramToDB(object):
...
 
-        spec_db = self.multiplier * torch.log10(spec / spec.max())  # power -> dB
+        # numerically stable implementation from librosa
+        # https://librosa.github.io/librosa/_modules/librosa/core/spectrum.html
+        spec_db = self.multiplier * torch.log10(torch.max(self.amin.to(spec.device), spec))
+        spec_db -= self.multiplier * torch.log10(torch.max(self.amin.to(spec.device), self.ref_value.to(spec.device)))
+
         if self.top_db is not None:
-            spec_db = torch.max(spec_db, spec_db.new_full((1,), self.top_db))
+            spec_db = torch.max(spec_db, spec_db.new_full((1,), spec_db.max() - self.top_db))
```
f0k: The docstring needs a strong warning about using `top_db`, as it will make the output depend on the maximum value in the spectrogram -- e.g., you may get different outputs when feeding an audio file in snippets and concatenating it compared to feeding it as a whole. The same applies to `MFCC`, which uses a default `top_db` of 80.

torchaudio/transforms.py
```diff
@@ -284,22 +291,102 @@ class SpectogramToDB(object):
...
+        if norm == 'ortho':
+            V_ndim_0 = V.narrow(-1, 0, 1)
+            V_ndim_rest = V.narrow(-1, 1, x_shape[-1] - 1)
+
+            V_ndim_0.div_(torch.sqrt(l) * 2)
+            V_ndim_rest.div_(torch.sqrt(l / 2) * 2)
+
+        V.mul_(2.)
+
+        return V
```
f0k: I find this implementation hard to read without comments. The `torch.cat` looks potentially expensive. Is this really faster than just constructing a DCT matrix and doing a matmul? The latter also easily allows to compute a limited number of `self.n_mfcc` components. Old implementation by me:
```python
def create_dct(dim, outdim=None, orthogonal=True):
    """
    Creates a DCT transformation matrix.
    @param dim: Dimensionality of input data to transform
    @param outdim: If given, create a matrix for the first 'outdim' DCT
        coefficients only. If omitted, create a full (squared) DCT matrix.
    @param orthogonal: If given, makes the transform orthogonal
    @return The transformation matrix, to be right-multiplied to row-wise data.
    """
    if outdim is None:
        outdim = dim
    elif outdim > dim:
        raise ValueError("DCT outdim cannot be larger than dim")
    # http://en.wikipedia.org/wiki/Discrete_cosine_transform#DCT-II
    n = np.arange(dim)
    k = np.arange(outdim)[:, np.newaxis]
    dct = np.cos(np.pi/dim * (n + 0.5) * k)
    if orthogonal:
        dct[0] *= 1.0 / np.sqrt(2)
        dct *= np.sqrt(2.0 / dim)
    return dct
```
This matrix could be computed at construction time (in torch or numpy) and then multiplied with in the forward pass.

torchaudio/transforms.py
```diff
@@ -167,18 +167,23 @@ class Spectrogram(object):
...
-
     """
-    def __init__(self, ws=400, hop=None, n_fft=None,
-                 pad=0, window=torch.hann_window, wkwargs=None):
-        self.window = window(ws) if wkwargs is None else window(ws, **wkwargs)
-        self.ws = ws
-        self.hop = hop if hop is not None else ws // 2
+    def __init__(self, n_fft=400, ws=None, hop=None,
+                 pad=0, window=torch.hann_window,
+                 power=2, normalize=False, wkwargs=None):
```
f0k: The docstring needs to be adapted to match the argument order (to be fair, it was wrong before).

torchaudio/transforms.py
```diff
@@ -327,25 +414,24 @@ class MelSpectrogram(object):
...
-        self.top_db = -80.
         self.f_max = f_max
         self.f_min = f_min
-        self.spec = Spectrogram(self.ws, self.hop, self.n_fft,
-                                self.pad, self.window, self.wkwargs)
+        self.spec = Spectrogram(n_fft=self.n_fft, ws=self.ws, hop=self.hop,
+                                pad=self.pad, window=self.window, power=2,
+                                normalize=False, wkwargs=self.wkwargs)
         self.fm = MelScale(self.n_mels, self.sr, self.f_max, self.f_min)
-        self.s2db = SpectogramToDB("power", self.top_db)
```
f0k: It's still called `spec_mel_db` in `__call__()`, this is potentially confusing.

torchaudio/transforms.py
PCerles: great call.

torchaudio/transforms.py
PCerles: Yes I was surprised by that too. I think we could add a parameter for log vs. db-scale and not complicate things too much. What do you think?

torchaudio/transforms.py
PCerles: Would you suggest we make a relative vs. absolute parameter?

torchaudio/transforms.py
PCerles: Yep, good idea. This actually gives about a ~10x speed up on CPU, maybe more on GPU. Integrating now.

dhpollack:
Is this pull request ready to go?

f0k:
> Is this pull request ready to go?

Yes, thanks for addressing my comments @PCerles! I think it's good to merge, but we may want to introduce some breaking changes later.

test/test_transforms.py
```diff
@@ -136,42 +136,157 @@ def test_mu_law_companding(self):
...
+
+        out_torch = spect_transform(sound).squeeze().cpu().numpy().T
+        self.assertTrue(np.allclose(out_torch, out_librosa, atol=1e-5))
+
+        # test mel spectrogram
+        melspect_transform = torchaudio.transforms.MelSpectrogram(sr=sample_rate, window=torch.hann_window,
+                                                                  hop=hop_length, n_mels=n_mels, n_fft=n_fft)
+        librosa_mel = librosa.feature.melspectrogram(y=sound_librosa, sr=sample_rate,
+                                                     n_fft=n_fft, hop_length=hop_length, n_mels=n_mels,
+                                                     htk=True, norm=None)
```
EgorLakomkin: Is there a test comparing melspectrogram call with the default htk=False for example?

dhpollack: no, there is not.  The purpose of this is not to create an exhaustive consistency test.  More so that it works similarly.

EgorLakomkin: Thank @dhpollack. I see, the reason I asked is that somehow I cannot match the default librosa's melspectrogram behaviour. Curious if you ever tried?

PCerles: You would have to implement non-htk mel-bins to be able to match Librosa's melspectrogram default behavior.
https://librosa.github.io/librosa/_modules/librosa/core/time_frequency.html#hz_to_mel
The librosa consistency test is more to show which options to set librosa at to match the current torchaudio implementation. Shouldn't be too hard to add non-htk mel frequency bins to torchaudio but in practice the difference doesn't seem to matter for speech recog tasks so I didn't work on it.

PR #86, opened 2019-03-06
clarify sox info length field

faroit:
I fell for this myself. The sox info struct is specifying the total number of samples and not the number of sample per channel as it is normally used.

Its probably better to clarify this, since not everyone knows the difference between samples, length, frames....


soumith:
thanks @faroit !

PR #94, opened 2019-04-04
Normalized range should be [-1, 1]

tengyifei:
If sox uses 32-bit signed audio internally, dividing by `1 << 31` should yield -1 to 1.


engineerchuan:
Thanks @tengyifei, we fixed this as part of another open Issue in https://github.com/pytorch/audio/pull/283.  Thank you for noticing!

vincentqb:
Thanks for reporting! This should be fixed now. Please feel free to re-open if not.

PR #105, opened 2019-05-16
Migrate audio transform computations into functional.py

jamarshon:
T44497670 Refactor transforms.py into functional.py.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -2,6 +2,22 @@ import numpy as np
 import torch
 
 
+__all__ = [
+    'scale',
+    'pad_trim',
+    'downmix_mono',
+    'LC2CL',
+    'spectrogram',
+    'create_fb_matrix',
+    'mel_scale',
+    'spectrogram_to_DB',
+    'create_dct',
+    'MFCC',
+    'BLC2CBL',
+    'mu_law_encoding',
+    'mu_law_expanding'
+]
+
 def scale(tensor, factor):
     # type: (Tensor, int) -> Tensor
     """Scale audio tensor from a 16-bit integer (represented as a FloatTensor)
@@ -34,7 +50,6 @@ def pad_trim(tensor, ch_dim, max_len, len_dim, fill_value):
 
     Outputs:
         Tensor: Padded/trimmed tensor
-
     """
     assert tensor.size(ch_dim) < 128, \
         "Too many channels ({}) detected, see channels_first param.".format(tensor.size(ch_dim))
@@ -316,7 +331,7 @@ def mu_law_encoding(x, qc):
     return x_mu
 
 
-def mu_law_expanding(x, qc):
+def mu_law_expanding(x_mu, qc):
     # type: (Tensor/ndarray, int) -> Tensor/ndarray
     """Decode mu-law encoded signal.  For more info see the
     `Wikipedia Entry <https://en.wikipedia.org/wiki/%CE%9C-law_algorithm>`_
@@ -325,7 +340,7 @@ def mu_law_expanding(x, qc):
     and returns a signal scaled between -1 and 1.
 
     Inputs:
-        x (Tensor): Input tensor
+        x_mu (Tensor): Input tensor
         qc (int): Number of channels (i.e. quantization channels)
 
     Outputs:
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -1,5 +1,5 @@
+import numpy as np
 import torch
-
 
 def scale(tensor, factor):
     # type: (Tensor, int) -> Tensor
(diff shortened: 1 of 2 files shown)


cpuhrsch:
Looks good! But the functionals still need tests.

torchaudio/functional.py
```diff
@@ -0,0 +1,182 @@
...
+                        window, center=True,
+                        normalized=False, onesided=True,
+                        pad_mode='reflect').transpose(1, 2)
+    if normalize:
+        spec_f /= window.pow(2).sum().sqrt()
+    spec_f = spec_f.pow(power).sum(-1)  # get power of "complex" tensor (c, l, n_fft)
+    return spec_f
+
+
+def create_fb_matrix(n_stft, f_min, f_max, n_mels):
```
cpuhrsch: We also need to add docs to the functionals. This can be akin to [functional.py](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py), but we by far don't need that much detail for now, nor the structure needed to generated the html docs. So a few lines should be neough.

torchaudio/functional.py
```diff
@@ -0,0 +1,182 @@
...
+    """
+    Creates a DCT transformation matrix with shape (num_mels, num_mfcc),
+    normalized depending on norm
+    Returns:
+        The transformation matrix, to be right-multiplied to row-wise data.
+    """
+    outdim = n_mfcc
+    dim = n_mels
+    # http://en.wikipedia.org/wiki/Discrete_cosine_transform#DCT-II
+    n = np.arange(dim)
```
cpuhrsch: We need to get rid of stuff like this in a future pull request

torchaudio/functional.py
```diff
@@ -0,0 +1,182 @@
+import numpy as np
```
cpuhrsch: We shouldn't have to depend on numpy. Let's remove this in a future pullrequest.

torchaudio/transforms.py
```diff
@@ -507,18 +409,7 @@ def __call__(self, x):
             x_mu (LongTensor or ndarray)
 
         """
-        mu = self.qc - 1.
-        if isinstance(x, np.ndarray):
```
cpuhrsch: In general we don't want to support this. If Someone wants to use numpy, they'll need to use our 0-cost conversions from their numpy array to a tensor, then pass it into this and then convert it back. We also don't want to use numpy to do our computations.

torchaudio/transforms.py
```diff
@@ -549,17 +440,7 @@ def __call__(self, x_mu):
             x (FloatTensor or ndarray)
 
         """
-        mu = self.qc - 1.
-        if isinstance(x_mu, np.ndarray):
```
cpuhrsch: Here is another instance of this conversion. Let's get rid of this kind of stuff in a future pull request.

torchaudio/functional.py
```diff
@@ -0,0 +1,359 @@
...
+        ch_dim (int): Dimension of channel (not size)
+        max_len (int): Length to which the tensor will be padded
+        len_dim (int): Dimension of length (not size)
+        fill_value (float): Value to fill in
+
+    Outputs:
+        Tensor: Padded/trimmed tensor
+    """
+    assert tensor.size(ch_dim) < 128, \
+        "Too many channels ({}) detected, see channels_first param.".format(tensor.size(ch_dim))
```
cpuhrsch: There is no "channels_first" argument to this functional. Either we migrate channel_first to become a parameter here, or we move the error message outside or we raise an Error with a different message. This also seems like an arbitrary restriction.

torchaudio/functional.py
```diff
@@ -0,0 +1,359 @@
...
+    if max_len > tensor.size(len_dim):
```
cpuhrsch: This logic still needs explaining in the docs.

jamarshon:
cc @soumith

torchaudio/functional.py
```diff
@@ -0,0 +1,358 @@
...
+    'MFCC',
+    'BLC2CBL',
+    'mu_law_encoding',
+    'mu_law_expanding'
+]
+
+
+def scale(tensor, factor):
+    # type: (Tensor, int) -> Tensor
+    """Scale audio tensor from a 16-bit integer (represented as a FloatTensor)
```
cpuhrsch: This documentation isn't quite right, because this function literally just converts to float32 and divides the input. Let's update this in the next PR.

torchaudio/functional.py
```diff
@@ -0,0 +1,358 @@
...
+        spec_f (Tensor): normal STFT
+        f_min (float): minimum frequency
+        f_max (float): maximum frequency
+        n_mels (int): number of mel bins
+        fb (Optional[Tensor]): triangular filter banks (fb matrix)
+
+    Outputs:
+        Tuple[Tensor, Tensor]: triangular filter banks (fb matrix) and mel frequency STFT
+    """
+    if fb is None:
```
cpuhrsch: We should make this mandatory, because creating this much memory every time the function is called is inefficient and not strictly restricted to compute. We can do that in a PR where we check for performance as well.

PR #107, opened 2019-05-20
 Remove numpy support in transforms.py and functional.py

jamarshon:
From previous pull request:
"In general we don't want to support this. If someone wants to use numpy, they'll need to use our 0-cost conversions from their numpy array to a tensor, then pass it into this and then convert it back. We also don't want to use numpy to do our computations."

The major difference is mu_law_expanding and mu_law_encoding no longer accepts numpy arrays as input/output.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -316,6 +316,7 @@ def mu_law_encoding(x, qc):
     Outputs:
         Tensor: Input after mu-law companding
     """
+    assert(isinstance(x, torch.Tensor)), 'mu_law_encoding expects a Tensor'
     mu = qc - 1.
     if not x.dtype.is_floating_point:
         x = x.to(torch.float)
@@ -341,6 +342,7 @@ def mu_law_expanding(x_mu, qc):
     Outputs:
         Tensor: Input after decoding
     """
+    assert(isinstance(x_mu, torch.Tensor)), 'mu_law_expanding expects a Tensor'
     mu = qc - 1.
     if not x_mu.dtype.is_floating_point:
         x_mu = x_mu.to(torch.float)
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -316,7 +316,7 @@ def mu_law_encoding(x, qc):
     Outputs:
         Tensor: Input after mu-law companding
     """
-    assert(isinstance(x, torch.Tensor)), 'mu_law_encoding expects a Tensor'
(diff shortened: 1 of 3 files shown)


torchaudio/functional.py
```diff
@@ -245,15 +245,15 @@ def create_dct(n_mfcc, n_mels, norm):
     outdim = n_mfcc
     dim = n_mels
     # http://en.wikipedia.org/wiki/Discrete_cosine_transform#DCT-II
-    n = np.arange(dim)
-    k = np.arange(outdim)[:, np.newaxis]
-    dct = np.cos(np.pi / dim * (n + 0.5) * k)
+    n = torch.arange(dim, dtype=torch.float32)
```
cpuhrsch: Don't pass the type to the constructor in case someone wants to change the default dtype from float32 to float64 (for e.g. debugging purposes).

torchaudio/functional.py
```diff
@@ -316,22 +316,19 @@ def mu_law_encoding(x, qc):
...
-        mu = torch.tensor(mu, dtype=x.dtype)
-        x_mu = torch.sign(x) * torch.log1p(mu *
-                                           torch.abs(x)) / torch.log1p(mu)
-        x_mu = ((x_mu + 1) / 2 * mu + 0.5).long()
+    if not x.dtype.is_floating_point:
+        x = x.to(torch.float)
+    mu = torch.tensor(mu, dtype=x.dtype)
+    x_mu = torch.sign(x) * torch.log1p(mu *
+                                       torch.abs(x)) / torch.log1p(mu)
+    x_mu = ((x_mu + 1) / 2 * mu + 0.5).long()
```
cpuhrsch: Instead of .long() use .to(torch.int64)

torchaudio/transforms.py
```diff
@@ -403,10 +403,10 @@ def __call__(self, x):
         """
 
         Args:
-            x (FloatTensor/LongTensor or ndarray)
+            x (FloatTensor/LongTensor)
```
cpuhrsch: Just write "Tensor" here. The functions converts anything to float anyway.

torchaudio/transforms.py
```diff
@@ -434,10 +434,10 @@ def __call__(self, x_mu):
         """
 
         Args:
-            x_mu (FloatTensor/LongTensor or ndarray)
+            x_mu (FloatTensor/LongTensor)
```
cpuhrsch: Same here

PR #111, opened 2019-05-21
Add Kaldi IO as a dependency + put  a wrapper to convert to Tensor + add test to check correct type

jamarshon:


cpuhrsch:
Nice! Could you update the docs as well. There's a [Makefile](https://github.com/pytorch/audio/tree/master/docs).

README.md
```diff
@@ -12,6 +12,7 @@ torchaudio: an audio library for PyTorch
 
 Dependencies
 ------------
+* vesis84/kaldi-io-for-python commit cb46cb1f44318a5d04d4941cf39084c5b021241e or above
```
soumith: mark this as Optional dependency. 
So move it to the bottom of the list, and prefix it with `[optional] `

README.md
```diff
@@ -6,13 +6,15 @@ torchaudio: an audio library for PyTorch
     - mp3, wav, aac, ogg, flac, avr, cdda, cvs/vms,
     - aiff, au, amr, mp2, mp4, ac3, avi, wmv,
     - mpeg, ircam and any other format supported by libsox.
+    - [Kaldi (arc/scp)](http://pytorch.org/audio/kaldi_io.html)
```
cpuhrsch: It's "ark" no?

torchaudio/kaldi_io.py
```diff
@@ -0,0 +1,99 @@
...
+    'read_vec_int_ark',
+    'read_vec_flt_scp',
+    'read_vec_flt_ark',
+    'read_mat_scp',
+    'read_mat_ark',
+]
+
+
+def _default_not_imported_method():
+    raise ImportError('Could not import kaldi_io')
```
cpuhrsch: Add an "Did you install it?" to suggest a solution (since it's probably the fact that they didn't install it).

PR #112, opened 2019-05-21
Remove numpy dependency from test_transforms.py, skip tests with missing dependencies.

jamarshon:
From https://github.com/pytorch/audio/pull/107, we removed numpy from the transforms.py and functional.py. We want to do the same for the test_transforms.py (i.e. using math.pi/math.sqrt as well as torch.allclose instead of their np equivalent).


test/test_transforms.py
```diff
@@ -224,31 +222,31 @@ def test_librosa_consistency(self):
                                                             hop_length=hop_length,
                                                             power=2)
 
-        out_torch = spect_transform(sound).squeeze().cpu().numpy().T
-        self.assertTrue(np.allclose(out_torch, out_librosa, atol=1e-5))
+        out_torch = spect_transform(sound).squeeze().cpu().t()
+        self.assertTrue(torch.allclose(out_torch, torch.from_numpy(out_librosa), atol=1e-5))
```
cpuhrsch: Make sure np.allclose and torch.allclose have similar standards on their relative and absolute errors that they consider close. Also make sure that this can't be the stricter equal.

test/test_transforms.py
```diff
@@ -224,31 +222,31 @@ def test_librosa_consistency(self):
...
                                                                   hop=hop_length, n_mels=n_mels, n_fft=n_fft)
         librosa_mel = librosa.feature.melspectrogram(y=sound_librosa, sr=sample_rate,
                                                      n_fft=n_fft, hop_length=hop_length, n_mels=n_mels,
                                                      htk=True, norm=None)
-        torch_mel = melspect_transform(sound).squeeze().cpu().numpy().T
+        torch_mel = melspect_transform(sound).squeeze().cpu().t()
 
         # lower tolerance, think it's double vs. float
-        self.assertTrue(np.allclose(torch_mel, librosa_mel, atol=5e-3))
+        self.assertTrue(torch.allclose(torch_mel.type(torch.DoubleTensor), torch.from_numpy(librosa_mel), atol=5e-3))
```
cpuhrsch: Use torch.double instead of torch.DoubleTensor.

Since you're using torch.from_numpy and melspect_transform is returning a numpy array, you still need numpy to execute this code, but this comes via the librosa dependency.

Flag this test as skipped via [@unittest.skip](https://docs.python.org/2/library/unittest.html#skipping-tests-and-expected-failures) if librosa and scipy can't be imported and instead of returning as it currently does. Also add some docs that mention this.

Again check if we can use equal instead of all close.

test/test_transforms.py
```diff
@@ -224,31 +222,31 @@ def test_librosa_consistency(self):
...
 
         # test s2db
 
         db_transform = torchaudio.transforms.SpectrogramToDB("power", 80.)
-        db_torch = db_transform(spect_transform(sound)).squeeze().cpu().numpy().T
+        db_torch = db_transform(spect_transform(sound)).squeeze().cpu().t()
         db_librosa = librosa.core.spectrum.power_to_db(out_librosa)
-        self.assertTrue(np.allclose(db_torch, db_librosa, atol=5e-3))
+        self.assertTrue(torch.allclose(db_torch, torch.from_numpy(db_librosa), atol=5e-3))
```
cpuhrsch: check if we can use equal instead of all close.

test/test_transforms.py
```diff
@@ -272,9 +270,9 @@ def test_librosa_consistency(self):
 #                                             n_mels=n_mels)
 
         librosa_mfcc = scipy.fftpack.dct(db_librosa, axis=0, type=2, norm='ortho')[:n_mfcc]
-        torch_mfcc = mfcc_transform(sound).squeeze().cpu().numpy().T
+        torch_mfcc = mfcc_transform(sound).squeeze().cpu().t()
 
-        self.assertTrue(np.allclose(torch_mfcc, librosa_mfcc, atol=5e-3))
+        self.assertTrue(torch.allclose(torch_mfcc.type(torch.DoubleTensor), torch.from_numpy(librosa_mfcc), atol=5e-3))
```
cpuhrsch: check if we can use equal instead of all close.

test/test_transforms.py
```diff
@@ -224,31 +222,31 @@ def test_librosa_consistency(self):
...
         db_librosa = librosa.core.spectrum.power_to_db(out_librosa)
-        self.assertTrue(np.allclose(db_torch, db_librosa, atol=5e-3))
+        self.assertTrue(torch.allclose(db_torch, torch.from_numpy(db_librosa), atol=5e-3))
 
-        db_torch = db_transform(melspect_transform(sound)).squeeze().cpu().numpy().T
+        db_torch = db_transform(melspect_transform(sound)).squeeze().cpu().t()
         db_librosa = librosa.core.spectrum.power_to_db(librosa_mel)
 
-        self.assertTrue(np.allclose(db_torch, db_librosa, atol=5e-3))
+        self.assertTrue(torch.allclose(db_torch.type(torch.DoubleTensor), torch.from_numpy(db_librosa), atol=5e-3))
```
cpuhrsch: check if we can use equal instead of all close and don't use torch.DoubleTensor.

test/test_transforms.py
cpuhrsch: And don't use torch.DoubleTensor, but .to(torch.double).

cpuhrsch:
"test_librosa_consistency" has a fixed set of arguments. Can you create an "_test_librosa_consistency" and call it with a few different arguments. Also check if we can use equal instead of allClose.

jamarshon:
equal doesn't seem to work 5e-3 is 0.005 which is fairly large absolute error. both torch.allclose and np.allclose use the same formula. 

absolute(a - b) <= (atol + rtol * absolute(b))
https://docs.scipy.org/doc/numpy/reference/generated/numpy.allclose.html

https://pytorch.org/docs/stable/torch.html#torch.allclose
|self − other| ≤ atol + rtol × |other|

test/test_transforms.py
```diff
@@ -272,9 +264,41 @@ def test_librosa_consistency(self):
...
         librosa_mfcc = scipy.fftpack.dct(db_librosa, axis=0, type=2, norm='ortho')[:n_mfcc]
-        torch_mfcc = mfcc_transform(sound).squeeze().cpu().numpy().T
-
-        self.assertTrue(np.allclose(torch_mfcc, librosa_mfcc, atol=5e-3))
+        torch_mfcc = mfcc_transform(sound).squeeze().cpu().t()
+
+        self.assertTrue(torch.allclose(torch_mfcc.type(torch.double), torch.from_numpy(librosa_mfcc), atol=5e-3))
+
+    def test_(self):
+        kwargs1 = {
```
cpuhrsch: What is the reasoning behind these particular parameters?

test/test_transforms.py
```diff
@@ -194,61 +192,55 @@ def test_mfcc(self):
...
 
         self.assertTrue(torch_mfcc_norm_none.allclose(norm_check))
 
-    def test_librosa_consistency(self):
+    def _test_librosa_consistency_helper(self, n_fft, hop_length, power, n_mels, n_mfcc, sample_rate):
         try:
             import librosa
             import scipy
         except ImportError:
+            self.skipTest("Ensure that librosa and scipy are installed")
```
cpuhrsch: This should be used as an annotation "@unittest.skip("[...]")" on the method that calls this. Also, issue a warning instead of an error. "Librosa and scipy are not available" or such.

test/test_transforms.py
cpuhrsch: Instead of calling into "self._test_..." make "self._test_..." a local function "_test_" of def "test_" and rename "test_" to something more meaningful. This should show up as a separate test on the command line when run with "python test_transforms.py --verbose". The local function is an abstraction to cover multiple arguments as you do here.

jamarshon: Done. Not sure about those parameters, I just increased/decreased the sample rate, number of fourier/mel bins arbitrarily.

cpuhrsch: Alright, let's revisit that as a follow-up PR.

PR #114, opened 2019-05-22
Use common_utils.py to check for correct import in torchaudio/kaldi_io.py

jamarshon:
Now that common_utils.py provides a robust method to check for whether an import is successful, we should use the same function to check for librosa in torchaudio/kaldi_io.py.
Change autodata to autofunction because we no longer need to wrap methods.


jamarshon:
@cpuhrsch

Issue #101, opened 2019-05-04
test_dataloader.py fails


hagenw:
Under Ubuntu 18.04 I did the following:
```bash
$ git clone git@github.com:pytorch/audio.git pytorch-audio
$ cd pytorch-audio
$ sudo apt-get install sox libsox-dev libsox-fmt-all libpython3.6-dev
$ virtualenv --python=/usr/bin/python3.6 --no-site-packages _env
$ source _env/bin/activate
$ pip install torch
$ python setup.py install
$ python -m unittest test/test_dataloader.py
python: effects_i_dsp.c:189: update_fft_cache: Assertion `fft_len >= 0' failed.
Aborted (core dumped)
```

It also fails for Python 3.5 unedr Ubuntu Xenial on travis, see #100.

PR #115, opened 2019-05-23
Setup Travis CI so that unit tests are run automatically

jamarshon:


test/test_dataloader.py
```diff
@@ -36,11 +38,9 @@ def test_1(self):
         expected_size = (2, 1, 16000)
         ds = TORCHAUDIODS()
         dl = DataLoader(ds, batch_size=2)
-        for x in dl:
-            # print(x.size())
-            continue
-
-        self.assertTrue(x.size() == expected_size)
```
cpuhrsch: Just flagging this to be uncommented if ready.

jamarshon:
Also updated README.md to have badges for test and code coverage. Some existing tests that are failing include test_sox_effects.py and test_dataloader.py are commented out. Future PR will uncomment and fix them. See https://travis-ci.org/jamarshon/audio

[Build Status](https://travis-ci.org/jamarshon/audio)
[Code Coverage](https://codecov.io/gh/jamarshon/audio)

test/test_dataloader.py
```diff
@@ -1,48 +1,47 @@
...
-
-if __name__ == '__main__':
-    torchaudio.initialize_sox()
-    unittest.main()
-    torchaudio.shutdown_sox()
+# import unittest
```
cpuhrsch: Why is this still uncommented?

cpuhrsch:
@jamarshon can you send a PR that fixes the tests before we land this? I rather not land a lot of uncommented code.

jamarshon:
sox is very buggy. initialize_sox() and shutdown_sox() can be called only once in the entire program execution. An example of segmentation fault would be:
```
torchaudio.initialize_sox()
torchaudio.load(...)
torchaudio.shutdown_sox()
torchaudio.initialize_sox()
torchaudio.load(...)
```
One solution is to call initialize_sox() when the torchaudio module is loaded because it is a singleton. Cleanup would be handled by atexit when the program terminates.
http://man7.org/linux/man-pages/man3/atexit.3.html
https://github.com/pytorch/pytorch/blob/dfcd7b0185d479f186ddb100c761f4df3495f8e8/caffe2/python/_import_c_extension.py#L52

The issue with fixing sox is out of the scope of this PR. Instead, we will fix the tests which do not call initialize_sox/shutdown_sox as python -m unittest or py.test does not call the \_\_main\_\_parts of the test.
https://github.com/pytorch/audio/issues/101

pytest defaults to running all tests in a single process in sequence which would have initialize_sox/shutdown_sox issues and when a seg fault occurs, the remaining tests are not executed. the solution is to use pytest-forked which runs each test on a separate process.

requirements.txt
```diff
@@ -0,0 +1,25 @@
...
+# Coverage statistics
+pytest-cov
+codecov
+
+# Used for comparison of outputs in tests
+librosa
+scipy
+
+# Parallel tests
+pytest-forked
```
cpuhrsch: I'm not a fan of this. Their [README](https://github.com/pytest-dev/pytest-forked) reads

"future maintenance beyond the bare minimum is not planned until a new maintainer is found."

It also says that it doesn't support Windows, but PyTorch has Windows users.

PR #117, opened 2019-05-24
Setup Travis CI so that unit tests are run automatically

jamarshon:
https://github.com/pytorch/audio/pull/115


test/test_dataloader.py
```diff
@@ -9,11 +10,12 @@
 
 class TORCHAUDIODS(Dataset):
 
-    test_dirpath = os.path.dirname(os.path.realpath(__file__))
+    test_dirpath, test_dir = test.common_utils.create_temp_assets_dir()
 
     def __init__(self):
         self.asset_dirpath = os.path.join(self.test_dirpath, "assets")
-        self.data = [os.path.join(self.asset_dirpath, fn) for fn in os.listdir(self.asset_dirpath)]
+        sound_files = list(filter(lambda x: '.wav' in x or '.mp3' in x, os.listdir(self.asset_dirpath)))
```
cpuhrsch: You can use "os.path" to get the ending of a file which will be able to better deal with (admittedly unlikely) cases like "3.wav.ogg". Let's mark that for the follow-up work of deduping test.py and test_legacy.py

PR #118, opened 2019-05-28
Modifying modules and functions to support JIT and cuda

jamarshon:


build_tools/travis/install.sh
```diff
@@ -53,7 +53,7 @@ pip install -r requirements.txt
  # Install the following only if running tests
 if [[ "$SKIP_TESTS" != "true" ]]; then
     # PyTorch
-    conda install --yes pytorch -c pytorch
+    conda install --yes pytorch-nightly-cpu -c pytorch
```
cpuhrsch: Make sure you communicate that this is what we expect users to install, if we're testing on it. In the future, we'll need to standardize on minor releases of PyTorch and use the import checks to guard against usage of potentially unsupported features.

jamarshon:
@zhangguanheng66

build_tools/travis/install.sh
```diff
@@ -52,8 +52,8 @@ pip install -r requirements.txt
 
  # Install the following only if running tests
 if [[ "$SKIP_TESTS" != "true" ]]; then
-    # PyTorch
-    conda install --yes pytorch -c pytorch
+    # PyTorch (nightly as 1.1 does not have Optional for type annotations)
```
cpuhrsch: Can you work around that?

torchaudio/functional.py
```diff
@@ -123,84 +140,57 @@ def spectrogram(sig, pad, window, n_fft, hop, ws, power, normalize):
...
+def create_fb_matrix(n_stft, sr, f_min, f_max, n_mels):
+    # type: (int, int, float, Optional[float], int) -> Tensor
     """ Create a frequency bin conversion matrix.
 
     Inputs:
         n_stft (int): number of filter banks from spectrogram
+        sr (int): sample rate of audio signal
         f_min (float): minimum frequency
-        f_max (float): maximum frequency
+        f_max (float, optional): maximum frequency. default: `sr` // 2
```
cpuhrsch: This is marked as optional, but the argument is not a keyword argument. Is this to support "None"?

jamarshon: Yes in JIT, Optional[x] means it can be type x or None. I guess for Python optional it means it has a default value e.g. ```def foo(x=1.0)```. I'll update this to be more clear

zhangguanheng66: I remember you could use None-type key argument as Option Tensor.

test/test_jit.py
```diff
@@ -0,0 +1,442 @@
...
+    @unittest.skipIf(not RUN_CUDA, "no CUDA")
+    def test_scriptmodule_pad_trim(self):
+
+        class MyModule(torch.jit.ScriptModule):
+            def __init__(self, max_len):
+                super(MyModule, self).__init__()
+                self.module = transforms.PadTrim(max_len)
+                self.module.eval()
+
+            @torch.jit.script_method
```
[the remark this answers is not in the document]
jamarshon: I think you need to. "You do this using the @torch.jit.script decorator (for functions) or @torch.jit.script_method decorator (for methods) on subclasses of ScriptModule."
https://pytorch.org/docs/stable/jit.html#torch.jit.ScriptModule
Their example:
```
import torch
class MyModule(torch.jit.ScriptModule):
    def __init__(self, N, M):
        super(MyModule, self).__init__()
        self.weight = torch.nn.Parameter(torch.rand(N, M))

    @torch.jit.script_method
    def forward(self, input):
        return self.weight.mv(input)
```

torchaudio/functional.py
```diff
@@ -123,84 +140,57 @@ def spectrogram(sig, pad, window, n_fft, hop, ws, power, normalize):
...
-    window = window.to(sig.device)
+        # TODO add "with torch.no_grad():" back when JIT supports it
+        sig = torch.nn.functional.pad(sig, (pad, pad), "constant")
 
     # default values are consistent with librosa.core.spectrum._spectrogram
-    spec_f = torch.stft(sig, n_fft, hop, ws,
-                        window, center=True,
-                        normalized=False, onesided=True,
-                        pad_mode='reflect').transpose(1, 2)
+    spec_f = _stft(sig, n_fft, hop, ws, window,
```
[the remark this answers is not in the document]
jamarshon: we just need the type annotation
```
# type: (Tensor, int, Optional[int], Optional[int], Optional[Tensor], bool, str, bool, bool) -> Tensor
```

test/test_jit.py
```diff
@@ -0,0 +1,485 @@
+import torch
+import torchaudio.functional as F
+import torchaudio.transforms as transforms
+import unittest
+
+RUN_CUDA = torch.cuda.is_available()
+print('Run test with cuda:', RUN_CUDA)
+
+
+class Test_ScaleJIT(unittest.TestCase):
```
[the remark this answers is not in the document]
jamarshon: Done.

test/test_transforms.py
```diff
@@ -36,7 +36,7 @@ def test_scale(self):
         result = transforms.Scale()(audio_orig)
         self.assertTrue(result.min() >= -1. and result.max() <= 1.)
 
-        maxminmax = float(max(abs(audio_orig.min()), abs(audio_orig.max())))
+        maxminmax = int(max(abs(audio_orig.min()), abs(audio_orig.max())))
```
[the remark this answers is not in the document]
jamarshon: Done.

torchaudio/functional.py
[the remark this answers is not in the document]
zhangguanheng66: Then, I would suggest to write this function as a wrapper so you aren't responsible for any future changes of torch.stft function.

test/test_jit.py
```diff
@@ -0,0 +1,420 @@
...
+
+        jit_out = jit_method(tensor)
+        py_out = F.LC2CL(tensor)
+
+        self.assertTrue(torch.allclose(jit_out, py_out, atol=5e-4, rtol=1e-4))
+
+    @unittest.skipIf(not RUN_CUDA, "no CUDA")
+    def test_scriptmodule_LC2CL(self):
+
+        class MyModule(torch.jit.ScriptModule):
```
cpuhrsch: It might be possible to abstract this into a helper function that you give the specific transform and then returns the class.

test/test_jit.py
```diff
@@ -0,0 +1,420 @@
...
+            # type: (Tensor, int) -> Tensor
+            return F.mu_law_encoding(tensor, qc)
+
+        tensor = torch.rand((10, 1))
+        qc = 256
+
+        jit_out = jit_method(tensor, qc)
+        py_out = F.mu_law_encoding(tensor, qc)
+
+        self.assertTrue(torch.allclose(jit_out, py_out, atol=5e-4, rtol=1e-4))
```
cpuhrsch: Why are you changing the atol and rtol? I assume this is for floats. How did you choose these values? The default [is much stricter](https://pytorch.org/docs/stable/torch.html#torch.allclose).

torchaudio/functional.py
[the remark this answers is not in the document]
cpuhrsch: Accepted https://github.com/pytorch/pytorch/pull/21302

torchaudio/functional.py
```diff
@@ -123,84 +131,57 @@ def spectrogram(sig, pad, window, n_fft, hop, ws, power, normalize):
...
+def create_fb_matrix(n_stft, sr, f_min, f_max, n_mels):
+    # type: (int, int, float, Optional[float], int) -> Tensor
     """ Create a frequency bin conversion matrix.
 
     Inputs:
         n_stft (int): number of filter banks from spectrogram
+        sr (int): sample rate of audio signal
         f_min (float): minimum frequency
-        f_max (float): maximum frequency
+        f_max (Optional[float]): maximum frequency. If None, it will use `sr` // 2
```
cpuhrsch: It's marked as optional here, but f_max isn't a kwarg. I assume this is used simply to enable None?

torchaudio/functional.py
```diff
@@ -123,84 +131,57 @@ def spectrogram(sig, pad, window, n_fft, hop, ws, power, normalize):
...
+    """
+    f_max = f_max if f_max is not None else float(sr // 2)
     # get stft freq bins
     stft_freqs = torch.linspace(f_min, f_max, n_stft)
     # calculate mel freq bins
-    m_min = 0. if f_min == 0 else _hertz_to_mel(f_min)
-    m_max = _hertz_to_mel(f_max)
+    # hertz to mel(f) is 2595. * math.log10(1. + (f / 700.))
+    m_min = 0. if f_min == 0 else 2595. * math.log10(1. + (f_min / 700.))
+    m_max = 2595. * math.log10(1. + (f_max / 700.))
```
cpuhrsch: Add a comment that this does hertz to mel etc.

torchaudio/functional.py
```diff
@@ -224,69 +214,44 @@ def spectrogram_to_DB(spec, multiplier, amin, db_multiplier, top_db=None):
     spec_db -= multiplier * db_multiplier
 
     if top_db is not None:
-        spec_db = torch.max(spec_db, spec_db.new_full((1,), spec_db.max() - top_db))
+        spec_db = torch.max(
+            spec_db,
+            torch.tensor(float(spec_db.max()) - top_db, dtype=spec_db.dtype, device=spec_db.device))
```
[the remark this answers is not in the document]
jamarshon: it's a one element tensor which means layout doesn't really matter?

torchaudio/transforms.py
```diff
@@ -200,20 +220,30 @@ class MelScale(object):
...
+            tmp_fb = F.create_fb_matrix(spec_f.size(2), self.sr, self.f_min, self.f_max, self.n_mels)
+            # Attributes cannot be reassigned outside __init__ so workaround
+            self.fb.resize_(tmp_fb.size())
+            self.fb.copy_(tmp_fb)
+        spec_m = torch.matmul(spec_f, self.fb)  # (c, l, n_fft) dot (n_fft, n_mels) -> (c, l, n_mels)
         return spec_m
 
 
-class SpectrogramToDB(object):
+class SpectrogramToDB(torch.jit.ScriptModule):
```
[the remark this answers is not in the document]
jamarshon: https://pytorch.org/docs/stable/jit.html#torch.jit.ScriptModule
It seems to be the JIT version of nn.module meaning it has TorchScript methods, etc.

vince62s:
Hello guys,
it would be great if you coudl start to release some versions.
We have the OpenNMT-py repo depending on torchaudio but we don't want to use pytorch-nightly in Travis.
In the meantime we will pin to the previous commit but this is quite dirty.
Cheers.
Vincent

cpuhrsch:
Thanks for reporting this @vince62s .

cc @jamarshon

PR #119, opened 2019-06-06
Implementing Kaldi Spectrogram

jamarshon:
We want to have the exact same interface as Kaldi's compute-spectrogram-feats except it receives a tensor instead of a filepath. The processing will occur without Kaldi as a dependency


jamarshon:
```
"blackman_coeff-dither-energy_floor-frame_length-frame_shift-preemphasis_coefficient-raw_energy-remove_dc_offset-round_to_power_of_two-snip_edges-subtract_mean-window_type.ark"

1.5883-0-3.3674-0.4375-0.4375-0.08-true-true-true-true-false-povey.ark
mse: 9.094947017729282e-13 max_error: 1.9073486328125e-06
1.6869-0-2.8518-0.5000-0.5000-0.85-true-true-false-false-true-povey.ark
mse: 0.0 max_error: 0.0
2.7415-0-3.9695-0.5000-0.3125-0.66-false-true-true-false-true-blackman.ark
mse: 3.6379788613018216e-13 max_error: 1.9073486328125e-06
3.3803-0-2.8843-0.3125-0.3750-0.17-true-true-true-false-true-blackman.ark
mse: 0.0 max_error: 0.0
1.8771-0-2.0562-0.5625-0.3125-0.32-true-false-true-false-true-hamming.ark
mse: 1.152026651157767e-11 max_error: 9.5367431640625e-06
4.3282-0-1.0466-0.2500-0.1250-0.85-false-false-false-false-false-povey.ark
mse: 4.4330286663729623e-10 max_error: 8.153915405273438e-05
0.2862-0-0.2384-0.3750-0.0625-0.64-false-false-true-false-true-blackman.ark
mse: 2.091837770709648e-12 max_error: 5.7220458984375e-06
2.2217-0-2.5091-0.3750-0.4375-0.55-false-false-false-false-false-hamming.ark
mse: 0.0 max_error: 0.0
3.5905-0-3.8384-0.1875-0.1250-0.23-false-false-true-true-false-hamming.ark
mse: 0.0 max_error: 0.0
1.4329-0-2.0812-0.3125-0.3750-0.27-true-false-true-true-true-hanning.ark
mse: 0.0 max_error: 0.0
1.4515-0-1.2974-0.3750-0.1875-0.45-false-true-false-true-true-hamming.ark
mse: 2.1032064978498966e-11 max_error: 9.5367431640625e-06
2.0427-0-0.6448-0.3750-0.1250-0.35-true-false-false-true-true-rectangular.ark
mse: 4.926429490043072e-12 max_error: 5.7220458984375e-06
0.2032-0-2.7057-0.2500-0.1875-0.10-false-true-true-false-false-rectangular.ark
mse: 0.0 max_error: 0.0
4.3341-0-1.2792-0.2500-0.4375-0.18-false-false-false-true-false-hamming.ark
mse: 1.515824466814808e-12 max_error: 1.9073486328125e-06
```

Current max_error is < 1e-4

jamarshon:
Current max error over 131 files is 4.93E-04

jamarshon:
cc for review: @okhonko @cpuhrsch @zhangguanheng66

test/compliance/test_kaldi.py
```diff
@@ -0,0 +1,140 @@
...
+        test_dirpath = os.path.dirname(os.path.dirname(os.path.realpath(__file__)))
+        test_filepath = os.path.join(test_dirpath, 'assets', 'kaldi_file.wav')
+        sr = 16000
+        x = torch.arange(0, 20).float()
+        # between [-6,6]
+        y = torch.cos(2 * math.pi * x) + 3 * torch.sin(math.pi * x) + 2 * torch.cos(x)
+        # between [-2^30, 2^30]
+        y = (y / 6 * (1 << 30)).long()
+        # clear the last 16 bits because they aren't used anyways
+        y = ((y >> 16) << 16).float()
```
cpuhrsch: Should we write a compliance load and write based on this? Otherwise I'd expect people to be confused why torchaudio.load + kaldi spectogram doesn't give you the same results as kaldi spectogram directly

torchaudio/compliance/kaldi.py
```diff
@@ -0,0 +1,209 @@
...
+        sample_frequency (float): Waveform data sample frequency (must match the waveform file, if
+            specified there) (default = 16000.0)
+        snip_edges (bool): If True, end effects will be handled by outputting only frames that completely fit
+            in the file, and the number of frames depends on the frame_length.  If False, the number of frames
+            depends only on the frame_shift, and we reflect the data at the ends. (default = true)
+        subtract_mean (bool): Subtract mean of each feature file [CMS]; not recommended to do
+            it this way.  (default = False)
+        window_type (str): Type of window ('hamming'|'hanning'|'povey'|'rectangular'|'blackman') (default = 'povey')
+
+    Not used Inputs:
```
zhangguanheng66: what are those "not used inputs"? why leave them here?

torchaudio/compliance/kaldi.py
```diff
@@ -0,0 +1,209 @@
...
+        output_format (str): Format of the output files [kaldi, htk] (default = 'kaldi')
+
+    Outputs:
+        Tensor: a spectrogram identical to what Kaldi would output. The shape is (, `padded_window_size` // 2 + 1)
+    """
+    # numeric_limits<float>::epsilon()
+    epsilon = torch.tensor(1.19209290e-07, dtype=torch.get_default_dtype())
+
+    waveform = sig[max(channel, 0), :]  # size (n)
+    window_shift = int(sample_frequency * 0.001 * frame_shift)
```
okhonko: nit: create named variables for constants like this to improve readability.

torchaudio/compliance/kaldi.py
```diff
@@ -0,0 +1,209 @@
...
+    window_size = int(sample_frequency * 0.001 * frame_length)
+    padded_window_size = _next_power_of_2(window_size) if round_to_power_of_two else window_size
+
+    if len(waveform) < min_duration * sample_frequency:
```
okhonko: nit: create some assert for invalid `sample_frequency` (e.g. -1). 
Also, maybe move asserts lines 152-156 before this if check on 148.

okhonko:
@jamarshon Thanks for the great work!

Quick question: Which code generated all the test/assets/kaldi/*.ark files? I checked code in audio/test/* but didn't find where are they generated. Or are they copied from somewhere?

torchaudio/compliance/kaldi.py
jamarshon: Done.

torchaudio/compliance/kaldi.py
jamarshon: Done.

torchaudio/compliance/kaldi.py
jamarshon: Done.

test/compliance/test_kaldi.py
jamarshon: for torchaudio load if you set normalization=(1<<16) you will read exactly what is in the wav file because all the samples are stored as consecutive 16 bits. therefore, loading from a wav file and what kaldi reads is exactly the same.

for torchaudio save however, we must write with precision=16 (kaldi throws an error ie does not support other formats). the input to torchaudio save is not necessarily what will be saved as the 16 bits due to sox. 
https://github.com/rbouqueau/SoX/blob/master/src/wavpack.c#L138
https://github.com/chirlu/sox/blob/4927023d0978615c74e8a511ce981cf4c29031f1/src/sox.h#L800
A summary of the mapping would be:
```
// x > 0b1111111111111110111111111111111 or 2147450879
if(x > (1 << 31) - 1 - (1 << 15)) {
	// '0b111111111111111' or 32767
	return (1 << 15) - 1
} else {
	return (x + (1 << 15)) >> 16
}
```
so it takes the first 16 bits and adds one if the 17th bit is one and if the summation doesn't cause overflow.

I don't think we need a compliance load because the normalization is pretty simple/straight forward. For compliance write, it is a question of what the user expects those 16 bits to be. They could assume 32 bit to 16 bit sox mapping or they could make sure their input is just 16 bits and within the first 16 bits of 32 so that their input matches exactly what is written to file. Both cases are not kaldi specific though.

jamarshon:
@okhonko I ran kaldi's src/featbin/compute-spectrogram-feats.cc offline on the same wav file using a random over all the flags

test/compliance/test_kaldi.py
cpuhrsch: Even if the wrapper is simple / small, it's still good to add even if it's effectively just documentation.

For saving, we could throw a warning if the given format doesn't match a convention we'll simply choose and fix for consistency.

okhonko:
@jamarshon thanks for clarification. Can you please leave a comment somewhere in the code regarding  src/featbin/compute-spectrogram-feats.cc . Maybe add script/command line you used in the test folder/comment.

test/compliance/test_kaldi.py
jamarshon: Done.

jamarshon:
@okhonko Done. let me know if it's okay to land

cpuhrsch:
You need to generate docs for this also

torchaudio/compliance/kaldi.py
```diff
@@ -0,0 +1,209 @@
...
+import random
+import torch
+
+
+__all__ = [
+    'spectrogram'
+]
+
+# numeric_limits<float>::epsilon()
+EPSILON = torch.tensor(1.19209290e-07, dtype=torch.get_default_dtype())
```
gchanan: drive by comment: you can use `torch.finfo(torch.get_default_dtype()).eps`

okhonko: @gqchen, indeed, good catch. And it give the same value as cpp numeric_limits<float>::epsilon() so it will be giving same results as kaldi
>>> torch.finfo(torch.get_default_dtype()).eps
1.1920928955078125e-07

torchaudio/compliance/kaldi.py
```diff
@@ -0,0 +1,209 @@
...
+        round_to_power_of_two=True, sample_frequency=16000.0, snip_edges=True,
+        subtract_mean=False, window_type=POVEY):
+    """Create a spectrogram from a raw audio signal. This matches the input/output of Kaldi's
+    compute-spectrogram-feats.
+
+    Inputs:
+        sig (Tensor): Tensor of audio of size (c, n) where c is in the range [0,2)
+        blackman_coeff (float): Constant coefficient for generalized Blackman window. (default = 0.42)
+        channel (int): Channel to extract (-1 -> expect mono, 0 -> left, 1 -> right) (default = -1)
+        dither (float): Dithering constant (0.0 means no dither). If you turn this off, you should set
```
vincentqb: @cpuhrsch -- Do you know where this comment about `dither=0` and `energy_floor` comes from? This is the earliest I could track it to, and there is no discussion about it here.

test/compliance/generate_test_stft_data.py
```diff
@@ -0,0 +1,80 @@
...
+def generate_rand_window_type():
+    # Generates a random window type
+    return WINDOWS[random.randint(0, len(WINDOWS) - 1)]
+
+
+def run():
+    for i in range(NUMBER_OF_OUTPUTS):
+        inputs = {
+            'blackman_coeff': '%.4f' % (random.random() * 5),
+            'dither': '0',
```
vincentqb: All the tests were done with `dither=0.` and the `energy_floor` set to a random value.

torchaudio/compliance/kaldi.py
cpuhrsch: It's discussed in this function https://github.com/kaldi-asr/kaldi/blob/master/src/feat/feature-mfcc.h#L38. One of the suggested values is 1.0 for energy_floor. The default value for dither might have be chosen accidentally based on a misread, whereas for the tests it was set to 0 explicitly.

So we can choose a different default value without changing anything if it helps people

jamarshon: The options come from here (https://github.com/kaldi-asr/kaldi/blob/master/src/feat/feature-window.h#L79). All the defaults in torchaudio match the defaults in Kaldi. If there is a flag that is not set in Kaldi and if its corresponding argument is not passed into the compliance function, they would have the same default value. It may seem strange if they were to mismatch as the user would not input all 16 flags into the compliance function or similarly, input all command line flags to the Kaldi executable.

vincentqb: Thanks for the reply! It looks like kaldi does indeed use `dither=1.` as default, see [here](https://github.com/kaldi-asr/kaldi/blob/master/src/feat/feature-window.h#L57). We'll need to check if the results are different then.

PR #122, opened 2019-06-09
Add some transforms

marc-moreaux:
In their paper https://arxiv.org/abs/1711.10282, the authors use 2D convolution of monoChannel raw audio sounds with some data augmentation like stretching the sound, randomly cropping a piece of sound and padding the sound on both extremities to allow random crops on the borders

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -18,8 +18,8 @@ __all__ = [
     'mu_law_expanding',
     'crop_in_between',
     'random_crop',
-    'strech',
-    'random_strech',
+    'stretch',
+    'random_stretch',
     'opposite',
     'random_opposite'
 ]
@@ -399,7 +399,7 @@ def random_crop(tensor, size, ch_dim):
     return crop_in_between(tensor, start.item(), end.item(), ch_dim)
 
 
-def strech(tensor, factor, interpolate, ch_dim):
+def stretch(tensor, factor, interpolate, ch_dim):
     """Strech a tensor on the time dimention (not the channel one) with
     the given factor.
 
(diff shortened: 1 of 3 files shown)


torchaudio/transforms.py
```diff
@@ -444,3 +452,136 @@ def __call__(self, x_mu):
 
     def __repr__(self):
         return self.__class__.__name__ + '()'
+
+
+class Pad(object):
```
jamarshon: This class seems very similar to PadTrim (https://github.com/pytorch/audio/blob/7ee1d389bb07a77e3fcbc4afcfa59fbb96306d04/torchaudio/transforms.py#L75).
Is there any difference?

Mistobaan: does not look like. They bot end up using the same `torch.functional.pad` function.

Mistobaan:
Would be nice to merge those extra utility functions @marc-moreaux

torchaudio/functional.py
```diff
@@ -350,3 +356,120 @@ def mu_law_expanding(x_mu, qc):
...
+    Returns:
+        Tensor: a piece of the tensor
+    """
+    orig_size = tensor.size(1 - ch_dim)
+    start = torch.randint(0, orig_size - size, (1,))
+    end = start + size
+    return crop_in_between(tensor, start.item(), end.item(), ch_dim)
+
+
+def stretch(tensor, factor, interpolate, ch_dim):
```
vincentqb: Is this similar to `Resample` or `phase_vocoder`?

vincentqb:
Since this PR is out of date and contains significant overlap with existing features, I will close this PR. Please feel free to open a new issue to discuss the transforms you would like to add within torchaudio so we can discuss how to proceed.

PR #127, opened 2019-06-21
Kaldi Fbank

jamarshon:
Implementation of Kaldi's compute-fbank-feats


jamarshon:
Screen Shot 2019-06-25 at 3 49 15 PM
Screen Shot 2019-06-25 at 3 50 50 PM
```
max absolute error: 1536
max relative error: 0.08182557
```

for the case of 1536 "fbank-1.6134-0.6389-1.0625-0.8125-7384-false-184-7-0.08-true-false-true-true-false-false-false-true-2759-306-1.0000-rectangular.ark
", it has a max relative error of 2.8154142910352675e-06. It seems that since the output of fft is large (~1e4/5) then the power spectrum is large (~1e9) so while the error between ffts is small, the error between power spectrums will be large which could explain the errors.

```
 >> abs(fft) - abs(expected_fft)
tensor([[[ 0.0000,  0.0000],
         [ 0.0000,  0.0000],
         [-0.0007,  0.0000],
         [ 0.0010,  0.0000],
         [ 0.0017,  0.0000],
         [ 0.0078,  0.0044],
         [ 0.0029,  0.0039],
         [ 0.0007,  0.0005],
         [ 0.0015,  0.0029],
         [-0.0012, -0.0012],
         [-0.0037,  0.0039],
         [-0.0010, -0.0054],
         [ 0.0017, -0.0012],
         [-0.0024,  0.0003],
         [ 0.0007, -0.0005],
         [-0.0010, -0.0020],
         [ 0.0000,  0.0000]]])
 >> fft.pow(2).sum(2) - expected_fft.pow(2).sum(2)
tensor([[[  0.,   0.,   0.,   8.,  32., 896., 320.,   6.,  64., -16.,  16.,
          -24.,   6., -22.,   4.,  -8.,   0.]]])
```

jamarshon:
cc for review: @cpuhrsch @okhonko @zhangguanheng66

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,174 @@
...
+            # kaldi_output_dict = {k: v for k, v in torchaudio.kaldi_io.read_mat_ark(out_fn)}
+            # print(kaldi_output_dict)
+            print('success')
+        except Exception:
+            if os.path.exists(out_fn):
+                os.remove(out_fn)
+
+
+def parse(token):
+    if token == 'true':
```
zhangguanheng66: This parse function returns different type of results (bool, int, float). It may be a better way to parse the token.

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,174 @@
...
+# The number of samples inside the input wave file read from `SCP_PATH`
+WAV_LEN = 20
+
+# How many output files should be generated.
+NUMBER_OF_OUTPUTS = 100
+
+VERBOSE = False
+
+
+def generate_rand_boolean():
```
zhangguanheng66: Since the function has a single line, you may consider to leave them onsite.

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,174 @@
...
+    # Generates a random boolean ('true', 'false')
+    return 'true' if random.randint(0, 1) else 'false'
+
+
+def generate_rand_window_type():
```
zhangguanheng66: same as above

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,174 @@
...
+    for i in range(NUMBER_OF_OUTPUTS):
+        try:
+            nyquist = 16000 // 2
+            high_freq = random.randint(1, nyquist)
+            low_freq = random.randint(0, high_freq - 1)
+            vtln_low = random.randint(low_freq + 1, high_freq - 1)
+            vtln_high = random.randint(vtln_low + 1, high_freq - 1)
+            vtln_warp_factor = random.uniform(0.0, 10.0) if random.random() < 0.3 else 1.0
+
+        except Exception:
```
zhangguanheng66: If you plan to pass the exception, why put try/exception here?

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,174 @@
...
+            'use_power': generate_rand_boolean(),
+            'vtln_high': str(vtln_high),
+            'vtln_low': str(vtln_low),
+            'vtln_warp': '%.4f' % (vtln_warp_factor),
+            'window_type': generate_rand_window_type()
+        }
+
+        fn = 'fbank-' + ('-'.join(list(inputs.values())))
+        out_fn = OUTPUT_DIR + fn + '.ark'
+        arg = [
```
zhangguanheng66: You should be able to pass the inputs directly to "arg" without the parameter "inputs".

test/compliance/generate_fbank_data.py
jamarshon: since the token is used in keyword arguments to `fbank`, they are different types. is there a better way of doing this?

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,149 @@
+import os
+import random
+import subprocess
+import torch
+import torchaudio
+
+# Path to the compute-fbank-feats executable.
+EXE_PATH = '/scratch/jamarshon/kaldi/src/featbin/compute-fbank-feats'
```
cpuhrsch: Why is this hardcoded?

test/compliance/generate_fbank_data.py
cpuhrsch: randbool would be more consistent.

test/compliance/generate_fbank_data.py
```diff
@@ -0,0 +1,149 @@
+import os
```
cpuhrsch: If you wanted to go above and beyond you'd add argparser and parameterize the path to the executable etc.

This way, locally, someone can generate their own files for testing.

test/compliance/test_kaldi.py
```diff
@@ -94,46 +104,117 @@ def _create_data_set(self):
         self.assertTrue(sample_rate == sr)
         self.assertTrue(torch.allclose(y, sound))
 
-    def test_spectrogram(self):
+    def _parse_arg(self, arg):
```
cpuhrsch: You seem to be rewriting this. Why not create a separate "utilities" file?

cpuhrsch: Or have the "generate" files import from this.

torchaudio/compliance/kaldi.py
```diff
@@ -195,6 +175,59 @@ def spectrogram(
     if not raw_energy:
         signal_log_energy = _get_log_energy(strided_input, EPSILON, energy_floor)  # size (m)
 
+    return strided_input, signal_log_energy
+
+
+def spectrogram(
```
cpuhrsch: Do you think this can easily be JIT-able?

torchaudio/compliance/kaldi.py
```diff
@@ -4,11 +4,19 @@
...
+    'mel_scale_scalar',
+    'spectrogram',
+    'vtln_warp_freq',
+    'vtln_warp_mel_freq',
 ]
 
-# numeric_limits<float>::epsilon()
-EPSILON = torch.tensor(1.19209290e-07, dtype=torch.get_default_dtype())
+# numeric_limits<float>::epsilon() 1.1920928955078125e-07
+EPSILON = torch.tensor(torch.finfo(torch.float).eps, dtype=torch.get_default_dtype())
```
cpuhrsch: Did you compare to Kaldi when running this using doubles? If not, it might make sense to throw an error in our compliance functions if the users passes in a Tensor of doubles, or add some documentation that sets the expectation that this only works with doubles (which is enough anyway).

torchaudio/compliance/kaldi.py
jamarshon: I guess we'll see in next PR

test/compliance/generate_test_stft_data.py
```diff
@@ -1,80 +1,69 @@
...
         print(inputs)
         print(' '.join(arg))
 
         try:
-            subprocess.call(arg)
+            if verbose:
+                subprocess.call(arg)
+            else:
+                subprocess.call(arg, stderr=open(os.devnull, 'wb'), stdout=open(os.devnull, 'wb'))
+            print('success')
```
cpuhrsch: Ideally you use the "logging" package for this, but it's not that important.

test/compliance/generate_test_stft_data.py
```diff
@@ -1,80 +1,69 @@
...
         except Exception:
-            pass
+            if os.path.exists(out_fn):
```
cpuhrsch: I wouldn't remove things by default. If someone runs the command by accident this can be very frustrating.

PR #130, opened 2019-07-03
ISTFT

jamarshon:
Inverse STFT

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -37,11 +37,8 @@ then
              -O miniconda.sh
     fi
     chmod +x miniconda.sh && ./miniconda.sh -b -f
-    echo "jasonb"
-    # conda update --yes conda
     echo "Creating environment to run tests in."
     conda create -n testenv --yes python="$PYTHON_VERSION"
-    echo "jasonc"
 fi
 cd ..
 popd
@@ -49,7 +46,6 @@ popd
  # Activate the python environment we created.
 source activate testenv
 
-echo "jasona"
  # Install requirements via pip in our conda environment
 pip install -r requirements.txt
 
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -37,9 +37,11 @@ then
              -O miniconda.sh
     fi
     chmod +x miniconda.sh && ./miniconda.sh -b -f
+    echo "jasonb"
     conda update --yes conda
     echo "Creating environment to run tests in."
     conda create -n testenv --yes python="$PYTHON_VERSION"
+    echo "jasonc"
 fi
 cd ..
 popd
@@ -47,6 +49,7 @@ popd
  # Activate the python environment we created.
 source activate testenv
 
+echo "jasona"
  # Install requirements via pip in our conda environment
 pip install -r requirements.txt
 
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -38,7 +38,7 @@ then
     fi
     chmod +x miniconda.sh && ./miniconda.sh -b -f
     echo "jasonb"
-    conda update --yes conda
+    # conda update --yes conda
     echo "Creating environment to run tests in."
     conda create -n testenv --yes python="$PYTHON_VERSION"
     echo "jasonc"
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -37,6 +37,7 @@ then
              -O miniconda.sh
     fi
     chmod +x miniconda.sh && ./miniconda.sh -b -f
+    conda update --yes conda
     echo "Creating environment to run tests in."
     conda create -n testenv --yes python="$PYTHON_VERSION"
 fi
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -1,0 +1,99 @@
+import os
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    # size (2,20)
+    test_data = torch.tensor([
+        [45.4243, 81.9316, 19.1100, 32.4998, 45.3313, 68.8204, 42.0782, 19.7222,
+         76.8721, 69.9104, 27.7188, 86.3579, 30.3251, 92.0308, 70.0568, 74.8940,
+         94.3127, 82.9875, 88.8303, 96.3460],
+        [59.4262, 91.0040, 74.7672, 79.8533, 46.7943, 13.6757, 85.5145, 33.0060,
+         88.5102, 25.6912, 57.9501, 33.3326, 71.5654, 90.0321, 81.8218, 91.6907,
+         87.9834, 16.4177, 62.4474, 0.2146]
+    ]).float()
+
+    def _test_istft_helper(self, sound, kwargs):
+        stft = torch.stft(sound, **kwargs)
+        estimate = torchaudio.functional.istft(stft, length=sound.size(1), **kwargs)
+
+        # trim sound for case when constructed signal is shorter than original
+        # print(sound)
+        # print(estimate)
+        sound = sound[:, :estimate.size(1)]
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+
+        # print((sound-estimate))
+        # print((sound-estimate).abs().max())
+        self.assertTrue(torch.allclose(sound, estimate, atol=1e-4))
+
+    def test_istft(self):
+        kwargs1 = {
+            'n_fft': 12,
+            'hop_length': 4,
+            'win_length': 12,
+            'window': torch.hann_window(12),
+            'center': True,
+            'pad_mode': 'reflect',
+            'normalized': True,
+            'onesided': True,
+        }
+
+        kwargs2 = {
+            'n_fft': 12,
+            'hop_length': 2,
+            'win_length': 8,
+            'window': torch.hann_window(8),
+            'center': True,
+            'pad_mode': 'reflect',
+            'normalized': False,
+            'onesided': False,
+        }
+
+        kwargs3 = {
+            'n_fft': 15,
+            'hop_length': 3,
+            'win_length': 11,
+            'window': torch.hamming_window(11),
+            'center': True,
+            'pad_mode': 'constant',
+            'normalized': True,
+            'onesided': False,
+        }
+
+        kwargs4 = {
+            'n_fft': 5,
+            'hop_length': 2,
+            'win_length': 5,
+            'window': torch.hamming_window(5),
+            'center': False,
+            'pad_mode': 'constant',
+            'normalized': False,
+            'onesided': True,
+        }
+
+        kwargs5 = {
+            'n_fft': 3,
+            'hop_length': 2,
+            'win_length': 3,
+            'window': torch.hamming_window(3),
+            'center': False,
+            'pad_mode': 'reflect',
+            'normalized': False,
+            'onesided': False,
+        }
+
+        self._test_istft_helper(self.test_data, kwargs1)
+        self._test_istft_helper(self.test_data, kwargs2)
+        self._test_istft_helper(self.test_data, kwargs3)
+        self._test_istft_helper(self.test_data, kwargs4)
+        self._test_istft_helper(self.test_data, kwargs5)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -151,7 +151,7 @@ class TestFunctional(unittest.TestCase):
 
     def _test_istft_of_sine(self, amplitude, L, n):
         # stft of amplitude*sin(2*pi/L*n*x) with the hop length and window size equaling L
-        x = torch.arange(2 * L, dtype=torch.get_default_dtype())
+        x = torch.arange(2 * L + 1, dtype=torch.get_default_dtype())
         sound = amplitude * torch.sin(2 * math.pi / L * x * n)
         # stft = torch.stft(sound, L, hop_length=L, win_length=L,
         #                   window=torch.ones(L), center=False, normalized=False)
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -7,15 +7,8 @@ import test.common_utils
 
 
 class TestFunctional(unittest.TestCase):
-    # size (2,20)
-    test_data = torch.tensor([
-        [45.4243, 81.9316, 19.1100, 32.4998, 45.3313, 68.8204, 42.0782, 19.7222,
-         76.8721, 69.9104, 27.7188, 86.3579, 30.3251, 92.0308, 70.0568, 74.8940,
-         94.3127, 82.9875, 88.8303, 96.3460],
-        [59.4262, 91.0040, 74.7672, 79.8533, 46.7943, 13.6757, 85.5145, 33.0060,
-         88.5102, 25.6912, 57.9501, 33.3326, 71.5654, 90.0321, 81.8218, 91.6907,
-         87.9834, 16.4177, 62.4474, 0.2146]
-    ]).float()
+    data_sizes = (2,20)
+    number_of_trials = 10
 
     def _test_istft_helper(self, sound, kwargs):
         stft = torch.stft(sound, **kwargs)
@@ -27,7 +20,8 @@ class TestFunctional(unittest.TestCase):
         self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
         self.assertTrue(torch.allclose(sound, estimate, atol=1e-4))
 
-    def test_istft(self):
+    def test_istft1(self):
+        # hann_window, centered, normalized, onesided
         kwargs1 = {
             'n_fft': 12,
             'hop_length': 4,
@@ -39,6 +33,12 @@ class TestFunctional(unittest.TestCase):
             'onesided': True,
         }
 
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs1)
+
+    def test_istft2(self):
+        # hann_window, centered, not normalized, not onesided
         kwargs2 = {
             'n_fft': 12,
             'hop_length': 2,
@@ -50,6 +50,12 @@ class TestFunctional(unittest.TestCase):
             'onesided': False,
         }
 
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs2)
+
+    def test_istft3(self):
+        # hamming_window, centered, normalized, not onesided
         kwargs3 = {
             'n_fft': 15,
             'hop_length': 3,
@@ -61,6 +67,13 @@ class TestFunctional(unittest.TestCase):
             'onesided': False,
         }
 
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs3)
+
+    def test_istft4(self):
+        # hamming_window, not centered, not normalized, onesided
+        # window same size as n_fft
         kwargs4 = {
             'n_fft': 5,
             'hop_length': 2,
@@ -72,6 +85,13 @@ class TestFunctional(unittest.TestCase):
             'onesided': True,
         }
 
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs4)
+
+    def test_istft5(self):
+        # hamming_window, not centered, not normalized, not onesided
+        # window same size as n_fft
         kwargs5 = {
             'n_fft': 3,
             'hop_length': 2,
@@ -83,11 +103,9 @@ class TestFunctional(unittest.TestCase):
             'onesided': False,
         }
 
-        self._test_istft_helper(self.test_data, kwargs1)
-        self._test_istft_helper(self.test_data, kwargs2)
-        self._test_istft_helper(self.test_data, kwargs3)
-        self._test_istft_helper(self.test_data, kwargs4)
-        self._test_istft_helper(self.test_data, kwargs5)
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs5)
 
 
 if __name__ == '__main__':
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -1,5 +1,4 @@
 import math
-import os
 
 import torch
 import torchaudio
@@ -26,7 +25,7 @@ class TestFunctional(unittest.TestCase):
         # generates a random sound signal for each tril and then does the stft/istft
         # operation to check whether we can reconstruct signal
         for data_size in self.data_sizes:
-            for i in range(self.number_of_trials):
+            for _ in range(self.number_of_trials):
                 sound = torch.rand(data_size)
 
                 stft = torch.stft(sound, **kwargs)
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -8,7 +8,7 @@ import test.common_utils
 
 
 class TestFunctional(unittest.TestCase):
-    data_sizes = (2, 20)
+    data_sizes = [(2, 20), (3, 15)]
     number_of_trials = 100
 
     def setUp(self):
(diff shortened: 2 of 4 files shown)


torchaudio/functional.py
```diff
@@ -105,6 +106,136 @@ def _stft(input, n_fft, hop_length, win_length, window, center, pad_mode, normal
...
+          win_length,   # type: Optional[int]
+          window,       # type: Optional[Tensor]
+          center,       # type: bool
+          pad_mode,     # type: str
+          normalized,   # type: bool
+          onesided,     # type: bool
+          length        # type: Optional[int]
+          ):
+    # type: (...) -> Tensor
+    r""" Inverse short time Fourier Transform. This is expected to be the inverse of torch.stft.
```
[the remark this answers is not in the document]
jamarshon: test/test_functional.py

test/test_functional.py
```diff
@@ -0,0 +1,94 @@
...
+            'pad_mode': 'reflect',
+            'normalized': False,
+            'onesided': False,
+        }
+
+        self._test_istft_helper(self.test_data, kwargs1)
+        self._test_istft_helper(self.test_data, kwargs2)
+        self._test_istft_helper(self.test_data, kwargs3)
+        self._test_istft_helper(self.test_data, kwargs4)
+        self._test_istft_helper(self.test_data, kwargs5)
```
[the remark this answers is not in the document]
jamarshon: yes it is in this file. checks if x == istft(stft(x))

torchaudio/functional.py
```diff
@@ -105,6 +106,136 @@ def _stft(input, n_fft, hop_length, win_length, window, center, pad_mode, normal
...
+    could be useful. If :attr:`length` is None then padding will be aggressively removed (some loss of signal).
+
+    [1] D. W. Griffin and J. S. Lim, “Signal estimation from modified short-time Fourier transform,”
+    IEEE Trans. ASSP, vol.32, no.2, pp.236–243, Apr. 1984.
+
+    Inputs:
+        stft_matrix (Tensor): output of stft where each row of a batch is a frequency and each column is
+            a window. it has a shape of (batch, fft_size, n_frames, 2)
+        n_fft (int): size of Fourier transform
+        hop_length (Optional[int]): the distance between neighboring sliding window frames
```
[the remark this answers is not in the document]
jamarshon: Done.

jamarshon:
@vincentqb I think that [1] is sufficient to describe the overlap add method

vincentqb:
@jamarshon Thanks for specifying the default values in the header of the function.

For acceptance, we need to validate correctness by documenting and testing enough to confirm the pull request is indeed correct. Evidences make the approval process faster.

For documentation, we need to provide references for the details of the implementation provided.  For example, beside [Griffin and Lim], did you base your implementation on [scipy's](https://github.com/scipy/scipy/blob/v1.3.0/scipy/signal/spectral.py#L1180-L1459)?

For testing,  
1. we need to show important properties are fulfilled: for example, the composition of ISTFT and STFT (and of STFT and ISTFT) is the identity under appropriate conditions, 
2. the algorithm converges to a known analytical solution with appropriate speed,
3. the algorithm matches the result of a reference implementation.

If for example scipy's implementation was used as a reference, it could be mentioned here along with a notebook showing that the two results are the same (see for instance #21250).

(For reference, here are general acceptance criteria [here](https://github.com/pytorch/pytorch/issues/21988#issuecomment-503822859).)

test/test_functional.py
```diff
@@ -0,0 +1,94 @@
...
+        stft = torch.stft(sound, **kwargs)
+        estimate = torchaudio.functional.istft(stft, length=sound.size(1), **kwargs)
+
+        # trim sound for case when constructed signal is shorter than original
+        sound = sound[:, :estimate.size(1)]
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=1e-4))
+
+    def test_istft(self):
```
cpuhrsch: Can you split this into separate tests?

test/test_functional.py
```diff
@@ -0,0 +1,94 @@
...
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    # size (2,20)
+    test_data = torch.tensor([
```
cpuhrsch: Instead of using a static size, I'd suggest you create a function with a few parameters that can be used to deterministically (hence repeatably) create Tensors of various properties (sizes, dtypes, ranges, etc.).

test/test_functional.py
```diff
@@ -0,0 +1,94 @@
...
+        [45.4243, 81.9316, 19.1100, 32.4998, 45.3313, 68.8204, 42.0782, 19.7222,
+         76.8721, 69.9104, 27.7188, 86.3579, 30.3251, 92.0308, 70.0568, 74.8940,
+         94.3127, 82.9875, 88.8303, 96.3460],
+        [59.4262, 91.0040, 74.7672, 79.8533, 46.7943, 13.6757, 85.5145, 33.0060,
+         88.5102, 25.6912, 57.9501, 33.3326, 71.5654, 90.0321, 81.8218, 91.6907,
+         87.9834, 16.4177, 62.4474, 0.2146]
+    ]).float()
+
+    def _test_istft_helper(self, sound, kwargs):
+        stft = torch.stft(sound, **kwargs)
```
cpuhrsch: Don't only rely on this invariant. Also add some simpler sanity checks / edge cases. If stft and istft fail because a common component fails in a similar manner (say something like conv1d) then they could both produce garbage, but still be the inverse of the other.

test/test_functional.py
```diff
@@ -0,0 +1,94 @@
...
+            'hop_length': 4,
+            'win_length': 12,
+            'window': torch.hann_window(12),
+            'center': True,
+            'pad_mode': 'reflect',
+            'normalized': True,
+            'onesided': True,
+        }
+
+        kwargs2 = {
```
cpuhrsch: Also add some coverage of edge cases. Like empty input, window length shorter than hop length, n_fft bigger than window length, etc. - You can also check if the test predictably fails for cases in which the window isn't appropriate.

Also, add some comments as to why you chose these parameters and not some other ones specifically.

torchaudio/functional.py
```diff
@@ -105,6 +106,137 @@ def _stft(input, n_fft, hop_length, win_length, window, center, pad_mode, normal
...
+        onesided (bool): whether the STFT is onesided
+        length (Optional[int]): the amount to trim the signal by (i.e. the
+            original signal length). (Default: whole signal)
+
+    Outputs:
+        Tensor: least squares estimation of the original signal of size (batch, signal_length)
+    """
+    device = stft_matrix.device
+    fft_size = stft_matrix.size(1)
+    assert (onesided and n_fft // 2 + 1 == fft_size) or (not onesided and n_fft == fft_size)
```
cpuhrsch: Instead of asserts throw errors with descriptions. You can find examples of this within PyTorch.

jamarshon:
for reference, stft tests are being added for https://github.com/pytorch/audio/pull/131

test/test_functional.py
```diff
@@ -0,0 +1,133 @@
...
+    number_of_trials = 10
+
+    def _compare_estimate(self, sound, estimate):
+        # trim sound for case when constructed signal is shorter than original
+        sound = sound[..., :estimate.size(-1)]
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=1e-4))
+
+    def _test_istft_helper(self, sound, kwargs):
```
vincentqb: The goal of the test would be clearer if the name was `_test_istft_is_inverse_of_stft`. Similarly for tests invoking this.

test/test_functional.py
```diff
@@ -0,0 +1,133 @@
...
+        stft = torch.tensor([
+            [[4., 0.], [4., 0.], [4., 0.], [4., 0.], [4., 0.]],
+            [[0., 0.], [0., 0.], [0., 0.], [0., 0.], [0., 0.]],
+            [[0., 0.], [0., 0.], [0., 0.], [0., 0.], [0., 0.]]
+        ])
+
+        estimate = torchaudio.functional.istft(stft, n_fft=4, length=4)
+        self._compare_estimate(torch.ones(4), estimate)
+
+    def test_istft7(self):
```
vincentqb: The goal would be clearer with a name like `test_istft_of_zero`.

test/test_functional.py
```diff
@@ -0,0 +1,133 @@
...
+            'pad_mode': 'reflect',
+            'normalized': False,
+            'onesided': False,
+        }
+
+        for i in range(self.number_of_trials):
+            test_data = torch.rand(self.data_sizes)
+            self._test_istft_helper(test_data, kwargs5)
+
+    def test_istft6(self):
```
vincentqb: Is there a different name you could give to that test to describe the property you are validating?

torchaudio/functional.py
```diff
@@ -105,6 +106,150 @@ def _stft(input, n_fft, hop_length, win_length, window, center, pad_mode, normal
...
+    # we need to trim the front padding away if center
+    start = half_n_fft if center else 0
+    end = -half_n_fft if length is None else start + length
+
+    y = y[:, :, start:end]
+    window_envelop = window_envelop[:, :, start:end]
+
+    # check NOLA non-zero overlap condition
+    window_envelop_lowest = window_envelop.abs().min()
+    assert window_envelop_lowest > 1e-11, ('window overlap add min: %f' % (window_envelop_lowest))
```
vincentqb: You can add a test that triggers this error.

test/test_functional.py
jamarshon: Done.

vincentqb:
Let's use sines as analytical solution. Could you try the following analytical solution? Take `f(t) = sin(2*pi*n*t/L)` where `L` is the window length and hop. The result should be a band at frequency `2*pi*n/L`.

test/test_functional.py
```diff
@@ -0,0 +1,169 @@
...
+        # throw an error.
+        torchaudio.functional.istft(stft, **kwargs_ok)
+        self.assertRaises(AssertionError, torchaudio.functional.istft, stft, **kwargs_not_ok)
+
+    def test_istft_of_sine(self):
+        # stft of 123*sin(2*pi/5*x) with the hop length and window size equaling the period of L = 5
+        x = torch.arange(10, dtype=torch.get_default_dtype())
+        L = 5
+        amplitude = 123
+        sound = amplitude * torch.sin(2 * math.pi / L * x)
```
vincentqb: Great! Please also add the general case with nonzero positive integer `n`, i.e. `sin(2*pi*n*x/L)`.

test/test_functional.py
```diff
@@ -0,0 +1,169 @@
...
+        # stft = torch.stft(sound, L, hop_length=L, win_length=L,
+        #                   window=torch.ones(L), center=False, normalized=False)
+        stft = torch.zeros((3, 2, 2))
+        stft[1, :, 1] = -(amplitude * L) / 2.0
+
+        estimate = torchaudio.functional.istft(stft, L, hop_length=L, win_length=L,
+                                               window=torch.ones(L), center=False, normalized=False)
+        self._compare_estimate(sound, estimate, atol=1e-4)
```
vincentqb: Please use a consistent tolerance value by reducing this to `1e-6` as done elsewhere in this code.

test/test_functional.py
jamarshon: Done.

test/test_functional.py
jamarshon: An issue here is that the stft is actually 1e-5 around everything that is not the main frequency of the signal (not actually zero). So I'm guessing the fft of a pure sine wave still has very tiny bit of noise. The other tests use 1e-6 because we feed the output of stft directly into istft.
Example:
```
# amplitude=234, L=5, n=2
tensor([[[-3.0518e-05,  0.0000e+00],
         [ 8.6690e-05,  0.0000e+00]],

        [[ 4.1749e-05, -5.8500e+02],
         [ 7.2119e-05, -5.8500e+02]],

        [[-2.6490e-05,  3.1527e-05],
         [-1.3180e-05,  1.5360e-04]]])
```

jamarshon: Actually I think It's the scaling of the signal. 0.001 error for a signal of ~100 amplitude in this test instead of 0.000001 error for a signal of amplitude ~1 in the other tests.

vincentqb: In this case, let's specify the relative tolerance of 1e-5 instead of the absolute tolerance as it conveys what you are saying here more explicitly.

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # A window of ones meets NOLA but a window of zeros does not. This should
+        # throw an error.
+        torchaudio.functional.istft(stft, **kwargs_ok)
+        self.assertRaises(AssertionError, torchaudio.functional.istft, stft, **kwargs_not_ok)
+
+    def _test_istft_of_sine(self, amplitude, L, n):
+        # stft of amplitude*sin(2*pi/L*n*x) with the hop length and window size equaling L
+        x = torch.arange(2 * L, dtype=torch.get_default_dtype())
+        sound = amplitude * torch.sin(2 * math.pi / L * x * n)
```
vincentqb: The way you have set it up, L here is your number of point, so I would remove it from the solution: `sin(2 pi n x)`, and this way it will not be needed in the definition of `stft_largest_val`.

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        self._compare_estimate(sound, estimate, atol=1e-3)
+
+    def test_istft_of_sine(self):
+        self._test_istft_of_sine(amplitude=123, L=5, n=1)
+        self._test_istft_of_sine(amplitude=150, L=5, n=2)
+        self._test_istft_of_sine(amplitude=111, L=5, n=3)
+        self._test_istft_of_sine(amplitude=160, L=7, n=4)
+        self._test_istft_of_sine(amplitude=145, L=8, n=5)
+        self._test_istft_of_sine(amplitude=80, L=9, n=6)
+        self._test_istft_of_sine(amplitude=99, L=10, n=7)
```
vincentqb: The frequency n is becoming high relative to the length: we should expect to have at least 5 points per oscillation to resolve a wave properly, depending on how the inverse fourier transform works. This might be a source of higher error.

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        if n < stft.size(0):
+            stft[n, :, 1] = -stft_largest_val
+
+        if 0 <= L - n < stft.size(0):
+            # symmetric about L // 2
+            stft[L - n, :, 1] = stft_largest_val
+
+        estimate = torchaudio.functional.istft(stft, L, hop_length=L, win_length=L,
+                                               window=torch.ones(L), center=False, normalized=False)
+        self._compare_estimate(sound, estimate, atol=1e-3)
```
vincentqb: See [comment](https://github.com/pytorch/audio/pull/130#discussion_r301664547).

test/test_functional.py
jamarshon: I'm just going to keep it as it describes how long the window function is

test/test_functional.py
jamarshon: No increasing the number of points does not reduce absolute error. Decreasing the amplitude to 1 does though (absolute error of 1e-3 to 1e-5)

build_tools/travis/install.sh
```diff
@@ -37,16 +37,19 @@ then
...
     conda create -n testenv --yes python="$PYTHON_VERSION"
+    echo "jasonc"
 fi
 cd ..
 popd
 
  # Activate the python environment we created.
 source activate testenv
 
+echo "jasona"
```
vincentqb: Was this added intentionally?

build_tools/travis/install.sh
```diff
@@ -37,7 +37,6 @@ then
              -O miniconda.sh
     fi
     chmod +x miniconda.sh && ./miniconda.sh -b -f
-    conda update --yes conda
```
cpuhrsch: Why do you remove this here?

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        sound = sound[..., :estimate.size(-1)]
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=atol, rtol=rtol))
+
+    def _test_istft_is_inverse_of_stft(self, kwargs):
+        # generates a random sound signal for each tril and then does the stft/istft
+        # operation to check whether we can reconstruct signal
+        for i in range(self.number_of_trials):
+            sound = torch.rand(self.data_sizes)
```
cpuhrsch: this should be deterministic. If the test fails you don't have the data to reproduce it. So then you'll start creating all kinds of data to see if you can get a deterministic reproduction. you can write out some simple pseudo-random number generator that takes a seed.

cpuhrsch: that'd be useful in general for other case s as well

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+            'center': False,
+            'pad_mode': 'reflect',
+            'normalized': False,
+            'onesided': False,
+        }
+
+        self._test_istft_is_inverse_of_stft(kwargs5)
+
+    def test_istft_of_ones(self):
+        # stft = torch.stft(torch.ones(4), 4)
```
cpuhrsch: should this be removed?

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+            [[4., 0.], [4., 0.], [4., 0.], [4., 0.], [4., 0.]],
+            [[0., 0.], [0., 0.], [0., 0.], [0., 0.], [0., 0.]],
+            [[0., 0.], [0., 0.], [0., 0.], [0., 0.], [0., 0.]]
+        ])
+
+        estimate = torchaudio.functional.istft(stft, n_fft=4, length=4)
+        self._compare_estimate(torch.ones(4), estimate)
+
+    def test_istft_of_zeros(self):
+        # stft = torch.stft(torch.zeros(4), 4)
```
cpuhrsch: same here

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        kwargs_not_ok = {
+            'n_fft': 4,
+            'win_length': 4,
+            'window': torch.zeros(4),
+        }
+
+        # A window of ones meets NOLA but a window of zeros does not. This should
+        # throw an error.
```
cpuhrsch: Nice!

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        torchaudio.functional.istft(stft, **kwargs_ok)
+        self.assertRaises(AssertionError, torchaudio.functional.istft, stft, **kwargs_not_ok)
+
+    def _test_istft_of_sine(self, amplitude, L, n):
+        # stft of amplitude*sin(2*pi/L*n*x) with the hop length and window size equaling L
+        x = torch.arange(2 * L + 1, dtype=torch.get_default_dtype())
+        sound = amplitude * torch.sin(2 * math.pi / L * x * n)
+        # stft = torch.stft(sound, L, hop_length=L, win_length=L,
```
cpuhrsch: should this be removed?

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+import os
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    data_sizes = (2, 20)
```
cpuhrsch: I'd add multiple here, this won't help you find various boundary cases.

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+            'center': True,
+            'pad_mode': 'reflect',
+            'normalized': True,
+            'onesided': True,
+        }
+
+        self._test_istft_is_inverse_of_stft(kwargs1)
+
+    def test_istft_is_inverse_of_stft2(self):
+        # hann_window, centered, not normalized, not onesided
```
cpuhrsch: Did you also add cases for when the windows don't overlap?

test/test_transforms.py
```diff
@@ -230,10 +230,10 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
             librosa_mel = librosa.feature.melspectrogram(y=sound_librosa, sr=sample_rate,
                                                          n_fft=n_fft, hop_length=hop_length, n_mels=n_mels,
                                                          htk=True, norm=None)
+            librosa_mel_tensor = torch.from_numpy(librosa_mel)
             torch_mel = melspect_transform(sound).squeeze().cpu().t()
 
-            # lower tolerance, think it's double vs. float
```
cpuhrsch: why did you remove the comment?

torchaudio/functional.py
```diff
@@ -105,6 +106,138 @@ def _stft(input, n_fft, hop_length, win_length, window, center, pad_mode, normal
...
+          win_length=None,      # type: Optional[int]
+          window=None,          # type: Optional[Tensor]
+          center=True,          # type: bool
+          pad_mode='reflect',   # type: str
+          normalized=False,     # type: bool
+          onesided=True,        # type: bool
+          length=None           # type: Optional[int]
+          ):
+    # type: (...) -> Tensor
+    r""" Inverse short time Fourier Transform. This is expected to be the inverse of torch.stft.
```
cpuhrsch: great documentation!

test/test_transforms.py
jamarshon: it might be a double or float depending on what the function returns/default dtype

test/test_functional.py
jamarshon: yes test_istft_requires_overlap_windows

test/test_functional.py
jamarshon: I'm trying to say ```torch.stft(torch.ones(4), 4)``` returns the same result as the hardcoded value there. All the ```stft = ...``` means what pytorch command/configuration is required to get the same value.

test/test_functional.py
jamarshon: Done.

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
+import math
+import os
```
vincentqb: import os is unused :)

test/test_functional.py
```diff
@@ -0,0 +1,182 @@
...
+        # trim sound for case when constructed signal is shorter than original
+        sound = sound[..., :estimate.size(-1)]
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=atol, rtol=rtol))
+
+    def _test_istft_is_inverse_of_stft(self, kwargs):
+        # generates a random sound signal for each tril and then does the stft/istft
+        # operation to check whether we can reconstruct signal
+        for i in range(self.number_of_trials):
```
vincentqb: since i is unused, you can replace it by _

build_tools/travis/install.sh
jamarshon: There seems to be an error with the build. I'm not sure what changed but after installing conda and trying to update the packages, it will say the package already exists/cannot be installed.
```
UnsatisfiableError: The following specifications were found to be in conflict
- cffi
```

test/test_functional.py
jamarshon: Done.

build_tools/travis/install.sh
cpuhrsch: But that's a general CI error then? Why is it part of this specifically? If so, we should make that a separate PR so it's easier to revert.

test/test_functional.py
```diff
@@ -8,9 +7,13 @@
 
 
 class TestFunctional(unittest.TestCase):
-    data_sizes = (2, 20)
+    data_sizes = [(2, 20), (3, 15)]
     number_of_trials = 100
 
+    def setUp(self):
+        # we want to make sure that the random values are reproducible
+        torch.manual_seed(0)
```
cpuhrsch: This isn't always going to work. It requires a deterministic execution of calls to the random generator to be reproducible. As soon as you change that order, it'll generate different data at its call-sites.

A random number generator could be very simple like adding the last two numbers, adding some constant and then modding it by some fixed number

cpuhrsch: *some fixed prime number

test/test_functional.py
cpuhrsch: @jamarshon - please add in the comment and write out an explanation as to why this is ok.

An analytical example is something I'd expect to have 0 tolerance (intuitively).

test/test_transforms.py
cpuhrsch: The function being "librosa"? If that matters, please check it and branch or assert.

test/test_transforms.py
```diff
@@ -230,10 +230,10 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
             librosa_mel = librosa.feature.melspectrogram(y=sound_librosa, sr=sample_rate,
                                                          n_fft=n_fft, hop_length=hop_length, n_mels=n_mels,
                                                          htk=True, norm=None)
-            librosa_mel_tensor = torch.from_numpy(librosa_mel)
             torch_mel = melspect_transform(sound).squeeze().cpu().t()
 
-            self.assertTrue(torch.allclose(torch_mel.type(librosa_mel_tensor.dtype), librosa_mel_tensor, atol=5e-3))
+            # lower tolerance, think it's double vs. float
+            self.assertTrue(torch.allclose(torch_mel.type(torch.double), torch.from_numpy(librosa_mel), atol=5e-3))
```
vincentqb: That error is essentially 1e-2. What is the relative scale we are taking about here? rtol=1e-4?

jamarshon: A separate PR, if needed, can be made to fix this as it is unrelated to istft. There is a PR to fix the CI but that is strictly for fixing the build

Issue #36, opened 2018-03-16
No handler for .mp4 format.


bjuncek:
When I try to read an audio track from .mp4 file (Kinetics dataset), 
I get the following error:
`formats: no handler for file extension 'mp4'`
with the full stack trace:

```
~/bin/miniconda3/envs/pytorch03/lib/python3.6/site-packages/torch/utils/ffi/__init__.py in safe_call(*args, **kwargs)
    178                      for arg in args)
    179         args = (function,) + args
--> 180         result = torch._C._safe_call(*args, **kwargs)
    181         if isinstance(result, ffi.CData):
    182             typeof = ffi.typeof(result)

FatalError: [read_audio_file] Failure to read file at torchaudio/src/generic/th_sox.c:42
```

I take it it has something to do with my SoX install, but for the life of me can't figure out what I'm missing. Any ideas?

faroit:
on linux you may have luck installing `libsox-fmt-ffmpeg`, but I think this where we hit the limits of libsox....

bjuncek:
Is there a compile-it-from-source installation? The only thing I see are `.deb` packages and I have neither debian nor sudo access for package installation.

bjuncek:
Edit - closing this issue, as that is out of scope - support for this would have to come from SoX

PR #131, opened 2019-07-03
torchaudio-contrib: Adding (some) functionals 

ksanjeevan:
Adding functionals `stft` multichannel wrapper, `complex_norm`, `angle`, `magphase` and `phase_vocoder` from [torchaudio_contrib](https://github.com/keunwoochoi/torchaudio-contrib).

As discussed with @cpuhrsch & @jamarshon through slack and issues, here is the PR corresponding with the functionals for [phase 1](https://github.com/keunwoochoi/torchaudio-contrib/issues/61). We'll definitely have to review/change things, and we said to do it in the context of this PR.

#### Scope of this first PR
- naming convention. More thorough discussion going on in [#36](https://github.com/keunwoochoi/torchaudio-contrib/issues/36).
- shape of grams: we're using `(batch, channel, freq, time)` which is the same as `torch.stft` but not  `torchaudio`.
- `pytest` vs. `unittest`: there's been some comments through slack indicating that using `pytest` could be preferable (?).

#### Larger scope
- in `torchaudio_contrib` we've made all the low level transforms `nn.Modules`, to then be chained with `nn.Sequential` to make common transforms (e.g. calling `Spectrogram(...)` returns a `nn.Sequential(STFT(), ComplexNorm(power=1.0) )`, a stretched Melspectrogram is `nn.Sequential(STFT(), TimeStretch(fixed_rate=1.3), ComplexNorm(power=2), ApplyFilterbank())
`, etc.).
- We also have the transform parameters as buffers and not including them in the `state_dict`, which maybe we should revisit.

(cc @keunwoochoi, @faroit)


test/common_utils.py
```diff
@@ -16,3 +16,11 @@ def create_temp_assets_dir():
     copytree(os.path.join(TEST_DIR_PATH, "assets"),
              os.path.join(tmp_dir.name, "assets"))
     return tmp_dir.name, tmp_dir
+
+
+def _num_stft_bins(signal_len, fft_len, hop_length, pad):
+    return (signal_len + 2 * pad - fft_len + hop_length) // hop_length
+
+
+def _approx_all_equal(x, y, atol=1e-7):
```
jamarshon: would torch.allclose be better here?
https://pytorch.org/docs/stable/torch.html#torch.allclose

torchaudio/functional.py
```diff
@@ -305,3 +305,176 @@ def mu_law_expanding(x_mu, qc):
     x = ((x_mu) / mu) * 2 - 1.
     x = torch.sign(x) * (torch.exp(torch.abs(x) * torch.log1p(mu)) - 1.) / mu
     return x
+
+
+def stft(waveforms, fft_length, hop_length=None, win_length=None, window=None,
```
jamarshon: My understanding is this function takes waveforms of size (..., signal_len) and does stft on the last dimension to return size (..., freq, time, 2).

Is there a way to incorporate this into [spectrogram](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L109)? I think it would be preferable to have a single function for stft

test/common_utils.py
keunwoochoi: Absolutely. Strange that I missed it :o

torchaudio/functional.py
keunwoochoi: It'd be definitely possible and we can also remove `_stft()` then. But I understood this PR to be not changing the previous implementation and behavior, hence not sure about it. (It should definitely happen after the freezing anyway)

test/test_transforms.py
```diff
@@ -305,5 +305,120 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
...
+
+@pytest.mark.parametrize('rate', [0.5, 1.01, 1.3])
+@pytest.mark.parametrize('complex_specgrams', [
+    torch.randn(1, 2, 1025, 400, 2),
+    torch.randn(1, 1025, 400, 2)
+])
+@pytest.mark.parametrize('hop_length', [256])
+def test_phase_vocoder(complex_specgrams, rate, hop_length):
+
+    class use_double_precision:
```
cpuhrsch: This is neat, but "get_default_type" is only used once, so it feels like overkill, instead of passing it in via torch.float64 and adding a comment.

torchaudio/functional.py
cpuhrsch: What's the advantage of this over torch.stft? We're in the process of adding istft which will be build around being a sane inverse of torch.stft. Having torchaudio.stft and torch.stft will be confusing.

test/test_transforms.py
```diff
@@ -305,5 +305,120 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
         _test_librosa_consistency_helper(**kwargs3)
 
 
+import pytest
```
jamarshon: Can we move these imports to the top of the file?

cpuhrsch:
Complex number support is something we should revisit separately maybe. torch has a [complex dtype](https://github.com/pytorch/pytorch/search?q=complex&unscoped_q=complex), but no support for the used ops. If we treat complex as a dtype we should implicitly dispatch on that and use it instead of doing shape-based inference.

This comes down to a discussion around standards.

test/test_transforms.py
```diff
@@ -305,5 +305,120 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
...
+@pytest.mark.parametrize('waveform', [
+    (torch.randn(1, 100000)),
+    (torch.randn(1, 2, 100000)),
+    pytest.param(torch.randn(1, 100), marks=xfail(raises=RuntimeError)),
+])
+@pytest.mark.parametrize('pad_mode', [
+    # 'constant',
+    'reflect',
+])
+def test_stft(waveform, fft_length, hop_length, pad_mode):
```
jamarshon: Can we have test guards (e.g
```
@unittest.skipIf(not IMPORT_LIBROSA...
```

cpuhrsch: stft as a pytorch operator should be properly tested.

test/test_transforms.py
```diff
@@ -305,5 +305,120 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
...
+    expected_size = list(waveform.size()[:-1])
+    expected_size += [fft_length // 2 + 1, test.common_utils._num_stft_bins(
+        waveform.size(-1), fft_length, hop_length, pad), 2]
+    assert complex_spec.dim() == waveform.dim() + 2
+    assert complex_spec.size() == torch.Size(expected_size)
+
+    # == Test values
+    fft_config = dict(n_fft=fft_length, hop_length=hop_length, pad_mode=pad_mode)
+    # note that librosa *automatically* pad with fft_length // 2.
+    expected_complex_spec = np.apply_along_axis(librosa.stft, -1,
```
jamarshon: could we not import numpy? like have a for loop instead of np.apply_along_axis, torch.allclose for np.allclose, math.pi for np.pi, and math.ceil for np.ceil

test/test_transforms.py
jamarshon: pytorch tests their own method so it would be interesting to see what we need to test on our end

https://github.com/pytorch/pytorch/blob/042da2171e95ae5cf39452ae17a5000a5a24eb85/test/test_torch.py#L6374-L6393

cpuhrsch:
cc @ezyang 

We need to revisit our policy around complex numbers. it doesn't make sense to have torchaudio specific complex number support, since complex numbers are used across all kinds of domains.

Other than that, implementing complex numbers via an inner dimension of 2 can lead to subtle errors.

For example, let's say we standardize (most likely) on channel x time. We can then have a real signal from an audio file of say 2 channels and 1000 samples. So, if we align with channel x time, we get a Tensor from torchaudio.load of 2 x 1000. If you pass this into torchaudio.stft you get 2 x 1000 x 2.

Now if someone loads an audio file from somewhere else or isn't familiar with this they might pass 1000 x 2 into stft, which then is interpreted as a complex signal of length 1000 instead of what was intended, which is a signal with 2 channels and 1000 samples.

So, if we emulate complex numbers using two real numbers we should imo create a "ComplexTensor" extension, which yields support via two real tensors hidden from users. In above case that'd then yield 1000 x 2 being interpreted as 1000 channels of 2 samples, which, while still wrong, is at least consistent and could more easily raise a warning.

But then again, this kind of extension might not be what we want given that we already have a complex dtype. There it might make more sense to write a torchaudio c++ operator that dispatches on this dtype for things like magnitude etc.

soumith:
I think we should start with the convention of `trailing dimension size = 2`, and then graduate to a Complex boxed Tensor

test/test_transforms.py
keunwoochoi: that's valid, but also we can think of some relaxation as long as we're import `librosa`, which will definitely rely on `numpy`.

keunwoochoi:
> we should start with the convention of `trailing dimension size = 2`

+1, at least for this PR. I feel like having our own complex data type is a bit of stretch when `torch.stft` doesn't use it.

torchaudio/functional.py
keunwoochoi: AFAIR the issue was that we didn't like the default names/values/behaviors of `torch.stft`. I think this also boils down to to what extent we'd like these functionals to be less hacky in the phase 1.

cpuhrsch:
@soumith - we run the risk of introducing and releasing a standard we need to revert later on. We already have the complex dtype and we also already have C++ extensions for torchaudio.

cpuhrsch:
The currently accepted solution to complex number support is the complex dtype, so we should run with that. Otherwise we're introducing a domain specific unicorn solution with domain specific interfaces of generic operations such as stft and its inverse. I propose we instead look into adding a complex flag (or such) to the core stft.

I want to get a sense of urgency and eventual operator coverage of complex number support @keunwoochoi on your end, could you advice?

keunwoochoi:
@cpuhrsch Alright, the main reason we had this less-neat operation was because of the way current `torch.stft` represent the complex number. Especially when we're implementing `istft` as well here, I'm sure you have a plan to cover these all with consistency :)

I don't think we'd need anything super specific. We definitely need `angle`, `norm`. Besides, the very basic stuff should be fine? `+`, `-`, `*`, `/` (between complex and complex-to-real) would be already in your list I assume and some of them are used in applying phase vocoder in our implementation.

ezyang:
cc @Roger-luo

torchaudio/functional.py
cpuhrsch: @keunwoochoi - we'll need to revisit the discussion around the default names, values and behavior. We cannot add a torchaudio.stft, it'll be very confusing and hacks around an issue that's related to the core of pytorch and we have the power resolve over there.

cpuhrsch:
@keunwoochoi - @ezyang and I just spoke offline about this issue. Depending on your timeline there are two proposals here

1) If we need complex number support quickly, we'll go with the implicit PyTorch convention of returning a Tensor with last dimension 2. torch.fft does this and so does torch.stft. This will happen with the expectation that it'll be changed in the future.

2) If there is time, we'd want to introduce a ComplexTensor wrapper, which links into our wider story around wrappers. It'll be used in core and also change the type torch.stft will return it.

On torchaudio.stft: In general there cannot be a torchaudio.stft if there is already a torch.stft and even [torchaudio.istft](https://github.com/pytorch/audio/pull/130) will be moved into torch.istft once it's great, because we also have a torch.ifft for torch.fft.

=======

In summary, I suggest we run with the size (... x 2) convention for complex number support, if we need to have this soon, and not add duplicate functions (stft) to torchaudio to work around an issue specific to torch.

torchaudio/functional.py
ksanjeevan: @keunwoochoi maybe initially, but at this point the default values are the same as `torch.stft`. I think it was more about the names and `torchaudio.stft` being multichannel.

ksanjeevan:
> On torchaudio.stft: In general there cannot be a torchaudio.stft if there is already a torch.stft and even torchaudio.istft will be moved into torch.istft once it's great, because we also have a torch.ifft for torch.fft.

Makes sense, we still keep it and change the name to `multichannel_stft` or something.

torchaudio/functional.py
cpuhrsch: I see there is https://github.com/keunwoochoi/torchaudio-contrib/issues/36 in torchaudio-contrib around this. The main thing I can tell from your code is that you want batch support for the input (waveforms) and the hann_window as a standard if none is specified. You're also implying the preference of input to be named "waveforms" instead of using [input in torch core](https://pytorch.org/docs/master/torch.html#torch.stft).

You're also implying the standard of having "channel" first, which is something @jamarshon is looking into as the framework-wide standard. We talked about this today and it seems the most sane, since it's ML specific (and this is an ML library) and since its consistent with the standards used in common operations [such as conv1d](https://pytorch.org/docs/master/nn.html#torch.nn.Conv1d).

cpuhrsch: cc @ksanjeevan

cpuhrsch:
@ksanjeevan - I'd say "batched_stft" or we just add these batching semantics to the core.

keunwoochoi:
> Depending on your timeline

@cpuhrsch  I think we were rather relying on your timeline for finishing Phase 1. The plan of phase 1 and 2 (freezing, then making breaking changes) is still valid and this PR is in Phase 1, right? Then I'd be inclined to the option 1.

Roger-luo:
I think in short term you could just use two real tensor as complex tensor in Python or C++ API. And if you want a bit better support I think I once made element-wise thing like `+`, `-` work in torch-complex extension for CPU, but it doesn't work due to the change of C++ part and cause some issue to CUDA. The ATen library has a generic implementation of broadcast I think, which should make this not hard to implement.

cpuhrsch:
@keunwoochoi - yes indeed, we're currently in the process of pushing the binaries and have a [release branch](https://github.com/pytorch/audio/tree/v0.2.0) cut.

Option 1 sounds reasonable.

cpuhrsch:
@ksanjeevan - Is there anything blocking you at this point for this PR?

ksanjeevan:
@cpuhrsch don't think so, there are a couple discussions that seem to be semi-ongoing but I think I've addressed most changes.

test/test_transforms.py
```diff
@@ -15,6 +15,11 @@
 if IMPORT_SCIPY:
     import scipy
 
+import pytest
+import numpy as np
```
jamarshon: ``import numpy as np`` should be under the ``if IMPORT_LIBROSA`` right?

ksanjeevan: Yes, I'll fix that and add a skipif for the tests

jamarshon:
@ksanjeevan 
Yes that sounds good. Could you also do the following:
-merge master into this PR (so we can see that it merges cleanly and to get the status of the CI)
-add pytest to requirements.txt
-change the CI e.g ``python -m pytest -v $FILE`` in [test_script.sh](https://github.com/pytorch/audio/blob/master/build_tools/travis/test_script.sh#L21)
-resolve the comments ([here](https://github.com/pytorch/audio/pull/131#discussion_r300778704) and [here](https://github.com/pytorch/audio/pull/131/files#r304030689))

I think it should look okay after that 😄

torchaudio/functional.py
```diff
@@ -449,3 +449,176 @@ def mu_law_expanding(x_mu, qc):
     x = ((x_mu) / mu) * 2 - 1.
     x = torch.sign(x) * (torch.exp(torch.abs(x) * torch.log1p(mu)) - 1.) / mu
     return x
+
+
+def batched_stft(waveforms, fft_length, hop_length=None, win_length=None, window=None,
```
cpuhrsch: I think we can rename this back to stft, since [torchaudio's istft also supports batching](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L151). Eventually this should be lowered down to the core of PyTorch, but we might also find [a more general solution to batching](https://github.com/pytorch/pytorch/issues/22169) beforehand.

Until then, for consistency, let's rename this back to stft and create documentation (flags etc.) in compliance with our istft.

build_tools/travis/test_script.sh
```diff
@@ -23,6 +23,10 @@ run_tests() {
     EXIT_STATUS="$(($EXIT_STATUS+STATUS))"
   done
 
+  python -m pytest -v -k "test_transforms"
```
jamarshon: I meant run every test with pytest. the for loop above iterates over every test with unittest currently, just modify it to pytest

ksanjeevan: ooh gotcha, thought you meant just for the tests we introduced. I'll change it.

test/common_utils.py
```diff
@@ -48,3 +47,7 @@ def random_int_tensor(seed, size, low=0, high=2 ** 32, a=22695477, c=1, m=2 ** 3
     """ Same as random_float_tensor but integers between [low, high)
     """
     return torch.floor(random_float_tensor(seed, size, a, c, m) * (high - low)) + low
+
+
+def _num_stft_bins(signal_len, fft_len, hop_length, pad):
```
cpuhrsch: I'd move this into functional.py, especially if it starts with an underscore which indicates that its a local convenience function.

ksanjeevan: It's only used for tests though, should it be in `test_transforms.py` then? Or rather, I should move all the contrib functional tests (and this helper) to `test_functional.py` right?

test/test_functional.py
```diff
@@ -183,5 +193,119 @@ def test_istft_of_sine(self):
...
+@pytest.mark.parametrize('rate', [0.5, 1.01, 1.3])
+@pytest.mark.parametrize('complex_specgrams', [
+    torch.randn(1, 2, 1025, 400, 2),
+    torch.randn(1, 1025, 400, 2)
+])
+@pytest.mark.parametrize('hop_length', [256])
+@unittest.skipIf(not IMPORT_LIBROSA, 'Librosa is not available')
+def test_phase_vocoder(complex_specgrams, rate, hop_length):
+
+    class use_double_precision:
```
cpuhrsch: What's the advantage of this? Do we still plan to remove it?

ksanjeevan: It's needed to compare the results of the `phase_vocoder` against `librosa`. Because it uses a cumulative sum, small numerical errors when not using double precision make them accumulate and differ. Some discussion on this [here](https://github.com/keunwoochoi/torchaudio-contrib/pull/65#discussion_r291855782).

cpuhrsch: Can this only done using the "with" construct or can the dtype be passed explicitly?

ksanjeevan: I think so, because `F.phase_vocoder` create a tensor [`time_steps`](https://github.com/pytorch/audio/blob/c8596a683f1ccdd53e8e1c31fa4c86974d81bba4/torchaudio/functional.py#L573) and that needs to be a `double()` if we want to compare it with `librosa`. Doing the `with` ensures that auxiliary tensors created by the phase vocoder have the correct precision.

cpuhrsch: If that is the fear, then I think we should change phase_vocoder overall to respect the type of the input. Because, for testing this might be fine, but what if a user wants to do this herself and isn't aware of the need to create this context? Wanting to use doubles is still a legitimate feature and also being invariant under datatypes makes it easier to later on add float16 etc.

torchaudio/functional.py
cpuhrsch: We talked about this separately again today and the main thing this wrapper does is add batching support.

Batching is a big topic overall. I'd prefer if we would add batching support in a separate PR and, if we add it, do so globally, since none of the other transforms have batching support either.

Once you add batching support user will probably want to have support for masks etc., so then we should think it about it more principled or maybe even wait for something like NestedTensor.

This also doesn't block people from using batching if they're aware of the reshape trick.

So, could I please ask you to remove this from the PR and then send again later after we talked more about batching (and after the standardization)?

test/test_functional.py
ksanjeevan: Ok I'll make it so they have the same dtype as the input tensor and remove the with.

torchaudio/functional.py
cpuhrsch: cc @ksanjeevan

PR #135, opened 2019-07-10
ISTFT

jamarshon:
Inverse STFT (https://github.com/pytorch/audio/pull/130). Can't reopen previous one so here it is again

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -1,6 +1,7 @@
 import os
 from shutil import copytree
 import tempfile
+import torch
 
 
 TEST_DIR_PATH = os.path.dirname(os.path.realpath(__file__))
@@ -16,3 +17,28 @@ def create_temp_assets_dir():
     copytree(os.path.join(TEST_DIR_PATH, "assets"),
              os.path.join(tmp_dir.name, "assets"))
     return tmp_dir.name, tmp_dir
+
+
+class RandomTensorGenerator:
+    """ Generates random tensors given a seed and size
+    https://en.wikipedia.org/wiki/Linear_congruential_generator
+    X_{n + 1} = (a * X_n + c) % m
+    Using Borland C/C++ values
+    """
+    a = 22695477
+    c = 1
+    m = 2 ** 32
+
+    def __init__(self, seed, size):
+        # seed is an int and size is a tuple of ints
+        torch.manual_seed(seed)
+        self.x_n = torch.randint(low=0, high=self.m, size=size)
+
+    def rand_int_tensor(self, rand_min=0, rand_max=2 ** 32):
+        # returns an integer tensor between [rand_min, rand_max)
+        return torch.floor(self.rand_float_tensor() * (rand_max - rand_min)) + rand_min
+
+    def rand_float_tensor(self):
+        # returns a tensor between [0, 1)
+        self.x_n = (self.a * self.x_n + self.c) % self.m
+        return self.x_n.float() / self.m
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -37,7 +37,7 @@ def random_float_tensor(seed, size, a=22695477, c=1, m=2 ** 32):
     for s in size:
         num_elements *= s
 
-    arr = [seed % m]
+    arr = [(a * seed + c) % m]
     for i in range(num_elements - 1):
         arr.append((a * arr[i] + c) % m)
 
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -31,8 +31,14 @@ class RandomTensorGenerator:
 
     def __init__(self, seed, size):
         # seed is an int and size is a tuple of ints
-        torch.manual_seed(seed)
-        self.x_n = torch.randint(low=0, high=self.m, size=size)
+        num_elements = 1
+        for s in size:
+            num_elements *= s
+
+        arr = [seed]
+        for i in range(num_elements - 1):
+            arr.append((self.a * arr[i] + self.c) % self.m)
+        self.x_n = torch.tensor(arr, dtype=torch.int64).view(size)
 
     def rand_int_tensor(self, rand_min=0, rand_max=2 ** 32):
         # returns an integer tensor between [rand_min, rand_max)
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -19,32 +19,32 @@ def create_temp_assets_dir():
     return tmp_dir.name, tmp_dir
 
 
-class RandomTensorGenerator:
+def random_float_tensor(seed, size, a=22695477, c=1, m=2 ** 32):
     """ Generates random tensors given a seed and size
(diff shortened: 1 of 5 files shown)


test/test_functional.py
```diff
@@ -0,0 +1,202 @@
...
+class TestFunctional(unittest.TestCase):
+    data_sizes = [(2, 20), (3, 15)]
+    number_of_trials = 100
+    stored_rand_data = []
+    fixed_precision = int(1e10)
+
+    def setUp(self):
+        # we want to make sure that the random values are reproducible
+        self.stored_rand_data.clear()
+        torch.manual_seed(0)
```
cpuhrsch: We still need to resolve this

test/test_functional.py
```diff
@@ -0,0 +1,202 @@
...
+
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=atol, rtol=rtol))
+
+    def _test_istft_is_inverse_of_stft(self, kwargs):
+        # generates a random sound signal for each tril and then does the stft/istft
+        # operation to check whether we can reconstruct signal
+        for i in range(len(self.data_sizes)):
+            for _ in range(self.number_of_trials):
+                sound = self._get_random_tensor(i)
```
cpuhrsch: Each trial has the same data, right?

test/test_functional.py
```diff
@@ -0,0 +1,202 @@
...
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    data_sizes = [(2, 20), (3, 15)]
+    number_of_trials = 100
+    stored_rand_data = []
+    fixed_precision = int(1e10)
```
cpuhrsch: This should be a prime number

cpuhrsch: If used in the random number generator

test/test_functional.py
```diff
@@ -0,0 +1,202 @@
...
+        self.stored_rand_data.clear()
+        torch.manual_seed(0)
+        for data_size in self.data_sizes:
+            rand_data1 = torch.randint(low=-self.fixed_precision, high=self.fixed_precision, size=data_size)
+            rand_data2 = torch.randint(low=-self.fixed_precision, high=self.fixed_precision, size=data_size)
+            self.stored_rand_data.append([rand_data1, rand_data2])
+
+    def _get_random_tensor(self, i):
+        # gets a random tensor of size data_sizes[i]. adds to previous tensors and then mods it.
+        rand_data = self.stored_rand_data[i]
```
cpuhrsch: You don't need to store data for this. Take the first value (i), fix some constant to add (say c) and then some prime number (say p) to mod by.

So for element 0 you have i, for element 1 you have (i + c) mod  p for element 2 you have (element1 + c) mod p

test/test_functional.py
```diff
@@ -0,0 +1,202 @@
+import math
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    data_sizes = [(2, 20), (3, 15)]
```
cpuhrsch: We could add a few more and also degenerate cases like (1, 1) or (1, 0) etc.

test/test_functional.py
```diff
@@ -9,25 +9,17 @@
...
     def setUp(self):
-        # we want to make sure that the random values are reproducible
-        self.stored_rand_data.clear()
-        torch.manual_seed(0)
+        self.random_tensor_generators.clear()
         for data_size in self.data_sizes:
-            rand_data1 = torch.randint(low=-self.fixed_precision, high=self.fixed_precision, size=data_size)
-            rand_data2 = torch.randint(low=-self.fixed_precision, high=self.fixed_precision, size=data_size)
-            self.stored_rand_data.append([rand_data1, rand_data2])
+            rtg = test.common_utils.RandomTensorGenerator(seed=0, size=data_size)
```
cpuhrsch: Why do you create them here as opposed to on demand?

test/test_functional.py
```diff
@@ -0,0 +1,184 @@
+import math
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    data_sizes = [(2, 20), (3, 15)]
```
cpuhrsch: Please add more shapes here

test/test_functional.py
```diff
@@ -0,0 +1,184 @@
...
+        self.assertTrue(sound.shape == estimate.shape, (sound.shape, estimate.shape))
+        self.assertTrue(torch.allclose(sound, estimate, atol=atol, rtol=rtol))
+
+    def _test_istft_is_inverse_of_stft(self, kwargs):
+        # generates a random sound signal for each tril and then does the stft/istft
+        # operation to check whether we can reconstruct signal
+        for data_size in self.data_sizes:
+            rtg = test.common_utils.RandomTensorGenerator(seed=0, size=data_size)
+            for _ in range(self.number_of_trials):
+                sound = rtg.rand_float_tensor()
```
cpuhrsch: This suffers from the same issue has using a regular number generator, which is that its stateful and the results dependent on the number of previous calls to rand_float_tensor

vincentqb:
What are the differences with #130?

test/common_utils.py
```diff
@@ -16,3 +17,34 @@ def create_temp_assets_dir():
...
+    c = 1
+    m = 2 ** 32
+
+    def __init__(self, seed, size):
+        # seed is an int and size is a tuple of ints
+        num_elements = 1
+        for s in size:
+            num_elements *= s
+
+        arr = [seed]
```
cpuhrsch: seed is an int64_t if its a python integer, so it can't be the first number if you always mod by m for the other values.

test/common_utils.py
```diff
@@ -16,3 +17,34 @@ def create_temp_assets_dir():
...
+        size (Tuple[int]): the size of the output tensor
+        a (int): the multiplier constant to the generator
+        c (int): the additive constant to the generator
+        m (int): the modulus constant to the generator
+    """
+    num_elements = 1
+    for s in size:
+        num_elements *= s
+
+    arr = [seed % m]
```
cpuhrsch: this will give you the same return data for a whole range of seeds as opposed to using it within the first iteration of data generation.

cpuhrsch: it can't be the first element of the generated data.

test/test_functional.py
```diff
@@ -0,0 +1,183 @@
+import math
+
+import torch
+import torchaudio
+import unittest
+import test.common_utils
+
+
+class TestFunctional(unittest.TestCase):
+    data_sizes = [(2, 20), (3, 15), (4, 10)]
```
cpuhrsch: Please exercise (2, 0) or assert that an exception is raised.

Issue #143, opened 2019-07-16
Error when loading VCTK


vincentqb:
I'm getting an error from SOX when loading VCTK.

```
>>> import torchaudio
>>> torchaudio.datasets.VCTK("~/VCTK", download=True)
Downloading http://homepages.inf.ed.ac.uk/jyamagis/release/VCTK-Corpus.tar.gz
Processing...
Found 44257 audio files and 44085 utterences
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/private/home/vincentqb/anaconda3/envs/audiotutorial/lib/python3.7/site-packages/torchaudio-0.2-py3.7-linux-x86_64.egg/torchaudio/datasets/vctk.py", line 109, in __init__
    self.download()
  File "/private/home/vincentqb/anaconda3/envs/audiotutorial/lib/python3.7/site-packages/torchaudio-0.2-py3.7-linux-x86_64.egg/torchaudio/datasets/vctk.py", line 220, in download
    sig = read_audio(f, downsample=self.downsample)[0]
  File "/private/home/vincentqb/anaconda3/envs/audiotutorial/lib/python3.7/site-packages/torchaudio-0.2-py3.7-linux-x86_64.egg/torchaudio/datasets/vctk.py", line 39, in read_audio
    E = torchaudio.sox_effects.SoxEffects()
AttributeError: module 'torchaudio.sox_effects' has no attribute 'SoxEffects'
```

rfalcon100:
Is this fix present in the stable version? I just installed the latest release, and I am still getting the same error:

```console
>>> import torchaudio
>>> import torchaudio.datasets as dsets
>>> b = dsets.VCTK('~/datasets/VCTK/', download=True)
Downloading http://homepages.inf.ed.ac.uk/jyamagis/release/VCTK-Corpus.tar.gz
Processing...
Found 44257 audio files and 44085 utterences
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/u/77/falconr1/unix/.conda/envs/audio/lib/python3.7/site-packages/torchaudio/datasets/vctk.py", line 109, in __init__
    self.download()
  File "/u/77/falconr1/unix/.conda/envs/audio/lib/python3.7/site-packages/torchaudio/datasets/vctk.py", line 220, in download
    sig = read_audio(f, downsample=self.downsample)[0]
[...]

PR #145, opened 2019-07-16
Fix vctk.read_audio #143

jamarshon:
https://github.com/pytorch/audio/issues/143
A quick fix to vctk.read_audio. Added a few tests to make sure that it works.


jamarshon:
Yes specifically the `test_read_audio` function tests `read_audio` which was not working previously

PR #148, opened 2019-07-17
Cleanup torch_sox.h/.cpp

jamarshon:
-Ran clangtidy on torch_sox.h/.cpp
-Add the header "torch_sox.h" to torch_sox.cpp. Previously torch_sox.h was not used by any files
-Reorder functions in torch_sox.h/.cpp to match the order they are defined in PYBIND11_MODULE

### DIFF
--- a/torchaudio/torch_sox.cpp
+++ b/torchaudio/torch_sox.cpp
@@ -1,14 +1,15 @@
-#include <torch/extension.h>
-
-#include <sox.h>
+#include "torch_sox.h"
 
 #include <algorithm>
 #include <cstdint>
 #include <stdexcept>
+#include <string>
+#include <tuple>
 #include <vector>
 
-namespace torch {
-namespace audio {
+#include <sox.h>
+#include <torch/extension.h>
+
 namespace {
 /// Helper struct to safely close the sox_format_t descriptor.
 struct SoxDescriptor {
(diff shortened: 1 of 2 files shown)


cpuhrsch:
I'm generally in favor of code hygiene, but you're not adding any functionality and are significantly changing git blame. Typically we only format new code or code we touch and not all files at once. I know that it's cleaner, but if it's just for the sake of formatting then that stands in contrast of having code where the development is easily traced.

jamarshon:
@cpuhrsch most of the git [blame](https://github.com/pytorch/audio/blame/adcbc8509662e770c01733b58a39a12a692e5de7/torchaudio/torch_sox.cpp) is maintained (compared to the [original](https://github.com/pytorch/audio/blame/master/torchaudio/torch_sox.cpp)). 

also I think most developer use git history instead of blame as it is more fine tuned so it should be fine

cpuhrsch:
@jamarshon by history you mean log? blame is a lot more detailed than that. But even aside from that, if you now look at a range of commits the majority of changes will be dominated by this PR even though this PR doesn't fix any bugs nor adds any features but merely deals with aesthetics.

jamarshon:
By history I mean [this](https://github.com/pytorch/audio/commits/master/torchaudio/torch_sox.h). I don't think preserving git blame should be a reason not to fix code (i.e. refactoring, etc.)

cpuhrsch:
@jamarshon - It's used commonly and important to determine ownership and resolve bugs. Cosmetics will always come secondary to something like this.

PR #149, opened 2019-07-17
CI fix

jamarshon:
Travis CI using pytorch-nightly has been annoying. Occasionally pytorch/audio master would be passing and then the next day the same test would be fail due to changes in pytorch.
The last passing installs for our python 3.5/3.6 tests were:

```
pytorch-nightly-cpu-1.1.0.dev20190516|      py3.5_cpu_0
pytorch-nightly-cpu-1.2.0.dev20190716|      py3.6_cpu_0
```

so want to fix it on 1.1.0.dev20190516

### DIFF
--- a/build_tools/travis/install.sh
+++ b/build_tools/travis/install.sh
@@ -53,7 +53,7 @@ pip install -r requirements.txt
  # Install the following only if running tests
 if [[ "$SKIP_TESTS" != "true" ]]; then
     # PyTorch (nightly as 1.1 does not have Optional for type annotations)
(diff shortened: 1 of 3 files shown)


cpuhrsch:
This is dangerous because we're now developing against an arbitrary version. If nightlies keep on breaking our build we should create an issue around that work on the process under which we pick a nightly. Also, its important feedback for pytorch to have us fail on nightlies, since we provide additional coverage. If anything we should maybe look into linking these errors back to the main repo.

PR #150, opened 2019-07-17
Adding initial tutorial

vincentqb:
Initial tutorial with an overview of dataset loading and transformers. Created using the following [notebook](https://github.com/vincentqb/audio-tutorial/blob/master/Tutorial.ipynb). See [demo](https://vincentqb.github.io/audio/tutorial.html).

Things that may or may not make it in this PR:

- [x]  Add as a follow-up a part on migrating to pytorch from kaldi, and kaldi compliance support.
- [ ]  Update as soon as the transformers have been standardized, see [PR](https://github.com/pytorch/audio/pull/152). For instance, we expect LC2CL to no longer be needed.
- [ ] Confirm location to put tutorial
- [ ] Link to a code/notebook/lab, as done [here](https://pytorch.org/tutorials/beginner/audio_classifier_tutorial.html?highlight=audio).
- [ ] Find a better solution to correct the path to images than adding a command in Makefile.

In a later iteration, we should demo a neural network implementation, maybe inspired by [a prior version](https://pytorch.org/tutorials/beginner/audio_classifier_tutorial.html?highlight=audio) (which no longer appears in the left hand side menu but is still searchable).


vincentqb:
Since there is an ongoing discussion about dataset abstraction, I changed to load one of our test file, converted from mp3 to wav.

vincentqb:
Discussion with @jamarshon and @cpuhrsch.

* Let's expand the introduction, and talk about how the library in general is useful and leverages GPUs.
* In the kaldi section, let's showcase fbank and spectrogram. Some of the [others](https://github.com/pytorch/audio/blob/master/torchaudio/compliance/kaldi.py) could be made private.
* torchaudio is the official spelling 😄

vincentqb:
I talked with @nairbv and he suggested putting the tutorials in https://github.com/pytorch/tutorials

PR #151, opened 2019-07-17
Mention how to install sox through Anaconda in README

vincentqb:
Let's mention how to install sox through Anaconda in README, #133.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,10 @@ Linux (Ubuntu):
 ```bash
 sudo apt-get install sox libsox-dev libsox-fmt-all
 ```
+Anaconda
+```bash
+conda install -c conda-forge sox
+```
 
 Installation
 ------------


faroit:
As I understand, libsox is currently not available through conda-forge thus the installation of torchaudio would still fail on systems without root access. 

Can we make progress on this? In the meantime I would suggest to make it more clear in the docs.

soumith:
@faroit I personally added libsox to conda-forge.

The package is called sox, not libsox, but it includes libsox: https://github.com/conda-forge/sox-feedstock

faroit:
@soumith sorry that I have to go back on this. I tried again in a clean (docker) environment.

Again, the use case here is that a user has only access to conda and g++/gcc but cannot install sox through apt.

Using this [gist here](https://gist.github.com/faroit/10fbddb0e9de42d130b3327a37c3a869), I get the following error:

```
cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++                                   
torchaudio/torch_sox.cpp:3:17: fatal error: sox.h: No such file or directory                                                      
 #include <sox.h>
                 ^
compilation terminated.
error: command 'gcc' failed with exit status 1
```

indicating that `sox.h` cannot be found (even though it correctly lands in the conda env include paths). Would users need to adjust the search path in this case?

soumith:
@faroit I'm sorry I dont understand. Can you not just do: `conda install sox -c conda-forge` as specified now in the README?

faroit:
Yes, I did that (see gist) but setup.py doesn't find the conda include path for some reason.

faroit:
Shall we use the `IS_CONDA` flag in this case?

soumith:
oh i see what you mean now. Yea does the `IS_CONDA` flag work for you? If so, we can make it default, i.e. including those dirs in default case as well

faroit:
I checked back: using `IS_CONDA` and torch 1.2 nighly allows to compile on my non-root environment. However it couldn't manage to import because – I guess – the conda gcc toolchain is not ABI compabtible with pytorch. For now this would really make it tricky to install torchaudio using just conda...

Issue #110, opened 2019-05-21
Merging plan from torchaudio-contrib


keunwoochoi:
Hi all,
I think it's good timing to discuss a potential merging plan from [torchaudio-contrib](https://github.com/keunwoochoi/torchaudio-contrib) to here, especially because there's going to be new features and changes by @jamarshon @cpuhrsch. 

## Main idea
A lot of things are well summarized in https://github.com/keunwoochoi/torchaudio-contrib. In short, we wanted to re-design torch-based audio processing so that

- things can be `Layers`, which are based on corresponding `Functionals`
- names for layers and arguments are carefully chosen
- all work for multi-channel
- complex numbers are supported when it makes sense (e.g., STFTs)

## Review - layers
. torchaudio-contrib already covers lots of functions that `transform.py` is covering now, but not all of them. And that's why I feel like it's time to discuss this here.
Let me list the classes in `transform.py` one by one with some notes.

- `class Compose`: we wouldn't need it because once things are based on `Layers` people can simply build a `nn.Sequential()`. 
- `class Scale`: It does `16int` --> `float`. I think we need to deprecate this because if we really need this, it should be with a more intuitive and precise name, and probably should support other conversions as well. 
- `class PadTrim`: I don't actually know why we need it exactly, would love to hear about this!
- `class DownmixMono`: I like it, we don't have it yet but would like to implement one. But we also consider having a [time-frequency representation-based downmix (energy-preserving operation)](https://github.com/keunwoochoi/torchaudio-contrib/pull/45) (@faroit). I'm open for discussion.
- `class LC2CL`: So far, torchaudio-contrib code hasn't considered `channel-first` tensors. If it's a thing, we'd i) update our code to make them compatible and ii) have the same or a similar class to this. But, ..do we really need this?
- `class Spectrogram`: we have it in torchaudio-contrib. On top of this, we also have `STFT` layer which outputs complex representations (same as `torch.stft` since we're wrapping it).
- `class MelScale`: we have it and would like to suggest to change the name to something more general. We named it `class MelFilterbank`, assuming there can be other types of filterbanks, too. It also supports `htk` and non-`htk` mel filterbanks.
- `class SpectrogramToDB`: we would like to propose a more general approach -- `class AmplitudeToDb(ref=1.0, amin=1e-7)` and `class DbToAmplitude(ref=1.0)`, because decibel-scaling is about changing it's unit, not the core content of the input. 
- `class MFCC`: we currently don't have it but would like to add one.
[...]

PR #152, opened 2019-07-18
[BC] Standardization of Transforms/Functionals

jamarshon:
## Removal of Legacy Transforms
-Compose: From https://github.com/pytorch/audio/issues/110, "we wouldn't need it because once things are based on Layers people can simply build a nn.Sequential()."
-Aliases and deprecated modules: SPECTROGRAM, F2M, MEL If we are going to start breaking things in this new release, then it makes sense to do as much as possible so that the future stable versions do not have technical debt
-BLC2CBL: there is no input/output to this library that uses BLC or CBL format (going to keep LC2CL mainly because all the inputs are (c, n) meaning it would be convenient for users to switch this format).

## Channel x Time
The input (some outputs) of all transforms/functions would be (c, n) e.g Spectrogram, MFCC, MelSpectrogram, Resample, etc. This is to be consistent with PyTorch which has channel followed by the number of samples.
This means removing the channel dimension flags that some transforms were using which could accept either (n, c) or (c, n).

## Channel x Frequency x Time
The output of stft is (c, f, t, 2) meaning for each channel, the columns are the fft of a certain window so as we travel horizontally we can see each column (fft) change over time. This matches the output of librosa so we no longer need to transpose in our test comparisions with spectrogram, melscale, and mfcc. Previously it was time x frequency.
image

## DCT and Filterbank matrices
Going to keep them as they are. These matrices should be shaped so that they can be multiplied without a transpose e.g for filterbank (..., n_freqs) x (n_freqs, n_mels) meaning each column is a filterbank and DCT (..., n_mels) x (n_mels, n_mfcc) meaning each column is a cos function

## Misc
Going to keep scale, padtrim, downmixmono as while they are simple, they could be operations that are performed often so it would be convenient to have the code to reduce amount of code the user has to write as well as provides documentation as the name of the transforms/functions communicates the intent clearly

### DIFF
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -2,6 +2,7 @@ import math
 
 import torch
 import torchaudio
+import torchaudio.functional as F
 import pytest
 import unittest
 import test.common_utils
@@ -212,8 +213,7 @@ def test_phase_vocoder(complex_specgrams, rate, hop_length):
     complex_specgrams = complex_specgrams.type(torch.float64)
     phase_advance = torch.linspace(0, np.pi * hop_length, complex_specgrams.shape[-3], dtype=torch.float64)[..., None]
 
-    complex_specgrams_stretch = torchaudio.functional.phase_vocoder(
-        complex_specgrams, rate=rate, phase_advance=phase_advance)
+    complex_specgrams_stretch = F.phase_vocoder(complex_specgrams, rate=rate, phase_advance=phase_advance)
 
     # == Test shape
     expected_size = list(complex_specgrams.size())
@@ -244,7 +244,7 @@ def test_phase_vocoder(complex_specgrams, rate, hop_length):
 @pytest.mark.parametrize('power', [1, 2, 0.7])
 def test_complex_norm(complex_tensor, power):
     expected_norm_tensor = complex_tensor.pow(2).sum(-1).pow(power / 2)
-    norm_tensor = torchaudio.functional.complex_norm(complex_tensor, power)
+    norm_tensor = F.complex_norm(complex_tensor, power)
 
     assert torch.allclose(expected_norm_tensor, norm_tensor, atol=1e-5)
 
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -2,6 +2,7 @@ import math
 
 import torch
 import torchaudio
+import pytest
 import unittest
 import test.common_utils
 
@@ -11,8 +12,6 @@ if IMPORT_LIBROSA:
     import numpy as np
     import librosa
 
-import pytest
-import torchaudio.functional as F
 xfail = pytest.mark.xfail
 
 
@@ -197,54 +196,6 @@ def _num_stft_bins(signal_len, fft_len, hop_length, pad):
     return (signal_len + 2 * pad - fft_len + hop_length) // hop_length
 
 
-@pytest.mark.parametrize('fft_length', [512])
-@pytest.mark.parametrize('hop_length', [256])
-@pytest.mark.parametrize('waveform', [
-    (torch.randn(1, 100000)),
-    (torch.randn(1, 2, 100000)),
-    pytest.param(torch.randn(1, 100), marks=xfail(raises=RuntimeError)),
-])
-@pytest.mark.parametrize('pad_mode', [
-    # 'constant',
-    'reflect',
-])
-@unittest.skipIf(not IMPORT_LIBROSA, 'Librosa is not available')
-def test_stft(waveform, fft_length, hop_length, pad_mode):
-    """
-    Test STFT for multi-channel signals.
-
-    Padding: Value in having padding outside of torch.stft?
-    """
-    pad = fft_length // 2
-    window = torch.hann_window(fft_length)
-    complex_spec = F.stft(waveform,
-                          fft_length=fft_length,
-                          hop_length=hop_length,
-                          window=window,
-                          pad_mode=pad_mode)
-    mag_spec, phase_spec = F.magphase(complex_spec)
-
-    # == Test shape
-    expected_size = list(waveform.size()[:-1])
-    expected_size += [fft_length // 2 + 1, _num_stft_bins(
-        waveform.size(-1), fft_length, hop_length, pad), 2]
-    assert complex_spec.dim() == waveform.dim() + 2
-    assert complex_spec.size() == torch.Size(expected_size)
-
-    # == Test values
-    fft_config = dict(n_fft=fft_length, hop_length=hop_length, pad_mode=pad_mode)
-    # note that librosa *automatically* pad with fft_length // 2.
-    expected_complex_spec = np.apply_along_axis(librosa.stft, -1,
-                                                waveform.numpy(), **fft_config)
-    expected_mag_spec, _ = librosa.magphase(expected_complex_spec)
-    # Convert torch to np.complex
-    complex_spec = complex_spec.numpy()
-    complex_spec = complex_spec[..., 0] + 1j * complex_spec[..., 1]
-
-    assert np.allclose(complex_spec, expected_complex_spec, atol=1e-5)
-    assert np.allclose(mag_spec.numpy(), expected_mag_spec, atol=1e-5)
-
-
 @pytest.mark.parametrize('rate', [0.5, 1.01, 1.3])
 @pytest.mark.parametrize('complex_specgrams', [
     torch.randn(1, 2, 1025, 400, 2),
@@ -261,7 +212,8 @@ def test_phase_vocoder(complex_specgrams, rate, hop_length):
     complex_specgrams = complex_specgrams.type(torch.float64)
     phase_advance = torch.linspace(0, np.pi * hop_length, complex_specgrams.shape[-3], dtype=torch.float64)[..., None]
 
-    complex_specgrams_stretch = F.phase_vocoder(complex_specgrams, rate=rate, phase_advance=phase_advance)
+    complex_specgrams_stretch = torchaudio.functional.phase_vocoder(
+        complex_specgrams, rate=rate, phase_advance=phase_advance)
 
     # == Test shape
     expected_size = list(complex_specgrams.size())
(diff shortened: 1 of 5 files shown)


cpuhrsch:
@keunwoochoi @ksanjeevan please let us know what you think. This is going to resolve a lot of discussion :)

ksanjeevan:
@cpuhrsch on the dimension side all great 👍. There still will be much to discuss on some variable/transform names I guess.

keunwoochoi:
Great! All the dimension configs are aligned with what we've discussed in the `contrib` repo :) 

If the var names were an issue, we reached at [this conclusion](https://github.com/keunwoochoi/torchaudio-contrib/issues/36#issuecomment-502144987) if you wanna borrow some idea @cpuhrsch @jamarshon. They're probably slightly more verbose than usual, and that's what we wanted so that even beginners in either Pytorch or audio processing would get it only by reading the code.

test/test_transforms.py
```diff
@@ -127,43 +98,37 @@ def test_mu_law_companding(self):
         sig_exp = transforms.MuLawExpanding(quantization_channels)(sig_mu)
         self.assertTrue(sig_exp.min() >= -1. and sig_exp.max() <= 1.)
 
-        repr_test = transforms.MuLawEncoding(quantization_channels)
```
vincentqb: Was this block meant to be removed?

jamarshon: Yes the check was to see if the class has a __repr__. Every class has a __repr__ so it did not  really do anything

jamarshon:
In terms of variable name standardization, some concepts were taken from https://github.com/keunwoochoi/torchaudio-contrib/issues/36#issuecomment-502144987, mainly:

`waveform`: a tensor of audio samples
`sample_rate`: the rate of audio samples (samples per second)
`specgram`: a tensor of spectrogram
`mel_specgram`: a mel spectrogram
`hop_length`: the number of samples between the starts of consecutive frames
`n_freqs`: the number of bins in a linear spectrogram (i.e., n_freqs == n_fft // 2 + 1)
`min_freq`: the lowest frequency of the lowest band in a mel/CQT spectrogram
`max_freq`: the highest frequency of the highest band in a mel/CQT spectrogram

Other variable names
`n_fft`: the number of fourier bins (matches torch.stft naming)
`win_length`: the length of the STFT window (matches torch.stft naming)
`n_mfcc`, `n_mels`: just to be consistent with other n_* variables

Currently there is no complex tensors anywhere so no `complex_specgram`. Also no batching so no `waveforms`, `specgrams`, etc.

jamarshon:
@vincentqb @keunwoochoi could you take a look and let me know what you think?

keunwoochoi:
Looks good to me. The things mentioned in [the original thread](https://github.com/pytorch/audio/pull/152#issue-298981045) and [the comment](https://github.com/pytorch/audio/pull/152#issuecomment-513968536) seem aligned to what we had discussed :) 
Curious though - what do you mean by no batching?

cpuhrsch:
@keunwoochoi - see [this comment](https://github.com/pytorch/audio/pull/131#discussion_r305991797) for a wider discussion around this topic.

jamarshon:
https://github.com/pytorch/audio/pull/131#discussion_r305991797
None of the current transforms/functional receive batched inputs/outputs. I think the goal is to have no batch support in this version and then add batching in the next version

vincentqb:
I'm happy with the convention, and the code looks good to me. Thanks for updating the code and the tests.

torchaudio/functional.py
```diff
@@ -9,26 +9,23 @@
...
+    'create_fb_matrix',
     'create_dct',
-    'BLC2CBL',
     'mu_law_encoding',
     'mu_law_expanding'
 ]
 
 
 @torch.jit.script
 def scale(tensor, factor):
```
cpuhrsch: Do we even still want to keep this one? I mean, it's just "tensor / factor" with an "to(torch.float32)" before.

torchaudio/functional.py
```diff
@@ -40,52 +37,44 @@ def scale(tensor, factor):
...
-        ch_dim (int): Dimension of channel (not size)
+        waveform (torch.Tensor): Tensor of audio of size (c, n)
 
     Returns:
-        torch.Tensor: Mono signal
+        torch.Tensor: Tensor that has been downmixed of size (1, n)
     """
-    if not tensor.is_floating_point():
-        tensor = tensor.to(torch.float32)
+    if not waveform.is_floating_point():
```
cpuhrsch: I'm not a big fan of these implicit conversions. What if the users isn't aware this needs floats? Also, it'll return floats for an integer input. Didn't we agree on all waveforms to be floats in range -1 to 1?

torchaudio/functional.py
```diff
@@ -254,59 +244,93 @@ def istft(stft_matrix,          # type: Tensor
...
+    if normalized:
         spec_f /= window.pow(2).sum().sqrt()
-    spec_f = spec_f.pow(power).sum(-1)  # get power of "complex" tensor (c, l, n_fft)
+    spec_f = spec_f.pow(power).sum(-1)  # get power of "complex" tensor
     return spec_f
 
 
 @torch.jit.script
-def create_fb_matrix(n_stft, f_min, f_max, n_mels):
+def spectrogram_to_DB(specgram, multiplier, amin, db_multiplier, top_db=None):
```
cpuhrsch: Did you just move this around? Why?

torchaudio/transforms.py
```diff
@@ -7,441 +7,362 @@
...
 
-    def __repr__(self):
-        return self.__class__.__name__ + '()'
+        Returns:
+            torch.Tensor: Tensor that has been downmixed of size (1, n)
+        """
+        return F.downmix_mono(waveform)
 
 
 class LC2CL(torch.jit.ScriptModule):
```
cpuhrsch: What about this: If we agree that all signal must be of form C x L and torch.load produces C x L then any user should be responsible for this conversion herself? - Similar to scale, this wraps a simple function 'transpose'.

torchaudio/transforms.py
```diff
@@ -7,441 +7,362 @@
...
-        Args:
-        sr (int) : sample rate of audio signal
-        n_mfcc (int) : number of mfc coefficients to retain
-        dct_type (int) : type of DCT (discrete cosine transform) to use
-        norm (string, optional) : norm to use
-        log_mels (bool) : whether to use log-mel spectrograms instead of db-scaled
-        melkwargs (dict, optional): arguments for MelSpectrogram
-    """
-    __constants__ = ['sr', 'n_mfcc', 'dct_type', 'top_db', 'log_mels']
+class MelScale(torch.jit.ScriptModule):
```
cpuhrsch: Can you move this higher up again? It'll be much easier to review the changes.

torchaudio/functional.py
jamarshon: Done.

torchaudio/functional.py
jamarshon: Done.

torchaudio/transforms.py
jamarshon: Done.

torchaudio/functional.py
jamarshon: https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L11
It makes sense to have `spectrogram` and `spectrogram_to_DB` beside each other and `create_fb_matrix` and `create_dct` beside each other.

jamarshon:
Removing scale and lc2cl because we want to assume the user will do the necessary formatting of their input. Also want to keep library lean and only have complex, multiline algorithms.

Removing downmixmono from feedback. Related PR https://github.com/keunwoochoi/torchaudio-contrib/pull/45

torchaudio/functional.py
cpuhrsch: ok, then lets do that in a follow-up PR, but not while changing them at the same time

vincentqb:
If we are making the code leaner, should we only have DBScale/MelScale transforms without having the SpectrogramDB/MelSpectrogram if they are easy to chain?

vincentqb:
> If we are making the code leaner, should we only have DBScale/MelScale transforms without having the SpectrogramDB/MelSpectrogram if they are easy to chain?

If we do this, we'll do it as part of a separate PR.

vincentqb:
Here are things we can consider for a later PR:

- [ ] Remove SpectrogramToDB, and create a DBScale transformation.
- [ ] Remove MelSpectrogram and keep only MelScale.
- [ ] Remove PadTrim.
- [ ] Introduce Padding-only transformation.

test/test_functional.py
```diff
@@ -197,54 +195,6 @@ def _num_stft_bins(signal_len, fft_len, hop_length, pad):
...
-    (torch.randn(1, 100000)),
-    (torch.randn(1, 2, 100000)),
-    pytest.param(torch.randn(1, 100), marks=xfail(raises=RuntimeError)),
-])
-@pytest.mark.parametrize('pad_mode', [
-    # 'constant',
-    'reflect',
-])
-@unittest.skipIf(not IMPORT_LIBROSA, 'Librosa is not available')
-def test_stft(waveform, fft_length, hop_length, pad_mode):
```
cpuhrsch: Why might want to bring this back / add it to core later on for the regular torch.stft

jamarshon:
> Based on the work done here, here are things we can consider for a later PR:
> 
> * [ ]  Remove SpectrogramToDB, and create a DBScale transformation [like so](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/functional.py#L277).
> * [ ]  Remove MelSpectrogram and keep only MelScale.
> * [ ]  Remove PadTrim.
> * [ ]  Introduce Padding-only transformation, [like so](https://github.com/pytorch/audio/pull/122/files#diff-a3428ec41790ed12faea36fcc6dd8f06R457).
> * [ ]  Rename MuLawExpanding to MuLawDecoding.

Rename MuLawExpanding to MuLawDecoding. https://github.com/pytorch/audio/pull/159

keunwoochoi:
awesome!!!

Issue #153, opened 2019-07-18
Resample supports only one channel


vincentqb:
torchaudio.transforms.Resample appears to only apply to single channels.

```
tensor.size()
# torch.Size([2, 276858])

torchaudio.transforms.Resample(frequency, new_frequency)(tensor).size()
# RuntimeError: Given groups=1, weight of size 1 1 121, expected input[1, 2, 276972] to have 1 channels, but got 2 channels instead

torchaudio.transforms.Resample(frequency, new_frequency)(tensor[0,:].view(1,-1))
# torch.Size([1, 27686])

torchaudio.transforms.Compose([
    torchaudio.transforms.LC2CL(),
    torchaudio.transforms.DownmixMono(),
    torchaudio.transforms.LC2CL(),
    torchaudio.transforms.Resample(frequency, new_frequency)
])(tensor).size()
# torch.Size([1, 27686])
```

Since we are in the process of standardizing #152, would it make sense to apply the resample on the last dimension, assumed to be time?

vincentqb:
@jamarshon -- thoughts?

jamarshon:
Fixed in #154

PR #154, opened 2019-07-18
Multichannel Resample

jamarshon:
Fix for https://github.com/pytorch/audio/issues/153


jamarshon:
Waiting on https://github.com/pytorch/pytorch/issues/23042 to figure out why the error is not zero

jamarshon:
Based on https://github.com/pytorch/pytorch/issues/23042, convolution over multiple channels is different from over single channel due to numerical accuracy issue. This means the test would not expect exact matches in terms of outputs.

PR #156, opened 2019-07-19
Adding __version__, git_version to torchaudio

jamarshon:
From pytorch [__init__.py](https://github.com/pytorch/pytorch/blob/master/torch/__init__.py) and [setup.py](https://github.com/pytorch/pytorch/blob/798d5d9771412bea95b44a0c8eb72337cbf73c81/setup.py#L284), they generate version.py and expose it to the user.

This would be useful in torchaudio to see which what configuration is currently installed

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -36,8 +36,9 @@ class Tester(unittest.TestCase):
         result = transforms.Scale()(audio_orig)
         self.assertTrue(result.min() >= -1. and result.max() <= 1.)
 
-        maxminmax = float(max(abs(audio_orig.min()), abs(audio_orig.max())))
+        maxminmax = max(abs(audio_orig.min()), abs(audio_orig.max())).item()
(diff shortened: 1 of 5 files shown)


test/test_transforms.py
```diff
@@ -36,8 +36,9 @@ def test_scale(self):
         result = transforms.Scale()(audio_orig)
         self.assertTrue(result.min() >= -1. and result.max() <= 1.)
 
-        maxminmax = float(max(abs(audio_orig.min()), abs(audio_orig.max())))
+        maxminmax = max(abs(audio_orig.min()), abs(audio_orig.max())).item()
```
cpuhrsch: Touching tests seems to be out of scope for this PR. Can you split this out?

cpuhrsch:
I'm going to yield this to @soumith since it's related to release engineering - not sure how important this is.

PR #160, opened 2019-07-24
Remove PadTrim

jamarshon:


vincentqb:
#152

cpuhrsch:
We're going to revisit batching support.

PR #176, opened 2019-07-29
Make it possible to change the name of torch (e.g., to torch_nightly)

ezyang:


jamarshon:
@ezyang We probably want this in v0.2.0 as well right?

ezyang:
I don't plan to build 0.2.0 nightlies, so no, not really :)

Excerpts from jamarshon's message of 2019-07-29 12:38:53 -0700:
> @ezyang We probably want this in v0.2.0 as well right?
>

PR #186, opened 2019-07-30
Make test scripts runnable without being modules.

ezyang:
This makes it easier to test against an installed wheel, as the
torchaudio folder is no longer preferentially picked up when
you run a test module.

I had to move all tests in subfolders into the top level test
directory to make this work, since you can't access .. modules
without mucking around with sys.path (which I don't want to do.)

NB: this BREAKS the syntax where you can run a test by
saying `python -m test.test`.  Instead, do `python test/test.py`
or use the pytest runner.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -1,5 +1,5 @@
 import unittest
-import test.common_utils
+import common_utils
 import torch
 import torchaudio
 import math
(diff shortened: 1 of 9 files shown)


jamarshon:
Is there a reason for the test/compliance/__init__.py

ezyang:
Yeah, so I can `import compliance.blah`.

PR #193, opened 2019-07-31
Support for building nightlies on master.

ezyang:


jamarshon:
What about osx_wheel.sh

ezyang:
Oh, good point.

ezyang:
OK OS X bits are updated, but I haven't tested them yet

PR #194, opened 2019-08-01
Removal of torchaudio.legacy

jamarshon:
The legacy load/save are wrappers around the regular load/save functions. We want to remove them for this release.

```
def load(filepath, out=None, normalization=None, num_frames=0, offset=0):
  return torchaudio.load(filepath, out, normalization, False, num_frames, offset)

def save(filepath, src, sample_rate, precision=32):
  torchaudio.save(filepath, src, sample_rate, precision, False)
```


vincentqb:
If this goes into the release, then we need to update the notes.

PR #197, opened 2019-08-01
Mention complex number support in README.

vincentqb:
* Mention standard complex number in README, with functionals available.
* Update formatting for `torch.hann_window`

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,7 +127,7 @@ Transforms expect the following dimensions.
 * `MuLawDecode`: (channel, time) -> (channel, time)
 * `Resample`: (channel, time) -> (channel, time)
 
-Complex numbers are supported via tensors of dimension (..., 2), and provide `magphase` to convert such a tensor into its magnitude and phase, and also `complex_norm` and `angle`.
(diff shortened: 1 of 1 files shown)


vincentqb:
@cpuhrsch @jamarshon -- thoughts?

README.md
```diff
@@ -127,6 +127,8 @@ Transforms expect the following dimensions.
 * `MuLawDecode`: (channel, time) -> (channel, time)
 * `Resample`: (channel, time) -> (channel, time)
 
+Complex numbers are supported via tensors of dimension (..., 2), and provide `magphase` to convert such a tensor into its magnitude and phase, and also `complex_norm` and `angle`.
+
```
jamarshon: maybe "its magnitude through `complex_norm` and phase through `angle`."

jamarshon:
Yes I think this is a good idea to mention the dimensions of complex tensors until we move to a better solution (dtype, etc.) https://github.com/pytorch/audio/pull/131#issuecomment-508853162

README.md
vincentqb: I agree we can simplify this. Let's only talk about `complex_norm` and `angle` then. How does that sound?

PR #198, opened 2019-08-01
Packaging improvements for nightlies

ezyang:
- Instructions for how to install stable and master binaries,
  without hardcoding wheel filenames
- Turn on Python 2 build (numba is only needed for testing)
- Make builds a little more hermetic
- Prefer building against nightly CPU-only torch build
- Don't install all of requirements.txt, it's not necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,20 +49,22 @@ Installation
 
 ### Binaries
 
-To install pip wheels of 0.2.0, select the appropriate pip wheel for
-your version of Python:
+To install the latest pip wheels, run:
 
 ```
-# Wheels for Python 2 are NOT supported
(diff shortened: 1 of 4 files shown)


jamarshon:
I prefer to have the requirements in a file just for visibility (maybe packaging_requirements.txt). The current requirements.txt has a bunch of test dependencies like librosa, etc. which is used in the CI

ezyang:
> I prefer to have the requirements in a file just for visibility (maybe packaging_requirements.txt). The current requirements.txt has a bunch of test dependencies like librosa, etc. which is used in the CI

Sure, and this PR doesn't remove requirements.txt. It just stops using them for wheel builds.

PR #199, opened 2019-08-01
Packaging improvements for 0.2.0

ezyang:
Based off of #198 but nightly specific changes were dropped.

### DIFF
--- a/build_tools/packaging/wheel/build_from_source.sh
+++ b/build_tools/packaging/wheel/build_from_source.sh
@@ -28,6 +28,8 @@ popd
 
 # build mad, statically
 pushd libmad-0.15.1b
+# See https://stackoverflow.com/a/12864879/23845
+sed -i.bak 's/-march=i486//' configure
 ./configure --disable-shared --enable-static --prefix="$PREFIX/audio/third_party/mad" CFLAGS=-fPIC CXXFLAGS=-fPIC \
     --with-pic --disable-debug --disable-dependency-tracking
 make -s -j && make install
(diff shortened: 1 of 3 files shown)


build_tools/packaging/wheel/build_from_source.sh
```diff
@@ -28,6 +28,8 @@ popd
 
 # build mad, statically
 pushd libmad-0.15.1b
+# See https://stackoverflow.com/a/12864879/23845
```
soumith: if you want, you can use: https://github.com/conda-forge/mad-feedstock/blob/master/recipe/mad.patch

I had issues with clang  as well, which need this

build_tools/packaging/wheel/osx_wheel.sh
```diff
@@ -1,7 +1,7 @@
-if [[ ":$PATH:" == *"conda"* ]]; then
```
soumith: without this, you are prone to accidents. but up to you.

build_tools/packaging/wheel/build_from_source.sh
ezyang: I was able to build wheels after only this change, so that's kind of interesting.

build_tools/packaging/wheel/osx_wheel.sh
ezyang: I am not sure how the accident occurs; the script initializes a fresh conda environment and then sources it?

soumith: if you already have a conda install in path, and you install + source a new conda install, the PATH will have both conda installs at the same time. Then your cmake will find stuff in the first OR the second conda install

ezyang: I see. OK I'll bring it back

PR #203, opened 2019-08-01
Conda nightly packaging updates.

ezyang:
Billing of changes:
- Can now toggle what branch is built using BRANCH envvar
- PyTorch nightly we were built against is hardcoded into conda meta.yaml
- Disable Conda Python 3.5 builds as we don't have nightlies for torch at the moment
- No more `torchaudio-cpu` or `torchaudio-nightly` suffix
- Remove `librosa` from testing stanza, so that conda build works with stock anaconda channel
- Change convention on `PREFIX` to be what we previously called `$PREFIX/audio`
- Make build_from_source.sh cd-neutral

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ cache:
 matrix:
   fast_finish: true
   include:
-    # TODO add this back in when there is a pytorch 1.2 for python 3.5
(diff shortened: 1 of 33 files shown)


ezyang:
This is stacked on top of #209

PR #208, opened 2019-08-02
Updates to wheel scripts for 0.2 build.

ezyang:


build_tools/packaging/wheel/linux_manywheel.sh
```diff
@@ -18,7 +19,7 @@ done
 OLD_PATH=$PATH
 cd /tmp
 rm -rf audio
-git clone https://github.com/pytorch/audio -b v${TORCHAUDIO_BUILD_VERSION}
+git clone https://github.com/pytorch/audio -b pr/0.2-wheel-updates
```
jamarshon: TORCHAUDIO_BUILD_VERSION isn't used anywhere now, is this intended? I thought v0.2.0 was the branch we intend to build against

ezyang: Yes, I need to revert this before landing.

PR #209, opened 2019-08-02
Make torchaudio work on Python 2.

ezyang:
- Apply `__future__` imports uniformly (future division is the biggy, but absolute
  imports mattered too)
- Hotfix use of tempfile.TemporaryDirectory using a BC library (DO NOT
  add this library as a dependency to torchaudio; it's only for testing)
- Replace math.gcd with fractions.gcd
- Fix a weird pytest collection bug involving parametrized tests
- Turn on Python 2 and Python 3.7 in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,10 +12,11 @@ cache:
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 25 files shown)


ezyang:
Implicitly approved in #203

PR #217, opened 2019-08-07
Rewrite our packaging from scratch

ezyang:
Structure of the PR:
- New entry points are `packaging/build_wheel.sh` and `packaging/build_conda.sh`. The only mandatory environment variable you have to set is `PYTHON_VERSION`
- CircleCI configuration uses 2.1-style parametrized builds to let you toggle python version, etc. as you do builds. We create a separate job per build configuration for maximum parallelism
- `build_tools/packaging` got moved to `packaging`, to be in-line with directory structure in torchvision
- The `build_conda.sh` and `build_wheel.sh` delegate most of the heavy lifting to `pkg_helpers.bash`, which defines a number of bash functions for performing common operations. The intent is that I'll copy-paste this file between other domain API projects. 
- `TORCHAUDIO_` prefix removed from envvars, so that I can more easily share packaging scripts between projects
- `BUILD_VERSION` is completely gone; just change the version number if you need to rebuild
- No more logic for cloning and checking out a fresh copy of torchaudio


packaging/pkg_helpers.bash
```diff
@@ -0,0 +1,133 @@
+# A set of useful bash functions for common functionality we need to do in
+# many build scripts
+
+# Respecting PYTHON_VERSION and UNICODE_ABI, add (or install) the correct
+# version of Python to perform a build.  Relevant to wheel builds.
+setup_python() {
+  if [[ "$(uname)" == Darwin ]]; then
+    eval "$(conda shell.bash hook)"
+    conda create -yn "env$PYTHON_VERSION" python="$PYTHON_VERSION"
```
soumith: what if such an env is already created, and potentially corrupted?

packaging/torchaudio/meta.yaml
```diff
@@ -13,16 +13,16 @@ requirements:
...
   run:
     - python
     - typing
-{{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
+    {{ environ.get('CONDA_PYTORCH_CONSTRAINT') }}
 
 build:
-  number: {{ environ.get('TORCHAUDIO_BUILD_NUMBER') }}
-  string: py{{py}}_{{environ.get('TORCHAUDIO_BUILD_NUMBER')}}
+  string: py{{py}}
```
soumith: you dont have a BUILD_NUMBER, are you sure you will never botch a binary build ever?

packaging/pkg_helpers.bash
ezyang:
```
    conda env remove -n "env$PYTHON_VERSION" || true
```

:)

packaging/torchaudio/meta.yaml
ezyang: My thinking is, if I botch build of `1.2.0`, then I will next build `1.2.0.post2` (because I don't get build number in pip wheels). Now the version has changed and I don't need it. But I could be wrong!

PR #219, opened 2019-08-07
Experimental windows builds

ezyang:
cc @peterjc123 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,7 +110,8 @@ jobs:
       - run:
           command: |
             choco install miniconda3
-            refreshenv
+            # For some reason, refreshenv doesn't work
+            set PATH="C:\tools\miniconda3\Scripts;%PATH%"
             bash packaging/build_wheel.sh
       - store_artifacts:
           path: dist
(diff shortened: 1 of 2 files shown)


peterjc123:
@ezyang I believe this repo is not Windows-ready because it relies on sox that cannot be built statically.

ezyang:
Yeah, I think I agree

HsunGong:
> Yeah, I think I agree

Maybe use socket of `WSL` in order to get `sox` info ?? Seems feasible.

peterjc123:
This is not a good idea. First, it requires users to install WSL. It's a win 10 only feature. Second, if there's WSL, why don't we just run code in WSL? Nevertheless, it may be helpful if you want to use native libraries with torchaudio. But apparently, this is not something we should do here.

black-snow:
> Second, if there's WSL, why don't we just run code in WSL? 

Because it still has no CUDA support.

peterjc123:
@black-snow They actually announced that several days ago. But the discussion here is rather old. We have torchaudio nightly packages for Windows now.

black-snow:
> 
> 
> @black-snow They actually announced that several days ago. But the discussion here is rather old. We have torchaudio nightly packages for Windows now.

Thanks for the reply. Yes they did but only for WSL2 and for that you still have to be on insider/fast-ring (although the may update should be out soon).

How do I get the nightly? For now, I only need torchaudio.datasets (which uses `torchaudio.load` but I don't know if that requires sox already).

peterjc123:
@black-snow 
```
# Conda
conda install -c pytorch-nightly torchaudio
# Pip
pip install -f https://download.pytorch.org/whl/nightly/torch_nightly.html torchaudio
```

mthrok:
Addressed by https://github.com/pytorch/audio/pull/1345

PR #228, opened 2019-08-12
Kaldi MFCC

jamarshon:
We want to write compute-mfcc-feats for compliance.kaldi.mfcc in torchaudio


jamarshon:
current result on 70 files
image

jamarshon:
Found a bug. Now the code is:
```
Max absolute error: 0.00029754638671875
Max relative error: 5.5
```
The 5.5 max relative error has a max absolute error of 5.245208740234375e-06. For file 
```
mfcc-2.0603-1.7529-0.6875-0.6250-7900-false-5086-4-0.74-false-false-true-false-true-false-4-59.0671-7431-7014-1.0000-hamming.ark
```
which is acceptable. Now the graph looks more like expected 😄 
image

torchaudio/compliance/kaldi.py
```diff
@@ -513,13 +520,134 @@ def fbank(
...
+
+    # size (m, num_ceps)
+    feature = feature.matmul(dct_matrix)
+
+    if cepstral_lifter != 0.0:
+        # size (1, num_ceps)
+        lifter_coeffs = _get_lifter_coeffs(num_ceps, cepstral_lifter).unsqueeze(0)
+        feature *= lifter_coeffs
+
+    # if use_energy then replace the last column for htk_compat == true else first column
```
vincentqb: where is `htk_compat == True` checked?

vincentqb:
I see a commit about the README, but I don't see changes in README files. Did you mean to add one?

torchaudio/compliance/kaldi.py
jamarshon: that condition occurs on line 637 where htk_compat makes the last column the energy

jamarshon:
I added a small change to the README to say mfcc along with the related kaldi features we currently have
https://github.com/pytorch/audio/pull/228/files#diff-04c6e90faac2675aa89e2176d2eec7d8R25

torchaudio/compliance/kaldi.py
vincentqb: oh, i see now, thanks

vincentqb:
> I added a small change to the README to say mfcc along with the related kaldi features we currently have
> https://github.com/pytorch/audio/pull/228/files#diff-04c6e90faac2675aa89e2176d2eec7d8R25

Right, I thought I see what that meant now. Thanks!

vincentqb:
> I'm glad to see that there are over 2^6 test files to test all true/false combinations :)

They don't cover all cases since they are picked randomly.

As discussed with @cpuhrsch: 
* We also should get in the habit of building more thorough edge cases.
* Long list of parameters can be passed easily using locals.

jamarshon:
new statistic over 145 files (same overall shape just more points on the left region) + added test for empty tensors
image

PR #206, opened 2019-08-02
[WIP] Voice activity detection algorithm

vincentqb:
This is an initial import of a voice activity detection algorithm by [Moattar and Homayounpour](https://www.eurasip.org/Proceedings/Eusipco/Eusipco2009/contents/papers/1569192958.pdf). See also [notebook](https://github.com/vincentqb/audio-tutorial/blob/master/voice_activity_detection.ipynb).

- [x] Define APIs to invoke the algorithm (see [here](https://github.com/vincentqb/audio-tutorial/blob/master/voice_activity_detection_online.ipynb))
- [ ] Add documentation about what is being done, and docstrings.
- [ ] Optimize operations (e.g. replace python lists with appropriate PyTorch structures).
- [x] Run on streaming data from microphone on laptop (see [here](https://github.com/vincentqb/audio-tutorial/blob/master/voice_activity_detection_online_with_mic.ipynb))
- [x] Incorporate into speech recognition pipeline (see [here](https://github.com/fairinternal/fairseq-py/commit/3260bd422155d2369a67921f2c8f981c170905bb) and [here](https://github.com/cpuhrsch/asr-demo))

PR #229, opened 2019-08-12
Interactive speech recognition demo

cpuhrsch:
This is an interactive command line demo combining the voice activity detection algorithm from #206 with a speech recognition pipeline from fairseq. This is an example, and is not rendered like a tutorial.


examples/interactive_asr/infer_file.py
```diff
@@ -0,0 +1,195 @@
...
+        return mean, 1.0 / (np.sqrt(var) + sys.float_info.epsilon)
+    return mean, 1.0 / np.sqrt(var)
+
+    
+def calcMN(features):
+    mean, invstddev = calc_mean_invstddev(features)
+    res = (features - mean) * invstddev
+    return res
+
+import matplotlib.pyplot as plt
```
vincentqb: Move to beginning?

examples/interactive_asr/infer_file.py
```diff
@@ -0,0 +1,195 @@
...
+    assert len(hypos) == 1
+    transcription = []
+    for i in range(len(hypos)):
+        # Process top predictions
+        hyp_words = process_predictions(args, hypos[i], sp, tgt_dict)
+        transcription.append(hyp_words)
+
+    print('transcription:', transcription)
+    return transcription
+
```
vincentqb: Double space, as done elsewhere

examples/interactive_asr/vad.py
```diff
@@ -0,0 +1,254 @@
...
+
+def compute_energy(frame):
+    return frame.pow(2).sum(-1)
+
+
+class VoiceActivityDetection(object):
+
+    def __init__(self):
+
+        self.num_init_frames = 30
```
vincentqb: it would be nicer to expose `num_init_frames`, `ignore_silent_count`, `ignore_speech_count`, `energy_prim_thresh`, `frequency_prim_thresh`, `spectral_flatness_prim_thresh`, `ignore_silent_count`, `ignore_speech_count` through the `__init__` parameters.

examples/interactive_asr/vad.py
```diff
@@ -0,0 +1,254 @@
...
+        self.ignore_silent_count = 10
```
vincentqb: `ignore_silent_count` and `ignore_speech_count` are defined twice, please remove the first definition.

FYI the original variables in the algorithm had these default values, but they meant something slightly different.

examples/vad/vad.py
```diff
@@ -0,0 +1,254 @@
+"""
```
vincentqb: Duplicated file? One of the two is redundant.

vincentqb:
@cpuhrsch @jamarshon -- I cannot select you for review :) Any thoughts on the demo?

examples/interactive_asr/infer_file.py
```diff
@@ -0,0 +1,195 @@
...
+        if args.fp16:
+            model.half()
+        if use_cuda:
+            model.cuda()
+
+
+def calc_mean_invstddev(feature):
+    if len(feature.shape) != 2:
+        raise ValueError("We expect the input feature to be 2-D tensor")
+    mean = np.mean(feature, axis=0)
```
jamarshon: we could use torch for the np calls in this function (e.g. torch.mean)

jamarshon:
Looks good! Just a few requests:
 - could we share code between infer_file.py and interactive_asr.py by having a separate file (utils.py?) that has the functions (e.g. calcMN is shared between them)
 - also could we merge master into this PR just so we can see whether the tests pass?

vincentqb:
> Looks good! Just a few requests:
> 
> * could we share code between infer_file.py and interactive_asr.py by having a separate file (utils.py?) that has the functions (e.g. calcMN is shared between them)
> * also could we merge master into this PR just so we can see whether the tests pass?

Yup, merged the two script in one, and rebase.

examples/interactive_asr/infer_file.py
vincentqb: Good idea: removed numpy dependency and used pytorch.

examples/interactive_asr/vad.py
```diff
@@ -0,0 +1,278 @@
...
+            self.spectral_flatness_list = []
+
+    def iter(self, frame):
+
+        EPSILON = 1.0
+        frame_fft = torch.rfft(frame, 1)
+        amplitudes = torchaudio.functional.complex_norm(frame_fft)
+
+        # Compute frame energy
+        energy = compute_energy(frame)
```
jamarshon: energy is just amplitudes.pow(2)

examples/interactive_asr/vad.py
```diff
@@ -0,0 +1,278 @@
...
+        self.n = 0
+
+        if self.verbose:
+            self.energy_list = []
+            self.frequency_list = []
+            self.spectral_flatness_list = []
+
+    def iter(self, frame):
+
+        EPSILON = 1.0
```
jamarshon: we could use sys.float_info.epsilon for EPSILON here and in compute_spectral_flatness?

examples/interactive_asr/asr.py
```diff
@@ -0,0 +1,217 @@
...
+    optimize_models(args, use_cuda, models)
+
+    # Initialize generator
+    generator = task.build_generator(args)
+
+    sp = spm.SentencePieceProcessor()
+    sp.Load(os.path.join(args.data, "spm.model"))
+
+    if args.input_file:
+
```
jamarshon: maybe remove the empty line here

examples/interactive_asr/asr.py
```diff
@@ -0,0 +1,217 @@
...
+Run inference for pre-processed data with a trained model.
+"""
+
+import datetime as dt
+import logging
+import os
+import sys
+import time
+
+import numpy as np
```
cpuhrsch: Maybe we can remove this before merging?

vincentqb: thx, good catch!

examples/interactive_asr/asr.py
vincentqb: good idea

examples/interactive_asr/vad.py
vincentqb: move that `EPSILON` within the one place where it is used. there's another epsilon, with a different value, that was renamed to avoid confusion. the model is sensitive to the values of both, and were calibrated.

jamarshon:
need to resolve flake8
./examples/interactive_asr/asr.py:109:69: W291 trailing whitespace

vincentqb:
> need to resolve flake8
> ./examples/interactive_asr/asr.py:109:69: W291 trailing whitespace

good catch!

PR #237, opened 2019-08-14
Fix #236 Length Inaccurate for Multi-Channel

### LINKED ISSUE #236: Length Inaccurate on Multi-Channel Audio Files
## 🐛 Bug

When doing `torchaudio.info()` on a multi channel audio file, the length is multiplied by however many channels there are.

## To Reproduce

1. `torchaudio.load(dual_channel_path)` Look at the shape
1. `torchaudio.info(dual_channel_path)` Look at the length
1. It will be length of a single channel * # of channels

## Expected behavior

I would expect the length to be the length of one channel, not the length of all of them put together. 

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)? Pip
 - What does `torchaudio.__version__` print? (If applicable) 0.3.0

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
```

 - PyTorch Version (e.g., 1.0):
 - OS (e.g., Linux):
 - How you installed PyTorch (`conda`, `pip`, source):
 - Build command you used (if compiling from source):
 - Python version:
 - CUDA/cuDNN version:
 - GPU models and configuration:
 - Any other relevant information:

## Additional context

I think the best way to fix this issue is just to take length and divide it by # of channels.

kevinbird15:
https://github.com/pytorch/audio/issues/236

I believe this will fix the issue assuming that everybody agrees with me that length should be the length of a single channel.

### DIFF
--- a/torchaudio/torch_sox.cpp
+++ b/torchaudio/torch_sox.cpp
@@ -120,7 +120,7 @@ int read_audio_file(
 
   const int number_of_channels = fd->signal.channels;
   const int sample_rate = fd->signal.rate;
-  const int64_t total_length = fd->signal.length;
+  const int64_t total_length = fd->signal.length/fd->signal.channels;
 
   // multiply offset and number of frames by number of channels
   offset *= number_of_channels;


jamarshon:
@kevinbird15 Hi thanks for the PR! I don't believe this is correct though because that line of code is where the buffer_length is determined meaning it should be all the samples (e.g num_channels * samples_per_channel). The buffer_length is then split into its individual channels.
https://github.com/pytorch/audio/blob/cb97f06957309f66e225e0f014697f506f1abb88/torchaudio/torch_sox.cpp#L63
What I think you want to do is modify the SoxDescriptor in [get_info](https://github.com/pytorch/audio/blob/cb97f06957309f66e225e0f014697f506f1abb88/torchaudio/torch_sox.cpp#L90).

The question whether changing this line of logic makes sense should be discussed though. We're returning a `sox_signalinfo_t` and according to [sox length](https://fossies.org/dox/sox-14.4.2/structsox__signalinfo__t.html), this is defined as `samples * chans in file, 0 if unknown, -1 if unspecified`. Whether we want to provide our own definition of length can be discussed 😄

vincentqb:
Let's move the discussion to #236. Please feel free to comment there.

Issue #240, opened 2019-08-16
Suggestion: The device-aware declaration of variable 'window' in istft function


ws-choi:
Hello. I'm implementing an end-to-end model for speech enhancement.
During the implementation, I encountered the error caused by the conflict between cuda-cpu variables.
I think the variable window should be declared with the device of the given input (i.e. stft_matrix).
After fixing that [line](https://github.com/ws-choi/audio/blob/ws-choi-patch-1/torchaudio/functional.py#L117), my code is running well without errors.

Thank you.

p.s. I'm sorry that the post is not written in the given format.

vincentqb:
Good point. We should also consider inheriting the `dtype` from `stft_matrix`.

PR #242, opened 2019-08-16
ISTFT window device for #240

jamarshon:
https://github.com/pytorch/audio/issues/240
the window tensor should be the same device as the stft matrix


jamarshon:
Yes that makes sense. The window should be the same dtype as the stft matrix

Issue #245, opened 2019-08-18
Lazy initialization of MelScale.fb throws when loading module


zh217:
Lazy initialization refers to [the declaration here](https://github.com/pytorch/audio/blob/a424509dda5b57c932fa8b5b780de93e60ed7ee2/torchaudio/transforms.py#L135) and [the initialization here](https://github.com/pytorch/audio/blob/a424509dda5b57c932fa8b5b780de93e60ed7ee2/torchaudio/transforms.py#L146-L150). When loading saved model (saved with torch.save()), this causes problem:

```
size mismatch for reduction.mel_scale.fb: copying a param with shape torch.Size([41, 16]) from checkpoint, the shape in current model is torch.Size([0]).
```

The problem can be sidestepped by running the model at least once before loading weights. But I'd rather not have to rely on such tricks.

Why is this being lazily initialized anyway?

jamarshon:
@zh217 can we have small code snippet of the load/save?

zh217:
Sure.

```python
from torchaudio.transforms import MelSpectrogram

ms = MelSpectrogram()
ms(torch.randn(1, 1000))

ms2 = MelSpectrogram()
ms2.load_state_dict(ms.state_dict())
```

jamarshon:
Thanks for the snippet 😄!  Yes I am not sure why that is that way:
https://github.com/pytorch/audio/pull/72/files#diff-a3428ec41790ed12faea36fcc6dd8f06R228
The PR says that they do not want to reconstruct the fb matrix every time `__call__` is used so they initialize it in `__init__` or for the first input.

We can remove lazy initialization by forcing `n_fft` to be required so that the fb matrix is always created in `__init__` (the user must know the dimensions of their matrix). Then for the `__call__`, it just checks that the input dimensions match expectations and does a matrix multiplication. 

Thoughts, @zh217 @dhpollack @vincentqb @cpuhrsch

zh217:
Making `n_fft` required sounds reasonable to me. I can't think of a use case where I need `n_fft` to be dynamically determined at runtime. Even if there were such a use case, the current implementation doesn't work with changing `n_fft` anyway.

jamarshon:
The `MelSpectrogram` could derive `n_stft` from `n_fft`
https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L202
```
self.mel_scale = MelScale(self.n_mels, self.sample_rate, self.f_min,  
[...]

PR #246, opened 2019-08-19
MelSpectrogram: Use given n_fft to initialize fb immediately

jamarshon:
In the issue https://github.com/pytorch/audio/issues/245, it was determined that for MelSpectrogram the transform already knows how many bins the stft will have and thus can initialize the fb matrix without needing to infer the dimensions of its first input.
The benefit of initializing the fb matrix immediately as opposed to lazy load when given a first input is that loading and saving from state dict (`load_state_dict`) will not throw an error as the tensor sizes match

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -199,7 +199,7 @@ class MelSpectrogram(torch.jit.ScriptModule):
                                        hop_length=self.hop_length,
                                        pad=self.pad, window_fn=window_fn, power=2,
                                        normalized=False, wkwargs=wkwargs)
-        self.mel_scale = MelScale(self.n_mels, self.sample_rate, self.f_min, self.f_max)
+        self.mel_scale = MelScale(self.n_mels, self.sample_rate, self.f_min, self.f_max, self.n_fft // 2 + 1)
 
     @torch.jit.script_method
     def forward(self, waveform):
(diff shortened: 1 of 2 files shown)


torchaudio/transforms.py
```diff
@@ -199,7 +199,7 @@ def __init__(self, sample_rate=16000, n_fft=400, win_length=None, hop_length=Non
                                        hop_length=self.hop_length,
                                        pad=self.pad, window_fn=window_fn, power=2,
                                        normalized=False, wkwargs=wkwargs)
-        self.mel_scale = MelScale(self.n_mels, self.sample_rate, self.f_min, self.f_max)
+        self.mel_scale = MelScale(self.n_mels, self.sample_rate, self.f_min, self.f_max, self.n_fft // 2 + 1)
```
vincentqb: Is there a way to read off the size of the spectrogram instead of recomputing it?

jamarshon: Not unless the spectrogram computes a specgram and we look at the tensor's dimension

vincentqb: If that constant `n_fft // 2 + 1` is used in a few places within Spectrogram, we could attach it to Spectrogram. If that's our inferred default dimension, we could also just attach a `default_dimension` to Spectrogram, in case that ever changes, but that would add clutter to the Spectrogram interface.

Alright, I'm ok with having the value coded there. If there's ever a change in this, the test will fail :)

PR #248, opened 2019-08-20
Increasing test coverage (ASR demo)

jamarshon:
For the ASR demo, we should test that the `transcribe_file` works correctly (i.e. reads "sample.wav" and produces `[['THE QUICK BROWN FOX JUMPS OVER THE LAZY DOG']]`).

To do this, functions inside `examples/interactive_asr/asr.py` are moved into `examples/interactive_asr/utils.py` so that the functions can be called in `examples/test/test_interactive_asr.py`. This also reduces the number of lines of code for `asr.py`

The Travis CI has been updated accordingly

### DIFF
--- a/examples/interactive_asr/README.md
+++ b/examples/interactive_asr/README.md
@@ -16,6 +16,9 @@ and the following models
 
 We recommend that you use [conda](https://docs.conda.io/en/latest/miniconda.html) to install the dependencies when available.
 ```bash
+# Assume that all commands are from the examples folder
+cd examples
+
 # Install dependencies
 conda install -c pytorch torchaudio
 conda install -c conda-forge librosa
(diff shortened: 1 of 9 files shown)


examples/interactive_asr/README.md
```diff
@@ -16,33 +16,48 @@ and the following models
 
 We recommend that you use [conda](https://docs.conda.io/en/latest/miniconda.html) to install the dependencies when available.
 ```bash
+# Assume that all commands are from the examples folder
+cd examples
+
 # Install dependencies
 conda install -c pytorch torchaudio
 conda install -c conda-forge librosa
 conda install pyaudio
```
vincentqb: I don't see pyaudio installed in the scripts. I needed when I experimented with this. Is it needed?

examples/test/test_interactive_asr.py
```diff
@@ -0,0 +1,56 @@
...
+        'max_len_b': 200, 'min_len': 1, 'match_source_len': False,
+        'no_early_stop': False, 'unnormalized': False, 'no_beamable_mm': False,
+        'lenpen': 1, 'unkpen': 0, 'replace_unk': None, 'sacrebleu': False,
+        'score_reference': False, 'prefix_size': 0, 'no_repeat_ngram_size': 0,
+        'sampling': False, 'sampling_topk': -1, 'sampling_topp': -1.0,
+        'temperature': 1.0, 'diverse_beam_groups': -1, 'diverse_beam_strength': 0.5,
+        'print_alignment': False, 'ctc': False,
+        'rnnt': False, 'kspmodel': None, 'wfstlm': None, 'rnnt_decoding_type': 'greedy',
+        'lm_weight': 0.2, 'rnnt_len_penalty': -0.5, 'momentum': 0.99, 'weight_decay': 0.0,
+        'force_anneal': None, 'lr_shrink': 0.1, 'warmup_updates': 0}
```
vincentqb: I'm never too happy with blocks of parameters like that, but I don't have a good alternative to offer :) Oh well :)

examples/interactive_asr/README.md
jamarshon: Yes this is correct, we need pyaudio. It is installed on this line
https://github.com/pytorch/audio/pull/248/files#diff-36b82146ca0e8499b5b86ff3882e063bR61

vincentqb: Oh, right, thanks :)

Issue #262, opened 2019-09-03
Commit d26e112d8b2171ceb78f4c2566245c977b2ff7d1 appears to have broken the build


mcollins42:
## 🐛 Bug

The changes in commit d26e112d8b2171ceb78f4c2566245c977b2ff7d1 to move away from the deprecated torch type `at::DeprecatedTypeProperties` seems to have broken the build: https://travis-ci.org/pytorch/audio/builds/578519920

```
                 from torchaudio/torch_sox.cpp:1:
torchaudio/torch_sox.cpp: In lambda function:
torchaudio/torch_sox.cpp:40:42: error: expected primary-expression before ���>’ token
     auto* data = tensor.data_ptr<scalar_t>();
                                          ^
/home/travis/miniconda3/envs/testenv/lib/python3.7/site-packages/torch/include/ATen/Dispatch.h:12:12: note: in definition of macro ‘AT_PRIVATE_CASE_TYPE’
     return __VA_ARGS__();                          \
            ^
torchaudio/torch_sox.cpp:39:3: note: in expansion of macro ‘AT_DISPATCH_ALL_TYPES’
   AT_DISPATCH_ALL_TYPES(tensor.scalar_type(), "write_audio_buffer", [&] {
   ^
torchaudio/torch_sox.cpp:40:44: error: expected primary-expression before ‘)’ token
     auto* data = tensor.data_ptr<scalar_t>();
                                            ^
```

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)?

Fails on travis and when building/installing with:

`pip install git+git://github.com/pytorch/audio.git`

vincentqb:
Thanks for pointing this out. Reverting in #264.

vincentqb:
Thanks for reporting! Fixed with #265. Please feel free to-reopen or open a new PR if the issue persists.

PR #264, opened 2019-09-03
Revert "add asound for linux. fix deprecated torch type. (#254)"

vincentqb:
Compilation error noticed in #262.

This reverts commit d26e112d8b2171ceb78f4c2566245c977b2ff7d1.


vincentqb:
This does not appear to be the commit that prevents travis from building, see [here](https://travis-ci.org/pytorch/audio/pull_requests).

mcollins42:
Thanks for the quick fix!

PR #268, opened 2019-09-05
Compute deltas

vincentqb:
This PR adds a functional/transform/compliance to compute the deltas of MFCC features, see also [here](http://practicalcryptography.com/miscellaneous/machine-learning/guide-mel-frequency-cepstral-coefficients-mfccs/#deltas-and-delta-deltas).
* The effect at the boundary follows Kaldi's convention of replicating the values (see [here](https://kaldi-asr.org/doc/feature-functions_8h_source.html)).
* To compute the deltas-deltas, simply compute the deltas on the deltas (though Kaldi has a `order` parameter to compute higher order deltas). The compliance interface loops over the order to computer higher order differences.
* This function simply returns the deltas. Kaldi instead appends them to the set of available features, see [here](https://github.com/kaldi-asr/kaldi/blob/master/src/featbin/add-deltas.cc)

To do:
- [ ] ~~Add a test comparing with Kaldi's add_deltas. (leaving off this PR)~~

CC @cpuhrsch @mravanelli

### DIFF
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -193,23 +193,23 @@ class TestFunctional(unittest.TestCase):
 
 
 class TestDeltas(unittest.TestCase):
-    waveform = torch.tensor([1., 2., 3., 4.]).unsqueeze(0)
+    specgram = torch.tensor([1., 2., 3., 4.])
 
-    def _test(self, waveform, expected, n_diff=1, atol=1e-6, rtol=1e-8):
-        computed = F.compute_deltas(waveform, n_diff=1)
+    def _test(self, specgram, expected, n_diff=1, atol=1e-6, rtol=1e-8):
+        computed = F.compute_deltas(specgram, n_diff=1)
         self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
         self.assertTrue(torch.allclose(computed, expected, atol=atol, rtol=rtol))
 
     def test_onechannel(self):
-        waveform = self.waveform
-        expected = torch.tensor([[1.0, 1.0, 1.0, -1.5]])
-        self._test(waveform, expected)
+        specgram = self.specgram.unsqueeze(0).unsqueeze(0)
+        expected = torch.tensor([[[1.0, 1.0, 1.0, -1.5]]])
+        self._test(specgram, expected)
 
     def test_twochannel(self):
-        waveform = torch.cat([self.waveform, self.waveform], dim=0)
-        expected = torch.tensor([[1.0, 1.0, 1.0, -1.5],
-                                 [1.0, 1.0, 1.0, -1.5]])
-        self._test(waveform, expected)
+        specgram = self.specgram.repeat(1, 2, 1)
+        expected = torch.tensor([[[1.0, 1.0, 1.0, -1.5],
+                                  [1.0, 1.0, 1.0, -1.5]]])
+        self._test(specgram, expected)
 
 
 def _num_stft_bins(signal_len, fft_len, hop_length, pad):
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -192,6 +192,26 @@ class TestFunctional(unittest.TestCase):
         self._test_istft_of_sine(amplitude=99, L=10, n=7)
 
 
+class TestDeltas(unittest.TestCase):
+    waveform = torch.tensor([1.,2.,3.,4.]).unsqueeze(0)
+
+    def _test(self, waveform, expected, n_diff=1, atol=1e-6, rtol=1e-8):
+        computed = F.compute_deltas(waveform, n_diff=1)
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+        self.assertTrue(torch.allclose(computed, expected, atol=atol, rtol=rtol))
+
+    def test_onechannel(self):
+        waveform = self.waveform
+        expected = torch.tensor([[ 1.0,  1.0,  1.0, -1.5]])
+        self._test(waveform, expected)
+
+    def test_twochannel(self):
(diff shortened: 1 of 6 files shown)


test/test_functional.py
```diff
@@ -192,6 +192,35 @@ def test_istft_of_sine(self):
         self._test_istft_of_sine(amplitude=99, L=10, n=7)
 
 
+class TestDeltas(unittest.TestCase):
```
vincentqb: @zhangguanheng66 recommends merging this class with the one above

PS the test file should be reorganized a bit too :)

test/test_functional.py
```diff
@@ -192,6 +192,35 @@ def test_istft_of_sine(self):
...
+    specgram = torch.tensor([1., 2., 3., 4.])
+
+    def _test(self, specgram, expected, window=1, atol=1e-6, rtol=1e-8):
+        computed = F.compute_deltas(specgram, window=1)
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+        self.assertTrue(torch.allclose(computed, expected, atol=atol, rtol=rtol))
```
zhangguanheng66: there is a allclose func `torch.testing.assert_allclose`

test/test_transforms.py
```diff
@@ -281,5 +281,16 @@ def test_resample_size(self):
...
 
+    def test_compute_deltas(self):
+        channel = 13
+        n_mfcc = channel * 3
+        time = 1021
+        window = 7
+        specgram = torch.randn(channel, n_mfcc, time)
+        transform = transforms.ComputeDeltas(window=window)
+        computed = transform(specgram)
+        self.assertTrue(computed.shape == specgram.shape, (computed.shape, specgram.shape))
```
zhangguanheng66: You could add a deterministic test here to cover the transform function.

torchaudio/transforms.py
```diff
@@ -365,3 +365,36 @@ def forward(self, waveform):
             return kaldi.resample_waveform(waveform, self.orig_freq, self.new_freq)
 
         raise ValueError('Invalid resampling method: %s' % (self.resampling_method))
+
+
+class ComputeDeltas(torch.jit.ScriptModule):
```
zhangguanheng66: if this depends on the jib mode, it's better to add a jit test to make sure that it works properly.

torchaudio/transforms.py
```diff
@@ -365,3 +365,36 @@ def forward(self, waveform):
...
+        window (int): A nonzero number of differences to use in computing delta
+    """
+    __constants__ = ['window']
+
+    def __init__(self, window=2):
+        super(ComputeDeltas, self).__init__()
+        self.window = window
+
+    @torch.jit.script_method
+    def forward(self, specgram):
```
zhangguanheng66: should the window be the argument of the forward func, instead of the class attribute? Do you call the func with a different value of window?

test/test_functional.py
zhangguanheng66: yeap, we don't need a separate class to cover the dalta func.

torchaudio/transforms.py
vincentqb: To follow the convention of the other transform, forward only takes the waveform. We won't change the window very often.

torchaudio/transforms.py
vincentqb: Do you have an example of what you have in mind?

zhangguanheng66: You just need to add a text case to ensure every operator is traceable.

vincentqb: There was already a test verifying that the shape of the output is correct. I added a test to verify that the output matches the functional.  Thanks for pointing out [this link](https://l.workplace.com/l.php?u=https%3A%2F%2Fgithub.com%2Fpytorch%2Fpytorch%2Fblob%2F00d967c39da15b1654c7ec3afa57f2a983c1e70c%2Ftest%2Ftest_jit.py%23L14033&h=AT3e-r0lqKPwcdpluC73rNGwvXCLZNbeSRbnRr8wMB77YlMdbZfrMnqqBLeArbmNDV7pDkYoKI7ADwRW4MSoLY12N94bE0brPfYc4CnCiUEPnlE41RKQlqh6kUzJOuYTL91wjyAPZme0XUmKRWeq5Q_6TECMx9EI3olwSg), but could you clarify what you would like to verify? If the module does not compile, the test would not run at all.

torchaudio/functional.py
```diff
@@ -652,3 +652,51 @@ def lowpass_biquad(waveform, sample_rate, cutoff_freq, Q=0.707):
     a1 = -2 * math.cos(w0)
     a2 = 1 - alpha
     return biquad(waveform, b0, b1, b2, a0, a1, a2)
+
+
+def compute_deltas(specgram, win_length=5):
+    # type: (Tensor, int) -> Tensor
+    r"""Compute delta coefficients of a spectogram:
```
cpuhrsch: Nit: This technically isn't limited to spectrograms right? I mean, we could just talk about regular Tensors as inputs.

Also, the edge behavior might be something the user could want to change in the future. Some people rather not replicate the deltas, but just drop those frames altogether.

vincentqb: > Nit: This technically isn't limited to spectrograms right? I mean, we could just talk about regular Tensors as inputs.

fair enough, it's commonly used on spectrogram, but i could replace this by `waveform`, since there is a time-component, though it is not limited to `waveform`. the list of [standardized words](https://github.com/pytorch/audio/blob/master/README.md) might not be useful here? :)

> Also, the edge behavior might be something the user could want to change in the future. Some people rather not replicate the deltas, but just drop those frames altogether.

This is the behavior that kaldi does, and also does not make configurable :) I'll add the option to change that, by simply passing along the mode option for `pad`, and default to 'replicate'.

PR #272, opened 2019-09-09
Use older numpy for python 2.7

vincentqb:
Conda builds for torchaudio have been failing recently. This may be due to numpy having dropped python 2.7 support in recent versions.


vincentqb:
Testing that #274 fails to build so as to conclude that the build success really is due to the change here :)

vincentqb:
Based on #274, the build failures were flaky from upstream.

Issue #260, opened 2019-08-28
torch transforms inspired by sox effects


vincentqb:
## 🚀 Feature

Convert [sox effects](http://sox.sourceforge.net/sox.html#EFFECTS) into transforms. Ideally, we provide enough building blocks to make those transforms easy to assemble.

- [x] allpass (with biquad)
- [x] band (with biquad)
- [x] bandpass/bandreject (with biquad)
- [x] bass (with biquad)
- [x] treble (with biquad)
- [ ] bend
- [x] biquad
- [x] ~channels (we already decided downmix, etc, is easy to do by hand)~
- [ ] chorus
- [ ] compand/mcompand
- [ ] contrast
- [x] dcshift
- [x] deemph (with biquad, related to [this](https://github.com/keunwoochoi/torchaudio-contrib/issues/61#issuecomment-524418049))
- [ ] delay
- [x] dither
- [x] downsample/upsample (available with resample)
- [ ] earwax
- [ ] echo/echos
- [x] equalizer (with biquad, see also [here](https://ccrma.stanford.edu/~jos/filters/Peaking_Equalizers.html))
- [x] fade
- [ ] fir
- [x] flanger
- [x] gain
- [x] highpass/lowpass (with biquad)
- [ ] hilbert
- [ ] ladspa
- [ ] loudness
- [ ] mcompand
- [ ] noiseprof/noisered
- [ ] ~norm (duplicate of gain)~
- [x] ~~oops (simply the difference between two channels)~~
- [x] overdrive
- [x] ~pad/trim (easy to do without transforms)~
- [ ] phaser
- [ ] pitch
- [x] ~rate (resample already available)~
- [x] ~remix (we already decided downmix, etc, is easy to do by hand)~
- [x] ~repeat (`tensor.repeat` or [pad with replicate](https://pytorch.org/docs/stable/nn.functional.html?highlight=pad#torch.nn.functional.pad))~
- [ ] reverb
- [ ] ~reverse~
- [x] riaa (with biquad)
- [ ] silence
- [ ] sinc
- [x] spectrogram
- [x] ~speed (change sample rate)~
- [x] stretch/tempo (vocoder functional available)
- [ ] splice
- [x] ~~stat/stats (only prints information about the waveform)~~
- [x] ~swap~
- [ ] synth
- [ ] tremolo
- [x] vad (in [example](https://github.com/pytorch/audio/blob/master/examples/interactive_asr/vad.py))
- [x] vol

[...]

PR #275, opened 2019-09-10
[WIP] Exploring Filtering Capabilities

engineerchuan:
Use biquad filter similar to SoX.


engineerchuan:
Hi Community,

We are currently exploring ways to make torchaudio's dependence on SoX optional (per https://github.com/pytorch/audio/issues/260).

A subset of SoX's functionality is its frequency filtering operations e.g. highpass, lowpass, bandpass, etc..  all of which are based on the biquad effect (https://en.wikipedia.org/wiki/Digital_biquad_filter). 
 This WIP PR ports SoX's implementation of the biquad effect, lowpass, and highpass filters.

**Question:  Which parts of the frequency filtering functions would be helpful to include in torchaudio vs keep separate?**

- One approach would be to provide only the core "execution layer" and leave the filter design outside the library.
- For an IIR filter like biquad, this could mean implementing the biquad effect but asking the user to supply the coeffs (b0, b1, b2, a0, a1, a2).  Or perhaps implementing a general IIR difference equation execution engine.
- For a general FIR filter, the user could use filter design libraries like scipy.signal to design the impulse response.  Then we can use torch's GPU accelerated convolution functions to execute.

@vincentqb 

Thank you,

Chuan

vincentqb:
Relates to #260

examples/filter_file.py
```diff
@@ -0,0 +1,14 @@
+import os
+import sys
+import torchaudio
+
+if __name__ == '__main__':
```
vincentqb: This is a nice code snippet, but demonstrating the `lowpass` would be better in [the audio tutorial](https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html) from [here](https://github.com/pytorch/tutorials/blob/master/beginner_source/audio_preprocessing_tutorial.py).

The good things from this example would be either if we want to offer some command line parsing, or if we want to use this as part of integrated testing. I'd open separate PRs though.

Thoughts?

test/test_functional_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+
+        white_noise_path = os.path.join(os.path.join(self.test_dirpath, "assets"), 'whitenoise.mp3')
+        audio, sample_rate = torchaudio.load(white_noise_path, normalization=False)
+        high_to_low_pre_lowpass = compute_ratio_high_vs_low(audio, 512)
+        self.assertTrue(0.25 < high_to_low_pre_lowpass < 4)
+        CUTOFF_FREQ = sample_rate // 4 # half of nyquist
+        lowpassed_audio = F.lowpass(audio, sample_rate, 512, CUTOFF_FREQ)
+        high_to_low_post_lowpass = compute_ratio_high_vs_low(lowpassed_audio, 512)
+        
+        print(high_to_low_pre_lowpass, high_to_low_post_lowpass)
```
vincentqb: nit: This is useful for [WIP] but let's not forget to remove it before the final version :)

test/test_functional_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+        noise_filepath = os.path.join(self.test_dirpath, "assets", "whitenoise.mp3")
+        E = torchaudio.sox_effects.SoxEffectsChain()
+        E.set_input_file(noise_filepath)
+        E.append_effect_to_chain("lowpass", [CUTOFF_FREQ])
+        sox_signal_out, sr = E.sox_build_flow_effects()
+
+        noise, sample_rate = torchaudio.load(noise_filepath, normalization=True)
+        signal_out = F.lowpass_biquad(noise, sample_rate, CUTOFF_FREQ)
+
+        print(sox_signal_out, signal_out)
```
vincentqb: Same

test/test_functional_filtering.py
```diff
@@ -0,0 +1,117 @@
...
+        noise_filepath = os.path.join(self.test_dirpath, "assets", "whitenoise.mp3")
+        E = torchaudio.sox_effects.SoxEffectsChain()
+        E.set_input_file(noise_filepath)
+        E.append_effect_to_chain("highpass", [CUTOFF_FREQ])
+        sox_signal_out, sr = E.sox_build_flow_effects()
+
+        noise, sample_rate = torchaudio.load(noise_filepath, normalization=True)
+        signal_out = F.highpass_biquad(noise, sample_rate, CUTOFF_FREQ)
+
+        print(sox_signal_out, signal_out)
```
vincentqb: Same

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+import math
+import torch
+import math
+import torchaudio
+
+"""
+Biquad filtering approach based on SoX implementation 14.4.2
+
+Performance is currently very slow
```
vincentqb: Have you seen performance gain by going FIR instead?

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
...
+"""
+
+__all__ = [
```
vincentqb: I'm not for or against, but why did you decide to move those functionals outside of functionals? Are you suggesting all of sox re-implementation should be in separate files?

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
...
+        o0 = input[:, i] * b0 + pi1 * b1 + pi2 * b2 - po1 * a1 - po2 * a2
+        pi2 = pi1
+        pi1 = input[:, i]
+        po2 = po1
+        po1 = o0
+        output[:, i] = o0
+
+    return output
+
+def lowpass1(waveform,           # type: Tensor
```
vincentqb: Let's leave it in for now, but this will be superseded by the other low pass. Out of curiosity, have you seen performance difference between the two low pass filters?

torchaudio/functional_filtering.py
vincentqb: and also compare to sox :)

examples/filter_file.py
engineerchuan: I agree.  I have removed this from latest update.

test/test_functional_filtering.py
engineerchuan: I have removed this particular one.

test/test_functional_filtering.py
engineerchuan: I have also removed this one.

test/test_functional_filtering.py
engineerchuan: I have also removed this one.

torchaudio/functional_filtering.py
engineerchuan: I will follow up with performance tests.  This was just a guess but I will test.

torchaudio/functional_filtering.py
engineerchuan: Yes - may be nice to separate into functional layer and convenience layers for SoX.

torchaudio/functional_filtering.py
engineerchuan: I have decided to remove this.  I will do comparisons of performance between FIR and IIR filters.

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
...
+    b0 = (1 - math.cos(w0))/2
+    b1 = 1 - math.cos(w0)
+    b2 = b0
+    a0 = 1 + alpha
+    a1 = -2 * math.cos(w0)
+    a2 = 1 - alpha
+    return biquad(input, b0, b1, b2, a0, a1, a2)
+
+
+def biquad(input, b0, b1, b2, a0, a1, a2):
```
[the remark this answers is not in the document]
engineerchuan: OK.

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
...
+    b0 = (1 + math.cos(w0))/2
+    b1 = -1 - math.cos(w0)
+    b2 = b0
+    a0 = 1 + alpha
+    a1 = -2 * math.cos(w0)
+    a2 = 1 - alpha
+    return biquad(input, b0, b1, b2, a0, a1, a2)
+
+
+def lowpass_biquad(input, sr, cutoff_freq, Q = 0.707):
```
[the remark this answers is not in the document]
engineerchuan: OK.

torchaudio/functional_filtering.py
```diff
@@ -0,0 +1,160 @@
...
+    'biquad',
+    'lowpass_biquad',
+    'highpass_biquad',
+]
+
+
+def _dB2Linear(x):
+    return math.exp(x * math.log(10) / 20.0)
+
+def highpass_biquad(input, sr, cutoff_freq, Q = 0.707):
```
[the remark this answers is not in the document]
engineerchuan: OK.

torchaudio/functional_sox_convenience.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+import math
+import torch
+import torchaudio
+from _torch_filtering import diff_eq as diffeq_cpp
+
+__all__ = ["lowpass_biquad", "highpass_biquad", "biquad"]
```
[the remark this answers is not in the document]
engineerchuan: dibe,

torchaudio/functional_sox_convenience.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def biquad(input_waveform, b0, b1, b2, a0, a1, a2):
+    output_waveform = torch.zeros_like(input_waveform)
+    assert input_waveform.dtype == torch.float32
```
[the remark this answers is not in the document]
engineerchuan: done

torchaudio/functional.py
```diff
@@ -466,3 +471,63 @@ def phase_vocoder(complex_specgrams, rate, phase_advance):
...
+        N.B. all tensors must be of float32 type
+
+    Returns:
+        torch.Tensor: output waveform filtered, shape `(n_channels, n_frames)`
+    """
+
+    output_waveform = torch.zeros_like(input_waveform)
+    assert input_waveform.dtype == torch.float32
+
+    diffeq_cpp(input_waveform, output_waveform, a_coeffs, b_coeffs)
```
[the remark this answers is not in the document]
engineerchuan: done.

torchaudio/functional.py
```diff
@@ -466,3 +471,63 @@ def phase_vocoder(complex_specgrams, rate, phase_advance):
...
+
+    n_channels, n_frames = input_waveform.size()
+
+    # Perform convolution for each channel
+    assert input_waveform.dtype == torch.float32
+    assert impulse_response.dtype == torch.float32
+
+    # N.B. - can we do all channels at once?
+    channel_output_waveforms = []
+    for i_channel in range(n_channels):
```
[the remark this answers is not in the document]
engineerchuan: Let's remove the convolve function since all it does is wrap.

torchaudio/functional.py
```diff
@@ -466,3 +471,63 @@ def phase_vocoder(complex_specgrams, rate, phase_advance):
...
+        impulse_response (torch.Tensor): Tensor shape of `(n_response_length)`
+
+    Returns:
+        torch.Tensor: output waveform filtered, shape `(n_channels, n_frames + n_response_length - 1)`
+    """
+
+    n_channels, n_frames = input_waveform.size()
+
+    # Perform convolution for each channel
+    assert input_waveform.dtype == torch.float32
```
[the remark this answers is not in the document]
engineerchuan: Done.

test/test_functional_filtering.py
```diff
@@ -0,0 +1,203 @@
...
+import unittest
+import common_utils
+import time
+
+
+class TestFunctionalFiltering(unittest.TestCase):
+    test_dirpath, test_dir = common_utils.create_temp_assets_dir()
+
+    def test_lfilter_basic(self):
+        """l
```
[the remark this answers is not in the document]
engineerchuan: Yes.

torchaudio/functional.py
```diff
@@ -466,3 +471,63 @@ def phase_vocoder(complex_specgrams, rate, phase_advance):
     complex_specgrams_stretch = torch.stack([real_stretch, imag_stretch], dim=-1)
 
     return complex_specgrams_stretch
+
+
+def convolve(input_waveform, impulse_response):
```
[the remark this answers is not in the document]
engineerchuan: OK I've removed it.

torchaudio/filtering.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    // assumes waveform is normalized between 1 and -1
+
+    int64_t n_channels = input_waveform.size(0);
+    int64_t n_frames = input_waveform.size(1);
+
+    assert(output_waveform.size(0) == n_channels);
+    assert(output_waveform.size(1) == n_frames);
+
+    auto input_accessor = input_waveform.accessor<float,2>();
```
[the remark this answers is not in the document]
engineerchuan: I added GPU typedefs based on my understanding of code but we need to test this.

torchaudio/filtering.cpp
```diff
@@ -0,0 +1,80 @@
...
+    auto output_accessor = output_waveform.accessor<float,2>();    
+    auto a_coeffs_accessor = a_coeffs.accessor<float,1>();
+    auto b_coeffs_accessor = b_coeffs.accessor<float,1>();
+
+    int n_order = a_coeffs.size(0); // n'th order - 1 filter
+    assert(a_coeffs.size(0) == b_coeffs.size(0));
+
+    for (int64_t i_channel = 0; i_channel < n_channels; ++i_channel) {
```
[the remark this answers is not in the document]
engineerchuan: I'm not sure there is necessarily a faster way.  Would you use tensors to slice all channels at each frame?  Would that be much faster?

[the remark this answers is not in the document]
vincentqb: Slicing would be great, yes. I'd expect this to work faster on GPU.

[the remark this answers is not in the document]
vincentqb: @yf225 -- We want to implement a transformation `lfilter` that could "almost" be implemented by convolutions. We can't because of the way the terms depend on each other. @engineerchuan suggests implementing it in C++ (see here, in `torchaudio/filtering.cpp`), since the for-loop in time is much faster and comparable to scipy in speed on CPU. We shouldn't need a for-loop over the channels though. Thoughts on things to be careful about here?

engineerchuan:
3 different implementations of lfilter were explored in this commit.
1. `lfilter` "Element-wise" computation using tensor accessors (https://pytorch.org/cppdocs/notes/tensor_basics.html#efficient-access-to-tensor-elements)
2. `lfilter_tensor` "Slice" computation where the difference equation is evaluated simultaneously across all channels by adding and subtracting slices.
3. `lfilter_tensor_matrix` "Matrix" computation where we recast the difference equation computation as a matrix multiply.

All 3 implementations return same result at 1e-5 variety of random inputs.  We have confidence they are doing the same maths.

Performance (all on CPU):

Input (2 channel x 100K samples):
Looped Implementation took       :  0.0026006698608398438
Tensor Implementation took       :  7.084993362426758
Tensor Matrix Implementation took:  2.870526075363159

Input (10 channel x 10K samples):
Looped Implementation took       :  0.0015499591827392578
Tensor Implementation took       :  0.8290731906890869
Tensor Matrix Implementation took:  0.29657483100891113

Input (100 channel x 10K samples):
Looped Implementation took       :  0.01202535629272461
Tensor Implementation took       :  0.9394674301147461
Tensor Matrix Implementation took:  0.4104940891265869

The basic element wise implementation seems much faster.  I need help evaluating my tensor based implementation.  Is the method shown of slicing the most efficient way to access? Do I need to contiguous() at all?

Thank you,

Chuan

torchaudio/filtering.cpp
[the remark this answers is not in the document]
yf225: Since the computation for each channel doesn't depend on each other, we might be able to use `at::parallel_for` (https://github.com/pytorch/pytorch/blob/cc61af3c3d8ca8b46f7234383513b5166e10150c/aten/src/ATen/Parallel.h#L48) to speed up on CPU, and `thrust::for_each` (or a custom CUDA kernel launch) to speed up on GPU.

engineerchuan:
Unwound the cpp lfilter implementations.  We will move these to a separate PR.

docs/source/functional_sox_compatibility.rst
```diff
@@ -0,0 +1,24 @@
+.. role:: hidden
+    :class: hidden-section
+
+torchaudio.functional_sox_compatibility
```
vincentqb: For now, let's actually move all those functionals into the functional. We should only claim compatibility if we test agreement between sox and pytorch. We may revisit that decision at a later time.

test/test_datasets_vctk.py
```diff
@@ -23,10 +23,13 @@ def test_is_audio_file(self):
     def test_make_manifest(self):
         audios = vctk.make_manifest(self.test_dirpath)
         files = ['kaldi_file.wav', 'kaldi_file_8000.wav',
-                 'sinewave.wav', 'steam-train-whistle-daniel_simon.mp3']
+                 'sinewave.wav', 'steam-train-whistle-daniel_simon.mp3',
+                 'dtmf_30s_stereo.mp3', 'whitenoise_1min.mp3', 'whitenoise.mp3']
```
vincentqb: Why is this change here?

test/test_functional_filtering.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+
+        filepath = os.path.join(self.test_dirpath, "assets", "whitenoise.mp3")
+        waveform, sample_rate = torchaudio.load(filepath, normalization=True)
+        output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
+        assert len(output_waveform.size()) == 2
+        assert output_waveform.size(0) == waveform.size(0)
+        assert output_waveform.size(1) == waveform.size(1)
+
+    def test_lowpass_sox_compliance(self):
```
vincentqb: We'll need to rename this test then :)

docs/source/index.rst
```diff
@@ -13,6 +13,7 @@ The :mod:`torchaudio` package consists of I/O, popular datasets and common audio
    kaldi_io
    transforms
    functional
+   functional_sox_compatibility
```
vincentqb: We'll need to remove this then :)

test/test_functional_filtering.py
```diff
@@ -0,0 +1,158 @@
...
+        assert torch.allclose(
+            waveform[:, 0:-3], output_waveform[:, 3:], atol=1e-5
+        )
+
+    def test_lfilter(self):
+        """
+        Design an IIR lowpass filter using scipy.signal filter design
+        https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.iirdesign.html#scipy.signal.iirdesign
+
+        from scipy.signal import iirdesign
```
vincentqb: Each line of the example should start with a `>`, see [here](https://github.com/pytorch/pytorch/blob/master/torch/optim/lr_scheduler.py#L121) for instance.

test/test_functional_filtering.py
```diff
@@ -0,0 +1,158 @@
...
+        sox_output_waveform, sr = E.sox_build_flow_effects()
+
+        waveform, sample_rate = torchaudio.load(
+            noise_filepath, normalization=True
+        )
+        output_waveform = FSOX.lowpass_biquad(waveform, sample_rate, CUTOFF_FREQ)
+
+        assert torch.allclose(sox_output_waveform, output_waveform, atol=1e-4)
+
+    def test_highpass_sox_compliance(self):
```
vincentqb: Same

torchaudio/functional_sox_compatibility.py
```diff
@@ -0,0 +1,98 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
```
vincentqb: As mentioned above, let's just merge these into the functional file.

test/test_datasets_vctk.py
engineerchuan: We added more audio files (generated files, license free) for testing.  The VCTK test was failing.  Is sinewav, steam-train-whistle, actually a dataset or its just checking which files it sees in the directory.

test/test_functional_filtering.py
engineerchuan: OK

test/test_functional_filtering.py
engineerchuan: OK

test/test_functional_filtering.py
engineerchuan: OK

test/test_datasets_vctk.py
vincentqb: I see. This test is only to see that the manifest is generated correctly with the given test folder. Indeed, we added files, and did not update the test.

PR #276, opened 2019-09-11
Fix version import

vincentqb:
Address version import, see [here](https://our.internmc.facebook.com/intern/diff/D17289027/).


zhangguanheng66:
Just a suggestion. We should separate lint fix and the real issue in two PRs for the tracking purpose.

PR #279, opened 2019-09-14
add linearity test for istft

swchen1234:
Add four test cases of linearity check for inverse stft in test_functional.py.

### DIFF
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -267,7 +267,7 @@ class TestFunctional(unittest.TestCase):
         }
         data_size = (2, 12, 7, 2)
         self._test_linearity_of_istft(data_size, kwargs3)
-    #
+
     def test_linearity_of_istft4(self):
         # hamming_window, not centered, not normalized, onesided
         kwargs4 = {
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -218,6 +218,77 @@ class TestFunctional(unittest.TestCase):
         self._test_istft_of_sine(amplitude=80, L=9, n=6)
         self._test_istft_of_sine(amplitude=99, L=10, n=7)
 
+    def _test_linearity_of_istft(self, kwargs):
+        for data_size in self.data_sizes:
+            for i in range(self.number_of_trials):
+                sound1 = common_utils.random_float_tensor(i, data_size)
+                sound2 = common_utils.random_float_tensor(i * 2, data_size)
+                a, b = torch.rand(2)
+                stft1 = torch.stft(sound1, **kwargs)
+                stft2 = torch.stft(sound2, **kwargs)
+                istft1 = torchaudio.functional.istft(stft1, **kwargs)
+                istft2 = torchaudio.functional.istft(stft2, **kwargs)
+                sound = a * istft1 + b * istft2
+                estimate = torchaudio.functional.istft(a * stft1 + b * stft2,  **kwargs)
+                self._compare_estimate(sound, estimate)
+
+    def test_linearity_of_istft1(self):
+        # hann_window, centered, normalized, onesided
+        kwargs1 = {
+            'n_fft': 12,
+            'hop_length': 4,
+            'win_length': 12,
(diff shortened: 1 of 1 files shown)


test/test_functional.py
```diff
@@ -191,6 +191,76 @@ def test_istft_of_sine(self):
...
+            for i in range(self.number_of_trials):
+                sound1 = common_utils.random_float_tensor(i, data_size)
+                sound2 = common_utils.random_float_tensor(i * 2, data_size)
+                a, b = torch.rand(2)
+                stft1 = torch.stft(sound1, **kwargs)
+                stft2 = torch.stft(sound2, **kwargs)
+                istft1 = torchaudio.functional.istft(stft1, **kwargs)
+                istft2 = torchaudio.functional.istft(stft2, **kwargs)
+                sound = a * istft1 + b * istft2
+                estimate = torchaudio.functional.istft(a * stft1 + b * stft2,  **kwargs)
```
vincentqb: Thanks for looking into this! This is testing the linearity of STFT, and that ISTFT is the inverse of ISTFT. Let's do a test instead that generates 2 frequency-domain tensors, say specgram1 and specgram2, of the shape you would get from `torch.stft`, and then assert that 
```
a * istft(specgram1) + b * istft(specgram2) = istft(a * specgram1 + b * specgram2)
```

engineerchuan:
We recently added retries to the build process to be robust to network timeouts.  That seems to be why the tests failed, versus anything wrong with your test itself.

test/test_functional.py
vincentqb: Thanks for looking into this! Let's avoid using STFT in this test, and generate a random `specgram` tensor by generating directly a tensor of the form of the output of STFT.

vincentqb:
Can you rebase your changes on master? There seems to be changes that are not available here. In particular, if you look at the error logs, `curl` is called without retry, see [here](https://github.com/pytorch/audio/blob/master/packaging/build_from_source.sh#L13).

test/test_functional.py
```diff
@@ -191,6 +191,69 @@ def test_istft_of_sine(self):
...
+            'n_fft': 12,
+            'window': torch.hamming_window(12),
+            'center': True,
+            'pad_mode': 'constant',
+            'normalized': True,
+            'onesided': False,
+        }
+        data_size = (2, 12, 7, 2)
+        self._test_linearity_of_istft(data_size, kwargs3)
+    #
```
vincentqb: I'm assuming you pushed this to force a refresh of the test. Please mark it as "TODO: Remove this line after the tests are all passed."

swchen1234:
still gets lots of  Failed to connect to downloads.sourceforge.net port 80: Operation timed out failure.

vincentqb:
> still gets lots of Failed to connect to downloads.sourceforge.net port 80: Operation timed out failure.

Have you rebased on master? It seems like your version doesn't have the `curl` fix, see [here](https://github.com/pytorch/audio/blob/1e06c98f638ba892a0b09ec46ef62ef8734a2a71/packaging/build_from_source.sh#L13).

test/test_functional.py
```diff
@@ -191,6 +191,69 @@ def test_istft_of_sine(self):
         self._test_istft_of_sine(amplitude=80, L=9, n=6)
         self._test_istft_of_sine(amplitude=99, L=10, n=7)
 
+    def _test_linearity_of_istft(self, data_size, kwargs):
+        for i in range(self.number_of_trials):
+            tensor1 = common_utils.random_float_tensor(i, data_size)
```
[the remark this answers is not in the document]
vincentqb: `data_size` is of (channels, freq, time, complex) on L215 etc. Good!

vincentqb:
flake8 and rebased

vincentqb:
One test needs to be relaxed from `atol=1e-6` to `atol=1e-5`.

vincentqb:
pytorch/pytorch#3775

Issue #278, opened 2019-09-13
[docs] Function arguments in function signature are missing for many functions


vadimkantorov:
https://pytorch.org/audio/functional.html

Many functions like `mu_law_encoding` are listed as `torchaudio.functional.mu_law_encoding()` with no arguments (unlike the `istft` function that has them in the signature). Arguments are still specified below in the Parameters section, but it is confusing.

vincentqb:
The problem comes from the fact that those functions are decorated with `@torch.jit.script`. The forward looking solution is probably to remove the decorator (still test that they can be decorated). 

PS In the same vein, transforms should probably also be transferred from `torch.jit.ScriptModule` to `nn.Module`.

mthrok:
Closing the issue as they look fine in the latest documentation. https://pytorch.org/audio/0.7/functional.html

Issue #165, opened 2019-07-25
Remove _docs.py


jamarshon:
In https://github.com/pytorch/audio/pull/166, we add _docs.py because we can't get the doc string of a class in a `@torch.jit.script_method` without creating an instance (e.g using the syntax `Class.method.__doc__`).
So the workaround is to have _docs.py
```
class Spectrogram:
    forward = torchaudio.transforms.Spectrogram().forward
```
This creates an instance to expose the docstring to sphinx. 
An issue has been made for the pytorch jit https://github.com/pytorch/pytorch/issues/23394

PR #282, opened 2019-09-16
Add explicit signature for Sphinx autodoc when function wrapped by @t…

engineerchuan:
We use sphinx to auto generate docs.

Some of these files are wrapped in decorator @torch.jit.script.

To adjust for this, for these functions, we explicitly specify the signature in the first line of the docstring.  

N.B. Maybe there could be a better way to introspect the wrapper decorator but could not figure out:  https://stackoverflow.com/questions/3687046/python-sphinx-autodoc-and-decorated-members

Tested locally by generating the HTML docs, look good.  Will do a separate PR to check them into the gh-pages branch.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -176,7 +176,10 @@ def istft(stft_matrix,          # type: Tensor
 @torch.jit.script
 def spectrogram(waveform, pad, window, n_fft, hop_length, win_length, power, normalized):
     # type: (Tensor, int, Tensor, int, int, int, int, bool) -> Tensor
-    r"""Create a spectrogram from a raw audio signal.
+    r"""
+    spectrogram(waveform, pad, window, n_fft, hop_length, win_length, power, normalized)
+    
+    Create a spectrogram from a raw audio signal.
 
     Args:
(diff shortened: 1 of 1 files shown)


engineerchuan:
This is related to issue:  https://github.com/pytorch/audio/issues/278

engineerchuan:
image

engineerchuan:
This seems related to https://github.com/pytorch/audio/issues/165 although we are not trying to document a Class method.  @jamarshon  is there a better way to do this?

vincentqb:
Do you have an example of where we do this elsewhere in pytorch?

engineerchuan:
I do not unfortunately.  There are not many standalone methods that are decorated by @torch.jit.script.   I checked out vision, text, and pytorch main.

vincentqb:
This is also what's done [here](https://github.com/pytorch/pytorch/blob/master/torch/_torch_docs.py).

vincentqb:
I'll merge this for now. In the longer term, the solution will be [this](https://github.com/pytorch/audio/issues/278#issuecomment-533618910). Let's wait for the next release to render the documentation

Issue #259, opened 2019-08-28
Merging items torchaudio-contrib that are not yet in torchaudio


vincentqb:
When looking into [torchaudio-contrib](https://github.com/keunwoochoi/torchaudio-contrib/tree/master/torchaudio_contrib), I see the following items that are not into torchaudio:
- [ ] [Harmonic-Percussive Source Separation](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/beta_hpss.py)
- [ ] DB to Amplitude [functional](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/functional.py#L299) and [layer](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/layers.py#L384) (though we have Amplitude to DB)
- [ ] [mel-hertz functionals](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/functional.py#L5)
- [ ] ~[Complex Norm layer](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/layers.py#L112) (in discussion in pytorch/pytorch#755)~
- [ ] [Time stretch layer](https://github.com/keunwoochoi/torchaudio-contrib/blob/master/torchaudio_contrib/layers.py#L112) but phase vocoder functional is available

See 
* torchaudio-contrib merging plans in #110 
* keunwoochoi/torchaudio-contrib#61
* keunwoochoi/torchaudio-contrib#71
* sox discussion in #260.

keunwoochoi:
Right that’s how we planed. But I guess we can be more flexible though? At the moment I don’t have a strong opinion about doing it here first :)

vincentqb:
We're also flexible here, and want to be compatible with your workflow. We're really excited about your work, so we want to make sure you can have as much impact as you want :)

ksanjeevan:
@vincentqb great! So I'll make a pull request with the four above and we can all have a discussion in the the context of the PR for changes like where to put them, pitch shift stuff, etc. 👍

keunwoochoi:
@ksanjeevan Perfect, that'd be great! One thing, for a few reasons (we might need to make some changes and etc just to make sure), HPSS would fit with me better. I'd happily do it :)

PR #285, opened 2019-09-17
torchaudio-contrib: some augmentations

ksanjeevan:
**Adding augmentations: `TimeStretch`, `TimeMasking` and `FrequencyMasking`.**

@vincentqb @cpuhrsch @keunwoochoi
## On complex STFT

We had discussions about this in torchaudio-contrib, about how it's useful to have a transform that outputs the complex stft. So basically the same as the [current](https://github.com/pytorch/audio/blob/a2d35df2e4d1013f3a58b2726f7b92074d842c78/torchaudio/transforms.py#L22) `Spectrogram` but without normalizing the output. In our previous PR we added the [`complex_norm`](https://github.com/pytorch/audio/blob/a2d35df2e4d1013f3a58b2726f7b92074d842c78/torchaudio/functional.py#L361) to the functionals so then the `Spectrogram` transform could just call the complex stft functional (which will basically just wrap `torch.stft` plus whatever padding we're currently doing, batching (?)) and then pass the output to `complex_norm` functional. 
An example where this can be needed is for example with the introduction of the `TimeStretch` layer. It takes as input a complex spectrogram so if a user wanted to make a module where a spectrogram is computed and stretched it could look something like:
```python
model = nn.Sequential( STFT(n_fft=n_fft, hop_length=hop_length),
                           TimeStretch(freq=num_freqs, hop_length=hop_length, fixed_rate=1.3),
                           ComplexNorm(power=2.0),
                           AmplitudeToDB() )
```
I have added a `STFT` transform as well as a `stft` functional to be able to showcase the `TimeStretch` augmentation (we have something like this in [torchaudio-contrib](https://github.com/keunwoochoi/torchaudio-contrib/blob/61fc6a804c941dec3cf8a06478704d19fc5e415a/torchaudio_contrib/layers.py#L35)), but of course this should be reworked once we get some feedback!

## On batching

I've made `STFT` be able to handle a batch dimension, and so can the augmentation layers. I can't seem to find it but I remember a discussion on how the layers should handle batch of inputs. Should the layers apply transforms only to single inputs for now? I can make changes if that is the case.

## On the augmentations
So we talked about what augmentations to do in #259 about doing `TimeStretch`, `TimeMasking`, `FrequencyMasking` and `PitchShift` but I've only included the first three in this PR since pitch shifting I think will require a bit more discussion and this can be a solid first step. I haven't included the tests for them yet since I figured many things might change after feedback is given. A quick summary of what they do:

### Time Stretching
Simply wrapping and allowing for batching for the `phase_vocoder` functional [introduced in the last PR](https://github.com/pytorch/audio/pull/131). A fixed rate can be passed when initializing, or a (randomly generated if doing augmentation) rate can be given to the forward method each time. Looks like:
pv

### Time and Frequency masking
From [SpecAugment](https://ai.googleblog.com/2019/04/specaugment-new-data-augmentation.html), apply masks of a desired value to an input spectrogram. I've included a flag in case the input is batched and the user wants the masks to be independent of each other. Examples:
Same mask for batch:
no_iid
Independent mask for batch:
iid

Some demo code:
```python
num_freqs, hop_length = 400, 512
model = nn.Sequential( STFT(n_fft=(num_freqs - 1) * 2, hop_length=hop_length),
                           TimeStretch(freq=num_freqs, hop_length=hop_length, fixed_rate=1.3),
                           ComplexNorm(power=2.0),
                           FrequencyMasking(freq_mask_param=60, iid_masks=False),
                           TimeMasking(time_mask_param=30, iid_masks=False),
                           AmplitudeToDB() )
inp = torch.as_tensor(librosa.load('file.wav')[0]).view(1, -1)
out = model(inp)
plot_heatmap(out)
```
gives:
plot


torchaudio/functional.py
```diff
@@ -466,3 +468,104 @@ def phase_vocoder(complex_specgrams, rate, phase_advance):
     complex_specgrams_stretch = torch.stack([real_stretch, imag_stretch], dim=-1)
 
     return complex_specgrams_stretch
+
+
+@torch.jit.script
+def stft(waveform, pad, window, n_fft, hop_length, win_length):
```
cpuhrsch: stft is already in pytorch core. I know that not having batching is annoying but replicating a function across both libraries is too. We're working on a more principled abstraction around batching, but we'd like to avoid having a reimplementation of this. Maybe we can do this in a separate PR?

ksanjeevan: >We're working on a more principled abstraction around batching, but we'd like to avoid having a reimplementation of this. Maybe we can do this in a separate PR?

👍 

Gotcha, so I can take this out no problem and then have the `STFT` layer just use `torch.stft`? I think the bigger point for this was to have the non-normalized output of stft as a functional/layer.

vincentqb: > ## On batching
> I've made `STFT` be able to handle a batch dimension, and so can the augmentation layers. I can't seem to find it but I remember a discussion on how the layers should handle batch of inputs. Should the layers apply transforms only to single inputs for now? I can make changes if that is the case.

In this PR, we can focus on single inputs in order to avoid adding STFT here again, as mentioned by @cpuhrsch. For batching, I would prefer waiting on a standardized approach. If that is so important, I'd open a separate PR for that so we don't block the rest :)

vincentqb:
> I've made `STFT` be able to handle a batch dimension, and so can the augmentation layers. I can't seem to find it but I remember a discussion on how the layers should handle batch of inputs. Should the layers apply transforms only to single inputs for now? I can make changes if that is the case.

In this PR, we can focus on single inputs in order to avoid adding STFT here again, as mentioned by @cpuhrsch. For batching, I would prefer waiting on a standardized approach. If that is so important, I'd open a separate PR for that so we don't block the rest :)

vincentqb:
> We had discussions about this in torchaudio-contrib, about how it's useful to have a transform that outputs the complex stft. So basically the same as the [current](https://github.com/pytorch/audio/blob/a2d35df2e4d1013f3a58b2726f7b92074d842c78/torchaudio/transforms.py#L22) `Spectrogram` but without normalizing the output. In our previous PR we added the [`complex_norm`](https://github.com/pytorch/audio/blob/a2d35df2e4d1013f3a58b2726f7b92074d842c78/torchaudio/functional.py#L361) to the functionals so then the `Spectrogram` transform could just call the complex stft functional (which will basically just wrap `torch.stft` plus whatever padding we're currently doing, batching (?)) and then pass the output to `complex_norm` functional.

Quick thought: I'm not necessarily suggesting to change something, but maybe `Spectrogram` should have been two operations then :)

ksanjeevan:
> Quick thought: I'm not necessarily suggesting to change something, but maybe `Spectrogram` should have been two operations then :)

Yeah that's actually how we have it in -contrib. The layers are STFT, ApplyFilterbank, TimeStretch, etc. So then if say you want a db melspectrogram, simply chain `nn.Sequential(STFT, ComplexNorm, ApplyFilterbank, AmplitudeToDb)`. So we don't have an explicit Spectrogram layer, although I guess having both options can be ok?

vincentqb:
> Yeah that's actually how we have it in -contrib. The layers are STFT, ApplyFilterbank, TimeStretch, etc. So then if say you want a db melspectrogram, simply chain `nn.Sequential(STFT, ComplexNorm, ApplyFilterbank, AmplitudeToDb)`. So we don't have an explicit Spectrogram layer, although I guess having both options can be ok?

Let's write the implementations here as if Spectrogram does not exist. We can then decide to axe it, keep it, or turn it into a wrapper, keeping in mind backward compatibility.

torchaudio/functional.py
vincentqb: Are you saying there's an operation provided by torchaudio that normalizes the output when you would like it not to?

ksanjeevan: Yeah, so [`Spectrogram`](https://github.com/pytorch/audio/blob/8273c3f4500ac9c01e95f1d78fa3f5752aafdca7/torchaudio/transforms.py#L22) calls the `spectrogram` functional which is basically computing the stft and then [getting the power of the complex tensor](https://github.com/pytorch/audio/blob/8273c3f4500ac9c01e95f1d78fa3f5752aafdca7/torchaudio/functional.py#L247) (saying "normalizing" might have confusing on my part, sorry).

But in order to use the phase vocoder, the `TimeStretch` transform has to be applied to the output of the `stft` before the complex norm (i.e. before doing`spec_f = spec_f.pow(power).sum(-1)`).

So that's why I was kinda wanting torchaudio to have a transform that gives the complex `STFT` like we have in contrib, because if someone then wants to work with randomly stretched spectrograms, they can simply do:
```python
nn.Sequential(STFT, TimeStretch, ComplexNorm, ...)
```
whereas currently `STFT` and `ComplexNorm` are "coupled" together as `Spectrogram`, and we can't put the time stretching in between (i.e. `nn.Sequential(Spectrogram, TimeStretch, ...)` won't work).

ksanjeevan:
So I've removed the functional and transform *complex-and-also-batched* `stft`, per @cpuhrsch and @vincentqb comments. Also added some tests for the masking functionals. Are the checks I wrote enough? I can add comments and/or more checks if needed. Should the next steps be writing some preliminary tests for the augmentation layers? I can also put a gist together showing how to chain these transforms.

test/test_functional.py
```diff
@@ -251,5 +250,47 @@ def test_complex_norm(complex_tensor, power):
...
+    torch.randn(4, 2, 1025, 400),
+])
+@pytest.mark.parametrize('mask_param', [100])
+@pytest.mark.parametrize('mask_value', [0., 30.])
+@pytest.mark.parametrize('axis', [2, 3])
+def test_mask_along_axis_iid(specgrams, mask_param, mask_value, axis):
+
+    mask_specgrams = F.mask_along_axis_iid(specgrams, mask_param, mask_value, axis)
+
+    torch.save(mask_specgrams, 'ex.pth')
```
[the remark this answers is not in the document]
ksanjeevan: Oh no, that was left over from when I was debugging some stuff my bad! Fixed.

ksanjeevan:
> Is there something in SpecAugment we could compare against? Otherwise, the tests and the demo you provided here seems good to me.

Hmm I'll look into it! I think the check we have to make sure the masked # of columns is < the mask parameter is a good start but there may be other things to compare with.

> If you have a demo of some code that could be fun to show, it could be shown as a new torchaudio tutorial, e.g. here, or example. :) These are useful since they also provide an integration test. This can be done as a separate PR.

Yeah absolutely. Happy to do a separate PR with a demo/tutorial once we've sorted out the `STFT` stuff (just so it's cleaner for the user).

vincentqb:
PS: let's not worry about this in this PR, but we try to make the transforms be thin wrappers around functionals as much as possible. :)

PR #286, opened 2019-09-17
istft: torch.eye dtype conforms to stft_matrix.

zhanwenchen:
In `torchaudio.functional.istft`, `eye = torch.eye` should manually be typecast to that of `stft_matrix` or `ytmp`, because otherwise, when `stft_matrix`/`ytmp` is `torch.float64`, `eye` still defaults to `torch.float32` and causes a `RuntimeError: expected scalar type Double but found Float`

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -137,7 +137,7 @@ def istft(stft_matrix,          # type: Tensor
     # each column of a channel is a frame which needs to be overlap added at the right place
     ytmp = ytmp.transpose(1, 2)  # size (channel, n_fft, n_frames)
 
-    eye = torch.eye(n_fft, requires_grad=False,
+    eye = torch.eye(n_fft, requires_grad=False, dtype=ytmp.dtype,
                     device=device, dtype=dtype).unsqueeze(1)  # size (n_fft, 1, n_fft)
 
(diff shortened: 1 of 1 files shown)


engineerchuan:
Hi Phil,

Thank you for submitting this PR.

I tried to reproduce this and am not getting the same error that you are.  Specifically, when stft_matrix is of dtype torch.float64, this is saved in the `dtype` variable in line 99.  Then, when `eye` is instantiated, it also is cast into this in line 141.

Am I missing something or do we have some version skew?  Would you mind submitting your calling code snippet?

Thanks,

Chuan

```
        # https://github.com/pytorch/audio/pull/286
        stft = torch.ones((2, 257, 1000, 2), dtype=torch.float64)
        torchaudio.functional.istft(stft, 512)

```

zhanwenchen:
Hi @engineerchuan. After digging into the commit history, it looks like the this problem was fixed in #135. I do not believe this commit has been included in the last release however. So the problem still exists with the official 0.3.0 release.

engineerchuan:
Thanks @zhanwenchen.  Feel free to use master to get around this until the next release comes out.  

I'm sure you know this but you can clone, cd into the repo and do a  `pip install .` to install the latest.

PR #291, opened 2019-09-19
Fix lfilter for GPU machine

engineerchuan:
I took master and tested on a GPU machine, passing in inputs that are tensors on GPU.  Failed due to internal tensors being created on CPU.

Fixed this, and added testing for GPU path if available.

### DIFF
--- a/test/test_functional_filtering.py
+++ b/test/test_functional_filtering.py
@@ -12,7 +12,7 @@ import time
 class TestFunctionalFiltering(unittest.TestCase):
     test_dirpath, test_dir = common_utils.create_temp_assets_dir()
 
-    def _test_lfilter_basic(self, device):
+    def _test_lfilter_basic(self, dtype, device):
         """
         Create a very basic signal,
         Then make a simple 4th order delay
@@ -20,25 +20,22 @@ class TestFunctionalFiltering(unittest.TestCase):
         """
 
         torch.random.manual_seed(42)
-        waveform = torch.rand(2, 44100 * 1, device=device)
-        b_coeffs = torch.tensor([0, 0, 0, 1], dtype=torch.float32, device=device)
-        a_coeffs = torch.tensor([1, 0, 0, 0], dtype=torch.float32, device=device)
+        waveform = torch.rand(2, 44100 * 1, dtype=dtype, device=device)
+        b_coeffs = torch.tensor([0, 0, 0, 1], dtype=dtype, device=device)
+        a_coeffs = torch.tensor([1, 0, 0, 0], dtype=dtype, device=device)
         output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
 
         assert torch.allclose(waveform[:, 0:-3], output_waveform[:, 3:], atol=1e-5)
 
     def test_lfilter_basic(self):
-        """
-        Will a basic difference engine run, on CPU
-        and GPU (if available)
-        """
+        self._test_lfilter_basic(torch.float32, torch.device("cpu"))
 
-        cpu = torch.device("cpu")
-        self._test_lfilter_basic(cpu)
+    def test_lfilter_basic_double(self):
+        self._test_lfilter_basic(torch.float64, torch.device("cpu"))
 
+    def test_lfilter_basic_gpu(self):
         if torch.cuda.is_available():
-            cuda0 = torch.device("cuda:0")
-            self._test_lfilter_basic(cuda0)
+            self._test_lfilter_basic(torch.float32, torch.device("cuda:0"))
         else:
             print("skipping GPU test for lfilter_basic because device not available")
             pass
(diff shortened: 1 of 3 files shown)


torchaudio/functional.py
```diff
@@ -524,34 +524,38 @@ def lfilter(waveform, a_coeffs, b_coeffs):
 
     Returns:
         output_waveform (torch.Tensor): Dimension of `(n_channel, n_frames)`.  Output will be clipped to -1 to 1.
+                                        Will be on the same device as the inputs.
 
     """
 
     assert(waveform.dtype == torch.float32)
```
vincentqb: Why do we limit to `float32`? It might be a better idea to pass the `dtype` to the different computations.

vincentqb: Same below. By the way, let's be careful when we create other tensors to :)

test/test_functional_filtering.py
```diff
@@ -59,16 +74,31 @@ def test_lfilter(self):
...
 
+    def test_lfilter(self):
+
+        filepath = os.path.join(self.test_dirpath, "assets", "whitenoise.mp3")
+        waveform, _ = torchaudio.load(filepath, normalization=True)
+
+        cpu = torch.device("cpu")
+        self._test_lfilter(waveform, cpu)
+
+        if torch.cuda.is_available():
```
vincentqb: Let's make that a separate test.

test/test_functional_filtering.py
```diff
@@ -12,24 +12,38 @@
...
+    def test_lfilter_basic(self):
+        """
+        Will a basic difference engine run, on CPU
+        and GPU (if available)
+        """
+
+        cpu = torch.device("cpu")
+        self._test_lfilter_basic(cpu)
+
+        if torch.cuda.is_available():
```
vincentqb: Let's make this a separate test

torchaudio/functional.py
```diff
@@ -524,34 +524,38 @@ def lfilter(waveform, a_coeffs, b_coeffs):
...
     assert(a_coeffs.size(0) == b_coeffs.size(0))
     assert(len(waveform.size()) == 2)
```
vincentqb: Is that something mandatory?

torchaudio/functional.py
```diff
@@ -524,34 +524,38 @@ def lfilter(waveform, a_coeffs, b_coeffs):
...
+    assert(b_coeffs.device == a_coeffs.device)
 
+    device = waveform.device
     n_channels, n_frames = waveform.size()
     n_order = a_coeffs.size(0)
     assert(n_order > 0)
 
     # Pad the input and create output
-    padded_waveform = torch.zeros(n_channels, n_frames + n_order - 1)
+    padded_waveform = torch.zeros(n_channels, n_frames + n_order - 1, device=device)
```
vincentqb: Let's pass also the dtype being used, here and below.

torchaudio/functional.py
vincentqb: In a similar vein, we should probably avoid using the math library were we can use pytorch, e.g. [here](https://github.com/pytorch/audio/blob/c9926814cd57fa8dc176fe1387ff05eb60fef5fe/torchaudio/functional.py#L599).

torchaudio/functional.py
```diff
@@ -524,34 +524,38 @@ def lfilter(waveform, a_coeffs, b_coeffs):
 
     Returns:
         output_waveform (torch.Tensor): Dimension of `(n_channel, n_frames)`.  Output will be clipped to -1 to 1.
+                                        Will be on the same device as the inputs.
```
vincentqb: This is usually assumed. We probably don't have to say it here, but it's good to keep in mind :)

torchaudio/functional.py
engineerchuan: No, this is probably the easiest way.  We could also have the user specify where they want the output or have the user pass in an output container.

torchaudio/functional.py
engineerchuan: I'd prefer not to cast to a tensor just to do an exp.  This is done very few times, just in the filter design portion.  Also `math` is a python standard library.  https://pytorch.org/docs/stable/torch.html?highlight=exp#torch.exp

torchaudio/functional.py
```diff
@@ -584,15 +588,19 @@ def biquad(waveform, b0, b1, b2, a0, a1, a2):
...
-        waveform, torch.tensor([a0, a1, a2]), torch.tensor([b0, b1, b2])
+        waveform,
+        torch.tensor([a0, a1, a2], dtype=dtype, device=device),
+        torch.tensor([b0, b1, b2], dtype=dtype, device=device)
     )
     return output_waveform
 
 
 def _dB2Linear(x):
+    import math
```
vincentqb: nit: let's keep the import at the beginning :)

torchaudio/functional.py
vincentqb: We don't need to worry about this here. This can be as part of a separate PR.

torchaudio/functional.py
vincentqb: I see, they are used only for scalars. I agree that's fine.

Issue #292, opened 2019-09-20
Incorrect mel filterbank when f_max != sample_rate // 2


jongwook:
## 🐛 Bug

`MelScale` and thus `MelSpectrogram` will not use the correct filterbank when `f_max` is not equal to `sample_rate // 2`.

## To Reproduce

The following should give the same filterbank matrix:

```python
# Equivalent filterbank construction by Librosa
fb = librosa.filters.mel(sr=22050, n_fft=2048, n_mels=128, fmin=0.0, fmax=8000, htk=True, norm=None)

# TorchAudio's Filterbank
fb = torchaudio.transforms.MelScale(sample_rate=22050, n_mels=128, f_max=8000.0, n_stft=1025).fb
```

## Expected behavior

Librosa's Mel filterbank stops at certain frequency bin, since `f_max` is lower than the Nyquist frequency.

![image](https://user-images.githubusercontent.com/266841/65362593-dd9e8980-dbbc-11e9-8e5d-550ccdc679b8.png)

`MelScale`'s filter will fill up until the Nyquist frequency no matter what:

![image](https://user-images.githubusercontent.com/266841/65362642-032b9300-dbbd-11e9-9381-4ad7b0f2995c.png)

The function `F.create_fb_matrix` uses `f_max` but places at the last frequency bin regardless of `sample_rate`. This ends up creating filters in a wrong Mel scale.

## Environment

Same as #287 

## Additional context

Currently we can work around this bug by assigning Librosa's filterbank in TorchAudio module's filterbank attribute.

engineerchuan:
Hi,

I was able to confirm this issue.  I have created a pull request to fix this.  I ran the following comparison code on the PR and see similar results.

```
# Librosa vs Pytorch
import torchaudio 
import librosa
import matplotlib.pyplot as plt

SAMPLE_RATE = 22050
N_MEL = 10
FMIN = 2000.
FMAX = 8000.

plt.subplot(2, 1, 1)
plt.title('LibRosa Filter Bank')
librosa_fb = librosa.filters.mel(sr=SAMPLE_RATE, n_fft=2048, n_mels=N_MEL, fmin=FMIN, fmax = FMAX, htk=True, norm=None)
print("Librosa Filter Bank is of size: ", len(librosa_fb), len(librosa_fb[0]))
for i in range(len(librosa_fb)):
    plt.plot(librosa_fb[i])


plt.subplot(2, 1, 2)
plt.title('Torchaudio Filter Bank')
[...]

PR #294, opened 2019-09-22
Fix mel filter bank

engineerchuan:
Fixes to address issue https://github.com/pytorch/audio/issues/292

### DIFF
--- a/test/test_jit.py
+++ b/test/test_jit.py
@@ -59,17 +59,18 @@ class Test_JIT(unittest.TestCase):
 
     def test_torchscript_create_fb_matrix(self):
         @torch.jit.script
-        def jit_method(n_stft, f_min, f_max, n_mels):
-            # type: (int, float, float, int) -> Tensor
-            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels)
+        def jit_method(n_stft, f_min, f_max, n_mels, sample_rate):
+            # type: (int, float, float, int, int) -> Tensor
+            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate)
 
         n_stft = 100
         f_min = 0.
         f_max = 20.
         n_mels = 10
+        sample_rate = 16000
 
-        jit_out = jit_method(n_stft, f_min, f_max, n_mels)
-        py_out = F.create_fb_matrix(n_stft, f_min, f_max, n_mels)
+        jit_out = jit_method(n_stft, f_min, f_max, n_mels, sample_rate)
+        py_out = F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate)
 
         self.assertTrue(torch.allclose(jit_out, py_out))
 
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -313,7 +313,7 @@ def create_fb_matrix(n_freqs, f_min, f_max, n_mels, sample_rate):
     """
     # freq bins
     # Equivalent filterbank construction by Librosa
-    all_freqs = torch.linspace(0.0, float(sample_rate) / 2., n_freqs)
+    all_freqs = torch.linspace(0, sample_rate // 2, n_freqs)
     i_freqs = all_freqs.ge(f_min) & all_freqs.le(f_max)
     freqs = all_freqs[i_freqs]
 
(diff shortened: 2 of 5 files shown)


test/test_jit.py
```diff
@@ -59,17 +59,18 @@ def test_scriptmodule_Spectrogram(self):
 
     def test_torchscript_create_fb_matrix(self):
         @torch.jit.script
-        def jit_method(n_stft, f_min, f_max, n_mels):
-            # type: (int, float, float, int) -> Tensor
-            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels)
+        def jit_method(n_stft, f_min, f_max, n_mels, sample_rate):
+            # type: (int, float, float, int, int) -> Tensor
```
vincentqb: I see `sample_rate` being converted to float later. Does it need to be an integer instead of a float?

torchaudio/functional.py
```diff
@@ -290,18 +290,19 @@ def amplitude_to_DB(x, multiplier, amin, db_multiplier, top_db=None):
...
+    create_fb_matrix(n_freqs, f_min, f_max, n_mels, sample_rate)
 
     Create a frequency bin conversion matrix.
 
     Args:
         n_freqs (int): Number of frequencies to highlight/apply
-        f_min (float): Minimum frequency
-        f_max (float): Maximum frequency
+        f_min (float): Minimum frequency (Hz)
+        f_max (float): Maximum frequency (Hz)
```
vincentqb: Ths units depend on the units of `sample_rate` which need not be in Hz, right?

torchaudio/functional.py
```diff
@@ -311,17 +312,21 @@ def create_fb_matrix(n_freqs, f_min, f_max, n_mels):
         ``A * create_fb_matrix(A.size(-1), ...)``.
     """
     # freq bins
-    freqs = torch.linspace(f_min, f_max, n_freqs)
+    # Equivalent filterbank construction by Librosa
+    all_freqs = torch.linspace(0.0, float(sample_rate) / 2., n_freqs)
```
vincentqb: This construction generates `n_freqs` bins between 0 and sample_rate/2. Since the line after filters out what's outside of `f_min` and `f_max`, we may end up with less than `n_freqs` bins, right? Is that expected?

Also, what happens if there are no bins between `f_min` and `f_max` in the range? Do we get an error?

torchaudio/functional.py
vincentqb: You are correct, that's indeed what we use in the readme. :) WIth/without is fine by me.

engineerchuan: @Vincent - I think it does have to be in Hz.  In the create_fb_matrix function there are hardcoded constants like 2595 which are conversions from HZ to Mel Scale.

test/test_jit.py
engineerchuan: I guess technically you could sample at a non integer rate per second but I've never seen this before - typically audio is always 8kHz, 44.1kHz, etc..

torchaudio/functional.py
engineerchuan: It is expected.  If we do a STFT with N = 2048, we should expect 1025 frequency bins corresponding to the entire frequency range from DC (0) to Nyquist.  We should not have 1025 bins between an arbitrary f_min and f_max set by the user.

Our code for slopes / zero / down_slopes / up_slopes is quite complicated but it will still create a size [n_freqs, n_mels] data structure that we have verified against librosa implementation in the test.

Yes in general, f_min and f_max can be set to bad values depending on what n_mels, n_freqs are and sample rate are.  Even if f_max > f_min but only slightly greater, we could return some results.  In this case we will still produce a fb matrix but some filters will be entirely zero.  This is consistent with librosa behavior and I have added a test for it. 

However, librosa will also output the following warning message ( n_mel is 56, fmin is 800, fmax is 900, n_fft is 1025 and sample_rate is 22050.)

```
Empty filters detected in mel frequency basis. Some channels will produce empty responses. Try increasing your sampling rate (and fmax) or reducing n_mels.
  warnings.warn('Empty filters detected in mel frequency basis. '
```

test/test_jit.py
vincentqb: In any case, this would be a discussion for elsewhere :) `sample_rate` is consistently an `int` elsewhere.

vincentqb:
One test is failing due to a timeout. I'll go ahead and merge.

PR #296, opened 2019-09-24
Fix curl failure

engineerchuan:
Tests are failing because downloading flac has some certificate expired.

When looking at the download.

A.) Why is flac being downloaded from some other subdomain other than downloads.sourceforge
B.) Why are sox and lame not using https.

Trying this one to see if it works.

Chuan


engineerchuan:
create a new pull request.

PR #293, opened 2019-09-22
Small fixes to README.md

engineerchuan:
`n_channels` and `n_samples` will come up a lot.  We should put this in the Guidelines.

PR #298, opened 2019-09-24
standardizing n_*

vincentqb:
We want to clarify the naming of `n_*` variables, initiated by #293. Documentation of standardization in [README](https://github.com/pytorch/audio/blob/master/README.md).

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -83,7 +83,7 @@ def istft(
     Example: Suppose the last window is:
     [17, 18, 0, 0, 0] vs [18, 0, 0, 0, 0]
 
-    The n_frames, hop_length, win_length are all the same which prevents the calculation of right padding.
+    The n_frame, hop_length, win_length are all the same which prevents the calculation of right padding.
     These additional values could be zeros or a reflection of the signal so providing ``length``
     could be useful. If ``length`` is ``None`` then padding will be aggressively removed
     (some loss of signal).
@@ -93,8 +93,8 @@ def istft(
 
     Args:
         stft_matrix (torch.Tensor): Output of stft where each row of a channel is a frequency and each
-            column is a window. it has a size of either (channel, fft_size, n_frames, 2) or (
-            fft_size, n_frames, 2)
+            column is a window. it has a size of either (channel, fft_size, n_frame, 2) or (
+            fft_size, n_frame, 2)
         n_fft (int): Size of Fourier transform
         hop_length (Optional[int]): The distance between neighboring sliding window frames.
             (Default: ``win_length // 4``)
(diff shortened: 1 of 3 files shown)


vincentqb:
We should be careful about the distinction between samples and frames. The latter is often a small group of samples.

vincentqb:
Moved to #293

torchaudio/functional.py
```diff
@@ -537,7 +537,7 @@ def lfilter(waveform, a_coeffs, b_coeffs):
     Performs an IIR filter by evaluating difference equation.
 
     Args:
-        waveform (torch.Tensor): audio waveform of dimension of `(n_channel, n_frames)`.  Must be normalized to -1 to 1.
+        waveform (torch.Tensor): audio waveform of dimension of `(channel, time)`.  Must be normalized to -1 to 1.
```
zhangguanheng66: Should this be `frame'.

torchaudio/functional.py
```diff
@@ -558,37 +557,37 @@ def lfilter(waveform, a_coeffs, b_coeffs):
 
     device = waveform.device
     dtype = waveform.dtype
-    n_channels, n_frames = waveform.size()
+    n_channel, n_sample = waveform.size()
```
zhangguanheng66: For consistency, `channel` and `frame`, rather than `n_channel` and `n_sample`?

torchaudio/functional.py
```diff
@@ -558,37 +557,37 @@ def lfilter(waveform, a_coeffs, b_coeffs):
...
     n_order = a_coeffs.size(0)
     assert(n_order > 0)
 
     # Pad the input and create output
-    padded_waveform = torch.zeros(n_channels, n_frames + n_order - 1, dtype=dtype, device=device)
+    padded_waveform = torch.zeros(n_channel, n_sample + n_order - 1, dtype=dtype, device=device)
```
zhangguanheng66: Same as above.

torchaudio/functional.py
```diff
@@ -558,37 +557,37 @@ def lfilter(waveform, a_coeffs, b_coeffs):
...
     padded_waveform[:, (n_order - 1):] = waveform
-    padded_output_waveform = torch.zeros(n_channels, n_frames + n_order - 1, dtype=dtype, device=device)
+    padded_output_waveform = torch.zeros(n_channel, n_sample + n_order - 1, dtype=dtype, device=device)
```
zhangguanheng66: Same as above.

torchaudio/functional.py
```diff
@@ -558,37 +557,37 @@ def lfilter(waveform, a_coeffs, b_coeffs):
...
-    a_coeffs_filled = a_coeffs.flip(0).repeat(n_channels, 1).t()
-    b_coeffs_filled = b_coeffs.flip(0).repeat(n_channels, 1).t()
+    a_coeffs_filled = a_coeffs.flip(0).repeat(n_channel, 1).t()
+    b_coeffs_filled = b_coeffs.flip(0).repeat(n_channel, 1).t()
 
     # Set up a few other utilities
-    a0_repeated = torch.ones(n_channels, dtype=dtype, device=device) * a_coeffs[0]
-    ones = torch.ones(n_channels, n_frames, dtype=dtype, device=device)
+    a0_repeated = torch.ones(n_channel, dtype=dtype, device=device) * a_coeffs[0]
+    ones = torch.ones(n_channel, n_sample, dtype=dtype, device=device)
```
zhangguanheng66: I saw `frame`, `time`, `sample` are used. Do they refer to the same thing?

torchaudio/functional.py
vincentqb: We use `n_*` to refer to the actual value, and `*` to the name of the dimension. Based on this convention, this would be ok. We could decide to change the convention itself, but I'd leave that to a separate PR if at all.

torchaudio/functional.py
vincentqb: Before this PR, `time` was used to indicate time dimension. `n_frames` and `n_samples` were used inconsistently to refer to the number of time sample available. Here, I'm suggesting to use `frame` (and `n_frame`) to refer to "groups of samples" (e.g. in STFT, we have rolling windows of samples to operate on), and "n_sample" to the number of samples in a "time" dimension.

We could also decide to change "time" dimension to "sample" dimension, but that does not seem to have been a source of confusion, so I would avoid changing this in this PR.

PR #303, opened 2019-10-05
new dataset format with librispeech and commonvoice

vincentqb:
Introducing a new dataset format based on generators, along with two new datasets: LibriSpeech and CommonVoice.

Related to
* pytorch/pytorch#24915
* keunwoochoi/torchaudio-contrib#67
* [notebook for experiment](https://github.com/vincentqb/audio-tutorial/blob/master/datasets.ipynb)


torchaudio/datasets/commonvoice.py
```diff
@@ -0,0 +1,87 @@
...
+    Input: path, file name identifying a row of data
+    Output: dictionary with keys from tsv along with, waveform, sample_rate
+    """
+
+    for path, fileid in fileids:
+        filename = os.path.join(path, "clips", fileid)
+        tsv = os.path.join(path, tsv)
+
+        found = False
+        with open(tsv) as tsv:
```
cpuhrsch: You can use csvreader for this or even better write a unicode_csv_reader akin to https://github.com/pytorch/text/blob/fd31bf3722e17dfd5998b8c4f32c0431f3016d59/torchtext/utils.py#L108 (which we still need because of Python2). 

Only difference is that the delimiter is a tab instead of a comma.

cpuhrsch: It also has support for headers and such.

cpuhrsch: https://docs.python.org/3/library/csv.html

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        item = self._cache.pop(0)
+        self._fill()
+        return item
+
+
+def download(urls, root_path):
```
cpuhrsch: If you want to write an abstraction that explicitly downloads a file to disk I'd call this "download_to_file".

In general a generic download function that returns a stream is a bit tricky, because if a user doesn't consume the stream the connection might break down (?). Also when writing a file straight to disk from the web we can potentially bypass system memory.

However, having something that does indeed give you a stream of data from the web is very useful.

If you want to get fancy you could also try to auto-discover the filename to be used as in here: https://github.com/pytorch/text/blob/fd31bf3722e17dfd5998b8c4f32c0431f3016d59/torchtext/utils.py#L33

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+    Input: url generator, folder inside archive
+    Output: downloaded archive, folder inside archive
+    """
+    for url, folder in urls:
+        torchaudio.datasets.utils.download_url(url, root_path)
+        file = os.path.join(root_path, os.path.basename(url))
+        yield file, folder
+
+
+def extract(files):
```
cpuhrsch: You should add a flag to have the user choose a format and also I don't think you need to make this specific to files. Have a look at https://github.com/pytorch/text/blob/fd31bf3722e17dfd5998b8c4f32c0431f3016d59/torchtext/utils.py#L152 , which however, regrettably, still requires files.

For an example of how to extract "on-the-fly" from an arbitrary file / buffer check out this snippet

```
def get_data(URL):
    r = requests.get(URL)
    file_like_object = io.BytesIO(r.content)
    tar = tarfile.open(fileobj=file_like_object)
    d = {}
    for member in tar.getmembers():
        if member.isfile() and member.name.endswith('csv'):
            k = 'train' if 'train' in member.name else 'test'
            d[k] = tar.extractfile(member)
    return d```

With something generic like this users can do

```
for tar_buffer in extract(open("/path/to/file.tar")
    tar_buffer.write_to_disk("/path...")
```

In general https://docs.python.org/3/library/io.html is a good library to look at for this.

If you want to make it explicit that this extracts form and to a file, I'd change the name to extract_file_to_disk or such.

File: torchaudio/datasets/datasets.py [hunk 4]
```diff
@@ -0,0 +1,419 @@
...
+
+
+def walk(paths, extension):
+    """
+    Walk inside a path recursively to find all files with given extension.
+
+    Input: path
+    Output: path, file name identifying a row of data
+    """
+    for path in paths:
```
cpuhrsch: You can potentially write this as as a list expression, which might be more performant.

In general this function is more akin to the http://man7.org/linux/man-pages/man1/find.1.html tool, so that might be a better name?

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+        for dp, dn, fn in os.walk(path):
+            for f in fn:
+                if extension in f:
+                    yield path, f
+
+
+def shuffle(generator):
```
cpuhrsch: I don't think this is necessary, except that random.shuffle is annoying because it doesn't give you a reference to the shuffled object.

Instead of generator you can also say "iterable"

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+
+    # Load whole generator in memory
+    generator = list(generator)
+    # print(len(generator))
+    random.shuffle(generator)
+    for g in generator:
+        yield g
+
+
+def filtering(fileids, reference):
```
cpuhrsch: This is similar to https://linux.die.net/man/1/comm and can also be extended to work on generic streams of text plus one constant source. It's similar to set intersection etc.

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+        file = os.path.join(path, fileid)
+        waveform, sample_rate = torchaudio.load(file)
+        label = os.path.basename(fileid).split(".")[0].split("_")
+
+        yield {"label": label, "waveform": waveform, "sample_rate": sample_rate}
+
+
+def YESNO(root):
+    """
+    Cache a pipeline loading YESNO.
```
cpuhrsch: I'd say you're really building a pipeline here. The caching is optional and I'd expect that to see in a dataset class for convenience.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,341 @@
...
+        with open(file, "rb") as file:
+            item = pickle.load(file)
+        return item
+
+    def __len__(self):
+        # Return length of cache
+        return len(self._cache)
+
+
+class Buffer:
```
cpuhrsch: For this to provide a performance gain it will probably want to make use of multithreading. This does introduce various constraints on the given generator that need to be determined. I imagine this to be most useful if applied to function that perform IO operations such as reading from multiple files, downloading multiple URLs etc.

See the following snippet

```
import time
import concurrent.futures
import urllib.request

def read_50(fn):
    for _ in range(50):
        yield fn()

def read_fn():
    with urllib.request.urlopen(<your_image_url>) as response:
        return response.read()

def read_50_threaded(fn):
    with concurrent.futures.ThreadPoolExecutor(max_workers=5) as executor:
        futures = [executor.submit(fn) for _ in range(50)]
        for future in futures:
            yield future.result()

t1 = time.time()
results = list(read_50(read_fn))
t2 = time.time()

print("t: " + str(t2 - t1))

t1 = time.time()
results = list(read_50_threaded(read_fn))
t2 = time.time()

print("t: " + str(t2 - t1))
```

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,341 @@
...
+        self.capacity = capacity
+        self._cache = []
+        self._fill()
+
+    def _fill(self):
+        # TODO non-blocking after the first item
+        while len(self._cache) <= self.capacity:
+            self._cache.append(next(self.generator))
+
+    def __getitem__(self, n):
```
cpuhrsch: I'm surprised this has a getitem function if it is to store a certain amount of data returned asynchronously by a generator.

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+import random
+from functools import partial, reduce
+from warnings import warn
+
+import torch
+
+import torchaudio
+
+
+class Cache:
```
cpuhrsch: you could even make this a function annotations that assumes the input/output relationship is deterministic.

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+            item = self[self._internal_index]
+        else:
+            item = next(self.generator)
+
+            file = str(self._id) + "-" + str(len(self))
+            file = os.path.join(self.location, file)
+            self._cache.append(file)
+
+            os.makedirs(self.location, exist_ok=True)
+            with open(file, "wb") as file:
```
cpuhrsch: pickling and files are a relatively strong assumption. lots of things can't be pickled. in terms of caching things on list for each input argument, you might end up with *a lot* of small files for something like images. if you want to write a read-only file cache (which is what this kind of is) you want to have a single file you append to. but random seek is not very good when it comes to files.

i'd say for now we stick to in-memory only caching

cpuhrsch: Or if you want to serialize it, read out the entire generator into a list first and then save that, or append to an open file etc.

torchaudio/datasets/datasets.py
cpuhrsch: The other thing we (I) semi under-appreciate is the fact that classes can print details about the pipeline pretty nicely. So that's also a nice side-effect of using a class for this.

torchaudio/datasets/datasets.py
```diff
@@ -0,0 +1,419 @@
...
+        )
+    ]
+
+    path = download(url, root_path=root)
+    path = extract(path)
+    path = walk(path, extension=".flac")
+    path = shuffle(path)
+    data = load_librispeech(path)
+
+    return Cache(data, "tmp/")
```
cpuhrsch: I'd suggest we don't Cache this by default to enable streaming applications if people want. If it's past of a class we could make random access (caching) a flag.

torchaudio/datasets/legacy/vctk.py
```diff
@@ -0,0 +1,243 @@
...
+            for fname in fnames:
+                if is_audio_file(fname):
+                    path = os.path.join(root, fname)
+                    item = path
+                    audios.append(item)
+    return audios
+
+
+def read_audio(fp, downsample=True):
+    if downsample:
```
cpuhrsch: I wonder if we can use our resample transform or if this is necessary for this dataset

torchaudio/datasets/legacy/vctk.py
```diff
@@ -0,0 +1,243 @@
...
+        E.append_effect_to_chain("gain", ["-rh"])
+        E.append_effect_to_chain("dither", ["-s"])
+        sig, sr = E.sox_build_flow_effects()
+    else:
+        sig, sr = torchaudio.load(fp)
+    sig = sig.contiguous()
+    return sig, sr
+
+
+def load_txts(dir):
```
cpuhrsch: I think you could first run a "find" function to get a dictionary of file paths plus the corresponding text (in UTF-8) format.

Then somewhere you need a map between transcription files and audio files.

torchaudio/datasets/legacy/vctk.py
```diff
@@ -0,0 +1,243 @@
...
+            downloaded again. (Default: ``True``)
+        dev_mode(bool, optional): If true, clean up is not performed on downloaded
+            files.  Useful to keep raw audio and transcriptions. (Default: ``False``)
+    """
+    raw_folder = 'vctk/raw'
+    processed_folder = 'vctk/processed'
+    url = 'http://homepages.inf.ed.ac.uk/jyamagis/release/VCTK-Corpus.tar.gz'
+    dset_path = 'VCTK-Corpus'
+
+    def __init__(self, root, downsample=True, transform=None, target_transform=None, download=False, dev_mode=False):
```
cpuhrsch: I assume the transforms key are still part of the current API, otherwise, in the future, I'd suggest we not make transforms necessary for datasets at all, but instead provide an easy block that combines transforms.

torchaudio/datasets/librispeech.py
```diff
@@ -0,0 +1,72 @@
...
+        found = False
+        for line in open(file_text):
+            fileid_text, content = line.strip().split(" ", 1)
+            if fileid == fileid_text:
+                found = True
+                break
+        if not found:
+            from warnings import warn
+
+            warn("Translation not found for {}.".format(fileid))
```
cpuhrsch: Should this ever happen? If not I'd not check for existence at all and just have the open function error out.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,341 @@
...
+
+
+def check_md5(fpath, md5, **kwargs):
+    return md5 == calculate_md5(fpath, **kwargs)
+
+
+def check_integrity(fpath, md5=None):
+    if not os.path.isfile(fpath):
+        return False
+    if md5 is None:
```
cpuhrsch: Is this actually desired?

torchaudio/datasets/datasets.py
vincentqb: agree. this file wasn't supposed to be in the PR :)

torchaudio/datasets/legacy/vctk.py
vincentqb: This is part of the legacy datasets which I simply moved in a legacy folder.

torchaudio/datasets/datasets.py
vincentqb: For reference: [SpooledTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.SpooledTemporaryFile).

ryanleary:
For best efficiency (if that's a goal here), it might make some sense to format the audio such that utterances are shuffled and packed into one or more tars (or something like it) such that sequential i/o can be used. Especially for these larger datasets, dealing with thousands of files and accessing them randomly over network can become a significant challenge.

torchaudio/datasets/datasets.py
vincentqb: pytorch also has `torch.save`.

torchaudio/datasets/commonvoice.py
```diff
@@ -0,0 +1,92 @@
...
+from torchaudio.datasets.utils import (
+    download,
+    extract,
+    shuffle,
+    unicode_csv_reader,
+    walk,
+)
+
+
+def load_commonvoice_item(line, header, path, folder_audio):
```
cpuhrsch: This function is only called once. Is there an anticipation for it to be used more widely in the future? Otherwise I suggest you inline it given that it's quite short.

vincentqb: In all of those, I used a `load_thisdataset_item`, so that I decouple the "load this particular item" from `__getitem__` or `__iter__`` etc, and make the class easy to convert from an list to an iterable or something else. Thoughts?

torchaudio/datasets/librispeech.py
```diff
@@ -0,0 +1,81 @@
...
+            "test-other",
+            "train-clean-100",
+            "train-clean-360",
+            "train-other-500",
+        ]
+
+        assert selection in selections
+
+        ext_archive = ".tar.gz"
+        base = "http://www.openslr.org/resources/12/"
```
cpuhrsch: I suggest we split this out into a URL constant within the file that's referenced as the default value in a kwarg, but the user should be able to specify e.g. a mirror on S3 or within an internal host or maybe even a file

torchaudio/datasets/librispeech.py
```diff
@@ -0,0 +1,81 @@
...
+        "sample_rate": sample_rate,
+    }
+
+
+class LIBRISPEECH(data.IterableDataset):
+
+    _ext_txt = ".trans.txt"
+    _ext_audio = ".flac"
+
+    def __init__(self, root, selection):
```
cpuhrsch: select could also be a list. might be useful if someone wants to select all train datasets. root could also have a default value (something like './cached'), but this is a wider decision.

torchaudio/datasets/librispeech.py
```diff
@@ -0,0 +1,81 @@
...
+        archive = os.path.basename(url)
+        archive = os.path.join(root, archive)
+        self._path = os.path.join(root, folder_in_archive)
+
+        if not os.path.isdir(self._path):
+            if not os.path.isfile(archive):
+                torchaudio.datasets.utils.download_url(_url, root)
+            torchaudio.datasets.utils.extract_archive(archive)
+
+        walker = torchaudio.datasets.utils.walk_files(
```
cpuhrsch: Not sure walking here is the right approach. This is unfortunately one of those situations where the "input" to the pipeline is a list of paths that then are randomized (sometimes based on some per-path metadata). We could offer the user to provide these paths as an argument as well and reference the walk_files util as a means of retrieving them.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,500 @@
...
+            pbar.total = total_size
+        progress_bytes = count * block_size
+        pbar.update(progress_bytes - pbar.n)
+
+    return bar_update
+
+
+def calculate_md5(fpath, chunk_size=1024 * 1024):
+    md5 = hashlib.md5()
+    with open(fpath, "rb") as f:
```
cpuhrsch: I assume this is a copy. In general this shouldn't be restricted to files, but binary blobs. we might very easily want to provide this functionality to files downloaded on the fly.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,500 @@
...
+        for chunk in iter(lambda: f.read(chunk_size), b""):
+            md5.update(chunk)
+    return md5.hexdigest()
+
+
+def check_md5(fpath, md5, **kwargs):
```
cpuhrsch: This seems overkill

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,500 @@
...
+
+def _is_gzip(filename):
+    return filename.endswith(".gz") and not filename.endswith(".tar.gz")
+
+
+def _is_zip(filename):
+    return filename.endswith(".zip")
+
+
+def extract_archive(from_path, to_path=None, overwrite=False):
```
cpuhrsch: This doesn't need to be restricted to paths.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,500 @@
...
+        >>> from_path = './validation.tar.gz'
+        >>> to_path = './'
+        >>> torchtext.utils.download_from_url(url, from_path)
+        >>> torchtext.utils.extract_archive(from_path, to_path)
+    """
+
+    if to_path is None:
+        to_path = os.path.dirname(from_path)
+
+    if from_path.endswith((".tar.gz", ".tgz")):
```
cpuhrsch: It's possible to ```try...except``` opening something as a tar file. The tar library will read the first few bytes and then decided whether it's a tarfile or not. We shouldn't make this dependent on the extension.

torchaudio/datasets/utils.py
```diff
@@ -0,0 +1,500 @@
...
+                zfile.extract(file_, to_path)
+        return files
+
+    else:
+        raise NotImplementedError(
+            "We currently only support tar.gz, .tgz and zip achives."
+        )
+
+
+class Cache:
```
cpuhrsch: I'd give this a more specific name. Like "DiskCache".

torchaudio/datasets/yesno.py
```diff
@@ -1,150 +1,45 @@
...
+    return {"label": label, "waveform": waveform, "sample_rate": sample_rate}
 
-    def _check_exists(self):
-        return os.path.exists(os.path.join(self.root, self.processed_folder, self.processed_file))
 
-    def download(self):
-        """Download the yesno data if it doesn't exist in processed_folder already."""
-        from six.moves import urllib
-        import tarfile
+class YESNO(data.IterableDataset):
```
cpuhrsch: Can we change this back into a regular data.Dataset to maintain backwards compatibility? Same for the other 3 datasets, I think they can all be indexable, which then works with the Samplers etc. we have in pytorch.

torchaudio/datasets/commonvoice.py
vincentqb: thanks for pointing this out!

torchaudio/datasets/datasets.py
vincentqb: Agreed.

torchaudio/datasets/utils.py
vincentqb: Agreed.

torchaudio/datasets/yesno.py
vincentqb: Agreed. The datasets are naturally indexable, even without loading the data in memory.

torchaudio/datasets/utils.py
vincentqb: Good point.

cpuhrsch:
@ryanleary - I can't agree more that pipelining and chunking are necessary optimizations. For now we're aiming for decreased complexity and coverage for these new datasets without breaking backwards compatibility. After that we'll work on performance.

torchaudio/datasets/utils.py
vincentqb: Good point.

vincentqb:
I've defined the interface close to what was existing, threw a few deprecation warnings, and removed the "legacy" folder. @cpuhrsch thoughts?

PR #313, opened 2019-10-24
Pitch detection

### LINKED ISSUE #257: Fundamental Frequency
## 🚀 Feature

We want to implement a fundamental frequency feature such as [kaldi's pitch-functions.cc](https://github.com/kaldi-asr/kaldi/blob/master/src/feat/pitch-functions.cc). We could also add a Kaldi compliance interface.

## Additional context

[Paper](https://www.danielpovey.com/files/2014_icassp_pitch.pdf) documenting the implementation in Kaldi

vincentqb:
Since the pitch is important in translating languages such as mandarin, we want a pitch detection algorithm. See also [notebook](https://github.com/vincentqb/audio-tutorial/blob/master/pitch.ipynb).

We implement this using normalized cross-correlation function (NCCF) and median smoothing, mentioned in [RAPT](https://www.ee.columbia.edu/~dpwe/papers/Talkin95-rapt.pdf). Kaldi also uses NCCF, but uses an algorithm based on viterbi instead of median smoothing for smoothing, see [here](https://www.danielpovey.com/files/2014_icassp_pitch.pdf).

- [x] Both pure tones from [here](https://www.mediacollege.com/audio/tone/download/) are detected successfully.
- [x] Move out of example

Closes #257


examples/pitch_dectection/pitch.py
```diff
@@ -0,0 +1,132 @@
...
+
+    best = _combine_max(half, best)
+    indices = best[1]
+
+    # Add back minimal lag
+    indices += lag_min
+    # Add 1 empirical calibration offset
+    indices += 1
+
+    # Median smoothing
```
vincentqb: I'll move the median smoothing as an another function.

cpuhrsch:
Would these transforms already work with @torch.jit.script ? :)

vincentqb:
> Would these transforms already work with @torch.jit.script ? :)

Yup :)

PetrochukM:
What's the difference between this and something like CREPE, pYIN and [SPICE](https://ai.googleblog.com/2019/11/spice-self-supervised-pitch-estimation.html)? Which is best to use for a typical voice-over?

vincentqb:
> What's the difference between this and something like CREPE, pYIN and [SPICE](https://ai.googleblog.com/2019/11/spice-self-supervised-pitch-estimation.html)? Which is best to use for a typical voice-over?

The algorithm included here is a non-neural-network algorithm (based on RAPT, and related to Kaldi's) to identify pitch that can be used as a baseline. The algorithms you mention look promising, but I'd simply experiment and find what's best for a particular case :)

PR #319, opened 2019-10-28
Add functionals gain, dither, scale_to_interval

CamiWilliams:
## Added functionality for [SoX](http://sox.sourceforge.net/sox.html) capabilities.
See [comment](https://github.com/pytorch/audio/issues/260#issuecomment-545116426) in [issue #260](https://github.com/pytorch/audio/issues/260) for details.

**Gain** : Apply amplification or attenuation to the audio signal.
``` 
def gain(waveform, gain_db=1.0):
    # type: (Tensor, float) -> Tensor
```
 
**Scale to Interval** : Scales waveform to an interval.
```
def scale_to_interval(waveform, interval=1.0):
    # type: (Tensor, float) -> Tensor
```

**Dither**: Maximize the dynamic range of audio stored at a particular bit-depth.
```
def dither(waveform, probability_density_function="TPDF", noise_shaping=False, ns_filter=""):
    # type: (Tensor, str, bool, str) -> Tensor
```

## Updated [VCTK](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/vctk.py) dataset code to replace SoX

```
E.append_effect_to_chain("gain", ["-h"])
E.append_effect_to_chain("channels", [1])
E.append_effect_to_chain("rate", [16000])
E.append_effect_to_chain("gain", ["-rh"])
E.append_effect_to_chain("dither", ["-s"])
```
is equivalent to 
```
wf_vctk, sr_vctk = torchaudio.load(test_filepath_vctk)
sample = T.Resample(sr_vctk, 16000, resampling_method='sinc_interpolation')
wf_vctk = sample(wf_vctk)
wf_vctk = F.dither(wf_vctk, noise_shaping=True)
```


vincentqb:
See [comment](https://github.com/pytorch/audio/issues/260#issuecomment-545116426) in #260 for VCTK.

torchaudio/functional.py
```diff
@@ -801,3 +801,21 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
     return torch.nn.functional.conv1d(
         specgram, kernel, groups=specgram.shape[1] // specgram.shape[0]
     ) / denom
+
```
vincentqb: Have you checked for `flake8` on the PR? It's a good way to see if there's lint to correct.

torchaudio/functional.py
```diff
@@ -801,3 +801,74 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    abs_max = torch.max(torch.abs(waveform))
+    ratio = abs_max / interval
+    waveform /= ratio
+
+    assert torch.max(torch.abs(waveform)) == interval
+
+    return waveform
+
+
+def channels(waveform, channel):
```
vincentqb: Thanks for looking at this :) As noted in the body of the [PR](https://github.com/pytorch/audio/issues/260#issue-486504615), `channels` doesn't need its own function.
* If it's used to inform `torchaudio.load` of the number of channels in a raw audio file, this would go in the `load` function instead.
* If it's used to down mix, then the user can simply compute a mean, or other tensor operations directly.
* If it's used to increase the number of channels, then the user can just use tensor operations to repeat (or use their own algorithm directly).

vincentqb: Just to clarify: from what VCTK use, we need gain and dither, since channels doesn't need its own implementation (see list in #260), "rate" already has "resample".

CamiWilliams: AH! Just saw the note about channels. Will remove!

CamiWilliams:
TODO: Add filters to dither
```
Options:            Lipshitz - `L`
                    F-Weighted - `FW`
                    Modified-E-Weighted - `MEW`
                    Improved-E-Weighted - `IEW`
                    Gesemann - `G`
                    Shibata - `S`
                    Low-Shibata - `LS`
                    High-Shibata - `HS`
```

test/test_functional.py
```diff
@@ -247,6 +251,59 @@ def test_create_fb(self):
...
+        sox_dither_waveform, sox_sr = torchaudio.load(test_filepath_sox_dither)
+
+        self.assertTrue(torch.allclose(waveform_dithered, sox_dither_waveform, rtol=1e-03, atol=1e-03))
+
+        test_filepath_sox_dither_ns = os.path.join(self.test_dirpath, "assets", "sinewave_soxdither_noiseshaping.wav")
+        sox_dither_waveform_ns, sox_sr = torchaudio.load(test_filepath_sox_dither_ns)
+
+        self.assertTrue(torch.allclose(waveform_dithered_noiseshaped, sox_dither_waveform_ns, rtol=1e-03, atol=1e-03))
+
+    def test_vctk(self):
```
vincentqb: Let's make the name more precise `test_vctk_transform_pipeline` for instance.

test/test_functional.py
```diff
@@ -247,6 +251,59 @@ def test_create_fb(self):
...
+        # dither
+        wf_vctk = F.dither(wf_vctk, noise_shaping=True)
+
+        test_filepath_vctk_sox = os.path.join(self.test_dirpath, "assets", "p224_002_sox.wav")
+        # E.append_effect_to_chain("gain", ["-h"])
+        # E.append_effect_to_chain("channels", [1])
+        # E.append_effect_to_chain("rate", [16000])
+        # E.append_effect_to_chain("gain", ["-rh"])
+        # E.append_effect_to_chain("dither", ["-s"])
+        waveform_vctk_sox, sr_vctk_sox = torchaudio.load(test_filepath_vctk_sox)
```
vincentqb: Where are the sox filters applied?

torchaudio/datasets/vctk.py
```diff
@@ -21,18 +21,15 @@ def load_vctk_item(
...
-        E.set_input_file(file_audio)
-        E.append_effect_to_chain("gain", ["-h"])
-        E.append_effect_to_chain("channels", [1])
-        E.append_effect_to_chain("rate", [16000])
-        E.append_effect_to_chain("gain", ["-rh"])
-        E.append_effect_to_chain("dither", ["-s"])
-        waveform, sample_rate = E.sox_build_flow_effects()
-    else:
-        waveform, sample_rate = torchaudio.load(file_audio)
+        F = torchaudio.functional
```
vincentqb: The comment `# Legacy` meant the API will change. I suggest we keep a label like `# TODO Remove this parameter after deprecation`

test/test_functional.py
CamiWilliams: I did this in the terminal, so the .wav file is the output

torchaudio/datasets/vctk.py
CamiWilliams: Will do!

test/test_functional.py
cpuhrsch: I'd make this a runtime operation during test time. See the other tests. sox itself might be faulty or change (across platforms) and we want to track that as well.

vincentqb: Cool, looks like that's what @CamiWilliams did :)

torchaudio/functional.py
```diff
@@ -801,3 +802,132 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
     return torch.nn.functional.conv1d(
         specgram, kernel, groups=specgram.shape[1] // specgram.shape[0]
     ) / denom
+
+
```
vincentqb: To compile your function with jit, you can add the decorator `@torch.jit.script` before your function. Since `@torch.jit.script` is applied recursively, only the functions exposed to the users need to be decorated.

When rerunning the tests, there may be new compilation errors that need to be addressed. :)

CamiWilliams: Dine!

torchaudio/functional.py
```diff
@@ -844,6 +844,147 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+       waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+       gain_db (float) Gain adjustment in decibels (dB) (Default: `1.0`).
+
+    Returns:
+       torch.Tensor: waveform amplified by gain_db
+    """
+    if (gain_db == 0):
+        return waveform
+
+    # gain_db = 20log(Tensor_output/tensor_input)
```
cpuhrsch: Should this be removed?

torchaudio/functional.py
```diff
@@ -844,6 +844,147 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+           bound (Default: `1.0`)
+           Example: interval=1.0 -> [-1.0, 1.0]
+    Returns:
+       torch.Tensor: waveform scaled to interval
+    """
+    abs_max = torch.max(torch.abs(waveform))
+    ratio = abs_max / interval
+    waveform /= ratio
+
+    assert torch.max(torch.abs(waveform)) == interval
```
cpuhrsch: This could be expensive and only useful for debugging. @vincentqb - what policy do we have currently in place?

torchaudio/functional.py
```diff
@@ -844,6 +844,147 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    if (probability_density_function == "RPDF"):
+        RPDF_dither = waveform[0][_get_rand_tensor(wave_size)] - 0.5
+
+        signal_scaled_RPDF_dithered = signal_scaled + RPDF_dither
+        quantised_signal_scaled_RPDF_dithered = torch.round(signal_scaled_RPDF_dithered)
+        quantised_signal_RPDF_dithered = quantised_signal_scaled_RPDF_dithered / down_scaling
+
+        dithered = quantised_signal_RPDF_dithered
+    elif (probability_density_function == "GPDF"):
+        gaussian_dither = (waveform[0][_get_rand_tensor(wave_size)]
```
cpuhrsch: You could write this as ```sum(waveform[0][_get_rand_tensor(wave_size)] for 6 * [wave_size])```. The reason that ```sum(6 * [waveform[0][_get_rand_tensor(wave_size)]])``` doesn't work, is that the list expression will just replicate the one value, instead of evaluating the expression ```[waveform[0][_get_rand_tensor(wave_size)]]``` 6 times as desired.

torchaudio/functional.py
```diff
@@ -844,6 +844,147 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    # type: (int) -> int
+    r"""Returns tensor of a random int according to the size.
+    """
+    return torch.randint(size, [1, ]).item()
+
+
+@torch.jit.script
+def dither(waveform, probability_density_function="TPDF", noise_shaping=False, ns_filter=""):
+    # type: (Tensor, str, bool, str) -> Tensor
+    r"""Maximize the dynamic range of audio stored at a particular bit-depth.
```
cpuhrsch: Could you expand a bit on this as to why adding that noise maximizes this?

torchaudio/functional.py
CamiWilliams: If gain_db is 0, shouldn't we return the original waveform? Otherwise the logic is `ratio = 0`, `waveform * 0`

torchaudio/functional.py
```diff
@@ -844,6 +844,147 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+    number_of_bits = 16
+
+    up_scaling = 2 ** (number_of_bits - 1) - 2
+    signal_scaled = waveform * up_scaling
+    down_scaling = 2 ** (number_of_bits - 1)
+
+    dithered = waveform
+
+    if (probability_density_function == "RPDF"):
```
cpuhrsch: It should be worth splitting this out into a helper method instead of ```_get_rand_tensor```. Common pdfs should be useful in general. We already have some here: https://pytorch.org/docs/stable/distributions.html - Are the ones you needed included there?

torchaudio/functional.py
vincentqb: if `gain` is zero, the ratio is 1, and so it's fine to just return the waveform.

@cpuhrsch refers to the comment on line 861, no?

torchaudio/functional.py
vincentqb: We currently use assertions to do some lightweight validation of the input. We do not need to validate the output like this.

torchaudio/functional.py
CamiWilliams: Ah, I got this from the [SoX documentation](http://sox.sourceforge.net/sox.html)

"Dithering is a technique used to maximise the dynamic range of audio stored at a particular bit-depth. Any distortion introduced by quantisation is decorrelated by adding a small amount of white noise to the signal. In most cases, SoX can determine whether the selected processing requires dither and will add it during output formatting if appropriate."

torchaudio/functional.py
CamiWilliams: Quick scan through, I didn't find rectangular, triangular, but there are some options for gaussian ([normal](https://pytorch.org/docs/stable/distributions.html#normal)). I will read a little more in depth and test to make sure the output is the same!

cpuhrsch: Rectangular and triangular might be achieved by using one of the [torch.*_window functions](https://pytorch.org/docs/stable/torch.html#torch.bartlett_window).

torchaudio/functional.py
```diff
@@ -844,6 +844,192 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        err = error[index]
+        error_offset = torch.cat((torch.zeros(1), err))
+        error[index] = error_offset[:waveform.size()[1]]
+
+    noise_shaped = dithered_waveform + error
+    return noise_shaped
+
+
+@torch.jit.script
+def retangular_probability_distribution(waveform):
```
cpuhrsch: I suggest we create a single "probability_distribution" and pass a kwarg to select e.g. rectangular, gaussian, etc.

torchaudio/functional.py
cpuhrsch: It'd be interesting to just add a few lines explaining why this works

torchaudio/functional.py
```diff
@@ -903,7 +903,8 @@ def _add_noise_shaping(dithered_waveform, waveform):
 @torch.jit.script
 def probability_distribution(waveform, density_function="TPDF"):
     # type: (Tensor, str) -> Tensor
-    r"""Perform a probability distribution function on a waveform
+    r"""Perform a probability distribution function on a waveform.
```
cpuhrsch: What does "perform" mean here?

torchaudio/functional.py
```diff
@@ -844,6 +844,163 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        distributed = quantised_signal_gaussian
+    else:
+        TPDF = torch.bartlett_window(wave_size + 1)
+
+        signal_scaled_TPDF = signal_scaled
+        for index in range(signal_scaled_TPDF.size()[0]):
+            signal_scaled_TPDF[index] += TPDF
+
+        quantised_signal_scaled_TPDF = torch.round(signal_scaled_TPDF)
+        quantised_signal_TPDF = quantised_signal_scaled_TPDF / down_scaling
```
vincentqb: I see these lines with `round` and `/ down_scaling` repeated. Would it be cleaner to move them out of the conditional?

torchaudio/functional.py
```diff
@@ -844,6 +844,163 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    number_of_bits = 16
+
+    up_scaling = 2 ** (number_of_bits - 1) - 2
+    signal_scaled = waveform * up_scaling
+    down_scaling = 2 ** (number_of_bits - 1)
+
+    distributed = waveform
+    if (density_function == "RPDF"):
+        random_tensor = int(torch.randint(wave_size, [1, ]).item())
+        RPDF = waveform[0][random_tensor] - 0.5
```
vincentqb: Should this be its own small distribution function to mirror the syntax of the other cases?

torchaudio/functional.py
```diff
@@ -844,6 +844,163 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+        signal_scaled_RPDF = signal_scaled + RPDF
+        quantised_signal_scaled_RPDF = torch.round(signal_scaled_RPDF)
+        quantised_signal_RPDF = quantised_signal_scaled_RPDF / down_scaling
+
+        distributed = quantised_signal_RPDF
+    elif (density_function == "GPDF"):
+        gaussian = torch.distributions.normal.Normal(torch.mean(waveform), 1).sample()
```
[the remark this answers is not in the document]
CamiWilliams: This is the `torch.jit.frontend.NotSupportedError: comprehension ifs not supported yet:` error. Should I wait on that to be fixed, or remove the JIT tags for these functions? (see [issue #29843](https://github.com/pytorch/pytorch/issues/29843))

torchaudio/functional.py
CamiWilliams: Changed to "apply"

torchaudio/functional.py
CamiWilliams: I could do that, should I make a separate PR for it?

torchaudio/functional.py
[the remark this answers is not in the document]
vincentqb: Oh right. While waiting for this to be implemented, we discussed two options, no?

1. Comment out the jit decoration 
```
# TODO Add once https://github.com/pytorch/pytorch/issues/29843 is resolved
# @torch.jit.script
```

2. Use the alternative that didn't rely on `torch.distributions`.
```
# TODO Replace by distribution code once https://github.com/pytorch/pytorch/issues/29843 is resolved
# gaussian = torch.distributions.normal.Normal(torch.mean(waveform), 1).sample()
```

Option 1 is acceptable for this PR since it's a new feature and we can add jit support in stage. Option 2 is preferrable, since it doesn't sacrifice functionality, but takes a little more effort now and in maintenance. Thoughts?

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+        probability_density_function (string): The density function of a
+           continuous random variable (Default: `TPDF`)
+           Options: Triangular Probability Density Function - `TPDF`
+                    Rectangular Probability Density Function - `RPDF`
+                    Gaussian Probability Density Function - `GPDF`
+    Returns:
+        torch.Tensor: waveform dithered with TPDF
+    """
+    wave_size = waveform.size()[1] - 1
```
vincentqb: This assumes a waveform of the form (channel, time), right? It would be more easily batchable to assume (..., time), and put -1 instead of 1 here.

There's a few more of these manipulation that trickle in here. If this is too much effort, we could make it a separate PR to make the transforms batchable.

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    number_of_bits = 16
+
+    up_scaling = 2 ** (number_of_bits - 1) - 2
+    signal_scaled = waveform * up_scaling
+    down_scaling = 2 ** (number_of_bits - 1)
+
+    signal_scaled_dis = waveform
+    if (density_function == "RPDF"):
+        random_tensor = int(torch.randint(wave_size, [1, ]).item())
+        RPDF = waveform[0][random_tensor] - 0.5
```
vincentqb: `waveform[0]` only looks at the first channel, right?

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+        signal_scaled_dis = signal_scaled + RPDF
+    elif (density_function == "GPDF"):
+        gaussian = torch.distributions.normal.Normal(torch.mean(waveform), 1).sample()
```
vincentqb: This takes the mean of the whole waveform across batch/channel/time. We should double check that this would be indeed the desired computation for batching.

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+        signal_scaled_dis = signal_scaled + gaussian
+    else:
+        TPDF = torch.bartlett_window(wave_size + 1)
+
+        signal_scaled_dis = signal_scaled
+        for index in range(signal_scaled_dis.size()[0]):
```
vincentqb: Same `[0]`.

torchaudio/functional.py
vincentqb: Yes, that would be ok in a separate PR.

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    error[n] = dithered[n] - original[n]
+    noise_shaped_waveform[n] = dithered[n] + error[n-1]
+    """
+    error = dithered_waveform - waveform
+
+    # add error[n-1] to dithered_waveform[n], so offset the error by 1 index
+    for index in range(error.size()[0]):
+        err = error[index]
+        error_offset = torch.cat((torch.zeros(1), err))
+        error[index] = error_offset[:waveform.size()[1]]
```
vincentqb: Same for batching

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+def _add_noise_shaping(dithered_waveform, waveform):
+    r"""Noise shaping is calculated by error:
+    error[n] = dithered[n] - original[n]
+    noise_shaped_waveform[n] = dithered[n] + error[n-1]
+    """
+    error = dithered_waveform - waveform
+
+    # add error[n-1] to dithered_waveform[n], so offset the error by 1 index
+    for index in range(error.size()[0]):
```
vincentqb: Same for batching

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    Args:
+       waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+       interval (float): The bounds of the interval, where the float indicates
+           the upper bound and the negative of the float indicates the lower
+           bound (Default: `1.0`)
+           Example: interval=1.0 -> [-1.0, 1.0]
+    Returns:
+       torch.Tensor: waveform scaled to interval
+    """
+    abs_max = torch.max(torch.abs(waveform))
```
vincentqb: Let's add a note in the documentation to say that this normalize the "whole waveform" ? (as opposed to "per channel", or "per batch").

torchaudio/functional.py
```diff
@@ -844,6 +844,152 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+       waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+       gain_db (float) Gain adjustment in decibels (dB) (Default: `1.0`).
+
+    Returns:
+       torch.Tensor: waveform amplified by gain_db
+    """
+    if (gain_db == 0):
+        return waveform
+
+    ratio = 10 ** (gain_db / 20)
```
vincentqb: Same about "whole waveform"?

torchaudio/functional.py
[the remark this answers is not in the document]
CamiWilliams: Implementing option 2!

torchaudio/functional.py
```diff
@@ -844,6 +844,162 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+
+    return waveform * ratio
+
+
+@torch.jit.script
```
vincentqb: By the way, we just changed how we are scripting, see #314. Fortunately, it's just cosmetic, since we've already made your code jitable :)
* We are no longer decorating with `@torch.jit.script`
* We are adding tests instead, see [here](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L511) for instance.

vincentqb: If you remove the `@torch.jit.script` from the PR, please add a corresponding test in `test/test_functional.py`.

torchaudio/functional.py
CamiWilliams: This is resolved with me reshaping the waveform, correct?

vincentqb: `torch.mean(waveform)` takes the mean of all the entries (e.g. also across all batch/channels). I expect you mean `torch.mean(waveform, -1)` to take the mean per batch/channel. Is that the case?

test/test_functional.py
```diff
@@ -29,8 +31,14 @@ class TestFunctional(unittest.TestCase):
     data_sizes = [(2, 20), (3, 15), (4, 10)]
     number_of_trials = 100
     specgram = torch.tensor([1., 2., 3., 4.])
-
+
```
vincentqb: nit: is this an empty line with spaces in it? :)

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    if (density_function == "RPDF"):
+        RPDF = waveform[random_channel][random_time] - 0.5
+
+        signal_scaled_dis = signal_scaled + RPDF
+    elif (density_function == "GPDF"):
+        # TODO Replace by distribution code once 
+        # https://github.com/pytorch/pytorch/issues/29843 is resolved
+        # gaussian = torch.distributions.normal.Normal(torch.mean(waveform), 1).sample()
+        
+        EPOCH = 6
```
vincentqb: "epoch" has quite an established meaning in the context of pytorch. What does it mean here?

torchaudio/functional.py
vincentqb: As you pointed out offline, it would be even better to follow what you do on Line 984 in all three cases.
```
        TPDF = torch.bartlett_window(time_size + 1)
```

torchaudio/functional.py
CamiWilliams: Changed it to be the "number of random variables" in the distribution (see [Normal distribution](https://en.wikipedia.org/wiki/Normal_distribution))

torchaudio/functional.py
CamiWilliams: The window functions we have defined are 
- `torch.bartlett_window`
- `torch.blackman_window`
- `torch.hamming_window`
- `torch.hann_window`

So since we don't have one for Rectangular (Dirichlet window) or Gaussian (Gaussian window), should I write both of those?

It looks like a [Hamming window](https://upload.wikimedia.org/wikipedia/commons/thumb/4/4f/Window_function_and_frequency_response_-_Hamming_%28alpha_%3D_0.53836%2C_n_%3D_0...N%29.svg/960px-Window_function_and_frequency_response_-_Hamming_%28alpha_%3D_0.53836%2C_n_%3D_0...N%29.svg.png) could be similar to a [gaussian window](https://upload.wikimedia.org/wikipedia/commons/thumb/0/05/Window_function_and_frequency_response_-_Gaussian_%28sigma_%3D_0.4%29.svg/960px-Window_function_and_frequency_response_-_Gaussian_%28sigma_%3D_0.4%29.svg.png), but they're definitely [defined as different window functions everywhere](https://www.dsprelated.com/freebooks/sasp/Spectrum_Analysis_Windows.html). It DOES however, say that a hamming window is a "Rectangular + scaled-cosine window", so in theory I could device by one period of a cosine.

CamiWilliams: If I write these, should I include them in `torch` or just keep them in `functions.py`?

vincentqb: Let's start i in `functional.py` for now. We can migrate them later.

torchaudio/functional.py
```diff
@@ -844,6 +844,175 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        for ws in EPOCH * [time_size]:
+            rand_chan = int(torch.randint(channel_size, [1, ]).item())
+            gaussian += waveform[rand_chan][int(torch.randint(ws, [1, ]).item())]
+
+        signal_scaled_dis = signal_scaled + gaussian
+    else:
+        TPDF = torch.bartlett_window(time_size + 1)
+
+        signal_scaled_dis = signal_scaled
+        for index in range(channel_size + 1):
```
vincentqb: would `TPDF.repeat(...)` avoid a for-loop here?

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        return waveform
+
+    ratio = 10 ** (gain_db / 20)
+
+    return waveform * ratio
+
+
+def scale_to_interval(waveform, interval=1.0):
+    # type: (Tensor, float) -> Tensor
+    r"""Scales the whole waveform to an interval.
```
vincentqb: nit: "Scale the waveform to the interval [-interval_max, interval_max] across all dimensions."

nit: How about `interval_max` instead of `interval`?

torchaudio/functional.py
vincentqb: As commented [here](https://github.com/pytorch/audio/pull/319#discussion_r346965072), let's punt this for a subsequent PR.

test/test_functional.py
```diff
@@ -31,6 +32,12 @@ class TestFunctional(unittest.TestCase):
     specgram = torch.tensor([1., 2., 3., 4.])
 
     test_dirpath, test_dir = common_utils.create_temp_assets_dir()
+
+    test_filepath_sinewave = os.path.join(test_dirpath, "assets", "sinewave.wav")
+    waveform_sinewave, sr_sinewave = torchaudio.load(test_filepath_sinewave)
+    E = torchaudio.sox_effects.SoxEffectsChain()
```
vincentqb: Doing this means we should definitely clear the chain at the beginning of each test so we don't save much code. Let's create the chain inside in each test, so we are sure they do not interact.

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    Args:
+       waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+       probability_density_function (string): The density function of a
+           continuous random variable (Default: `TPDF`)
+           Options: Triangular Probability Density Function - `TPDF`
+                    Rectangular Probability Density Function - `RPDF`
+                    Gaussian Probability Density Function - `GPDF`
+       noise_shaping (boolean): a filtering process that shapes the spectral
+           energy of quantisation error (Default: `False`)
+       ns_filter (string): TODO The noise shaping filter (Default: `""`)
```
vincentqb: Is this something you mean to do as part of this PR? If not, let's instead create an issue in torchaudio to add this functionality in dither on top of what we have here.

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    """
+    if (gain_db == 0):
+        return waveform
+
+    ratio = 10 ** (gain_db / 20)
+
+    return waveform * ratio
+
+
+def scale_to_interval(waveform, interval=1.0):
```
vincentqb: Can you provide the link to where we discuss adding this? I can't find it :)

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        signal_scaled_dis = signal_scaled
+        for index in range(channel_size + 1):
+            signal_scaled_dis[index] += TPDF
+
+    quantised_signal_scaled = torch.round(signal_scaled_dis)
+    quantised_signal = quantised_signal_scaled / down_scaling
+    return quantised_signal.reshape(shape[:-1] + quantised_signal.shape[-1:])
+
+
+def dither(waveform, probability_density_function="TPDF", noise_shaping=False, ns_filter=""):
```
vincentqb: nit: let's just use `density_function` in the parameters.

torchaudio/functional.py
```diff
@@ -860,6 +860,172 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+        error_offset = torch.cat((torch.zeros(1), err))
+        error[index] = error_offset[:waveform.size()[1]]
+
+    noise_shaped = dithered_waveform + error
+    return noise_shaped.reshape(dithered_shape[:-1] + noise_shaped.shape[-1:])
+
+
+def probability_distribution(waveform, density_function="TPDF"):
+    # type: (Tensor, str) -> Tensor
+    r"""Apply a probability distribution function on a waveform.
```
vincentqb: This should just be an internal function, and it in fact does the core of applying dither. Let's rename this to `_apply_dither`.

nit: "Apply dither to the waveform using the chosen density function."

test/test_functional.py
```diff
@@ -31,6 +32,10 @@ class TestFunctional(unittest.TestCase):
     specgram = torch.tensor([1., 2., 3., 4.])
 
     test_dirpath, test_dir = common_utils.create_temp_assets_dir()
+
+    test_filepath_sinewave = os.path.join(test_dirpath, "assets", "sinewave.wav")
```
cpuhrsch: I don't think a sinewave is enough as a test example. There should be a few others that more generic. Sinewaves don't cover a broad, mixes range of frequencies.

torchaudio/functional.py
```diff
@@ -858,6 +858,162 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    """
+    if (gain_db == 0):
+        return waveform
+
+    ratio = 10 ** (gain_db / 20)
+
+    return waveform * ratio
+
+
+def scale_to_interval(waveform, interval_max=1.0):
```
cpuhrsch: Do we want to expose and maintain this as a separate function? I worry that it's too simple to warrant an explicit abstraction. We could preprend an underscore for now to signal that it is private. I also don't see it used anywhere outside of the tests.

torchaudio/functional.py
```diff
@@ -858,6 +858,162 @@ def compute_deltas(specgram, win_length=5, mode="replicate"):
...
+    for index in range(error.size()[0]):
+        err = error[index]
+        error_offset = torch.cat((torch.zeros(1), err))
+        error[index] = error_offset[:waveform.size()[1]]
+
+    noise_shaped = dithered_waveform + error
+    return noise_shaped.reshape(dithered_shape[:-1] + noise_shaped.shape[-1:])
+
+
+def probability_distribution(waveform, density_function="TPDF"):
```
cpuhrsch: I'd expect this to return a probability distribution and not to apply it to a waveform. Why doesn't applying require more than simple addition? And if it does, I think it's worth creating a separate function "apply noise" or such that explains why it's necessary.

cpuhrsch: Since this is only used internally for now, maybe we can prepend it with an underscore and eventually make it public.

torchaudio/functional.py
vincentqb: Let's remove this function from this PR. We can have a separate PR for it if we want to keep the history.

test/test_functional.py
vincentqb: We have other asset files we can use instead which I'd say would be good enough.

torchaudio/functional.py
vincentqb: Let's keep this function internal for now.

torchaudio/functional.py
vincentqb: I see this comment marked as resolved. What was the resolution of this?

CamiWilliams: Just pushed the change, I renamed it to _apply_probability_distribution. How does that sound?

CamiWilliams: per [this comment](https://github.com/pytorch/audio/pull/319/files/2d601c2ffd2ed7e2cf1a4a808e53dba9537ce114#r352820190)

CamiWilliams: And [this comment](https://github.com/pytorch/audio/pull/319#discussion_r346965072)... should I still rename to "apply dither"?

PR #320, opened 2019-10-29
Resume download

vincentqb:
This adds resume to download function, and a validate function with md5 or sha256, from pytorch/pytorch#24915.

Following [this](https://tobiasraabe.github.io/blog/how-to-download-files-with-python.html), it was tested using:
```python
    def test_download_url(self):

        url = "http://www.patentsview.org/data/20171226/botanic.tsv.zip"
        hash_url = "94c642405619b20ecaf657b30e84bab787320649e751ed6ac629c0be613ded44"

        download_folder = "."
        download_url(url, download_folder)
        validate_download_url(url, download_folder, hash_url)
```
Do we have a url we can use to test? I'm leaning toward not having a test like this that would ping a url with every batch of test.

See 
* [another example](https://gist.github.com/mjohnsullivan/9322154#file-download-py-L32).
* [using keep alive](https://stackoverflow.com/questions/25239650/python-requests-speed-up-using-keep-alive)

### DIFF
--- a/torchaudio/datasets/utils.py
+++ b/torchaudio/datasets/utils.py
@@ -126,13 +126,11 @@ def download_url(
         resume (bool): Enable resuming download.
     """
 
-    # Detect filename
-    if filename is None:
-        req = urllib.request.Request(url, method="HEAD")
-        filename = urllib.request.urlopen(
-            req
-        ).info().get_filename() or os.path.basename(url)
+    req = urllib.request.Request(url, method="HEAD")
+    req_info = urllib.request.urlopen(req).info()
 
+    # Detect filename
+    filename = filename or req_info.get_filename() or os.path.basename(url)
     filepath = os.path.join(download_folder, filename)
 
     if resume and os.path.exists(filepath):
@@ -146,17 +144,26 @@ def download_url(
         mode = "wb"
         local_size = None
 
-    # If we already have the whole file, there is no need to download it again
+    if hash_value and local_size == int(req_info.get("Content-Length", -1)):
+        if validate_file(filepath, hash_value, hash_type):
+            return
+        else:
+            raise RuntimeError(
+                "The hash of {} does not match. Delete the file manually and retry.".format(
+                    filepath
+                )
+            )
+
+    with open(filepath, mode) as fpointer:
+        for chunk in stream_url(url, start_byte=local_size, progress_bar=progress_bar):
+            fpointer.write(chunk)
+
     if hash_value and not validate_file(filepath, hash_value, hash_type):
         raise RuntimeError(
             "The hash of {} does not match. Delete the file manually and retry.".format(
                 filepath
             )
         )
-
-    with open(filepath, mode) as fpointer:
-        for chunk in stream_url(url, start_byte=local_size, progress_bar=progress_bar):
-            fpointer.write(chunk)
 
 
 def validate_file(filepath, hash_value, hash_type="sha256"):
--- a/torchaudio/datasets/utils.py
+++ b/torchaudio/datasets/utils.py
@@ -10,7 +10,6 @@ import threading
 import zipfile
 from queue import Queue
 
-import requests
 import six
 import torch
 import torchaudio
@@ -68,44 +67,6 @@ def makedir_exist_ok(dirpath):
             raise
 
 
-def download_url_resume(url, download_folder, resume_byte_pos=None):
-    """Download url to disk with possible resumption.
-
-    Args:
-        url (str): Url.
-        download_folder (str): Folder to download file.
-        resume_byte_pos (int): Position of byte from where to resume the download.
-    """
-    # Get size of file
-    r = requests.head(url)
-    file_size = int(r.headers.get("content-length", 0))
-
-    # Append information to resume download at specific byte position to header
-    resume_header = (
-        {"Range": "bytes={}-".format(resume_byte_pos)} if resume_byte_pos else None
-    )
-
-    # Establish connection
-    r = requests.get(url, stream=True, headers=resume_header)
-
-    # Set configuration
-    n_block = 32
-    block_size = 1024
-    initial_pos = resume_byte_pos if resume_byte_pos else 0
-    mode = "ab" if resume_byte_pos else "wb"
-
-    filename = os.path.basename(url)
-    filepath = os.path.join(download_folder, os.path.basename(url))
-
-    with open(filepath, mode) as f:
-        with tqdm(
-            unit="B", unit_scale=True, unit_divisor=1024, total=file_size
-        ) as pbar:
-            for chunk in r.iter_content(n_block * block_size):
-                f.write(chunk)
-                pbar.update(len(chunk))
-
-
 def download_url(url, download_folder, hash_value=None, hash_type="sha256"):
     """Execute the correct download operation.
     Depending on the size of the file online and offline, resume the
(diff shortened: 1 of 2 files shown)


vincentqb:
For reference: [sample code](https://stackoverflow.com/questions/7243750/download-file-from-web-in-python-3) that opens a gzip url and decode on-the-fly.

cpuhrsch:
Yes, in an ideal world our tests do not depend on the internet. But it's hard to test a function that operates on URLs without URLs. I guess we could use a ```file://``` type URLs? This might be a bit harder to test rigorously within the CI.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,85 @@ def makedir_exist_ok(dirpath):
             raise
 
 
-def download_url(url, root, filename=None, md5=None):
-    """Download a file from a url and place it in root.
+def download_url(url, download_folder, hash_value=None, hash_type="sha256"):
+    """Execute the correct download operation.
+    Depending on the size of the file online and offline, resume the
```
cpuhrsch: I'd make resuming an option, similar to how something like rsync allows you to decide whether you want to just simply overwrite or attempt to detect differences.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,85 @@ def makedir_exist_ok(dirpath):
             raise
 
 
-def download_url(url, root, filename=None, md5=None):
-    """Download a file from a url and place it in root.
+def download_url(url, download_folder, hash_value=None, hash_type="sha256"):
```
cpuhrsch: This makes the assumption that we want to download to a folder. Is it possible to split this out?

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,85 @@ def makedir_exist_ok(dirpath):
...
+        while True:
+            chunk = upointer.read(block_size)
+            if not chunk:
+                break
+            fpointer.write(chunk)
+            num_bytes += len(chunk)
+            pbar.update(len(chunk))
+
+
+def validate_download_url(filepath, hash_value, hash_type="sha256"):
```
cpuhrsch: This works with any file right? So it could just be called "validate" and take a path. Even better, take a buffer.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,85 @@ def makedir_exist_ok(dirpath):
...
-                urllib.request.urlretrieve(url, fpath, reporthook=gen_bar_updater())
-            else:
-                raise e
+            )
+
+        return
+
+    with open(filepath, mode) as fpointer, urllib.request.urlopen(
+        req
+    ) as upointer, tqdm(
```
cpuhrsch: I know there are tqdm options that prevent things from being printed, but I think this should also warrant a "quiet" or "verbose" flag. Something like this can fill up a log with a lot of lines of unimportant information. This kind of plays into a global logging level flag.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,85 @@ def makedir_exist_ok(dirpath):
...
+        hash_value (str): Hash for url.
+        hash_type (str): Hash type.
     """
-    from six.moves import urllib
 
-    root = os.path.expanduser(root)
-    if not filename:
-        filename = os.path.basename(url)
-    fpath = os.path.join(root, filename)
+    filepath = os.path.join(download_folder, os.path.basename(url))
```
cpuhrsch: Some URLs might no yield a sane filename, but people will still want to download from this. There's an option to attempt to detect the filename that the underlying file resolves to in https://github.com/pytorch/text/blob/master/torchtext/utils.py#L53

cpuhrsch:
Until we define torchdata we'll likely copy-paste functions like this. Although download has the potential of living within core pytorch since it's also required by torchhub.

torchaudio/datasets/utils.py
vincentqb: Yes, it can be done.

torchaudio/datasets/utils.py
vincentqb: Sure

torchaudio/datasets/utils.py
vincentqb: Indeed.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,130 @@ def makedir_exist_ok(dirpath):
             raise
 
 
-def download_url(url, root, filename=None, md5=None):
-    """Download a file from a url and place it in root.
+def stream_url(url, start_byte=None, block_size=32 * 1024, progress_bar=True):
```
cpuhrsch: This could be an excellent target for a multithreaded buffer function :D

vincentqb:
> Bonus points for a multithreaded stream_urls function that accepts a list of urls and yields the results URL by URL (or alternatively as_completed). For this you can create a buffer that you fill up with chunks concurrently and can whose size be specified by the user. My main worry here would be a timeout, since the URL chunks will be streamed sequentially. Presumably you can simply read ahead. Maybe this can be achieved by a generic Buffer function :)
> 
> You can use concurrent [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor) for that. Based on https://python3statement.org/ PyTorch will drop Python2.7 support in 2020. So it's ok to have features that will not work with Python2.7 at this point, as long as we still guard against them not being available.

If the goal is to get an iterator over multiple files, then we can wrap `download_url` by simply applying it to a list of urls.

If instead we want to add parallel at the at the stream_url level, then we'd need a buffer also to make the download parallel. This is more specialized in its application though. Thoughts?

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,161 @@ def makedir_exist_ok(dirpath):
...
+
+    if hash_value and not validate_file(filepath, hash_value, hash_type):
+        raise RuntimeError(
+            "The hash of {} does not match. Delete the file manually and retry.".format(
+                filepath
+            )
+        )
+
+
+def download_url(urls, *args, max_workers=5, **kwargs):
```
vincentqb: This notation and concurrent futures are not supported in python 2. We could leave the multithreaded download for the first release with python 3. Otherwise, I could add some mechanism to offer this functionality only when python 3 is available and default to single thread otherwise.

torchaudio/datasets/utils.py
```diff
@@ -76,41 +65,161 @@ def makedir_exist_ok(dirpath):
...
+        if not isinstance(value, list):
+            kwargs[key] = [value] * len(urls)
+    kwargs = [dict(zip(kwargs.keys(), values)) for values in zip(*(kwargs.values()))]
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=max_workers) as executor:
+        futures = [
+            executor.submit(download_single_url, url, *arg, **kwarg)
+            for url, arg, kwarg in zip(urls, args, kwargs)
+        ]
+        return concurrent.futures.as_completed(futures)
```
vincentqb: This notation and concurrent futures are not supported in python 2. We could leave the multithreaded download for the first release with python 3. Otherwise, I could add some mechanism to offer this functionality only when python 3 is available and default to single thread otherwise.

torchaudio/datasets/utils.py
vincentqb: [Indeed](https://github.com/pytorch/audio/pull/323) :D

torchaudio/datasets/utils.py
```diff
@@ -65,85 +65,123 @@ def makedir_exist_ok(dirpath):
...
-def validate_download_url(filepath, hash_value, hash_type="sha256"):
+def validate_file(filepath, hash_value, hash_type="sha256"):
     """Validate a given file with its hash.
 
     Args:
-        url (str): Url.
-        download_folder (str): Folder to download file.
+        filepath (str): File to read.
         hash_value (str): Hash for url.
         hash_type (str): Hash type.
```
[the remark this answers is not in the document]
vincentqb: Agree.

torchaudio/datasets/utils.py
cpuhrsch: Sounds good

vincentqb: For reference: [this](https://github.com/pytorch/audio/pull/320/commits/33156c4ba761342beee77d1448ffe9cf8c41bb2a) is the commit that adds multidownload.

cpuhrsch:
Should we merge this?

torchaudio/datasets/utils.py
vincentqb: So you want to suggest removing the `with open(filepath, "rb")` statement?

vincentqb:
> Bonus points for a multithreaded stream_urls function that accepts a list of urls and yields the results URL by URL (or alternatively as_completed). For this you can create a buffer that you fill up with chunks concurrently and can whose size be specified by the user. My main worry here would be a timeout, since the URL chunks will be streamed sequentially. Presumably you can simply read ahead. Maybe this can be achieved by a generic Buffer function :)
> 
> You can use concurrent [ThreadPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#threadpoolexecutor) for that. Based on https://python3statement.org/ PyTorch will drop Python2.7 support in 2020. So it's ok to have features that will not work with Python2.7 at this point, as long as we still guard against them not being available.

Indeed, see [here](https://github.com/pytorch/audio/pull/323) :D

vincentqb:
> Should we merge this?

Rebased. Ready when you are :)

torchaudio/datasets/utils.py
cpuhrsch: Yes and instead accept a file_like object.

torchaudio/datasets/utils.py
```diff
@@ -78,41 +65,129 @@ def makedir_exist_ok(dirpath):
...
+    """
+
+    if hash_type == "sha256":
+        hash_func = hashlib.sha256()
+    elif hash_type == "md5":
+        hash_func = hashlib.md5()
+    else:
+        raise ValueError
+
+    with open(filepath, "rb") as f:
```
cpuhrsch: We're assuming a file_path as an input here for validation. I think this would be useful broadly for file-like objects. If you want to make this specific I'd suggest something like ```validate_filepath(filepath): validate_file(open(filepath))```.

cpuhrsch:
I think it's worthwhile standardizing on file-like objects and create convenience wrapper functions for objects that can be referenced to via a file path.

torchaudio/datasets/utils.py
vincentqb: Sure, I'm ok with that.

torchaudio/datasets/utils.py
vincentqb: I made `validate_file` take a file_obj and iterate through it. `download_url` calls it with `with open(...)`.

torchaudio/datasets/utils.py
```diff
@@ -78,41 +65,130 @@ def makedir_exist_ok(dirpath):
...
+    if hash_type == "sha256":
+        hash_func = hashlib.sha256()
+    elif hash_type == "md5":
+        hash_func = hashlib.md5()
+    else:
+        raise ValueError
+
+    while True:
+        # Read by chunk to avoid filling memory
+        chunk = f.read(1024 ** 2)
```
zhangguanheng66: Does this function actually consume `file_obj`?

vincentqb: This consumes an object with `read(chunk_size)` signature, see line 147. Is that what you meant?

zhangguanheng66: Yes, but should it be `file_obj.read` rather than `f.read`?

vincentqb: Oh! Thanks for pointing out. Opened #352.

PR #326, opened 2019-11-01
make all functional jit script

vincentqb:
Update functional to be jit script, as discussed [here](https://github.com/pytorch/audio/pull/315#issuecomment-548604295).

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -442,7 +442,9 @@ def complex_norm(complex_tensor, power=1.0):
     return torch.norm(complex_tensor, 2, -1).pow(power)
 
 
+@torch.jit.script
 def angle(complex_tensor):
+    # type: (Tensor) -> Tensor
     r"""Compute the angle of complex tensor input.
 
     Args:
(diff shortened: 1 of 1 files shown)


vincentqb:
> I would suggest to test those functions under the jit mode in case that some ops are not supported by the jit team.

The current convention is to compile with jit by default by adding the `@torch.jit.script` decorator, and so this PR is extending this convention to functions that were not covered in their respective PRs. It may be a good idea to not compile them by default, and let the user it themselves. This is what is discussed in #314. Thoughts?

All the functions in this PR are being run in jit mode with `@torch.jit.script` and so are compiled successfully during import.

zhangguanheng66:
> > I would suggest to test those functions under the jit mode in case that some ops are not supported by the jit team.
> 
> The current convention is to compile with jit by default by adding the `@torch.jit.script` decorator, and so this PR is extending this convention to functions that were not covered in their respective PRs. It may be a good idea to not compile them by default, and let the user it themselves. This is what is discussed in #314. Thoughts?
> 
> All the functions in this PR are being run in jit mode with `@torch.jit.script` and so are compiled successfully during import.

I'm not suggested to compile them by default. Instead, it would be great if we could test the func under the JIT mode. Based on my experience, some of the ops (`hasattr`, `for` loop with index, etc) are not supported by JIT).

cpuhrsch:
We do have JIT tests https://github.com/pytorch/audio/blob/master/test/test_jit.py - @vincentqb we should evaluate whether these are still necessary.

vincentqb:
> We do have JIT tests https://github.com/pytorch/audio/blob/master/test/test_jit.py - @vincentqb we should evaluate whether these are still necessary.

Something like this will be needed for #314. Since there is no harm in keeping in in this PR, I will leave them here for now.

PR #327, opened 2019-11-01
Complex STFT transform from spectrogram

vincentqb:
We offer a layer that applies complex STFT without the power/normalization done in spectrogram. However, we do not want to overload the name "STFT" existing in pytorch, as done in the original proposal in [the proposal](https://github.com/pytorch/audio/pull/327/commits/008791c06a7ea187056f1ee48072ec37705c9148) from #285.

@ksanjeevan -- thoughts?

CC @ksanjeevan @keunwoochoi @cpuhrsch

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -313,6 +313,20 @@ class Tester(unittest.TestCase):
         computed = transform(specgram)
         self.assertTrue(computed.shape == specgram.shape, (computed.shape, specgram.shape))
 
+    def test_batch_spectrogram(self):
+        waveform, sample_rate = torchaudio.load(self.test_filepath)  # (2, 278756), 44100
+
+        # Single then transform then batch
+        expected = transforms.Spectrogram()(waveform).unsqueeze(0).repeat(3,1,1,1)
+
+        # Batch then transform
+        waveform = waveform.unsqueeze(0).repeat(3,1,1)
+        computed = transforms.Spectrogram()(waveform)
+
+        # shape = (3, 2, 201, 1394)
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+        self.assertTrue(torch.allclose(computed, expected))
+
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -314,16 +314,14 @@ class Tester(unittest.TestCase):
         self.assertTrue(computed.shape == specgram.shape, (computed.shape, specgram.shape))
 
     def test_batch_spectrogram(self):
-        waveform, sample_rate = torchaudio.load(self.test_filepath)  # (2, 278756), 44100
+        waveform, sample_rate = torchaudio.load(self.test_filepath)
 
         # Single then transform then batch
-        expected = Transform()(waveform).unsqueeze(0).repeat(3, 1, 1, 1)
+        expected = transforms.Spectrogram()(waveform).repeat(3, 1, 1, 1)
 
         # Batch then transform
-        waveform = waveform.unsqueeze(0).repeat(3, 1, 1)
-        computed = transforms.Spectrogram()(waveform)
+        computed = transforms.Spectrogram()(waveform.repeat(3, 1, 1))
 
-        # shape = (3, 2, 201, 1394)
         self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
         self.assertTrue(torch.allclose(computed, expected))
 
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -317,10 +317,10 @@ class Tester(unittest.TestCase):
         waveform, sample_rate = torchaudio.load(self.test_filepath)  # (2, 278756), 44100
 
         # Single then transform then batch
-        expected = transforms.Spectrogram()(waveform).unsqueeze(0).repeat(3,1,1,1)
+        expected = Transform()(waveform).unsqueeze(0).repeat(3, 1, 1, 1)
 
         # Batch then transform
-        waveform = waveform.unsqueeze(0).repeat(3,1,1)
+        waveform = waveform.unsqueeze(0).repeat(3, 1, 1)
         computed = transforms.Spectrogram()(waveform)
 
         # shape = (3, 2, 201, 1394)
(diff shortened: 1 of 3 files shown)


ksanjeevan:
@vincentqb Yeah that would work! So I'm not seeing anything on `Spectrogram` that checks that `power` is `int` or anything like that so just with this change of the functional we could have a complex stft transform correct?

torchaudio/functional.py
```diff
@@ -246,14 +247,22 @@ def spectrogram(
...
+    shape = waveform.size()
+    waveform = waveform.reshape(-1, shape[-1])
+
     # default values are consistent with librosa.core.spectrum._spectrogram
     spec_f = _stft(
         waveform, n_fft, hop_length, win_length, window, True, "reflect", False, True
     )
 
+    # unpack batch
+    spec_f = spec_f.reshape(shape[:-1] + spec_f.shape[-3:])
```
ksanjeevan: @vincentqb is support for batching going to be done optionally then? :grin: I can add it to the augmentation transforms if needed.

vincentqb: Yeah, we decided to go ahead and add batching using reshape now :) we'll then update these codes when nested tensor comes out

vincentqb:
> @vincentqb Yeah that would work! So I'm not seeing anything on `Spectrogram` that checks that `power` is `int` or anything like that so just with this change of the functional we could have a complex stft transform correct?

Right, there should only be a check for none to disable the power computation

torchaudio/functional.py
```diff
@@ -218,42 +218,51 @@ def istft(
...
-    # type: (Tensor, int, Tensor, int, int, int, int, bool) -> Tensor
+    # type: (Tensor, int, Tensor, int, int, int, Optional[int], bool) -> Tensor
     r"""
     spectrogram(waveform, pad, window, n_fft, hop_length, win_length, power, normalized)
 
     Create a spectrogram from a raw audio signal.
 
     Args:
-        waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+        waveform (torch.Tensor): Tensor of audio of dimension ([batch,] channel, time)
```
keunwoochoi: It's not directly about this PR, but probably the optional last dimension for complex stft should be mentioned as well.

torchaudio/functional.py
```diff
@@ -218,42 +218,51 @@ def istft(
 def spectrogram(
     waveform, pad, window, n_fft, hop_length, win_length, power, normalized
 ):
-    # type: (Tensor, int, Tensor, int, int, int, int, bool) -> Tensor
+    # type: (Tensor, int, Tensor, int, int, int, Optional[int], bool) -> Tensor
     r"""
     spectrogram(waveform, pad, window, n_fft, hop_length, win_length, power, normalized)
 
     Create a spectrogram from a raw audio signal.
```
keunwoochoi: I think the {optional / magnitude} should be also mentioned here, too.

test/test_transforms.py
```diff
@@ -313,6 +313,20 @@ def test_compute_deltas_twochannel(self):
         computed = transform(specgram)
         self.assertTrue(computed.shape == specgram.shape, (computed.shape, specgram.shape))
 
+    def test_batch_spectrogram(self):
+        waveform, sample_rate = torchaudio.load(self.test_filepath)  # (2, 278756), 44100
+
+        # Single then transform then batch
+        expected = transforms.Spectrogram()(waveform).unsqueeze(0).repeat(3,1,1,1)
```
keunwoochoi: Linter will fix this anyway, but `(3,1,1,1)` --> `(3, 1, 1, 1)`

test/test_transforms.py
```diff
@@ -313,6 +313,20 @@ def test_compute_deltas_twochannel(self):
...
+
+        # Batch then transform
+        waveform = waveform.unsqueeze(0).repeat(3,1,1)
```
keunwoochoi: also `(3, 1, 1)`

torchaudio/functional.py
keunwoochoi: The convention was specifying the optional batch dimension as `(*, channel, time)`. Maybe it's nicer to follow?

torchaudio/functional.py
keunwoochoi: So is the optional batch dimension too. So.. something like
```
Create a spectrogram or a batch of spectrograms from a raw audio signal. The spectrogram can be
either magnitude-only or complex. 
```

torchaudio/functional.py
vincentqb: Yes, I agree, thanks for pointing this out!

torchaudio/functional.py
```diff
@@ -218,42 +217,52 @@ def istft(
...
     r"""
     spectrogram(waveform, pad, window, n_fft, hop_length, win_length, power, normalized)
 
-    Create a spectrogram from a raw audio signal.
+    Create a spectrogram or a batch of spectrograms from a raw audio signal.
+    The spectrogram can be either magnitude-only or complex.
 
     Args:
-        waveform (torch.Tensor): Tensor of audio of dimension (channel, time)
+        waveform (torch.Tensor): Tensor of audio of dimension (*, channel, time)
```
cpuhrsch: I'd explicitly call this 'batch', i.e. ([batch], channel, time) or such.

vincentqb: This is more general than batch. It works with a tensor of any dimension ending in (channel, time). In fact, this could be updated to (*, time) by making channel optional. Saying "(batch, channel, time) or (channel, time) or (time)" is a little heavy. And saying "([batch], [channel], time)" hides the generality. Thoughts?

keunwoochoi: I can't think of anything more compact. Just by following what it is something like `..(*, time), e.g., (time), (channel, time), (batch, channel, time)`? It's kinda verbose but if we can reduce its ambiguity :)

vincentqb: I agree with reducing ambiguity :)

Is `(..., time)` clearer than `(*, time)`?

keunwoochoi: Hmm............ maybe? (But if I might have confused you, I meant `..audio of dimensions (*, time), e.g., (time), blah -`.) The current `torch.stft` doc uses `(*, time, ..)` to indicate an optional batch dimension which is more limited than we want here. Probably `(..., time)` makes more sense then.

vincentqb: Good point, the better globbing to use would be `(**, time)` not `(*, time)` for `(..., time)` :)

I do not recommend using that though :)

PR #330, opened 2019-11-05
Data points remain tuples

vincentqb:
Since this is what has been most used in domains, we keep tuples as the item to return for now.

for CommonVoice:
```
    # dictionary with keys:
    # client_id, path, sentence, up_votes, down_votes, age, gender, accent
    return waveform, sample_rate, dictionary
```
for LibriSpeech:
```
    return (
        waveform,
        sample_rate,
        utterance,
        int(speaker_id),
        int(chapter_id),
        int(utterance_id),
    )
```
for VCTK:
```
    return waveform, sample_rate, utterance, speaker_id, utterance_id
```
for YesNo:
```
    # labels = [int(c), ...]
    return waveform, sample_rate, labels
```

CC #303

### DIFF
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -29,13 +29,14 @@ def load_librispeech_item(fileid, path, ext_audio, ext_txt):
     waveform, sample_rate = torchaudio.load(file_audio)
 
     # Load text
-    for line in open(file_text):
-        fileid_text, content = line.strip().split(" ", 1)
-        if fileid_audio == fileid_text:
-            break
-    else:
-        # Translation not found
-        raise FileNotFoundError("Translation not found for " + fileid_audio)
+    with open(file_text) as ft:
+        for line in ft:
+            fileid_text, content = line.strip().split(" ", 1)
+            if fileid_audio == fileid_text:
+                break
+        else:
+            # Translation not found
+            raise FileNotFoundError("Translation not found for " + fileid_audio)
 
     return {
         "speaker_id": speaker,
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -1,8 +1,7 @@
 import os
 
-from torch.utils.data import Dataset
-
 import torchaudio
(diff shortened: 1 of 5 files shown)


torchaudio/datasets/librispeech.py
```diff
@@ -16,38 +15,43 @@
...
-
-    return {
-        "speaker_id": speaker,
-        "chapter_id": chapter,
-        "utterance_id": utterance,
-        "utterance": content,
-        "waveform": waveform,
-        "sample_rate": sample_rate,
-    }
+    with open(file_text) as ft:
```
vincentqb: Adding this as part of this PR to make sure that the file is closed.

torchaudio/datasets/librispeech.py
```diff
@@ -16,38 +15,43 @@
...
+            # Translation not found
+            raise FileNotFoundError("Translation not found for " + fileid_audio)
+
+    return (
+        waveform,
+        sample_rate,
+        utterance,
+        int(speaker_id),
+        int(chapter_id),
+        int(utterance_id),
```
vincentqb: Also converting to int though they come from file names.

torchaudio/datasets/yesno.py
```diff
@@ -11,16 +11,20 @@
 
 def load_yesno_item(fileid, path, ext_audio):
     # Read label
-    label = fileid.split("_")
+    labels = [int(c) for c in fileid.split("_")]
```
vincentqb: Also converting to int though they come from file names.

vincentqb:
Just modified CommonVoice to return a tuple (waveform, sample_rate, dictionary), where the dictionary contains the data and header from the tsv. This is closer to how the data is provided by CommonVoice. The only assumption on the tsv is that the second column is the name of the audio file to retrieve.

The previous iteration was:
```
    # waveform, sample_rate, client_id, path, sentence, up_votes, down_votes, age, gender, accent
    return(waveform, sample_rate, *line)
```
P.S. The notation with *line is not compatible with python 2.

PR #331, opened 2019-11-05
Multi download

vincentqb:
Now that  python 2 has been deprecated, we want to add parallel download, see [comment](https://github.com/pytorch/audio/pull/320#discussion_r340872063).

We are using concurrent futures. An alternative would be if this could be fed into `DataLoader` which would then take care of the multiprocessing and download.


vincentqb:
To tests:

```
    def test_download_multiple_url(self):

        url = "http://www.patentsview.org/data/20171226/botanic.tsv.zip"
        url = [url] * 6

        download_folder = "."
        fs = download_url(url, download_folder, hash_value=None)
```

torchaudio/datasets/utils.py
```diff
@@ -171,6 +172,38 @@ def download_url(
...
+    if isinstance(urls, str):
+        return download_single_url(urls, *args, **kwargs)
+
+    # Turn arguments into lists
+    args = list(args)
+    for i, item in enumerate(args):
+        if not isinstance(item, list):
+            args[i] = [item] * len(urls)
+    args = list(zip(*args))
+
```
tomassosorio: If its passed 10 urls and a list with 6 lists this will explode, maybe should be added an 
`assert len(args) == len(urls)`

tomassosorio:
```suggestion
assert len(args) == len(urls)
```

If its passed something like:
`download_url(["www.google.com", "www.yahoo.com"], [1, ])`

It will download only 1 file and not both files, since when you zip len of args will be lower then urls

PR #337, opened 2019-11-11
Batching for transforms

### LINKED ISSUE #336: Problems with compute_deltas
Hi,
I found some weird behaviour in the compute_deltas function when adding multiple channels. I'm not sure it is a bug, maybe I just expect this function to work in a way different to the actual one.
As you can see from the minimal example below, the derivatives are different when I switch from 1 to 2 channels. I think the channels should be processed independently, while here there is a kind of interaction. 

```
from torchaudio.functional import compute_deltas
import torch

# single channel audio
spectrogram=torch.rand(1,20,75)
deltas=compute_deltas(spectrogram)

print(deltas)
print(deltas.shape)


# audio with 2 channels: the signals are the same 
# and the derivatives should be the same on the different channels.

spectrogram=torch.cat([spectrogram,spectrogram])
deltas=compute_deltas(spectrogram)

print(deltas[0,:])
print(deltas[1,:])
print(deltas.shape)

# the derivatives are different now. This should't happen because the channels must be processed independently
```

According to my understanding the issue is that the function compute_deltas uses in some points specgram.shape[0], where dim=0 is the channel dimension (i.e. the batch dimension). The problem can be simply fixed by replacing all "specgram.shape[0]" with 1.  I noticed this issue because the performance of my speech recognizer didn't improve when adding the derivatives of the MFCC coefficients. With this fix everything works as expected.

vincentqb:
Extends #327 to other transforms.

Fixes #336 also.


cpuhrsch:
I'd also update the README alongside this to explain the ellipsis convention.

vincentqb:
> I'd also update the README alongside this to explain the ellipsis convention.

Updated. We were already using ellipses in the README. :)

README.md
```diff
@@ -129,7 +129,7 @@ Transforms expect and return the following dimensions.
 * `MuLawDecode`: (channel, time) -> (channel, time)
 * `Resample`: (channel, time) -> (channel, time)
 
-Complex numbers are supported via tensors of dimension (..., 2), and torchaudio provides `complex_norm` and `angle` to convert such a tensor into its magnitude and phase.
+Complex numbers are supported via tensors of dimension (..., 2), and torchaudio provides `complex_norm` and `angle` to convert such a tensor into its magnitude and phase. Here, and in the documentation, we use ellipsis "..." as a placeholder for the rest of the dimensions of a tensor.
```
cpuhrsch: I'd also briefly mention the purpose of that (i.e. batching, etc.). "rest of the dimensions" to me at least doesn't imply a functionality.

vincentqb: Agree: "we use an ellipsis "..." as a placeholder for the rest of the dimensions of a tensor, e.g. an optional batching dimension."

PR #348, opened 2019-11-21
Move augmentations in transforms

vincentqb:
* Moving new augmentations from #285 in transforms.
* Linking missing transforms in documentations


zhangguanheng66:
Are those transforms in the last release? Any BC breaking issue?

vincentqb:
> Are those transforms in the last release? Any BC breaking issue?

None of these were on the last release, only in master.

cpuhrsch:
As a TODO we should add splitting the compute component of these augmentations out into functionals.

PR #350, opened 2019-11-21
Move batch from vocoder transform to functional

vincentqb:
* Moving batch pack/unpack from transform (vocoder) to functional, as done elsewhere.
* Also use ellipses for batch in docstring.


vincentqb:
#337

PR #355, opened 2019-11-26
Backend switch

### LINKED ISSUE #329: Inconsistent length between info() and load() for MP3 files
## 🐛 Bug

For single channel MP3 files, the length returned when calling torchaudio.info() is slightly longer than torchaudio.load(). For files with a sample rate of 16KHz the length provided by info is always 576 longer and for a sample rate of 48KHz the length is 1152 longer. 

## To Reproduce

```python
import torchaudio
path = ... # some MP3 file location
print(torchaudio.info(path)[0].channels)  # Verify that file is single channel
print(torchaudio.info(path)[0].length)
print(torchaudio.load(path)[0].size(1))
```

## Expected behavior

The above code should output identical lengths, e.g.
```
1
96768
96768
```

## Environment

Installed with pip using `conda env create -f environment.yml` where `environment.yml` contains the dependencies for the project

`0.3.0`

```
PyTorch version: 1.2.0
Is debug build: No
CUDA used to build PyTorch: 10.0.130

OS: Ubuntu 18.04.3 LTS
GCC version: (Ubuntu 7.4.0-1ubuntu1~18.04.1) 7.4.0
CMake version: version 3.10.2

Python version: 3.7
Is CUDA available: Yes
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: GeForce GTX 1050 Ti
Nvidia driver version: 430.50
cuDNN version: Could not collect

[pip] numpy==1.16.4
[pip] torch==1.2.0
[pip] torchaudio==0.3.0
[pip] torchvision==0.4.0a0+6b959ee
[conda] _pytorch_select           0.1                       cpu_0    anaconda
[conda] blas                      1.0                         mkl    conda-forge/label/cf201901
[conda] mkl                       2019.4                      243    anaconda
[conda] mkl-service               2.3.0            py37he904b0f_0    anaconda
[conda] mkl_fft                   1.0.12           py37ha843d7b_0    anaconda
[conda] mkl_random                1.0.2            py37hd81dba3_0    anaconda
[conda] pytorch                   1.2.0           py3.7_cuda10.0.130_cudnn7.6.2_0    pytorch
[conda] torchaudio                0.3.0                    pypi_0    pypi
[conda] torchvision               0.4.0                py37_cu100    pytorch
```

## Additional context

Checking the length of the files with `soxi -D filename`, the length reported by info appears to be the correct one. The example code above behaves as expected (i.e. info and load agree on lengths) when using a FLAC file in my environment.

The MP3 files tested came from the [Common Voice](https://voice.mozilla.org/en) data set and converting some of the FLAC files in the [LibriSpeech](http://www.openslr.org/12) data set to MP3s.

vincentqb:
Introduce a backend switch in a similar way to [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/__init__.py#L18) from pytorch/vision#153.

- [x] Offer an option to change backend to load files, as in torchvision (https://pytorch.org/docs/stable/torchvision/index.html).
- [x] Import sox only when sox is used at runtime, e.g. sox_effects or loading files.
- [x] Offer wrapper function to switch between backends for load/save. (Maintained current interface to avoid BC-breaking change.)
- [x] Move sox functions from `__init__` to `torchaudio.sox_backend`. (`torchaudio.sox_effects` does not change.)
- [x] Add deprecation warning if calling sox functions using for instance `torchaudio.initialize_sox`.
- [x] Offer pysoundfile to experiment with new interface.
- [x] Add tests for soundfile backend.
- [x] Add test to load/save with different backends
- [ ] ~Fix librosa test appearing here if not flaky? #372~

For later:
- [ ] Add libsndfile directly (pysoundfile uses numpy arrays when wrapping libsndfile).
- [ ] Add decorator to restrict which backend is supported by given function, see [comment](https://github.com/pytorch/audio/pull/355#discussion_r354948733).
- [ ] Make sure mechanism works in parallel context (e.g. could add a backend parameter to functions needing it)
- [ ] Make sure mechanism is torchscriptable (e.g. torchscript currently does not support global variable)

Add to release notes:
- [ ] `SoxEffectsChain.EFFECTS_AVAILABLE` replaced by `SoxEffectsChain().EFFECTS_AVAILABLE`

[Comparison of backends](https://github.com/faroit/python_audio_loading_benchmark)
[Internal doc](https://fb.quip.com/3kdTAwnXWC34)

Fixes #329 by offering other backend as options. As such, this also is a first step in addressing #357.


test/test.py
```diff
@@ -7,160 +7,240 @@
 import os
 
 
+class AudioBackendScope:
```
vincentqb: This would be convenient in general, but I wasn't planning to introduce this as a feature yet since torchvision doesn't have this.

torchaudio/__init__.py
```diff
@@ -67,36 +91,35 @@ def load(filepath,
...
-    sample_rate = _torch_sox.read_audio_file(filepath,
-                                             out,
-                                             channels_first,
-                                             num_frames,
-                                             offset,
-                                             signalinfo,
-                                             encodinginfo,
-                                             filetype)
+    if get_audio_backend() == "sox":
+        waveform, sample_rate = sox_backend.load(
```
vincentqb: A better way to branch would be by doing a conditional import but with the same name
```
if get_audio_backend() == "sox":
  from sox_backend import load
elif get_audio_backend() == "soundfile":
  from _soundfile_backend import load
else:
  raise NotImplementedError

waveform, sample_rate = load(
            filepath,
            out=out,
            normalization=normalization,
            channels_first=channels_first,
            num_frames=num_frames,
            offset=offset,
            filetype=filetype,
        )
```

vincentqb:
@cpuhrsch -- For this PR, I'd focus on the interface for the user, and leave a new backend for later. This still addresses the main issue of completely blocking the import of torchaudio when there is an issue with sox. Thoughts?

test/test.py
```diff
@@ -7,160 +7,230 @@
...
-        # test precision
-        new_precision = 32
-        new_filepath = os.path.join(self.test_dirpath, "test.wav")
-        si, ei = torchaudio.info(sinewave_filepath)
-        torchaudio.save(new_filepath, y, sr, new_precision)
-        si32, ei32 = torchaudio.info(new_filepath)
-        self.assertEqual(si.precision, 16)
-        self.assertEqual(si32.precision, new_precision)
-        os.unlink(new_filepath)
+        for backend in ["sox"]:
```
cpuhrsch: You don't have to do this. You can call a function within a score. Prefix this with an underscore and then write another test that calls it into within different scopes.

torchaudio/__init__.py
```diff
@@ -1,33 +1,61 @@
...
-    if src.is_cuda:
-        raise TypeError('Expected a CPU based tensor, got %s' % type(src))
+_audio_backend = "sox"
+
+
+def set_audio_backend(backend):
+    """
+    Specifies the package used to load.
+    Args:
+        backend (string): Name of the backend. one of {'sox'}.
```
cpuhrsch: nit: add soundfile

Having references to the sources of those backends in the doctstring could be useful too

torchaudio/__init__.py
cpuhrsch: You might be able to save on these if statements if you overwrite the load functions etc. when switching backends, but i'd almost chalk that up under a performance optimization, so it's not necessary yet

cpuhrsch: You could still assign to a local load function and then move these branches higher (which will save on indentations) and make the code more readable

cpuhrsch:
Looks good so far!

From what I gather from the PR description there are no BC-breaking changes introduced here?

test/test.py
```diff
@@ -7,160 +7,230 @@
...
+                    offset = 201
+                    # load entire mono sinewave wav file, load a partial copy and then compare
+                    input_sine_path = os.path.join(
+                        self.test_dirpath, "assets", "sinewave.wav"
+                    )
+                    x_sine_full, sr_sine = torchaudio.load(input_sine_path)
+                    x_sine_part, _ = torchaudio.load(
+                        input_sine_path, num_frames=num_frames, offset=offset
+                    )
+                    l1_error = (
```
cpuhrsch: I don't want to be "that guy", but these code format changes make this harder to review. You could make them the last commit and continue to maintain that, or you could send a separate PR later on. They also introduce a lot of meaningless git blame changes.

torchaudio/__init__.py
```diff
@@ -125,85 +148,26 @@ def save(filepath, src, sample_rate, precision=16, channels_first=True):
...
-
-def save_encinfo(filepath,
-                 src,
-                 channels_first=True,
-                 signalinfo=None,
-                 encodinginfo=None,
-                 filetype=None):
-    r"""Saves a tensor of an audio signal to disk as a standard format like mp3, wav, etc.
+    if get_audio_backend() == "sox":
+        from torchaudio import sox_backend
```
cpuhrsch: I'd carefully make sure that these repeated imports aren't expensive due to some kind of initialization code in sox.

cpuhrsch: or soundfile for that matter

torchaudio/__init__.py
```diff
@@ -125,85 +148,26 @@ def save(filepath, src, sample_rate, precision=16, channels_first=True):
...
-    extension = os.path.splitext(filepath)[1]
-    filetype = extension[1:] if len(extension) > 0 else filetype
-    # transpose from C x L -> L x C
-    if channels_first:
-        src = src.transpose(1, 0)
-    # save data to file
-    src = src.contiguous()
-    _torch_sox.write_audio_file(filepath, src, signalinfo, encodinginfo, filetype)
+
+def save_encinfo(*args, **kwargs):
```
cpuhrsch: The [current docs for encinfo](https://github.com/pytorch/audio/blob/c74e580f92c11419a9a3c02fd2d0e301e3347597/torchaudio/__init__.py#L140) don't even reference the function in the example. This is a strange one.

torchaudio/__init__.py
```diff
@@ -220,149 +184,66 @@ def info(filepath):
...
-            (Default: ``sox_option_default`` or ``2``)
-    Returns:
-        sox_option_t: A sox_option_t type
-    """
-    if i is None:
-        return _torch_sox.sox_option_t
-    else:
-        return _torch_sox.sox_option_t(i)
 
+def sox_encodinginfo_t(*args, **kwargs):
```
cpuhrsch: encoding information, signal information etc. are very useful functions in general. we could also think about adding some backend independent interfaces, but maybe after the release.

torchaudio/_soundfile_backend.py
```diff
@@ -0,0 +1,125 @@
...
+
+def load(
+    filepath,
+    out=None,
+    normalization=True,
+    channels_first=True,
+    num_frames=0,
+    offset=0,
+    filetype=None,
+):
```
cpuhrsch: I notice you're repeating the docstring here. But you call into the load backends separately. I don't think the user will be able to see this unless she looks at this function which is part of a private backend. Here assigning a backend function to the module unction could help use actually choose the correct doc string at runtime.

However, the static documentation won't be able to do that.

So instead I'd say it makes sense to have a single docstring for our load function and then reference it here.

torchaudio/_soundfile_backend.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def info(filepath):
+    r"""Gets metadata from an audio file without loading the signal.
+
+     Args:
+        filepath (str): Path to audio file
+
+     Returns:
+        Tuple[sox_signalinfo_t, sox_encodinginfo_t]: A si (sox_signalinfo_t) signal
```
cpuhrsch: nit: you probably don't mean to mention sox here

test/test.py
vincentqb: Yeah, (1) this was definitely not meant to be in, and (2) this was not quite ready for review :)

torchaudio/__init__.py
vincentqb: One way of avoiding repeated import is to import at the beginning and catch import errors, as done in [vision](https://github.com/pytorch/vision/blob/master/torchvision/transforms/transforms.py#L7). However, local test on my mac don't seem to see a cost to repeated import and seem to properly fetch the cached version:
```
In [1]: %timeit import torchaudio                                                                  
The slowest run took 17.73 times longer than the fastest. This could mean that an intermediate result is being cached.
637 ns ± 1.07 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: %timeit import _torch_sox                                                                  
76.3 ns ± 1.84 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

torchaudio/__init__.py
vincentqb: Good point

torchaudio/sox_effects.py
```diff
@@ -84,6 +92,9 @@ def __init__(self, normalization=True, channels_first=True, out_siginfo=None, ou
         self.normalization = normalization
         self.channels_first = channels_first
 
+        # Define in __init__ to avoid calling at import time
+        self.EFFECTS_AVAILABLE = set(effect_names())
```
vincentqb: This is BC-breaking in the sense that users could check the `EFFECTS_AVAILABLE` without creating an instance.

torchaudio/_soundfile_backend.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+def load(
+    filepath,
+    normalization=True,
+    channels_first=True,
+    num_frames=0,
+    offset=0,
+    filetype=None,
+    **_
```
vincentqb: Other parameters will be silently dropped. This may not be desired, but it convenient in code :)

vincentqb:
I can't reproduce locally the librosa error.
```
❯ conda create -n librosa-conda python=3.7
❯ conda activate librosa-conda
❯ conda install -c pytorch pytorch
❯ conda install -c conda-forge sox pysoundfile librosa
❯ conda install backports.tempfile
❯ python setup.py clean --all
❯ MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ NO_CUDA=1 python setup.py install

❯ python test/test_transforms.py
Ran 23 tests in 0.845s

OK
❯ python test/test.py                   
Ran 7 tests in 0.164s

OK
```

test/test.py
```diff
@@ -7,14 +7,48 @@
 import os
 
 
+def _extract_digits(s):
+    return int("".join([i for i in s if i.isdigit()]))
```
cpuhrsch: You can use "filter"

test/test.py
```diff
@@ -7,14 +7,48 @@
...
+
+
 class Test_LoadSave(unittest.TestCase):
+
+    default_audio_backend = torchaudio.get_audio_backend()
+
     test_dirpath, test_dir = common_utils.create_temp_assets_dir()
     test_filepath = os.path.join(test_dirpath, "assets",
                                  "steam-train-whistle-daniel_simon.mp3")
+    test_filepath_wav = os.path.join(test_dirpath, "assets",
```
cpuhrsch: Isn't this the same path as above?

test/test.py
```diff
@@ -77,7 +115,54 @@ def test_1_save(self):
         self.assertEqual(si32.precision, new_precision)
         os.unlink(new_filepath)
 
+        torchaudio.set_audio_backend(self.default_audio_backend)
+
+    def test_1_save_soundfile_sine(self):
```
cpuhrsch: Doesn't the same advise as above apply where you don't need to copy paste the test code from the sox test

cpuhrsch: It could also be interesting to try saving with sox, loading with sndfile and vice-versa.

test/test.py
```diff
@@ -171,5 +300,21 @@ def test_5_get_info(self):
         self.assertEqual(si.rate, rate)
         self.assertEqual(ei.bits_per_sample, precision)
 
+        torchaudio.set_audio_backend(self.default_audio_backend)
+
+    def _test_5_get_info_soundfile(self):
```
cpuhrsch: I think there's repetition again?

test/test.py
```diff
@@ -171,5 +300,21 @@ def test_5_get_info(self):
...
+        torchaudio.set_audio_backend("soundfile")
+
+        input_path = os.path.join(self.test_dirpath, 'assets', 'sinewave.wav')
+        channels, samples, rate, precision = (1, 64000, 16000, 16)
+        si = torchaudio.info(input_path)
+        self.assertEqual(si.channels, channels)
+        self.assertEqual(si.frames, samples)
+        self.assertEqual(si.samplerate, rate)
+        si_precision = _extract_digits(si.subtype)
```
cpuhrsch: If this is necessary we should at least write out as a todo to make this consistent. That can be done via a wrapper class that uses getattribute etc to align these attributes consistently.

torchaudio/__init__.py
```diff
@@ -242,6 +280,11 @@ def sox_signalinfo_t():
         >>> si.precision = 16
         >>> si.length = 0
     """
+
+    if get_audio_backend() != "sox":
```
cpuhrsch: You could create a generic decorator to write this less often

```
def _backend_guard(backends):
    def decorator(fn):
        @functools.docs-decorator(fn) # not sure about the name
        def _fn(*args, **kwargs):
            if get_audio_backend() not in backends:
                raise Runtime("fn {} requires backend to be one of".format(fn.__name__, backends)
           fn(*args, **kwargs)
       return _fn

```


```
@backend_support(['sox'])
def sox_signalinfo_t():
      ....
```

test/test.py
cpuhrsch: There isn't because info returns a different struct depending on the backend.

test/test.py
cpuhrsch: It should make sense to standardize on sox since otherwise we'll introduce BC-breaking changes to support this new backend.

torchaudio/__init__.py
```diff
@@ -128,13 +149,17 @@ def save(filepath, src, sample_rate, precision=16, channels_first=True):
...
-    si = sox_signalinfo_t()
-    ch_idx = 0 if channels_first else 1
-    si.rate = sample_rate
-    si.channels = 1 if src.dim() == 1 else src.size(ch_idx)
-    si.length = src.numel()
-    si.precision = precision
-    return save_encinfo(filepath, src, channels_first, si)
+
+    if get_audio_backend() == "sox":
+        func = _sox_backend.save
```
cpuhrsch: You could do
```
getattr(get_audio_backend_module() , 'save')
```

Do we want get_audio_backend()  to return a module or a string that is the name?

A module can yield the name as well via introspection

test/test.py
vincentqb: `wav` vs `mp3` file, since soundfile doesn't support `mp3`.

torchaudio/__init__.py
vincentqb: Nice, but let's make that a separate PR.

cpuhrsch:
A small note on load/save - we're currently using allclose to check the identity. I'm not sure if we should restrict ourselves to bit-wise here. Especially if loading and saving across contexts.

torchaudio/__init__.py
```diff
@@ -220,7 +246,9 @@ def info(filepath):
          >>> si, ei = torchaudio.info('foo.wav')
          >>> rate, channels, encoding = si.rate, si.channels, ei.encoding
      """
-    return _torch_sox.get_info(filepath)
+
+    info = getattr(_get_audio_backend_module(), 'info')
```
cpuhrsch: nit: you can write this in one line ```getattr(_get_audio_backend_module(), 'info')(filepath)``` if you want.

torchaudio/sox_effects.py
cpuhrsch: Why is this change necessary?

vincentqb: If `effect_names` is a class variable, the python function `effect_names` gets run at import time, which then imports `_torch_sox` for `_torch_sox.get_effect_names`.

torchaudio/__init__.py
vincentqb: It is shorter, but is it more readable?

vincentqb:
> A small note on load/save - we're currently using allclose to check the identity. I'm not sure if we should restrict ourselves to bit-wise here. Especially if loading and saving across contexts.

You mean make the test stricter by checking for exact bitwise match?

torchaudio/_soundfile_backend.py
vincentqb: When you say "reference", I assume you mean 
```
"""See torchaudio.save"""
```

Unless you meant copying docstring from another with something like `functools.wraps()`?

torchaudio/_soundfile_backend.py
vincentqb: I removed this.

test/test.py
vincentqb: I removed the need for this function.

vincentqb:
> A small note on load/save - we're currently using allclose to check the identity. I'm not sure if we should restrict ourselves to bit-wise here.

I'm leaning for staying with allclose for this PR since that's what all the other tests are doing. (Added a test to make sure load/save across backend is allclose.) Thoughts?

torchaudio/__init__.py
```diff
@@ -274,6 +304,11 @@ def sox_encodinginfo_t():
         >>> ei.opposite_endian = torchaudio.get_sox_bool(0)
 
     """
+
+    if get_audio_backend() != "sox":
+        raise ImportError
```
fmassa: Instead of just raising an `ImportError`, it might be better user experience to explicitly say the "function YY not supported for backend XX".

test/test.py
```diff
@@ -78,34 +113,36 @@ def test_1_save(self):
...
+    def _test_2_load(self, test_filepath, length):
         # check normal loading
-        x, sr = torchaudio.load(self.test_filepath)
+        x, sr = torchaudio.load(test_filepath)
         self.assertEqual(sr, 44100)
-        self.assertEqual(x.size(), (2, 278756))
-
-        # check no normalizing
-        x, _ = torchaudio.load(self.test_filepath, normalization=False)
-        self.assertTrue(x.min() <= -1.0)
```
cpuhrsch: Why was this removed?

torchaudio/__init__.py
cpuhrsch: The trace should include the name of the function or you can maybe get the name of the function from locals()

torchaudio/_soundfile_backend.py
cpuhrsch: Yes that's correct.

torchaudio/sox_effects.py
cpuhrsch: So you're concerned that it's possible that someone imports torchaudio, checks this variable and finds there are no effects available? But then once that persons uses load or such once it'll be populated?

test/test.py
vincentqb: Good point. Added.

torchaudio/__init__.py
vincentqb: Alright, added to standardize import error. :)

torchaudio/__init__.py
vincentqb: Agree, added along with decorator.

torchaudio/sox_effects.py
vincentqb: (Corrected typo in last response.)

I was just highlighting that it won't be able to do `SoxEffectsChain.EFFECTS_AVAILABLE` anymore. Instead, a user will have to do `SoxEffectsChain().EFFECTS_AVAILABLE`. I do not think anyone was doing the first. 

Calling load won't change this behavior.

cpuhrsch: Ok that seems fine, I'd only mention it in the release notes for completeness.

vincentqb: Added to PR description. Resolving discussion.

vincentqb:
Signals from the tests from the tests are clouded by #372.

test/test.py
```diff
@@ -126,7 +186,38 @@ def test_3_load_and_save_is_identity(self):
...
+            input_path = os.path.join(self.test_dirpath, 'assets', 'sinewave.wav')
+            tensor1, sample_rate1 = torchaudio.load(input_path)
+
+            output_path = os.path.join(self.test_dirpath, 'test.wav')
+            torchaudio.save(output_path, tensor1, sample_rate1)
+
+        with AudioBackendScope(backend2):
+            tensor2, sample_rate2 = torchaudio.load(output_path)
+
+        # tensor1 = tensor1.type(torch.FloatTensor)
```
cpuhrsch: nit: maybe you wanted to remove these?

vincentqb: Thanks for catching this!

tadas-subonis:
Is there any reason why 
```
    # normalize if needed
    # _audio_normalization(out, normalization)
```
was commented out?

PR #359, opened 2019-12-03
remove support of smoke test on python 2.7

mingbowan:
reference: [smoke tests](https://circleci.com/workflow-run/1610f891-7bd1-4665-9040-56da917a006c) failed on python 2.7 due to unicode support error,   since python 2.7 will not be supported in next release, it's safe to remove. cc: @vincentqb


mingbowan:
Since current python 2.7 smoke tests are failing, merge this now will give us better signal. Discussed with @vincentqb , will merge it now

PR #361, opened 2019-12-04
Move import _torch_sox inside function calls

vincentqb:
Import sox only when sox is used at runtime, e.g. sox_effects or loading files, from #355. As noted [here](https://github.com/pytorch/audio/pull/355#discussion_r353811539), repeated imports of `_torch_sox` do seem to properly hit cache.

### DIFF
--- a/test/test.py
+++ b/test/test.py
@@ -7,12 +7,32 @@ import math
 import os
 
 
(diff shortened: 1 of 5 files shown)


vincentqb:
Pushed work to #355

PR #362, opened 2019-12-07
JIT resample waveform

Oktai15:
Make `resample_waveform` jitable.

### DIFF
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -32,6 +32,22 @@ BLACKMAN = 'blackman'
 WINDOWS = [HAMMING, HANNING, POVEY, RECTANGULAR, BLACKMAN]
 
 
+def gcd(a, b):
+    # type: (int, int) -> int
+    """Calculate the Greatest Common Divisor of a and b.
+
+    Unless b==0, the result will have the same sign as b (so that when
+    b is divided by it, the result comes out positive).
+    """
+
+    try:
+        return math.gcd(a, b)
+    except AttributeError:
+        while b:
+            a, b = b, a % b
+        return a
+
+
 def _next_power_of_2(x):
     r"""Returns the smallest power of 2 that is greater than x
     """
@@ -733,7 +749,7 @@ def _get_LR_indices_and_weights(orig_freq, new_freq, output_samples_in_unit, win
 
 def _lcm(a, b):
     # type: (int, int) -> int
-    return abs(a * b) // math.gcd(a, b)
+    return abs(a * b) // gcd(a, b)
 
 
 def _get_num_LR_output_samples(input_num_samp, samp_rate_in, samp_rate_out):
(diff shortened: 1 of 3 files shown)


Oktai15:
I have no problem with `torch.device` as type when test `resample_waveform` function. However, when I use `resample_waveform` in some `torch.nn.Module`, I got the following error:
```
RuntimeError:                                                                                                                                                                                               
Unknown type name 'torch.device'
```

torchaudio/compliance/kaldi.py
```diff
@@ -779,7 +787,9 @@ def _get_num_LR_output_samples(input_num_samp, samp_rate_in, samp_rate_out):
     return num_output_samp
 
 
+@torch.jit.export
 def resample_waveform(waveform, orig_freq, new_freq, lowpass_filter_width=6):
```
vincentqb: We no longer compile functionals and transforms by default. Therefore, we no longer add annotations. However, we do test that the [functionals](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L20) and [transforms](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L23) can be compiled.

torchaudio/compliance/kaldi.py
```diff
@@ -731,11 +735,15 @@ def _get_LR_indices_and_weights(orig_freq, new_freq, output_samples_in_unit, win
     return min_input_index, weights
 
 
+@torch.jit.export
 def _lcm(a, b):
+    # type: (int, int) -> int
     return abs(a * b) // fractions.gcd(a, b)
 
 
+@torch.jit.export
```
vincentqb: By the way, compilation is now recursive, and so functions called within functions that are compiled no longer need to be annotated.

torchaudio/compliance/kaldi.py
```diff
@@ -813,7 +823,8 @@ def resample_waveform(waveform, orig_freq, new_freq, lowpass_filter_width=6):
 
     window_width = lowpass_filter_width / (2.0 * lowpass_cutoff)
     first_indices, weights = _get_LR_indices_and_weights(orig_freq, new_freq, output_samples_in_unit,
-                                                         window_width, lowpass_cutoff, lowpass_filter_width)
+                                                         window_width, lowpass_cutoff, 
+                                                         lowpass_filter_width, waveform.device)
     weights = weights.to(waveform.device)  # TODO Create weights on device directly
```
vincentqb: nit: Please remove this line once the weights are moved to device.

Oktai15:
@vincentqb 
 > ... I infer that you would like to make sure that Resample can be compiled. If this is so, then you should instead add the relevant test for the Resample transform, see comment below.

Yes, that's right. I will modify code in few days according to your comments and add tests. But I suppose that I will have the same problem that I mentioned above about module (in the same time, transform as function should work).

Issue #351, opened 2019-11-25
Reverse MelScale, Griffin-Lim transformation


jaeyeun97:
## 🚀 Feature

I have implemented PyTorch versions of `librosa`'s ` mel_to_stft` and `griffinlim`, which I believe would be useful for any generation task using mel-spectrograms. You can see it as a part of my Melnet implementation here: https://github.com/jaeyeun97/MelNet/blob/master/audio.py. While it currently does not conform to the standards of this repo with a little touch here and there I think it would be a good addition.

## Motivation

1. `librosa` has `mel_to_stft`, which solves for a spectrogram of linear frequency scale from a Mel scale spectrogram. Having a pure PyTorch implementation would remove the need of a `librosa` dependency and add support for GPU processing (which speeds up the processes significantly)

2. Similarly, Griffin-Lim algorithm provides an approximation for phase in order to provide a better spectrogram inversion, and since this project provides an implementation for `istft` I thought I would use it to port the `griffinlim` function from `librosa`.

## Pitch

I would like to add the code for Griffin-Lim and mel-to-linear scaling I referenced above to this repository.

I don't know why this hasn't been done before--was there any license issue, etc. regarding porting the code? Otherwise, I think this would be a relatively a painless merge into the repository.

maximecb:
This would be useful to me :+1:

vincentqb:
Yes, that sounds like good additions. Please feel free to open pull requests.

jaeyeun97:
Cool, had a look into the test codes--I'll open a pull request soon.

jaeyeun97:
@vincentqb Would you like these as separate pull requests or as one?

vincentqb:
> @vincentqb Would you like these as separate pull requests or as one?

I usually separate, but please do what's convenient for you for this :)

vincentqb:
Thanks again for working on this @jaeyeun97! #448 merged the second request made in this issue.

PR #365, opened 2019-12-13
Griffin-Lim Transformation Implementation

jaeyeun97:
Implementation of Griffin-Lim Transformation for feature request #351 
@vincentqb I am not quite sure how I might go about writing the test code for this.
Meanwhile, I have a simple comparison of the waveforms of the `steam-train-whistle-daniel_simon.mp3` file, which by the looks and sounds seem to be working properly.

GriffinLim


vincentqb:
Thanks for submitting the PR!

* We put the computation part of a transform in `torchaudio/functional.py` and then make the code in `torchaudio/transforms.py` just a wrapper around the functional.
* torchaudio's code can be compared against librosa's output. And torchaudio's code is also tested for jitability. See examples in `test/test_functional.py` and `test/test_transforms.py`. 
* Python 2 support will be dropped after the upcoming release, so this pull request only needs to support Python 3.

jaeyeun97:
@vincentqb Since `torchaudio.functional.istft` is not jitable, neither is this one--should I omit the latter test?

jaeyeun97:
@vincentqb As a part of implementing jitability for GriffinLim, I ended up writing jitability for istft as well. It passes the tests on my machine.

torchaudio/functional.py
```diff
@@ -151,7 +152,8 @@ def istft(
     assert 0 < win_length <= n_fft
 
     if window is None:
-        window = torch.ones(win_length, requires_grad=False, device=device, dtype=dtype)
```
vincentqb: nit: indentation should be 4 spaces -- I recommend running the code in flake8 to catch those typos :)

torchaudio/functional.py
```diff
@@ -151,7 +152,8 @@ def istft(
...
+       window = torch.ones(win_length)
```
vincentqb: Is there a reason to create the tensor and then move to a different device?

torchaudio/functional.py
```diff
@@ -174,9 +176,8 @@ def istft(
     # each column of a channel is a frame which needs to be overlap added at the right place
     ytmp = ytmp.transpose(1, 2)  # size (channel, n_fft, n_frame)
 
-    eye = torch.eye(n_fft, requires_grad=False, device=device, dtype=dtype).unsqueeze(
-        1
-    )  # size (n_fft, 1, n_fft)
+    eye = torch.eye(n_fft)
```
vincentqb: Same about moving to device.

torchaudio/transforms.py
```diff
@@ -70,6 +71,68 @@ def forward(self, waveform):
...
+    """
+    __constants__ = ['n_fft', 'n_iter', 'win_length', 'hop_length', 'power', 'normalized',
+                     'length', 'momentum']
+
+    def __init__(self, n_fft=400, n_iter=32, hop_length=None, win_length=None,
+                 window_fn=torch.hann_window, wkwargs=None, normalized=False,
+                 power=2, length=None, momentum=0.99, rand_init=True):
+        super(GriffinLim, self).__init__()
+
+        assert momentum < 1, 'momentum=%s > 1 can be unstable' % momentum
```
vincentqb: How about momentum = 1?

nit: the message sounds more like a warning, then an error. Is the algorithm compatible with > 1?

torchaudio/functional.py
```diff
@@ -118,15 +119,15 @@ def istft(
     """
     stft_matrix_dim = stft_matrix.dim()
     assert 3 <= stft_matrix_dim, "Incorrect stft dimension: %d" % (stft_matrix_dim)
-    assert stft_matrix.nelement() > 0
+    assert stft_matrix.numel() > 0
```
vincentqb: STFT is decorated with `@torch.jit.ignore` and so is not compiled. What are the changes to STFT?

torchaudio/functional.py
jaeyeun97: ah I have been having a problem with my coc.nvim settings recently--thanks for pointing it out!

torchaudio/functional.py
jaeyeun97: It has to do with the JITability of the function: the signature of `torch.ones` seem not to allow for `device` and `dtype` setting when the `out` parameter is not specified.

torchaudio/functional.py
jaeyeun97: Yep see the comment above.

torchaudio/functional.py
jaeyeun97: This is also a JITability workaround; `nelements` and `numel` are basically the same functions but one is supported by Torchscript and the other is not.

torchaudio/transforms.py
jaeyeun97: I agree, this is more of a warning. I had the assertions set because Perraudin and Sondergaard states that momentum = 1 is the limit of the stability of the algorithm with some results indicating that it fails to converge with momentum > 1 (see figure 3 of P&S). It'll still run and return some (although not ideal) results so if there is some warning mechanism torchaudio is using that would be more appropriate. (Speaking of which, is there any?)

vincentqb: Thanks for the clarification! Let's keep the assertion then. 

(For warnings, we often use `warnings.warn`.)

vincentqb:
> @vincentqb As a part of implementing jitability for GriffinLim, I ended up writing jitability for istft as well. It passes the tests on my machine, but please double check?

Thanks! jitability test looks good to me! and thanks for taking care of `istft`.

FYI: The `ones` and `eye` tensor can be created on device directly. No need for two separate steps.

torchaudio/functional.py
vincentqb: The `ones` and `eye` tensor can be created on device directly. No need for two separate steps.

jaeyeun97: The last time I checked, the function signatures required an `out` parameter when specifying the device. Since `out` is of type `Tensor` and `window` is of `Optional[Tensor]`, it complained. I might be misinterpreting the errors, though.

torchaudio/functional.py
vincentqb: Opened issue pytorch/pytorch#31626 for this.

PR #366, opened 2019-12-13
InverseMelScale Implementation

jaeyeun97:
InverseMelScale implementation for #351 
Test codes are coming!

Edit: It would be great if someone could chime in on the test code guideline for torchaudio?


vincentqb:
If the code is in progress, you can prefix the pull request with [WIP] to indicate it is not yet ready for review/merging.

See also [comment](https://github.com/pytorch/audio/pull/365).

vincentqb:
Test automated testing [here](https://travis-ci.org/pytorch/audio/jobs/624731273?utm_medium=notification&utm_source=github_status) picks up flake8 errors.
```
./torchaudio/transforms.py:206:9: E303 too many blank lines (2)
./torchaudio/transforms.py:206:51: W291 trailing whitespace
./torchaudio/transforms.py:210:1: W293 blank line contains whitespace
./torchaudio/transforms.py:222:37: W291 trailing whitespace
```

jaeyeun97:
`InverseMelScale` code changed significantly to be JITable--it now uses SVD to compute the inverse of the Mel basis matrix.

jaeyeun97:
No longer the case; using SGD for optimization yields best results when auditioned, along with the lowest and most stable loss values. JITability will have to suffer, I guess?

vincentqb:
> No longer the case; using SGD for optimization yields best results when auditioned, along with the lowest and most stable loss values. JITability will have to suffer, I guess?

What do you mean "best results"? Can it be quantified? Can we compare the outputs with librosa, SGD and SVD? If the goal is to compute the inverse of a matrix, what other ways have you tried? inverse, pseudo-inverse? (btw [this post](https://discuss.pytorch.org/t/torch-pinverse-seems-to-be-inaccurate/33616) may be relevant)

In general, I'd say jitability is important, and leave the code snippet with SGD in a comment here, and a TODO note in code with a link to here for when optimizers become jitable too :)

jaeyeun97:
Okay, I'll try to post reasonable test results between original, SVD, and SGD examples.

As a heads up--it's easier when you look at the resulting spectrograms of the tests, which shows a stark difference (mostly by having reasonable data where the missing bases result in random data on the case of SVD or pseudo-inverse methods).

jaeyeun97:
Okay, I have some results here, although they aren't quantitative ones. They generally favor solving by SGD, although the difference between SGD and SVD methods are small if `n_fft == 2^n`.

`n_fft == 400` and `n_mels == 50`:


`n_fft == 512` and `n_mels == 64`:


`n_fft == 2048` and `n_mels == 256`:


@vincentqb Should I still just progress with SVD methods?

vincentqb:
Thanks for running those tests. 

1. Can you provide the codes for each version (e.g. using SVD and SGD)? in the comment here is fine.
1. We need to quantify the differences, though I do agree that we'll need to go with SGD if we measure that SVD is not stable.
1. Do you have a reference we can compare against? Something like [librosa's](https://librosa.github.io/librosa/generated/librosa.feature.inverse.mel_to_audio.html) that we could introduce as a [test](https://github.com/pytorch/audio/blob/master/test/test_transforms.py).

jaeyeun97:
1. You can find the code at https://gist.github.com/jaeyeun97/8651dff509d5b084636ac6c3a7547108, where the SGD version is the one you can currently find in this branch.

2. Comparing with the Librosa version I'd say SGD is the right way to go. Librosa uses L-BGFS for solving the minimum squared difference as well, so having such a process here seems natural. The only quantifying measure I can think of is comparing the Euclidean distance (`torch.dist`?) between the proposed method, librosa's method, and the original spectrogram, since `torch.allclose` does not really work on any of the methods.

jaeyeun97:
The following shows the results including librosa:

`n_fft == 400`, `n_mels == 50`:


`n_fft == 2048`, `n_mels == 256`:

vincentqb:
Let's look at `torch.dist` with `p=0` and `p=1` between 
* `LibrosaInverse(melspec)` and `SGDInverse(melspec)`
* `LibrosaInverse(melspec)` and `SVDInverse(melspec)`

This will help us see the average difference and max difference.

Qualitatively, it's clear the the SVD implementation wouldn't do, so we'll end up with the SGD implementation. We'll need to add a test to make sure that the `p=0` or `p=1` distance between librosa and torchaudio is smaller than some threshold.

jaeyeun97:
Sorry, was on a long trip, will be finishing up this feature in the next few days.

jaeyeun97:
@vincentqb here are the `torch.dist`  results:

|        | SGDInverse | SVDInverse |
| ----- | ---------- | ------------ |
| p=0 | 522046. | 516637. |
| p=1 | 14831.9863 | 1549607.8750 |
| p=2 | 77.0535 | 14585.5244 |

jaeyeun97:
The `p=1` difference above should be enough to justify SGD?

Now for the test I'd say it makes more sense to compare the difference between:
* `torch.dist(LibrosaInverse(MelScale(spec), spec))`
* `torch.dist(SGDInverse(MelScale(spec), spec))`
if we are going to move ahead with SGDInverse.

vincentqb:
Yes, based on these results, let's use the SGD variant. Thanks for providing a gist in [comment](https://github.com/pytorch/audio/pull/366#issuecomment-569685524) with SVD. There's an alternative version of SVD coming up in pytorch/pytorch#29488, but we can revisit after the latter is merged.

> I am not quite sure why `p=0` may be significant, since it is very unlikely (or so I presume) for any corresponding "cells" to be exactly the same after a numerical solving process.

Oops, this was a typo :) I meant `p=1` and `p=2`. Thanks for providing them!

vincentqb:
Let's make the tests green :) Can you rebase your code onto master?

jaeyeun97:
> Let's make the tests green :) Can you rebase your code onto master?

yep done!

vincentqb:
Thanks! Can you do the following?
- [ ] We put the computation part in `torchaudio.functional` and then wrap this in `torchaudio.transforms`. The documentation also needs to be updated for functionals.
- [ ] We need tests (e.g. test/test_functional.py and test/test_transform.py) for correctness. Can you add in the corresponding test file (1) Inverse(specgram) is close to librosa, (2) Inverse(MelSpecgram(waveform)) is close to waveform (3) MelSpecgram(Inverse(specgram)) is close to specgram?

mthrok:
Hi 👋

I am working to wrap up this nice PR. (@jaeyeun97 let me know if you have a compelling reason that I should not.)

Following @vincentqb 's suggestion, I extracted the main part of `InverseMelScale` from `transform` module to `functional` module, and then I noticed that `functional` module does not have an equivalent of `MelScale` implementation there.

So I had offline discussion with @vincentqb, and we agreed that it's okay to leave both `MelScale` and `InverseMelScale` in `transform` module.

The followings are the points brought up;

1. MelScale is creating filter banks and applying it with `matmul`. Filter bank generation is defined in `functional` module. So it's mostly done in `functional` module.
2. The consistency across MelScale and InverseMelScale is important here. (I was writing a numerical compatibility test for functional version of InverseMelScale, but writing it without an equivalent of MelScale was weird and code looks ugly.)

So I will keep the implementation of InverseMelScale transform as is and will add test for numerical compatibility to supplement this PR.

jaeyeun97:
@mthrok Ah sorry, I decided to seriously look for jobs now so that took priority.

> I noticed that `functional` module does not have an equivalent of `MelScale` implementation there.

I was also hesitant on moving code to functional because of this issue, thanks for sorting it out!

mthrok:
@jaeyeun97 I see. Good luck with your job hunting!

jaeyeun97:
@mthrok Thanks! Just reviewed #448 and looks good to me.

jaeyeun97:
Some checks failed but they look like `curl` failures.

PR #368, opened 2019-12-14
resample jit

vincentqb:
Thanks to @Oktai15 in #362 for bringing up support for jit and `resample_waveform`.

### DIFF
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -32,6 +32,22 @@ BLACKMAN = 'blackman'
 WINDOWS = [HAMMING, HANNING, POVEY, RECTANGULAR, BLACKMAN]
 
 
+def gcd(a, b):
+    # type: (int, int) -> int
+    """Calculate the Greatest Common Divisor of a and b.
+
+    Unless b==0, the result will have the same sign as b (so that when
+    b is divided by it, the result comes out positive).
+    """
+
+    try:
+        return math.gcd(a, b)
+    except AttributeError:
+        while b:
+            a, b = b, a % b
+        return a
+
+
 def _next_power_of_2(x):
     r"""Returns the smallest power of 2 that is greater than x
     """
@@ -733,7 +749,7 @@ def _get_LR_indices_and_weights(orig_freq, new_freq, output_samples_in_unit, win
 
 def _lcm(a, b):
     # type: (int, int) -> int
-    return abs(a * b) // math.gcd(a, b)
+    return abs(a * b) // gcd(a, b)
 
 
 def _get_num_LR_output_samples(input_num_samp, samp_rate_in, samp_rate_out):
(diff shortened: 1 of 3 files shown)


torchaudio/compliance/kaldi.py
```diff
@@ -20,7 +19,7 @@
 ]
 
 # numeric_limits<float>::epsilon() 1.1920928955078125e-07
-EPSILON = torch.tensor(torch.finfo(torch.float).eps, dtype=torch.get_default_dtype())
+EPSILON = torch.tensor(torch.finfo(torch.float).eps)
```
Oktai15: It will create EPSILON always on CPU, it is not right?

torchaudio/compliance/kaldi.py
```diff
@@ -92,10 +107,10 @@ def _feature_window_function(window_type, window_size, blackman_coeff):
         # like hanning but goes to zero at edges
         return torch.hann_window(window_size, periodic=False).pow(0.85)
     elif window_type == RECTANGULAR:
-        return torch.ones(window_size, dtype=torch.get_default_dtype())
+        return torch.ones(window_size)
```
Oktai15: The same question: CPU tensor?

Oktai15:
By the way, you can try to do jit kaldi.fbank in this PR (it needs very small changes I suppose)

torchaudio/compliance/kaldi.py
vincentqb: This code needs updating to work on GPU, and we don't officially support GPU for now in torchaudio. I won't do this as part of this PR.

torchaudio/compliance/kaldi.py
```diff
@@ -92,10 +107,10 @@ def _feature_window_function(window_type, window_size, blackman_coeff):
         # like hanning but goes to zero at edges
         return torch.hann_window(window_size, periodic=False).pow(0.85)
     elif window_type == RECTANGULAR:
-        return torch.ones(window_size, dtype=torch.get_default_dtype())
```
zhangguanheng66: why don't we force the torch default type?

vincentqb: torchscript does not support calling `torch.get_default_dtype`. In any case, the `dtype` of `torch.tensor`, `torch.ones` and `torch.arange` is already set to `torch.get_default_dtype`, see [torch.tensor](https://pytorch.org/docs/stable/torch.html#torch.tensor), [torch.ones](https://pytorch.org/docs/stable/torch.html#torch.ones) and [torch.arange](https://pytorch.org/docs/stable/torch.html#torch.arange). Is that what you meant?

vincentqb: (relates to pytorch/pytorch#30774)

vincentqb:
Pushed to #362.

PR #369, opened 2019-12-15
Module GPU test fixes

jaeyeun97:
Tensors initialized but not registered as buffers during module instantiation do not move when the module is moved to another device. This fix registers such tensors as buffers so that they are automatically moved when the module moves.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -82,7 +82,7 @@ class Tester(unittest.TestCase):
         self.assertTrue(waveform_exp.min() >= -1. and waveform_exp.max() <= 1.)
 
     def test_scriptmodule_AmplitudeToDB(self):
-        spec = torch.rand((6, 201))
+        spec = torch.rand((1, 201, 6))
         _test_script_module(transforms.AmplitudeToDB, spec)
 
     def test_scriptmodule_MelScale(self):
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -82,7 +82,7 @@ class Tester(unittest.TestCase):
         self.assertTrue(waveform_exp.min() >= -1. and waveform_exp.max() <= 1.)
 
     def test_scriptmodule_AmplitudeToDB(self):
-        spec = torch.rand((1, 201, 6))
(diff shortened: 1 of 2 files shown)


vincentqb:
Did you have a test failing to do this change? See [FAQ](https://pytorch.org/docs/stable/jit.html#frequently-asked-questions) for `register_buffer` which should be equivalent to what was done.

test/test_transforms.py
```diff
@@ -82,7 +82,7 @@ def test_mu_law_companding(self):
         self.assertTrue(waveform_exp.min() >= -1. and waveform_exp.max() <= 1.)
 
     def test_scriptmodule_AmplitudeToDB(self):
-        spec = torch.rand((6, 201))
```
vincentqb: Reason to change the test?

jaeyeun97: Sorry, that is plain old poor git management. yikes. In its defence, however, it tests for batch dimension.

jaeyeun97:
Pretty much all CUDA tests were failing because of these errors. I read the FAQs you have linked, and it might be a problem of upstream, but the tests kept failing until I explicitly registered them.

test/test_transforms.py
vincentqb: Just checking :)

jaeyeun97:
In hindsight, I should have opened an issue first--do you know where the error might reside? I'm using the latest version of PyTorch release on anaconda.

test/test_transforms.py
jaeyeun97: Also, removed or not, this change is completely irrelevant to the GPU script changes.

jaeyeun97: Reverted.

vincentqb:
> In hindsight, I should have opened an issue first--do you know where the error might reside? I'm using the latest version of PyTorch release on anaconda.

Can you provide more information about the errors and the machine you are using so we can reproduce?

vincentqb:
Thanks for fixing this!

jaeyeun97:
Thanks for bringing it upstream!

PR #370, opened 2019-12-15
Fix MelScale test and documentation

jaeyeun97:
Minor fixes to documentation and tests for `MelScale`.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -86,7 +86,7 @@ class Tester(unittest.TestCase):
         _test_script_module(transforms.AmplitudeToDB, spec)
 
     def test_scriptmodule_MelScale(self):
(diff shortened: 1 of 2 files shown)


test/test_transforms.py
```diff
@@ -86,7 +86,7 @@ def test_scriptmodule_AmplitudeToDB(self):
         _test_script_module(transforms.AmplitudeToDB, spec)
 
     def test_scriptmodule_MelScale(self):
-        spec_f = torch.rand((1, 6, 201))
+        spec_f = torch.rand((1, 201, 6))
```
vincentqb: We don't need to change this. The important part is just that it tests with a random matrix.

jaeyeun97:
yeah it was just a pet peeve of mine since it looked like it was targeted for `n_fft = 400`. All good though.

PR #377, opened 2019-12-26
create tensor directly on device.

vincentqb:
Create directly on device for #365.

Tested on GPU device.


vincentqb:
I self-approve this change.

PR #378, opened 2019-12-27
Adopt native-Python code generation convention

### LINKED ISSUE #304: Apply "Adopt native-Python code generation convention" to this repo
https://github.com/pytorch/vision/pull/1321 wasn't applied to this repo

kostmo:
Compare to application to vision repo in pytorch/vision#1321
See rationale writeup: https://github.com/pytorch/vision/pull/1321#issuecomment-531033978

Fixes #304

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,7 +48,7 @@ jobs:
       - checkout
       - run:
           command: |
-            pip install --user --progress-bar off jinja2
+            pip install --user --progress-bar off jinja2 pyyaml
             python .circleci/regenerate.py
             git diff --exit-code || (echo ".circleci/config.yml not in sync with config.yml.in! Run .circleci/regenerate.py to update config"; exit 1)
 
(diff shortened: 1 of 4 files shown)


.circleci/regenerate.py
```diff
@@ -1,13 +1,121 @@
...
+    base_workflow_name = f"{prefix}binary_{os_type}_{btype}_py{python_version}{unicode_suffix}"
+
+    w.append(generate_base_workflow(base_workflow_name, python_version, unicode, filter_branch, os_type, btype))
+
+    if upload:
+
+        is_py3_linux = os_type == 'linux' and not python_version.startswith("2.")
+
+        # XXX This logic is suspect...
+        upload_job_filter_branch = not is_py3_linux and filter_branch
```
kostmo: Not sure if this was a bug introduced at some point in the in the existing logic.  Currently, the `filters:` yaml chunk is excluded from `*_upload` jobs on Linux with Python 3, so I've crafted my logic to match this behavior.  Not sure what the rationale is, though, or whether that's intentional.

.circleci/sort-yaml.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+
```
kostmo: I have added this little script to prove that my change yields identical output, modulo ordering of YAML keys.

vincentqb: So this is a test file, right? How about adding it as `.circleci/test/test_sort_yaml.py` ?

.circleci/config.yml
```diff
@@ -48,7 +48,7 @@ jobs:
       - checkout
       - run:
           command: |
-            pip install --user --progress-bar off jinja2
+            pip install --user --progress-bar off jinja2 pyyaml
```
vincentqb: This seems to be the only non-cosmetic change to this file, is this the case? It would have been easier to read if all the other changes were done in a separate PR.

.circleci/regenerate.py
vincentqb: Is this code generation (with exception) the same across domains? What are the differences?

kostmo: There are some minor differences with the `vision` repo's `config.yml` file, and by extension the codegen.  To name a few:
* `vision` exercises an extra dimension of configuration with different versions of CUDA + cpu
* it does not have `smoke_test` jobs
* it lacks `filters:` sections in the YAML

.circleci/sort-yaml.py
kostmo: Will do, though it shouldn't be considered a standalone test, but rather a utility to facilitate a manual comparison for this somewhat extensive code change.

.circleci/config.yml
kostmo: True.  I didn't realize I needed this addition until late in the process.

PR #391, opened 2020-01-02
extend batch support

### LINKED ISSUE #383: Is batching-for-transform done?
Hi, I assumed it's done by #337 (which is great!), but at least `transforms.MFCC` is not. What's the current state?
Also related, the docstrings of `transforms` need to be updated. The current version reads like it only deals with a single audio signal.

vincentqb:
* Extend batch support from #337 to MelScale, MelSpectrogram, MFCC. 
* Update documentation for transforms.
* `mask_along_axis_iid` supports batching, so is not updated here, see #401 and #350.

Closes #383

### DIFF
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -103,18 +103,21 @@ class TestFunctional(unittest.TestCase):
 
         self.assertTrue(torch.allclose(ta_out, lr_out, atol=5e-5))
 
-        # test batch
+    def test_batch_griffinlim(self):
 
-        # Single then transform then batch
-        expected = ta_out.unsqueeze(0).repeat(3, 1, 1)
+        tensor = torch.rand((1, 201, 6))
 
-        # Batch then transform
-        specgram = specgram.unsqueeze(0).repeat(3, 1, 1, 1)
-        computed = F.griffinlim(specgram, window, n_fft, hop, ws, 1, normalize,
-                                n_iter, momentum, length, rand_init)
+        n_fft = 400
+        ws = 400
+        hop = 200
+        window = torch.hann_window(ws)
+        power = 2
+        normalize = False
+        momentum = 0.99
+        n_iter = 32
+        length = 1000
 
-        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
-        self.assertTrue(torch.allclose(computed, expected, atol=5e-5))
+        self._test_batch(F.griffinlim, tensor, window, n_fft, hop, ws, power, normalize, n_iter, momentum, length, 0)
 
     def _test_compute_deltas(self, specgram, expected, win_length=3, atol=1e-6, rtol=1e-8):
         computed = F.compute_deltas(specgram, win_length=win_length)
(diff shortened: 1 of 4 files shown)


test/test_transforms.py
```diff
@@ -433,6 +446,30 @@ def test_batch_spectrogram(self):
         self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
         self.assertTrue(torch.allclose(computed, expected))
 
+    def test_batch_melspectrogram(self):
+        waveform, sample_rate = torchaudio.load(self.test_filepath)
+
+        # Single then transform then batch
+        expected = transforms.MelSpectrogram()(waveform).repeat(3, 1, 1, 1)
```
cpuhrsch: This is repetitive. I'm wondering if there is a way of creating a test generator that explicitly exercises this decorator. It's a function that accepts a function with some args and kwargs and then assumes that the first input is to be batchable (so it applies various reshapes to args[0] etc.)

```
def _gen_batchable(self, func, *args, **kwargs):
    self.assertEqual(func(*args, **kwargs), func(*(args[0].reshape(3, -1, -1, -1) + args[1:]), **kwargs)
```
(not sure on the reshape behavior, could also be a lambda).

torchaudio/functional.py
```diff
@@ -946,7 +957,7 @@ def gain(waveform, gain_db=1.0):
     r"""Apply amplification or attenuation to the whole waveform.
 
     Args:
-       waveform (torch.Tensor): Tensor of audio of dimension (channel, time).
+       waveform (torch.Tensor): Tensor of audio of dimension (..., time).
```
cpuhrsch: Maybe ```([...], channel, time)``` to indicate that we still follow that convention of channel x time?

cpuhrsch: Maybe ```([...], channel, time)``` to indicate that we still follow that convention of channel x time?

vincentqb: The `...` is described in the [readme](https://github.com/pytorch/audio/blob/master/README.md), so I would not use `[...]` as we discussed in #337.

We can add `channel` everywhere to emphasize the convention, but that seems repetitive to me, so I wouldn't go for it. Thoughts/

test/test_transforms.py
vincentqb: I followed the format that I did for testing jitability, and simply introduced common functions to test batching. Thoughts?

torchaudio/functional.py
vincentqb: Duplicated comment.

test/test_functional.py
```diff
@@ -103,6 +103,25 @@ def test_griffinlim(self):
 
         self.assertTrue(torch.allclose(ta_out, lr_out, atol=5e-5))
 
+    def test_batch_griffinlim(self):
+
+        torch.random.manual_seed(42)
+        tensor = torch.rand((1, 201, 6))
+
+        n_fft = 400
```
cpuhrsch: Are these parameters special to the batch test?

test/test_functional.py
```diff
@@ -389,6 +394,16 @@ def test_linearity_of_istft4(self):
         data_size = (2, 7, 3, 2)
         self._test_linearity_of_istft(data_size, kwargs4, atol=1e-5, rtol=1e-8)
 
+    def test_batch_istft(self):
+
+        stft = torch.tensor([
+            [[4., 0.], [4., 0.], [4., 0.], [4., 0.], [4., 0.]],
+            [[0., 0.], [0., 0.], [0., 0.], [0., 0.], [0., 0.]],
```
cpuhrsch: Why repeat all 0s instead of something more interesting (maybe just ```[0., 4.]```)

test/test_transforms.py
cpuhrsch: Yeah I'm mostly just trying to iterate on top of that. Might be one of these non-problems I'm trying to fix.

torchaudio/functional.py
```diff
@@ -310,11 +310,15 @@ def griffinlim(
...
     Returns:
-        torch.Tensor: waveform of (channel, time), where time equals the ``length`` parameter if given.
+        torch.Tensor: waveform of (..., time), where time equals the ``length`` parameter if given.
     """
     assert momentum < 1, 'momentum=%s > 1 can be unstable' % momentum
     assert momentum > 0, 'momentum=%s < 0' % momentum
 
+    # pack batch
+    shape = specgram.size()
+    specgram = specgram.reshape([-1] + list(shape[-2:]))
```
cpuhrsch: Instead of reshape I suggest you use view. Reshape dispatches to view if it doesn't require a copy. If this does require a copy I think it'd be good to know about since it's not something we'd expect to be necessary.

torchaudio/functional.py
cpuhrsch: As long as it's documented and standardized so we can apply it consistently this is fine by me

test/test_functional.py
vincentqb: Using the same as `test_torchscript_spectrogram`, but we can change them or we could set them in the class directly.

test/test_functional.py
vincentqb: I'm reusing the parameters from `test_istft_of_ones`.

torchaudio/functional.py
vincentqb: Good idea, but that means I need to change all of the batching. I suggest to change that all at the same time explicitly as a separate PR.

PR #372, opened 2019-12-18
deactivate failing test

vincentqb:
Deactivate temporarily librosa consistency test which is currently failing on travis. Cannot reproduce offline. Blocking #355.

PR #392, opened 2020-01-02
[Bug Fix] fix power of spectrogram. makes power a float

### LINKED ISSUE #389: 'Power' in spectrogram
In `functional.spectrogram`, the docstring says
```
        power (int): Exponent for the magnitude spectrogram,
            (must be > 0) e.g., 1 for energy, 2 for power, etc.
            If None, then the complex spectrum is returned instead.
```
which is implemented as ([link](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L268))
```python
    if power is not None:
        spec_f = spec_f.pow(power).sum(-1)  # get power of "complex" tensor
```

If I'm not missing something, I think the right implementation is i) convert complex to magnitude ii) then apply the power on it, i.e.,
```python
    if power is not None:
        magnitude_spec = torch.sqrt(spec_f.pow(2).sum(-1))  # i)
        return magnitude_spec.pow(power)  # ii)
```

Or equivalently, using `complex_norm`..
```python
    if power is not None:
        return complex_norm(spec_f, power=power)
```

Any thought?

vincentqb:
Align with librosa:
* fix power of spectrogram: closes #389
* makes power a float, so as to work with complex norm
* partially re-enabling librosa consistency test to test with different power: closes #373.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -16,7 +16,7 @@ if IMPORT_LIBROSA:
 if IMPORT_SCIPY:
     import scipy
 
-SKIP_LIBROSA_CONSISTENCY_TEST = True
+SKIP_LIBROSA_CONSISTENCY_TEST = False
 RUN_CUDA = torch.cuda.is_available()
 print("Run test with cuda:", RUN_CUDA)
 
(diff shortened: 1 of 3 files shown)


test/test_transforms.py
```diff
@@ -308,9 +304,20 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
...
+            'power': 3.0,
+            'n_mels': 128,
+            'n_mfcc': 40,
+            'sample_rate': 16000
+        }
+
         _test_librosa_consistency_helper(**kwargs1)
         _test_librosa_consistency_helper(**kwargs2)
-        _test_librosa_consistency_helper(**kwargs3)
+        # XXX Test started failing
```
cpuhrsch: Should this be a TODO then or a NOTE with some information as to why it fails?

vincentqb: Good point, I will change to NOTE and refer to #372, which is where it started.

PR #394, opened 2020-01-06
Add Windows CI

peterjc123:


peterjc123:
cc @vincentqb Should be ready.

torchaudio/_backend.py
```diff
@@ -1,11 +1,12 @@
 from functools import wraps
 
+import platform
 import torch
 
 from . import _soundfile_backend, _sox_backend
 
 
-_audio_backend = "sox"
+_audio_backend = "soundfile" if platform.system() == "Windows" else "sox"
```
vincentqb: You recommend making soundfile the default in windows? If that is the case, I'm thinking we may want to make soundfile the default everywhere. Since we are doing the 0.4.0 release now, it may be a good idea to bring this in now so that we have window supported.

Thoughts?

peterjc123: > You recommend making soundfile the default in windows?

I made the change because I was not able to compile the sox backend on Windows. So actually, soundfile is the only available choice for Windows.

> If that is the case, I'm thinking we may want to make soundfile the default everywhere.

The major limitation of soundfile is that it cannot read the file with the MP3 format, which is possible in sox.

peterjc123:
> Have you tried running them at all? 
Test results: https://gist.github.com/peterjc123/4f5b43ce372905c82333f7693a65a15c

torchaudio/_backend.py
peterjc123: Another difficult I met is that there are no available pysoundfile packages hosted on Anaconda Cloud for Windows. Although the current one is marked as noarch, it is depending on a gcc package which is not available on Windows. So for conda users, the installation script will look a bit weird like this.
```
conda install -c pytorch torchaudio
pip install soundfile
```

PR #402, opened 2020-01-10
Upgrading to UserWarning so that the user gets the warning.

vincentqb:
DeprecationWarning are hidden by default. Upgrading to UserWarning.


albanD:
Maybe you could create a custom `TorchAudioWarning` type to make them easier for the user to filter them out without removing all the warnings?

PR #403, opened 2020-01-12
Implementation of RandomCrop for audio

haideraltahan:
Similar to `RandomCrop` in torchvision. This implementation is for audio given a tensor.
I would've used `RandomCrop` in torchvision, however, it only takes PIL Image instead of a Tensor.
More on it in #416


mthrok:
Closing the PR as
1. `torchvision`'s `RandomCrop` now accepts `torch.Tensor`
2. This needs formal specification, as many audio training has corresponding metadata, which also needs to be cropped in the same time steps.

PR #413, opened 2020-01-16
ci: Remove 2.7 tests

seemethere:
Removes for both Travis and CircleCI

Python 2.7 was EOL on January 1, 2020

The last torchaudio release to support Python 2.7 was 0.4.0

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

PR #415, opened 2020-01-16
remove python 2.7 from CI

vincentqb:
As mentioned in release notes, python 2 is no longer officially supported. Some pathways will need to be removed from `regenerate.py`.


vincentqb:
#413

PR #419, opened 2020-01-22
testing with sox only when sox is available

vincentqb:
Address RuntimeError in Windows CI tests that occur due to sox not being present, see [here](https://github.com/pytorch/audio/pull/394#issuecomment-571413640).

@peterjc123 -- please let me know if this indeed reduces the number of errors in windows.


peterjc123:
Sorry for late reply. The test result is listed as follows:
```
(base) audio\test>pytest . --maxfail 10000000000
================================================= test session starts =================================================
platform win32 -- Python 3.7.4, pytest-5.2.1, py-1.8.0, pluggy-0.13.0
plugins: arraydiff-0.3, doctestplus-0.4.0, openfiles-0.4.0, remotedata-0.3.2
collected 75 items / 1 errors / 74 selected

======================================================= ERRORS ========================================================
______________________________________ ERROR collecting test/test_functional.py _______________________________________
test_functional.py:35: in <module>
    class TestFunctional(unittest.TestCase):
test_functional.py:44: in TestFunctional
    waveform_train, sr_train = torchaudio.load(test_filepath)
..\torchaudio\__init__.py:86: in load
    filetype=filetype,
..\torchaudio\_soundfile_backend.py:86: in load
    filepath, frames=num_frames, start=offset, dtype="float32", always_2d=True
\Anaconda3\lib\site-packages\soundfile.py:257: in read
    subtype, endian, format, closefd) as f:
\Anaconda3\lib\site-packages\soundfile.py:629: in __init__
    self._file = self._open(file, mode_int, closefd)
\Anaconda3\lib\site-packages\soundfile.py:1184: in _open
    "Error opening {0!r}: ".format(self.name))
\Anaconda3\lib\site-packages\soundfile.py:1357: in _error_check
    raise RuntimeError(prefix + _ffi.string(err_str).decode('utf-8', 'replace'))
E   RuntimeError: Error opening '\\AppData\\Local\\Temp\\tmp1nhtbbu0\\assets\\steam-train-whistle-daniel_simon.mp3': File contains data in an unknown format.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
================================================== 1 error in 11.87s ==================================================
```

peterjc123:
Updated test results with PyTorch 1.3.1: https://gist.github.com/peterjc123/02f6cb8a00d8cdbc69b9adeb14a7c3ca

vincentqb:
Thanks, I've updated the PR to get rid of more RuntimeError when sox is absent.

cpuhrsch:
This seems like a regression in test coverage due to an OS specific need? Ideally we could detect the available backends at runtime.

vincentqb:
> This seems like a regression in test coverage due to an OS specific need?

All the tests that were running before this PR will still be running after. Currently, the CI on Windows is not active precisely because Windows only has SoundFile, and all the tests depending on SoX would fail.

> Ideally we could detect the available backends at runtime.

The current mechanic to determine which backend is available is in `torchaudio/_backend.py` and runs at import time. We can indeed determine available backends at runtime.

test/test_compliance_kaldi.py
```diff
@@ -7,6 +7,22 @@
...
+
+BACKENDS = torchaudio._backend._audio_backends
+
+
+@contextmanager
+def AudioBackendScope(new_backend):
+    previous_backend = torchaudio.get_audio_backend()
+    try:
+        unittest.skipIf(not new_backend, "No backend supporting this test.")
+        unittest.skipIf(new_backend not in BACKENDS, new_backend + " is not available.")
```
vincentqb: FIXME: Too many tests are skipped here.

test/test_dataloader.py
```diff
@@ -9,6 +9,21 @@
...
 
+BACKENDS = torchaudio._backend._audio_backends
+
+
+@contextmanager
+def AudioBackendScope(new_backend):
+    previous_backend = torchaudio.get_audio_backend()
+    try:
+        unittest.skipIf(not new_backend, "No backend supporting this test.")
+        unittest.skipIf(new_backend not in BACKENDS, new_backend + " is not available.")
```
vincentqb: TODO: Import duplicated code.

vincentqb:
@peterjc123 -- The windows build is failing, have you encountered this issue before? Could you also tell me if this new version of this PR reduced the number of failing tests in windows?

vincentqb:
Linking #466 for general awareness

test/_test.py
```diff
@@ -0,0 +1,39 @@
...
+def AudioBackendScope(new_backend):
+    previous_backend = torchaudio.get_audio_backend()
+    try:
+        torchaudio.set_audio_backend(new_backend)
+        yield
+    finally:
+        torchaudio.set_audio_backend(previous_backend)
+
+
+def get_backends_with_mp3(backends):
```
mthrok: The purpose of this function was not immediately clear to me at the first time reading it. Adding docstring would be helpful.

test/test.py
```diff
@@ -5,20 +5,7 @@
...
-        self.previous_backend = torchaudio.get_audio_backend()
-
-    def __enter__(self):
-        torchaudio.set_audio_backend(self.new_backend)
-        return self.new_backend
-
-    def __exit__(self, type, value, traceback):
-        backend = self.previous_backend
-        torchaudio.set_audio_backend(backend)
+from _test import AudioBackendScope, BACKENDS, BACKENDS_MP3
```
mthrok: The name of the module, `_test`, is not very descriptive.
Also, to me, prefixing with underscore does not give additional context, as test modules are never intended for external use. So, something like `backend_utils` sound more appropriate.

vincentqb: Good point. In fact, I just moved those functions in `common_utils.py` :)

test/_test.py
vincentqb: Thanks for pointing this out!

PR #421, opened 2020-01-22
add cuda test for complex norm

vincentqb:
Adding missing test to cover all 13 transforms with cuda test.

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -326,6 +326,10 @@ class Tester(unittest.TestCase):
 
         _test_script_module(transforms.Spectrogram, tensor, sample_rate, sample_rate_2)
 
+    def test_scriptmodule_ComplexNorm(self):
+        tensor = torch.rand((1, 2, 201, 2))
+        _test_script_module(transforms.ComplexNorm, tensor)
+
     def test_resample_size(self):
         input_path = os.path.join(self.test_dirpath, 'assets', 'sinewave.wav')
         waveform, sample_rate = torchaudio.load(input_path)


vincentqb:
> This test will just prove that the jitable transform works. 

The test also checks that all the transform is compatible with cuda, see [here](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L23). Since [this test](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L607) was already testing jitability, this particular new test only really guarantees gpu support.

> There is no guarantee for the result. Should we add some deterministic result there?

Good point, though the correctness of the transform is already checked through its functional, see [here](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L697).

zhangguanheng66:
> > This test will just prove that the jitable transform works.
> 
> The test also checks that all the transform is compatible with cuda, see [here](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L23). Since [this test](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L607) was already testing jitability, this particular new test only really guarantees gpu support.
> 
> > There is no guarantee for the result. Should we add some deterministic result there?
> 
> Good point, though the correctness of the transform is already checked through its functional, see [here](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L697).

Thanks for pointing the test. Correct me if I'm wrong. The jit mode may give you slightly different result from the python one.

vincentqb:
> Thanks for pointing the test. Correct me if I'm wrong. The jit mode may give you slightly different result from the python one.

Yup, that is correct. That's why `_test_script_module` checks that the results with python and with jit agree, see [here](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L31). Is that what you meant?

Issue #287, opened 2019-09-18
amplitude normalization in create_fb_matrix


jongwook:
## 🐛 Not necessarily a Bug

Not exactly a bug, but I think it makes more sense if the triangular Mel filterbanks are area-normalized, as opposed to the current behavior where they are height-normalized.

## To Reproduce / Expected behavior

```python
import matplotlib.pyplot as plt
import torchaudio
import librosa

plt.figure(figsize=(16, 8))
plt.subplot(2, 1, 1)
plt.plot(torchaudio.transforms.MelScale(n_stft=1025, sample_rate=16000).fb.detach().cpu().numpy())
plt.title('torchaudio Mel filterbanks')

plt.subplot(2, 1, 2)
plt.plot(librosa.filters.mel(sr=16000, n_fft=2048).transpose())
plt.title('librosa Mel filterbanks')
plt.show()
```

![image](https://user-images.githubusercontent.com/266841/65094873-abe6b200-d973-11e9-9464-fd652f3a5faa.png)

The former has an advantage of preserving the energy similar to the STFT, while the latter keeps the overall spectral envelope unchanged.

If this is an intended behavior, feel free to close this issue. Thanks!

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)?

I did `pip install torchaudio -f https://download.pytorch.org/whl/torch_stable.html`, following `README.md`

 - What does `torchaudio.__version__` print? (If applicable)

'0.3.0'

Please copy and paste the output from our [environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py).

```
Collecting environment information...
PyTorch version: 1.2.0
Is debug build: No
CUDA used to build PyTorch: 10.0.130

OS: Ubuntu 16.04.5 LTS
GCC version: (Ubuntu 5.4.0-6ubuntu1~16.04.11) 5.4.0 20160609
CMake version: version 3.11.1

Python version: 3.7
Is CUDA available: Yes
CUDA runtime version: 10.0.130
GPU models and configuration:
GPU 0: Tesla V100-SXM2-16GB
GPU 1: Tesla V100-SXM2-16GB
GPU 2: Tesla V100-SXM2-16GB
GPU 3: Tesla V100-SXM2-16GB
GPU 4: Tesla V100-SXM2-16GB
GPU 5: Tesla V100-SXM2-16GB
GPU 6: Tesla V100-SXM2-16GB
[...]

PR #426, opened 2020-02-06
Adding normalized=True options to MelScale and MelSpectrogram modules

jongwook:
This allows area normalization for Mel spectrogram (See #287 ), where the filterbank 

In the same issue I also mentioned adding the Slaney mel frequencies in addition to HTK ones, but as the Slaney scale is quite cumbersome (manually connect the linear and logarithmic pieces), I'm not sure if the maintainers thinks it's worth adding it at the moment. So adding the area normalization only for now. 

One caveat might be that the namining might be confusing since `normalized` is also used in the `Spectrogram` module, but I couldn't come up with better choice. Other sensible choices for naming include `normalization=None | "area"` or `area_normalize=False | True`, and I am happy to change the naming to either.

For the test, I've added one more case in the librosa consistency checks. CR appreciated!

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -211,7 +211,7 @@ class Tester(unittest.TestCase):
 
     @unittest.skipIf(not IMPORT_LIBROSA or not IMPORT_SCIPY, 'Librosa and scipy are not available')
     def test_librosa_consistency(self):
-        def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, sample_rate):
+        def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, sample_rate, mel_norm=None):
             input_path = os.path.join(self.test_dirpath, 'assets', 'sinewave.wav')
             sound, sample_rate = torchaudio.load(input_path)
             sound_librosa = sound.cpu().numpy().squeeze()  # (64000)
@@ -229,10 +229,10 @@ class Tester(unittest.TestCase):
             # test mel spectrogram
             melspect_transform = torchaudio.transforms.MelSpectrogram(
                 sample_rate=sample_rate, window_fn=torch.hann_window,
-                hop_length=hop_length, n_mels=n_mels, n_fft=n_fft)
+                hop_length=hop_length, n_mels=n_mels, n_fft=n_fft, normalized=mel_norm == "slaney")
             librosa_mel = librosa.feature.melspectrogram(y=sound_librosa, sr=sample_rate,
                                                          n_fft=n_fft, hop_length=hop_length, n_mels=n_mels,
-                                                         htk=True, norm=None)
+                                                         htk=True, norm=mel_norm)
             librosa_mel_tensor = torch.from_numpy(librosa_mel)
             torch_mel = melspect_transform(sound).squeeze().cpu()
 
@@ -251,7 +251,7 @@ class Tester(unittest.TestCase):
             self.assertTrue(torch.allclose(db_torch.type(db_librosa_tensor.dtype), db_librosa_tensor, atol=5e-3))
 
             # test MFCC
-            melkwargs = {'hop_length': hop_length, 'n_fft': n_fft}
+            melkwargs = {'hop_length': hop_length, 'n_fft': n_fft, 'normalized': mel_norm == "slaney"}
             mfcc_transform = torchaudio.transforms.MFCC(sample_rate=sample_rate,
                                                         n_mfcc=n_mfcc,
                                                         norm='ortho',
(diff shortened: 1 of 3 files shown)


jongwook:
Fix incoming!

cpuhrsch:
Is there value in pulling out this normalization approach and making it a separate transform?

jongwook:
IMHO it wouldn't be necessary; it's still Mel spectrogram (and MFCC) conceptually, and people typically use whichever is the default in the toolkit. Many papers don't bother to report the Mel scale or normalization.

Still, It'd be nicer to have numerical interoperability as discussed in #287, hence this PR!

test/test_transforms.py
```diff
@@ -229,10 +229,10 @@ def _test_librosa_consistency_helper(n_fft, hop_length, power, n_mels, n_mfcc, s
             # test mel spectrogram
             melspect_transform = torchaudio.transforms.MelSpectrogram(
                 sample_rate=sample_rate, window_fn=torch.hann_window,
-                hop_length=hop_length, n_mels=n_mels, n_fft=n_fft)
+                hop_length=hop_length, n_mels=n_mels, n_fft=n_fft, normalized=mel_norm == "slaney")
```
vincentqb: From [comment](https://github.com/pytorch/audio/issues/287#issuecomment-582363778), there seems to be more than one way to normalize, is that the case? If so, wouldn't it be cleaner to just say `normalized="slaney"`?

However, if we go this direction, I second @cpuhrsch, it may be valuable to have a `SlaneyScale` transform like the `MelScale` transform that torchaudio has instead. Thoughts?

jongwook: Sorry for the confusion; the comments ("htk" vs "slaney") was about the mel-hz conversion methods, while this PR addresses an orthogonal option where the height of the triangular filters can be adjusted to have the same area. [(figure)](https://user-images.githubusercontent.com/266841/65094873-abe6b200-d973-11e9-9464-fd652f3a5faa.png)

This (and librosa's API) is somewhat confusing because there are the Mel scale from Slaney and Slaney normalization, both referring to [the auditory toolbox](https://engineering.purdue.edu/~malcolm/interval/1998-010/AuditoryToolboxTechReport.pdf) by Malcolm Slaney.

So for this specific issue I'd suggest using `normalized` or `area_normalize` for the names.

----
**NOT related to the PR but regarding the Mel scale options**:

We can later consider supporting other Mel scales (which is good to have for better librosa compliance but not my top priority).

Even in that case, having `SlaneyScale` and `MelScale` would be quite confusing; both are Mel scales, and there is no "standard" Mel scale used in literature (which is quite unfortunate). Someone uses this:

image

which is neither Slaney nor HTK... but this scale seems to be a pretty niche one.

torchaudio/functional.py
```diff
@@ -406,6 +406,7 @@ def create_fb_matrix(n_freqs, f_min, f_max, n_mels, sample_rate):
         f_max (float): Maximum frequency (Hz)
         n_mels (int): Number of mel filterbanks
         sample_rate (int): Sample rate of the audio waveform
+        normalized (bool, optional): area-normalize the FB, so that each filter has equal area
```
vincentqb: nit: "If True, normalize each filter to have equal area. If False, normalize each filter to have equal height.  (Default: ``False``)"

Thoughts?

jongwook: I love it! But before commiting these changes, I just learned about the [upcoming chainge in librosa.filters.mel](https://github.com/librosa/librosa/issues/1030) where the area normalizatioin has two variants, so people may choose from:

* `norm=None`: same as torchaudio's default MelScale
* `norm="slaney"`: the very new default (since 0.7.2, in Jan 13, 2020); same as the area normalization in this PR, dividing the filter by the width between the adjacent mel frequencies. [While this makes the L1 norm of each filter slightly different](https://github.com/librosa/librosa/issues/1030), this was preferred historically because it is regression tested to be bit identical with the MATLAB counterpart.
* `norm=1`: area normalization that has been the default historically and behaves exactly the same as `norm="slaney"`, but will change behavior in future librosa (likely 0.8), where the components will exactly sum to 1 (i.e. L1-normalized)
* `norm=p`: similar to the above, using Lp-norm, but there's no practical reason to use this.

Do you think we should follow their convention instead?

vincentqb: That's good to know! Given how the implementation is just a small division at the end of the computation on the filterbanks, I was still considering something like `MelScale`, `SlaneyScale`. However, if we add a `MelPScale` with a parameter for p-norm, it seems like we are back to having to provide a parameter.

Let' just use their new convention.

jongwook: Sure, I'll modify the current PR to have librosa-0.8-style `norm=` options instead.

---

(Again, below not pertaining to this PR)

I still think the naming `SlaneyScale` on top of `MelScale` would be confusing as commented above, because:

* The Mel scale from the MATLAB auditory toolbox (referred to as "slaney") is also a Mel scale. It might give an impression that it is not, or less legitimate than the HTK Mel scale.
* What the current `MelScale` class represents is actually a Mel filter bank, because the term "Mel scale" simply refer to a series of Mel frequencies mapped from Hz (by a function that is roughly linear below 1kHz and roughly logarithmic above 1kHz). The `MelScale` class contains a set of triangular filters constructed from the Mel scale. So `MelFilterBank` would have been a better naming.
* With `MelScale` encapsulating both the scale AND the filterbank, it is not quite clear in the name `SlaneyScale` if "slaney" is referring to the scale or the normalization method, and we too seem to be confused throughout the comments on this PR.
* Nobody uses the term "slaney scale"; when googled with quotes, I see only three irrelevant results and none in Google Scholar.

Regardless of the future naming of `SlaneyScale` (or `mel_scale="htk"|"slaney"` as I'd prefer), it should only apply to the Mel scale, and it is orthogonal to the `norm` options this PR is addressing.

jongwook:
I'll re-open the PR once the [librosa 0.8+ style `norm` options](https://github.com/pytorch/audio/pull/426#discussion_r376263816) are ready.

vincentqb:
> I'll re-open the PR once the [librosa 0.8+ style `norm` options](https://github.com/pytorch/audio/pull/426#discussion_r376263816) are ready.

The change will be around standardizing `p=1` and `p=infinity`, right? This is a reasonable standard to me. I'd also be ok going ahead since we know this will happen "soon".

jongwook:
Yes! I didn't mean to block on them but rather getting the code ready on my side 😄

PR #435, opened 2020-02-14
batch resample transform

### LINKED ISSUE #431: Resample Batch of Waveforms
Hello,
I would like to use the `Resample` class for up/down sampling the learned features during the training. Exactly the same way as `F.interpolate`. However, the batch dimension is not accepted by the `Resample` class so that I need to reshape the tensor to have only the channel and temporal dimensions.

Any help on how to resample a batch of waveforms/multi-channel features without any need to reshape?

Thanks in advance.

vincentqb:
Fixes #431


vincentqb:
The errors are not related to this, but to windows images not being found. @peterjc123

Issue #434, opened 2020-02-14
Adding Speech Command Dataset


tomassosorio:
## 🚀 Feature
Adding to the possible datasets to use the Speech Command Dataset
https://arxiv.org/abs/1804.03209

## Motivation
As a beginner steps into audio manipulation using the Speech Command Dataset is a must since it is simple and allows us to use it as a classification problem or end-to-end Automatic Speech Recognition.

## Pitch
Allowing beginner to have a starting point to a dataset that can allow the training of interesting models is an easy way.

vincentqb:
Since this dataset is very recent, I am not inclined to consider adding it at this time. Do you have references for such successful model? Have you considered LibriSpeech or CommonVoice?

tomassosorio:
It is a dataset that would be an easy start for a beginner to start touching with ASR since it does not necessitate many moving parts and does not need that many parameters to a model to have good performances with.

Sure, but if you want to detect, a phone number or something similar this dataset would be much simpler than training with Librispeech or CommonVoice.

repo using this dataset: https://github.com/LearnedVector/Wav2Letter

vincentqb:
Nice example, and good to see that the dataset appeared on [kaggle](https://www.kaggle.com/c/tensorflow-speech-recognition-challenge/data) :) Alright, we would welcome this dataset. In general, we try to keep the available datasets to the most popular and requested. Would you be interested in opening a pull request for it?

Please keep in mind that we keep the structure of the code very close to the ones currently available, see for instance [librispeech](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/librispeech.py).

tomassosorio:
Yes, I am totally interested in doing the PR! Thanks!!

Sure, I will take into account that!

PR #437, opened 2020-02-19
Adding Speech Command Dataset

tomassosorio:
Following the feature improvement on [#434](https://github.com/pytorch/audio/issues/434) it was add the Speech Command dataset.


tomassosorio:
@vincentqb For what I am seeing the tested failed due to not being downloaded yet, however, I dought that it is expected to be downloaded at the testing time

Those Errors occurs in my machine with the other datasets. So what is the process now?

vincentqb:
Also, the dataset test fails:
```
_______________________ TestDatasets.test_speechcommands _______________________
self = <test_datasets.TestDatasets testMethod=test_speechcommands>
    def test_speechcommands(self):
        path = os.path.join(self.path, "speechcommands")
        data = SPEECHCOMMANDS(path)
>       data[0]
test/test_datasets.py:54: 
self = <torchaudio.datasets.speechcommands.SPEECHCOMMANDS object at 0x7f88d40fe9d0>
n = 0
    def __getitem__(self, n):
>       fileid = self._walker[n]
E       IndexError: list index out of range
torchaudio/datasets/speechcommands.py:89: IndexError
========================= 1 failed, 6 passed in 0.50s ==========================
```

torchaudio/datasets/speechcommands.py
```diff
@@ -0,0 +1,93 @@
...
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    download_url(url, root)
+                extract_archive(archive, self._path)
+
+        walker = glob.glob(os.path.join(self._path, "**/*.wav"), recursive=True)
+        self._walker = [os.path.relpath(sample_path, self._path) for sample_path in walker]
+
+        if subset in subsets:
```
vincentqb: I do not recommend doing the logic of intersection and difference here as there are many ways of doing it, and the user should be in control. Moreover, there are already mechanism to combine datasets, e.g. [ConcatDataset](https://pytorch.org/docs/stable/data.html#torch.utils.data.ConcatDataset).

In [librispeech](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/librispeech.py#L63), we let the user pick one, and only one, of interest to load. The user can then produce more than one dataset and combine however they want afterward.

torchaudio/datasets/speechcommands.py
```diff
@@ -0,0 +1,93 @@
...
+
+        self._path = os.path.join(root, folder_in_archive)
+
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    download_url(url, root)
+                extract_archive(archive, self._path)
+
+        walker = glob.glob(os.path.join(self._path, "**/*.wav"), recursive=True)
```
vincentqb: We already provide `torchaudio.datasets.utils.walk_files` for this function.

torchaudio/datasets/speechcommands.py
```diff
@@ -0,0 +1,93 @@
...
+
+        if subset in subsets:
+            if subset == "training":
+                subsets.remove("training")
+                for subset in subsets:
+                    dataset_path = os.path.join(self._path, subset + suffix_subset_file)
+
+                    with open(dataset_path, "r") as f:
+                        walker = [sample[0] for sample in unicode_csv_reader(f)]
+                        self._walker = list(set(self._walker).difference(walker))
```
vincentqb: Let's define the walker once, and avoid redefining it later. This will allow us later to consider using a generator construction for dataset.

torchaudio/datasets/speechcommands.py
tomassosorio: To know the label is needed to the path of the file and currently that function only return the file and not the path that's why I implemented that way

torchaudio/datasets/speechcommands.py
tomassosorio: The split is obtained by a csv provided in the download.
 All the files for train validation and test are on the same folders. So the only way to separate is to do something similar to that.

torchaudio/datasets/speechcommands.py
tomassosorio: Yes, I saw that function but I did not notice the `param prefix`, will change for that

torchaudio/datasets/speechcommands.py
tomassosorio: I removed all the logic, the user, with the label and file_id, can have all the control to split however he wants

vincentqb:
I've rebased the code, changed the tuple type, and moved the path logic in the item function. Let's see what the tests give :)

PR #439, opened 2020-02-20
LJ Speech dataset

### LINKED ISSUE #429: Adding LJ Speech dataset
## 🚀 Feature
Adding publicly available [LJ Speech](https://keithito.com/LJ-Speech-Dataset/) dataset, widely used for TTS tasks.

## Motivation

It will be convenient for researchers in speech synthesis (and speech processing in general) to get directly to model development and spend less time on building custom pipelines for loading datasets. Also LJ Speech is considered as a benchmark dataset and many papers in speech synthesis report their performance scores based on it.

## Additional context

I'd like to contribute by adding LJ Speech dataset to torchaudio. What repository owners thing of having it in the library?

taras-sereda:
This PR adds LJ Speech dataset to torch audio. 

Code structure is similar to [librispeech](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/librispeech.py) as was requested by @vincentqb . But there is one difference, `load_ljspeech_item` is implemented as a class method, not as a standalone function. There is a reason for this this: all transcripts for corresponding audio files are stored in a single *metadata.csv* file (which contains 13100 rows) and it will be prohibitive loading this files and searching for a corresponding transcript for each call of `load_ljspeech_item`. So a content of *metadata.csv* is stored in `self._metadata` dictionary of `LJSPEECH` object and corresponding transcript can be retrieved by querying `self._metadata`

Note: This code supports different versions of LJSpeech, currently it downloads the most recent version of dataset **LJSpeech-1.1**. If one needs to use previous/future dataset versions, then passing corresponding url at the time of creating `LJSPEECH` object should be sufficient. Provided that dataset structure is not changing by dataset maintainer.

Closes #429.


torchaudio/datasets/ljspeech.py
```diff
@@ -0,0 +1,79 @@
...
+                    download_url(url, root)
+                extract_archive(archive)
+
+        walker = walk_files(
+            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
+        )
+        self._walker = list(walker)
+        self._metadata = self._load_metadata()
+
+    def _load_metadata(self):
```
vincentqb: We can avoid creating a variable just for the metadata by following the example done in [CommonVoice](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/commonvoice.py#L107). Essentially, in this case, we iterate over the rows of the file to generate the walker. The load item function then just opens the audio file and returns the desired tuple.

torchaudio/datasets/ljspeech.py
```diff
@@ -0,0 +1,79 @@
...
+        fileid_audio = fileid + ext_audio
+        fileid_audio = os.path.join(path, fileid_audio)
+
+        # Load audio
+        waveform, sample_rate = torchaudio.load(fileid_audio)
+
+        # Load transcription
+        fileid_metadata = self._metadata.get(fileid, None)
+        if fileid_metadata is None:
+            raise ValueError("Transcription not found for " + fileid_audio)
```
vincentqb: This won't be necessary if iterating over the metadata file instead of the path.

taras-sereda: Yeah, @vincentqb thank you for pointing on [commonvoice](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/commonvoice.py) example, not code is more concise and follows structure used across other dataset.

taras-sereda:
btw, should I squash the commits?

taras-sereda:
great, PR was accepted pretty fast, it's a pleasure to contribute :)

PR #445, opened 2020-02-26
[WIP] Add pitchShift

### LINKED ISSUE #438: Pitch Shift
## 🚀 Feature
Shift the pitch of a waveform by n_steps semitones.

## Motivation

Common data augmentation used for training in ASR

tomassosorio:
- Add PitchShift according with: https://librosa.github.io/librosa/_modules/librosa/effects.html#pitch_shift
- Add test for PitchShift

Closes #438

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -100,6 +100,10 @@ ENV/
 # Rope project settings
 .ropeproject
 
+# Pycharm Project settings
+*.idea
+.DS_Store
+
 # mkdocs documentation
 /site
 
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -21,6 +21,7 @@ __all__ = [
     'Resample',
     'ComplexNorm',
     'TimeStretch',
+    'PitchShift',
     'FrequencyMasking',
     'TimeMasking',
 ]
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -556,13 +556,13 @@ class TimeStretch(torch.jit.ScriptModule):
 
 
 class PitchShift(torch.nn.Module):
-    r"""Shift the pitch of a waveform by `n_steps` semitones.
+    r"""Shift the pitch of a waveform by n_steps semitones.
 
     Args:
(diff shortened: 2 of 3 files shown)


tomassosorio:
It was not introduced the test_scriptmodule due to this issue: [21478](https://github.com/pytorch/pytorch/issues/21478), however, it is fixed but I do not know if the builds being made by travis already have this in consideration, so I decided not to add it.

.gitignore
```diff
@@ -100,6 +100,10 @@ ENV/
 # Rope project settings
 .ropeproject
 
+# Pycharm Project settings
```
vincentqb: This should be a separate PR if we add this.

torchaudio/transforms.py
```diff
@@ -640,6 +641,81 @@ def forward(self, complex_specgrams, overriding_rate=None):
...
+            torch.Tensor: Tensor of audio of dimension (..., time).
+        """
+        complex_specgrams = self.Spectrogram(waveform)
+        complex_specgrams_stretch = self.TimeStretch(complex_specgrams)
+        waveform_stretch = F.istft(complex_specgrams_stretch,
+                                   n_fft=self.n_fft,
+                                   hop_length=self.hop_length,
+                                   win_length=self.win_length)
+
+        waveform_shift = self.Resample(waveform_stretch)
```
vincentqb: I'm surprise we need a Resample here. Can we do without?

ksanjeevan:
Hey @tomassosorio, so with this implementation the `n_steps` are always fixed right? I'm using a similar implementation but with functionals so that it can be used in an augmentation pipeline, thoughts on having it like that here?

tomassosorio:
> Hey @tomassosorio, so with this implementation the `n_steps` are always fixed right? I'm using a similar implementation but with functionals so that it can be used in an augmentation pipeline, thoughts on having it like that here?

@ksanjeevan Yes, as it is it's fixed. Actually I am going to close this PR's since I am leaving this field of study, audio/speech, and I do not have time to help with this repo any more.

PR #448, opened 2020-02-27
Add test for InverseMelScale

mthrok:
This PR follows up #366 and adds test for `InverseMelScale` (and `MelScale`) for librosa compatibility.

For `MelScale` compatibility test;
1. Generate spectrogram
2. Feed the spectrogram to `torchaudio.transforms.MelScale` instance
3. Feed the spectrogram to `librosa.feature.melspectrogram` function.
4. Compare the result from 2 and 3 elementwise.
Element-wise numerical comparison is possible because under the hood their implementations use the same algorith.

For `InverseMelScale` compatibility test, it is more elaborated than that.
1. Generate the original spectrogram
2. Convert the original spectrogram to Mel scale using `torchaudio.transforms.MelScale` instance
3. Reconstruct spectrogram using torchaudio implementation
3.1. Feed the Mel spectrogram to `torchaudio.transforms.InverseMelScale` instance and get reconstructed spectrogram.
3.2. Compute the sum of element-wise P1 distance of the original spectrogram and that from 3.1.
4. Reconstruct spectrogram using librosa
4.1. Feed the Mel spectrogram to `librosa.feature.inverse.mel_to_stft` function and get reconstructed spectrogram.
4.2. Compute the sum of element-wise P1 distance of the original spectrogram and that from 4.1. (this is the reference.)
5. Check that resulting P1 distance are in a roughly same value range.

Element-wise numerical comparison is not possible due to the difference algorithms used to compute the inverse. The reconstructed spectrograms can have some values vary in magnitude.
Therefore the strategy here is to check that P1 distance (reconstruction loss) is not that different from the value obtained using `librosa`. For this purpose, threshold was empirically chosen

```
print('p1 dist (orig <-> ta):', torch.dist(spec_orig, spec_ta, p=1))
print('p1 dist (orig <-> lr):', torch.dist(spec_orig, spec_lr, p=1))
>>> p1 dist (orig <-> ta): tensor(1482.1917)
>>> p1 dist (orig <-> lr): tensor(1420.7103)
```

This value can vary based on the length and the kind of the signal being processed, so it was handpicked.

Closes #366


mthrok:
@vincentqb Once we settle on the test, we can do;
1. Squash merge #366 to master
2. Rebase this branch to master
3. Squash merge this.
That way the credit of @jaeyeun97 is preserved properly.

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        spec_orig = spec_orig.sqrt()
+        spec_ta = spec_ta.sqrt()
+
+        # For debug
+        # _plot_melspecs(sample_rate, spec_orig, spec_ta, spec_lr)
+
+        # The spectrograms reconstructed by librosa and torchaudio are not very comparable elementwise.
+        threshold = 2.5
+        # This is because they use different approximation algorithms and resulting values can live
+        # in different magniude.
```
vincentqb: nit: "magnitude" :)

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        # See https://github.com/pytorch/audio/pull/366 for the discussion of the choice of algorithm
+        # To get this number do the following and pick something bigger than that;
+        # print('p1 dist (  lr <-> ta):', torch.dist(spec_lr, spec_ta, p=float('inf')))
```
vincentqb: nit: "p-inf" not "p1"

in any case, just put: `torch.dist(spec_lr, spec_ta, p=float('inf'))` along with the resulting value, as you have done below

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        assert torch.allclose(spec_ta, spec_lr, atol=threshold)
+
+        threshold = 1500.0
+        # This threshold was choosen empirically, based on the following observations
+        # print('p1 dist (orig <-> ta):', torch.dist(spec_orig, spec_ta, p=1))
+        # print('p1 dist (orig <-> lr):', torch.dist(spec_orig, spec_lr, p=1))
+        # print('p1 dist (  lr <-> ta):', torch.dist(spec_lr, spec_ta, p=1))
```
vincentqb: same

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        # >>> p1 dist (orig <-> ta): tensor(1482.1917)
+        # >>> p1 dist (orig <-> lr): tensor(1420.7103)
+        # >>> p1 dist (  lr <-> ta): tensor(881.7889)
+        assert torch.dist(spec_orig, spec_ta, p=1) < threshold
+
+
+def _plot_melspecs(sample_rate, spec_original, spec_ta, spec_lr):
```
vincentqb: Thanks for providing the code, but can you put this in a comment on the PR instead of in here?

vincentqb: That being said, please feel free to show the graph in the PR comment :)

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+            melspec_lr, sr=sample_rate, n_fft=n_fft, power=2.0, htk=True, norm=None)
+        spec_lr = torch.from_numpy(spec_lr[None, ...])
+
+        # Align dimensions
+        # librosa does not return power spectrogram
+        spec_orig = spec_orig.sqrt()
+        spec_ta = spec_ta.sqrt()
+
+        # For debug
+        # _plot_melspecs(sample_rate, spec_orig, spec_ta, spec_lr)
```
vincentqb: can remove this

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+
+        # The spectrograms reconstructed by librosa and torchaudio are not very comparable elementwise.
+        threshold = 2.5
```
vincentqb: offline discussion with @mthrok: the error appears on the boundary mostly.

can you compute the error with and without the boundary? let's definitely add an assertion without the boundary to show that the error is very small.

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        # This is because they use different approximation algorithms and resulting values can live
+        # in different magniude.
+        # See https://github.com/pytorch/audio/pull/366 for the discussion of the choice of algorithm
+        # To get this number do the following and pick something bigger than that;
+        # print('p1 dist (  lr <-> ta):', torch.dist(spec_lr, spec_ta, p=float('inf')))
+        assert torch.allclose(spec_ta, spec_lr, atol=threshold)
+
+        threshold = 1500.0
```
vincentqb: thanks for providing the experiments below: this looks fine in that context.

test/test_transforms.py
```diff
@@ -509,5 +510,130 @@ def test_scriptmodule_TimeMasking(self):
...
+        return sound.cpu().numpy().squeeze()
+
+    def _get_sample_data(self, *asset_paths, **kwargs):
+        file_path = os.path.join(self.test_dirpath, 'assets', *asset_paths)
+
+        sound, sample_rate = torchaudio.load(file_path, **kwargs)
+        return sound.mean(dim=0, keepdim=True), sample_rate
+
+    @unittest.skipIf(not IMPORT_LIBROSA, 'Librosa is not available')
+    def test_MelScale(self):
```
vincentqb: thanks for adding this!

test/test_transforms.py
mthrok: As I look closer, what I thought was boundary effect was not really boundary effect.
I summarized the frequency-wise P-inf distance [here](https://gist.github.com/mthrok/b8749ae2d2c5e010e900b0823e2fa8b1).
Around [i==53, 54 and 68](https://gist.github.com/mthrok/b8749ae2d2c5e010e900b0823e2fa8b1#file-frequency-wise-p-inf-distance-txt-L53), are the distance largest.
I could hand-pick the threshold for these frequencies but it would make the test flaky so I ended up not doing that.

mthrok:
For the record; 
the following is what the reconstructed spectrograms look like in `test_InverseMelScale` test.
As mentioned in [my comment](https://github.com/pytorch/audio/pull/448/files#r385747021) and as in [the original PR](https://github.com/pytorch/audio/pull/366), these values are mostly close but in some places they can be extremely different.

test_InverseMelScale

<Details> <Summary>code for plot</Summary>

```
# based on https://gist.github.com/jaeyeun97/8651dff509d5b084636ac6c3a7547108
# thanks @jaeyeun97

def _plot_melspecs(sample_rate, spec_original, spec_ta, spec_lr):
    def log_mag(spec):
        ref = spec.max().clamp(1e-16).log10_()
        spec = spec.clamp(1e-16).log10().sub_(ref).mul_(20).clamp(min=-80)
        return spec.squeeze().numpy()

    spec_original = log_mag(spec_original)
    spec_ta = log_mag(spec_ta)
    spec_lr = log_mag(spec_lr)

    import librosa.display
    import matplotlib

    matplotlib.use('TkAgg')
    plt.figure(figsize=(20, 10))

    plt.subplot(3, 1, 1)
    librosa.display.specshow(spec_original, sr=sample_rate)
    plt.title('Original')

    plt.subplot(3, 1, 2)
    librosa.display.specshow(spec_lr, sr=sample_rate)
    plt.title('Librosa')

    plt.subplot(3, 1, 3)
    librosa.display.specshow(spec_ta, sr=sample_rate)
    plt.title('Torchaudio')
    plt.savefig(f'test_InverseMelScale.png')
```

</Details>

test/test_transforms.py
vincentqb: Alrighty :)

test/test_transforms.py
```diff
@@ -509,5 +510,97 @@ def test_scriptmodule_TimeMasking(self):
...
+        melspec_lr = melspec_ta.cpu().numpy().squeeze()
+        # Perform InverseMelScale with torch audio and librosa
+        spec_ta = transforms.InverseMelScale(
+            n_stft, n_mels=n_mels, sample_rate=sample_rate)(melspec_ta)
+        spec_lr = librosa.feature.inverse.mel_to_stft(
+            melspec_lr, sr=sample_rate, n_fft=n_fft, power=2.0, htk=True, norm=None)
+        spec_lr = torch.from_numpy(spec_lr[None, ...])
+
+        # Align dimensions
+        # librosa does not return power spectrogram
```
vincentqb: Do you mean that neither librosa nor torchaudio return the power spectrum?

torchaudio/transforms.py
```diff
@@ -233,6 +234,79 @@ def forward(self, specgram):
...
+
+        assert f_min <= self.f_max, 'Require f_min: %f < f_max: %f' % (f_min, self.f_max)
+
+        fb = F.create_fb_matrix(n_stft, self.f_min, self.f_max, self.n_mels, self.sample_rate)
+        self.register_buffer('fb', fb)
+
+    def forward(self, melspec):
+        r"""
+        Args:
+            melspec (torch.Tensor): A Mel frequency spectrogram of dimension (channel, ``n_mels``, time)
```
vincentqb: Could we support batching by doing `view` as in other transforms?

jaeyeun97: @vincentqb `channel` here is effectively the batch dimension?

jaeyeun97: Added support for arbitrary batch dimensions. https://github.com/pytorch/audio/pull/366/commits/590939e07327d13a854bcdccac77a93ebac1a95d

vincentqb: Can you also add a test as done [here](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L329) for this?

vincentqb: (just saw you added tests as done [here](https://github.com/pytorch/audio/blob/master/test/test_transforms.py#L329), thanks!)

test/test_transforms.py
mthrok: No, only librosa returns spectrogram. TorchAudio returns power spectrogram. I updated the comment to clarify that.

test/test_transforms.py
```diff
@@ -410,6 +410,25 @@ def test_batch_MelScale(self):
...
+        # Single then transform then batch
+        expected = transforms.InverseMelScale(n_stft, n_mels)(mel_spec).repeat(3, 1, 1, 1)
+
+        # Batch then transform
+        computed = transforms.InverseMelScale(n_stft, n_mels)(mel_spec.repeat(3, 1, 1, 1))
+
+        # shape = (3, 2, n_mels, 32)
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+
+        # Because InverseMelScale runs SGD on randomly initialized values so they do not yield
```
vincentqb: Does setting the random seed fix this, see [here](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L83)?

torchaudio/transforms.py
mthrok: Thanks @jaeyeun97. I incorporated your commit and made small changes for Python2 compatibility. (just in case)
Also added batch test.

vincentqb: Most transforms are reshaped using view to fold the batches into the channel dimensions, and then unfolded, see [here](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L216).

test/test_transforms.py
mthrok: It seems not. I ran multiple of run with different seed value but the P-inf distance was pretty much similar.

|seed| P-inf |
|-----|------|
|57011|0.8953|
|38810|0.9012|
|70091|0.9201|
|80441|0.9746|
|7957|0.9411|
|30379|0.9522|
|32164|0.9534|
|92637|0.9684|
|23494|0.9216|
|94651|0.9352|

torchaudio/transforms.py
mthrok: @vincentqb Will do so.

test/test_transforms.py
vincentqb: Interesting, how did you set the seed?

torchaudio/transforms.py
```diff
@@ -233,6 +234,88 @@ def forward(self, specgram):
...
+            diff = melspec - specgram.matmul(self.fb)
+            new_loss = diff.pow(2).sum(axis=2).mean()
+            new_loss.backward()
+            optim.step()
+            specgram.data = specgram.data.clamp(min=0)
+
+            new_loss = new_loss.item()
+            if new_loss < self.tolerance_loss or abs(loss - new_loss) < self.tolerance_change:
+                break
+            loss = new_loss
```
mthrok: @jaeyeun97 @vincentqb 
While I was applying batching with view, I noticed that this iteration loop has some issues.
from obvious one to less obvious one
1. `loss` variable (previous loss) was not updated so `tolerance_change` was not really working. 
2. checking `tolerance_loss` **AND** `tolerance_change` seems too strict. The break condition I often see is; 'either the target threshold is met or the value is not change as much'
3. During the loss computation, it took sum of all the elements. This means that the magnitude of loss value can vary based on the length of the input signal, the number of channels, the size of batch.
    So for example, even if you fix`tolerance_loss` as 1e-3, input signal length are same, but if you batch 10 signals, then on average, the threshold for one timeframe segment becomes 1e-4.
    I think tolerance values should be par timeframe. so I suggest changing loss computation to taking sum over frequency dimension and taking mean over the other dimensions. See 668f0f6 for the detail

test/test_transforms.py
mthrok: I looped the test, changing the seed every time.

```
import random
for _ in range(10):
    seed = random.randint(0, 100000)
    print(seed)
    torch.random.manual_seed(seed)
    # run test.
```

torchaudio/transforms.py
vincentqb: Good catch! I agree with your feedback.

torchaudio/transforms.py
```diff
@@ -233,6 +234,88 @@ def forward(self, specgram):
...
+        specgram = torch.rand(melspec.size()[0], time, freq, requires_grad=True,
+                              dtype=melspec.dtype, device=melspec.device)
+
+        optim = torch.optim.SGD([specgram], **self.sgdargs)
+
+        loss = float('inf')
+        for _ in range(self.max_iter):
+            optim.zero_grad()
+            diff = melspec - specgram.matmul(self.fb)
+            new_loss = diff.pow(2).sum(axis=2).mean()
```
vincentqb: nit: Let's use negative indices, so that would be `-1`, right? If we change the way we do batching, this is a more likely invariant.

test/test_transforms.py
vincentqb: For reference, there's a discussion about reproducibility [here](https://pytorch.org/docs/stable/notes/randomness.html). We don't need to worry about this here though. Thanks for the inline comment though :)

vincentqb:
As you noticed, test is failing. 
```
        # This threshold was choosen empirically, based on the following observations
        #
        # torch.dist(spec_orig, spec_ta, p=1)
        # >>> tensor(1482.1917)
        # torch.dist(spec_orig, spec_lr, p=1)
        # >>> tensor(1420.7103)
        # torch.dist(spec_lr, spec_ta, p=1)
        # >>> tensor(881.7889)
>       assert torch.dist(spec_orig, spec_ta, p=1) < threshold
E       AssertionError: assert tensor(1658.8206) < 1500.0
E        +  where tensor(1658.8206) = <built-in method dist of type object at 0x7fe5e2937880>(tensor([[[1.5051e+00, 7.9520e-01, 3.0737e-01,  ..., 6.0686e-01,\n          8.8888e-01, 2.8056e+00],\n         [7.7557e-0...e-04, 2.0129e-03],\n         [2.0623e-04, 4.6134e-05, 6.3226e-05,  ..., 6.4552e-05,\n          1.0691e-03, 2.0032e-03]]]), tensor([[[0.3844, 0.4208, 0.1461,  ..., 0.9873, 0.7092, 0.8993],\n         [0.7756, 1.1142, 0.9477,  ..., 0.8303, 1.985...0466, 0.0493, 0.0557,  ..., 0.0246, 0.0529, 0.0000],\n         [0.8467, 0.9353, 0.5868,  ..., 0.6988, 0.8108, 0.9290]]]), p=1)
E        +    where <built-in method dist of type object at 0x7fe5e2937880> = torch.dist
test/test_transforms.py:620: AssertionError
```

test/test_transforms.py
```diff
@@ -410,6 +410,25 @@ def test_batch_MelScale(self):
         self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
         self.assertTrue(torch.allclose(computed, expected))
 
+    def test_batch_InverseMelScale(self):
+        n_fft = 8
```
cpuhrsch: I'm wondering if it makes sense to vary some of these parameters a bit in a few subsequent tests. In particular to also include edge cases to see what the error behavior is. Unless this is verified through other tests.

test/test_transforms.py
```diff
@@ -509,5 +528,97 @@ def test_scriptmodule_TimeMasking(self):
...
+            sound, pad=0, window=torch.hann_window(n_fft), n_fft=n_fft,
+            hop_length=hop_length, win_length=n_fft, power=2, normalized=False)
+        spec_lr = spec_ta.cpu().numpy().squeeze()
+        # Perform MelScale with torchaudio and librosa
+        melspec_ta = transforms.MelScale(n_mels=n_mels, sample_rate=sample_rate)(spec_ta)
+        melspec_lr = librosa.feature.melspectrogram(
+            S=spec_lr, sr=sample_rate, n_fft=n_fft, hop_length=hop_length,
+            win_length=n_fft, center=True, window='hann', n_mels=n_mels, htk=True, norm=None)
+        # Note: Using relaxed rtol instead of atol
+        assert torch.allclose(melspec_ta, torch.from_numpy(melspec_lr[None, ...]), rtol=1e-3)
```
cpuhrsch: using self.assertTrue might yield a nicer message if this fails.

In the future and in a separate PR we might want to look into introducing some of the Unittest extensions that PyTorch implements that'll enable things such as self.assertAllClose and also does torch.Tensor specific checks such as dtype,memory layout etc. . allclose might do upcasting, broadcasting etc., but actually we care that those properties match. cc @vincentqb

torchaudio/transforms.py
cpuhrsch: Agreed.

test/test_transforms.py
mthrok: I agree with running test on multiple parameters.
However, I think that it can be accomplished better with reorganizing the whole test suite.

Right now Tester class contains all kinds of test, (like batch test, torch script test, librosa compatibility test etc...) and it was hard to tell what type of test I should add and where to add.
By creating separate test suite for different test, it will be much easier.
So I think, I would add that kind parameterized test later, and add similar things to the existing ones.

test/test_transforms.py
mthrok: > using self.assertTrue might yield a nicer message if this fails.

I found it opposite. Using `assertTrue` on `torch.allclose` only says 

```
>       self.assertTrue(torch.allclose(spec_ta, spec_lr, atol=threshold))
E       AssertionError: False is not true

test/test_transforms.py:618: AssertionError
```

whereas `assert` says
(although this is still hard to read due to combination of multiple line messages and pytest's annotation)
```
>       assert torch.allclose(spec_ta, spec_lr, atol=threshold)
E       AssertionError: assert False
E        +  where False = <built-in method allclose of type object at 0x121552eb0>(tensor([[[0.8752, 0.8655, 0.6858,  ..., 0.7232, 0.3609, 0.2115],\n         [0.7756, 1.1142, 0.9477,  ..., 0.8303, 1.985...0338, 0.0434, 0.0437,  ..., 0.0581, 0.0294, 0.0445],\n         [0.4310, 0.7263, 0.4167,  ..., 0.1131, 0.5628, 0.8183]]]), tensor([[[0.0000e+00, 0.0000e+00, 0.0000e+00,  ..., 0.0000e+00,\n          0.0000e+00, 0.0000e+00],\n         [7.7557e-0...e-04, 2.5366e-04],\n         [3.0709e-11, 5.1357e-12, 3.0357e-12,  ..., 0.0000e+00,\n          4.2634e-11, 1.0426e-10]]]), atol=1.0)
E        +    where <built-in method allclose of type object at 0x121552eb0> = torch.allclose

test/test_transforms.py:618: AssertionError
```

mthrok: Combined with your comment on parameterized test, I think reorganizing test structure and using PyTorch's helper functions to show a good example of how to write a test will be great benefit for all developers.

PR #449, opened 2020-02-28
Add fade

### LINKED ISSUE #447: Add Fade
#260 Working on add fade into transformations

tomassosorio:
Closes #447 

- Add test
- Add transform for Fade

Transform based on:
- https://pysox.readthedocs.io/en/latest/api.html#sox.transform.Transformer.fade
- https://manual.audacityteam.org/man/fade_and_crossfade.html

### DIFF
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -500,6 +500,20 @@ class Tester(unittest.TestCase):
         tensor = torch.rand((10, 2, n_freq, 10, 2))
         _test_script_module(transforms.TimeStretch, tensor, n_freq=n_freq, hop_length=hop_length, fixed_rate=fixed_rate)
 
+    def test_batch_Fade(self):
+        waveform, sample_rate = torchaudio.load(self.test_filepath)
+        fixed_fade_in_len = 3000
+        fixed_fade_out_len = 3000
+
+        # Single then transform then batch
+        expected = transforms.Fade(fixed_fade_in_len, fixed_fade_out_len)(waveform).repeat(3, 1, 1)
+
+        # Batch then transform
+        computed = transforms.Fade(fixed_fade_in_len, fixed_fade_out_len)(waveform.repeat(3, 1, 1))
+
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+        self.assertTrue(torch.allclose(computed, expected))
+
     def test_scriptmodule_FrequencyMasking(self):
         tensor = torch.rand((10, 2, 50, 10, 2))
         _test_script_module(transforms.FrequencyMasking, tensor, freq_mask_param=60, iid_masks=False)
--- a/test/test_transforms.py
+++ b/test/test_transforms.py
@@ -521,17 +521,24 @@ class Tester(unittest.TestCase):
 
     def test_batch_Fade(self):
         waveform, sample_rate = torchaudio.load(self.test_filepath)
-        fixed_fade_in_len = 3000
-        fixed_fade_out_len = 3000
+        fade_in_len = 3000
+        fade_out_len = 3000
 
         # Single then transform then batch
-        expected = transforms.Fade(fixed_fade_in_len, fixed_fade_out_len)(waveform).repeat(3, 1, 1)
+        expected = transforms.Fade(fade_in_len, fade_out_len)(waveform).repeat(3, 1, 1)
 
         # Batch then transform
-        computed = transforms.Fade(fixed_fade_in_len, fixed_fade_out_len)(waveform.repeat(3, 1, 1))
+        computed = transforms.Fade(fade_in_len, fade_out_len)(waveform.repeat(3, 1, 1))
 
         self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
         self.assertTrue(torch.allclose(computed, expected))
+
+    def test_scriptmodule_Fade(self):
+        tensor = torch.rand((1, 10000))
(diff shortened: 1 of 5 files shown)


tomassosorio:
@vincentqb Could you please take a look at this PR?

Because of test_scriptmodule I had to remove on-the-fly Fade alteration since I would get the following error: `Could not cast value of type Optional[int] to bool`

torchaudio/transforms.py
```diff
@@ -640,6 +641,83 @@ def forward(self, complex_specgrams, overriding_rate=None):
         return F.phase_vocoder(complex_specgrams, rate, self.phase_advance)
 
 
+class Fade(torch.nn.Module):
+    r"""Add a fade in and/or fade out to an waveform.
+
+    Args:
+        fade_in_len (int, optional): Length of fade-in (time frames). (Default: ``0``)
+        fade_out_len (int, optional): Length of fade-out (time frames). (Default: ``0``)
+        fade_shape (str, optional): Shape of fade. Must be one of:
```
vincentqb: Where does this notation come from?

test/test_transforms.py
```diff
@@ -519,6 +519,27 @@ def test_scriptmodule_TimeStretch(self):
...
+
+        self.assertTrue(computed.shape == expected.shape, (computed.shape, expected.shape))
+        self.assertTrue(torch.allclose(computed, expected))
+
+    def test_scriptmodule_Fade(self):
+        waveform, sample_rate = torchaudio.load(self.test_filepath)
+        fade_in_len = 3000
+        fade_out_len = 3000
+
+        _test_script_module(transforms.Fade, waveform, fade_in_len, fade_out_len)
```
vincentqb: Can you show the code snippet and error stack that fails as you mentioned above?

torchaudio/transforms.py
```diff
@@ -640,6 +641,83 @@ def forward(self, complex_specgrams, overriding_rate=None):
...
+        if self.fade_shape == "l":
+            fade = torch.log10(1.1 - fade) + 1
+
+        if self.fade_shape == "q":
+            fade = torch.sin(fade * math.pi / 2 + math.pi / 2)
+
+        if self.fade_shape == "h":
+            fade = torch.sin(fade * math.pi + math.pi / 2) / 2 + 0.5
+
+        return torch.cat((ones, fade)).clamp_(0, 1)
```
vincentqb: Is `clamp` necessary? Should we let the user decide what to do with this? Which step would you expect would make the signal go outside of `[0,1]`?

torchaudio/transforms.py
tomassosorio: The notions come from https://pysox.readthedocs.io/en/latest/api.html#sox.transform.Transformer.fade and looking into fade in http://sox.sourceforge.net/sox.html they are also coherent.

'p' for inverted parabola are not working properly using Sox so I opted to follow Audacity nomenclature using 'e' for exponential fade instead.

torchaudio/transforms.py
tomassosorio: To create fade-in and fade-out is calculated a percentage by which the audio or signal will be multiplied, this way simulating the fade. So, to guarantee that it will be between [0, 1] I added `clamp_` to enforce this.

test/test_transforms.py
tomassosorio: It was because to use `torch.jit.script` is necessary to:

> var = None
>  if var is None:

instead of:
> var = None
>  if var:

So it was my fault.

tomassosorio:
@vincentqb Tests were added, but effect sox is not correctly creating logarithmic and inverse parabola fades so I did not include this ones. I checked these by plotting both outputs.

torchaudio/transforms.py
vincentqb: We don't have to use the exact same notation. How about `quarter_sine`, `half_sine`, `linear`, `logarithmic`, and `exponential`?

vincentqb:
Could you provide the code snippet and plots here in the chats?

vincentqb:
The failures in the build are not related to this PR. :)

torchaudio/transforms.py
vincentqb: Updated.

tomassosorio:
```
import torchaudio
import torch
import math
from matplotlib import pyplot as plt


test_filepath = "/test/assets/steam-train-whistle-daniel_simon.mp3"
y, sr = torchaudio.load(test_filepath)

E = torchaudio.sox_effects.SoxEffectsChain()
E.set_input_file(test_filepath)

E.append_effect_to_chain("fade", [fade_type[1], 1, "0", 1])
x, sr = E.sox_build_flow_effects()

fade = Fade(44100, 44100, fade_type[0])

y_fade = fade(y)

fig, (ax1, ax2, ax3) = plt.subplots(3, 1, figsize=(20, 10))

ax1.set_title("Torch - Torchaudio")
ax1.plot(y_fade.numpy()[0], 'r');

ax2.set_title("Sox")
ax2.plot(x.numpy()[0], 'g');

ax3.set_title("Sox minus Torchaudio")
ax3.plot(x.numpy()[0] - y_fade.numpy()[0]);
```

## Linear
`fade_type = ("linear", "t")`
linear

## Quarter-Sine
`fade_type = ("quarter_sine", "q")`
quarter

## Half-sine
`fade_type = ("half_sine", "h")`
half

## Logarithmic
`fade_type = ("logarithmic", "l")`
log

## Exponential
`fade_type = ("exponential", "p")`
exponential

PR #456, opened 2020-03-05
Migrate TimeStretch and AmplitudeToDB to torch.nn.Module

tomassosorio:
Fixes #454

PR #457, opened 2020-03-05
phase_advance should be a buffer so it moves device correctly

ksanjeevan:
@vincentqb, fixing what was mentioned https://github.com/pytorch/audio/pull/456.

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -611,7 +611,7 @@ class TimeStretch(torch.nn.Module):
         n_fft = (n_freq - 1) * 2
         hop_length = hop_length if hop_length is not None else n_fft // 2
         self.register_buffer('phase_advance', torch.linspace(0, math.pi * hop_length, n_freq)[..., None])
-        
+
     def forward(self, complex_specgrams, overriding_rate=None):
         # type: (Tensor, Optional[float]) -> Tensor
         r"""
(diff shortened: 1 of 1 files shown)


vincentqb:
Can you show an error that would occur without that change? Some attributes/etc are inferred by `nn.Module`, see [here](https://pytorch.org/docs/stable/jit.html#module-attributes). If this is indeed needed here, it may be needed elsewhere in the transforms.

torchaudio/transforms.py
```diff
@@ -610,8 +610,8 @@ def __init__(self, hop_length=None, n_freq=201, fixed_rate=None):
 
         n_fft = (n_freq - 1) * 2
         hop_length = hop_length if hop_length is not None else n_fft // 2
-        self.phase_advance = torch.linspace(0, math.pi * hop_length, n_freq)[..., None]
-
+        self.register_buffer('phase_advance', torch.linspace(0, math.pi * hop_length, n_freq)[..., None])
+
```
vincentqb: nit: flake8 :)

ksanjeevan:
Yeah so here's an example:
```python
import torch
from torchaudio.transforms import TimeStretch
m = TimeStretch().cuda()
m(torch.randn(1, 2, 201, 400, 2).cuda(), 1.3)
```
Will give a `RuntimeError: expected device cuda:0 but got device cpu`. Also we're using buffers on the other transforms as far as I'm aware, see [here](https://github.com/pytorch/audio/blob/db1e7da9b309e69ab719155de9521d0c00c6e1c4/torchaudio/transforms.py#L58) for example.

Issue #446, opened 2020-02-26
Adding Module Models


tomassosorio:
## 🚀 Feature
Similar to [torchvision](https://pytorch.org/docs/stable/torchvision/models.html) would be interesting to have the architecture of the most important models available, such as Wav2Letter, DeepSpeech, DeepSpeech2 between others... 

## Motivation
When trying to train a model it is already available most features needed such as Datasets, Data Transformations, however it is not available any architecture of a known model.

## Pitch
Adding a module with the architecture of the most known models.

vincentqb:
This is a good idea. Would you like to contribute some?

tomassosorio:
Sure!

I will inspire into torchvision structure and add at first Wav2Letter.

At the time I cannot offer the model with the trained weights, but I can talk with my company to see if we could offer this to torchaudio in the future.

vincentqb:
Let's start with just the model class, untrained. Hosting trained models (and datasets) is a separate discussion :)

vincentqb:
@tomassosorio -- thanks again for merging wav2letter! Which other ones do you believe would be a good idea to request help to implement?

- [x] Wav2Letter
- [ ] DeepSpeech
- [ ] DeepSpeech2

tomassosorio:
Yes, I fell both of them would be good alternatives, such as the [Wav2Letter++](https://github.com/facebookresearch/wav2letter/tree/master/recipes/models), in this link there is also some other models that could be interesting to implement :)

koukyo1994:
How about [PANNs](https://github.com/qiuqiangkong/audioset_tagging_cnn)? It has rich collections of pre-trained audio models.

limazix:
Hi guys, I'd like to help with this task. Where can I start?

Edresson:
@vincentqb  I think it would be interesting to implement [Jasper](https://arxiv.org/pdf/1904.03288.pdf) too, as it reported state of the art on LibreSpeech :). 

It is available on [OpenSeq2Seq](https://github.com/NVIDIA/OpenSeq2Seq)

What do you think about converting the checkpoints of these models to pytorch and using them in the initialization of model weights?

vincentqb:
> @vincentqb I think it would be interesting to implement [Jasper](https://arxiv.org/pdf/1904.03288.pdf) too, as it reported state of the art on LibreSpeech :).
> 
> It is available on [OpenSeq2Seq](https://github.com/NVIDIA/OpenSeq2Seq)

[...]

PR #462, opened 2020-03-11
Add model Wav2Letter

tomassosorio:
- Add Documentation related to module models
- Add documentation to model wav2letter
- Add Unit-test to wav2-letter
- Add model Wav2Letter according to its paper [wav2letter](https://arxiv.org/pdf/1609.03193.pdf)
 #446


torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,79 @@
...
+            torch.Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
+        """
+
+        x = self.acoustic_model(x)
+        x = nn.functional.log_softmax(x, dim=1)
+        x = x.permute(2, 0, 1)
+        return x
+
+
+def wav2letter(**kwargs):
```
cpuhrsch: What's the advantage of having this separate factory function?

test/test_models.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
+import unittest
+
+import torch
+from torchaudio.models import wav2letter
+
+
+class ModelTester(unittest.TestCase):
+    def test_wav2letter(self):
```
cpuhrsch: Should we extend these tests further (i.e. specific input / output tests for fixed weights) or is there little value in it?

cpuhrsch:
As a generic reminder: Since this creates a whole new top-level folder and design (collection of models) I think it's worth pausing to make sure we're not setting us up for some hard-to-reverse decisions down the road.

test/test_models.py
tomassosorio: For what I saw in [torchvision](https://github.com/pytorch/vision/blob/master/test/test_models.py) they had a problem with doing extensive test due to slowing Travis.

The way of the test that I added is that the model should give a letter every 20ms, since the sample rate that they use is 16000, 20ms would be 320 points, however, because of the padding at the 10-th layer it would add 1 extra letter as output.

I used padding same with formula padding = ceil(kernel-stride)/2 due to wav2letter++ [issue](https://github.com/facebookresearch/wav2letter/issues/410#issuecomment-535054856)

torchaudio/models/wav2letter.py
tomassosorio: I used to mimic [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/models/alexnet.py), however, for now, is pointless yes... so I removed to minimize pointless functions.

vincentqb:
> As a generic reminder: Since this creates a whole new top-level folder and design (collection of models) I think it's worth pausing to make sure we're not setting us up for some hard-to-reverse decisions down the road.

Yes, we need to make sure we think the interface correctly. This provides a nice forcing function for us to do so. :)

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+        num_classes (int, optional): Number of classes to be classified. (Default: ``40``)
+        version (str, optional): Wav2Letter can use as input: ``waveform``, ```power_spectrum``
+         or ```mfcc``. (Default: ``waveform``)
+        n_input_features (int, optional): If is used ``power_spectrum`` or ```mfcc`` must be
+        specified the number of features that were extracted. (Default: ``None``)
+    """
+
+    def __init__(self, num_classes=40, version="waveform", n_input_features=None):
+        super(Wav2Letter, self).__init__()
+        n_input_features = 250 if not n_input_features else n_input_features
```
vincentqb, line 21: Can a user want to specify a number of feature different from 1 when using the waveform version?

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+        Args:
+            x (torch.Tensor): Tensor of dimension (batch_size, n_features, input_length).
+
+        Returns:
+            torch.Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
+        """
+
+        x = self.acoustic_model(x)
+        x = nn.functional.log_softmax(x, dim=1)
+        x = x.permute(2, 0, 1)
```
vincentqb, line 70: That's useful for CTC/ASG, but we should be careful around standardizing for this here. I'm tempted to leave the output shape by default for any models as `batch, class, length`, and let the user do what they want. DataParallel in particular breaks batches on the first dimension by default. This may not play as nicely with LSTM/etc though. Thoughts?

test/test_models.py
vincentqb: > I used padding same with formula padding = ceil(kernel-stride)/2 due to wav2letter++ [issue](https://github.com/facebookresearch/wav2letter/issues/410#issuecomment-535054856)

I feel like this should be mentioned in the documentation.

test/test_models.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import absolute_import, division, print_function, unicode_literals
```
vincentqb, line 1: nit: no more python 2 as you know :)

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+                nn.Conv1d(in_channels=1, out_channels=250, kernel_size=250, stride=160, padding=45),
+                nn.ReLU(inplace=True)
+            )
+            self.acoustic_model = nn.Sequential(waveform_model, acoustic_model)
+
+        if version in ["power_spectrum", "mfcc"]:
+            self.acoustic_model = acoustic_model
+
+    def forward(self, x):
+        # type: (Tensor) -> Tensor
```
vincentqb, line 59: nit: inline typing as you know :)

torchaudio/models/wav2letter.py
vincentqb: Since we don't currently offer pre-train model, I agree such factory functions are not needed yet.

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        num_classes (int, optional): Number of classes to be classified. (Default: ``40``)
+        version (str, optional): Wav2Letter can use as input: ``waveform``, ```power_spectrum``
+         or ```mfcc``. (Default: ``waveform``)
+        n_input_features (int, optional): If is used ``power_spectrum`` or ```mfcc`` must be
+        specified the number of features that were extracted. (Default: ``None``)
+    """
+
+    def __init__(self, num_classes=40, version="waveform", n_input_features=None):
```
vincentqb, line 19: nit: I'd use `num_features` as a name. thoughts?

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+            self.acoustic_model = acoustic_model
+
+    def forward(self, x):
+        # type: (Tensor) -> Tensor
+        r"""
+        Args:
+            x (torch.Tensor): Tensor of dimension (batch_size, n_features, input_length).
+
+        Returns:
+            torch.Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
```
vincentqb, line 65: If we were initializing the module lazily, we could infer the number of features. Maybe a factory function could help with that? I don't see it done with torchvision though, so I won't be advocating this for this PR. :)

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+class Wav2Letter(nn.Module):
+    r"""Wav2Letter model architecture from the `"Wav2Letter: an End-to-End ConvNet-based Speech Recognition System"
+     <https://arxiv.org/abs/1609.03193>`_ paper.
+
+    Args:
+        num_classes (int, optional): Number of classes to be classified. (Default: ``40``)
+        version (str, optional): Wav2Letter can use as input: ``waveform``, ```power_spectrum``
+         or ```mfcc``. (Default: ``waveform``)
+        n_input_features (int, optional): If is used ``power_spectrum`` or ```mfcc`` must be
+        specified the number of features that were extracted. (Default: ``None``)
```
vincentqb, line 16: nit: phrasing and number of quotes
```
If used with ``power_spectrum`` or ``mfcc`` version, ``num_features`` must specify the number of features extracted.
```

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class Wav2Letter(nn.Module):
+    r"""Wav2Letter model architecture from the `"Wav2Letter: an End-to-End ConvNet-based Speech Recognition System"
+     <https://arxiv.org/abs/1609.03193>`_ paper.
+
+    Args:
+        num_classes (int, optional): Number of classes to be classified. (Default: ``40``)
+        version (str, optional): Wav2Letter can use as input: ``waveform``, ```power_spectrum``
+         or ```mfcc``. (Default: ``waveform``)
```
vincentqb, line 14: nit: number of quotes? some are triple

vincentqb:
> One open point is to ensure correctness when someone suggests a new model. The tests added here are very limited, and testing that the implementation is right is not obvious.

Is there an invariant that could be verified? say the shape remains the same at each layers?

torchaudio/models/wav2letter.py
tomassosorio: It seems better than the previous name

vincentqb:
Test is failling:

```
_________________________ ModelTester.test_wav2letter __________________________
self = <test_models.ModelTester testMethod=test_wav2letter>
    def test_wav2letter(self):
        batch_size = 2
        n_features = 1
        input_length = 320
    
        model = Wav2Letter()
        x = torch.rand(batch_size, n_features, input_length)
>       out = model(x)
test/test_models.py:15: 
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/module.py:532: in __call__
    result = self.forward(*input, **kwargs)
torchaudio/models/wav2letter.py:71: in forward
    x = self.acoustic_model(x)
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/module.py:532: in __call__
    result = self.forward(*input, **kwargs)
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/container.py:100: in forward
    input = module(input)
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/module.py:532: in __call__
    result = self.forward(*input, **kwargs)
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/container.py:100: in forward
    input = module(input)
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/module.py:532: in __call__
    result = self.forward(*input, **kwargs)
self = Conv1d(1, 250, kernel_size=(48,), stride=(2,), padding=(23,))
input = tensor([[[0.0000, 0.0000],
         [0.0000, 0.0000],
         [0.0000, 0.0000],
         [0.0000, 0.0000],
         [....0000],
         [0.0000, 0.0973],
         [0.3981, 0.2518],
         [0.4468, 0.1599]]], grad_fn=<AsStridedBackward>)
    def forward(self, input):
        if self.padding_mode == 'circular':
            expanded_padding = ((self.padding[0] + 1) // 2, self.padding[0] // 2)
            return F.conv1d(F.pad(input, expanded_padding, mode='circular'),
                            self.weight, self.bias, self.stride,
                            _single(0), self.dilation, self.groups)
        return F.conv1d(input, self.weight, self.bias, self.stride,
>                       self.padding, self.dilation, self.groups)
E       RuntimeError: Given groups=1, weight of size 250 1 48, expected input[2, 250, 2] to have 1 channels, but got 250 channels instead
../../../miniconda3/envs/testenv/lib/python3.7/site-packages/torch/nn/modules/conv.py:202: RuntimeError
```

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+            self.acoustic_model = nn.Sequential(waveform_model, acoustic_model)
+
+        if version in ["power_spectrum", "mfcc"]:
+            self.acoustic_model = acoustic_model
+
+    def forward(self, x: Tensor) -> Tensor:
+        r"""
+        Args:
+            x (Tensor): Tensor of dimension (batch_size, n_features, input_length).
```
vincentqb, line 65: nit: `num_features`

torchaudio/models/wav2letter.py
```diff
@@ -0,0 +1,73 @@
...
+
+        Returns:
+            Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
```
vincentqb, line 68: This should be `(batch_size, number_of_classes, input_length)`, right?

vincentqb:
Overall, this looks very good to me, and serves as a great template for other models to come :)

tomassosorio:
> Overall, this looks very good to me, and serves as a great template for other models to come :)

Thanks! I will try to take a look today! Did not have the time yet :/

torchaudio/models/wav2letter.py
tomassosorio: Yes, it might want two use two channels for example

torchaudio/models/wav2letter.py
tomassosorio: Removed to allow more control for the user

tomassosorio:
@vincentqb Sorry for the delay, I was a bit busy.
If you could take a look I would appreciate :)

PR #463, opened 2020-03-12
Add test transform and functional AmplitudeToDB

tomassosorio:
- Compare results from `functional.AmplitudeToDB` with librosa corresponding functions
- Add batch test to `transform.AmplitudeToDB`
- Add to `test_librosa_consistency` test for magnitude


cpuhrsch:
@vincentqb - Merged based on your comment.

PR #468, opened 2020-03-20
Add Vol Transformation

tomassosorio:
Relates to #260 

- Add transform to mimic Sox Vol effect
- Add tests to compare with Sox


vincentqb:
Let's go with Vol as you suggest, since one of the cases for `gain_type` doesn't use `gain`.

PR #469, opened 2020-03-20
Add Functional DB_to_amplitude

tomassosorio:
- Add functional DB_to_amplitude
- Add tests to functional DB_to_amplitude

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -399,6 +399,21 @@ def amplitude_to_DB(x, multiplier, amin, db_multiplier, top_db=None):
     return x_db
 
 
+def DB_to_amplitude(x, ref, power):
+    # type: (Tensor, float) -> Tensor
+    r"""Turn a tensor from the decibel scale to the power/amplitude scale.
+
+    Args:
+        x (torch.Tensor): Input tensor before being converted to power/amplitude scale.
+        ref (float): Reference which the output will be scaled by.
+        power (float): If power equal to 1 will compute DB to power if 0.5 will compute DB to magnitude.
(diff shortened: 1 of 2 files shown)


torchaudio/functional.py
```diff
@@ -399,6 +399,21 @@ def amplitude_to_DB(x, multiplier, amin, db_multiplier, top_db=None):
...
+    # type: (Tensor, float, float) -> Tensor
+    r"""Turn a tensor from the decibel scale to the power/amplitude scale.
+
+    Args:
+        x (torch.Tensor): Input tensor before being converted to power/amplitude scale.
+        ref (float): Reference which the output will be scaled by. If computing DB to amplitude ref should be ref**2.
+        power (float): If power equal to 1 will compute DB to power, if 0.5 will compute DB to amplitude.
+
+    Returns:
+        torch.Tensor: Output tensor in power/amplitude scale.
```
vincentqb: The description of the arguments needs retouching:
* `If power equals 1, will compute DB to power. If 0.5, will compute DB to amplitude.`
* What is ref? "ref should be ref**2" seems circle. Would simply say "Scaling factor" be simpler?
* Should ref in code be exponentiated by `1/power` then? If so, then why not just taking `ref` out of `torch.pow`? This would avoid having to say this "ref should be ref**2". 
* Could an example of how to revert `amplitude_to_DB` help in the documentation?

tomassosorio: I did not add an example since I feel that with the implemented tests and the suggested improvements on the docstrings are enough, however, is your call :)

Issue #466, opened 2020-03-19
Refactor tests into small test suits


mthrok:
## 🚀 Refactor tests into small test suits.

As suggested/discussed in some [PR](https://github.com/pytorch/audio/pull/448#discussion_r385953906), I would like to clean up how unit tests are organized, parameterized and ran.

## Motivation

Functionalities in `torchaudio` have many common test patterns, such as, comparison against `librosa`, jit-ability and batch consistency. So if a function has nicely-organized test suite, adding the same type of function will be easier. (E.g. "I added this new SoX filter which is very similar to A, and there is a suite of test defined for A, so those are the tests I should be adding.") That way, it becomes easier for new contributor to write tests.

Let's take an example of `./test/test_functional.py` module. There are a lot of test methods under `TestFunctional` class.

The difficulty I had with this structure when I worked on https://github.com/pytorch/audio/pull/448 is that it was not immediately clear what kind of tests are typical to add.

<Details><Summary>The current organization of test_functional.py </Summary>

```
$ grep '\bclass\b\|def test' -r ./test/test_functional.py
class TestFunctional(unittest.TestCase):
    def test_torchscript_spectrogram(self):
    def test_torchscript_griffinlim(self):
    def test_griffinlim(self):
    def test_batch_griffinlim(self):
    def test_compute_deltas_onechannel(self):
    def test_compute_deltas_twochannel(self):
    def test_compute_deltas_randn(self):
    def test_batch_pitch(self):
    def test_jit_pitch(self):
    def test_istft_is_inverse_of_stft1(self):
    def test_istft_is_inverse_of_stft2(self):
    def test_istft_is_inverse_of_stft3(self):
    def test_istft_is_inverse_of_stft4(self):
    def test_istft_is_inverse_of_stft5(self):
    def test_istft_of_ones(self):
    def test_istft_of_zeros(self):
    def test_istft_requires_overlap_windows(self):
    def test_istft_requires_nola(self):
    def test_istft_requires_non_empty(self):
    def test_istft_of_sine(self):
    def test_linearity_of_istft1(self):
    def test_linearity_of_istft2(self):
    def test_linearity_of_istft3(self):
    def test_linearity_of_istft4(self):
    def test_batch_istft(self):
    def test_create_fb(self):
    def test_gain(self):
    def test_dither(self):
    def test_vctk_transform_pipeline(self):
    def test_pitch(self):
    def test_torchscript_create_fb_matrix(self):
[...]

PR #480, opened 2020-03-30
Extract JIT tests from test_functional to the dedicated test module

mthrok:
This PR is derived from #466. 
It extracts torchscript consistency test from `test_functional` module and put them in a dedicated module/class (`test_torchscript::TestFunctional`).


mthrok:
Yay

PR #481, opened 2020-03-30
Extract librosa tests from test_functional to the dedicated test module

mthrok:
From #466.
Similar to #480 
It extracts librosa consistency test from `test_functional` module and put them in a dedicated module/class (`test_librosa_consistency::TestFunctional`).


vincentqb:
Should librosa test in `test_transforms.py` also be moved?

Issue #477, opened 2020-03-30
Python 2 Deprecated


vincentqb:
The 0.4.0 release of torchaudio was the last one supporting python 2, and master no longer officially supports python 2. We're looking to strip the code of python 2 references.
- [x] No longer use package `six` and `backports` for cross-compatibility
- [x] Convert to inline type hinting
- [x] No `__future__` import
- [x] ~~Change string formatting style~~
- [ ] Remove mention of python 2.7 in `setup.py`
- [x] Remove older code path in [_check_module_exists](https://github.com/pytorch/audio/blob/master/torchaudio/common_utils.py#L26) and no longer need to check python 3 is not used [at the end of the file](https://github.com/pytorch/audio/blob/master/torchaudio/common_utils.py#L38)
- [x] Update `unicode_decoder` to python 3 only, [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/utils.py#L22).
- [x] Replace calls to [makedir_exist_ok](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/utils.py#L51) to `os.makedirs(.., exist_ok=True)`

tomassosorio:
I can take care of type hinting tomorrow :)

james-mchugh:
Do you need an extra hand with this or is everything covered?

vincentqb:
For completeness, I did a quick grep through files, and I see the following reference to python 2. 

```
setup.py:        "Programming Language :: Python :: 2.7",
test/test_librosa_compatibility.py:    # Using a decorator here causes parametrize to fail on Python 2
torchaudio/common_utils.py:    if not PY3:  # Python 2
torchaudio/common_utils.py:    elif not PY34:  # Python [3, 3.4)
torchaudio/common_utils.py:    else:  # Python >= 3.4
torchaudio/common_utils.py:# On Py2, importing librosa 0.6.1 triggers a TypeError (if using newest joblib)
torchaudio/common_utils.py:# TODO: allow Py2 when librosa 0.6.2 releases
torchaudio/datasets/utils.py:    r"""Since the standard csv library does not handle unicode in Python 2, we need a wrapper.
torchaudio/datasets/utils.py:    Python2 support for os.makedirs(.., exist_ok=True)
torchaudio/datasets/utils.py:    # Required for Python 2.7 compatibility
```

In here, also a reference to python less than 3.5 which is no longer supported on master, see [here](https://github.com/pytorch/audio/pull/538#issuecomment-613468667).

[...]

PR #482, opened 2020-03-31
Add inline typing to functional

tomassosorio:
relates to #477
- Use inline type hinting
- Fix Docstrings according with typing

During this fix two functions have a parameter that is not being used for anything: `griffinlim` with `normalized` and `istft` with `pad_mode`, probably they should be removed since they can mislead.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -140,11 +140,11 @@ def istft(
     device = stft_matrix.device
     fft_size = stft_matrix.size(1)
     assert (onesided and n_fft // 2 + 1 == fft_size) or (
-            not onesided and n_fft == fft_size
+        not onesided and n_fft == fft_size
     ), (
-            "one_sided implies that n_fft // 2 + 1 == fft_size and not one_sided implies n_fft == fft_size. "
-            + "Given values were onesided: %s, n_fft: %d, fft_size: %d"
-            % ("True" if onesided else False, n_fft, fft_size)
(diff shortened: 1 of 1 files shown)


vincentqb:
Thanks for catching up those parameters. We should remove them, but not in this PR since these changes are technically BC-breaking now.

vincentqb:
Thanks for working on this :)

vincentqb:
By the way, thanks for fixing some other lint (space after `assert`, typo, etc) though ideally the rest would have gone in a separate PR :) I looked into all of it and what's there is fine :)

tomassosorio:
Completely forgot to pass the flake8 :/

vincentqb:
> During this fix two functions have a parameter that is not being used for anything: `griffinlim` with `normalized` and `istft` with `pad_mode`, probably they should be removed since they can mislead.

@tomassosorio -- Do you want to open a PR to remove these? We should raise a `UserWarning` in this version if the values are changed from the default saying that these parameters are going away in the next. If these parameters meant something to convey that a specific algorithm was being used, we should also add that information in the documentation.

tomassosorio:
> > During this fix two functions have a parameter that is not being used for anything: `griffinlim` with `normalized` and `istft` with `pad_mode`, probably they should be removed since they can mislead.
> 
> @tomassosorio -- Do you want to open a PR to remove these? We should raise a `UserWarning` in this version if the values are changed from the default saying that these parameters are going away in the next. If these parameters meant something to convey that a specific algorithm was being used, we should also add that information in the documentation.

Sure @vincentqb ! I will take a look into that

PR #484, opened 2020-03-31
[BC breaking] fix issue with VCTK dataset

### LINKED ISSUE #483: VCTK missing txt file for 'p315'
## 🐛 Bug
The 'p315' text is missing from the VCTK Corpus. This leads to "FileNotFoundError" error when accessing './VCTK-Corpus/txt/p315'

> Please note while text files containing transcripts of the speech are provided for 109 of the 110 recordings, in the '/txt' folder, the 'p315' text was lost due to a hard disk error.

- https://datashare.is.ed.ac.uk/handle/10283/3443

## To Reproduce

```
!pip install torch>=1.2.0
!pip install torchaudio
!pip install librosa
%matplotlib inline

import torch
import torchaudio
import librosa
import numpy as np

import torchaudio.datasets as dsets
vctk_data = dsets.VCTK(".", download=True)
vctk_data[1]
```

## Expected behavior
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-25-a25a0369fbf9> in <module>()
----> 1 vctk_data[1]

1 frames
/usr/local/lib/python3.6/dist-packages/torchaudio/datasets/vctk.py in __getitem__(self, n)
    106             self._ext_txt,
    107             self._folder_audio,
--> 108             self._folder_txt,
    109         )
    110 

/usr/local/lib/python3.6/dist-packages/torchaudio/datasets/vctk.py in load_vctk_item(fileid, path, ext_audio, ext_txt, folder_audio, folder_txt, downsample)
     17     # Read text
     18     file_txt = os.path.join(path, folder_txt, speaker_id, fileid + ext_txt)
---> 19     with open(file_txt) as file_text:
     20         utterance = file_text.readlines()[0]
     21 

FileNotFoundError: [Errno 2] No such file or directory: './VCTK-Corpus/txt/p315/p315_041.txt'
```
## Environment

pip
Not applicable
1.4.0

```Collecting environment information...
PyTorch version: 1.4.0
Is debug build: No
CUDA used to build PyTorch: 10.1

OS: Ubuntu 18.04.3 LTS
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
CMake version: version 3.12.0

Python version: 3.6
Is CUDA available: No
CUDA runtime version: 10.1.243
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5

[pip3] numpy==1.18.2
[pip3] torch==1.4.0
[pip3] torchaudio==0.4.0
[pip3] torchsummary==1.5.1
[pip3] torchtext==0.3.1
[pip3] torchvision==0.5.0
[conda] Could not collect
```

tomassosorio:
Fixes #483

### DIFF
--- a/torchaudio/datasets/vctk.py
+++ b/torchaudio/datasets/vctk.py
@@ -97,9 +97,7 @@ class VCTK(Dataset):
         walker = walk_files(
             self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
         )
-
         walker = filter(lambda w: self._except_folder not in w, walker)
-
         self._walker = list(walker)
 
     def __getitem__(self, n):
--- a/torchaudio/datasets/vctk.py
+++ b/torchaudio/datasets/vctk.py
@@ -7,7 +7,6 @@ from torchaudio.datasets.utils import download_url, extract_archive, walk_files
 
 URL = "http://homepages.inf.ed.ac.uk/jyamagis/release/VCTK-Corpus.tar.gz"
 FOLDER_IN_ARCHIVE = "VCTK-Corpus"
-EXCEPT_FOLDER = "p315"
 
 
 def load_vctk_item(
@@ -49,6 +48,7 @@ class VCTK(Dataset):
     _folder_audio = "wav48"
     _ext_txt = ".txt"
     _ext_audio = ".wav"
+    _except_folder = "p315"
 
     def __init__(
         self,
@@ -98,7 +98,7 @@ class VCTK(Dataset):
             self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
         )
 
-        walker = filter(lambda w: EXCEPT_FOLDER not in w, walker)
+        walker = filter(lambda w: self._except_folder not in w, walker)
 
         self._walker = list(walker)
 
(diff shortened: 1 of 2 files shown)


torchaudio/datasets/vctk.py
```diff
@@ -91,7 +91,7 @@ def __init__(
             )
 
         walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
+            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True, ignore=["p315"]
```
vincentqb, line 94: We don't need to modify `walk_files`, since we can just use filter, as done [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/speechcommands.py#L67).

vincentqb, line 94: Let's make the list of files a module hidden variable.

vincentqb:
Added a note that this is BC-breaking since it changes the layout of the dataset.

vincentqb:
By the way, this issue is also arising with some of the other datasets. It would be a good idea to update the others too.

vincentqb:
The error appears to be due to not having rebased, which will happen automatically. I will go ahead and merge.

PR #490, opened 2020-04-01
Add inline typing to SoxEffects

tomassosorio:
- Add inline typing

relates to #477

Where are `Any` I do not know how to classify its types.

The `class SoxEffectsChain(object):` does not need anymore the `object` and the lines 52-55 of the docstrings I think are not right since is the method `sox_build_flow_effects`  that returns that, however, I left as it is.

### DIFF
--- a/torchaudio/sox_effects.py
+++ b/torchaudio/sox_effects.py
@@ -21,7 +21,7 @@ def effect_names() -> List[str]:
 
 
 @_audio_backend_guard("sox")
-def SoxEffect() -> Any:
+def SoxEffect():
     r"""Create an object for passing sox effect information between python and c++
 
     Returns:
--- a/torchaudio/sox_effects.py
+++ b/torchaudio/sox_effects.py
@@ -1,12 +1,13 @@
+from typing import Any, Callable, List, Optional, Tuple, Union
+
 import torch
-
 import torchaudio
-
+from torch import Tensor
 from torchaudio._backend import _audio_backend_guard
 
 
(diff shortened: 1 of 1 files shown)


torchaudio/sox_effects.py
```diff
@@ -77,9 +78,14 @@ class SoxEffectsChain(object):
 
     """
 
-    EFFECTS_UNIMPLEMENTED = set(["spectrogram", "splice", "noiseprof", "fir"])
+    EFFECTS_UNIMPLEMENTED = {"spectrogram", "splice", "noiseprof", "fir"}
```
vincentqb, line 81: Is there a reason you are changing this?

torchaudio/sox_effects.py
```diff
@@ -20,7 +21,7 @@ def effect_names():
 
 
 @_audio_backend_guard("sox")
-def SoxEffect():
+def SoxEffect() -> Any:
```
vincentqb, line 24: Instead of putting `Any` here, let's not deal with this case in this PR. We didn't have typing here, so it wouldn't be a regression to omit this one from this PR :)

torchaudio/sox_effects.py
tomassosorio: flake8 -> `torchaudio/sox_effects.py:81:29: C405 Unnecessary list literal - rewrite as a set literal.`

vincentqb: Fair enough :)

PR #491, opened 2020-04-01
Extract batch test from test_functional and move to the dedicated module

mthrok:
From #466.
This PR extracts batch consistency test from `test_functional` module and put them in a dedicated module/class (`test_batch_consistency::TestFunctional`).

### DIFF
--- a/test/test_batch_consistency.py
+++ b/test/test_batch_consistency.py
@@ -1,0 +1,104 @@
+"""Test numerical consistency among single input and batched input."""
+import os
+import unittest
+
+import torch
+import torchaudio
+import torchaudio.functional as F
+
+import common_utils
+
+
+def _test_batch_shape(functional, tensor, *args, **kwargs):
+
+    kwargs_compare = {}
+    if 'atol' in kwargs:
+        atol = kwargs['atol']
+        del kwargs['atol']
+        kwargs_compare['atol'] = atol
+
+    if 'rtol' in kwargs:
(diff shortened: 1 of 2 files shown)


test/test_functional.py
```diff
@@ -383,65 +350,6 @@ def test_pitch(self):
...
-
-        ind = [3] + [1] * (int(tensors.dim()) - 1)
-        tensors = tensor.repeat(*ind)
-
-        ind = [3] + [1] * (int(expected.dim()) - 1)
-        expected = expected.repeat(*ind)
-
-        torch.random.manual_seed(42)
-        computed = functional(tensors.clone(), *args, **kwargs)
-
```
mthrok, line 444: @vincentqb As I go through this, I noticed that unlike the presumed intention of this test 3-batch test is not performed, so I removed it. If you think we should add 3-Batch test then I can do so either in a same PR or a separate one.

vincentqb: Let's do this as a separate PR.

test/test_batch_consistency.py
```diff
@@ -0,0 +1,65 @@
...
+import unittest
+
+import torch
+import torchaudio
+import torchaudio.functional as F
+
+import common_utils
+
+
+# TODO: it is wrong to set manual seed in helper function. Move this to input function
```
vincentqb, line 12: Let's focus this PR on moving, and not changing the code yet.

PR #497, opened 2020-04-02
Inline typing Compliance Kaldi

tomassosorio:
- Add inline typing to compliance kaldi
- Minor fixes on docstrings
- Minor fixes for Flake8 rules

cc #477


tomassosorio:
due to github incident https://www.githubstatus.com/ it might have cause the builds to fail

PR #498, opened 2020-04-02
Fix common voice dataset

tomassosorio:
- CommonVoice Dataset was not downloading the dataset nether extracting to the correct folder according to other datasets structures
- Update dataset version to last one
- Add newly available languages


tomassosorio:
@vincentqb If you could take a look.

When I was doing the inline typing for dataset I tried to use this dataset but it wasn't downloading anything. I maintain the same structure as other datasets

torchaudio/datasets/commonvoice.py
```diff
@@ -77,29 +86,40 @@ def __init__(self, root, tsv=TSV, url=URL, download=False):
...
-                + ".s3.amazonaws.com/cv-corpus-3/"
-            )
-            url = base_url + language + ext_archive
+            base_url = "https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com"
+            url = os.path.join(base_url, version, language + ext_archive)
 
-        archive = os.path.basename(url)
-        archive = os.path.join(root, archive)
-        self._path = root
+        basename = os.path.basename(urlparse(url).path)
```
vincentqb, line 108: Is `urlparse` needed here?

vincentqb:
Added BC-breaking label since this changes how the data is being stored.

Issue #492, opened 2020-04-01
Add checksum to datasets


vincentqb:
All datasets that download an archive should have a checksum when downloading from the default address, see [comment](https://github.com/pytorch/audio/issues/483#issuecomment-607487378).

md5 would be enough for capturing obvious archive corruption.

- [x] VCTK: `45e8dede780278ef5541fde0b82ac292`
- [x] YesNo:
- [x] LibriSpeech:
- [x] SpeechCommands:
- [x] LJSpeech:
- [ ] CommonVoice:

tomassosorio:
speech_commands_v0.02 -> `6b74f3901214cb2c2934e98196829835`

PR #499, opened 2020-04-03
Add datasets checksum

bhargavkathivarapu:
Hey ,Regarding the issue ( #492 ) . 

I have added a new folder checksums to datasets folder.
I have modified all custom datasets classes to fetch corresponding checksum for url  from checksums folder.

Regarding checksums folder 
- One file for every dataset 
- File name as "dataset class name" + ".txt"
- Each file has url and hash_value separated by space for every url of that dataset

MD5 hash taken for  below datasets from  #492 
- VCTK
- Speech commands
- YesNo

SHA256 hash taken for below datasets from  [tensorflow datasets repo](https://github.com/tensorflow/datasets/tree/master/tensorflow_datasets/url_checksums)
- LIBRISPEECH
- LJSPEECH

Couldn't find hash values for COMMONVOICE online.
For now COMMONVOICE will not perform the checksum validation.But it will function as it is before.

Tested if downloads are starting  for all datasets . All downloads are starting. 
For YesNo dataset i have completed downloaded  and Verified if checksum is working by altering a checksum character.

@vincentqb  could you review these changes


torchaudio/datasets/utils.py
```diff
@@ -264,6 +264,25 @@ def walk_files(root, suffix, prefix=False, remove_suffix=False):
...
+    with open(checksum_file) as f:
+        checksum_dict = {k: v for line in f for k, v in [line.strip().split()]}
+        try:
+            checksum = checksum_dict[url]
+        except KeyError:
+            checksum = None
+            pass
+        return checksum
+
+
```
vincentqb, line 285: Reading such files could be useful, but this particular function wouldn't be generic enough. This also means that new files need to be created to host the checksums.

Since the urls are already hardcoded in each module file, I suggest instead to simply add a hidden dictionary directly in each module along the URLs, e.g.
```python
_CHECKSUMS = {
"http://www.openslr.org/resources/12/dev-clean.tar.gz": "76f87d090650617fca0cac8f88b9416e0ebf80350acb97b343a85fa903728ab3",
"http://www.openslr.org/resources/12/dev-other.tar.gz":  "12661c48e8c3fe1de2c1caa4c3e135193bfb1811584f11f569dd12645aa84365",
...
}
```
and then we can get the checksum by doing
```
                    checksum = _CHECKSUMS.get(url, None)
```

Thoughts?

bhargavkathivarapu: ok will make the suggested changes. I thought it might clutter the module for datasets like commonvoice where we have many urls ,so went for file  approach.

bhargavkathivarapu:
Changes completed . @vincentqb could you review these changes

vincentqb:
Thanks for working on this! Overall, the code looks good to me. I see the checksums for VCTK, SpeechCommands, and YesNo match what's listed in #492.

I'll merge this PR after the upcoming release so that we have some time to double check the checksums.

cpuhrsch:
Given that we have URLs and Checksums for datasets it might make sense, at some point, to introduce a separate .yaml or .json config file to store this information instead of including it into the implementation.

vincentqb:
> Given that we have URLs and Checksums for datasets it might make sense, at some point, to introduce a separate .yaml or .json config file to store this information instead of including it into the implementation.

As discussed [above](https://github.com/pytorch/audio/pull/499#discussion_r403075713), since most of the datasets only have 1-2 URLs supported, we decided to keep the checksums in the implementation.

PR #511, opened 2020-04-03
Datasets inline typing

tomassosorio:
Add inline typing to Datasets
relates to: #477


torchaudio/datasets/commonvoice.py
```diff
@@ -125,9 +130,9 @@ def __init__(self, root,
             self._header = next(walker)
             self._walker = list(walker)
 
-    def __getitem__(self, n):
+    def __getitem__(self, n: int):
```
vincentqb, line 133: nit: no return type

PR #513, opened 2020-04-06
Use torch.testing.assert_allclose

mthrok:
Currently there are three different assertions used for numerical comparison.

```python
assert torch.allclose(...

self.assertTrue(torch.allclose(...

torch.assert_allclose(...
```

`torch.assert_allclose` has nice features compared to the others;
 - More readable error
    `AssertionError: Not within tolerance rtol=0.0001 atol=1e-05 at input[0, 1] (-0.8857436776161194 vs. 1.0887877941131592) and 2 other locations (100.00%)`
 - Will raise an error if the shapes are different.
    `RuntimeError: expand(torch.FloatTensor{[1, 1, 3]}, size=[1, 3]): the number of sizes provided (2) must be greater or equal to the number of dimensions in the tensor (3)`

Example:

```python
import unittest

import torch


t1 = torch.randn(1, 3)
t2 = torch.randn(1, 3)


class TestNotClose(unittest.TestCase):
    def test_0(self):
        self.assertTrue(torch.allclose(t1, t2))

    def test_1(self):
        assert torch.allclose(t1, t2)

    def test_2(self):
        torch.testing.assert_allclose(t1, t2)


t3 = t1.view(1, 1, 3)


class TestWrongShape(unittest.TestCase):
    def test_0(self):
        self.assertTrue(torch.allclose(t1, t3))  # This passes
	assert t1.shape == t3.shape

    def test_1(self):
        assert torch.allclose(t1, t3)  # This passes
        assert t1.shape == t3.shape

    def test_2(self):
        torch.testing.assert_allclose(t1, t3)
```

```shell
$ python -m unittest foo.py

FFFFFE
ERROR: test_2 (foo.TestWrongShape)
  File "/scratch/moto/torchaudio/foo.py", line 34, in test_2
    torch.testing.assert_allclose(t1, t3)
  File "/scratch/moto/pytorch/torch/testing/__init__.py", line 24, in assert_allclose
    expected = expected.expand_as(actual)
RuntimeError: expand(torch.FloatTensor{[1, 1, 3]}, size=[1, 3]): the number of sizes provided (2) must be greater or equal to the number of dimensions in the tensor (3)

FAIL: test_0 (foo.TestNotClose)
  File "/scratch/moto/torchaudio/foo.py", line 12, in test_0
    self.assertTrue(torch.allclose(t1, t2))
AssertionError: False is not true

FAIL: test_1 (foo.TestNotClose)
  File "/scratch/moto/torchaudio/foo.py", line 15, in test_1
    assert torch.allclose(t1, t2)
AssertionError

FAIL: test_2 (foo.TestNotClose)
  File "/scratch/moto/torchaudio/foo.py", line 18, in test_2
    torch.testing.assert_allclose(t1, t2)
  File "/scratch/moto/pytorch/torch/testing/__init__.py", line 57, in assert_allclose
    raise AssertionError(msg.format(
AssertionError: Not within tolerance rtol=0.0001 atol=1e-05 at input[0, 1] (-0.8857436776161194 vs. 1.0887877941131592) and 2 other locations (100.00%)

FAIL: test_0 (foo.TestWrongShape)
  File "/scratch/moto/torchaudio/foo.py", line 27, in test_0
    assert t1.shape == t3.shape
AssertionError

FAIL: test_1 (foo.TestWrongShape)
  File "/scratch/moto/torchaudio/foo.py", line 31, in test_1
    assert t1.shape == t3.shape
AssertionError

Ran 6 tests in 0.007s

FAILED (failures=5, errors=1)
```

### DIFF
--- a/test/test_batch_consistency.py
+++ b/test/test_batch_consistency.py
@@ -23,8 +23,7 @@ def _test_batch_shape(functional, tensor, *args, atol=1e-8, rtol=1e-5, **kwargs)
     torch.random.manual_seed(42)
     computed = functional(tensors.clone(), *args, **kwargs)
 
-    assert expected.shape == computed.shape, (expected.shape, computed.shape)
-    torch.testing.assert_allclose(expected, computed, rtol=rtol, atol=atol)
+    torch.testing.assert_allclose(computed, expected, rtol=rtol, atol=atol)
 
     return tensors, expected
 
@@ -43,8 +42,7 @@ def _test_batch(functional, tensor, *args, atol=1e-8, rtol=1e-5, **kwargs):
     torch.random.manual_seed(42)
     computed = functional(tensors.clone(), *args, **kwargs)
 
-    assert expected.shape == computed.shape, (expected.shape, computed.shape)
-    torch.testing.assert_allclose(expected, computed, rtol=rtol, atol=atol)
+    torch.testing.assert_allclose(computed, expected, rtol=rtol, atol=atol)
 
 
 class TestFunctional(unittest.TestCase):
@@ -96,7 +94,6 @@ class TestTransforms(unittest.TestCase):
         # Batch then transform
         computed = torchaudio.transforms.AmplitudeToDB()(spec.repeat(3, 1, 1))
 
-        assert computed.shape == expected.shape, (computed.shape, expected.shape)
         torch.testing.assert_allclose(computed, expected)
 
     def test_batch_Resample(self):
@@ -108,7 +105,6 @@ class TestTransforms(unittest.TestCase):
         # Batch then transform
         computed = torchaudio.transforms.Resample()(waveform.repeat(3, 1, 1))
 
-        assert computed.shape == expected.shape, (computed.shape, expected.shape)
         torch.testing.assert_allclose(computed, expected)
 
     def test_batch_MelScale(self):
@@ -121,7 +117,6 @@ class TestTransforms(unittest.TestCase):
         computed = torchaudio.transforms.MelScale()(specgram.repeat(3, 1, 1, 1))
 
         # shape = (3, 2, 201, 1394)
-        assert computed.shape == expected.shape, (computed.shape, expected.shape)
         torch.testing.assert_allclose(computed, expected)
 
     def test_batch_InverseMelScale(self):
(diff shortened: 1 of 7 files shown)


test/test_librosa_compatibility.py
```diff
@@ -91,18 +92,18 @@ def test_amplitude_to_DB(self):
 
         ta_out = F.amplitude_to_DB(spec, multiplier, amin, db_multiplier, top_db)
         lr_out = librosa.core.power_to_db(spec.numpy())
-        lr_out = torch.from_numpy(lr_out).unsqueeze(0)
```
vincentqb, line 94: does this have to do with the assert change?

test/test_batch_consistency.py
```diff
@@ -84,7 +84,8 @@ def test_griffinlim(self):
         length = 1000
         tensor = torch.rand((1, 201, 6))
         _test_batch(
-            F.griffinlim, tensor, window, n_fft, hop, ws, power, normalize, n_iter, momentum, length, 0, atol=5e-5
+            F.griffinlim, tensor, window, n_fft, hop, ws, power, normalize, n_iter, momentum, length, 0,
+            atol=5e-5, rtol=1e-5,
```
vincentqb, line 88: why is the tolerance changed?

test/test_librosa_compatibility.py
mthrok: Yes, the shapes are not matching in the original test. Now with `assert_allclose` that is causing the issue so it has to be removed.

test/test_batch_consistency.py
mthrok: To use `assert_allclose`, you have to either 1. give both `atol` and `rtol` or 2. give neither of them. So I filled the values which was used originally.

mthrok:
Note: the order of the argument is `assert_allclose(COMPUTED, EXPECTED)`.

test/test_librosa_compatibility.py
vincentqb: sounds good!

vincentqb: sounds good!

test/test_batch_consistency.py
vincentqb: sounds good!

cpuhrsch:
On the tolerances: pytorch defines [default values for various dtypes](https://github.com/pytorch/pytorch/blob/591b5da2c83cf2cf29b80be985c889bfac422441/torch/testing/__init__.py#L113).

Also as a general (probably obvious) reminder: I wouldn't use ```allclose``` indiscriminately. For some tests it makes sense to check for bitwise correctness, etc.

vincentqb:
> Also as a general (probably obvious) reminder: I wouldn't use `allclose` indiscriminately. For some tests it makes sense to check for bitwise correctness, etc.

@cpuhrsch -- I agree with the importance of the distinction between the two. All the tests replaced here though were `torch.allclose` already. Did you mean something else?

cpuhrsch:
Sure, within the context of standardization this seems fine.

Another aspect when it comes to these assertions is whether they're checking if the dtypes, device and layout of the passed arguments match and whether that is for importance. I don't know if that differs across these three checks.

Issue #515, opened 2020-04-06
Making test code device aware.


mthrok:
Currently, test code does not have a standard way to handle device, which led to CUDA tests added disorganized ways, such as [1](https://github.com/pytorch/audio/blob/5f5df1d649de8504c68ec532195ab1ba3fb54680/test/test_functional_filtering.py#L36-L41), [2](https://github.com/pytorch/audio/blob/5f5df1d649de8504c68ec532195ab1ba3fb54680/test/test_functional_filtering.py#L90-L99) and [3](https://github.com/pytorch/audio/blob/5f5df1d649de8504c68ec532195ab1ba3fb54680/test/test_torchscript_consistency.py#L311-L331). 
Also it is worth noting that **Travis CI is not running any of CUDA test**. [See here, search for `test_lfilter_cuda`](https://travis-ci.org/github/pytorch/audio/jobs/671779349).

We should have a standard way to add CUDA test.

FYI: PyTorch has [a mechanism to instantiate test class for different devices at runtime](https://github.com/pytorch/pytorch/blob/4ef383d5db38c5a837093089fe50770d132cc47c/test/test_torch.py#L5298).

mthrok:
I also noticed that in [1](https://github.com/pytorch/audio/blob/master/test/test_functional_filtering.py#L36-L41), the test result looks like it passes when it was actually skipped.

vincentqb:
* I would be in favor of deprecating Travis CI in favor of CircleCI only, e.g. #493.
* We should also add GPU testing in CircleCI.

mthrok:
The choice of CI is adjacent and we should discuss down the road. (though I have no objection to the migration to CircleCI)

The followings are what I think needs to be sorted out for the next steps.

(here I consider only CPU and CUDA for device option.)

1. Add CUDA test
   1.1 Define a standard way to add device name to tests (or add CUDA test).
     - The bottom line is that device name (if not on CPU) should be visible in test result. so that we can see which tests PASSED/FAILED/ERROR/SKIPPED on which device.
     - CUDA test will be automatically skipped if CUDA device is not found, but also we should have option to force CUDA test and fail if CUDA is not available so that we can catch misconfiguration on CI.
   
   1.2 Find an efficient way to actually add device name to tests.
     - We can do this manually, but it can get tedious.
     - PyTorch parameterize device and instantiate test for different devices (CPU, CUDA, XLA ROCm).

2. Identify which test we should run on CUDA.
   - `torchaudio.functional`
[...]

PR #520, opened 2020-04-07
Separate CPU and GPU tests for Transforms torchscript test

mthrok:
Currently torchscript tests for Transforms module runs GPU tests optionally when GPU is available.
This makes it hard for understand what test is ran/skipped/success/fail.

This PR separates CPU test and GPU test and make them explicit in test report.

Relates to #515 

[Screen Shot 2020-04-07 at 13 25 46](https://travis-ci.org/github/pytorch/audio/jobs/672182174)

### DIFF
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -5,9 +5,18 @@ import unittest
 import torch
 import torchaudio
 import torchaudio.functional as F
-import torchaudio.transforms
+import torchaudio.transforms as T
 
 import common_utils
+
+
+def _assert_transforms_consistency(transform, tensor, device):
+    tensor = tensor.to(device)
+    transform = transform.to(device)
+    ts_transform = torch.jit.script(transform)
+    output = transform(tensor)
+    ts_output = ts_transform(tensor)
+    torch.testing.assert_allclose(ts_output, output)
 
 
 def _assert_functional_consistency(py_method, *args, shape_only=False, **kwargs):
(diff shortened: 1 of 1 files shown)


test/test_torchscript_consistency.py
```diff
@@ -304,85 +304,75 @@ def test_perf_biquad_filtering(self):
...
-
-        py_method = py_method.cuda()
-        jit_method = torch.jit.script(py_method)
+class _TransformsTestMixin:
+    """Implements test for Transforms that are performed for different devices"""
+    device = None
 
-        py_out = py_method(tensor)
-        jit_out = jit_method(tensor)
+    def _assert_conssitency(self, transform, tensor):
```
vincentqb, line 321: nit: typo "consistency"

test/test_torchscript_consistency.py
```diff
@@ -304,85 +304,75 @@ def test_perf_biquad_filtering(self):
...
+        _assert_transforms_consistency(transform, tensor, self.device)
 
-        torch.testing.assert_allclose(jit_out, py_out)
-
-
-class TestTransforms(unittest.TestCase):
     def test_Spectrogram(self):
         tensor = torch.rand((1, 1000))
-        _test_script_module(torchaudio.transforms.Spectrogram, tensor)
+        self._assert_conssitency(T.Spectrogram(), tensor)
```
vincentqb, line 326: Were you trying to reduce the use of classes/methods in each test?

test/test_torchscript_consistency.py
```diff
@@ -304,85 +304,75 @@ def test_perf_biquad_filtering(self):
         _test_torchscript_functional(F.lfilter, waveform, a, b)
 
 
-RUN_CUDA = torch.cuda.is_available()
-print("Run test with cuda:", RUN_CUDA)
+def _assert_transforms_consistency(transform, tensor, device):
```
vincentqb, line 307: Does this make the interface different for functionals?

test/test_torchscript_consistency.py
```diff
@@ -391,23 +381,32 @@ def test_Fade(self):
...
+        self._assert_conssitency(T.Vol(1.1), waveform)
+
+
+class TestTransformsCPU(_TransformsTestMixin, unittest.TestCase):
+    """Test suite for Transforms module on CPU"""
+    device = torch.device('cpu')
+
+
+@unittest.skipIf(not torch.cuda.is_available(), 'CUDA not available')
+class TestTransformsCUDA(_TransformsTestMixin, unittest.TestCase):
```
vincentqb, line 407: Overall, this looks reasonable to me. What would be an alternative way?

test/test_torchscript_consistency.py
mthrok: Yes, functionals do not have `to` method, so Tensor arguments have to be moved to the target device manually.

test/test_torchscript_consistency.py
mthrok: Yes, that's why the actual implementation is placed outside of the class and this helper method is a syntax sugar to pass `self.device` automatically. This way, even if we want to get rid of class structure, it's easy.

test/test_torchscript_consistency.py
mthrok: Actually, I do not come up with an alternative that is not hacky and as good as this one.
PyTorch's test dynamically generates this by manipulating global variable scope, but that's basically same thing as this implementation., yet more messy and it does not bring any additional benefit for us.

test/test_torchscript_consistency.py
vincentqb: Would there be a way to keep the two implementations closer?

mthrok: Refer to my [WIP](https://github.com/mthrok/audio/commit/c736e21746653c1132654d80adb83e71dc96d757#diff-04757c68c2f80adf8dac82791cd9d5beR13). To me this is very close.

PR #523, opened 2020-04-07
Use istft from torch

mthrok:
Corresponding PR https://github.com/pytorch/pytorch/pull/35569

This PR will use `istft` from torch repo, which is ported to C++.

This PR will not succeed until the upstream PR is merged and `torch.istft` becomes available.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -2,6 +2,7 @@
 
 import math
 from typing import Optional, Tuple
+import warnings
 
 import torch
 from torch import Tensor
@@ -49,7 +50,7 @@ def istft(
         win_length: Optional[int] = None,
         window: Optional[Tensor] = None,
         center: bool = True,
-        pad_mode: str = "reflect",
+        pad_mode: Optional[str] = None,
         normalized: bool = False,
         onesided: bool = True,
         length: Optional[int] = None,
@@ -94,8 +95,7 @@ def istft(
         center (bool, optional): Whether ``input`` was padded on both sides so
             that the :math:`t`-th frame is centered at time :math:`t \times \text{hop\_length}`.
             (Default: ``True``)
-        pad_mode (str, optional): Controls the padding method used when ``center`` is True. (Default:
-            ``"reflect"``)
+        pad_mode: This argument was ignored and to be removed.
         normalized (bool, optional): Whether the STFT was normalized. (Default: ``False``)
         onesided (bool, optional): Whether the STFT is onesided. (Default: ``True``)
         length (int or None, optional): The amount to trim the signal by (i.e. the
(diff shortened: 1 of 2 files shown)


cpuhrsch:
Did you double check that ```warnings``` is compatible with pytorch/pytorch?

torchaudio/functional.py
```diff
@@ -124,105 +125,15 @@ def istft(
...
-
-    # unpack batch
-    y = y.view(shape[:-3] + y.shape[-1:])
-
-    if stft_matrix_dim == 3:  # remove the channel dimension
-        y = y.squeeze(0)
-
-    return y
+    warnings.warn(
+        'istft has been moved to PyTorch and will be deprecated, please use torch.istft instead.')
```
gchanan, line 129: this should produce a UserWarning, which is our convention.

I don't know what it means that something "will be deprecated" -- something being deprecated means it will be changed eventually, so there's not really a delta between "will be deprecated" and "is deprecated".

mthrok: Oops I meant, 'will be removed'...

torchaudio/functional.py
```diff
@@ -124,105 +125,15 @@ def istft(
...
-    if stft_matrix_dim == 3:  # remove the channel dimension
-        y = y.squeeze(0)
-
-    return y
+    warnings.warn(
+        'istft has been moved to PyTorch and will be removed, please use torch.istft instead.')
+    warnings.warn(
+        '`pad_mode` for istft function is deprecated and will be removed. '
+        'It was introduced accidentally and never used. '
+        'You can safely remove it from your function call')
```
vincentqb, line 133: In the sense that `pad_mode` was added to mirror the signature of `stft`, `pad_mode` in `istft` is indeed not needed. Also, `pad_mode="reflect"` in `istft` still provides a reasonable inverse to `stft` with `pad_mode="constant"`, as shown in [torchaudio tests](https://github.com/pytorch/audio/blob/98fe8b46af5582f602f58ca9f9633393f6a931a7/test/test_functional.py#L90).

If we are to show this warning, let's only do so if the user attempts to change the default value. An alternative is to set the default value to `None`, and show the warning if the user changes `pad_mode` to something else than that.

nit: 'The parameter `pad_mode` was ignored in isftft, and is thus being deprecated. Please set pad_mode to None to suppress this warning.')

vincentqb:
Marked as WIP to emphasize that this is not ready to be merged.

vincentqb:
Now that pytorch/pytorch#35569 has been merged, removed label marking as do not merge.

torchaudio/functional.py
```diff
@@ -104,105 +104,16 @@ def istft(
...
-    assert 3 <= stft_matrix_dim, "Incorrect stft dimension: %d" % (stft_matrix_dim)
-    assert stft_matrix.numel() > 0
-
-    if stft_matrix_dim == 3:
-        # add a channel dimension
-        stft_matrix = stft_matrix.unsqueeze(0)
-
-    # pack batch
-    shape = stft_matrix.size()
-    stft_matrix = stft_matrix.reshape(-1, shape[-3], shape[-2], shape[-1])
```
vincentqb, line 117: This is what enabled the implementation to support batching.

PR #524, opened 2020-04-07
Add mypy Config File

tomassosorio:
In order to start using Mypy full potential, we must start fixing the issues that he might throw.
Relates to #477

As a starter I add the following definitions to the config file:

allow_redefinition: True
`Allows variables to be redefined with an arbitrary type, as long as the redefinition is in the same block and nesting level as the original definition.`
ignore_missing_imports: True
`Suppresses error messages about imports that cannot be resolved.`

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,0 +1,4 @@
+[mypy]
+allow_redefinition = True
+disallow_untyped_defs = False
+ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -1,4 +1,3 @@
 [mypy]
 allow_redefinition = True
-disallow_untyped_defs = False
 ignore_missing_imports = True


vincentqb:
For reference, [pytorch](https://github.com/pytorch/pytorch/blob/master/mypy.ini) does not have anything beyond python 2.7 compatibility. We do not want python 2.7 of course.

tomassosorio:
> For reference, [pytorch](https://github.com/pytorch/pytorch/blob/master/mypy.ini) does not have anything beyond python 2.7 compatibility. We do not want python 2.7 of course.

@vincentqb I did not understand what you meant by that, mypy currently is not passing at all and a lot of errors occur, this configuration will ignore small details such as a variable being change of its type.
 This will allow us to better focus in the errors that occur on torchaudio.

vincentqb:
Sorry for the misunderstanding :) My comment was meant as a note to self to reference what pytorch is doing.

Those two change from default seems reasonable to me, and a good start. Thanks for adding this :)

PR #526, opened 2020-04-08
Inline typing init

tomassosorio:
- Add inline typing to `__init__`
- [Atention] function `save_encinfo` had major changes due to error that mypy throw, parameters `signalinfo`, `encodinginfo` are mandatory and not optional

relates to: #477

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -18,6 +18,7 @@ from torchaudio._backend import (
     get_audio_backend,
     set_audio_backend,
 )
+from torchaudio._soundfile_backend import SignalInfo, EncodingInfo
 
 try:
     from .version import __version__, git_version  # noqa: F401
@@ -31,8 +32,8 @@ def load(filepath: Union[str, Path],
          channels_first: bool = True,
          num_frames: int = 0,
          offset: int = 0,
-         signalinfo: Any = None,
-         encodinginfo: Any = None,
+         signalinfo: Optional[SignalInfo] = None,
+         encodinginfo: Optional[EncodingInfo] = None,
          filetype: Optional[str] = None) -> Tuple[Tensor, int]:
     r"""Loads an audio file from disk into a tensor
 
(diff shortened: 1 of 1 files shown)


torchaudio/__init__.py
```diff
@@ -122,23 +124,21 @@ def save(filepath, src, sample_rate, precision=16, channels_first=True):
...
-                 channels_first=True,
-                 signalinfo=None,
-                 encodinginfo=None,
-                 filetype=None):
+def save_encinfo(filepath: str,
+                 src: Tensor,
+                 signalinfo: SignalInfo,
+                 encodinginfo: EncodingInfo,
+                 channels_first: bool = True,
+                 filetype: Optional[str] = None) -> None:
```
vincentqb, line 132: Let's not make a BC-breaking change in this PR. Can you clarify what issue you are trying to fix?

As used [here](https://github.com/pytorch/audio/blob/master/torchaudio/_sox_backend.py#L65), `encodinginfo` is indeed optional.

tomassosorio: Sure :) 
 in line 168 we have:
```
    if not isinstance(signalinfo.rate, float):
        if float(signalinfo.rate) == signalinfo.rate:
            signalinfo.rate = float(signalinfo.rate)
        else:
            raise TypeError('Sample rate should be a float or int')
    # check if the bit precision (i.e. bits per sample) is an integer
    if not isinstance(signalinfo.precision, int):
        if int(signalinfo.precision) == signalinfo.precision:
            signalinfo.precision = int(signalinfo.precision)
        else:
            raise TypeError('Bit precision should be an integer')
```
however if `signalinfo` is `None` we cannot get `rate` so will break, the solution that I first work was using an `if signalinfo:` and indent the rest (lines 168-178), which would only do those checks if the parameter was actually defined, however when I tried to run `_torch_sox.write_audio_file` and if you look into its typing is mandatory to have `signalinfo` defined, so I opted by changing it to mandatory, however, it can be fixed by using the if-condition and maybe adding a raise if its not defined. Your call :)

tomassosorio: @vincentqb Updated to not break BC

torchaudio/__init__.py
```diff
@@ -165,17 +167,18 @@ def save_encinfo(filepath,
...
-            signalinfo.rate = float(signalinfo.rate)
-        else:
-            raise TypeError('Sample rate should be a float or int')
-    # check if the bit precision (i.e. bits per sample) is an integer
-    if not isinstance(signalinfo.precision, int):
-        if int(signalinfo.precision) == signalinfo.precision:
-            signalinfo.precision = int(signalinfo.precision)
-        else:
-            raise TypeError('Bit precision should be an integer')
+    if signalinfo:
```
vincentqb, line 170: Can you provide a code example with the error that you want to fix here?

vincentqb, line 170: Thanks for catching this :)

PR #528, opened 2020-04-09
Separate CPU and GPU tests for functions torchscript test + Fix devices in two functionals

mthrok:
Similar to #520 this PR make

1. torchscript consistency test on `functional` module  device aware
2. Define the above test for CPU and GPU
3. Fix some device compatibility issues in `functional` module found in Step 2.

Relates to #515

### DIFF
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -10,6 +10,18 @@ import torchaudio.transforms as T
 import common_utils
 
 
+def _assert_functional_consistency(func, tensor, device, shape_only=False):
+    tensor = tensor.to(device)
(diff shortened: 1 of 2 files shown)


mthrok:
That test only failed on PY3.6 and it succeeded after re-run, meaning that the test is flaky.
The randomness in [`mask_along_axis_iid` function](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L1213-L1214) and [in the `test_functional.py`](https://github.com/pytorch/audio/blob/master/test/test_functional.py#L365) are causing it.

The proper fix will be to set random seed in `test_functional.py`. I can create a PR for that.

PR #531, opened 2020-04-10
Call torch.stft directly

mthrok:
Found the note

`# TODO: remove this once https://github.com/pytorch/pytorch/issues/21478 gets solved`

And the issue is resolved and we should not need to wrap `torch.stft` anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,13 +13,13 @@ cache:
 matrix:
   fast_finish: true
   include:
-    - env: PYTHON_VERSION="3.5"
     - env: PYTHON_VERSION="3.6"
     - env: PYTHON_VERSION="3.7"
-    - env: PYTHON_VERSION="3.5" RUN_FLAKE8="true" SKIP_INSTALL="true" SKIP_TESTS="true"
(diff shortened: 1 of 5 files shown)


mthrok:
Interesting! It does not fail on my environment 🤔

vincentqb:
> Interesting! It does not fail on my environment 🤔

The fix pytorch/pytorch#33504 isn't part of pytorch 1.4.0 on which the torchaudio test environment depends, see `.circleci/config.yaml.in`.

vincentqb:
We'll merge this PR once Pytorch 1.5.0 has landed.

mthrok:
Actually, that's not where test environment is defined. it's in https://github.com/pytorch/audio/blob/master/requirements.txt#L1 via https://github.com/pytorch/audio/blob/master/build_tools/travis/install.sh#L51 via https://github.com/pytorch/audio/blob/master/.travis.yml

vincentqb:
Yes, torchaudio master should depend on pytorch master; and the latest torchaudio release should depend on the latest pytorch release. Do you want to a PR to make torchaudio master depend on pytorch nightly?

PR #544, opened 2020-04-14
Fix inline typing for mypy

tomassosorio:
relates to #477 
This should fix all mypy issues and could be used to check if PR follow all its agreements running:
` mypy torchaudio/`
getting:
`Success: no issues found in 19 source files
`

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -168,8 +168,8 @@ def save_encinfo(filepath: str,
     # sox stores the sample rate as a float, though practically sample rates are almost always integers
     # convert integers to floats
     if signalinfo:
-        if not isinstance(signalinfo.rate, float):
-            if signalinfo.rate and float(signalinfo.rate) == signalinfo.rate:
+        if signalinfo.rate and not isinstance(signalinfo.rate, float):
+            if float(signalinfo.rate) == signalinfo.rate:
                 signalinfo.rate = float(signalinfo.rate)
             else:
                 raise TypeError('Sample rate should be a float or int')
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -169,12 +169,12 @@ def save_encinfo(filepath: str,
     # convert integers to floats
     if signalinfo:
         if not isinstance(signalinfo.rate, float):
-            if float(signalinfo.rate) == signalinfo.rate:
+            if signalinfo.rate and float(signalinfo.rate) == signalinfo.rate:
                 signalinfo.rate = float(signalinfo.rate)
             else:
                 raise TypeError('Sample rate should be a float or int')
         # check if the bit precision (i.e. bits per sample) is an integer
-        if not isinstance(signalinfo.precision, int):
+        if signalinfo.precision and not isinstance(signalinfo.precision, int):
             if int(signalinfo.precision) == signalinfo.precision:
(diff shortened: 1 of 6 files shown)


torchaudio/datasets/utils.py
```diff
@@ -7,17 +7,18 @@
...
 
 import torch
 import urllib
+import urllib.request
 from torch.utils.data import Dataset
 from torch.utils.model_zoo import tqdm
 
 
-def unicode_csv_reader(unicode_csv_data: TextIOWrapper, **kwargs: Any) -> str:
+def unicode_csv_reader(unicode_csv_data: TextIOWrapper, **kwargs: Any) -> Any:
```
vincentqb, line 21: This is not BC breaking yet since the inline typing isn't part of the release. `str` was added in [here](https://github.com/pytorch/audio/blame/master/torchaudio/datasets/utils.py#L20).

torchaudio/datasets/utils.py
```diff
@@ -129,7 +130,8 @@ def download_url(url: str,
 
     if resume and os.path.exists(filepath):
         mode = "ab"
-        local_size = os.path.getsize(filepath)
+        local_size = os.path.getsize(filepath)  # type: Optional[int]
```
vincentqb, line 133: First, we need to check that the rest of pytorch is following that style. In the meanwhile, I suggest to skip this test in mypy.

By the way, since torchaudio's master branch only support python 3.6 and up, we could use
`local_size: Optional[int] = os.path.getsize(filepath)` instead.

torchaudio/datasets/utils.py
```diff
@@ -286,7 +288,7 @@ def __init__(self, dataset: Dataset, location: str = ".cached") -> None:
         self.location = location
 
         self._id = id(self)
-        self._cache = [None] * len(dataset)
+        self._cache = [None] * len(dataset)   # type: List
```
vincentqb, line 291: same

torchaudio/datasets/utils.py
```diff
@@ -325,7 +327,7 @@ class _End:
 
     def __init__(self, generator: Iterable, maxsize: int) -> None:
         threading.Thread.__init__(self)
-        self.queue = Queue(maxsize)
+        self.queue = Queue(maxsize)  # type: Queue
```
vincentqb, line 330: same

torchaudio/datasets/utils.py
tomassosorio: This was changed due to `csv.reader` has as output Any instead of str so this function actually will give `Any` and not solely `str` has output

torchaudio/__init__.py
```diff
@@ -168,13 +168,13 @@ def save_encinfo(filepath: str,
     # sox stores the sample rate as a float, though practically sample rates are almost always integers
     # convert integers to floats
     if signalinfo:
-        if not isinstance(signalinfo.rate, float):
+        if signalinfo.rate and not isinstance(signalinfo.rate, float):
```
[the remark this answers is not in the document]
tomassosorio: actually this is due to [SignalInfo](https://github.com/pytorch/audio/blob/master/torchaudio/_soundfile_backend.py#L16) is defined as `None` in its variables. 
So this will check if the variable exists if so then will check if the variable `isinstance` `float`

PR #548, opened 2020-04-14
backend change instructions in readme

vincentqb:
We could show the instructions on how to change the backend in the readme.


mthrok:
Looking at README, there is no mention on backend. So I think it's better to explain what is backend and what backends are available for which OS. (and optionally how to add backend.)

vincentqb:
superseded by #553

PR #551, opened 2020-04-15
Add contrast to functional

bhargavkathivarapu:
Hi ,

Regarding the issue #260 ( Reducing dependency in Sox)
Implemented below function in functional.py
 - contrast

Added a test case  to test_sox_compatibility.py

All tests in test_sox_compatibility.py passed on CPU.

```
................gain: can't reclaim headroom
.
Ran 17 tests in 501.774s

OK
```
@vincentqb could you please review changes

### DIFF
--- a/test/test_batch_consistency.py
+++ b/test/test_batch_consistency.py
@@ -65,6 +65,18 @@ class TestFunctional(unittest.TestCase):
         ])
         _test_batch(F.istft, stft, n_fft=4, length=4)
 
+    def test_contrast(self):
+        filenames = [
+            'steam-train-whistle-daniel_simon.wav',  # 2ch 44100Hz
+            # Files from https://www.mediacollege.com/audio/tone/download/
+            '100Hz_44100Hz_16bit_05sec.wav',  # 1ch
+            '440Hz_44100Hz_16bit_05sec.wav',  # 1ch
+        ]
+        for filename in filenames:
+            filepath = common_utils.get_asset_path(filename)
+            waveform, sample_rate = torchaudio.load(filepath)
+            _test_batch(F.contrast, waveform, enhancement_amount=80.)
+
 
 class TestTransforms(unittest.TestCase):
     """Test suite for classes defined in `transforms` module"""
--- a/test/test_batch_consistency.py
+++ b/test/test_batch_consistency.py
@@ -66,16 +66,8 @@ class TestFunctional(unittest.TestCase):
         _test_batch(F.istft, stft, n_fft=4, length=4)
 
     def test_contrast(self):
-        filenames = [
-            'steam-train-whistle-daniel_simon.wav',  # 2ch 44100Hz
-            # Files from https://www.mediacollege.com/audio/tone/download/
-            '100Hz_44100Hz_16bit_05sec.wav',  # 1ch
-            '440Hz_44100Hz_16bit_05sec.wav',  # 1ch
-        ]
-        for filename in filenames:
-            filepath = common_utils.get_asset_path(filename)
-            waveform, sample_rate = torchaudio.load(filepath)
-            _test_batch(F.contrast, waveform, enhancement_amount=80.)
+        waveform = torch.rand(2, 100) - 0.5
+        _test_batch(F.contrast, waveform, enhancement_amount=80.)
 
 
 class TestTransforms(unittest.TestCase):
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -31,6 +31,7 @@ __all__ = [
     "deemph_biquad",
     "riaa_biquad",
     "biquad",
+    "contrast",
     'mask_along_axis',
     'mask_along_axis_iid'
 ]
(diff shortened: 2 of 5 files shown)


torchaudio/functional.py
```diff
@@ -1160,6 +1161,38 @@ def riaa_biquad(
...
+            Note that enhancement_amount = 0 still gives a significant contrast enhancement
+
+    Returns:
+        Tensor: Waveform of dimension of `(..., time)`
+
+    References:
+        http://sox.sourceforge.net/sox.html
+    """
+
+    if (enhancement_amount < 0 or enhancement_amount > 100):
```
mthrok, line 1184: nit You can also write

```
if not 0 <= enhancement_amount <= 100:
```

torchaudio/functional.py
```diff
@@ -1160,6 +1161,38 @@ def riaa_biquad(
...
+        waveform: Tensor,
+        enhancement_amount: float = 75.
+) -> Tensor:
+    r"""Apply contrast effect.  Similar to SoX implementation.
+    Comparable with compression, this effect modifies an audio signal to make it sound louder
+
+    Args:
+        waveform (Tensor): audio waveform of dimension of `(..., time)`
+        enhancement_amount (float): controls the amount of the enhancement
+            Allowed range of values for enhancement_amount : 0-100
```
vincentqb, line 1174: nit: I would have been tempted to scale this parameter to be between 0 and 1, instead of 0 and 100. Since sox is using 0 and 100, we could just keep that. Thoughts?

mthrok:
@bhargavkathivarapu If you need help in adding the tests, let me know.
For batch test you can add `test_contrast` under `TestFunctional` and call `_test_batch` function.
For torchscript test, you can add `test_contrast` under `_FunctionalTestMixin` and define an inner function that takes input Tensor as the only argument (other parameters should be embedded in the function), then call `self._assert_consistency` with input Tensor.

bhargavkathivarapu:
@vincentqb  
- Added batching test for contrast
- Added torchscript test for contrast
- Updated functional.rst for contrast and other missing functions

@mthrok , @vincentqb  
I think we should have a CONTRIBUTING.md just like the pytorch repo.
Detailing orgainization of torchaudio code base and with the checklists indicating what tests  to do and where to add them when adding new features to functional , transforms or datasets

torchaudio/functional.py
bhargavkathivarapu: I think we should keep it 0-100 , so that those who are using existing sox effects will find it easy when migrating to  functional . Let me know if we want to scale 0 to 1 . I will make the change

test/test_batch_consistency.py
```diff
@@ -65,6 +65,18 @@ def test_istft(self):
         ])
         _test_batch(F.istft, stft, n_fft=4, length=4)
 
+    def test_contrast(self):
```
mthrok, line 68: Can you use simple random tensor of value range [-1.0, 1.0] i.e. `torch.rand(2, 100) - 0.5` ?

The reason is that we do not want this unit test to rely on other library function `torchaudio.load`.
That way when `torchaudio.load` is broken, this test is not affected.
I know that `test_detect_pitch_frequency` does this but it's only because of recent refactoring and in my opinion it should not.

mthrok:
> I think we should have a CONTRIBUTING.md just like the pytorch repo.
> Detailing orgainization of torchaudio code base and with the checklists indicating what tests to do and where to add them when adding new features to functional , transforms or datasets

I agree. Recently I also had difficult time figuring out what tests need to be added so I reorganized most of the test, and I was planning to add overview of the test structure. Thanks for bringing this up.

test/test_batch_consistency.py
bhargavkathivarapu: @mthrok I have this modified . 
Are torchaudio functionals conditioned on normalization ?
- Because  sox in some functions it internally clips the values as per max and min of input. 
- I assumed normalized waveform in writing contrast function . 
- In tests also i didn't see anywhere normalization=False.
- If unnormalized waveform is passed, the outputs will differ between existing SoxEffects and functional contrast

mthrok: @bhargavkathivarapu 

I am not sure if there is/was a design principle for audio normalization but since `torchaudio.load` has `normalization=True` by default, and all the tests are written in `normalization=True` so I think that's de-facto standard.

Talking about batch consistency test here specifically, the purpose of the test is to assure that function returns the same result regardless of batching and not to assure the result is same as SoX. So it's okay to put whatever the tensor, but to be closer to the actual situation, I thought rather than `torch.rand` (which produces [0, 1]), `torch.rand - 0.5` would be slightly better. You can normalize it to be further closer if you would like.

mthrok:
Tests look good to me. Thanks! @bhargavkathivarapu

torchaudio/functional.py
vincentqb: Sounds good :)

test/test_batch_consistency.py
vincentqb: > I am not sure if there is/was a design principle for audio normalization but since `torchaudio.load` has `normalization=True` by default, and all the tests are written in `normalization=True` so I think that's de-facto standard.

Yes, we assume that waveforms are in the range [-1, 1] in torchaudio.

We should consider documenting this convention in the readme, and guaranteeing it in some form. A caveat is that, when doing a transformation that makes the waveform fall outside (e.g. gain), there may be more than one way to bring the signal back to the [-1, 1] range.

PR #553, opened 2020-04-15
backend change instructions in readme

vincentqb:
Show the instructions on how to switch the backend amongst available ones, in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ having all the computations be through Pytorch operations which makes it easy
 to use and feel like a natural extension.
 
 - [Support audio I/O (Load files, Save files)](http://pytorch.org/audio/)
-  - Load the following formats into a torch Tensor using sox
+  - Load the following formats into a torch Tensor using SoX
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,10 +98,29 @@ Quick Usage
...
+torchaudio.save('foo_save.wav', waveform, sample_rate)  # save tensor to file
 ```
 
+Backend Dispatch
+----------------
+
+By default in OSX and Linux, torchaudio uses SoX as a backend to load and save files.
+The backend can be changed to [SoundFile](https://pysoundfile.readthedocs.io/en/latest/)
+using the following. See [SoundFile](https://pysoundfile.readthedocs.io/en/latest/)
+for installation instructions.
```
cpuhrsch, line 112: I would suggest providing some more information here specific to pip and conda. You can directly link to the package on pypi for example to avoid confusion with projects such as [sndfile](https://pypi.org/project/sndfile/), which is the name used by the underlying libsndfile.

Issue #557, opened 2020-04-16
Small clean up `test_sox_compatibility.py`


mthrok:
Among the formatting issues in [`test_sox_compatibility.py`](https://github.com/pytorch/audio/blob/master/test/test_sox_compatibility.py#L122), there are two that can be immediately fixed.
 - local variables somehow all capitalized such as [`CENTRAL_FREQ`](https://github.com/pytorch/audio/blob/master/test/test_sox_compatibility.py#L122) and [`Q`](https://github.com/pytorch/audio/blob/master/test/test_sox_compatibility.py#L123) should be lowercased `central_freq` and `q`
 - [unused variable](https://github.com/pytorch/audio/blob/master/test/test_sox_compatibility.py#L316) should be `_`

I can work on this but if anyone is interested in contributing to the code base. It's highly welcomed and appreciated.
**Note**: Please limit the change to the ones mentioned above.

PR #559, opened 2020-04-17
Fix  variable names in test_sox_compatibility.py

bhargavkathivarapu:
Hi ,
Regarding the issue #557 . 
I have changed all upper case variables to lower case in test_sox_compatibility.py

@mthrok  could you please review these changes


mthrok:
Thanks, LGTM.

class #557

PR #540, opened 2020-04-14
add cmvn

wanglong001:
HI ，[issues  ](https://github.com/pytorch/audio/issues/535)， add feature of kaldi cmvn

ths

Closes #535.

PR #560, opened 2020-04-17
Add Kaldi compatibility test for `sliding_window_cmn`

mthrok:
Follow up of #540 

This PR adds
 - Kaldı compatibility test
 - ~~batch consistency test~~
for `sliding_window_cmn`

~~and make `sliding_window_cmn` handle batched input.~~

cc @wanglong001 

EDIT: batch ompatibility part was moved to #570

### DIFF
--- a/test/test_kaldi_compatibility.py
+++ b/test/test_kaldi_compatibility.py
@@ -1,0 +1,53 @@
+"""Test suites for checking numerical compatibility against Kaldi"""
+import shutil
+import unittest
+import subprocess
+
+import kaldi_io
+import torch
+import torchaudio.functional as F
+
+
+def _exe_exists(cmd):
+    return shutil.which(cmd) is not None
+
+
+def _convert_args(**kwargs):
+    args = []
+    for key, value in kwargs.items():
+        key = '--' + key.replace('_', '-')
+        value = str(value).lower() if value in [True, False] else str(value)
+        args.append('%s=%s' % (key, value))
+    return args
+
+
+def _run_kaldi(command, input_tensor):
+    """Run provided Kaldi command, pass a tensor and get the resulting tensor
+
+    Assumption:
+        The provided Kaldi command consumes one ark and produces one ark.
+        i.e. 'ark:- ark:-'
+    """
+    process = subprocess.Popen(command, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
+    kaldi_io.write_mat(process.stdin, input_tensor.numpy(), key='foo')
+    process.stdin.close()
+    result = dict(kaldi_io.read_mat_ark(process.stdout))['foo']
+    return torch.from_numpy(result.copy())  # copy supresses some torch warning
+
+
+class TestFunctional:
+    @unittest.skipUnless(_exe_exists('apply-cmvn-sliding'), '`apply-cmvn-sliding` not available')
+    def test_sliding_window_cmn(self):
+        """sliding_window_cmn should be numerically compatible with apply-cmvn-sliding"""
+        kwargs = {
+            'cmn_window': 600,
+            'min_cmn_window': 100,
+            'center': False,
+            'norm_vars': False,
+        }
+
+        tensor = torch.randn(40, 10)
+        result = F.sliding_window_cmn(tensor, **kwargs)
+        command = ['apply-cmvn-sliding'] + _convert_args(**kwargs) + ['ark:-', 'ark:-']
+        kaldi_result = _run_kaldi(command, tensor)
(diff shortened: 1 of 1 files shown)


test/test_batch_consistency.py
```diff
@@ -69,6 +69,13 @@ def test_contrast(self):
         waveform = torch.rand(2, 100) - 0.5
         _test_batch(F.contrast, waveform, enhancement_amount=80.)
 
+    def test_sliding_window_cmn(self):
+        waveform = torch.randn(2, 1024) - 0.5
+        _test_batch(F.sliding_window_cmn, waveform, center=True, norm_vars=True)
+        _test_batch(F.sliding_window_cmn, waveform, center=True, norm_vars=False)
+        _test_batch(F.sliding_window_cmn, waveform, center=False, norm_vars=True)
+        _test_batch(F.sliding_window_cmn, waveform, center=False, norm_vars=False)
```
mthrok, line 77: We should definitely use `pytest.mark.parametrize` to simplify this.

mthrok:
The changes I made in `functional.py` are necessary to pass the test, but I am not sure if the tests cover all the branches in the function. Adding coverage report to test would help finding out this kind of edge case.

test/test_batch_consistency.py
vincentqb: You can try `subTest` in `unittest`.

mthrok: Our test runner does not understand `subTest`. The test name does not show up in the execution result, so `subTest` does not add any value despite more code and indentation.

test/test_kaldi_compatibility.py
```diff
@@ -0,0 +1,53 @@
...
+    args = []
+    for key, value in kwargs.items():
+        key = '--' + key.replace('_', '-')
+        value = str(value).lower() if value in [True, False] else str(value)
+        args.append('%s=%s' % (key, value))
+    return args
+
+
+def _run_kaldi(command, input_tensor):
+    """Run provided Kaldi command, pass a tensor and get the resulting tensor
```
vincentqb, line 25: Since kaldi is not installed in CI, the test is skipped in travis
```
(test/test_kaldi_compatibility.py::TestFunctional::test_sliding_window_cmn SKIPPED [100%]
```

For the other tests, we simply saved the output from kaldi, and compare against that. This allows us to run the test on CI. Let's do that.

If you want to keep the infrastructure you have added to run kaldi commands, we can do it too, along with the test just mentioned. We can rename this as e.g. `test_sliding_window_cmn_against_installed_kaldi`.

mthrok: My concern with the previous approach is that it adds overhead for both adding a new test and maintaining the test, which leads to the situation where every time a new Kaldi-compatible functionality is added, one of us has to add all the related files, because it is a complicated process to review.

When we run tests on Circle CI, we can bake Kaldi binaries into base test image. Making Kaldi accessible on CI should be easy.

vincentqb: I do also prefer using kaldi commands to test against for new transforms, over testing against saved kaldi output files. This has the advantage of making sure we stay in sync with kaldi, and we can regenerate the exact command that produces the kaldi output we want.

Adding kaldi is easy (though there is always a risk that it won't run easily on some platform like windows), but we should isolate this change in a separate PR from the testing. And, since kaldi is not yet available in CircleCI, this test is not running and so is not showing us that `sliding_window_cmn` is compatible with kaldi.

Even if kaldi were available, we do not have convincing evidence in this PR's CircleCI that the new test is indeed providing the signal we want.

I see two options:
* We add kaldi in the testing build now.
* We add a second test against the kaldi output file, along with what you have here.

Let's go for the latter in this PR, and follow up with the former.

mthrok:
@vincentqb 

I used the new image being added in https://github.com/pytorch/audio/pull/581 to run the Kaldi compatibility test and it passed.

Search `test_kaldi_compatibility` in `Run test` step.

https://app.circleci.com/pipelines/github/pytorch/audio/1514/workflows/e2be1e0d-641b-46e9-ae16-1775e3fed72d

mthrok:
> * We could decide to switch all the kaldi tests to this.

I think we can create `help-wanted` tasks for that. I think it is really nice if we can get rid of the previous test and helper scripts.

> * We could start thinking of migrating tests from travis to circleci only once all tests on travis also run on circleci.

Sure.

vincentqb:
> I think we can create `help-wanted` tasks for that. I think it is really nice if we can get rid of the previous test and helper scripts.

Good idea: #597

PR #569, opened 2020-04-21
Add overdrive to functional

bhargavkathivarapu:
Hi ,
Regarding the issue #260 ( Reducing dependency in Sox)
Implemented ```overdrive``` function in functional.py

   

-  function.rst - updated with new fucntion

  

-   test_sox_compatibility.py - Added a test for overdrive . Tests passed

```
...................gain: can't reclaim headroom
.
Ran 20 tests in 509.658s

OK
```

  
-   test_torchscript_consistency.py - added test . Test passed

```
...................................sssssssssssssssssssssssssssssssssss................ssssssssssssssss
Ran 102 tests in 1400.141s

OK (skipped=51)
````

-  test_batch_consistency.py - added test for overdrive . Test for overdrive passed, but 
existing batch_mfcc  failed
```
...............F..
FAIL: test_batch_mfcc (__main__.TestTransforms)
  File "test/test_batch_consistency.py", line 201, in test_batch_mfcc
    torch.testing.assert_allclose(computed, expected, atol=1e-5, rtol=1e-5)
  File "/Users/ka387861/pytorch/torch/testing/__init__.py", line 59, in assert_allclose
    count - 1, 100 * count / actual.numel()))
AssertionError: Not within tolerance rtol=1e-05 atol=1e-05 at input[2, 0, 2, 1384] (0.5488258004188538 vs. 0.5488020181655884) and 8 other locations (0.00%)

Ran 18 tests in 185.586s
```

For making the torch script test work i added a redundant operation . 
Comment mentioning that is present in code.

@vincentqb could you please review these changes

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -1268,17 +1268,17 @@ def overdrive(
 
     gain = _dB2Linear(gain)
     colour = colour / 200
-    last_in = torch.zeros(waveform.shape[0],waveform.shape[1])
-    last_out = torch.zeros(waveform.shape[0],waveform.shape[1])
+    last_in = torch.zeros(waveform.shape[0], waveform.shape[1])
+    last_out = torch.zeros(waveform.shape[0], waveform.shape[1])
 
     temp = waveform * gain + colour
 
     mask1 = temp < -1
     temp[mask1] = temp[mask1] * 0 - 2 / 3
-    #Above redundant multiplcation to accomadate torchscript issue
+    # Above redundant multiplcation to accomadate torchscript issue
 
     mask2 = temp > 1
-    temp[mask2] =  temp[mask2] * 0 + 2 / 3
+    temp[mask2] = temp[mask2] * 0 + 2 / 3
 
     mask3 = (~mask1 & ~mask2)
     temp[mask3] = temp[mask3] - (temp[mask3]**3) * (1. / 3)
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -1268,16 +1268,17 @@ def overdrive(
 
     gain = _dB2Linear(gain)
     colour = colour / 200
-    last_in = 0.
-    last_out = 0.
+    last_in = torch.zeros(waveform.shape[0],waveform.shape[1])
+    last_out = torch.zeros(waveform.shape[0],waveform.shape[1])
 
     temp = waveform * gain + colour
 
     mask1 = temp < -1
-    temp[mask1] = waveform[mask1] * 0 + (-2 / 3)
+    temp[mask1] = temp[mask1] * 0 + 2 / 3
+    #Above redundant multiplcation to accomadate torchscript issue
 
     mask2 = temp > 1
-    temp[mask2] = waveform[mask2] * 0 + (2 / 3)
+    temp[mask2] =  temp[mask2] * 0 - 2 / 3
 
     mask3 = (~mask1 & ~mask2)
     temp[mask3] = temp[mask3] - (temp[mask3]**3) * (1. / 3)
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -33,6 +33,7 @@ __all__ = [
     "biquad",
     "contrast",
     "dcshift",
+    "overdrive",
     'mask_along_axis',
     'mask_along_axis_iid',
     'sliding_window_cmn',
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: torchaudio/functional.py [hunk 1]
```diff
@@ -1239,6 +1240,60 @@ def dcshift(
...
+def overdrive(
+        waveform: Tensor,
+        gain: float = 20,
+        colour: float = 20
+) -> Tensor:
+    r"""Apply a overdrive effect to the audio. Similar to SoX implementation.
+    This effect applies a non linear distortion to the audio signal.
+
+    Args:
+        waveform (Tensor): audio waveform of dimension of `(..., time)`
```
mthrok, line 1252: Looking at the implementation, it seems that this function expects some shape of Tensor. Can you elaborate that here?

I cannot infer what is the shape you are referring in the later part of the code.

```
last_in = torch.zeros(waveform.shape[0], waveform.shape[1])
last_out = torch.zeros(waveform.shape[0], waveform.shape[1])
```

torchaudio/functional.py
```diff
@@ -1239,6 +1240,60 @@ def dcshift(
...
+    """
+    actual_shape = waveform.shape
+
+    if len(actual_shape) == 2:
+        waveform = waveform.unsqueeze(0)
+
+    gain = _dB2Linear(gain)
+    colour = colour / 200
+    last_in = torch.zeros(waveform.shape[0], waveform.shape[1])
+    last_out = torch.zeros(waveform.shape[0], waveform.shape[1])
```
mthrok, line 1272: Please set the same `dtype` and `device` as input tensor.

i.e.

```diff
@@ -1262,23 +1262,24 @@ def overdrive(
         http://sox.sourceforge.net/sox.html
     """
     actual_shape = waveform.shape
+    device, dtype = waveform.device, waveform.dtype

     if len(actual_shape) == 2:
         waveform = waveform.unsqueeze(0)

     gain = _dB2Linear(gain)
     colour = colour / 200
-    last_in = torch.zeros(waveform.shape[0], waveform.shape[1])
-    last_out = torch.zeros(waveform.shape[0], waveform.shape[1])
+    last_in = torch.zeros(waveform.shape[0], waveform.shape[1], dtype=dtype, device=device)
+    last_out = torch.zeros(waveform.shape[0], waveform.shape[1], dtype=dtype, device=device)
```

torchaudio/functional.py
```diff
@@ -1239,6 +1240,60 @@ def dcshift(
...
+
+    temp = waveform * gain + colour
+
+    mask1 = temp < -1
+    temp[mask1] = temp[mask1] * 0 - 2 / 3
```
mthrok, line 1277: You should be able to do something like

```
temp[mask1] = torch.tensor(-2.0 / 3.0, dtype=dtype, device=device)
```

then leave a comment, like

```
# Wrapping the constant with Tensor is required for Torchscript
```

torchaudio/functional.py
mthrok: Also, I am not sure what is the expected shape of input tensor, but if the function needs to handle input tensor with dimension larger than 3, the shape of `last_in` and `last_out` should be able to handle it.

torchaudio/functional.py
```diff
@@ -1239,6 +1240,60 @@ def dcshift(
...
+
+    mask2 = temp > 1
+    temp[mask2] = temp[mask2] * 0 + 2 / 3
+
+    mask3 = (~mask1 & ~mask2)
+    temp[mask3] = temp[mask3] - (temp[mask3]**3) * (1. / 3)
+
+    output_waveform = torch.zeros_like(waveform)
+
+    # TODO: Implement a torch CPP extension
```
mthrok, line 1288: Out of curiosity, do you plan to do this? (I am not asking you to do so.)

torchaudio/functional.py
bhargavkathivarapu: updated

torchaudio/functional.py
bhargavkathivarapu: I have updated the function to take any shape waveform given that last dimension is time

torchaudio/functional.py
bhargavkathivarapu: I have updated the function to take any shape waveform given that last dimension is time. now last_in and last_out can handle any waveform shape

torchaudio/functional.py
bhargavkathivarapu: yeah . haven't written any cpp extension till now , but will try to do it sometime later . The python version is very slow

PR #572, opened 2020-04-22
Handle SoX init/shutdown automatically

mthrok:
This PR

- Adds safeguard to SoX initialization/shutdown
- When SoX is initialized, it register the exit hook to call shutdown using `atexit`
  - SoX is shutdown at the end of program execution.
  - Users can still manually shutdown.
- Automatically initialize SoX when a building effects chain. This frees the users from need to call `initialize_sox`/`shutdown_sox`, yet they can still do so. This gives better backend abstraction.

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -1,3 +1,4 @@
+import atexit
 import os.path
 from pathlib import Path
 from typing import Any, Callable, Optional, Tuple, Union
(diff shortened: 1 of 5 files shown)


torchaudio/__init__.py
```diff
@@ -338,25 +339,65 @@ def get_sox_bool(i: int = 0) -> Any:
...
+# True: Initialized
+# None: Already shut down (should not be initialized again.)
+
+_SOX_SUCCESS_CODE = 0
+# defined at
+# https://fossies.org/dox/sox-14.4.2/sox_8h.html#a8e07e80cebeff3339265d89c387cea93a9ef2b87ec303edfe40751d9a85fadeeb
+
+
 @_audio_backend_guard("sox")
 def initialize_sox() -> int:
```
mthrok, line 354: I think the whole logic of `initialize_sox`/`shutdown_sox` can be moved to effect chain and mark them as internal, because users no longer need to worry about it.

torchaudio/__init__.py
```diff
@@ -338,25 +339,65 @@ def get_sox_bool(i: int = 0) -> Any:
...
 
+_SOX_INITIALIZED = False
+# This variable has a micro lifecycle. (False -> True -> None)
+# False: Not initialized
+# True: Initialized
+# None: Already shut down (should not be initialized again.)
+
+_SOX_SUCCESS_CODE = 0
+# defined at
+# https://fossies.org/dox/sox-14.4.2/sox_8h.html#a8e07e80cebeff3339265d89c387cea93a9ef2b87ec303edfe40751d9a85fadeeb
```
mthrok, line 350: It kind of bugs me to have duplicated constant definition here.
An alternative approach would be, changing the CPP impl of `initialize_sox` to return boolean, indicating success / failure by performing return code check on CPP side.

torchaudio/__init__.py
```diff
@@ -338,25 +339,65 @@ def get_sox_bool(i: int = 0) -> Any:
...
+    Note:
+        This function is not required for simple loading.
+
+    Returns:
+        int: Code corresponding to sox_error_t enum. See
+        https://fossies.org/dox/sox-14.4.2/sox_8h.html#a8e07e80cebeff3339265d89c387cea93
+    """
+    global _SOX_INITIALIZED
+    if _SOX_INITIALIZED is None:
+        raise RuntimeError('SoX effects chain has been already shut down. Can not initialize again.')
```
mthrok, line 371: We do not need to do this, but raising `RuntimeError` with a message would be better than segfault.

torchaudio/__init__.py
```diff
@@ -338,25 +339,65 @@ def get_sox_bool(i: int = 0) -> Any:
...
+    if not _SOX_INITIALIZED:
+        import _torch_sox
```
mthrok, line 373: It would be better to have debug log here.

```
logger.debug('Initializing SoX effects chain.')
```

torchaudio/__init__.py
```diff
@@ -338,25 +339,65 @@ def get_sox_bool(i: int = 0) -> Any:
...
+    You do not need to call this function as it will be called automatically
+    at the end of program execution, only if ``initialize_sox`` was called.
+
+    Returns:
+        int: Code corresponding to sox_error_t enum. See
+        https://fossies.org/dox/sox-14.4.2/sox_8h.html#a8e07e80cebeff3339265d89c387cea93
+    """
+    global _SOX_INITIALIZED
+    if _SOX_INITIALIZED:
+        import _torch_sox
```
mthrok, line 395: Similarly, it would be better to have debug log here.

```
logger.debug('Shutting down SoX effects chain.')
```

torchaudio/__init__.py
vincentqb: This can be done in a separate PR later.

torchaudio/__init__.py
vincentqb: We could make an explicit commitment to add debugging everywhere, but let's not do that here.

torchaudio/__init__.py
vincentqb: Essentially, you suggest closing sox on exit only, right? We can discuss that on a separate pull request.

PR #573, opened 2020-04-22
Define proper logger for module

mthrok:
`logging.info` writes to the root logger, which should be reserved for application logging, and it makes hard for users to filter `torchaudio`.
Also, it is recommended to let logging module string substitution only when the log is actually emitted.

### DIFF
--- a/torchaudio/datasets/utils.py
+++ b/torchaudio/datasets/utils.py
@@ -17,6 +17,8 @@ import urllib.request
 from torch.utils.data import Dataset
 from torch.utils.model_zoo import tqdm
 
+_LG = logging.getLogger(__name__)
+
 
 def unicode_csv_reader(unicode_csv_data: TextIOWrapper, **kwargs: Any) -> Any:
     r"""Since the standard csv library does not handle unicode in Python 2, we need a wrapper.
@@ -214,14 +216,14 @@ def extract_archive(from_path: str, to_path: Optional[str] = None, overwrite: bo
 
     try:
         with tarfile.open(from_path, "r") as tar:
-            logging.info("Opened tar file {}.".format(from_path))
+            _LG.info("Opened tar file %s.", from_path)
             files = []
             for file_ in tar:  # type: Any
                 file_path = os.path.join(to_path, file_.name)
                 if file_.isfile():
                     files.append(file_path)
                     if os.path.exists(file_path):
-                        logging.info("{} already extracted.".format(file_path))
+                        _LG.info("%s already extracted.", file_path)
                         if not overwrite:
                             continue
                 tar.extract(file_, to_path)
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/utils.py
```diff
@@ -214,14 +216,14 @@ def extract_archive(from_path: str, to_path: Optional[str] = None, overwrite: bo
 
     try:
         with tarfile.open(from_path, "r") as tar:
-            logging.info("Opened tar file {}.".format(from_path))
+            _LG.info("Opened tar file %s.", from_path)
```
vincentqb, line 219: This should work too: ` _LG.info("Opened tar file {}.", from_path)`

mthrok:
> Shouldn't we just get rid of the logging though? We don't anywhere else.

It is true that the rest of torch audio (and PyTorch, except old Caffe2 portions) does not use logging, but I do not think that this fact is the good justification for us to make decision on getting rid of loggings. Proper logging always helps software development.

For this PR, I did not really give thoughts about the these logging mechanisms used in `dataset/utils` module. I just looked at the code and noticed the unit-patterns so I fixed them.

(the following does not concern this PR.)

Having said that, my personal experience is that it's good to have logging on data loader side when accessing a file on disk. When working on large scale dataset (thousands of files), one of which contains invalid data, and randomly iterating them, it is not easy to track down which file is broken. Having debug logging in place and turning on debug messages help in such case.
Which is also why I think it is a good idea to have logging in backend initialization/finalization as commented [here](https://github.com/pytorch/audio/pull/572#discussion_r413338706). Something happening behind the scene from users, deep down the call stack, can go wrong and if debug logging is available, it is easy to check what is happening.

vincentqb:
BTW, this is a BC breaking change.

vincentqb:
Logging is a topic that is relevant to other domains, and we should make sure we align on how we would want to do it, if at all.

mthrok:
> BTW, this is a BC breaking change.

Can you elaborate why?

vincentqb:
If someone was actually using the logging, now logging is done in a different location. Is that correct?

mthrok:
> If someone was actually using the logging, now logging is done in a different location. Is that correct?

No. Since the torchaudio still does not instantiate a Handler or Filter object, the log will be passed to the Handler of the root logger. Which is practically same as the previous behavior, writing to the root logger directory.
The users who were using `logging` will now see that these logs are coming from `torchaudio.dataset.util` module with `%(levelname)s `, then they gain the ability to filter logs from torchaudio in a clean way.

vincentqb:
> The users who were using `logging` will now see that these logs are coming from `torchaudio.dataset.util` module with `%(levelname)s `, then they gain the ability to filter logs from torchaudio in a clean way.

I see, so what was shown before?

PR #493, opened 2020-04-02
Turn on tests when building through conda-build

peterjc123:
Relates to #425

PR #576, opened 2020-04-23
Run unit test on CCI

mthrok:
This PR adds unit test on Circle CI.

These tests are added as a new workflow so that binary build and unit tests are independent each other.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -243,14 +243,14 @@ jobs:
       - restore_cache:
 
           keys:
-            - env-v1-{{ arch }}-<< parameters.python_version >>-{{ checksum "./packaging/build_from_source.sh" }}-{{ checksum ".circleci/unittest/environment.yml" }}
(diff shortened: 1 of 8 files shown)


cpuhrsch:
I'm currently in the process of setting up the same for torchtext https://github.com/pytorch/text/pull/740 - Let's chat about this if you want or I'll just wait for you to finish the work and then go from there since torchtext needs the same work done.

mthrok:
> I'm currently in the process of setting up the same for torchtext [pytorch/text#740](https://github.com/pytorch/text/pull/740) - Let's chat about this if you want or I'll just wait for you to finish the work and then go from there since torchtext needs the same work done.

Sounds good. This PR is mostly finished and ready for review, so we can chat about it. 🙂

vincentqb:
Do you want to add the unit tests for windows too? see [here](https://github.com/pytorch/audio/pull/493).

PR #578, opened 2020-04-23
Port sox::vad

astaff:
This PR adds Voice Activity Detector (vad) to Transforms to further reduce dependency on sox (see https://github.com/pytorch/audio/issues/260)

Original implementation: https://sourceforge.net/p/sox/code/ci/master/tree/src/vad.c

Notes:
* Although @zou3519 and I were able to get about 5x performance improvement by replacing loops with vector operations in `_measure`, it is still about 2-3x slower than original C implementation.
* ~~Couldn't get parity with sox when `normalization=True` as `torchaudio.load` and `sox` seem to normalize differently. (is that right?)~~
* ~~Need help with handling multi-channel audio correctly.~~
* ~~Some variable names reflect ones in the original code for troubleshooting purposes.~~


torchaudio/transforms.py
```diff
@@ -1,7 +1,8 @@
 # -*- coding: utf-8 -*-
 
 import math
-from typing import Callable, Optional
+from dataclasses import dataclass, InitVar
```
vincentqb, line 4: This is causing python 3.6 to fail, since `dataclasses` are from 3.7 onward.

vincentqb:
Since this is WIP, I suggest prefixing the title with [WIP] so we know when we are ready to consider merging. Github also allows the creation of draft pull request, though I don't see a way to convert an existing pull request to draft mode. Do you?

torchaudio/transforms.py
vincentqb: We should see if the next release will only support 3.7+. If so, then we don't need to worry about this.

vincentqb:
> * Need help with handling multi-channel audio correctly.

Can you give a little more details? Is the current implementation working on a single channel? How is sox handling multiple channel? Is the output meant to be per channel?

torchaudio/transforms.py
astaff: there is a python 3.6 [backport](https://pypi.org/project/dataclasses/) of `dataclasses` which is pretty lightweight, we can use it

astaff:
> In terms of code organization, we put the computation in a functional when possible so that the transform is more of a thin wrapper around this.

Got it. It was my original intent in the first commit, but after I ended up with two classes representing state I moved things to `transforms.py` and merged one into actual module. Do you want classes for holding the state in `functional.py` as well?

astaff:
> Can you give a little more details? Is the current implementation working on a single channel? How is sox handling multiple channel? Is the output meant to be per channel?

I tested current implementation on a single channel. I will confirm sox behavior on multiple channels and report back. Looking at the original C code, I think it triggers on voice activity in any channel and outputs all channels after the trigger.

vincentqb:
> Got it. It was my original intent in the first commit, but after I ended up with two classes representing state I moved things to `transforms.py` and merged one into actual module. Do you want classes for holding the state in `functional.py` as well?

The `functional.py` is for the computation, and `transforms.py` for the state. Is the state necessary to the computation? Is that what you meant?

vincentqb:
> I tested current implementation on a single channel. I will confirm sox behavior on multiple channels and report back. Looking at the original C code, I think it triggers on voice activity in any channel and outputs all channels after the trigger.

If the the VAD works per channel, then it's easy to add batching by simply reshaping the tensor from (batch, channels, time) to (batch * channels, time), and back.

If sox runs on each channel, and takes the union of detected regions over the channels, it may be a good idea to leave the detection per channel, and let the user decide what to do from there.

astaff:
> Is the state necessary to the computation? Is that what you meant?
That's right. I can technically pull the state inside the computation, but I think that will hurt readability.

torchaudio/transforms.py
```diff
@@ -907,3 +909,313 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        pre_trigger_time (float, optional): The amount of audio (in seconds) to preserve
+            before the trigger point and any found quieter/shorter bursts. (Default: 0.0)
+
+    References:
+        http://sox.sourceforge.net/sox.html
+        https://pysox.readthedocs.io/en/latest/api.html#sox.transform.Transformer.vad
+    """
+
+    @dataclass
+    class Channel:
```
vincentqb, line 944: If the information for each channel is held in a class, it will be more difficult to vectorize, no?

astaff: That's right! I am gonna experiment, refactor and report back.

torchaudio/transforms.py
vincentqb: Yeah, that sounds reasonable to me, though I would have preferred avoiding adding a special case for 3.6 :)

torchaudio/transforms.py
vincentqb: > That's right. I can technically pull the state inside the computation, declaring classes inside the function - let's see how it's gonna affect the readability and vectorization.

The state you refer to is the channel class, right? That would be one more reason to avoid implementing it :)

mthrok:
EDIT: Wait, I no longer see anywhere `dataclass` being used. Did you get rid of it?

<del>
I personally like `dataclass`, but I do not think it worth doing to add a new dependency. Here are my thoughts.

- Pros
  - `dataclass` is available

- Cons
  - The backport library
    - could be wrong
    - could break in the place where we do not have control.
  - This adds maintenance cost
    - Implementation
    - CI
      - Packaging
      - Testing
    - Documentation

It is also worth noting that common practice is to use `collections.namedtuple` in place of `dataclass` where Python 3.6 support matters.

Having said that, if we are adding backport module as a new dependencies, the following has to be corrected, in addition to the implementation.

 - Documentation
   This is most important because there is no automated system to tell if the instruction is missing from there but this is where users look at.
 - Update test 
   Unittest for Python 3.6 is failing as the backport is missing both on Travis and Circle CI. https://github.com/pytorch/audio/pull/578/checks?check_run_id=617323351
   The dependency has to be added to test environment. Note that this library, although it claims it does not work with 3.7, however, according to my quick check, `pip install dataclasses` works on Python 3.7 and the library is installed. Therefor ensuring backport being not installed on Python 3.7+ requires careful attention.
 - Update binary build
   Interestingly, not all binary build do not fail. https://github.com/pytorch/audio/pull/578/checks?check_run_id=617323353
   Wheel builds succeed while Conda builds properly fail.
   I think this suggests that Wheel package build does not check dependency list, presumably meaning that the package will fail at run time.

So I see a lot of maintenance overhead in adding backport library, while the benefit does not outweigh the overheads.
</del>

astaff:
> EDIT: Wait, I no longer see anywhere `dataclass` being used. Did you get rid of it?

@mthrok ugh, that's correct. sorry about that. having a state class was mostly the artifact of moving code from C. I ended up refactoring it, after I got numerical parity with SoX.

torchaudio/transforms.py
astaff: Thanks for your feedback @vincentqb and @mthrok. Got rid of `dataclass` and state class as a whole.

vincentqb:
> having a state class was mostly the artifact of moving code from C. I ended up refactoring it, after I got numerical parity with SoX.

Awesome you were able to remove it!

vincentqb:
I also see that you have the jit consistency test now. Marked test as added :)

astaff:
> I also see that you have the jit consistency test now. Marked test as added :)

Batch as well.

test/test_batch_consistency.py
```diff
@@ -83,6 +83,11 @@ def test_sliding_window_cmn(self):
         _test_batch(F.sliding_window_cmn, waveform, center=False, norm_vars=True)
         _test_batch(F.sliding_window_cmn, waveform, center=False, norm_vars=False)
 
+    def test_vad(self):
+        filepath = common_utils.get_asset_path("vad-hello-mono-32000.wav")
+        waveform, _ = torchaudio.load(filepath)
+        _test_batch(F.vad, waveform, sample_rate=32000)
```
vincentqb, line 89: nit: is the `sample_rate=32000` different than what `waveform, sample_rate = torchaudio.load(...)` would return as `sample_rate`?

test/test_torchscript_consistency.py
```diff
@@ -557,6 +557,11 @@ def test_SlidingWindowCmn(self):
         tensor = torch.rand((1000, 10))
         self._assert_consistency(T.SlidingWindowCmn(), tensor)
 
+    def test_Vad(self):
+        filepath = common_utils.get_asset_path("vad-hello-mono-32000.wav")
+        waveform, _ = torchaudio.load(filepath)
+        self._assert_consistency(T.Vad(32000), waveform)
```
vincentqb, line 563: Same about `sample_rate`

torchaudio/functional.py
```diff
@@ -1836,3 +1837,299 @@ def sliding_window_cmn(
...
+    search_time: float = 1.0,
+    allowed_gap: float = 0.25,
+    pre_trigger_time: float = 0.0,
+    # Fine-tuning parameters
+    boot_time: float = .35,
+    noise_up_time: float = .1,
+    noise_down_time: float = .01,
+    noise_reduction_amount: float = 1.35,
+    measure_freq: float = 20.0,
+    measure_duration: Optional[float] = None,  # by default, twice the measurement period; i.e. with overlap.
```
vincentqb, line 1939: nit: adding this comment in the documentation is enough to me

torchaudio/functional.py
```diff
@@ -1836,3 +1837,299 @@ def sliding_window_cmn(
...
+    measure_smooth_time: float = .4,
+    hp_filter_freq: float = 50.,
+    lp_filter_freq: float = 6000.,
+    hp_lifter_freq: float = 150.,
+    lp_lifter_freq: float = 2000.,
```
vincentqb, line 1944: nit: should we set defaults both on the functional and the transform or just the transform?

vincentqb:
Besides the minor point I mentioned, this looks good to me! Can you also add to the documentation in `docs/`?

torchaudio/functional.py
astaff: In the existing codebase there are cases of both. For example, `complex_norm` and `compute_deltas`, have defaults in `functional` and `transforms`. Some are implemented only in transforms (Fade), some are in functional (Overdrive). Personally, I think that the user experience of calling a `transform` should be no different from `functional`, that means both should have defaults.

The question that is out of scope of this Pull Request is: is there a value of synthesizing transform classes out of functions, including docstrings.

PR #581, opened 2020-04-24
Add base test image with Kaldi featbin and codecs

mthrok:
This PR adds Docker image definition that can be used for unittest on Circle CI, which includes Kaldi featbins and codecs.

* Previously the unittest workflow was using `pytorch/manylinux-cuda***` image, following the other jobs in build workflow, but this image is huge (3GB) and contains a lot of tools not used for testing, so I switched to building custom image. See `.circleci/unittest/docker/build_and_push.sh` for managing the image.

* Kaldi featbins are added to the base image in a following manner so as to minimize the final image size.
   - Use Docker multi-stage build to separate Kaldi build process (lots of artifacts) and the final image.
   - Use ATLAS as math library (instead of MKL) which can be separately installed in the final image.
   - Build Kaldi with as shared library so that the size of executables and libraries are small.
   - Only copy Kaldi libraries that are used executables to the final image.

The resulting Kaldi artifacts copied to the final image is 107MB, which is reasonable.
<Details><Summary>The following binaries are available for test</Summary>

```
Executables in /kaldi/bin
  /opt/kaldi/src/featbin/compare-feats
  /opt/kaldi/src/featbin/transform-feats
  /opt/kaldi/src/featbin/subsample-feats
  /opt/kaldi/src/featbin/append-vector-to-feats
  /opt/kaldi/src/featbin/fmpe-sum-accs
  /opt/kaldi/src/featbin/wav-reverberate
  /opt/kaldi/src/featbin/get-full-lda-mat
  /opt/kaldi/src/featbin/compose-transforms
  /opt/kaldi/src/featbin/extract-feature-segments
  /opt/kaldi/src/featbin/splice-feats
  /opt/kaldi/src/featbin/compute-plp-feats
  /opt/kaldi/src/featbin/add-deltas-sdc
  /opt/kaldi/src/featbin/process-kaldi-pitch-feats
  /opt/kaldi/src/featbin/compute-spectrogram-feats
  /opt/kaldi/src/featbin/extract-segments
  /opt/kaldi/src/featbin/wav-to-duration
  /opt/kaldi/src/featbin/shift-feats
  /opt/kaldi/src/featbin/paste-feats
  /opt/kaldi/src/featbin/fmpe-init
  /opt/kaldi/src/featbin/compute-fbank-feats
  /opt/kaldi/src/featbin/process-pitch-feats
  /opt/kaldi/src/featbin/concat-feats
  /opt/kaldi/src/featbin/post-to-feats
  /opt/kaldi/src/featbin/wav-copy
  /opt/kaldi/src/featbin/apply-cmvn
  /opt/kaldi/src/featbin/apply-cmvn-sliding
  /opt/kaldi/src/featbin/copy-feats-to-sphinx
  /opt/kaldi/src/featbin/compute-kaldi-pitch-feats
  /opt/kaldi/src/featbin/compute-mfcc-feats
  /opt/kaldi/src/featbin/copy-feats
  /opt/kaldi/src/featbin/interpolate-pitch
  /opt/kaldi/src/featbin/compute-cmvn-stats
  /opt/kaldi/src/featbin/modify-cmvn-stats
  /opt/kaldi/src/featbin/feat-to-dim
  /opt/kaldi/src/featbin/add-deltas
  /opt/kaldi/src/featbin/subset-feats
  /opt/kaldi/src/featbin/fmpe-est
  /opt/kaldi/src/featbin/extend-transform-dim
  /opt/kaldi/src/featbin/append-post-to-feats
  /opt/kaldi/src/featbin/compute-cmvn-stats-two-channel
  /opt/kaldi/src/featbin/copy-feats-to-htk
  /opt/kaldi/src/featbin/feat-to-len
  /opt/kaldi/src/featbin/select-feats
  /opt/kaldi/src/featbin/compute-and-process-kaldi-pitch-feats
  /opt/kaldi/src/featbin/fmpe-acc-stats
  /opt/kaldi/src/featbin/fmpe-apply-transform
Libraries in /kaldi/lib
  /opt/kaldi/src/lib/libkaldi-base.so
  /opt/kaldi/src/lib/libkaldi-feat.so
  /opt/kaldi/src/lib/libkaldi-gmm.so
  /opt/kaldi/src/lib/libkaldi-hmm.so
  /opt/kaldi/src/lib/libkaldi-matrix.so
  /opt/kaldi/src/lib/libkaldi-transform.so
  /opt/kaldi/src/lib/libkaldi-tree.so
  /opt/kaldi/src/lib/libkaldi-util.so
  /opt/kaldi/tools/openfst-1.6.7/lib/libfst.so.10
    Linking /kaldi/lib/libfst.so.10.0.0 -> /kaldi/lib/libfst.so.10
```

</Details>

Also this PR includes some tweak on comments which did not make its way in #576


mthrok:
@seemethere When you have time, could you build the image and push it to ECR? I would like to run CI using the resulting image but I do not have credential to push the image.

.circleci/config.yml
```diff
@@ -263,6 +263,31 @@ jobs:
...
+          name: upload docker image
+          no_output_timeout: "1h"
+          command: |
+            set +x
+            export AWS_ACCESS_KEY_ID=${ECR_AWS_ACCESS_KEY}
+            export AWS_SECRET_ACCESS_KEY=${ECR_AWS_SECRET_ACCESS_KEY}
+            eval $(aws ecr get-login --region us-east-1 --no-include-email)
+            set -x
+            docker tag ${image_name}:${CIRCLE_WORKFLOW_ID} 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
+            docker push 308535385114.dkr.ecr.us-east-1.amazonaws.com/${image_name}:${CIRCLE_WORKFLOW_ID}
```
seemethere, line 289: I actually think it'd be better to publish these to Docker Hub instead of ECR.

Management is easier there for us, i.e. creating new images, introspection into tags, etc.

WDYT @ezyang?

ezyang:
@seemethere and I talked about this today in 1:1 and we think it may be easier for you to push images yourself to DockerHub for now. Essentially the problem is we want to eventually setup a system that can automatically push Docker images to ECR (so it's not something seemethere has to manually do pushes for), but we're working on some infrastructure to make this better, so a half way solution would make it harder for us to put the infrastructure in later. But if you just upload images your self to DockerHub, that will be simple (as long as you only need one config) and will reduce the work we need to do later. Does that sound good?

mthrok:
> we want to eventually setup a system that can automatically push Docker images to ECR, 

I think the current solution already does this.
The current smoke test job is automatically pushing Docker images to ECR. For example [this](https://app.circleci.com/pipelines/github/pytorch/audio/1395/workflows/b411581b-2dbe-4e5f-9799-9663d6af70b5/jobs/37283) is the previous push from the last Sunday. 
And since this new job is added to the same workflow, trigger on Sundays, it should work in the same manner.

>  (so it's not something seemethere has to manually do pushes for)

I asked @seemethere to push once, only because I wanted to verify that the image works as expected from other PRs being blocked by this.

> @seemethere and I talked about this today in 1:1 and we think it may be easier for you to push images yourself to DockerHub for now. 

Yes I agree, assuming that your are talking about publishing it under `pytorch`, not my personal account.

> but we're working on some infrastructure to make this better, so a half way solution would make it harder for us to put the infrastructure in later. But if you just upload images your self to DockerHub, that will be simple (as long as you only need one config) and will reduce the work we need to do later. Does that sound good?

As long as the image becomes available to CircleCI's job, it is good for me.
I think by keeping the jobs that push Docker image to ECR (smoke test docker build and the new job) same, migration to a new solution later would not be hard, because it applies same changes to these jobs.

ezyang:
> I think the current solution already does this.

Not completely. There's other infrastructure you have to setup; for example, you need to garbage collect old images so that you don't run out of space on ECR (and make sure you whitelist the live images). You also need to document how to manually trigger docker image builds when an update is necessary. It's fairly involved, and even if we do set everything up for torchaudio, we still have to keep it up to date whenever there are updates in the main repo. I think you it is technically possible for you to set it up, but if something easy will unblock you for now, I'd prefer to do that (so that we can get all the repos on the same ground later.)

> Yes I agree, assuming that your are talking about publishing it under pytorch, not my personal account.

Yeah, that's probably better. @soumith should be able to set you up with access.

mthrok:
@ezyang @seemethere 

I update the PR to use the newly created docker repo and removed job definition to push to ECR.
As I discussed with @seemethere, once the docker repo credentials for CI is created, I will add the automated push job.

mthrok:
@seemethere Could you review this, please?

.circleci/unittest/docker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+################################################################################
+FROM ubuntu:18.04
+RUN apt update && apt install -y \
+        g++ \
+        gfortran \
+        git \
+        libatlas3-base \
+        wget \
+    && rm -rf /var/lib/apt/lists/*
+COPY --from=builder /kaldi /kaldi
```
seemethere, line 56: Nice use of multistage builds here :ok:

mthrok:
Thanks!

PR #558, opened 2020-04-17
Add dcshift to functional

bhargavkathivarapu:
Hi ,
Regarding the issue #260 ( Reducing dependency in Sox)
Implemented ```dcshift``` function in functional.py

-  function.rst  - updated with new fucntion
- test_sox_compatibility.py - Added two tests for dcshift and changes as per #557 . Tests passed
```
..................gain: can't reclaim headroom
.
----------------------------------------------------------------------
Ran 19 tests in 475.924s

OK
```
- test_batch_consistency.py - added test for dcshift 
```
.................
----------------------------------------------------------------------
Ran 17 tests in 204.327s

OK
```
- test_torchscript_consistency.py - added test but giving error, couldn't understand the issue.
Below is the log 
```
$ python test/test_torchscript_consistency.py 
.............E...................sssssssssssssssssssssssssssssssss...............sssssssssssssss
======================================================================
ERROR: test_dcshift (__main__.TestFunctionalCPU)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_torchscript_consistency.py", line 428, in test_dcshift
    self._assert_consistency(func, waveform)
  File "test/test_torchscript_consistency.py", line 38, in _assert_consistency
    return _assert_functional_consistency(func, tensor, self.device, shape_only=shape_only)
  File "test/test_torchscript_consistency.py", line 14, in _assert_functional_consistency
    ts_func = torch.jit.script(func)
  File "/Users/bhargav/pytorch/torch/jit/__init__.py", line 1296, in script
    fn = torch._C._jit_script_compile(qualified_name, ast, _rcb, get_default_args(obj))
  File "/Users/bhargav/pytorch/torch/jit/_recursive.py", line 559, in try_compile_fn
    return torch.jit.script(fn, _rcb=rcb)
  File "/Users/bhargav/pytorch/torch/jit/__init__.py", line 1293, in script
    ast = get_jit_def(obj)
  File "/Users/bhargav/pytorch/torch/jit/frontend.py", line 173, in get_jit_def
    return build_def(ctx, py_ast.body[0], type_line, self_name)
  File "/Users/bhargav/pytorch/torch/jit/frontend.py", line 206, in build_def
    build_stmts(ctx, body))
  File "/Users/bhargav/pytorch/torch/jit/frontend.py", line 129, in build_stmts
    stmts = [build_stmt(ctx, s) for s in stmts]
  File "/Users/bhargav/pytorch/torch/jit/frontend.py", line 129, in <listcomp>
[...]

PR #589, opened 2020-04-26
add slaney normalization

### LINKED ISSUE #287: amplitude normalization in create_fb_matrix

vincentqb:
Fixes #287

### DIFF
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -99,7 +99,7 @@ class _FunctionalTestMixin:
             f_max = 20.0
             n_mels = 10
             sample_rate = 16000
-            norm = None
+            norm = ""
             return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate, norm)
 
         dummy = torch.zeros(1, 1)
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -99,7 +99,8 @@ class _FunctionalTestMixin:
             f_max = 20.0
             n_mels = 10
             sample_rate = 16000
-            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate)
+            norm = None
+            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate, norm)
 
         dummy = torch.zeros(1, 1)
         self._assert_consistency(func, dummy)
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -467,7 +467,7 @@ def create_fb_matrix(
 
     if str(norm) == "slaney":
         # Slaney-style mel is scaled to be approx constant energy per channel
-        enorm = 2.0 / (f_pts[2 : n_mels + 2] - f_pts[:n_mels])
+        enorm = 2.0 / (f_pts[2:n_mels + 2] - f_pts[:n_mels])
         fb *= enorm.unsqueeze(0)
 
     return fb
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -424,7 +424,8 @@ def create_fb_matrix(
         f_min: float,
         f_max: float,
         n_mels: int,
-        sample_rate: int
+        sample_rate: int,
+        norm: Optional[str] = None,
 ) -> Tensor:
     r"""Create a frequency bin conversion matrix.
 
@@ -434,6 +435,8 @@ def create_fb_matrix(
         f_max (float): Maximum frequency (Hz)
         n_mels (int): Number of mel filterbanks
         sample_rate (int): Sample rate of the audio waveform
+        norm (Optional[str]): If 'slaney', divide the triangular mel weights by the width of the mel band
+        (area normalization). (Default: None)
 
     Returns:
         Tensor: Triangular filter banks (fb matrix) of size (``n_freqs``, ``n_mels``)
@@ -461,6 +464,12 @@ def create_fb_matrix(
     down_slopes = (-1.0 * slopes[:, :-2]) / f_diff[:-1]  # (n_freqs, n_mels)
     up_slopes = slopes[:, 2:] / f_diff[1:]  # (n_freqs, n_mels)
     fb = torch.max(zero, torch.min(down_slopes, up_slopes))
+
+    if norm == "slaney":
+        # Slaney-style mel is scaled to be approx constant energy per channel
+        enorm = 2.0 / (f_pts[2 : n_mels + 2] - f_pts[:n_mels])
+        fb *= enorm.unsqueeze(0)
+
     return fb
 
 
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -465,7 +465,7 @@ def create_fb_matrix(
     up_slopes = slopes[:, 2:] / f_diff[1:]  # (n_freqs, n_mels)
     fb = torch.max(zero, torch.min(down_slopes, up_slopes))
 
-    if norm == "slaney":
+    if str(norm) == "slaney":
         # Slaney-style mel is scaled to be approx constant energy per channel
         enorm = 2.0 / (f_pts[2 : n_mels + 2] - f_pts[:n_mels])
         fb *= enorm.unsqueeze(0)
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -425,7 +425,7 @@ def create_fb_matrix(
         f_max: float,
         n_mels: int,
         sample_rate: int,
-        norm: Optional[str] = None,
+        norm: str = "",
 ) -> Tensor:
     r"""Create a frequency bin conversion matrix.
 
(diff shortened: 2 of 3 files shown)


torchaudio/functional.py
```diff
@@ -430,6 +431,8 @@ def create_fb_matrix(
         f_max (float): Maximum frequency (Hz)
         n_mels (int): Number of mel filterbanks
         sample_rate (int): Sample rate of the audio waveform
+        norm (Optional[str]): If 'slaney', divide the triangular mel weights by the width of the mel band
```
cpuhrsch, line 434: Do we expected further normalization schemes? Does it make sense to split these normalizations out into its own layer? Are they useful in other contexts (maybe for volume normalization and such)?

vincentqb: > Do we expected further normalization schemes? 

Yes, we could, see librosa/librosa#1050. Librosa itself is in process of adding other normalization as mentioned in this pull request.

> Does it make sense to split these normalizations out into its own layer? 

Not according to [this comment](https://github.com/pytorch/audio/pull/426#discussion_r375998053).

> Are they useful in other contexts (maybe for volume normalization and such)?

The normalization is done against `f_pts` which is computed within `create_fb_matrix`. I'm not aware of other use case.

test/test_torchscript_consistency.py
```diff
@@ -99,7 +99,8 @@ def func(_):
             f_max = 20.0
             n_mels = 10
             sample_rate = 16000
-            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate)
+            norm = None
```
cpuhrsch, line 102: If you're exercising torchscript I'd pass the less trivial type which is a string instead of None.

torchaudio/functional.py
```diff
@@ -461,6 +464,12 @@ def create_fb_matrix(
     down_slopes = (-1.0 * slopes[:, :-2]) / f_diff[:-1]  # (n_freqs, n_mels)
     up_slopes = slopes[:, 2:] / f_diff[1:]  # (n_freqs, n_mels)
     fb = torch.max(zero, torch.min(down_slopes, up_slopes))
+
+    if str(norm) == "slaney":
```
cpuhrsch, line 468: Why the cast? You could raise an exception if the type doesn't match if it's not None.

vincentqb: Without the cast:
```
ERROR: test_create_fb_matrix (__main__.TestFunctionalCPU)
  File "test/test_torchscript_consistency.py", line 106, in test_create_fb_matrix
    self._assert_consistency(func, dummy)
  File "test/test_torchscript_consistency.py", line 38, in _assert_consistency
    return _assert_functional_consistency(func, tensor, self.device, shape_only=shape_only)
  File "test/test_torchscript_consistency.py", line 14, in _assert_functional_consistency
    ts_func = torch.jit.script(func)
  File "/Users/vincentqb/anaconda3/envs/audio-built/lib/python3.7/site-packages/torch/jit/__init__.py", line 1281, in script
    fn = torch._C._jit_script_compile(qualified_name, ast, _rcb, get_default_args(obj))
  File "/Users/vincentqb/anaconda3/envs/audio-built/lib/python3.7/site-packages/torch/jit/_recursive.py", line 555, in try_compile_fn
    return torch.jit.script(fn, _rcb=rcb)
  File "/Users/vincentqb/anaconda3/envs/audio-built/lib/python3.7/site-packages/torch/jit/__init__.py", line 1281, in script
    fn = torch._C._jit_script_compile(qualified_name, ast, _rcb, get_default_args(obj))
RuntimeError: 
Could not cast value of type Optional[str] to bool:
'create_fb_matrix' is being compiled since it was called from 'func'
  File "test/test_torchscript_consistency.py", line 103
            sample_rate = 16000
            norm = None
            return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate, norm)
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE


Ran 1 test in 0.032s

FAILED (errors=1)
```

test/test_torchscript_consistency.py
vincentqb: You mean: default to empty string? We don't use that elsewhere, but sounds good to me.

vincentqb: done

vincentqb:
@mthrok -- do you have any feedback?

torchaudio/functional.py
```diff
@@ -424,7 +424,8 @@ def create_fb_matrix(
         f_min: float,
         f_max: float,
         n_mels: int,
-        sample_rate: int
+        sample_rate: int,
+        norm: str = "",
```
mthrok, line 428: As a public API signature, I think `Optional[str]` looks cleaner.

vincentqb: This was made due to [comment](https://github.com/pytorch/audio/pull/589#discussion_r420324932). I'll leave it as it is for now. We can always extend the `str` to `Optional[str]` without BC breaking later :)

mthrok: It looks to me that, that comment was meant for the type of variable to pass when running Torchscript test, not about the function signature.

vincentqb: If we allow `None` in the signature, then the code should work with/without jit when passing `None`. It wasn't though. Is that what you meant?

mthrok: > If we allow None in the signature, then the code should work with/without jit when passing None.
Yes, it works.

```
from typing import Optional

import torch
from torch import Tensor


def bar(foo: Optional[str]=None) -> Tensor:
    if foo is None:
        return torch.zeros(1, 2)
    if foo == "a":
        return torch.ones(1, 1)

    return torch.empty(1, 1)


ts_bar = torch.jit.script(bar)

for v in [None, "a", "b"]:
    print(bar(v))
    print(ts_bar(v))
```

produces

```
None
tensor([[0., 0.]])
tensor([[0., 0.]])
a
tensor([[1.]])
tensor([[1.]])
b
tensor([[-2.8910e+12]])
tensor([[0.]])
```

also `dcshift` uses `Optional[float]` and it works fine. for both None and float input for torchscript. 
https://github.com/pytorch/audio/pull/558

mthrok: It's just when type is optional, it firstly needs to compare against `None` using `if var is None` or `if var is not None`.

mthrok: https://pytorch.org/docs/master/jit_language_reference.html#optional-type-refinement

vincentqb: Alrighty, #641 :)

PR #599, opened 2020-04-30
Update audio samples for vad testing

astaff:
This PR re-uses sample `test/assets/SpeechCommands/speech_commands_v0.02/go/0a9f9af7_nohash_0.wav` to produce stereo and mono samples for vad.


mthrok:
Although these files are small, if data are coming from existing file, I prefer to generate such data on the fly during the test. If we keep adding test assets, the size of repository increases, and that could be problematic at some point for downstream users.

astaff:
> Although these files are small, if data are coming from existing file, I prefer to generate such data on the fly during the test. If we keep adding test assets, the size of repository increases, and that could be problematic at some point for downstream users.

Agreed. Maybe this is something that can be mentioned in how to contribute along with the right way to manage temporary files in the test suite. unittest and pytest have different approaches, as far as I understand.

jeffreyksmithjr:
It also seems possible to maintain a small library of test data, if we think we can keep cover all uses cases with a sufficiently small collection.

Alternatively, pulling down datasets from a network location would be another option, but it introduces other complications that make unit tests more like integration tests.

PR #603, opened 2020-05-01
Tweak CCI config

mthrok:
Importing the CCI tweak I found useful while working on https://github.com/pytorch/text/pull/748 https://github.com/pytorch/text/pull/744

 - bust conda cache weekly
 - ~~rename utility directory `test` -> `utility`~~
 - Show slowest 20 tests in log `--durations 20`


vincentqb:
The cache is for conda install packages, right?

vincentqb:
Should the name of the file be changed as well?
`.circleci/test/test_sort_yaml.py → .circleci/utils/test_sort_yaml.py`

mthrok:
> The cache is for conda install packages, right?

Yes

> Should the name of the file be changed as well?
> `.circleci/test/test_sort_yaml.py` → `.circleci/utils/test_sort_yaml.py`

The thing is, I do not know what this script is for and I think it's obsolete. If we care as much I suggest we delete it.

> Are you planning to keep the config files in sync between domains?

I don't think that's possible. Torchtext has specific caching which does not apply to other domains.
But these changes felt "good practice to have" in torchaudio too.

> Oops, I meant to say: tests are failling :/

Weird, rebuilding test image and will retry.

vincentqb:
> The thing is, I do not know what this script is for and I think it's obsolete. If we care as much I suggest we delete it.

How about we don't modify it as part of this PR, and we simply delete it in the next? :) If we don't see it used anywhere, the risk is low.

PR #593, opened 2020-04-28
enable mel_scale option

vincentqb:
Add `mel_scale` option enabling the htk or slaney mel scale, see [here](https://github.com/pytorch/audio/pull/426#discussion_r375998053) for discussion and [librosa](https://librosa.github.io/librosa/_modules/librosa/core/time_frequency.html#mel_frequencies).

Relates #589, [internal](https://www.internalfb.com/intern/diff/D20895257/), (#608?), (#259 for mel-hz functional)

PR #608, opened 2020-05-04
Reduce tolerance for MFCC

mthrok:
`test_batch_mfcc` fails not only on Windows but also on [GPU linux machine](https://app.circleci.com/pipelines/github/pytorch/audio/1628/workflows/9c945119-bd14-40a4-9813-6d1f0fd764bc/jobs/41720).

I suspect this is resulted from the different version of CPU architecture ( `amd64-6_85` vs GPU `amd64-6_79` Windows `amd64-1_`) but cannot be sure. 

Looking at the error rate, more than 99.9% of elements are within the range of `1e-5` so I think it's safe to reduce it to `1e-4` so that Windows and GPU can pass this test.

### DIFF
--- a/test/test_batch_consistency.py
+++ b/test/test_batch_consistency.py
@@ -207,7 +207,6 @@ class TestTransforms(unittest.TestCase):
         computed = torchaudio.transforms.MelSpectrogram()(waveform.repeat(3, 1, 1))
         torch.testing.assert_allclose(computed, expected)
 
-    @unittest.skipIf(platform.system() == "Windows", "Test is known to fail on Windows.")
     def test_batch_mfcc(self):
         test_filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
         waveform, _ = torchaudio.load(test_filepath)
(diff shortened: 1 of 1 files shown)


vincentqb:
Let me double-check: but this seems to indeed be the issue I run into for #593, and in particular why [float64](https://github.com/pytorch/audio/blob/507eda511e04362ebc0fa959f78dcc9608114570/torchaudio/functional.py#L514) was needed.

mthrok:
I located the cause to `torch.log10` producing the different result. https://github.com/pytorch/pytorch/issues/37833
I think for functions that use `amplitude_to_DB` we can reduce the threshold.

vincentqb:
Thanks for opening the issue! Yup, I confirm `torch.log10` is where the difference was coming from in #593.

mthrok:
I did further digging and with `float64`, it is fine. So we can convert the `dtype` before and passing a Tensor to `torch.log10`.

vincentqb:
Yeah, that's the solution I used [here](https://github.com/pytorch/audio/blob/507eda511e04362ebc0fa959f78dcc9608114570/torchaudio/functional.py#L514).

vincentqb:
@mthrok -- are you suggesting the following?
- [ ] `test_librosa_consistency` can be reactivated
- [ ] tests for `amplitude_to_DB` can be reactivated

mthrok:
> @mthrok -- are you suggesting the following?
> 
> * [ ]  `test_librosa_consistency` can be reactivated
> * [ ]  tests for `amplitude_to_DB` can be reactivated

Yes, if you are fine with that approach, we can do that. I will file an open task for that.

mthrok:
Filed #611

cpuhrsch:
In general it's useful to do a quick check on performance before working with float64 in performance critical sections of this code (I'm assuming that's the case here). It can cause major slowdowns.

PR #612, opened 2020-05-06
Fix docstring of masking behavior

### LINKED ISSUE #555: iid_masks in Frequency and Time Masking
## 📚 Documentation of iid_masks Parameter in AxisMasking

Description of Problem
------
The Docstring and Documentation of Frequency and Axis Masking state the following for the 'iid_masks' Paramter:

> _iid_masks (bool): weather to apply the same mask to all
            the examples/channels in the batch. (Default: False)_

This explanation would let me conclude, that the default is using __different__ masks for all spectral representations in the batch.

However, the opposite is the case when using this on a 4D-Tensor (e.g. having a Batch dimension). Looking at the source code, the function _mask_along_axis_ is used. This function does not generate different masks as it also states in it's docstring.

The docstring of _iid_mask_along_axis_ __also__ states, that is is generating the same mask for all spectrograms. However, I am quite confident that this function is generating different masks for each spectrogram.

Concluding:
--------
Behaviour of the paramter is not as expected. _iid_mask_along_axis_ shoud be used instead of _mask_along_axis_.

Looking at the code I think at the moment it is not possible to create different masks for 3d tensors (e.g. without batch dimension).

Suggested Solution:
-------
* Fix Documentation that usage of Paramter is inversed
* Document that different masks only work for a tensor with batch dimension (not 100 % sure if I am right here)

mthrok:
Resolves #555
cc. @Ektoplasmakugel

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -779,6 +779,7 @@ class _AxisMasking(torch.nn.Module):
         mask_param (int): Maximum possible length of the mask.
         axis (int): What dimension the mask is applied on.
         iid_masks (bool): Applies iid masks to each of the examples in the batch dimension.
+            This option is applicable only when the input tensor is 4D.
     """
     __constants__ = ['mask_param', 'axis', 'iid_masks']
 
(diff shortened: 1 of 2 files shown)


torchaudio/transforms.py
```diff
@@ -798,7 +799,6 @@ def forward(self, specgram: Tensor, mask_value: float = 0.) -> Tensor:
         Returns:
             Tensor: Masked spectrogram of dimensions (..., freq, time).
         """
-
```
vincentqb, line 801: nit: in terms of cosmetic, you prefer without the space here? :)

mthrok: Yeah. I forgot which language it was about but one of the Google style guides also discourages it for the reason not contributing to readability.

Issue #613, opened 2020-05-06
torchaudio.compliance.kaldi.fbank does NOT support GPU


csukuangfj:
The readme.md says

> The aim of torchaudio is to apply PyTorch to the audio domain. By supporting PyTorch, torchaudio follows the same philosophy of providing strong GPU acceleration,

**However**, `strong GPU acceleration` is not available for  `torchaudio.compliance.kaldi.fbank`.

-------

The following test code

```python
    filename = '/path/to/test.wav'
    waveform, sample_rate = torchaudio.load(filename)
    device = torch.device('cuda', 0)
    waveform = waveform.to(device)

    fbank = torchaudio.compliance.kaldi.fbank(waveform)
```

produces the following error:

```
Traceback (most recent call last):
  File "./fbank_test.py", line 27, in <module>
    main()
  File "./fbank_test.py", line 23, in main
    fbank = torchaudio.compliance.kaldi.fbank(waveform)
  File "/xxxxxx/py35/lib/python3.5/site-packages/torchaudio/compliance/kaldi.py", 
line 554, in fbank
    snip_edges, raw_energy, energy_floor, dither, remove_dc_offset, preemphasis_coefficient)
  File "/xxxxx/py35/lib/python3.5/site-packages/torchaudio/compliance/kaldi.py", 
line 180, in _get_window
    signal_log_energy = _get_log_energy(strided_input, EPSILON, energy_floor)  # size (m)
  File "/xxxxx/py35/lib/python3.5/site-packages/torchaudio/compliance/kaldi.py", 
line 113, in _get_log_energy
    log_energy = torch.max(strided_input.pow(2).sum(1), epsilon).log()  # size (m)
RuntimeError: iter.device(arg).is_cuda() INTERNAL ASSERT FAILED at /pytorch/aten/src/ATen/native/cuda/Loops.cuh:56, 
please report a bug to PyTorch.
```

pytorchaudito is installed using 
```
pip install torchaudio
```

```python
    print(torch.version.__version__)
    print(torch.version.git_version)
    print(torchaudio.version.__version__)
    print(torchaudio.version.git_version)
```
produces

```
1.5.0
4ff3872a2099993bf7e8c588f7182f3df777205b
0.5.0
3305d5c2f935b1512c58faad0aa770432dea6d21
```

mthrok:
I could reproduce it with `'/scratch/moto/torchaudio/test/assets/kaldi_file_8000.wav'`.

<Details><Summary>Env</Summary>
[...]

PR #616, opened 2020-05-07
Introduce common utility for defining test matrix for device/dtype

mthrok:
To work on #613, I added a helper function which simplifies the logic to generate test suites for different device, which also extends it to dtype.
As a result, torchscript consistency test is performed on (cpu, cuda) x (float32, float64) matrix. (float64 is newly added)

1. This revealed that some transforms (`Resample`) is not compatible with float64, which is fixed in this PR.
1. Interestingly `lfilter` does not yield consistent result among Python and torchscript only when `float64`.
    I looked at it but do not have an immediate fix so I marked them as expected to fail.

### DIFF
--- a/test/test_functional.py
+++ b/test/test_functional.py
@@ -9,10 +9,7 @@ import pytest
 import common_utils
 
 
-class _LfilterMixin:
-    device = None
-    dtype = None
-
+class Lfilter(common_utils.TestBaseMixin):
     def test_simple(self):
         """
         Create a very basic signal,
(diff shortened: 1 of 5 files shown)


test/test_torchscript_consistency.py
```diff
@@ -578,27 +617,4 @@ def test_Vad(self):
...
-
-@unittest.skipIf(not torch.cuda.is_available(), 'CUDA not available')
-class TestTransformsCUDA(_TransformsTestMixin, unittest.TestCase):
-    """Test suite for Transforms module on GPU"""
-    device = torch.device('cuda')
-
-
-if __name__ == '__main__':
-    unittest.main()
+common_utils.define_test_suites(globals(), [Functional, Transforms])
```
vincentqb, line 620: alright, looks like pytorch is passing `globals()`

test/test_functional.py
```diff
@@ -33,25 +30,12 @@ def test_clamp(self):
         b_coeffs = torch.tensor([1, 0], dtype=self.dtype, device=self.device)
         a_coeffs = torch.tensor([1, -0.95], dtype=self.dtype, device=self.device)
         output_signal = F.lfilter(input_signal, a_coeffs, b_coeffs, clamp=True)
-        self.assertTrue(output_signal.max() <= 1)
+        assert output_signal.max() <= 1
```
vincentqb, line 33: nit: changing `self.assertTrue` to `assert`? oh that's because the class isn't inheriting from `unittest`. I see.

test/test_torchscript_consistency.py
```diff
@@ -578,27 +617,4 @@ def test_Vad(self):
...
-
-
-@unittest.skipIf(not torch.cuda.is_available(), 'CUDA not available')
-class TestTransformsCUDA(_TransformsTestMixin, unittest.TestCase):
-    """Test suite for Transforms module on GPU"""
-    device = torch.device('cuda')
-
-
-if __name__ == '__main__':
-    unittest.main()
```
vincentqb, line 604: removing unittest is good to me.

cpuhrsch:
I'm really not a fan of not inheriting from PyTorch's unittest class. Sure we can argue that some designs such as overwriting self.assertEqual aren't perfect, but the entire system relies on it and we should comply with whatever the core library defines as "equal". Otherwise we stand to fail to test for new developments such as type upcasting or new Tensor properties.

mthrok:
> I'm really not a fan of not inheriting from PyTorch's unittest class. Sure we can argue that some designs such as overwriting self.assertEqual aren't perfect, but the entire system relies on it and we should comply with whatever the core library defines as "equal". Otherwise we stand to fail to test for new developments such as type upcasting or new Tensor properties.

offline discussion: I will look into this.

vincentqb:
> I'm really not a fan of not inheriting from PyTorch's unittest class. 

I agree. I suggest we revert the PR, and redo each item separately.

@mthrok -- Can you elaborate on the action items you have discussed?

mthrok:
> > I'm really not a fan of not inheriting from PyTorch's unittest class.
> 
> I agree. I suggest we revert the PR, and redo each item separately.
> 

Reverting this will adds more work for the same outcome so I would not.
Things I am working on
 - Replace `pytest` framework with `unittest`
 - Replace `torch.assert_allclose` with `assertEqual` from PyTorch''s test framework.

PR #617, opened 2020-05-08
Add synth

tomassosorio:
relates to: #260
Test
- batch test
- soxeffects test
- torchscript consistency

Transform Synth to synthesize waves and noise
Due to not having GPU on development, I am not sure if it torchscript works using GPU
Waveforms results


torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+    Args:
+        sample_rate (int, optional): Sample rate of audio signal (Default: ``16000``).
+        duration (float, optional): Duration in seconds of audio to synthesise (Default: ``1.0``).
+        wave_type (str, optional): is one of sine, square, triangle, sawtooth, trapezium, exp, white noise
+         and brownian noise (Default: ``"sine"``).
+        freq (int or tuple, optional): are the frequencies at the beginning/end of synthesis in Hz (Default: ``440``).
+        chirp (str, optional): If `freq` defined the variation between to frequencies can be changed `linear`,
+            `square` or `exp` (Default: ```linear``).
+        amp (float, optional): Maximum amplitude (Default: ```1.0``).
+    """
```
mthrok, line 1043: At one glance this class is not device/dtype-aware.
Could you rebase to the latest master and see if CI tests pass?
#616 added separate test for float32/float64.

mthrok:
Hi

Thanks for the PR.
Can you generate the asset on the fly, using spicy, as a Tensor then write it to a temporary file?
We would like to refrain from adding a new asset as much as possible.

tomassosorio:
> Hi
> 
> Thanks for the PR.
> Can you generate the asset on the fly, using spicy, as a Tensor then write it to a temporary file?
> We would like to refrain from adding a new asset as much as possible.

Yes, it makes much more sense, I will do it :)

torchaudio/transforms.py
tomassosorio: It passed without warning, so I will assume that is not necessary to do anything

torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
             hp_lifter_freq=self.hp_lifter_freq,
             lp_lifter_freq=self.lp_lifter_freq,
         )
+
+
+class Synth(torch.nn.Module):
```
cpuhrsch, line 1030: I'd consider splitting this into separate transforms and functionals.

We don't have this kind of setup in torchaudio (monothlic classes with factory methods). Also, not all arguments that are passed to the constructor are necessary for each of the synths (for example square or sin).

torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+
+    def evaluate_exp(self, ts):
+        freqs = torch.logspace(math.log10(self.freq[0]), math.log10(self.freq[1]), len(ts) - 1)
+        return self.evaluate(ts, freqs)
+
+    def evaluate_square(self, ts):
+        freqs = torch.pow(torch.linspace(0, 1, len(ts) - 1), 2) * (self.freq[1] - self.freq[0]) + self.freq[0]
+        return self.evaluate(ts, freqs)
+
+    def sine(self, phases):
```
cpuhrsch, line 1118: Is the suggested usecase here the following?

```
s = Synth(...)
phases = s. phases()
result = s.sine(phases)
```

torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+    def phases(self):
+        ts = torch.arange(int(self.duration * self.sample_rate), dtype=torch.float) / self.sample_rate
+        if self.chirp == "linear":
+            return self.evaluate_linear(ts)
+        elif self.chirp == "square":
+            return self.evaluate_square(ts)
+        else:
+            return self.evaluate_exp(ts)
+
+    def wave(self, evaluate):
```
cpuhrsch, line 1083: You could also do ```getattr(self, self.wave_type)```

torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+
+
+class Synth(torch.nn.Module):
+    r"""This effect can be used to generate fixed or swept frequency audio tones with various wave shapes,
+     or to generate noise.
+
+    Args:
+        sample_rate (int, optional): Sample rate of audio signal (Default: ``16000``).
+        duration (float, optional): Duration in seconds of audio to synthesise (Default: ``1.0``).
+        wave_type (str, optional): is one of sine, square, triangle, sawtooth, trapezium, exp, white noise
```
cpuhrsch, line 1037: If the user initializes this with ```wave_type``` why then also have these as separate, public methods? Maybe you meant to make them private (mark with a prefix "_")?

torchaudio/transforms.py
```diff
@@ -1025,3 +1025,127 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+         and brownian noise (Default: ``"sine"``).
+        freq (int or tuple, optional): are the frequencies at the beginning/end of synthesis in Hz (Default: ``440``).
+        chirp (str, optional): If `freq` defined the variation between to frequencies can be changed `linear`,
+            `square` or `exp` (Default: ```linear``).
+        amp (float, optional): Maximum amplitude (Default: ```1.0``).
+    """
+
+    def __init__(self,
```
cpuhrsch, line 1045: The amount of state carried around seems pretty light, further most of these arguments don't apply to all methods and also the functions could be combined to yield the usecases.

For example we could do:
```sine(linear_freqs(start=50, end=500)(phases(sample_rate=16000, duration=1.0)))```

I think phases and linear_freqs might even be useful as standalone transforms anyway maybe even for other methods. We already have functions that return frequencies for e.g. melspectrogram. We could potentially use this to create transforms with different interesting filters, which is unexpectely beyond the scope of synthesization.

tomassosorio:
I am going to move out of this field, audio/speech, and for now, no longer have time for helping this repo so I will close this PR.

PR #625, opened 2020-05-08
Self-contain codecs library

mthrok:
This PR
 - always include codecs to `_torchaudio.so` so that it does not depend on external SoX.
 - adds `python setup.py clean` command which facilitates clean build.

These changes follow the identical pattern as https://github.com/pytorch/text/pull/755 .

* Note on build time requirements
  - Before this change, it was assumed that SoX was built somewhere in the System. (could be one that was installed by package management system like `apt`, or manually built one)
  - After this change, `curl` and `CMake` are required, so that it can build SoX and codecs.


vincentqb:
> * change the library name to `_torchaudio.so` under `torchaudio` directory

Can that be BC breaking?

mthrok:
> > * change the library name to `_torchaudio.so` under `torchaudio` directory
> 
> Can that be BC breaking?

I do not think so.
 - Python naming convention tells that `_torch_sox.so` is not intended for external use.
 -  We do not have any header file, so it is not possible to link against `_torch_sox.so`.

Added the list of things changed by PR in the description.

cpuhrsch:
I think we should consider breaking this into a few more PRs. Changes like the move of the files or the rename of the library could be separated from the build scripts, etc. and make for a more manageable / revertable changeset.

mthrok:
Sounds good. Will work on.

Moto 🛵

vincentqb:
offline discussion with @mthrok -- how about compilation on the internal systems? e.g. [internal](https://www.internalfb.com/intern/tasks/?t=53444370)

vincentqb:
cc @seemethere for awareness

mthrok:
> offline discussion with @mthrok -- how about compilation on the internal systems? e.g. [internal](https://www.internalfb.com/intern/tasks/?t=53444370)

I added bypass for the internal. In internal, nothing about dependencies changes. This PR cannot address T53444370.

build_tools/setup_helpers/extension.py
```diff
@@ -0,0 +1,111 @@
...
+    ]
+    if _BUILD_DEPS:
+        dirs.append(str(_TP_INSTALL_DIR / 'include'))
+    return dirs
+
+
+def _get_extra_objects():
+    objs = []
+    if _BUILD_DEPS:
+        # NOTE: The order of the library listed bellow matters.
```
vincentqb, line 63: nit: "below" :)

.circleci/config.yml
```diff
@@ -329,18 +329,19 @@ jobs:
       - restore_cache:
 
           keys:
-            - env-v1-linux-{{ arch }}-py<< parameters.python_version >>-{{ checksum ".circleci/unittest/linux/scripts/environment.yml" }}-{{ checksum ".circleci-weekly" }}
```
vincentqb, line 332: nit: `v1` will be deleted soon after merging this PR?

mthrok: Caches on CCI will not be deleted, and we do not have control over them.
We change cache key on weekly basis, but this PR also changes what is cached, so it's safer to change the cache key from the previous one.
That way existing PR can still use the old style cache without rebasing, as long as there is no merge conflict.

PR #627, opened 2020-05-09
Fix log10 for float32

Micheladar:
Changed the log10 code to execute all the computations in float64 and change back to the original tensor at the end of the function.


cpuhrsch:
As a general comment, I think it could be useful to have a very explicit test for this. I.e. some concrete, manually set input that will raise this error. In the future we might want to revisit some transforms for performance reasons and this could be one of them. Numerical precision is something that can be easily impacted by performance optimizations (think reordering of operations).

PR #629, opened 2020-05-12
Add flanger to functional

bhargavkathivarapu:
Hi ,
Regarding the issue ( Reducing dependency in Sox - #260  )
Implemented ```flanger``` function in functional.py

   - function.rst - updated with new function

   - test_sox_compatibility.py - Added four tests for flanger ( Tested if all variations are working fine . we can remove some tests as it adds more time to test run time ) 

   - test_batch_consistency.py - added test for flanger

   - test_torchscript_consistency.py - added test for flanger

All CPU test passed.

@vincentqb or @mthrok  could you please review these changes


torchaudio/functional.py
```diff
@@ -1433,6 +1433,152 @@ def _generate_wave_table(
...
+    output_waveform = torch.zeros_like(waveform, dtype=dtype, device=device)
+
+    delay_buf_pos = 0
+    lfo_pos = 0
+
+    for i in range(waveform.shape[-1]):
+
+        delay_buf_pos = (delay_buf_pos + delay_buf_length - 1) % delay_buf_length
+
+        for c in range(0, n_channels):
```
mthrok, line 1540: Correct me if I am wrong, it seems that the same operation with different parameters is applied to each channel individually. If operations among channels do not interweave, can we vectorize the operation?

Looking at the test result, the test of this function takes the longest.
https://app.circleci.com/pipelines/github/pytorch/audio/1768/workflows/4a2bae91-b123-4fcd-af13-a0a24c929c1e/jobs/44621

bhargavkathivarapu: @mthrok , I have vectorized the channels loop , But facing issue in torch script test , I am getting below error 
```
RuntimeError: 
E           Arguments for call are not valid.
E           The following variants are available:
E             
E             aten::frac(Tensor self) -> (Tensor):
E             Expected a value of type 'Tensor' for argument 'self' but instead found type 'float'.
E             
E             aten::frac.out(Tensor self, *, Tensor(a!) out) -> (Tensor(a!)):
E             Expected a value of type 'Tensor' for argument 'self' but instead found type 'float'.
E           
E           The original call is:
E             File "/Users/bhargav/audio/torchaudio/functional.py", line 1542
E                   cur_channel_phase = (torch.arange(0, n_channels) * lfo_length * channel_phase + .5).to(torch.int64)
E                   delay = lfo[(lfo_pos + cur_channel_phase) % lfo_length]
E                   frac_delay = torch.frac(delay)
E                                ~~~~~~~~~~ <--- HERE
E                   delay = torch.floor(delay)
E           'flanger' is being compiled since it was called from 'func'
E             File "/Users/bhargav/audio/test/test_torchscript_consistency.py", line 542
E                       phase = 60.
E                       sample_rate = 44100
E                       return F.flanger(tensor, sample_rate, delay, depth, regen, width, speed,
E                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~...  <--- HERE
E                                        phase, sinusoidal=True, linear_interpolation=True)
```

The delay variable is a tensor, But the torch script is saying it got float instead of tensor

mthrok: `delay` is used in function signature and annotated as `float`, so JIT compiler considers it float but you are also re-defining variable `delay` as Tensor.

Change the variable name to something like `delay_tensor = lfo[(lfo_pos + cur_channel_phase) % lfo_length]`, and it should work.

bhargavkathivarapu: @mthrok , Fixed it , Torch script test passed

mthrok: @bhargavkathivarapu 
Thanks. Can you rebase? In #644, I moved the test file.

Also, in Torchscript test and batch consistency test, instead of loading sample value from a file, can you generate input value, with fixed random seed?

bhargavkathivarapu: @mthrok , I have messed up rebasing ,  I am closing this pull request . Made a new pull request  with suggested changes( #651 ) , please check that .

PR #630, opened 2020-05-12
[BC Breaking] Reorganize C++ resources

mthrok:
Breakdown of #625 

This PR
 - Changes the location of C++ codes from
     - from `torchaudio/torch_sox.cpp` to `torchaudio/csrc/sox.cpp`
     - from `torchaudio/torch_sox.h` to `torchaudio/csrc/sox.h`
 - Changes the location where the resulting library is placed
     - from `_torch_sox.so` (outside of `torchaudio` module)
        to `torchaudio/_torchaudio.so`.

BTW: I noticed that `torch_sox.h` is not used by extension. We can remove it if we do not intend to make the extension library usable from external.


mthrok:
> LGTM, but travis is failing.
> 
> ```
> ImportError: cannot import name '_torchaudio' from 'torchaudio' 
> ```

Okay I will remove the unit tests

> LGTM, but travis is failing.
> 
> ```
> ImportError: cannot import name '_torchaudio' from 'torchaudio' 
> ```

Ref: https://github.com/pytorch/audio/pull/636

mthrok:
@vincentqb

mthrok:
> Find with me, but please mark as BC breaking.

Done so. Thanks!

Issue #476, opened 2020-03-26
Fix optional test with fairseq


vincentqb:
- [x] As seen [here](https://travis-ci.org/github/pytorch/audio/jobs/667465577?utm_medium=notification&utm_source=github_status), we need to bump the python version used in example test in travis to at least 3.6.
- [x] Fix test
- [x] ~~Disable test~~

[internal doc](https://fb.quip.com/TxAbAzyVBt7t)

tomassosorio:
Looking at for example:
[pytorch text](https://github.com/pytorch/text/tree/master/examples/text_classification), we can see that this show with pytorch and torchtext how to build a text classification, however I do not see the point of this example in the scope of torchaudio since basically just uses the `fairseq`, which already have examples in its repository.

I think would be beneficial to do an example only using the functionalities from torchaudio and pytorch to do an audio classification with SpeechCommands for example. This way really showing what torchaudio is meant for and its utility.

vincentqb:
Thanks for the feedback :) This example was a partnership, and was helpful during the hacktackon last summer.

I agree with you that it'd be nice to have reference pipelines for researchers that also showcase torchaudio's ability more specifically. What kind of example would you suggest with SpeechCommands?

tomassosorio:
Currently, tensorflow has a [tutorial](https://github.com/tensorflow/docs/blob/master/site/en/r1/tutorials/sequences/audio_recognition.md), maybe would be better to use yesno dataset and create a simple classification, as a beginner tutorial, and then for an intermediary tutorial a simple ASR example, using pytorch implemented CTC and SpeechCommands.

vincentqb:
@tomassosorio -- We would indeed welcome new tutorials that uses pytorch, torchaudio, and freely available datasets :) We could add it [here](https://pytorch.org/tutorials/) in https://github.com/pytorch/tutorials. Is that of interest to you?

tomassosorio:
Absolutely, I will start working on that!

vincentqb:
Awesome! you can open a pull request https://github.com/pytorch/tutorials and ping me on it :)

vincentqb:
btw, feel free to open an issue over [there](https://github.com/vincentqb/audio-tutorial/blob/master/PipelineTrain.ipynb) to discuss how we should turn this into a reference implementation

vincentqb:
This test is no longer ran as part of CircleCI.

PR #632, opened 2020-05-12
Example pipeline with wav2letter

vincentqb:
We implement a reference pipeline using wav2letter model to train on librispeech. The structure will be inspired by [torchvision's reference implementation](https://github.com/pytorch/vision/tree/master/references).

As discussed [here](https://github.com/pytorch/audio/issues/476#issuecomment-626933599), this code was initially implemented in this [python script](https://github.com/vincentqb/audio-tutorial/blob/master/PipelineTrain.py) which was converted from [this notebook](https://github.com/vincentqb/audio-tutorial/blob/master/PipelineTrain.ipynb) to be ran using SLURM using [bash script](https://github.com/vincentqb/audio-tutorial/blob/master/runbatch.sh) and `sbatch`.

There are at least a few more things to do:
- [x] Clean code from notebook conversion.
- [x] Replace model implementation by call to torchaudio.
- [x] Remove SLURM-specific code.
- [x] ~~Add an option to activate `torch.autograd.set_detect_anomaly(True)`~~
- [x] Remove ~~Bring back~~ viterbi decoder.
- [x] Make distributed work with 1 GPU per python process?
- [ ] ~~Add 10 ms shift data augmentation~~
- [ ] ~~Publish pre-trained weights~~

Note:
* 4795a72fc52182412f22c20c53a757ef927ea884 removed the modified wav2letter model with dropout and custom hidden units in order to use the exact implementation currently available within torchaudio. This commit also removes the SLURM termination signal capture.
* d8ee1e9f3d91cb01fae8a4d065d27a0d4dff23fe removed the 2-gram viterbi decoder as it was not improving CER, and unused dataclass.

See also [post by assemblyai](https://www.assemblyai.com/blog/end-to-end-speech-recognition-pytorch), and [internal](https://github.com/fairinternal/FAIR_Cluster/blob/master/checkpoint/main.py).
cc @zhangguanheng66  for pytorch/text#767


examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+# Empty CUDA cache
+torch.cuda.empty_cache()
+
+# Profiling performance
+pr = cProfile.Profile()
+pr.enable()
+
+
+# Create argument parser
+parser = argparse.ArgumentParser()
```
cpuhrsch, line 64: I'd move this to the bottom of the file and guard it by ```if __name__ ==  "__main__":```and create a "main" function that accepts args as an object. This way this code won't run on import and can be executed from another script. It's also more canonical

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+    """
+    Wrap a dataset so that, whenever a new item is returned, it is saved to memory.
+    """
+
+    def __init__(self, dataset):
+        self.dataset = dataset
+        self._cache = [None] * len(dataset)
+
+    def __getitem__(self, n):
+        if self._cache[n]:
```
cpuhrsch, line 417: To be pedantic, you could have dataset setup to return "False" (which I know it isn't in this case) and then this check would fail. To be more exact you'd use ```if self._cache[n] is not None:```.

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+
+    def __init__(self, process_datapoint, dataset):
+        self.process_datapoint = process_datapoint
+        self.dataset = dataset
+
+    def __getitem__(self, n):
+        try:
+            item = self.dataset[n]
+            return self.process_datapoint(item)
+        except (FileNotFoundError, RuntimeError):
```
cpuhrsch, line 439: What's the purpose of catching these exception? As in, in what circumstances do you expect to see them in this particular case?

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+        for new_value in self._iter:
+            self._values.append(new_value)
+            yield new_value
+        self._done = True
+
+    def __len__(self):
+        return len(self._iterable)
+
+
+class MapMemoryCache(torch.utils.data.Dataset):
```
cpuhrsch, line 407: This is great! Seems like a pretty generic object that should also be useful for core. It's effectively a readthrough cache backed by RAM similar to how diskcache_iterator is a readthrough cache backed by disk.

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+        data = Processed(process_datapoint, data)
+        # data = diskcache_iterator(data)
+        data = MapMemoryCache(data)
+        return data
+
+    return create("train-clean-100"), create("dev-clean"), None
+    # return create(["train-clean-100", "train-clean-360", "train-other-500"]), create(["dev-clean", "dev-other"]), None
+
+
+def which_set(filename, validation_percentage, testing_percentage):
```
cpuhrsch, line 490: Why is this necessary as opposed to a seeded shuffle + split?

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+    c = None
+
+    for _, label in training:
+        # Count bigrams
+        count = [((a.item(), b.item())) for (a, b) in zip(label, label[1:])]
+        count = Counter(count)
+        if c is None:
+            c = count
+        else:
+            c = c + count
```
cpuhrsch, line 606: I think you could also use update and initialize c as an empty Counter.

Creating the Counter and adding to it over each iteration might be quite slow in comparison.

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+
+    return torch.tensor(outputs).transpose(0, -1), torch.cat(scores)
+
+
+def top_batch_viterbi_decode(tag_sequence: torch.Tensor):
+    output, _ = batch_viterbi_decode(tag_sequence, transitions, top_k=1)
+    return output[:, 0, :]
+
+
+def levenshtein_distance_list(r, h):
```
cpuhrsch, line 718: This seems unused

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+        'optimizer': optimizer.state_dict(),
+        'scheduler': scheduler.state_dict(),
+        'history_training': history_training,
+        'history_validation': history_validation,
+    }, False)
+
+
+with tqdm(total=args.epochs, unit_scale=1, disable=args.distributed) as pbar:
+    for epoch in range(args.start_epoch, args.epochs):
+        torch.cuda.reset_max_memory_allocated()
```
cpuhrsch, line 932: Why is this done at the beginning of each epoch?

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+
+def datasets_librispeech():
+
+    def create(tag):
+
+        if isinstance(tag, str):
+            data = LIBRISPEECH(
+                root, tag, folder_in_archive=folder_in_archive, download=False)
+        else:
+            data = torch.utils.data.ConcatDataset([LIBRISPEECH(
```
cpuhrsch, line 478: This could also be done using ```sum``` since ConcatDataset can be created [via add](https://github.com/pytorch/pytorch/blob/master/torch/utils/data/dataset.py#L27).

cpuhrsch:
Looks pretty good :)

examples/pipeline/wav2letter.py
mthrok: Defining variables in global scope makes it very hard to refactor the code later when this script is found useful. And often times unintended variable shadowing or access to global variable occurs. So it's highly recommended to refactor the code.

`torchvision` does very good job in their reference code. https://github.com/pytorch/vision/blob/master/references/detection/train.py

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,1179 @@
...
+        'state_dict': model.state_dict(),
+        'best_loss': best_loss,
+        'optimizer': optimizer.state_dict(),
+        'scheduler': scheduler.state_dict(),
+        'history_training': history_training,
+        'history_validation': history_validation,
+    }, False)
+
+
+with tqdm(total=args.epochs, unit_scale=1, disable=args.distributed) as pbar:
```
mthrok, line 930: Use of `tqdm` prevents to run the script in background while keeping the log in readable manner. I'd not use it.

cpuhrsch: This could be made a flag for args (e.g. --verbose)

examples/pipeline/wav2letter.py
vincentqb: The previous version of LibriSpeech would raise an error since a file was missing by design of the dataset. This is no longer the case. Removed.

examples/pipeline/wav2letter.py
vincentqb: Good point

examples/pipeline/wav2letter.py
vincentqb: This is the script recommended for splitting train/dev/test in SpeechCommands' readme adapted to this use case. I suggest we include it with SpeechCommands in torchaudio.

examples/pipeline/wav2letter.py
vincentqb: Only one version is needed indeed. The list version is faster than the pytorch version though.

examples/pipeline/wav2letter.py
vincentqb: Debugging. I wanted to see if some loops were smaller than others. Will remove.

examples/pipeline/wav2letter.py
cpuhrsch: How are they better distributed in comparison to a random shuffle?

cpuhrsch:
I'd add "split into multiple files" as another todo as well.

examples/pipeline/wav2letter.py
vincentqb: Sorry, the real advantage of this approach is listed in the docstring:

> We want to keep files in the same training, validation, or testing sets even if new ones are added over time. This makes it less likely that testing samples will accidentally be reused in training when long runs are restarted for example.

examples/pipeline/metrics.py
```diff
@@ -0,0 +1,28 @@
+from typing import Optional
+
+import torch
+
+
+def levenshtein_distance(r: str, h: str, device: Optional[str] = None):
```
cpuhrsch, line 6: This seems worthwhile a separate PR already. Do you agree? In particular with the C++ extension we can create a JIT-able, fast version of this already.

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,508 @@
...
+        "--distributed", action="store_true", help="enable DistributedDataParallel"
+    )
+    parser.add_argument("--jit", action="store_true", help="if used, model is jitted")
+
+    args = parser.parse_args()
+
+    return args
+
+
+def signal_handler(a, b):
```
cpuhrsch, line 123: The need for functions like this worry me, because I'd imagine most users to not be aware of their necessity or purpose.

examples/pipeline/wav2letter.py
```diff
@@ -0,0 +1,508 @@
...
+    if is_best:
+        shutil.copyfile(filename, "model_best.pth.tar")
+    print("Checkpoint: saved", flush=True)
+
+
+def model_length_function(tensor):
+    return int(tensor.shape[0]) // 2 + 1
+
+
+def collate_fn(batch):
```
cpuhrsch, line 158: I'd move this into the file related to data just to shorten this file.

vincentqb:
> Are you going to publish the model?

Good point, adding mention in description. That could be a next step, though maybe not part of this original pull request.

examples/pipeline_wav2letter/metrics.py
```diff
@@ -0,0 +1,38 @@
+from typing import List, Union
+
+
+def levenshtein_distance(r: Union[str, List[str]], h: Union[str, List[str]]):
```
cpuhrsch, line 4: I'd move this into torchaudio before this lands

examples/pipeline_wav2letter/datasets.py
```diff
@@ -0,0 +1,113 @@
...
+        return data
+
+    # For performance, we cache all datasets
+    return tuple(create(dataset) for dataset in datasets)
+
+
+def collate_factory(model_length_function, transforms=None):
+
+    if transforms is None:
+        transforms = torch.nn.Sequential()
```
cpuhrsch, line 88: I'm not a fan of the declarative "nn.Sequential" approach and would write a custom function whose pointer I'd pass around, but I can see it being nice to aggregate transforms based on a sequence of decisions.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+
+from ctc_decoders import GreedyDecoder
+from datasets import collate_factory, split_process_librispeech
+from languagemodels import LanguageModel
+from metrics import levenshtein_distance
+from transforms import Normalize, UnsqueezeFirst
+from utils import MetricLogger, count_parameters, save_checkpoint
+
+
+def parse_args():
```
cpuhrsch, line 24: I'd specialize this file to doing just this and split out everything else into another file. You could call it "engine" or such.

cpuhrsch: That way you're separating this way of interfacing with your program (i.e. the CLI) from the actual logic.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        if clip_grad > 0:
+            metric["gradient"] = torch.nn.utils.clip_grad_norm_(
+                model.parameters(), clip_grad
+            )
+
+        optimizer.step()
+
+        compute_error_rates(outputs, targets, decoder, language_model, metric)
+
+        try:
```
cpuhrsch, line 300: Is this try/except due to this only being supported by a newer version of pytorch? Could could branch on the pytorch version string torch.__version__

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+    # Model
+
+    model = Wav2Letter(
+        num_classes=language_model.length,
+        input_type=args.type,
+        num_features=num_features,
+    )
+
+    if args.jit:
+        model = torch.jit.script(model)
```
cpuhrsch, line 468: What do you intent to use this for?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        )
+    elif args.optimizer == "adamw":
+        optimizer = AdamW(
+            model.parameters(),
+            lr=args.learning_rate,
+            momentum=args.momentum,
+            weight_decay=args.weight_decay,
+        )
+    else:
+        raise ValueError("Selected optimizer not supported")
```
cpuhrsch, line 515: Repeat the given option, i.e. "Selected optimizer %s not supported".format(args.optimizer) if you're going for this type of input sanitization to make it easier for the user to debug.

examples/pipeline_wav2letter/datasets.py
```diff
@@ -0,0 +1,113 @@
...
+
+        tensors = [transforms(b[0]) for b in batch if b]
+
+        tensors_lengths = torch.tensor(
+            [model_length_function(t) for t in tensors],
+            dtype=torch.long,
+            device=tensors[0].device,
+        )
+
+        tensors = torch.nn.utils.rnn.pad_sequence(tensors, batch_first=True)
```
cpuhrsch, line 100: A wrapped / generalized version of this could form a useful torchaudio function.

examples/pipeline_wav2letter/languagemodels.py
```diff
@@ -0,0 +1,38 @@
...
+        flipped = [(sub[1], sub[0]) for sub in enumerated]
+
+        d1 = collections.OrderedDict(enumerated)
+        d2 = collections.OrderedDict(flipped)
+        self.mapping = {**d1, **d2}
+
+    def encode(self, iterable):
+        if isinstance(iterable, list):
+            return [self.encode(i) for i in iterable]
+        else:
```
cpuhrsch, line 23: What if I pass an iterable that yields lists? What's the basecase type here? Maybe that's an easier case to branch on. Also as a very minor nit, I actually like using returns to avoid "else". So you could write

```
if isinstance(iterable, list):
    return [self.encode(i) for i in iterable]
return [self.mapping[i] + self.mapping[self.char_blank] for i in iterable]
```

examples/pipeline_wav2letter/ctc_decoders.py
```diff
@@ -0,0 +1,15 @@
+from torch import topk
+
+
+class GreedyDecoder:
```
cpuhrsch, line 4: You could generalize this file to be called "decoders.py" and also fold in things such as compute_error_rates

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        except AttributeError:
+            metric["lr"] = optimizer.param_groups[0]["lr"]
+
+        metric["batch size"] = len(inputs)
+        metric["n_channel"] = inputs.shape[1]
+        metric["n_time"] = inputs.shape[-1]
+        metric["dataset length"] += metric["batch size"]
+        metric["iteration"] += 1
+        metric["loss"] = loss.item()
+        metric["cumulative loss"] += metric["loss"]
```
cpuhrsch, line 311: I'd abstract this accumulation for both training evaluation and merge it into a single function. That way you'll always be sure that both training and evaluation are using the exact same calculations, since that's the last place you'd want to be buggy.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+    # Distributed setup
+
+    if args.distributed:
+        setup_distributed(rank, args.world_size)
+
+    not_main_rank = args.distributed and rank != 0
+
+    logging.info("Start time: %s", datetime.now())
+
+    # Explicitly set seed to make sure models created in separate processes
```
cpuhrsch, line 384: If this is for distributed training I'd worry that this isn't already happening. Did you have a case where this became necessary in order to avoid a bug?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+    loader_validation_params = loader_training_params.copy()
+    loader_validation_params["shuffle"] = False
+
+    loader_training = DataLoader(
+        training,
+        batch_size=args.batch_size,
+        collate_fn=collate_fn_train,
+        **loader_training_params,
+    )
+    loader_validation = DataLoader(
```
cpuhrsch, line 548: For validation "drop_last" is usually undesired because you can end up not running on the entire dataset.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+
+        model.load_state_dict(checkpoint["state_dict"])
+        optimizer.load_state_dict(checkpoint["optimizer"])
+        scheduler.load_state_dict(checkpoint["scheduler"])
+
+        logging.info(
+            "Checkpoint: loaded '%s' at epoch %s", args.checkpoint, checkpoint["epoch"]
+        )
+    else:
+        logging.info("Checkpoint: not found")
```
cpuhrsch, line 579: This seems like a case I'd error on. If the user intents to resume from this checkpoint and it wasn't found that's probably a mistake.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+            decoder,
+            language_model,
+            devices[0],
+            epoch,
+            args.clip_grad,
+            not_main_rank,
+            not args.reduce_lr_valid,
+        )
+
+        if not (epoch + 1) % args.print_freq or epoch == args.epochs - 1:
```
cpuhrsch, line 618: nit: I sometimes like to save on the indent and write something like:
```
if epoch < args.epoch - 1 and (epoch + 1) % args.print_freq:
    continue
```

examples/pipeline_wav2letter/transforms.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class Normalize(torch.nn.Module):
+    def forward(self, tensor):
+        return (tensor - tensor.mean(-1, keepdim=True)) / tensor.std(-1, keepdim=True)
+
+
+class UnsqueezeFirst(torch.nn.Module):
+    def forward(self, tensor):
+        return tensor.unsqueeze(0)
```
cpuhrsch, line 11: This is, in my opinion, the sort of issue that makes me dislike using nn.Sequential. You end up wrapping simple, small commands into modules.

examples/pipeline_wav2letter/utils.py
```diff
@@ -0,0 +1,55 @@
...
+    def __call__(self):
+        self._iter = (self._iter + 1) % self.print_freq
+        if not self.disable and not self._iter:
+            print(self, flush=True)
+
+
+def save_checkpoint(state, is_best, filename, disable):
+    """
+    Save the model to a temporary file first,
+    then copy it to filename, in case the signal interrupts
```
cpuhrsch, line 30: Has this happened? I think the scheduler is supposed to signal you and then you get a bunch of time to catch the signal and shutdown gracefully.

examples/pipeline_wav2letter/ctc_decoders.py
mthrok: This class is stateless. Can it be a function?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        metric["n_channel"] = inputs.shape[1]
+        metric["n_time"] = inputs.shape[-1]
+        metric["dataset length"] += metric["batch size"]
+        metric["iteration"] += 1
+        metric["loss"] = loss.item()
+        metric["cumulative loss"] += metric["loss"]
+        metric["average loss"] = metric["cumulative loss"] / metric["iteration"]
+        metric["iteration time"] = time() - start
+        metric["epoch time"] += metric["iteration time"]
+        metric()
```
mthrok, line 315: I assume this will flush the log but this should be given a descriptive method name.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+
+    if reduce_lr_on_plateau and isinstance(scheduler, ReduceLROnPlateau):
```
mthrok, line 317: Isn't `scheduler` automatically instance of `ReduceLROnPlateau`, when `reduce_lr_on_plateau`?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+            metric["lr"] = optimizer.param_groups[0]["lr"]
+
+        metric["batch size"] = len(inputs)
+        metric["n_channel"] = inputs.shape[1]
+        metric["n_time"] = inputs.shape[-1]
+        metric["dataset length"] += metric["batch size"]
+        metric["iteration"] += 1
+        metric["loss"] = loss.item()
+        metric["cumulative loss"] += metric["loss"]
+        metric["average loss"] = metric["cumulative loss"] / metric["iteration"]
```
mthrok, line 312: I do not think "average loss" for training is meaningful. The model changes for each batch.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        # keep batch first for data parallel
+        outputs = model(inputs).transpose(-1, -2).transpose(0, 1)
+
+        # CTC
+        # outputs: input length, batch size, number of classes (including blank)
+        # targets: batch size, max target length
+        # input_lengths: batch size
+        # target_lengths: batch size
+
+        loss = criterion(outputs, targets, tensors_lengths, target_lengths)
```
mthrok, line 286: Is this `loss` averaged over batch or summed?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+            # keep batch first for data parallel
+            outputs = model(inputs).transpose(-1, -2).transpose(0, 1)
+
+            # CTC
+            # outputs: input length, batch size, number of classes (including blank)
+            # targets: batch size, max target length
+            # input_lengths: batch size
+            # target_lengths: batch size
+
+            metric["cumulative loss"] += criterion(
```
mthrok, line 357: Is the loss value averaged over batch or summed over batch?

examples/pipeline_wav2letter/README.md
```diff
@@ -0,0 +1,45 @@
...
+    --normalize \
+    --optimizer adadelta \
+    --scheduler reduceonplateau \
+    --epochs 30
+```
+With these default parameters, we get a character error rate of 13.8% on dev-clean after 30 epochs.
+
+### Output
+
+The information reported at each iteration and epoch (e.g. loss, character error rate, word error rate) is printed to standard output in the form of one json per line, e.g.
```
mthrok, line 30: I would write log to a separate file alongside with saved model, otherwise users have to redirect all the time, which is not very convenient.

examples/pipeline_wav2letter/README.md
```diff
@@ -0,0 +1,45 @@
...
+```python
```
mthrok, line 31: Isn't it `json?`

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+def main(rank, args):
+
+    # Distributed setup
+
+    if args.distributed:
+        setup_distributed(rank, args.world_size)
+
+    not_main_rank = args.distributed and rank != 0
+
+    logging.info("Start time: %s", datetime.now())
```
mthrok, line 382: Doesn't logging have timestamp itself?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+                n_mfcc=args.n_bins,
+                melkwargs=melkwargs,
+            ),
+        )
+        num_features = args.n_bins
+    elif args.type == "waveform":
+        transforms = torch.nn.Sequential(UnsqueezeFirst())
+        num_features = 1
+    else:
+        raise ValueError("Model type not supported")
```
mthrok, line 417: This is unreachable from CLI. `NotImplementedError` makes more sense.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        root=args.dataset_root,
+        folder_in_archive=args.dataset_folder_in_archive,
+    )
+
+    # Decoder
+
+    if args.decoder == "greedy":
+        decoder = GreedyDecoder()
+    else:
+        raise ValueError("Selected decoder not supported")
```
mthrok, line 457: This is unreachable from CLI. `NotImplementedError` makes more sense.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        input_type=args.type,
+        num_features=num_features,
+    )
+
+    if args.jit:
+        model = torch.jit.script(model)
+
+    if args.distributed:
+        n = torch.cuda.device_count() // args.world_size
+        devices = list(range(rank * n, (rank + 1) * n))
```
mthrok, line 472: This device assignment seems to assume that all the nodes (if there are multiple) have the same number of CUDA devices, I think `device_id = [rank // torch.cuda.device_count()]` would do in that case.

examples/pipeline_wav2letter/main.py
mthrok: Also, this code is unreachable from CLI, so `NotImplementedError` makes more sense, because the only time you reach here is when you intend to add a new choice and changed CLI parser but forgot to add actual implementation.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+        model = model.to(devices[0])
+        model = torch.nn.parallel.DistributedDataParallel(model, device_ids=devices)
+    else:
+        devices = ["cuda" if torch.cuda.is_available() else "cpu"]
+        model = model.to(devices[0], non_blocking=True)
+        model = torch.nn.DataParallel(model)
```
mthrok, line 478: Isn't `DataParallel` discouraged? Is there a performance implications compared against bare model?

examples/pipeline_wav2letter/main.py
mthrok: Also I would extract this into a helper function so that main logic becomes readable. `_get_optimizer(...)`

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+    collate_fn_valid = collate_factory(model_length_function)
+
+    loader_training_params = {
+        "num_workers": args.workers,
+        "pin_memory": True,
+        "shuffle": True,
+        "drop_last": True,
+    }
+    loader_validation_params = loader_training_params.copy()
+    loader_validation_params["shuffle"] = False
```
mthrok, line 540: I do not see a readability improvement from these intermediate dictionaries. Rather, passing them to `DataLoader` constructor directly would be simple and straightforward.

examples/pipeline_wav2letter/main.py
mthrok: Also the logic here is strange. If user does not give the checkpoint option (training from scratch), there is no need to say not found.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+            args.checkpoint,
+            not_main_rank,
+        )
+
+    if args.distributed:
+        torch.distributed.barrier()
+
+    torch.autograd.set_detect_anomaly(False)
+
+    for epoch in range(args.start_epoch, args.epochs):
```
mthrok, line 599: Isn't it `range(start_epoch, start_epoch + epochs)`?

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+                    "best_loss": best_loss,
+                    "optimizer": optimizer.state_dict(),
+                    "scheduler": scheduler.state_dict(),
+                },
+                is_best,
+                args.checkpoint,
+                not_main_rank,
+            )
+
+        if args.reduce_lr_valid and isinstance(scheduler, ReduceLROnPlateau):
```
mthrok, line 646: I think `and isinstance(scheduler, ReduceLROnPlateau)` is redundant.

examples/pipeline_wav2letter/metrics.py
mthrok: If moving this into the library, docstring needs to be improved with the equation.

examples/pipeline_wav2letter/utils.py
```diff
@@ -0,0 +1,55 @@
...
+import os
+import shutil
+from collections import defaultdict
+
+import torch
+
+
+class MetricLogger(defaultdict):
+    def __init__(self, name, print_freq=1, disable=False):
+        super().__init__(lambda: 0.0)
```
mthrok, line 12: I think `super().__init__(float)` is better.

examples/pipeline_wav2letter/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+def save_checkpoint(state, is_best, filename, disable):
+    """
+    Save the model to a temporary file first,
+    then copy it to filename, in case the signal interrupts
+    the torch.save() process.
+    """
+
+    if disable:
+        return
```
mthrok, line 35: I do not think this should be the logic of `save_checkpoinit` function. This goes against single-responsibility principle. It's caller's responsibility to check when to save.

examples/pipeline_wav2letter/utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if filename == "":
+        return
```
mthrok, line 38: I think, giving empty string as file location should be error.

examples/pipeline_wav2letter/utils.py
mthrok: Also in my opinion this logic does not corresponds to the name of the function. `save_checkpoint` should do saving and only saving. The logic for Handling temporary file for the sake of interruption solved different concern and should live in a different function.

mthrok (commented):
I have concerns on accuracy metrics. I strongly believe that WER computation is incorrect and we should not re-invent the wheel and use SCTK or something.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+
+    cers = [levenshtein_distance(t, o) for t, o in zip(target, output)]
+    cers = sum(cers)
+    n = sum(len(t) for t in target)
+    metric["cer over target length"] = cers / n
+    metric["cumulative cer"] += cers
+    metric["total chars"] += n
+    metric["cumulative cer over target length"] = metric["cer"] / metric["total chars"]
+
+    # Compute WER
```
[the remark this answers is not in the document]
vincentqb: > This way of WER computation ignores the common false cases such as removal/insertion, and does not report the correct value.

Do you have a test case that is failing with the definition given here?

> To compute WER, dynamic programming os sort is required.

`output` and `target` are batches of samples that need to be compared zip-like (first `output` with first `target`, etc) using `levenshtein_distance`. The latter is where the dynamic programming occurs.

The normalization of the edit distance is simply the division by the number of reference characters/words.

> For this one should resort to a standard tools like NIST SCTK.

Are you suggesting adding a test against a toolkit like SCTK? Currently, the current tests are in the levenshtein distance file.

examples/pipeline_wav2letter/ctc_decoders.py
cpuhrsch: It could be functional corresponding to a transform, but really it's a step towards our beamsearch work

examples/pipeline_wav2letter/main.py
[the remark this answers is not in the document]
cpuhrsch: I don't think NIST SCTK is necessary, because other projects don't necessarily use it either. It might be overkill as a dependency, but we could use it as a reference and to test our implementation in CI.

examples/pipeline_wav2letter/main.py
vincentqb: simplified to a one-line for all

examples/pipeline_wav2letter/main.py
vincentqb: i wanted to offer the option to jit, but i agree this is not useful here.

examples/pipeline_wav2letter/main.py
vincentqb: not quite, since one could decide to reduce lr based on some training value instead of validation value. that being said, it may not be useful to expose so many options :)

examples/pipeline_wav2letter/main.py
vincentqb: It's a fair comment about the model changing, so a name like `batch loss` might be more relevant. the real reason i have that though is that i need to reduce the learning rate if some metric has not decreased much recently. for that i was using the loss over the epoch. i could experiment with other metric as criteria though :)

examples/pipeline_wav2letter/main.py
vincentqb: averaged by default -- but it's a good observation, i'll make it summed, and i'll average manually so that each iteration gets the same weight in case the last iteration of an epoch doesn't have the same number of samples

examples/pipeline_wav2letter/main.py
vincentqb: [comment](https://github.com/pytorch/audio/pull/632#discussion_r491197079)

examples/pipeline_wav2letter/README.md
vincentqb: i must admit i do like the standard output a lot -- but i can see users preferring writing to a file, so i'll add the option to choose :)

examples/pipeline_wav2letter/main.py
vincentqb: not by default. removed this line, and replaced by timestamp in logging.

cpuhrsch (approved):

cpuhrsch (commented):
This PR is getting very, very large and has been up for a long time. Let's merge it, since it's already working, and revisit some of these suggested improvements on a PR-by-PR basis. It'll also help us share code across examples etc.

vincentqb:
Following [comment](https://github.com/pytorch/audio/pull/632#pullrequestreview-492929463), reverted to [Aug 19](https://github.com/pytorch/audio/pull/632#event-3672689390), merging, and moved follow-up to vincentqb/audio#3.

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+    logging.info(args)
+    return args
+
+
+def setup_distributed(rank, world_size):
+    os.environ["MASTER_ADDR"] = "localhost"
+    os.environ["MASTER_PORT"] = "12355"
+
+    # initialize the process group
+    torch.distributed.init_process_group("nccl", rank=rank, world_size=world_size)
```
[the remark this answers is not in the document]
vincentqb: Good point, I'll switch between nccl/gloo as done for conv-tasnet, [here](https://github.com/pytorch/audio/pull/896/files#diff-c1bfa40dcc3ccedec4987ab6fee91bd6R123).

examples/pipeline_wav2letter/main.py
```diff
@@ -0,0 +1,668 @@
...
+
+
+def model_length_function(tensor):
+    if tensor.shape[1] == 1:
```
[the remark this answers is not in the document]
vincentqb: Replaced by a constructor with an explicit parameter in vincentqb/audio#2

examples/pipeline_wav2letter/main.py
vincentqb: The documentation does point out that distributed is usually faster than the simpler-to-use data parallel, see [here](https://pytorch.org/tutorials/intermediate/ddp_tutorial.html). Having distributed be the default requires more dependencies even on single machine?

examples/pipeline_wav2letter/main.py
vincentqb: epochs is meant to be the maximal epoch to reach, so in that sense, no. however, the name may be misleading, so renamed.

examples/pipeline_wav2letter/main.py
vincentqb: the logic meant to let the user choose between decaying during training or validation.

vincentqb:
As mentioned in the README, we can get less than 13.8% "cer over target length" after 30 epochs. See sample output grepped for validation: 

```
{"name": "validation", "epoch": 0, "cumulative loss": 70.70183157920837, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3667538847242082, "validation time": 25.846306562423706}
{"name": "validation", "epoch": 1, "cumulative loss": 69.96176052093506, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3315124057588124, "validation time": 6.5293145179748535}
{"name": "validation", "epoch": 2, "cumulative loss": 70.07647657394409, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3369750749497187, "validation time": 6.5231428146362305}
{"name": "validation", "epoch": 3, "cumulative loss": 69.94338345527649, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3306373073941185, "validation time": 5.516169548034668}
{"name": "validation", "epoch": 4, "cumulative loss": 69.97763347625732, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3322682607741583, "validation time": 6.777941942214966}
{"name": "validation", "epoch": 5, "cumulative loss": 69.72349190711975, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3201662812914168, "validation time": 6.666280746459961}
{"name": "validation", "epoch": 6, "cumulative loss": 70.0134003162384, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3339714436304, "validation time": 6.872417688369751}
{"name": "validation", "epoch": 7, "cumulative loss": 69.52196455001831, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3105697404770624, "validation time": 6.776198148727417}
{"name": "validation", "epoch": 8, "cumulative loss": 69.42171597480774, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.3057959988003685, "validation time": 6.710602760314941}
{"name": "validation", "epoch": 9, "cumulative loss": 69.25003266334534, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.2976206030164446, "validation time": 6.838287353515625}
{"name": "validation", "epoch": 10, "cumulative loss": 64.44925856590271, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 1.0, "cumulative cer": 280923.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0, "cumulative wer": 54008.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.069012312662034, "validation time": 6.446615695953369}
{"name": "validation", "epoch": 11, "cumulative loss": 63.2327446937561, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.9225932936722553, "cumulative cer": 258065.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0010559662090812, "cumulative wer": 54122.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 3.0110830806550526, "validation time": 7.578996181488037}
{"name": "validation", "epoch": 12, "cumulative loss": 57.679614543914795, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.6934829637641968, "cumulative cer": 195536.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.3079901443153819, "cumulative wer": 70233.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 2.7466483116149902, "validation time": 12.67360258102417}
{"name": "validation", "epoch": 13, "cumulative loss": 54.061622858047485, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.6738777717685235, "cumulative cer": 189715.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.3005983808518127, "cumulative wer": 69715.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 2.5743629932403564, "validation time": 14.44613003730774}
{"name": "validation", "epoch": 14, "cumulative loss": 42.6647093296051, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.6101270957274202, "cumulative cer": 169292.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 1.0091517071453713, "cumulative wer": 54202.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 2.0316528252192905, "validation time": 15.787317752838135}
{"name": "validation", "epoch": 15, "cumulative loss": 30.52291715145111, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.42664954029204977, "cumulative cer": 120448.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.9686730024639212, "cumulative wer": 51122.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 1.4534722453071958, "validation time": 18.890476942062378}
{"name": "validation", "epoch": 16, "cumulative loss": 22.910719513893127, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.3174012979989183, "cumulative cer": 91667.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.7529039070749736, "cumulative wer": 41637.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 1.0909866435187203, "validation time": 19.944958686828613}
{"name": "validation", "epoch": 17, "cumulative loss": 17.98588478565216, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.2589913466738778, "cumulative cer": 75363.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.6613868356212601, "cumulative wer": 36475.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.8564707040786743, "validation time": 19.98793649673462}
{"name": "validation", "epoch": 18, "cumulative loss": 15.67355227470398, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.22397241752298538, "cumulative cer": 64542.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.6117564237944386, "cumulative wer": 33368.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.7463596321287609, "validation time": 20.789494514465332}
{"name": "validation", "epoch": 19, "cumulative loss": 13.239276766777039, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1870605732828556, "cumulative cer": 55027.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.5381907778951074, "cumulative wer": 29982.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.6304417507989066, "validation time": 20.70693564414978}
{"name": "validation", "epoch": 20, "cumulative loss": 12.41661947965622, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.180164954029205, "cumulative cer": 51669.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.5255191833861317, "cumulative wer": 28399.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5912675942693438, "validation time": 20.92684292793274}
{"name": "validation", "epoch": 21, "cumulative loss": 11.557319521903992, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.17056517036235802, "cumulative cer": 49252.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.5072157690953889, "cumulative wer": 27600.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5503485486620948, "validation time": 21.347500801086426}
{"name": "validation", "epoch": 22, "cumulative loss": 10.85356280207634, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.15548945375878853, "cumulative cer": 46219.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.46673706441393875, "cumulative wer": 26201.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5168363239083972, "validation time": 20.922102212905884}
{"name": "validation", "epoch": 23, "cumulative loss": 10.527889400720596, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1535289345592212, "cumulative cer": 44505.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.47624076029567053, "cumulative wer": 25758.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5013280667009807, "validation time": 20.710388660430908}
{"name": "validation", "epoch": 24, "cumulative loss": 10.139740884304047, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.14710654407787993, "cumulative cer": 43546.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.44948961633227735, "cumulative wer": 25158.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.48284480401447843, "validation time": 21.18306064605713}
{"name": "validation", "epoch": 25, "cumulative loss": 10.286657720804214, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1484586262844781, "cumulative cer": 42859.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4463217177050334, "cumulative wer": 24725.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.48984084384781973, "validation time": 20.720789670944214}
{"name": "validation", "epoch": 26, "cumulative loss": 9.967010378837585, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.14264467279610601, "cumulative cer": 41970.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.43681802182330165, "cumulative wer": 24719.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.47461954184940885, "validation time": 20.847771406173706}
{"name": "validation", "epoch": 27, "cumulative loss": 9.573839098215103, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13737155219037317, "cumulative cer": 40117.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.42696233720520943, "cumulative wer": 23860.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.45589709991500493, "validation time": 20.785948276519775}
{"name": "validation", "epoch": 28, "cumulative loss": 9.409523874521255, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13608707409410492, "cumulative cer": 39419.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.42625835973248855, "cumulative wer": 23408.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.44807256545339313, "validation time": 21.98930525779724}
{"name": "validation", "epoch": 29, "cumulative loss": 9.617189735174179, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1341265548945376, "cumulative cer": 39536.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.42203449489616335, "cumulative wer": 23571.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4579614159606752, "validation time": 21.18846106529236}
{"name": "validation", "epoch": 30, "cumulative loss": 9.645921111106873, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13608707409410492, "cumulative cer": 39568.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4325941569869764, "cumulative wer": 23657.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4593295767193749, "validation time": 20.77275061607361}
{"name": "validation", "epoch": 31, "cumulative loss": 9.530102461576462, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13115197404002163, "cumulative cer": 38687.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.41358676522351284, "cumulative wer": 23131.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4538144029322125, "validation time": 20.9629008769989}
{"name": "validation", "epoch": 32, "cumulative loss": 9.8547984957695, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13662790697674418, "cumulative cer": 39482.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.42872228088701164, "cumulative wer": 23718.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4692761188461667, "validation time": 21.710347414016724}
{"name": "validation", "epoch": 33, "cumulative loss": 9.720289260149002, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13270686857760952, "cumulative cer": 38280.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4237944385779655, "cumulative wer": 23243.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4628709171499525, "validation time": 21.28827738761902}
{"name": "validation", "epoch": 34, "cumulative loss": 10.022859960794449, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13067874526771228, "cumulative cer": 38105.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.41534670890531505, "cumulative wer": 23047.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4772790457521166, "validation time": 20.7469642162323}
{"name": "validation", "epoch": 35, "cumulative loss": 9.8920236825943, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1333829096809086, "cumulative cer": 37309.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.41745864132347765, "cumulative wer": 22699.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.47104874679020475, "validation time": 21.03089475631714}
{"name": "validation", "epoch": 36, "cumulative loss": 10.471008330583572, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.1385884261763115, "cumulative cer": 40063.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.43400211193241817, "cumulative wer": 23698.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.49861944431350347, "validation time": 21.33878517150879}
{"name": "validation", "epoch": 37, "cumulative loss": 10.184874802827835, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13223363980530017, "cumulative cer": 38067.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4104188665962689, "cumulative wer": 22916.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.4849940382298969, "validation time": 21.298787117004395}
{"name": "validation", "epoch": 38, "cumulative loss": 10.797764033079147, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13770957274202272, "cumulative cer": 39543.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4297782470960929, "cumulative wer": 23707.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5141792396704356, "validation time": 21.03481435775757}
{"name": "validation", "epoch": 39, "cumulative loss": 11.389482617378235, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13547863710113575, "cumulative cer": 39638.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.42027455121436114, "cumulative wer": 23703.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.5423563151132493, "validation time": 20.84154224395752}
{"name": "validation", "epoch": 40, "cumulative loss": 12.254583895206451, "dataset length": 2688.0, "iteration": 21.0, "cer over target length": 0.13040832882639264, "cumulative cer": 37159.0, "total chars": 280923.0, "cer": 0.0, "cumulative cer over target length": 0.0, "wer over target length": 0.4171066525871172, "cumulative wer": 22646.0, "total words": 54008.0, "wer": 0.0, "cumulative wer over target length": 0.0, "average loss": 0.58355161405745, "validation time": 20.9519464969635}
```

PR #635, opened 2020-05-13
Remove multiple definitions of `SoxEffect`.

### LINKED ISSUE #628: multiple definitions of `SoxEffect`.
`SoxEffect` is defined both in

https://github.com/pytorch/audio/blob/00d38203e401b8d9472a8f8394a10e2c309be02c/torchaudio/torch_sox.h#L50-L54

and

https://github.com/pytorch/audio/blob/00d38203e401b8d9472a8f8394a10e2c309be02c/torchaudio/torch_sox.cpp#L73-L77

-----

-----

`torch_sox.h` can be included into `torch_sox.cpp` so that `SoxEffect` needs to be defined
only once in the header.

csukuangfj:
Fix #628

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ else:
             '_torch_sox',
             ['torchaudio/torch_sox.cpp'],
             libraries=libraries,
-            include_dirs=include_dirs,
+            include_dirs=include_dirs + [cwd],
             extra_compile_args=eca,
             extra_objects=extra_objects,
             extra_link_args=ela),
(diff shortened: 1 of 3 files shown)


csukuangfj:
@vincentqb 

Please have a review. Ready to merge.

csukuangfj:
>  @csukuangfj, can you tell us how you are using it, so we can understand what to expose?

I need only the python interface for sox. `torch_sox.h` is not needed at present.

PR #640, opened 2020-05-14
Adopt PyTorch's test util to torchscript test

mthrok:
@cpuhrsch 

Part 1 of PyTorch test framework adaptation. Applied to torchscript tests and some functional test.
If this looks good, I will expand the adaptation to other tests.

### DIFF
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -18,9 +18,9 @@ class Functional(common_utils.TestBaseMixin):
         output = func(tensor)
         ts_output = ts_func(tensor)
         if shape_only:
-            assert ts_output.shape == output.shape, (ts_output.shape, output.shape)
-        else:
-            self.assertEqual(ts_output, output)
+            ts_output = ts_output.shape
+            output = output.shape
+        self.assertEqual(ts_output, output)
 
     def test_spectrogram(self):
         def func(tensor):
--- a/test/test_torchscript_consistency.py
+++ b/test/test_torchscript_consistency.py
@@ -1,6 +1,5 @@
 """Test suites for jit-ability and its numerical compatibility"""
 import unittest
-import pytest
 
 import torch
 import torchaudio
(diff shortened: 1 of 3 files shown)


test/test_torchscript_consistency.py
```diff
@@ -10,29 +9,18 @@
...
     """Implements test for `functinoal` modul that are performed for different devices"""
     def _assert_consistency(self, func, tensor, shape_only=False):
         tensor = tensor.to(device=self.device, dtype=self.dtype)
-        return _assert_functional_consistency(func, tensor, shape_only=shape_only)
+
+        ts_func = torch.jit.script(func)
+        output = func(tensor)
+        ts_output = ts_func(tensor)
+        if shape_only:
+            assert ts_output.shape == output.shape, (ts_output.shape, output.shape)
```
cpuhrsch, line 21: Instead of using ```assert``` let's use one of the ```self.assertXYZ``` methods. That'll yield better error messages.

test/test_torchscript_consistency.py
```diff
@@ -209,7 +197,7 @@ def func(tensor):
 
     def test_lfilter(self):
         if self.dtype == torch.float64:
-            pytest.xfail("This test is known to fail for float64")
+            raise unittest.SkipTest("This test is known to fail for float64")
```
cpuhrsch, line 200: I'd make this a decorator then and have separate tests for each dtype

test/test_torchscript_consistency.py
mthrok: updated

test/test_torchscript_consistency.py
mthrok: I think that's outside of the scope of this PR.

vincentqb:
Following this PR, with pytorch 1.5.0 I get the following. I'm assuming this is only compatible with nighlties?
```
  File "test/test_librosa_compatibility.py", line 17, in <module>
    import common_utils
  File "/Users/vincentqb/audio/test/common_utils.py", line 9, in <module>
    from torch.testing._internal.common_utils import TestCase
ModuleNotFoundError: No module named 'torch.testing._internal'
```

vincentqb:
> I'm assuming this is only compatible with nighlties?

offline follow-up with @mthrok -- yup.

PR #641, opened 2020-05-14
Make parameter optional string

vincentqb:
Following up on [comment](https://github.com/pytorch/audio/pull/589#discussion_r425449900)

### DIFF
--- a/test/torchscript_consistency_impl.py
+++ b/test/torchscript_consistency_impl.py
@@ -83,7 +83,7 @@ class Functional(common_utils.TestBaseMixin):
             f_max = 20.0
             n_mels = 10
             sample_rate = 16000
-            norm = ""
+            norm = None
             return F.create_fb_matrix(n_stft, f_min, f_max, n_mels, sample_rate, norm)
 
         dummy = torch.zeros(1, 1)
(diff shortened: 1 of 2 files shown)


torchaudio/functional.py
```diff
@@ -376,7 +376,7 @@ def create_fb_matrix(
     up_slopes = slopes[:, 2:] / f_diff[1:]  # (n_freqs, n_mels)
     fb = torch.max(zero, torch.min(down_slopes, up_slopes))
 
-    if norm == "slaney":
+    if norm is not None and norm == "slaney":
```
mthrok, line 379: I think we should not allow values other than `"slaney"` or `None` to prevent the case where unintended value is passed, like typo.

mthrok:
@vincentqb Test needs to be fixed.

test/torchscript_consistency_impl.py
```diff
@@ -83,7 +83,7 @@ def func(_):
             f_max = 20.0
             n_mels = 10
             sample_rate = 16000
-            norm = ""
+            norm = None
```
mthrok, line 86: Rather than changing it to `None`, I suggest following [the original suggestion](https://github.com/pytorch/audio/pull/589#discussion_r420324494) of passing string value.

PR #580, opened 2020-04-24
CPP extension for overdrive effect in functional

bhargavkathivarapu:
Hi ,

I  have tried implementing existing fucntional ```overdrive``` using cpp as the python version is slow compared to Sox version . ( #260 ( Reducing dependency in Sox) )
Though cpp version is slight slower than sox version . It is much better compated to python version 

- comparing Sox with new overdrive(with cpp ext )
Screenshot 2020-04-24 at 7 52 00 PM

- old python overdrive effect took 80000 ms
Screenshot 2020-04-24 at 8 28 46 PM

Almost >700X speed compared to python implementation

- sox comapatibilty - passed
- batch test - passed
- Torch script - not passed . I think  cpp extension cannot be converted to torch script
```
$ python test/test_torchscript_consistency.py
.............................E.....sssssssssssssssssssssssssssssssssss................ssssssssssssssss
======================================================================
ERROR: test_overdrive (__main__.TestFunctionalCPU)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test/test_torchscript_consistency.py", line 474, in test_overdrive
    self._assert_consistency(func, waveform)
  File "test/test_torchscript_consistency.py", line 38, in _assert_consistency
    return _assert_functional_consistency(func, tensor, self.device, shape_only=shape_only)
  File "test/test_torchscript_consistency.py", line 14, in _assert_functional_consistency
    ts_func = torch.jit.script(func)
  File "/Users/ka387861/pytorch/torch/jit/__init__.py", line 1296, in script
    fn = torch._C._jit_script_compile(qualified_name, ast, _rcb, get_default_args(obj))
  File "/Users/ka387861/pytorch/torch/jit/_recursive.py", line 559, in try_compile_fn
    return torch.jit.script(fn, _rcb=rcb)
  File "/Users/ka387861/pytorch/torch/jit/__init__.py", line 1296, in script
    fn = torch._C._jit_script_compile(qualified_name, ast, _rcb, get_default_args(obj))
RuntimeError: 
Python builtin <built-in method _overdrive_helper of PyCapsule object at 0x143fe2270> is currently not supported in Torchscript:
  File "/Users/ka387861/audio/torchaudio/functional.py", line 1291

    # TODO: Implement a torch CPP extension
    _overdrive_helper(waveform, temp, last_in, last_out, output_waveform)
    ~~~~~~~~~~~~~~~~~ <--- HERE

    return output_waveform.clamp(min=-1, max=1).view(actual_shape)
[...]

PR #290, opened 2019-09-18
CPP Implementation of lfilter for CPU

engineerchuan:
To compare with the current python implementation of `lfilter` (implementation 0), we developed 2 cpp implementations.
1.) an element wise implementation using accessors (implementation 1)
2.) a matrix based implementation (implementation 2)

We test these implementations against each other using random inputs, and a variety of signal lengths and filter orders:

*Performance*
1. Element Wise CPP (implementation 1) is typically much faster, on the order of >800x for lower order filters and 400x for higher order filters.  This is an argument for including this into the codebase.  We should in the future also compile on GPU.

*Correctness*
1. For all data sizes and order filters, both matrix based implementations (0 and 2) return the same results at <3e-4 tolerance.
2. For higher order filters and longer inputs, the element wise implementation can diverge from the matrix.  This makes sense as floating point round offs will typically accumulate.  If the order of operations in matrix method and element wise method are not identical, these results would not expect to match.  Generally, the user should be expected to supply filter coefficients that are stable.
3. This has been tested against by using a 2 order filter compared to SoX biquad results.

*Questions / Comments*
1. Is there a better way to "slice" that is more efficient?  @cpuhrsch 
2. @yf225  Thank you for your suggestion on a previous PR to parallelize across channel.  I think that generally will reduce time by at most 4x because audio waveforms have usually 2 channels.  So I don't think it'd be meaningful in choosing between matrix
3. @vincentqb Any thoughts?

PR #642, opened 2020-05-15
Add windows binary jobs

peterjc123:


peterjc123:
@mthrok @vincentqb

peterjc123:
The wheels generated on Windows are noarch. I don't know whether it will break the search process for the wheel packages on Linux/MacOS.

mthrok:
> noarch

@seemethere Could you advise?

peterjc123:
@mthrok I tried to make the wheel package specific both on python version and platform. But I guess it is okay without this change. I quoted the related description from [PEP-425](https://www.python.org/dev/peps/pep-0425/#id14).
> Sometimes there will be more than one supported built distribution for a particular version of a package. For example, a packager could release a package tagged cp33-abi3-linux_x86_64 that contains an optional C extension and the same distribution tagged py3-none-any that does not. The index of the tag in the supported tags list breaks the tie, and the package with the C extension is installed in preference to the package without because that tag appears first in the list.

mthrok:
> @mthrok I tried to make the wheel package specific both on python version and platform. But I guess it is okay without this change. I quoted the related description from [PEP-425](https://www.python.org/dev/peps/pep-0425/#id1).

Thank you for the link.

So based on my understanding of PEP-425, and looking at the job logs, the binaries built by new jobs are the following. The Windows wheel binaries have `win_amd64` as part of the name, so it should not interfere with Linux binaries or macOS binaries. Conda does not have ABI as part of the name but I guess they store such info as meta data somewhere in the files uploaded altogether.

I think it's safe to merge this PR. If this does interfere with binaries for Linux or macOS, then we can tell with [the nightly smoke test job](https://app.circleci.com/pipelines/github/pytorch/audio?branch=nightly) tomorrow morning. 

Side note:

Is 94994ecd sufficient to add ABI for Windows? We plan to revisit and add C++ implementations like https://github.com/pytorch/audio/pull/580, https://github.com/pytorch/audio/pull/290 which might be able to build on Windows. In that case, we will need to add proper ABI. (though since `torchaudio` does not build extension on Windows so there are many steps ahead towards that.)

---

(from the last line of **Uploading artifacts** job log)

* **Wheel**
  - `torchaudio-0.6.0.dev20200516-cp38-none-win_amd64.whl`
     - Python: `CPython3.8`
     - ABI: **N/A**
     - Platform: `win_amd64`
  - `torchaudio-0.6.0.dev20200515-cp38-cp38-macosx_10_9_x86_64.whl`
     - Python: `CPython3.8`
     - ABI: `CPython3.8`
     - Platform: `macosx_10_9_x86_64`
  - `torchaudio-0.6.0.dev20200516-cp38-cp38-linux_x86_64.whl`
     - Python: `CPython3.8`
     - ABI: `CPython3.8`
     - Platform: `linux_x86_64`

* **Conda** 
  - `win-64\torchaudio-0.6.0.dev20200516-py38.tar.bz2`
     - Python: `Python 3.8`
     - ABI: **N/A**
     - Platform: `win-64`
  - `osx-64/torchaudio-0.6.0.dev20200515-py38.tar.bz2`
     - Python: `Python 3.8`
     - ABI: `cpython-38-darwin`
     - Platform: `osx-64`
  - `linux-64/torchaudio-0.6.0.dev20200516-py38.tar.bz2`
     - Python: `Python 3.8`
     - ABI: `cpython-38-x86_64-linux-gnu`
     - Platform: `linux-64`

PR #644, opened 2020-05-15
Split CUDA and CPU tests of `torchscript_consistency`

mthrok:
Currently there is no CUDA device for regular FB's infra, where `torchaudio`'s tests are ran.
Skipping these tests consistently raises a flag so we decided to split the tests into runnable and not-runnable on file-level.


cpuhrsch:
Looks like there's some lint issues

PR #647, opened 2020-05-16
Update smoke test docker image and tag new images latest

mthrok:
1. Docker image used for smoke test is old (No Python 3.8). Updating this to the latest one. build [here](https://app.circleci.com/pipelines/github/pytorch/audio/1763/workflows/56c846a5-acaa-41a7-92f5-46ec66186c61/jobs/44484)
2. Add `latest` tag to the newly built docker so that later we can always use `latest` version and we do not have to update docker image tag manually.


mthrok:
Thanks!

PR #651, opened 2020-05-19
Add flanger to functional.py

bhargavkathivarapu:
Hi ,
Regarding the issue ( Reducing dependency in Sox - #260  ),

This pull request is new version of existing pull request #629 

@mthrok I have made the changes to batch and torch script test .
Now taking random input generated from a seed

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -1363,8 +1363,8 @@ def flanger(
         width: float = 71.,
         speed: float = 0.5,
         phase: float = 25.,
-        sinusoidal: bool = True,
-        linear_interpolation: bool = True
+        modulation: str = 'sinusoidal',
+        interpolation: str = 'linear'
 ) -> Tensor:
     r"""Apply a flanger effect to the audio. Similar to SoX implementation.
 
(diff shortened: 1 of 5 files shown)


mthrok:
Thanks, looks like there is something not working for 2D input. The tests are failing. Could you take a look at it?

bhargavkathivarapu:
@mthrok , Fixed the issue , now all checks passed

vincentqb:
@cpuhrsch -- you mentioned offline performance improvements. can you point them here?

torchaudio/functional.py
```diff
@@ -1353,6 +1354,152 @@ def _generate_wave_table(
...
+        linear_interpolation (bool):  If ``True``, uses linear interpolation for delay-line interpolation
+            If ``False``, uses Quadratic interpolation
+            (Default: ``True``)
+
+    Returns:
+        Tensor: Waveform of dimension of `(...,channels,time)`
+
+    References:
+        http://sox.sourceforge.net/sox.html
+        Scott Lehman, Effects Explained, http://harmony-central.com/Effects/effects-explained.html
```
vincentqb, line 1399: This links does not work :)

torchaudio/functional.py
```diff
@@ -1353,6 +1354,152 @@ def _generate_wave_table(
...
+        speed (float):  modulation speed in Hz
+            Allowed range of values are 0.1 to 10
+        phase (float):  percentage phase-shift for multi-channel
+            Allowed range of values are 0 to 100
+        sinusoidal (bool):  If ``True``, uses sinusoidal modulation
+            If ``False``, uses triangular modulation
+            (Default: ``True``)
+        linear_interpolation (bool):  If ``True``, uses linear interpolation for delay-line interpolation
+            If ``False``, uses Quadratic interpolation
+            (Default: ``True``)
```
vincentqb, line 1392: In sox, these are flags were we pass strings. I can't find a link with a description of the links, can you point me to one? Are there more than two possibilities that could happen here? In any case, I may find the flags more explicit. Thoughts?

bhargavkathivarapu: @vincentqb  Only two possbilities for wave shape and interpolation mentioned in sox source code for flanger (https://github.com/rbouqueau/SoX/blob/master/src/flanger.c) ( check at bottom of file for function arguments description )

torchaudio/functional.py
bhargavkathivarapu: Original link is not longer working . Found this Archive link on google ( https://web.archive.org/web/20051125072557/http://www.harmony-central.com/Effects/effects-explained.html)  , Will submit a separate pull as it referenced in many functions

torchaudio/functional.py
vincentqb: Even though there are only two choices, let's use a string with only two values. This will be more explicit.

```
modulation (str): Use either "sinusoidal" or "triangular" modulation. (Default: ``sinusoidal``)
interpolation (str): Use either "linear" or "quadratic" for delay-line interpolation. (Default: ``linear``)
```

bhargavkathivarapu: @vincentqb  , Made the changes , all checks passed

PR #653, opened 2020-05-20
Fix io test

mthrok:
`set(["sox", "soundfile"]) not in set(["sox", "soundfile"])` is never true.

### DIFF
--- a/test/test_io.py
+++ b/test/test_io.py
@@ -171,7 +171,7 @@ class Test_LoadSave(unittest.TestCase):
         self.assertEqual(sample_rate, sample_rate2)
         os.unlink(output_path)
 
-    @unittest.skipIf(set(["sox", "soundfile"]) not in set(BACKENDS), "sox and soundfile are not available")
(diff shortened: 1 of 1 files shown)


test/test_io.py
```diff
@@ -171,7 +171,7 @@ def _test_3_load_and_save_is_identity(self):
         self.assertEqual(sample_rate, sample_rate2)
         os.unlink(output_path)
 
-    @unittest.skipIf(set(["sox", "soundfile"]) not in set(BACKENDS), "sox and soundfile are not available")
+    @unittest.skipIf(any(be not in BACKENDS for be in ["sox", "soundfile"]), "sox and soundfile are not available")
```
vincentqb, line 174: Good catch! `set(["sox", "soundfile"]).issubset(BACKENDS)` is more readable though.

mthrok: I think that's subjective.

mthrok:
thanks

PR #654, opened 2020-05-21
Add util to generate whitenoise

mthrok:
and remove I/O from torchscript test.

### DIFF
--- a/test/torchscript_consistency_impl.py
+++ b/test/torchscript_consistency_impl.py
@@ -2,6 +2,7 @@
 import unittest
 
 import torch
+import torchaudio
 import torchaudio.functional as F
 import torchaudio.transforms as T
 
@@ -64,7 +65,7 @@ class Functional(common_utils.TestBaseMixin):
         self._assert_consistency(func, tensor)
 
     def test_detect_pitch_frequency(self):
-        waveform = common_utils.get_whitenoise(sample_rate=44100)
+        waveform = common_utils.get_sinusoid(sample_rate=44100)
 
         def func(tensor):
             sample_rate = 44100
@@ -582,6 +583,6 @@ class Transforms(common_utils.TestBaseMixin):
         self._assert_consistency(T.SlidingWindowCmn(), tensor)
 
     def test_Vad(self):
-        sr = 16000
-        waveform = common_utils.get_whitenoise(sample_rate=sr)
-        self._assert_consistency(T.Vad(sample_rate=sr), waveform)
+        filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
+        waveform, sample_rate = torchaudio.load(filepath)
+        self._assert_consistency(T.Vad(sample_rate=sample_rate), waveform)
--- a/test/torchscript_consistency_impl.py
+++ b/test/torchscript_consistency_impl.py
@@ -2,7 +2,6 @@
 import unittest
 
 import torch
-import torchaudio
 import torchaudio.functional as F
 import torchaudio.transforms as T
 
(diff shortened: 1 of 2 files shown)


test/torchscript_consistency_impl.py
```diff
@@ -65,11 +64,10 @@ def func(tensor):
         self._assert_consistency(func, tensor)
 
     def test_detect_pitch_frequency(self):
-        filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
-        waveform, _ = torchaudio.load(filepath)
+        waveform = common_utils.get_whitenoise(sample_rate=16000)
```
vincentqb: This particular test is not suitable to reason with white noise.

test/torchscript_consistency_impl.py
```diff
@@ -596,15 +574,14 @@ def test_TimeMasking(self):
...
     def test_SlidingWindowCmn(self):
         tensor = torch.rand((1000, 10))
         self._assert_consistency(T.SlidingWindowCmn(), tensor)
 
     def test_Vad(self):
-        filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
-        waveform, sample_rate = torchaudio.load(filepath)
-        self._assert_consistency(T.Vad(sample_rate=sample_rate), waveform)
+        sr = 16000
+        waveform = common_utils.get_whitenoise(sample_rate=sr)
```
vincentqb, line 586: This particular test is not suitable to reason with white noise.

test/torchscript_consistency_impl.py
```diff
@@ -65,8 +64,7 @@ def func(tensor):
         self._assert_consistency(func, tensor)
 
     def test_detect_pitch_frequency(self):
-        filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
-        waveform, _ = torchaudio.load(filepath)
+        waveform = common_utils.get_whitenoise(sample_rate=44100)
```
vincentqb, line 67: This particular test is not suitable to reason with white noise.

mthrok: My reasoning is that the purpose of `torchscript_consistency_test` is not to check if the correct pitch is returned but to check if pure Python function and scripted function return the same result. The correctness check is performed in other test suits (such as `test_functional`). So as long as the input signal lies in the reasonable value range, it is practically same thing.

Also, loading the data from the file is making this test rely on I/O functionality, which is a separate unit of the library, which should be independently tested.
Looking at the implementation of `detect_pitch_frequency`, it does not have any conditional branching, so using white noise makes no difference in what this test covers.

mthrok: @vincentqb I am also working on generating sinusoid and using it elsewhere.
If you think the problem is whitenoise, then I can replace it with sinusoid


[sinusoid](https://github.com/mthrok/audio/blob/test-data2/test/common_utils.py#L157-L184), [how it replaces sinewave file](https://github.com/mthrok/audio/blob/test-data2/test/test_functional.py#L299-L303)

vincentqb: That's a fair point about decoupling. But the concern that I have is that if it's white noise, the fundamental frequency picked up might just be a constant set to the lowest value possible whether it works or not. There isn't anything that distinguishes something that would work or not :) 

sinusoid might work for pitch detection, but not for voice activity detection

mthrok: Okay, I replaced whitenoise with sinusoid for pitch and reverted vad.

PR #663, opened 2020-05-27
Apply tweaks required for running test in fbcode

mthrok:
This PR contains
 - changes to tweak for test import (relative import and addition of `__init__.py`)
 - Initialization required to use soundfile in fb infrastructure. This is harmless in other locations, yet `__init__.py` was the best place to put so I included it in public facing code base.

### DIFF
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,0 +1,10 @@
+def _init_fb_ctypes():
+    # Initiaization required only in facebook infrastructure to use soundfile
+    import libfb.py.ctypesmonkeypatch
+    libfb.py.ctypesmonkeypatch.install()
+
+
+try:
+    _init_fb_ctypes()
+except Exception:
+    pass
--- a/test/kaldi_compatibility_cpu_test.py
+++ b/test/kaldi_compatibility_cpu_test.py
@@ -1,5 +1,5 @@
-import common_utils
-from kaldi_compatibility_impl import Kaldi
+from . import common_utils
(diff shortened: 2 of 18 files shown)


mthrok:
cc @vincentqb Please review.

test/__init__.py
```diff
@@ -0,0 +1,10 @@
+def _init_fb_ctypes():
+    # Initiaization required only in facebook infrastructure to use soundfile
+    import libfb.py.ctypesmonkeypatch
+    libfb.py.ctypesmonkeypatch.install()
+
+
+try:
+    _init_fb_ctypes()
+except Exception:
```
vincentqb, line 9: nit: this would catch any exception, even ones that we may want to catch when on the fb infrastructure. i'm ok with this here for now though since this is not a regression for the community.

cpuhrsch: I'm wondering if we should make this internal only. Doesn't seem relevant to the OSS community

mthrok: > I'm wondering if we should make this internal only. Doesn't seem relevant to the OSS community

I felt the same and tried to exclude it, but 1. `__init__.py` file is required (even if it's empty) to make the imports relative, and 2. `__init__.py` is the most appropriate place to perform fbcode initialization. (otherwise we have to do it everywhere.)
So, my judgement was rather than maintaining two different `__init__.py` for OSS and internal, it's better to have the same things, because there will be a time someone working on fbcode forgets that we keep two different things and they get tangled.

PR #664, opened 2020-05-28
Separate style check from unit test

mthrok:
flake8 failure: https://app.circleci.com/pipelines/github/pytorch/audio/1930/workflows/96959f6f-f9e0-4a69-9d68-278afaa78200/jobs/48845
clang-format failure: https://app.circleci.com/pipelines/github/pytorch/audio/1931/workflows/049810da-6ea5-43c5-b95a-25478137376c/jobs/48886
both failure: https://app.circleci.com/pipelines/github/pytorch/audio/1932/workflows/da672eec-59fa-47b5-83c9-45faddda77f8/jobs/48900

### DIFF
--- a/.circleci/regenerate.py
+++ b/.circleci/regenerate.py
@@ -114,7 +114,7 @@ def unittest_workflows(indentation=6):
     jobs = []
     for os_type in ["linux", "windows"]:
         for device_type in ["cpu", "gpu"]:
-            for python_version in PYTHON_VERSIONS:
+            for i, python_version in enumerate(PYTHON_VERSIONS):
                 job = {
                     "name": f"unittest_{os_type}_{device_type}_py{python_version}",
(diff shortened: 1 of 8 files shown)


.circleci/regenerate.py
```diff
@@ -123,6 +123,14 @@ def unittest_workflows(indentation=6):
                 if device_type == 'gpu':
                     job['filters'] = gen_filter_branch_tree('master')
                 jobs.append({f"unittest_{os_type}_{device_type}": job})
+
+                if i == 0 and os_type == "linux" and device_type == "cpu":
```
vincentqb, line 127: nit: this condition is quite specific,  and could just be put outside of the loop. the order of jobs doesn't matter either. either way is readable though, so let's leave it like this :)

Issue #666, opened 2020-05-28
Undefined references when building `liblame`


mmxgn:
## 🐛 Bug

Hello all,
Thanks for maintaining torchaudio.

When trying to do `python setup.py install` from within anaconda it fails with error code 2 at the stage when building `liblame`. `make.log` in `third_party/tmp/lame-3.99.5` show errors similar to: 
```undefined reference to `tgetnum'```
 (see below for full list).

## To Reproduce

1. Activate anaconda3 environment
1. `cd` to the root `audio` directory
1. `python setup.py install`

```bash
Making all in mpglib
Making all in libmp3lame
Making all in i386
Making all in vector
Making all in frontend
libtool: link: /home/mmxgn/anaconda3/envs/tf_gpu/bin/x86_64-conda_cos6-linux-gnu-cc -Wall -pipe -fPIC -I/home/mmxgn/anaconda3/include -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z -Wl,relro -Wl,-z -Wl,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,-rpath -Wl,/home/mmxgn/anaconda3/lib -Wl,-rpath-link -Wl,/home/mmxgn/anaconda3/lib -o lame lame_main.o main.o brhist.o console.o get_audio.o lametime.o parse.o timestatus.o  -L/home/mmxgn/anaconda3/lib ../libmp3lame/.libs/libmp3lame.a -lncurses /home/mmxgn/anaconda3/lib/libiconv.so -lm -Wl,-rpath -Wl,/home/mmxgn/anaconda3/lib -Wl,-rpath -Wl,/home/mmxgn/anaconda3/lib
/home/mmxgn/anaconda3/envs/tf_gpu/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: console.o: in function `get_termcap_string.constprop.4':
console.c:(.text+0x1d): undefined reference to `tgetstr'
/home/mmxgn/anaconda3/envs/tf_gpu/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: console.o: in function `apply_termcap_settings.constprop.3':
console.c:(.text+0x80): undefined reference to `tgetent'
[...]

PR #667, opened 2020-05-28
Fix `undefined reference` errors (#666)

mmxgn:
There was an issue when building from source in an Anaconda environment under `Arch Linux` where building `liblame` would fail with `undefined reference` errors related to the `ncurses` library (`tgetstr`, `tgetnum`, `tgetent`). This PR allows `liblame` to build under this specific setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,21 @@ python setup.py install
 MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install
 ```
 
+If while building from within an anaconda environment you come across errors similar to the following:
+
+
+```
+../bin/ld: console.c:(.text+0xc1): undefined reference to `tgetnum'
+```
+
+Install `ncurses` from `conda-forge` before running `python setup.py install`:
+
+```
+# Install ncurses from conda-forge
+conda install -c conda-forge ncurses
+```
+
+
 Quick Usage
 -----------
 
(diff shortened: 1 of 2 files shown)


mthrok:
Thanks for opening PR.
Seems like the change interferes with the existing build.
Maybe we can add change the argument based on OS version.

mmxgn:
Hi,

So apparently the problem is when using the bundled anaconda `ncurses` and is fixed when using `ncurses 6-1` from `conda-forge` with:

```
conda install -c conda-forge ncurses
```
I don't know if that is specific to `Arch Linux` or not but I am happy to continue working on it with this manual fix. Might I suggest getting it into a FAQ or TROUBLESHOOTING subsection though in case it helps with others?

mthrok:
> Hi,
> 
> So apparently the problem is when using the bundled anaconda `ncurses` and is fixed when using `ncurses 6-1` from `conda-forge` with:
> 
> ```
> conda install -c conda-forge ncurses
> ```
> 
> In this case my change is not needed.
> 
> I don't know if that is specific to `Arch Linux` or not but I am happy to continue working on it with this manual fix. Might I suggest getting it into a FAQ or TROUBLESHOOTING subsection though in case it helps with others?

Hi @mmxgn 

Thanks for the report and I am glad to hear you found a simpler solution.
Yes, it sounds like a good idea to add your finding to README for the similar cases.
I think adding the subsection under [From Source](https://github.com/pytorch/audio#from-source) would do, but if you think somewhere else is a better fit, I am open to it too.

mmxgn:
Hi, thank you for your reply.

I added a short sentence for README.md  telling what someone can do if they have the same problem as me. If that's not ok it's fine to ditch the PR altogether and fix it otherwise.

mthrok:
@mmxgn 

Looks good to me. Thanks!

PR #669, opened 2020-05-29
Use environment variable for switching SoX dep

mthrok:
This PR changes
 
1. The way how to decide if codecs/SoX should be built when building torchaudio C++ extension.
    It check whether `BUILD_SOX` env is defined or not. (Originally, it was checking the existence of `.use_external_sox` file.)
2. Revert the default behavior of codecs/SoX build.
    It does not build codecs/SoX by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,18 @@ BUILD_SOX= python setup.py install
 BUILD_SOX= MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install
 ```
 
+This is known to work on a few major Linux distributions such as Ubuntu and CentOS 7 and macOS.
+If you try this on a new system and found a solution to make it work, feel free to share it by opening and issue.
+
 #### Troubleshooting:
+
+<Details><Summary>checking build system type... ./config.guess: unable to guess system type</Summary>
+
+Since the configuration file for codecs are old, they cannot correctly detect the new environments, such as Jetson Aarch. You need to replace the `config.guess` file in `./third_party/tmp/lame-3.99.5/config.guess` and/or `./third_party/tmp/libmad-0.15.1b/config.guess` with [the latest one](https://github.com/gcc-mirror/gcc/blob/master/config.guess).
+
+See also: [#658](https://github.com/pytorch/audio/issues/658)
+
+</Details>
 
 <Details><Summary>Undefined reference to `tgetnum' when using `BUILD_SOX`</Summary>
 
@@ -129,6 +140,9 @@ Install `ncurses` from `conda-forge` before running `python setup.py install`:
 # Install ncurses from conda-forge
 conda install -c conda-forge ncurses
 ```
+
+See also: [#666](https://github.com/pytorch/audio/issues/666)
+
 </Details>
 
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ to use and feel like a natural extension.
 Dependencies
 ------------
 * pytorch (nightly version needed for development)
-* libsox v14.3.2 or above (only when building from source. not required for binary installation)
+* libsox v14.3.2 or above (only required when building from source)
 * [optional] vesis84/kaldi-io-for-python commit cb46cb1f44318a5d04d4941cf39084c5b021241e or above
 
 Installation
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -93,8 +102,33 @@ python setup.py install
...
+BUILD_SOX= python setup.py install
 
+# OSX
+BUILD_SOX= MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install
+```
+
+This is known to work on a few major Linux distributions such as Ubuntu and CentOS 7 and macOS.
+If you try this on a new system and found a solution to make it work, feel free to share it by opening and issue.
+
+#### Troubleshooting:
```
mthrok, line 119: We can put this section into Wiki too.

build_tools/setup_helpers/extension.py
```diff
@@ -18,9 +19,7 @@
 _TP_BASE_DIR = _ROOT_DIR / 'third_party'
 _TP_INSTALL_DIR = _TP_BASE_DIR / 'build'
 
-# Temporary fix for building in fbcode
-# at the moment, we have to use external sox in fbcode
-_BUILD_DEPS = not (_ROOT_DIR / '.use_external_sox').exists()
+_BUILD_SOX = 'BUILD_SOX' in os.environ
```
vincentqb, line 22: I have seen more often a check for whether `BUILD_SOX=1` instead of just the existence. I could imagine someone doing `BUILD_SOX=0` and be surprise that sox is built.

README.md
```diff
@@ -27,29 +27,16 @@ to use and feel like a natural extension.
 Dependencies
 ------------
 * pytorch (nightly version needed for development)
-* libsox v14.3.2 or above
+* libsox v14.3.2 or above (only when building from source. not required for binary installation)
```
vincentqb, line 30: nit: (only required when building from source)

README.md
```diff
@@ -93,8 +102,33 @@ python setup.py install
...
+
+```bash
+# Linux
+BUILD_SOX= python setup.py install
 
+# OSX
+BUILD_SOX= MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install
+```
+
+This is known to work on a few major Linux distributions such as Ubuntu and CentOS 7 and macOS.
```
vincentqb, line 116: nit: This is known to work on linux and unix distributions such as Ubuntu and CentOS 7 and macOS.

README.md
vincentqb, line 119: nit: no need for ":"

vincentqb: The installation section should stay in the readme. I do like the folds you are adding, which avoids cluttering part of the wiki, at least when rendered such as on github. I'd say to leave troubleshooting here, with the installation steps, but I'm ok either for this.

Since we may leave some pull requests opened for troubleshooting, the troubleshooting section may be redundant, and more work to keep up-to-date. That would be a vote for just using issues for troubleshooting.

README.md
```diff
@@ -107,6 +141,10 @@ Install `ncurses` from `conda-forge` before running `python setup.py install`:
 conda install -c conda-forge ncurses
 ```
 
+See also: [#666](https://github.com/pytorch/audio/issues/666)
```
vincentqb, line 144: nit: does the link add something more?

vincentqb:
More downloads mean more instances of files that can't be downloaded when running circleci :)

vincentqb:
Closes #666?

build_tools/setup_helpers/extension.py
mthrok: So, for this kind of env var, there is no standard of what values are acceptable.
there are really a lot of variations like `0, 1, true, false, on, off, yes, no` etc...
and since this is env var, we should not really raise an error if the value is something unexpected.
I am open in defining values for our case, but my personal preference for binary behavior is just define variable or not.

vincentqb:
> > More downloads mean more instances of files that can't be downloaded when running circleci :)
> 
> I do not get what you mean. Can you elaborate ?

This was not meant as an actionable statement. :) I was just noting that if more files are downloaded when setting up circle ci, we may get more timeouts/etc.

build_tools/setup_helpers/extension.py
mthrok: I changed the behavior to expect 0/1 (and some others)

mthrok:
thanks!

Issue #597, opened 2020-04-29
Migrate Kaldi tests


vincentqb:
We used to test kaldi against a saved kaldi output [here](https://github.com/pytorch/audio/blob/master/test/test_compliance_kaldi.py), but we now have an infrastructure to run kaldi in our test environment [here](https://github.com/pytorch/audio/blob/master/test/test_kaldi_compatibility.py).

We would like to migrate the tests from ["saved kaldi"](https://github.com/pytorch/audio/blob/master/test/test_compliance_kaldi.py) to ["live kaldi"](https://github.com/pytorch/audio/blob/master/test/test_kaldi_compatibility.py).
- [ ] `test_spectrogram`
- [ ] `test_fbank`
- [ ] `test_mfcc`
- [ ] `test_mfcc_empty`
- [ ] `test_resample_waveform`
- [ ] `test_resample_waveform_upsample_size`
- [ ] `test_resample_waveform_downsample_size`
- [ ] `test_resample_waveform_identity_size`
- [ ] `test_resample_waveform_downsample_accuracy`
- [ ] `test_resample_waveform_upsample_accuracy`
- [ ] `test_resample_waveform_multi_channel`

vincentqb:
We should also have CI catch errors like #613

bhargavkathivarapu:
@vincentqb , For below kaldi  compliance tests, Do we need to write a compatibility test cases for each saved files configurations(function args in filenames)  in test/kaldi or one test configuration is enough for each
- test_spectorgram
- test_fbank ( one configuration already there in compatibility tests ) 
- test_mfcc
- test_resample_waveform

For remaining tests , they are not dependent on the saved kaldi files in test/kaldi ( though they are referencing a external .wav file , they are not comparing its contents ) , we can directly move them to compatibility without much changes

vincentqb:
The tests should cover the same cases. Is that what you meant?

bhargavkathivarapu:
> The tests should cover the same cases. Is that what you meant?

@vincentqb yes , Like for test_fbank there are 97 kaldi files , each file name is variation of kaldi  fbank function arguments 

```
Two example file names out of 97 files
- fbank-0.0939-4.5062-1.0625-0.6875-1841-true-479-5-0.84-true-true-true-true-true-true-true-false-1832-1824-1.0000-hanning.ark

[...]

PR #671, opened 2020-05-31
Migrating kaldi tests

bhargavkathivarapu:
Hi ,
Regarding  #597 , This PR moves tests from test/test_compliance_kaldi.py to test/kaldi_compatibility_impl.py


bhargavkathivarapu:
@vincentqb ,

```fbank, mfcc , spectrogram```  outputs between kaldi and torchaudio kaldi are not matching in a few cases as per the threshold defined . In other cases they are matching as per threshold .

For ```resample``` what is the kaldi command to get resampled waveform ? , for now  I kept the kaldi result of resample to  None

**Comment by mthrok** [comment]
Hi @bhargavkathivarapu 

Thanks for working on this.

First, let's put migration of different types of tests into separate PRs and start from existing one. `bank` or `mfcc`.

Second, instead of simple text file, can you use [JSON Lines format](http://jsonlines.org/examples/)? (with `.json` extension) so that no string parsing is required in test code. then we do not have to add `_get_func_args`, `_parse`, and `TEST_PREFIX`. 
JSON Lines format is one JSON object (`dict` with proper arg name and values) per line, which can be read like

```python
def _load_jsonl(path):
    with open('file.json', 'r') as file:
        return [json.loads(line) for line in file]
```

Third, when working on different types of test, can you use different parameter files? One JSON Line file for `fbank`, one for `mfcc` etc...

Forth, the current implementation stops when there is one test error. Let's try using [`parameterized.expand`](https://pypi.org/project/parameterized/) to see which ones fail. 
@vincentqb This is my temporary suggestion for parameterization. PyTorch has some discussion going on https://github.com/pytorch/pytorch/issues/11578 but this is stall. Once they implement their own parameterization, we can migrate it.

Combining the second and forth suggestion, the required change should be fairly simple.

From
```
def test_fbank(self):
```

to

```
@parameterized.expand(_load_jsonl(path_to_parameter_jsonl_file)
def test_fbank(self, kwargs):
```

test/kaldi_compatibility_impl.py
```diff
@@ -48,6 +89,8 @@ def _run_kaldi(command, input_type, input_value):
 
 
 class Kaldi(common_utils.TestBaseMixin):
+    test_8000_filepath = common_utils.get_asset_path('kaldi_file_8000.wav')
```
mthrok, line 92: Please keep this inside of test each test. The intended usage of asset file is not immediately clear when someone else work on adding a new test, so it is not very suited to be common property.

README.md
```diff
@@ -93,6 +93,21 @@ python setup.py install
...
+```
+
+Install `ncurses` from `conda-forge` before running `python setup.py install`:
+
+```
+# Install ncurses from conda-forge
+conda install -c conda-forge ncurses
+```
+
+
```
mthrok, line 110: Seems like this one is accidentally introduced.

bhargavkathivarapu: These lines are already there in audio master branch, don't know why it is showing up in diff

bhargavkathivarapu: I have created another PR only for fbank in that , this is resolved

bhargavkathivarapu:
@mthrok ,Submitted another PR ( #672 ) just for fbank , includes below 

1.  Separate JSON file for each fbank args
2.  Adding `parameterized` dependency to linux and windows yml files
3.  Using the load json and  parameterized.expand in tests 

But the CI tests are not getting trigged for that pull request 
I tried reverting the yaml to previous configuration without parametezied , but the commit is not triggering CI

PR #690, opened 2020-06-04
Clean up common_utils

mthrok:
This PR

1. ~~Removes compatibility code for `Python < 3.4` from `torchaudio.common_utils`.~~ #691 
2. Moves `IMPORT_KALDI_IO` global variable to `kaldi_io` module because it's only used there and does not seem to be useful elsewhere.
3. Moves `IMPORT_LIBROSA` global variable to test because it's only used in `test_librosa_compatibility.py`.
4. Removes `IMPORT_NUMPY` and `IMPORT_SCIPY` as they are redundant. (If `librosa` is available `numpy` and `scipy` should be available too, else installation is broken and that's not `torchaudio`'s responsibility to watch out such case.)

### DIFF
--- a/test/test_librosa_compatibility.py
+++ b/test/test_librosa_compatibility.py
@@ -7,9 +7,11 @@ import torch
 from torch.testing._internal.common_utils import TestCase
 import torchaudio
 import torchaudio.functional as F
-from torchaudio.common_utils import IMPORT_LIBROSA
+from torchaudio.common_utils import _check_module_exists
 
-if IMPORT_LIBROSA:
+LIBROSA_AVAILABLE = _check_module_exists('librosa')
+
+if LIBROSA_AVAILABLE:
     import numpy as np
     import librosa
     import scipy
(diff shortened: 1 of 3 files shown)


torchaudio/common_utils.py
```diff
@@ -1,17 +1,11 @@
 import importlib.util
 
 
-def _check_module_exists(name: str) -> bool:
+def _check_module_exists(*modules: str) -> bool:
```
vincentqb, line 4: nit: the typing for this works? If so, I like it :)

test/test_librosa_compatibility.py
```diff
@@ -115,12 +117,8 @@ def test_amplitude_to_DB(self):
 ])
 @pytest.mark.parametrize('rate', [0.5, 1.01, 1.3])
 @pytest.mark.parametrize('hop_length', [256])
+@unittest.skipIf(not LIBROSA_AVAILABLE, "Librosa not available")
```
vincentqb, line 120: Thanks for addressing [comment](https://github.com/pytorch/audio/issues/477#issuecomment-613528225)

mthrok:
thanks!

Issue #554, opened 2020-04-16
_audio_backends platform independent


cpuhrsch:
## 🐛 Bug

The constant [_audio_backends](https://github.com/pytorch/audio/blob/c29598d54185d73b4ed04103330573e190bbdb69/torchaudio/_backend.py#L12) is referenced by [BACKENDS](https://github.com/pytorch/audio/blob/0e5581cb2a9616205a00cbabf4c9a30613a1037f/test/common_utils.py#L10) in the common utilities of the test folder. [test_batch_mfcc](https://github.com/pytorch/audio/blob/0e5581cb2a9616205a00cbabf4c9a30613a1037f/test/test_batch_consistency.py#L181) is skipped if the 'sox' key is not present in that constant, but it always is. That means this test will be executed in environments where the package may not exist.

```
(base) PS C:\Users\chris\dev\audio> python .\test\test_batch_consistency.py TestTransforms.test_batch_mfcc
E
======================================================================
ERROR: test_batch_mfcc (__main__.TestTransforms)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\tools\Anaconda3\lib\contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File ".\test\test_batch_consistency.py", line 185, in test_batch_mfcc
    waveform, _ = torchaudio.load(test_filepath)
  File "c:\users\chris\dev\audio\torchaudio\__init__.py", line 87, in load
    filetype=filetype,
  File "c:\users\chris\dev\audio\torchaudio\_sox_backend.py", line 38, in load
    import _torch_sox
ModuleNotFoundError: No module named '_torch_sox'

----------------------------------------------------------------------
Ran 1 test in 0.001s

FAILED (errors=1)
```

## To Reproduce

1. Remove sox from your environment
2. Run ```python test/test_batch_consistency.py TestTransform.test_batch_mfcc```

I can provide more detailed information if required.

## Expected behavior

The test should be skipped if sox is not available.

## Environment

```
(base) PS C:\Users\chris\dev\audio> python .\collect_env_1.py
Collecting environment information...
PyTorch version: 1.6.0a0+8a60d8b
Is debug build: No
CUDA used to build PyTorch: None

OS: Microsoft Windows 10 Home
GCC version: Could not collect
CMake version: version 3.14.0

Python version: 3.7
Is CUDA available: No
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
[...]

PR #692, opened 2020-06-04
Base guard logic on module availability

mthrok:
The current implementation of `_audio_backend_guard` checks if the current backend is the desired one.
However as reported in https://github.com/pytorch/audio/issues/554, this does not reflect the actual installation.

A better approach is to check if the required modules to invoke a function are available at runtime.
The utility function `_check_module_exists` (renamed to `is_module_available`) we have in `common_utils` utilized `importlib.util.find_spec` and it allows to check module availability without import, so we can leverage this. 

This PR implements `requires_module` decorator which replaces the original `_audio_backend_guard`, which checks the module availability at import time, instead of current backend value. (Note that it is safe to use functions from `soundfile` backend when default backend is `sox` backend and vice versa).

This approach can also replace how functions in `torchaudio.kaldi_io` raise `ImportError` manually.

Also combining, `is_module_available` and `requires_module`, we can import optional modules (`kaldi_io`, `soundfile` and `torchaudio._torchaudio`) at module level, not inside of functions.

Example;

with proper installation of `torchaudio`
```
$ python -c 'import torchaudio;print(torchaudio.get_sox_bool(0))'
sox_bool.sox_false
```

if I remove `torchaudio._torchaudio` to imitate windows env;
```
$ rm torchaudio/_torchaudio.so
$ python -c 'import torchaudio;print(torchaudio.get_sox_bool(0))'
  File "<string>", line 1, in <module>
  File "/scratch/moto/torchaudio/torchaudio/common_utils.py", line 33, in wrapped
    raise RuntimeError(f'{func.__module__}.{func.__name__} requires {req}')
RuntimeError: torchaudio.get_sox_bool requires module: torchaudio._torchaudio
```


torchaudio/_internal/module_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    r"""Returns if a top-level module with :attr:`name` exists *without**
+    importing it. This is generally safer than try-catch block around a
+    `import X`. It avoids third party libraries breaking assumptions of some of
+    our tests, e.g., setting multiprocessing start method when imported
+    (see librosa/#747, torchvision/#544).
+    """
+    return all(importlib.util.find_spec(m) is not None for m in modules)
+
+
+def requires_module(*modules: str):
```
vincentqb, line 15: The new decorator reads a little different
```
@_audio_backend_guard("sox")
```
The former makes it clear that we are requiring "sox" in particular, whereas the latter simply requires `torchaudio._torchaudio` which could give the impression this requires more, and create confusion later (even though `torchaudio._torchaudio` currently only contains "sox"). Options I see here:
* the name `torchaudio._torchaudio` is not precise enough, and should be updated.
* we move the sox specific dependency in a module properly named
* we translate a more explicit name in `requires_module` mapping "sox" to the corresponding unique module.

Thoughts?

torchaudio/_internal/module_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    """Decorate function to give error message if invoked without required optional modules.
+
+    This decorator is to give better error message to users rather
+    than raising ``NameError:  name 'module' is not defined`` at random places.
+    """
+    missing = [m for m in modules if not is_module_available(m)]
+
+    if not missing:
+        # fall through. If all the modules are available, no need to decorate
+        def decorator(func):
```
vincentqb, line 25: nit: Can you confirm this code will preserve the docstrings of the wrapped functions?

torchaudio/_internal/module_utils.py
vincentqb: Yup, looks like we are on the same page.

I do agree that the previous solution was not ideal, and I am not advocating maintaining it :) It's only upside that I see was to say "sox" somewhere on the line.

That being said, I did not "request changes" to the pull request, because I also agree with you that now was likely not the time to fix this :)

mthrok:
Thanks!

PR #693, opened 2020-06-04
Move sox effect init/shut down to sox_effects module

mthrok:
For better [cohesion and coupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)#:~:text=Coupling%20versus%20cohesion,-Coupling%20and%20cohesion&text=Coupling%20refers%20to%20the%20interdependencies,as%20the%20module%20becomes%20large.), move sox_effects related functions to sox_effects module.
Then perform initialization in `__init__.py`.

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -1,4 +1,3 @@
-import atexit
 import os.path
 from pathlib import Path
 from typing import Any, Callable, Optional, Tuple, Union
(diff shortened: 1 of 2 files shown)


torchaudio/__init__.py
```diff
@@ -29,6 +29,7 @@
 
 if _mod_utils.is_module_available('torchaudio._torchaudio'):
     from . import _torchaudio
+    initialize_sox()
```
vincentqb, line 32: nit: We were not initializing sox at import time. I don't see a downside, so I'll just leave a note.

mthrok: Yup. 
With `is_module_available` and less function definition in `__init__.py`, we can actually do what `__init__.py` is supposed to do. Initialization :) .

mthrok:
Thanks!

PR #694, opened 2020-06-04
Move I/O related Tensor ops to misc_ops module

mthrok:
`check_input` and `audio_normalization` functions defined in `__init__.py`, `_backend.py` and `_soundfilebackend` (duplicated) are referred from multiple modules and cause cyclic module dependencies and preventing modularizing backends. This PR resolve this by putting them in a separate dedicated module `_internal.misc_ops`.

Also change the helper function name from `audio_normalization` (noun) to `normalize_audio` (verb) for consistency.

### DIFF
--- a/torchaudio/_internal/misc_ops.py
+++ b/torchaudio/_internal/misc_ops.py
@@ -1,0 +1,30 @@
+from typing import Union, Callable
+
+import torch
+from torch import Tensor
+
+
+def normalize_audio(signal: Tensor, normalization: Union[bool, float, Callable]) -> None:
+    """Audio normalization of a tensor in-place.  The normalization can be a bool,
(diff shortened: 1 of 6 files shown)


torchaudio/_internal/misc_ops.py
```diff
@@ -0,0 +1,30 @@
+from typing import Union, Callable
```
vincentqb, line 1: nit: That's a very generic name for the file :)

mthrok: Yeah, I do not know what to call, but this is just evicting these functions from `__init__.py` so if this module glows into something then we can have a better naming.

mthrok:
Thanks!

PR #695, opened 2020-06-04
Move SoX backend specific ops to _sox_backend module

mthrok:


mthrok:
Thanks for the world-record approval speed.

PR #696, opened 2020-06-04
Apply module availability guard to soundfile backend

mthrok:


torchaudio/_soundfile_backend.py
```diff
@@ -1,10 +1,16 @@
...
-from typing import Any, Optional, Tuple, Union
+from typing import Any, Optional, Tuple
 
 import torch
 from torch import Tensor
 
-from torchaudio._internal import misc_ops as _misc_ops
+from torchaudio._internal import (
+    module_utils as _mod_utils,
+    misc_ops as _misc_ops,
```
vincentqb, line 9: nit: this was like this before, but this is not used in the file at all?

mthrok: both of them are used. otherwise flake8 job would fail.

`_misc_ops.check_input(src)` is used in `save` function.

mthrok:
Thanks!

torchaudio/_soundfile_backend.py
vincentqb: good point :)

PR #697, opened 2020-06-04
List backends dynamically based on availability

mthrok:


torchaudio/_backend.py
```diff
@@ -1,43 +1,49 @@
...
+_BACKEND = None
+_BACKENDS = {}
 
-from . import _soundfile_backend, _sox_backend
+if _mod_utils.is_module_available('soundfile'):
+    _BACKENDS['soundfile'] = _soundfile_backend
+    _BACKEND = 'soundfile'
+if _mod_utils.is_module_available('torchaudio._torchaudio'):
+    _BACKENDS['sox'] = _sox_backend
+    _BACKEND = 'sox'
```
vincentqb, line 14: nit: here, "sox" would become the default since it's the last one selected

mthrok: Yes, that's the intension but it was a bit unusual so I updated to use regular `if 'sox' else if 'soundfile'`, which is easier to read.

mthrok:
Thanks!

PR #698, opened 2020-06-05
Refactor backend and not rely on global variables on switching

mthrok:
1. Do not rely on global variables for backend switch
   So that load/save/info/load_wav functions will be torchscript-able
2. Add no_backend module to for the case there is no backend module available
   [bonus] This allows the whole codebase importable on systems that do not have torchaudio C++ extension nor soundfile.

### DIFF
--- a/test/test_datasets.py
+++ b/test/test_datasets.py
@@ -29,11 +29,13 @@ class TestDatasets(unittest.TestCase):
         data[0]
 
     @unittest.skipIf("sox" not in common_utils.BACKENDS, "sox not available")
+    @common_utils.AudioBackendScope('sox')
     def test_commonvoice(self):
         data = COMMONVOICE(self.path, url="tatar")
         data[0]
 
     @unittest.skipIf("sox" not in common_utils.BACKENDS, "sox not available")
+    @common_utils.AudioBackendScope('sox')
     def test_commonvoice_diskcache(self):
         data = COMMONVOICE(self.path, url="tatar")
         data = diskcache_iterator(data)
@@ -43,6 +45,7 @@ class TestDatasets(unittest.TestCase):
         data[0]
 
     @unittest.skipIf("sox" not in common_utils.BACKENDS, "sox not available")
+    @common_utils.AudioBackendScope('sox')
     def test_commonvoice_bg(self):
         data = COMMONVOICE(self.path, url="tatar")
         data = bg_iterator(data, 5)
--- a/torchaudio/backend/utils.py
+++ b/torchaudio/backend/utils.py
@@ -1,62 +1,74 @@
+"""Defines utilities for switching audio backends"""
 import warnings
-from typing import Any, Optional
+from typing import Optional, List
 
-from torchaudio._internal import module_utils as _mod_utils
-from . import soundfile_backend, sox_backend
+import torchaudio
+from torchaudio._internal.module_utils import is_module_available
+from . import (
+    no_backend,
+    sox_backend,
+    soundfile_backend,
+)
 
-_BACKEND = None
-_BACKENDS = {}
+__all__ = [
+    'list_audio_backends',
+    'get_audio_backend',
+    'set_audio_backend',
+]
 
 
-def list_audio_backends():
-    return list(_BACKENDS.keys())
+def list_audio_backends() -> List[str]:
+    """List available backends"""
+    backends = []
+    if is_module_available('soundfile'):
+        backends.append('soundfile')
+    if is_module_available('torchaudio._torchaudio'):
+        backends.append('sox')
+    return backends
 
 
-def set_audio_backend(backend: str) -> None:
-    """
-    Specifies the package used to load.
+def set_audio_backend(backend: Optional[str]) -> None:
+    """Set the backend for I/O operation
+
     Args:
         backend (str): Name of the backend. One of "sox" or "soundfile",
             based on availability of the system.
     """
-    if backend not in _BACKENDS:
+    if backend is not None and backend not in list_audio_backends():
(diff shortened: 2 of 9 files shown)


test/test_datasets.py
```diff
@@ -29,11 +29,13 @@ def test_librispeech(self):
         data[0]
 
     @unittest.skipIf("sox" not in common_utils.BACKENDS, "sox not available")
+    @common_utils.AudioBackendScope('sox')
```
mthrok, line 32: See #715

torchaudio/backend/utils.py
```diff
@@ -1,62 +1,74 @@
...
+
+def get_audio_backend() -> Optional[str]:
+    """Get the name of the current backend"""
+    if torchaudio.load == no_backend.load:
+        return None
+    if torchaudio.load == sox_backend.load:
+        return 'sox'
+    if torchaudio.load == soundfile_backend.load:
+        return 'soundfile'
+    raise ValueError('Unknown backend.')
```
vincentqb, line 74: nit: `f'Unexpected backend "{backend}"'` ?

torchaudio/backend/utils.py
```diff
@@ -1,62 +1,74 @@
...
 
-def _get_audio_backend_module() -> Any:
-    """
-    Gets the module backend to load.
-    """
-    if _BACKEND is None:
-        raise RuntimeError('Backend is not initialized.')
-    return _BACKENDS[_BACKEND]
+    for func in ['save', 'load', 'load_wav', 'info']:
+        setattr(torchaudio, func, getattr(module, func))
```
vincentqb, line 52: Let's just make sure we make that conscious choice: `torchaudio.backend.__init__` invokes `utils._init_audio_backend()` which then modifies `torchaudio`.

Would it be advantageous to attach `load`, `save`, etc directly in `torchaudio.__init__` instead?

test/test_datasets.py
vincentqb: Does this fixes #715 completely?

mthrok: Nope. This is temporary fix. #719 gives principled fix which rely on this PR.

torchaudio/backend/utils.py
mthrok: This state is not reachable and if it does that means user assigned something else to `torchaudio.load` and we have no idea what it is so, there is no `backend` variable we can get.

torchaudio/backend/utils.py
mthrok: I am not sure what you mean by `advantageous` here, but the current approach gives cleaner separation of `torchaudio.__init__` and `torchaudio.backend` because `backend` module provides 1. backend implementations, 2. way to manipulate backend, and 3. way to fetch information about current backend and available backends. This way, all the responsibilities on backend is closed within `torchaudio.backend` module and the only contract between `torchaudio.__init__` and `torchaudio.backend` is that `torchaudio.load/save/load_wav/info` attributes are reserved for `torchaudio.backend` module (implying that as long as `torchaudio.__init__` does not touch these attributes, `torchaudio.backend` understands everything about it), and these attributes are the only attributes outside of `torchaudio.backend` that `torchaudio.backend` module is allowed to touch.

If the initialization is moved to `torchaudio.__init__`, that means there are more contract between `torchaudio.__init__` and `torchaudio.backend`, like how the backend switch function works, and it requires `torchaudio.__init__` to know the internal mechanics of `torchaudio.backend`. This introduces more coupling between `torchaudio.__init__` and `torchaudio.backend` and adds maintenance cost.

vincentqb: My main concern was the type of coupling that led to #712, when one module modifies another.

If I understand what you are saying: in order to avoid having the two depending on each other, we would need to bring `torchaudio.backend` completely within `torchaudio/__init__.py`. Is that what you are saying?

btw I am not advocating this :)

mthrok: > My main concern was the type of coupling that led to #712, when one module modifies another.

That's different. #712 was about new classes being defined inside of a module when a module function called from outside call.
Refactoring of backend switch is about achieving the proper [Separation on Concerns](https://en.wikipedia.org/wiki/Separation_of_concerns), by letting the `backend` module handle everything about backend, including initialization, so that main `torchaudio` module does not need to.

> If I understand what you are saying: in order to avoid having the two depending on each other, we would need to bring `torchaudio.backend` completely within `torchaudio/__init__.py`. Is that what you are saying?

No, that's totally opposite of what I am suggesting. It's better maitainable if `torchaudio/__init__.py` knows nothing about `backend` internal mechanism.

vincentqb: Let me rephrase my concern. If I read `__init__`, I don't have any way of knowing that `load/save/etc` are being added to it since it's added by another module. Am I missing something?

How about setting the attribute to `torchaudio.backend` and simply importing in `torchaudio/__init__.py`? Thoughts?

mthrok: > Let me rephrase my concern. If I read `__init__`, I don't have any way of knowing that `load/save/etc` are being added to it since it's added by another module. Am I missing something?

Yes, that's the correct way to achieve the Separate of Concerns.

> How about setting the attribute to `torchaudio.backend` and simply importing in `torchaudio/__init__.py`? Thoughts?

No that way, the responsibility of backend switching is spread to `torchaudio/__init__.py` and `torchaudio.backend`, which introcudes more coupling. That's anti-pattern.

Just adding some comment about `save/load/info` assigned in `torchaudio.backend` would suffice.

mthrok:
thanks!

PR #699, opened 2020-06-05
Migrate kaldi resample accuracy and size tests

bhargavkathivarapu:
This PR migrates the below tests (#597  ) from test/test_compliance_kaldi.py to test/kaldi_compatibility_impl.py

I am migrating all minor tests in this  PR, which are not dependent on other files for arguments .

- resample accuracy 
- resample size 
- mfcc empty input check

### DIFF
--- a/test/kaldi_compatibility_impl.py
+++ b/test/kaldi_compatibility_impl.py
@@ -3,6 +3,7 @@ import json
 import shutil
 import unittest
 import subprocess
+import math
 
 import kaldi_io
 import torch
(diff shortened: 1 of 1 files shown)


mthrok:
Hi @bhargavkathivarapu 

Thanks for working on this. 
Let me get back to this later as, unlike the other tests we've migrated, Kaldi does not provide a simple CLI for this.
I have to study the code and come up with the right approach. I feel the existing tests require improvements.

Note:
Kaldi's implementation is found [here](https://github.com/kaldi-asr/kaldi/blob/f2a89c232a0932a6cc79cb786990ef340ca131ef/src/feat/resample.cc#L368), and the corresponding test is found [here](https://github.com/kaldi-asr/kaldi/blob/master/src/feat/resample-test.cc#L147) and [here](https://github.com/kaldi-asr/kaldi/blob/master/src/feat/resample-test.cc#L229)

mthrok:
We have replaced the resample module in 0.9 release and Kaldi's resample code is no longer the reference point.
Closing the issue. Thanks @bhargavkathivarapu and sorry for the unfruitful collaboration here. Please feel free to pick up an any `help-wanted` task in the future. Now the team behind torchaudio has grown, we should be able to better respond.

PR #706, opened 2020-06-08
Put structs common to backends in backend.common

mthrok:
It is strange that `sox_backend` depends on custom classes defined in `soudfile_backend`. This PR introduces `backend.common` module where these custom classes are defined and have both `sox_backend` and `soundfile_backend` depend on it.


mthrok:
Thanks!

PR #708, opened 2020-06-09
Add sox_effects submodule and delegate sox_effects initialization

mthrok:


torchaudio/sox_effects/sox_effects.py
```diff
@@ -48,13 +47,13 @@ def initialize_sox() -> int:
...
             _SOX_INITIALIZED = True
-            atexit.register(shutdown_sox)
+            atexit.register(shutdown_sox_effects)
         return code
     return _SOX_SUCCESS_CODE
 
 
 @_mod_utils.requires_module("torchaudio._torchaudio")
-def shutdown_sox() -> int:
+def shutdown_sox_effects() -> int:
```
vincentqb, line 56: that's BC breaking, no? `torchaudio.sox_effects.initialize_sox()`

torchaudio/sox_effects/sox_effects.py
```diff
@@ -11,7 +10,7 @@
 )
 
 if _mod_utils.is_module_available('torchaudio._torchaudio'):
-    from . import _torchaudio
+    from torchaudio import _torchaudio
```
vincentqb, line 13: nit: that's a separate issue :)

mthrok:
> Can you detail what you are trying to do here?

There are couple of aspects of this PR that overall improves the maintainability of the code base, based on "decoupling" and "separation of concerns".

As you know, `sox_effects` functionalities are either available or unavailable. From the viewpoint of `torchaudio` main module, the looser the connection between the `torchaudio` module and `torchaudio.sox_effects`, the more manageable the code base become because you can change the two modules independently. This is mostly accomplished when the definitions of `initialize_sox` and `shutdown_sox` were moved from `torchaudio.__init__` to `torchaudio.sox_effects`, but the initialization of sox effects are still happening in `torchaudio.__init__`. If we move the initialization to `sox_effects` module, the responsibility of sox initialization is moved to `sox_effects` module, along with the required module availability check etc. The main `torchaudio` module can be carefree about how the `sox_effects` module should work.

In addition to that, I found that `initialize_sox` and `shutdown_sox` are confusing because it sound like they are required for `libsox` based I/O. To make it clear, I renamed them to include `sox_effect` in function name.

Also moving functions from the original places are BC breaking itself, therefore, these functions are re-imported in `torchaudio.__init__` and renamed to match the original names. Therefore the PR is not BC breaking.

In the next PR https://github.com/pytorch/audio/pull/709/commits/9519310c4096537d9daa7481c6c10b9c648d8d25, I am gonna mark the `initialize_sox` and `shutdown_sox` deprecated, so that the decoupling is completed.

Issue #550, opened 2020-04-14
Which datasets should torchaudio have?


vincentqb:
Which new datasets should we offer and prioritize in torchaudio?

I want to follow-up on #31 and a few of the recent PRs. Instead of aiming to have an exhaustive list of datasets, we should focus on a few important/common/representative dataset that can serve as templates for users to easily implement datasets of their choosing. All datasets should already be free/accessible/online/common with license permitting linking to them.

torchaudio currently has:
1. commonvoice
1. librispeech
1. ljspeech
1. speechcommands
1. vctk
1. yesno

Current open proposals:
* Free Universal Sound Separation (FUSS) #534
* CMU_ARCTIC #512

cyrta:
I would recommend to review list:
- https://github.com/JRMeyer/open-speech-corpora
- https://github.com/jim-schwoebel/voice_datasets
- https://wiki.inria.fr/rosp/Datasets
 - https://www.ismir.net/resources/datasets/
 - http://www.audiocontentanalysis.org/data-sets/

My suggestion are
- TED-LIUM 1-3 http://www.openslr.org/51/
- Spoken Wikipedia Corpus https://nats.gitlab.io/swc/
- CHiME 5 http://spandh.dcs.shef.ac.uk/chime_challenge/data.html
- AMI
- ICSI Meeting Corpus 
- LibriTTS http://www.openslr.org/60/

from noise
- NOIZEUS
- MUSAN
- ESC-50  https://github.com/karolpiczak/ESC-50

from music
- [AcousticBrainz-Genre](https://multimediaeval.github.io/2018-AcousticBrainz-Genre-Task/)
- [GTZAN](http://marsyas.info/downloads/datasets.html) 

All of them are considered onces you do benchmark evaluation for many tasks.

cHemingway:
For Noise, some thoughts:

- [NOIZEUS](https://ecs.utdallas.edu/loizou/speech/noizeus/) is already noisy speech, ideally a "pure" noise dataset should be used instead? However it is very popular, so possibly worth including for that reason.

- MUSAN looks good, has proper licensing, and is well cited, but I am unsure if it is sufficiently "varied", there is not much description of the contents of its noise in the paper, or how much care was made to make it representatives. 

- ESC-50 seems to meet all the requirements quite nicely, and is directly accessible as a GitHub zip download. It is aimed at Environmental Sound Classification however, and not speech processing.

[...]

PR #710, opened 2020-06-09
Add cmu_arctic dataset

### LINKED ISSUE #512: Add CMU_ARCTIC datasets
## 🚀 Feature
As for now, there is no such dataset in torchaudio.datasets. 
More information about it can be found on http://www.festvox.org/cmu_arctic/packed/

jimchen90:
Add the CMU_ARCTIC dataset used [here](https://github.com/r9y9/wavenet_vocoder).

Related to #550 
Fixes #512

### DIFF
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -26,7 +26,7 @@ All the datasets have almost similar API. They all have two common arguments:
 
 
 CMUARCTIC
-~~~~~
+~~~~~~~~~
 
 .. autoclass:: CMUARCTIC
   :members: __getitem__


torchaudio/datasets/cmu_arctic.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Load audio
+    waveform, sample_rate = torchaudio.load(file_audio)
+
+    # Load text
+    with open(file_text) as ft:
+        for line in ft:
+            file_id, utterance = line.strip().split(" ", 2)[1:]
+            if fileid == file_id:
+                utterance = utterance[1:-3]
```
vincentqb, line 73: Can you add a quick comment to document this line?
```
# clean "(utterance )" to "utterance"
```

torchaudio/datasets/cmu_arctic.py
```diff
@@ -0,0 +1,155 @@
...
+    file_text = os.path.join(path, "etc", ext_txt)
+
+    file_audio = os.path.join(path, "wav", utterance_id + ext_audio)
+
+    # Load audio
+    waveform, sample_rate = torchaudio.load(file_audio)
+
+    # Load text
+    with open(file_text) as ft:
+        for line in ft:
```
vincentqb, line 70: How long is the file? How many utterance does one contain? We should avoid having to scan the whole file for each data point if possible.

If there's only one text file with all the utterances, you can follow [commonvoice](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/commonvoice.py#L189) as a template.

torchaudio/datasets/cmu_arctic.py
```diff
@@ -0,0 +1,155 @@
...
+            "cmu_us_fem_arctic",
+            "cmu_us_gka_arctic",
+            "cmu_us_jmk_arctic",
+            "cmu_us_ksp_arctic",
+            "cmu_us_ljm_arctic",
+            "cmu_us_lnh_arctic",
+            "cmu_us_rms_arctic",
+            "cmu_us_rxr_arctic",
+            "cmu_us_slp_arctic",
+            "cmu_us_slt_arctic"
```
vincentqb, line 120: The abbreviations are all `cmu_us_???_arctic`, right? If so, let's just use `???` as the code here.

torchaudio/datasets/cmu_arctic.py
```diff
@@ -0,0 +1,155 @@
...
+        waveform,
+        sample_rate,
+        utterance,
+        utterance_id.split("_")[1]
+    )
+
+
+class CMU_ARCTIC(Dataset):
+    """
+    Create a Dataset for CMU_Arctic. Each item is a tuple of the form:
```
vincentqb, line 89: nit: no need for underscore

torchaudio/datasets/cmu_arctic.py
jimchen90: There is one text file with all the utterances. Yes, I will change it by following [commonvoice](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/commonvoice.py#L189).

torchaudio/datasets/cmu_arctic.py
jimchen90: Yes, only three letter are different. I will change it.

torchaudio/datasets/cmu_arctic.py
jimchen90: I will change it. Thanks.

torchaudio/datasets/cmu_arctic.py
jimchen90: I will change it. Thanks.

PR #712, opened 2020-06-10
Change parameterized testing system to be compatible with unittest

mthrok:
Summary: The previous implementation of parameterized testing worked by modifying test.common_utils inplace.  This doesn't work in general because unittest's contract with test modules is such that it must be able to load the module and run the test itself.  Because the previous implementation needed to load the module and modify it, it is incompatible.

Reviewed By: mthrok

Differential Revision: D21964676

### DIFF
--- a/test/common_utils.py
+++ b/test/common_utils.py
@@ -119,6 +119,15 @@ def define_test_suites(
                 scope[t.__name__] = t
 
 
+def common_test_class_parameters(
+    dtypes: Iterable[str] = ("float32", "float64"),
+    devices: Iterable[str] = ("cpu", "cuda"),
+):
+    for device in devices:
+        for dtype in dtypes:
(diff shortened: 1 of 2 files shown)


vincentqb:
> Summary: The previous implementation of parameterized testing worked by modifying test.common_utils inplace. This doesn't work in general because unittest's contract with test modules is such that it must be able to load the module and run the test itself. Because the previous implementation needed to load the module and modify it, it is incompatible.

Can you give a snippet of the error we were getting on the internal side?

vincentqb:
Are there other cases behaving like this that use `define_test_suites`?

mthrok:
> Are there other cases behaving like this that use `define_test_suites` and `globals`?

Ya, I am working to remove them now.

mthrok:
Thanks

PR #716, opened 2020-06-11
Get rid of dynamic test suite generation

mthrok:
`type` used in `common_utils` generates test class definition in `common_utils` and
this modifies the module state after it's imported. This is anti-pattern.
This PR get rid of the related utility functions and define test suite manually.

Follow up #712


mthrok:
Thanks!

PR #718, opened 2020-06-12
Add SignalInfo typedef, and extension module

mthrok:
This is a part of PRs to add new "sox_io" backend. https://github.com/pytorch/audio/pull/726

This PR adds `SignalInfo` structure, which is data exchange interface between Python and C++ in coming TorchScript-based sox IO backend.
For the case, where C++ extension is not available (i.e. Windows), this PR also adds dummy class and module that will be substituted.
This logic is implemented in `torchaudio.extension` moduel.

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -1,3 +1,4 @@
+from . import extension
 from torchaudio._internal import module_utils as _mod_utils
 from torchaudio import (
     compliance,
--- a/torchaudio/csrc/register.cpp
+++ b/torchaudio/csrc/register.cpp
@@ -1,0 +1,18 @@
+#ifndef TORCHAUDIO_REGISTER_H
+#define TORCHAUDIO_REGISTER_H
+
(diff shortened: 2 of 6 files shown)


mthrok:
> I just wonder if you think we should have tests that confirm that the dummy and the module behaves as expected.
For now, the dummy module's only purpose is to make import work in the system that does not have C++ extension. Starting from #728, without dummy, annotation would fail and it is not possible to `import torchaudio` in Windows system. The fact that Windows can still `import torchaudio` serves as a test for checking that dummy module fulfills its purpose.

If we decided to use `SignalInfo` class from dummy module in `soundfile` backend in Windows, then that's when the detail of Dummy module starts to matter, in that case, by adding tests specific to `soundfile`, then Windows CI jobs will catch if the dummy module is implemented correctly.

> Since "sox_io" will be used in place where we are currently using "sox", I believe these tests will cover what we need. Thoughts?
Can you elaborate `these tests` and `what we need`?

mthrok:
Thanks.

PR #705, opened 2020-06-08
Add MelResNet Block

jimchen90:
This MelResNet block is part of [WaveRNN model ](https://github.com/G-Wang/WaveRNN-Pytorch/blob/master/model.py#L29). Now the test is only for the input and output of this block. Other tests will be added after other blocks are combined.
Related to #446

Stack:
1. **Add MelResNet Block #705**
2. Add Upsampling Block (ongoing)
3. Add WaveRNN Model (ongoing)
4. Add example pipeline with WaveRNN (ongoing)

PR #725, opened 2020-06-17
MelResNet spacing and typing adjustment

jimchen90:
This is to adjust the spacing and typing of melresnet block #705 .

### DIFF
--- a/torchaudio/models/_wavernn.py
+++ b/torchaudio/models/_wavernn.py
@@ -1,5 +1,3 @@
-from typing import List
-
 from torch import Tensor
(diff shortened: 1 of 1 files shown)


torchaudio/models/_wavernn.py
```diff
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import List
```
vincentqb, line 1: nit: list is not used here (yet), so let's not add it

jimchen90: It has been removed.

Issue #618, opened 2020-05-08
Info length and rate returns different values for different backends


faroit:
## 🐛 Bug

`torchaudio.info` returns the info objects directly from the respective backend. Due to same property naming, users might forget to check how the metadata is calculated. This results in metadata being reported differently depending on which backend is reported.
E.g. sox calculates the `length` summed across channels whereas `soundfile` does this per channel (correct)

I would propose to add wrapper for the info objects that - independent of the backend - the most important metadata (`length` and `rate`) is identical.

Currently, the sox backend reports a missleading `length` and the `rate` parameter is of type `float` instead of `int`.

## To Reproduce

```python
path =  "any/wavfile.wav"
# soundfile
torchaudio.set_audio_backend("soundfile")
info = torchaudio.info(path)
print(si.length)
print(type(si.rate))

# sox
torchaudio.set_audio_backend("sox")
info = torchaudio.info(path)
print(si.length)
print(type(si.rate))
```

## Expected behavior

`soundfile` reports the correct metadata, `sox` should be corrected so that:

```python
# sox
torchaudio.set_audio_backend("sox")
info = torchaudio.info(path)
print(si.length // si.channels)
print(int(si.rate))
```

## Environment

torchaudio==0.5.0 from pypi

vincentqb:
Relates to #236?

vincentqb:
I agree that both backend should return the same values. We need to make sure we document the change of either so that users are aware of any changes. Thoughts?

mthrok:
I could reproduce this;

<Details><Summary>code</Summary>

```
import torchaudio

path = "test/assets/vad-go-stereo-44100.wav"

# soundfile
torchaudio.set_audio_backend("soundfile")
si, _ = torchaudio.info(path)
print(si.length)
print(type(si.rate))

# sox
torchaudio.set_audio_backend("sox")
si, _ = torchaudio.info(path)
print(si.length)
print(type(si.rate))
```

</Details>

```
43339
<class 'int'>
86678
<class 'float'>
```

mthrok:
I checked #236, and it does not fix the issue.
I think it's more appropriate to handle it on Python side, as suggested.
(Copy data from sox backend C++ struct to `SignalInfo` Python class and apply the fix there.)
That way, `info` function returns the same class regardless of backend.

faroit:
>  That way, info function returns the same class regardless of backend.

[...]

Issue #430, opened 2020-02-08
Save / Load Bugs


PetrochukM:
## 🐛 Bug

The process of saving and loading a file causes it to change significantly up to the 1st decimal point.

## To Reproduce

```python
import torch
import torchaudio

signal = torch.clamp(torch.randn(24000), -1.0, 1.0)
print('Signal Stats:', signal.max(), signal.min(), signal.std(), signal.mean())

torchaudio.save('wavefile.wav', signal, 24000)
loaded_signal, _ = torchaudio.load('wavefile.wav')
print('Loaded Signal Stats:', loaded_signal.max(), loaded_signal.min(), loaded_signal.std(),
      loaded_signal.mean())
```

Example output:
```bash
(venv) OS-XImage-9:Text-to-Speech michaelp$ python foo.py 
Signal Stats: tensor(1.) tensor(-1.) tensor(0.7185) tensor(0.0013)
Loaded Signal Stats: tensor(0.9998) tensor(-1.) tensor(0.6444) tensor(-0.3177)
```

Notice that the mean and standard deviation of the signal has significantly changed.

## Expected behavior

The WAV format should save and load without any loss in precision.

PetrochukM:
Another issue that I found is that if you load a 16-bit PCM file with the load function returns a `torch.Float32` which does not match the `dtype` of the file.

vincentqb:
This appears to be a problem with sox. Changing the backend to `soundfile` fixes the problem
```python
import torch
import torchaudio

torchaudio.set_audio_backend("sox")  # default

signal = torch.clamp(torch.randn(24000), -1.0, 1.0)
print('Signal Stats:', signal.max(), signal.min(), signal.std(), signal.mean())

torchaudio.save('wavefile.wav', signal, 24000)
loaded_signal, _ = torchaudio.load('wavefile.wav')
print('Loaded Signal Stats:', loaded_signal.max(), loaded_signal.min(), loaded_signal.std(), loaded_signal.mean())

# Signal Stats: tensor(1.) tensor(-1.) tensor(0.7190) tensor(0.0057)
# Loaded Signal Stats: tensor(0.9998) tensor(-1.) tensor(0.6426) tensor(-0.3226)

print(torch.norm(signal-loaded_signal, p=1))
# tensor(7880.0034)

torchaudio.set_audio_backend("soundfile")                                                  

signal = torch.clamp(torch.randn(24000), -1.0, 1.0)
[...]

Issue #400, opened 2020-01-10
Fbank features are different from Kaldi Fbank


jooan84:
## 🐛 Bug

The output of the fbank feature calculations differs from that of kaldi. 

## To Reproduce

using the following or even the defaults parameters:
```
 torchaudio.compliance.kaldi.fbank(waveform, blackman_coeff=0.42, channel=-1, dither=1.0, energy_floor=0.0, frame_length=25.0, frame_shift=10.0, high_freq=0.0, htk_compat=True, low_freq=20.0, min_duration=0.0, num_mel_bins=40, preemphasis_coefficient=0.97, raw_energy=True, remove_dc_offset=True, round_to_power_of_two=True, sample_frequency=16000.0, snip_edges=True, subtract_mean=False, use_energy=False, use_log_fbank=True,use_power=True, vtln_high=-500.0, vtln_low=100.0, vtln_warp=1.0, window_type='hamming')[0]
```

produce this output:
```
tensor([-0.7616, -0.4791,  0.2155,  0.7661,  2.0723,  1.4565,  2.9888,  3.2548,
         1.8460,  3.5807,  3.8290,  4.1785,  4.6776,  4.5801,  5.3610,  4.4910,
         5.1519,  5.3534,  5.2783,  5.6159,  6.0689,  5.5961,  5.8068,  5.0957,
         6.5200,  6.9314,  6.1741,  7.0430,  7.9394,  8.2380,  8.7115,  8.4105,
         8.3154,  8.2186,  7.9444,  8.4468,  8.4293,  8.9476,  9.1008,  9.2495])
```

with compute_fbank_feats of Kaldi
```
tensor([12.9911, 12.9795, 12.9127, 13.6171, 13.7416, 15.1579, 15.1996, 14.9468,
        14.1368, 14.8717, 14.8265, 13.8715, 15.2716, 15.0743, 15.2439, 15.3904,
        13.9460, 13.5932, 14.0038, 14.8721, 13.9944, 15.8337, 14.8682, 13.8247,
        15.0769, 15.1141, 15.1482, 14.7864, 13.6259, 14.4092, 14.1771, 13.6139,
        13.8014, 12.5796,  9.1051,  8.3382,  8.3738,  8.7829,  9.2973,  9.4913])
```

vincentqb:
* Can you provide the kaldi command you used? 
* Can you provide a sample file so we can reproduce?
* Note that you are using `dither=1.0` which adds dither.
* See also #332.

[...]

Issue #252, opened 2019-08-22
Saving and loading the downsampled audio results in a tensor with zeros.


hiromis:
## 🐛 Bug

I was using this particular audio file that I downloaded from https://www.lynxstudio.com/ to experiment with 6 channel audio. When I downsample this from 44.1kHz 8kHz, everything seems fine and I am able to play the audio. However, if I were to save the downsampled file and load it again, the tensor I get back have all zeros in it.  

```python
import torchaudio

waveform, sample_rate = torchaudio.load('ChannelPlacement.wav')

downsample_rate=8000

downsample_resample = torchaudio.transforms.Resample(
    sample_rate, downsample_rate, resampling_method='sinc_interpolation')

down_sampled = downsample_resample(waveform)

print(down_sampled)

torchaudio.save('temp.wav', down_sampled, downsample_rate)

waveform2, sample_rate2 = torchaudio.load('temp.wav')

print(waveform2)
```
```
tensor([[-3.7585e-09, -3.3725e-09,  9.2130e-09,  ..., -4.0691e-08,
          7.2912e-09, -5.7485e-08],
        [-3.2915e-09,  7.5441e-09,  9.4772e-10,  ..., -7.3543e-09,
          3.1981e-08, -2.3025e-08],
        [ 1.5473e-08,  1.8003e-08,  2.5778e-09,  ..., -1.0129e-08,
         -2.0479e-08,  2.6770e-08],
        [-2.1108e-08, -3.9693e-08, -2.2911e-08,  ..., -2.4338e-08,
          3.7029e-08, -3.1360e-09],
        [-2.1277e-08, -1.9114e-09, -4.4245e-09,  ..., -2.3023e-08,
          6.9994e-09, -7.5472e-09],
        [ 5.2013e-09,  2.5186e-08,  2.1362e-08,  ..., -2.6036e-07,
         -1.5355e-07,  3.7919e-08]])
tensor([[0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.],
        [0., 0., 0.,  ..., 0., 0., 0.]])
```

## To Reproduce

1. Load `ChannelPlacement.wav` which is included in the attached zip file
[...]

PR #726, opened 2020-06-17
Add TorchScript-based SoX I/O backend

mthrok:
This PR (and dependent PRs) adds a new backend "sox_io" backend.
 - [x] #718 Add SignalInfo typedef
 - [x] #728 Add "info"
 - [x] #734 Fix SignalInfo
 - [x] #731 Add "load"
 - [x] #732 Add "save"
 - [ ] This PR Add "sox_io" to the list of available backends.
   (So that users can opt-in, but we do not intend to change the default yet.)

The new "sox_io" backend has the following advantages;
 - TorchScript-able 
    The data process pipeline written using the new backend can be dumped and used from C++.
 - Correct
    The original "sox" backend had a number of issues,
     - #618 #236 Info length and rate returns different
     - #430 #252 Save then load would degrade the data
     - https://github.com/pytorch/audio/issues/400#issuecomment-615502956 `load` function cannot handle WAV file correctly.
    I have added bunch of tests to make sure that the new backend does not have the same issue.
    This includes read/write operaions of `wav`, `flac`, `mp3` and `ogg/vorbis` formats. *
    This backend can also read `opus`, ~~though it's not in unit test.~~ #755 
 - Cleaner interface
    - When loading WAV file, correct `dtype` is picked depending on the internal representation of WAV format. This behavior is same as [how SciPy handles WAV file](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html).
        - Load function also provides `normalize` option, which correctly maps integer value range to `[-1.0, 1.0]` with `float32`. 
    - The  existing"sox" backend exposes `sox_signalinfo_t` and `sox_encodinginfo_t` structs directly, but TorchScript does not allow this. Also setting the correct parameters for these structs is not easy. In the new backend, options related to sox-internal are handled in C++, and users only need to provide `compression` option that corresponds to `sox`'s `-C` option.

~~Note The current binary distribution of torchaudio does not contain `ogg/vorbis` codecs. To handle these files, you need to build torchaudio from the source. Refer to [README](https://github.com/pytorch/audio#from-source) for the instruction.~~ #750

### DIFF
--- a/test/test_io.py
+++ b/test/test_io.py
@@ -30,11 +30,15 @@ class Test_LoadSave(unittest.TestCase):
 
     def test_1_save(self):
         for backend in BACKENDS_MP3:
+            if backend == 'sox_io':
+                continue
             with self.subTest():
                 torchaudio.set_audio_backend(backend)
                 self._test_1_save(self.test_filepath, False)
 
         for backend in BACKENDS:
+            if backend == 'sox_io':
+                continue
             with self.subTest():
                 torchaudio.set_audio_backend(backend)
                 self._test_1_save(self.test_filepath_wav, True)
@@ -81,6 +85,8 @@ class Test_LoadSave(unittest.TestCase):
 
     def test_1_save_sine(self):
         for backend in BACKENDS:
+            if backend == 'sox_io':
+                continue
             with self.subTest():
                 torchaudio.set_audio_backend(backend)
                 self._test_1_save_sine()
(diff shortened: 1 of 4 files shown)


test/test_io.py
```diff
@@ -30,11 +30,15 @@ class Test_LoadSave(unittest.TestCase):
 
     def test_1_save(self):
         for backend in BACKENDS_MP3:
+            if backend == 'sox_io':
+                continue
```
vincentqb, line 34: Why is `sox_io` made a special case here?

mthrok: The test cases in this class depends on the global state of the previous test ran, which brakes the principle of unit test and having `sox_io` breaks it.

mthrok:
thanks!

PR #509, opened 2020-04-03
run tests again

vincentqb:
As noted [here](https://github.com/pytorch/audio/pull/508#pullrequestreview-387586964), tests were not running.

cc @mthrok

PR #727, opened 2020-06-18
Add subclass in model test classes

jimchen90:
Add subclass `common_utils.TorchaudioTestCase` in model test classes.
Related to #705

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -3,8 +3,10 @@ import unittest
 import torch
 from torchaudio.models import Wav2Letter, _MelResNet
 
+from . import common_utils
 
-class TestWav2Letter(unittest.TestCase):
+
+class TestWav2Letter(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
         batch_size = 2
@@ -33,7 +35,7 @@ class TestWav2Letter(unittest.TestCase):
         assert out.size() == (batch_size, num_classes, 2)
 
 
-class TestMelResNet(unittest.TestCase):
+class TestMelResNet(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
 
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,8 +1,10 @@
+import unittest
+
 import torch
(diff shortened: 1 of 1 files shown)


vincentqb:
oops, should have waited "ready for review" :)

test/test_models.py
```diff
@@ -1,8 +1,10 @@
+import unittest
+
 import torch
 from torchaudio.models import Wav2Letter, _MelResNet
 
 
-class TestWav2Letter:
+class TestWav2Letter(unittest.TestCase):
```
mthrok, line 7: Instead of `unittest.TestCase`, can you use `TorchaudioTestCase` from `common_utils`?
It's available in the latest master.

test/test_models.py
```diff
@@ -49,3 +51,6 @@ def test_waveform(self):
         out = model(x)
 
         assert out.size() == (batch_size, output_dims, num_features - pad * 2)
+
+if __name__ == '__main__':
+    unittest.main()
```
mthrok, line 56: nit/very subjective: although it does not hurt but I personally find adding `unittest.main` redundant and unnecessary.

vincentqb: :) since the other tests have it, let's keep that line in for consistency with the other files and other domains (e.g. [here](https://github.com/pytorch/vision/blob/master/test/test_datasets.py)), but feel free to open a pull request to discuss removing them all

mthrok: FYI: It was added to the other files without review or approval in https://github.com/pytorch/audio/pull/509, and it is not used by any official automated jobs. So it is not very convincing to keep this for the sake of consistency.

vincentqb: since we don't want to block this PR on this, feel free to open a separate PR to change this :)

vincentqb: @jimchen90 -- can you move this change to a separate pull request? thanks!

jimchen90: Thanks, do you mean that I open a separate pull request without these two lines after this file is merged?

jimchen90: Yes. Thanks

PR #731, opened 2020-06-18
Add TorchScript-able "load" func to sox_io backend

mthrok:
This is a part of PRs to add new "sox_io" backend. #726 and depends on #718 and #728 .

This PR adds `load` function to "sox_io" backend, which is  tested on the following audio formats;
 - `wav`
 - `mp3`
 - `flac`
 - `ogg/vorbis` *

By default, "sox_io" backend returns Tensor with `float32` dtype and the shape of `[channel, time]`. The samples are normalized to fit in the range of `[-1.0, 1.0]`.

Unlike existing "sox" backend, the new `load` function can handle WAV file natively, when the input format is WAV with integer type, (such as 32-bit signed integer, 16-bit signed integer and 8-bit unsigned integer) by providing `normalize=False`, this function can return integer Tensor, where the samples are expressed within the whole range of the corresponding dtype, that is, `int32` tensor for `32-bit PCM`, `int16` for `16-bit PCM` and `uint8` for `8-bit PCM`. This behavior follows [scipy.io.wavfile.read](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html). `normalize` parameter has no effect for other formats and the load function always return normalized value with `float32` Tensor.

__* Note__ The current binary distribution of torchaudio does not contain `ogg/vorbis` and `opus` codecs. To handle these files, one needs to build torchaudio from the source with proper codecs in the system.

__Note 2__ Since this PR, `scipy` becomes required module for running test.

### DIFF
--- a/test/sox_io_backend/common.py
+++ b/test/sox_io_backend/common.py
@@ -1,2 +1,90 @@
+from typing import Optional
+
+import torch
+import scipy.io.wavfile
+
+
 def get_test_name(func, _, params):
     return f'{func.__name__}_{"_".join(str(p) for p in params.args)}'
+
+
+def normalize_wav(tensor: torch.Tensor) -> torch.Tensor:
+    if tensor.dtype == torch.float32:
+        pass
+    elif tensor.dtype == torch.int32:
+        tensor = tensor.to(torch.float32)
+        tensor[tensor > 0] /= 2147483647.
+        tensor[tensor < 0] /= 2147483648.
+    elif tensor.dtype == torch.int16:
+        tensor = tensor.to(torch.float32)
+        tensor[tensor > 0] /= 32767.
+        tensor[tensor < 0] /= 32768.
+    elif tensor.dtype == torch.uint8:
+        tensor = tensor.to(torch.float32) - 128
+        tensor[tensor > 0] /= 127.
+        tensor[tensor < 0] /= 128.
+    return tensor
+
+
+def get_wav_data(
+        dtype: str,
+        num_channels: int,
+        *,
+        num_frames: Optional[int] = None,
+        normalize: bool = True,
+        channels_first: bool = True,
+):
+    """Generate linear signal of the given dtype and num_channels
+
+    Data range is
+        [-1.0, 1.0] for float32,
+        [-2147483648, 2147483647] for int32
+        [-32768, 32767] for int16
+        [0, 255] for uint8
+
+    num_frames allow to change the linear interpolation parameter.
+    Default values are 256 for uint8, else 1 << 16.
+    1 << 16 as default is so that int16 value range is completely covered.
+    """
+    dtype_ = getattr(torch, dtype)
+
+    if num_frames is None:
+        if dtype == 'uint8':
+            num_frames = 256
+        else:
+            num_frames = 1 << 16
+
+    if dtype == 'uint8':
+        base = torch.linspace(0, 255, num_frames, dtype=dtype_)
+    if dtype == 'float32':
+        base = torch.linspace(-1., 1., num_frames, dtype=dtype_)
+    if dtype == 'int32':
+        base = torch.linspace(-2147483648, 2147483647, num_frames, dtype=dtype_)
+    if dtype == 'int16':
+        base = torch.linspace(-32768, 32767, num_frames, dtype=dtype_)
+    data = base.repeat([num_channels, 1])
+    if not channels_first:
+        data = data.transpose(1, 0)
+    if normalize:
+        data = normalize_wav(data)
+    return data
+
+
+def load_wav(path: str, normalize=True, channels_first=True) -> torch.Tensor:
+    """Load wav file without torchaudio"""
+    sample_rate, data = scipy.io.wavfile.read(path)
+    data = torch.from_numpy(data.copy())
+    if data.ndim == 1:
+        data = data.unsqueeze(1)
+    if normalize:
+        data = normalize_wav(data)
+    if channels_first:
+        data = data.transpose(1, 0)
+    return data, sample_rate
+
+
+def save_wav(path, data, sample_rate, channels_first=True):
+    """Save wav file without torchaudio"""
+    if channels_first:
+        data = data.transpose(1, 0)
+    scipy.io.wavfile.write(path, sample_rate, data.numpy())
--- a/test/sox_io_backend/sox_utils.py
+++ b/test/sox_io_backend/sox_utils.py
@@ -26,6 +26,9 @@ def gen_audio_file(
         *, encoding=None, bit_depth=None, compression=None, attenuation=None, duration=1,
 ):
     """Generate synthetic audio file with `sox` command."""
+    if path.endswith('.wav'):
+        raise RuntimeError(
+            'Use get_wav_data and save_wav to generate wav file for accurate result.')
     command = [
         'sox',
         '-V',  # verbose
@@ -51,4 +54,17 @@ def gen_audio_file(
         command += ['vol', f'-{attenuation}dB']
     print(' '.join(command))
     subprocess.run(command, check=True)
-    subprocess.run(['soxi', path], check=True)
+
+
+def convert_audio_file(
+        src_path, dst_path,
+        *, bit_depth=None, compression=None):
+    """Convert audio file with `sox` command."""
+    command = ['sox', '-V', str(src_path)]
+    if bit_depth is not None:
+        command += ['--bits', str(bit_depth)]
+    if compression is not None:
+        command += ['--compression', str(compression)]
+    command += [dst_path]
+    print(' '.join(command))
+    subprocess.run(command, check=True)
--- a/test/sox_io_backend/test_load.py
+++ b/test/sox_io_backend/test_load.py
@@ -1,0 +1,245 @@
+import itertools
+
+from torchaudio.backend import sox_io_backend
+from parameterized import parameterized
+
+from ..common_utils import (
+    TempDirMixin,
+    PytorchTestCase,
+    skipIfNoExec,
+    skipIfNoExtension,
+)
+from .common import (
+    get_test_name,
+    get_wav_data,
+    load_wav,
+    save_wav,
+)
+from . import sox_utils
+
+
+class LoadTestBase(TempDirMixin, PytorchTestCase):
+    def assert_wav(self, dtype, sample_rate, num_channels, normalize, duration):
+        """`sox_io_backend.load` can load wav format correctly.
+
+        Wav data loaded with sox_io backend should match those with scipy
+        """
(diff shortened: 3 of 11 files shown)


test/sox_io_backend/test_load.py
```diff
@@ -0,0 +1,245 @@
...
+        we create reference wav file from mp3
+
+         x
+         |
+         | 1. Generate mp3 with Sox
+         |
+         v    2. Convert to wav with Sox
+        mp3 ------------------------------> wav
+         |                                   |
+         | 3. Load with torchaduio           | 4. Load with scipy
```
vincentqb, line 48: nit: torchaudio

test/sox_io_backend/test_torchscript.py
```diff
@@ -46,3 +51,28 @@ def test_info_wav(self, dtype, sample_rate, num_channels):
...
+    )), name_func=get_test_name)
+    def test_load_wav(self, dtype, sample_rate, num_channels, normalize, channels_first):
+        """`sox_io_backend.load` is torchscript-able and returns the same result"""
+        audio_path = self.get_temp_path(f'test_load_{dtype}_{sample_rate}_{num_channels}_{normalize}.wav')
+        data = get_wav_data(dtype, num_channels, normalize=False, num_frames=1 * sample_rate)
+        save_wav(audio_path, data, sample_rate)
+
+        script_path = self.get_temp_path('load_func')
+        torch.jit.script(py_load_func).save(script_path)
+        ts_load_func = torch.jit.load(script_path)
```
vincentqb, line 70: woohoo!!

torchaudio/backend/sox_io_backend.py
```diff
@@ -8,3 +10,66 @@
...
+            integer type, else ``float32`` type. If ``channels_first=True``, it has
+            ``[channel, time]`` else ``[time, channel]``.
+    """
+    signal_info = torch.classes.torchaudio.SignalInfo(0, 0, 0)
+    tensor = torch.ops.torchaudio.sox_io_load_audio_file(
+        filepath, signal_info, frame_offset, num_frames, normalize, channels_first)
+    return tensor, signal_info.get_sample_rate()
+
+
+load_wav = load
```
vincentqb, line 75: woohoo!! :D

torchaudio/csrc/sox_utils.cpp
```diff
@@ -0,0 +1,71 @@
...
+            return torch::kInt32;
+          default:
+            throw std::runtime_error(
+                "Only 16 and 32 bits are supported for signed PCM.");
+        }
+      case SOX_ENCODING_FLOAT: // 32-bit floating-point WAVE
+      case SOX_ENCODING_MP3:
+      case SOX_ENCODING_FLAC:
+      case SOX_ENCODING_VORBIS:
+      case SOX_ENCODING_OPUS:
```
vincentqb, line 29: This limits the codec supported, right? Are we reducing the support from what was mentioned in the readme? should we make an update to the [readme](https://github.com/pytorch/audio/blob/master/README.md) too?

vincentqb: This would be BC-breaking then, right?

vincentqb: This is not yet BC-breaking since this PR doesn't replace the current backend, but creates a new one instead.

mthrok: I gave some thoughts on this one, and I am okay with relaxing this, given that we clearly state somewhere (probably README AND document) that `wav`, `mp3`, `flac` and `vorbis` are the only formats rigorously tested and that the other formats might work but the correctness is not tested.

The original reason I wanted to limit the formats is that I wanted to make it clear that only some of the formats are verified its correctness, so that users are aware of it.

I think just passing through the other formats and the behavior should be same as the existing "sox" backend.

test/sox_io_backend/common.py
```diff
@@ -1,2 +1,95 @@
...
 def get_test_name(func, _, params):
     return f'{func.__name__}_{"_".join(str(p) for p in params.args)}'
+
+
+def normalize_wav(tensor: torch.Tensor) -> torch.Tensor:
+    if tensor.dtype == torch.float32:
+        pass
+    elif tensor.dtype == torch.int32:
+        tensor = tensor.to(torch.float32)
+        tensor[tensor > 0] /= 2147483647.
```
vincentqb, line 16: nit: is there a way of fetching these numbers instead of hard coding them?

torchaudio/csrc/sox_utils.cpp
```diff
@@ -0,0 +1,70 @@
+#include <c10/core/ScalarType.h>
+#include <sox.h>
+#include <torchaudio/csrc/sox_utils.h>
+
+namespace torchaudio {
+namespace sox_utils {
+
+caffe2::TypeMeta get_dtype(
```
vincentqb, line 8: nit: I was under the impression that we should avoid caffe2. Can you explain why it is necessary here?

torchaudio/backend/sox_io_backend.py
```diff
@@ -8,3 +10,72 @@
...
+         - OGG/VORBIS
+
+    By default, this function returns Tensor with ``float32`` dtype and the shape of ``[channel, time]``.
+    The samples are normalized to fit in the range of ``[-1.0, 1.0]``.
+
+    When the input format is WAV with integer type, such as 32-bit signed integer, 16-bit
+    signed integer and 8-bit unsigned integer (24-bit signed integer is not supported),
+    by providing ``normalize=False``, this function can return integer Tensor, where the samples
+    are expressed within the whole range of the corresponding dtype, that is, ``int32`` tensor
+    for 32-bit signed PCM, ``int16`` for 16-bit signed PCM and ``uint8`` for 8-bit unsigned PCM.
```
vincentqb, line 51: I prefer a function that would always return a float between -1 and 1 when normalize=True, and a float unnormalized when False. Someone who wants the integer version would call a different function. Thoughts?

torchaudio/csrc/register.cpp
```diff
@@ -7,19 +7,33 @@
...
+////////////////////////////////////////////////////////////////////////////////
+// sox_io.h
+////////////////////////////////////////////////////////////////////////////////
 static auto registerGetInfo = torch::RegisterOperators().op(
     torch::RegisterOperators::options()
         .schema(
             "torchaudio::sox_io_get_info(str path) -> __torch__.torch.classes.torchaudio.SignalInfo info")
         .catchAllKernel<decltype(sox_io::get_info), &sox_io::get_info>());
 
+static auto registerLoadAudioFile = torch::RegisterOperators().op(
```
vincentqb, line 29: nit: should there be a comment before, as introduced above?

torchaudio/csrc/sox_utils.cpp
```diff
@@ -0,0 +1,70 @@
...
+    const caffe2::TypeMeta dtype,
+    const bool normalize,
+    const bool channels_first) {
+  auto t = torch::from_blob(
+      buffer, {num_samples / num_channels, num_channels}, torch::kInt32);
+  // Note: Tensor created from_blob does not own data but borrwos
+  // So make sure to create a new copy after processing samples.
+  if (normalize || dtype == torch::kFloat32) {
+    t = t.to(torch::kFloat32);
+    t *= (t > 0) / 2147483647. + (t < 0) / 2147483648.;
```
vincentqb, line 50: this one is not in the tests though :)

torchaudio/csrc/sox_utils.cpp
mthrok: Simply that's what's returned when you call `dtype` on `torch::Tensor` class.
https://pytorch.org/cppdocs/api/classat_1_1_tensor.html#_CPPv4NK2at6Tensor5dtypeEv

torchaudio/csrc/register.cpp
mthrok: such as?

torchaudio/backend/sox_io_backend.py
mthrok: That way, if user wants to get integer value, they have to go through float, which might cause numerical issue around the edge of value range without care. I do not think that's a good UX.

I/O is fundamental to scientific computing so loading function should provide the functionality to load data as is, without modifying it.
See also https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html

torchaudio/csrc/sox_utils.cpp
mthrok: What do you mean? The combination of `float32` and `normalize=true|false` are tested everywhere.

torchaudio/backend/sox_io_backend.py
vincentqb: Yes, it should definitely be possible to load the data as is, and without going through floats first. That being said, we could delegate that to a different function, and keep the main load function as returning floats. However, the default with `normalize=True` does just that, so I'm happy enough as it is.

torchaudio/csrc/sox_utils.cpp
vincentqb: I meant the `2147483647` :)

torchaudio/csrc/register.cpp
vincentqb:
```
////////////////////////////////////////////////////////////////////////////////
// sox_io.h
////////////////////////////////////////////////////////////////////////////////
```
but then that's redundant :) Alright, let's keep things as is.

test/sox_io_backend/sox_utils.py
```diff
@@ -51,4 +54,17 @@ def gen_audio_file(
         command += ['vol', f'-{attenuation}dB']
     print(' '.join(command))
     subprocess.run(command, check=True)
-    subprocess.run(['soxi', path], check=True)
```
vincentqb, line 54: nit: why is this removed at this time?

mthrok: The command above it, `sox -V` prints the same information, so it turned out redundant.

torchaudio/csrc/sox_utils.cpp
mthrok: It is tested.

mthrok:
> nit: there's a C++ lint error :)

Yeah I was waiting for the approval first as I have six branches depend on this and had to rebase all of them.

mthrok:
thanks!

torchaudio/csrc/sox_utils.cpp
vincentqb: I meant something different: I have seen this hardcoded values appear in a few places in this PR. Most of them are in tests, so I don't worry much about them. This hardcoded value is not in a test file though, so I would have preferred if it were not hardcoded.

PR #732, opened 2020-06-18
Add TorchScript-able "save" func to sox_io backend

mthrok:
This is a part of PRs to add new "sox_io" backend. #726 and depends on #718, #728 and #731.

This PR adds `save` function to "sox_io" backend, which can save Tensor to a file with the following audio formats;
 - `wav`
 - `mp3`
 - `flac`
 - `ogg/vorbis` *

~~__* Note__ The current binary distribution of torchaudio does not contain `ogg/vorbis` codecs. To handle these files, one needs to build torchaudio from the source.~~

### DIFF
--- a/test/sox_io_backend/test_save.py
+++ b/test/sox_io_backend/test_save.py
@@ -1,0 +1,304 @@
+import itertools
+
+from torchaudio.backend import sox_io_backend
+from parameterized import parameterized
+
+from ..common_utils import (
+    TempDirMixin,
+    PytorchTestCase,
+    skipIfNoExec,
+    skipIfNoExtension,
+)
+from .common import (
+    get_test_name,
+    get_wav_data,
+    load_wav,
+    save_wav,
+)
+from . import sox_utils
+
+
+class SaveTestBase(TempDirMixin, PytorchTestCase):
+    def assert_wav(self, dtype, sample_rate, num_channels, num_frames):
+        """`sox_io_backend.save` can save wav format."""
+        path = self.get_temp_path('data.wav')
+        expected = get_wav_data(dtype, num_channels, num_frames=num_frames)
+        sox_io_backend.save(path, expected, sample_rate)
+        found, sr = load_wav(path)
+        assert sample_rate == sr
+        self.assertEqual(found, expected)
+
+    def assert_mp3(self, sample_rate, num_channels, bit_rate, duration):
+        """`sox_io_backend.save` can save mp3 format.
+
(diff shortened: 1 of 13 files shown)


torchaudio/backend/sox_io_backend.py
```diff
@@ -77,4 +72,58 @@ def load(
...
+        if ext == 'wav':
+            compression = 0.
+        elif ext == 'mp3':
+            compression = -4.5
+        elif ext == 'flac':
+            compression = 8.
+        elif ext in ['ogg', 'vorbis']:
+            compression = 3.
+        else:
+            raise RuntimeError(f'Unsupported file type: "{ext}"')
```
mthrok, line 124: @vincentqb I thought of calling for opening an issue for adding support for other audio formats, but differentiating the valid extensions that are supported by libsox from invalid ones is not trivial and there are many formats that require additional libraries to be installed. Meaning that adding support on our `get_signalinfo` function may not be enough for users to actually use a new format, depending on the format. So I think we want to be more cautious and not giving high expectations to users casually. Therefore I decided not to include call for opening a PR in the error message.

vincentqb: Can you remind me how the current backend exposes more backend? From our discussion, the user had to provide extra internal information, right?

test/sox_io_backend/test_save.py
```diff
@@ -0,0 +1,304 @@
...
+        [8000, 16000],
+        [1, 2],
+        [-1, 0, 1, 2, 3, 3.6, 5, 10],
+    )), name_func=get_test_name)
+    def test_vorbis(self, sample_rate, num_channels, quality_level):
+        """`sox_io_backend.save` can save vorbis format."""
+        self.assert_vorbis(sample_rate, num_channels, quality_level, duration=20)
+
+    # note: torchaudio can load large vorbis file, but cannot save large volbis file
+    # the following test causes Segmentation fault
```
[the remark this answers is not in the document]
mthrok: Nope I do not think this can be fixed. However, I wanted to leave a record of know limitation.

torchaudio/backend/sox_io_backend.py
mthrok: The existing implementation asks users to construct the following data structure, which requires users to know the internal of libsox.

https://fossies.org/dox/sox-14.4.2/structsox__signalinfo__t.html
https://fossies.org/dox/sox-14.4.2/structsox__encodinginfo__t.html

mthrok:
thanks

PR #734, opened 2020-06-19
Fix SignalInfo member name to frame

mthrok:
This PR fixes the wrong member name of SignalInfo introduced in #718. 

 - `num_samples` == `num_frames` * `num_channels`.


mthrok:
thanks

PR #724, opened 2020-06-17
UpsampleNetwork

jimchen90:
This Upsampling block is part of WaveRNN model. Now the test is to validate the output dimensions of this block. Other tests will be added after other blocks are combined.
Related to #446

Stack:

Add MelResNet Block #705  #751 
**Add Upsampling Block #724**
Add WaveRNN Model #735 
Add example pipeline with WaveRNN #749

PR #735, opened 2020-06-22
Add WaveRNN Model

jimchen90:
This is the WaveRNN model. 
Related to #446

Stack:

~~Add MelResNet Block #705, #751~~
~~Add Upsampling Block #724~~
**Add WaveRNN Model #735**
Add example pipeline with WaveRNN #749

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -86,7 +86,7 @@ class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
 class TestWaveRNN(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
-        """Validate the output dimensions of a _WaveRNN model with a waveform input.
+        """Validate the output dimensions of a _WaveRNN model in waveform mode.
         """
 
         upsample_scales = [5, 5, 8]
@@ -114,7 +114,7 @@ class TestWaveRNN(common_utils.TorchaudioTestCase):
         assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
 
     def test_mol(self):
-        """Validate the output dimensions of a _WaveRNN model with a mol input.
+        """Validate the output dimensions of a _WaveRNN model in mol mode.
         """
 
         upsample_scales = [5, 5, 8]
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -81,3 +81,36 @@ class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
 
         assert out1.size() == (n_batch, n_freq, total_scale * (n_time - kernel_size + 1))
         assert out2.size() == (n_batch, n_output, total_scale * (n_time - kernel_size + 1))
+
+
+class TestWaveRNN(common_utils.TorchaudioTestCase):
+
+    def test_waveform(self):
+        """
+        Create a tensor as the input of _WaveRNN model
+        and test if the output dimensions are correct.
+        """
+
+        upsample_scales = [5, 5, 8]
+        n_rnn = 512
+        n_fc = 512
+        n_bits = 9
+        sample_rate = 24000
+        hop_length = 200
+        batch_size = 2
+        n_time = 200
+        n_freq = 100
+        n_output = 256
+        n_res_block = 10
+        n_hidden = 128
+        kernel_size = 5
+        mode = 'RAW'
+
+        model = _WaveRNN(upsample_scales, n_bits, sample_rate, hop_length, n_res_block,
+                         n_rnn, n_fc, kernel_size, n_freq, n_hidden, n_output, mode)
+
+        x = torch.rand(batch_size, hop_length * (n_time - kernel_size + 1))
+        mels = torch.rand(batch_size, n_freq, n_time)
+        out = model(x, mels)
+
+        assert out.size() == (batch_size, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -107,11 +107,11 @@ class TestWaveRNN(common_utils.TorchaudioTestCase):
         model = _WaveRNN(upsample_scales, n_bits, sample_rate, hop_length, n_res_block,
                          n_rnn, n_fc, kernel_size, n_freq, n_hidden, n_output, mode)
 
-        x = torch.rand(n_batch, hop_length * (n_time - kernel_size + 1))
-        mels = torch.rand(n_batch, n_freq, n_time)
+        x = torch.rand(n_batch, 1, hop_length * (n_time - kernel_size + 1))
+        mels = torch.rand(n_batch, 1, n_freq, n_time)
         out = model(x, mels)
 
-        assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
+        assert out.size() == (n_batch, 1, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
 
     def test_mol(self):
         """Validate the output dimensions of a _WaveRNN model in mol mode.
@@ -135,8 +135,8 @@ class TestWaveRNN(common_utils.TorchaudioTestCase):
         model = _WaveRNN(upsample_scales, n_bits, sample_rate, hop_length, n_res_block,
                          n_rnn, n_fc, kernel_size, n_freq, n_hidden, n_output, mode)
 
-        x = torch.rand(n_batch, hop_length * (n_time - kernel_size + 1))
-        mels = torch.rand(n_batch, n_freq, n_time)
+        x = torch.rand(n_batch, 1, hop_length * (n_time - kernel_size + 1))
+        mels = torch.rand(n_batch, 1, n_freq, n_time)
         out = model(x, mels)
 
-        assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 30)
+        assert out.size() == (n_batch, 1, hop_length * (n_time - kernel_size + 1), 30)
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -86,9 +86,7 @@ class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
 class TestWaveRNN(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
-        """
-        Create a tensor as the input of _WaveRNN model
-        and test if the output dimensions are correct.
+        """test the output dimensions of waveform input after _WaveRNN model.
         """
 
         upsample_scales = [5, 5, 8]
@@ -97,20 +95,48 @@ class TestWaveRNN(common_utils.TorchaudioTestCase):
         n_bits = 9
         sample_rate = 24000
         hop_length = 200
-        batch_size = 2
+        n_batch = 2
         n_time = 200
         n_freq = 100
         n_output = 256
         n_res_block = 10
         n_hidden = 128
         kernel_size = 5
-        mode = 'RAW'
+        mode = 'waveform'
 
         model = _WaveRNN(upsample_scales, n_bits, sample_rate, hop_length, n_res_block,
                          n_rnn, n_fc, kernel_size, n_freq, n_hidden, n_output, mode)
 
-        x = torch.rand(batch_size, hop_length * (n_time - kernel_size + 1))
-        mels = torch.rand(batch_size, n_freq, n_time)
+        x = torch.rand(n_batch, hop_length * (n_time - kernel_size + 1))
+        mels = torch.rand(n_batch, n_freq, n_time)
         out = model(x, mels)
 
-        assert out.size() == (batch_size, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
+        assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
+
+    def test_mol(self):
+        """test the output dimensions of mol input after _WaveRNN model.
+        """
+
+        upsample_scales = [5, 5, 8]
+        n_rnn = 512
+        n_fc = 512
+        n_bits = 9
+        sample_rate = 24000
+        hop_length = 200
+        n_batch = 2
+        n_time = 200
+        n_freq = 100
+        n_output = 256
+        n_res_block = 10
+        n_hidden = 128
+        kernel_size = 5
+        mode = 'mol'
+
+        model = _WaveRNN(upsample_scales, n_bits, sample_rate, hop_length, n_res_block,
+                         n_rnn, n_fc, kernel_size, n_freq, n_hidden, n_output, mode)
+
+        x = torch.rand(n_batch, hop_length * (n_time - kernel_size + 1))
+        mels = torch.rand(n_batch, n_freq, n_time)
+        out = model(x, mels)
+
+        assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 30)
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -86,7 +86,7 @@ class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
 class TestWaveRNN(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
-        """test the output dimensions of waveform input after _WaveRNN model.
+        """Validate the output dimensions of a _WaveRNN model with a waveform input.
         """
 
         upsample_scales = [5, 5, 8]
@@ -114,7 +114,7 @@ class TestWaveRNN(common_utils.TorchaudioTestCase):
         assert out.size() == (n_batch, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
 
     def test_mol(self):
-        """test the output dimensions of mol input after _WaveRNN model.
+        """Validate the output dimensions of a _WaveRNN model with a mol input.
         """
 
         upsample_scales = [5, 5, 8]
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,5 +1,3 @@
-import unittest
-
 import torch
 from torchaudio.models import Wav2Letter, _MelResNet, _UpsampleNetwork, _WaveRNN
 
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,5 +1,7 @@
+import unittest
+
 import torch
-from torchaudio.models import Wav2Letter, _MelResNet, _UpsampleNetwork
+from torchaudio.models import Wav2Letter, _MelResNet, _UpsampleNetwork, _WaveRNN
 
 from . import common_utils
 
--- a/torchaudio/models/_wavernn.py
+++ b/torchaudio/models/_wavernn.py
@@ -1,9 +1,10 @@
 from typing import List
 
+import torch
 from torch import Tensor
 from torch import nn
 
-__all__ = ["_ResBlock", "_MelResNet", "_Stretch2d", "_UpsampleNetwork"]
+__all__ = ["_ResBlock", "_MelResNet", "_Stretch2d", "_UpsampleNetwork", "_WaveRNN"]
 
 
 class _ResBlock(nn.Module):
@@ -90,6 +91,7 @@ class _MelResNet(nn.Module):
 
 
 class _Stretch2d(nn.Module):
+<<<<<<< HEAD
     r"""Upscale the frequency and time dimensions of a spectrogram.
 
     Args:
@@ -144,13 +146,16 @@ class _UpsampleNetwork(nn.Module):
     "Efficient Neural Audio Synthesis". Nal Kalchbrenner, Erich Elsen, Karen Simonyan, Seb Noury,
     Norman Casagrande, Edward Lockhart, Florian Stimberg, Aaron van den Oord, Sander Dieleman,
     Koray Kavukcuoglu. arXiv:1802.08435, 2018.
+=======
+    r"""This is a two-dimensional stretch layer. It is a block used in WaveRNN.
+>>>>>>> add wavernn model
 
     Args:
-        x_scale: the scale factor in x axis (required)
-        y_scale: the scale factor in y axis (required)
+        x_scale: the scale factor in x axis
+        y_scale: the scale factor in y axis
 
     Examples::
-        >>> stretch2d = _Stretch2d(x_scale=1, y_scale=1)
+        >>> stretch2d = _Stretch2d(x_scale=10, y_scale=10)
 
         >>> input = torch.rand(10, 1, 100, 512)
         >>> output = stretch2d(input)
@@ -165,46 +170,50 @@ class _UpsampleNetwork(nn.Module):
         self.y_scale = y_scale
 
     def forward(self, x: Tensor) -> Tensor:
-        r"""Pass the input through the _Stretch2d layer.
-
+        r"""
         Args:
-            x: the input sequence to the _Stretch2d layer (required)
+            x: the input sequence to the _Stretch2d layer
 
         Shape:
-            - x: :math:`(batch_size, channel, freq, time)`
-            - output: :math:`(batch_size, channel, freq * y_scale, time * x_scale)`
+            - x: :math:`(..., freq, time)`
+            - output: :math:`(..., freq * y_scale, time * x_scale)`
         """
 
-        batch_size, channel, freq, time = x.size()
-        x = x.unsqueeze(-1).unsqueeze(3)
-        x = x.repeat(1, 1, 1, self.y_scale, 1, self.x_scale)
-        return x.view(batch_size, channel, freq * self.y_scale, time * self.x_scale)
+        return x.repeat_interleave(self.y_scale, 2).repeat_interleave(self.x_scale, 3)
 
 
 class _UpsampleNetwork(nn.Module):
     r"""This is an upsample block based on a stack of Conv2d and Strech2d layers.
-    It is a block used in WaveRNN. WaveRNN is based on the paper "Efficient Neural Audio Synthesis".
-    Nal Kalchbrenner, Erich Elsen, Karen Simonyan, Seb Noury, Norman Casagrande, Edward Lockhart,
-    Florian Stimberg, Aaron van den Oord, Sander Dieleman, Koray Kavukcuoglu. arXiv:1802.08435, 2018.
+    It is a block used in WaveRNN.
 
     Args:
-        upsample_scales: the list of upsample scales (required)
-        res_blocks: the number of ResBlock in stack (default=10)
-        input_dims: the number of input sequence (default=100)
-        hidden_dims: the number of compute dimensions (default=128)
-        output_dims: the number of output sequence (default=128)
-        pad: the kernel size (kernel_size = pad * 2 + 1) in the first Conv1d layer (default=2)
+        upsample_scales: the list of upsample scales
+        n_res_block: the number of ResBlock in stack (default=10)
+        n_freq: the number of bins in a spectrogram (default=128)
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
 
     Examples::
         >>> upsamplenetwork = _UpsampleNetwork(upsample_scales=[4, 4, 16],
-                                               res_blocks=10,
-                                               input_dims=100,
-                                               hidden_dims=128,
-                                               output_dims=128,
-                                               pad=2)
-        >>> input = torch.rand(10, 100, 512)
+                                               n_res_block=10,
+                                               n_freq=128,
+                                               n_hidden=128,
+                                               n_output=128,
+                                               kernel_size=5)
+        >>> input = torch.rand(10, 128, 512)
         >>> output = upsamplenetwork(input)
 
+<<<<<<< HEAD
+=======
+    def __init__(self,
+                 upsample_scales: List[int],
+                 n_res_block: int = 10,
+                 n_freq: int = 128,
+                 n_hidden: int = 128,
+                 n_output: int = 128,
+                 kernel_size: int = 5) -> None:
+>>>>>>> add wavernn model
         super().__init__()
 
         total_scale = 1
@@ -213,7 +222,10 @@ class _UpsampleNetwork(nn.Module):
 
         self.indent = (kernel_size - 1) // 2 * total_scale
         self.resnet = _MelResNet(n_res_block, n_freq, n_hidden, n_output, kernel_size)
+<<<<<<< HEAD
 
+=======
+>>>>>>> add wavernn model
         self.resnet_stretch = _Stretch2d(total_scale, 1)
 
         up_layers = []
@@ -253,15 +265,14 @@ class _UpsampleNetwork(nn.Module):
         return upsampling_output, resnet_output
 
     def forward(self, x: Tensor) -> Tensor:
-        r"""Pass the input through the _UpsampleNetwork layer.
-
+        r"""
         Args:
-            x: the input sequence to the _UpsampleNetwork layer (required)
+            x: the input sequence to the _UpsampleNetwork layer
 
         Shape:
-            - x: :math:`(batch_size, freq, time)`
-            - output: :math:`(batch_size, (time - 2 * pad) * total_scale, freq)`,
-                            `(batch_size, (time - 2 * pad) * total_scale, output_dims)`
+            - x: :math:`(batch, freq, time)`.
+            - output: :math:`(batch, (time - kernel_size + 1) * total_scale, freq)`,
+                            `(batch, (time - kernel_size + 1) * total_scale, n_output)`
         where total_scale is the product of all elements in upsample_scales.
         """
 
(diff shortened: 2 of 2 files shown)


jimchen90:
Three failed tests look irrelevant to this PR.

torchaudio/models/_wavernn.py
```diff
@@ -1,47 +1,45 @@
...
+__all__ = ["_ResBlock", "_MelResNet", "_Stretch2d", "_UpsampleNetwork", "_WaveRNN"]
 
 
 class _ResBlock(nn.Module):
     r"""This is a ResNet block layer. This layer is based on the paper "Deep Residual Learning
     for Image Recognition". Kaiming He,  Xiangyu Zhang, Shaoqing Ren, Jian Sun. CVPR, 2016.
-    It is a block used in WaveRNN. WaveRNN is based on the paper "Efficient Neural Audio Synthesis".
-    Nal Kalchbrenner, Erich Elsen, Karen Simonyan, Seb Noury, Norman Casagrande, Edward Lockhart,
-    Florian Stimberg, Aaron van den Oord, Sander Dieleman, Koray Kavukcuoglu. arXiv:1802.08435, 2018.
+    It is a block used in WaveRNN.
```
mthrok, line 13: In Python, the first line has to be complete and give the summary of what it is.
See https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

Starting with "This is ..." is lengthy but not adding a value, instead, try to come up with some useful information for first time reader. In this case something like `ResNet block based on "Deep Residual Learning
    for Image Recognition"` would do.

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+                 kernel_size: int = 5) -> None:
         super().__init__()
 
-        kernel_size = pad * 2 + 1
         ResBlocks = []
 
-        for i in range(res_blocks):
-            ResBlocks.append(_ResBlock(hidden_dims))
+        for i in range(n_res_block):
+            ResBlocks.append(_ResBlock(n_hidden))
```
mthrok, line 77: You can use one-liner here for better readability.

```
ResBlocks = [_ResBlock(n_hidden) for _ in range(n_res_block)]
```

Use underscore as variable name when you do not use the variable.

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+            total_scale *= upsample_scale
+
+        self.indent = (kernel_size - 1) // 2 * total_scale
+        self.resnet = _MelResNet(n_res_block, n_freq, n_hidden, n_output, kernel_size)
+        self.resnet_stretch = _Stretch2d(total_scale, 1)
+
+        up_layers = []
+        for scale in upsample_scales:
+            k_size = (1, scale * 2 + 1)
+            padding = (0, scale)
```
mthrok, line 178: defining variables like this looks uncommon yet it does not improve readability.

just put variables in keyword argument and it is readable enough.

```
conv = nn.Conv2d(
    in_channels=1, out_channels=1, kernel_size= (1, scale * 2 + 1), padding=(0, scale), bias=False)
```

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+            up_layers.append(conv)
+        self.upsample_layers = nn.Sequential(*up_layers)
+
+    def forward(self, x: Tensor) -> Tensor:
+        r"""
+        Args:
+            x: the input sequence to the _UpsampleNetwork layer
+
+        Shape:
+            - x: :math:`(batch, freq, time)`.
```
mthrok, line 192: In Sphinx docstring, code block is double back-ticks.

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+                                       n_rnn=512,
+                                       n_fc=512,
+                                       kernel_size=5,
+                                       n_freq=128,
+                                       n_hidden=128,
+                                       n_output=128,
+                                       mode='RAW')
+        >>> x = torch.rand(10, 24800, 512)
+        >>> mels = torch.rand(10, 128, 512)
+        >>> output = upsamplenetwork(x, mels)
```
mthrok, line 240: Are these examples ran by documentation generator?
I do not see anything special about this usage, so unless it is here to be tested at documentation generation time, it is not adding much value to the documentation.

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+        sample_rate: the rate of audio dimensions (samples per second)
+        hop_length: the number of samples between the starts of consecutive frames
+        n_res_block: the number of ResBlock in stack (default=10)
+        n_rnn: the dimension of RNN layer (default=512)
+        n_fc: the dimension of fully connected layer (default=512)
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
+        n_freq: the number of bins in a spectrogram (default=128)
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        mode: the type of input waveform (default='RAW')
```
mthrok, line 223: What are the other available options for `mode`? Can you list all of them in docstring?

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+                 mode: str = 'RAW') -> None:
+        super().__init__()
+
+        self.mode = mode
+        self.kernel_size = kernel_size
+
+        if self.mode == 'RAW':
+            self.n_classes = 2 ** n_bits
+        elif self.mode == 'MOL':
+            self.n_classes = 30
```
mthrok, line 264: Can you throw error when `mode` is invalid?

torchaudio/models/_wavernn.py
```diff
@@ -50,56 +48,277 @@ def forward(self, x: Tensor) -> Tensor:
...
+
+        Shape:
+            - x: :math:`(batch, time)`
+            - mels: :math:`(batch, freq, time_mels)`
+            - output: :math:`(batch, time, 2 ** n_bits)`
+        """
+
+        batch_size = x.size(0)
+        h1 = torch.zeros(1, batch_size, self.n_rnn, device=x.device)
+        h2 = torch.zeros(1, batch_size, self.n_rnn, device=x.device)
```
mthrok, line 298: `dtype` is missing from `h1` and `h2`, so if `float64`, is passed, this will fail.

test/test_models.py
```diff
@@ -36,18 +36,72 @@ def test_mfcc(self):
...
         out = model(x)
 
-        assert out.size() == (batch_size, output_dims, num_features - pad * 2)
+        assert out.size() == (batch_size, n_output, n_time - kernel_size + 1)
+
+
+class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
+
+    def test_waveform(self):
+
```
mthrok, line 58: For each test, can you add comment on what is tested?
Test docstring can be one line or multiple lines but it should express what you are trying to test here.
It is often common that the written test is not what was intended to, and such wrong test is very difficult to detect and fix.

Use `pytest test --collect-only -v` and see examples of the current tests.

test/test_models.py
```diff
@@ -36,18 +36,72 @@ def test_mfcc(self):
...
+        out1, out2 = model(x)
+
+        assert out1.size() == (batch_size, total_scale * (n_time - kernel_size + 1), n_freq)
+        assert out2.size() == (batch_size, total_scale * (n_time - kernel_size + 1), n_output)
+
+
+class TestWaveRNN(common_utils.TorchaudioTestCase):
+
+    def test_waveform(self):
+
```
mthrok, line 84: same here, explanation of test is required. otherwise it will be difficult to make proper changes to this test later.

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
 
 
 class _ResBlock(nn.Module):
-    r"""This is a ResNet block layer. This layer is based on the paper "Deep Residual Learning
-    for Image Recognition". Kaiming He,  Xiangyu Zhang, Shaoqing Ren, Jian Sun. CVPR, 2016.
-    It is a block used in WaveRNN. WaveRNN is based on the paper "Efficient Neural Audio Synthesis".
-    Nal Kalchbrenner, Erich Elsen, Karen Simonyan, Seb Noury, Norman Casagrande, Edward Lockhart,
-    Florian Stimberg, Aaron van den Oord, Sander Dieleman, Koray Kavukcuoglu. arXiv:1802.08435, 2018.
+    r"""ResNet block layer based on
+    `"Deep Residual Learning for Image Recognition" <https://arxiv.org/pdf/1512.03385.pdf>`_
```
mthrok, line 12: short description has to fit in one line.  https://www.python.org/dev/peps/pep-0257/#one-line-docstrings

"""ResNet block based on "Deep Residual Learning for Image Recognition"

more detailed explanation here.
"""

test/test_models.py
```diff
@@ -36,18 +36,85 @@ def test_mfcc(self):
 class TestMelResNet(common_utils.TorchaudioTestCase):
 
     def test_waveform(self):
+        """
+        Create a tensor as the input of _MelResNet layer
+        and test if the output dimensions are correct.
+        """
```
mthrok, line 42: Please use the same docstring notation.

```
"""Short description on the first line as the opening of docstring

Then the more detailed explanation only if necessary.
Every test should be simple so if you need to add more explanation,
you should also consider the design of test and simplify it.
"""
```

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
             nn.ReLU(inplace=True),
             *ResBlocks,
-            nn.Conv1d(in_channels=hidden_dims, out_channels=output_dims, kernel_size=1)
+            nn.Conv1d(in_channels=n_hidden, out_channels=n_output, kernel_size=1)
         )
 
     def forward(self, x: Tensor) -> Tensor:
-        r"""Pass the input through the _MelResNet layer.
-
+        r"""
```
mthrok, line 84: What's your rational behind removing the original short description, `Pass the input through the _MelResNet layer.`??

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
-        P is the number of output sequence, T is the length of input sequence.
+            - x: :math:`(batch, freq, time)`
+            - output: :math:`(batch, n_output, time - kernel_size + 1)`
         """
 
         return self.melresnet_model(x)
+
+
+class _Stretch2d(nn.Module):
+    r"""Two-dimensional stretch layer.
```
mthrok, line 97: I do not think `stretch layer` is common enough and reading the description does not help understand what it does. Can you mention the outcome of the layer?

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
+        Shape:
+            - x: :math:`(..., freq, time)`
+            - output: :math:`(..., freq * y_scale, time * x_scale)`
+        """
+
+        return x.repeat_interleave(self.y_scale, 2).repeat_interleave(self.x_scale, 3)
+
+
+class _UpsampleNetwork(nn.Module):
+    r"""Upsample block based on a stack of Conv2d and Strech2d layers.
```
mthrok, line 132: > based on a stack of Conv2d and Strech2d layers

Instead of mentioning how it's implemented internally, explain what it does.

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
+
+    Examples::
+        >>> upsamplenetwork = _UpsampleNetwork(upsample_scales=[4, 4, 16],
+                                               n_res_block=10,
+                                               n_freq=128,
+                                               n_hidden=128,
+                                               n_output=128,
+                                               kernel_size=5)
+        >>> input = torch.rand(10, 128, 512)
+        >>> output = upsamplenetwork(input)
```
mthrok, line 150: This example is missing the most important information of what the output shape looks like

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
+        x = x.unsqueeze(1)
+        upsampling_output = self.upsample_layers(x)
+        upsampling_output = upsampling_output.squeeze(1)[:, :, self.indent:-self.indent]
+
+        return upsampling_output.transpose(1, 2), resnet_output.transpose(1, 2)
+
+
+class _WaveRNN(nn.Module):
+    r"""WaveRNN model based on
+    `"Efficient Neural Audio Synthesis" <https://arxiv.org/pdf/1802.08435.pdf>`_
```
mthrok, line 208: Remove the link from the summary of docstring and put it in the detail section.

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
+
+        self.mode = mode
+        self.kernel_size = kernel_size
+
+        if self.mode == 'RAW':
+            self.n_classes = 2 ** n_bits
+        elif self.mode == 'MOL':
+            self.n_classes = 30
+        else:
+            raise ValueError("Unknown input mode - {}".format(self.mode))
```
mthrok, line 265: Adding what is expected values would improve user experience.

```python
ValueError(f"Expected mode: `RAW` or `MOL`, but found {self.mode}")
```

mthrok: also instead of `.format`, using f-string will improve the readability of code.

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,325 @@
...
+                                       sample_rate=24000,
+                                       hop_length=200,
+                                       n_res_block=10,
+                                       n_rnn=512,
+                                       n_fc=512,
+                                       kernel_size=5,
+                                       n_freq=128,
+                                       n_hidden=128,
+                                       n_output=128,
+                                       mode='RAW')
```
vincentqb, line 236: This whole block needs to be a code block. Would it be nicer to write
```
>>> ... = _waveRNN(
>>>     n_bits=9,
>>>     ...
>>>  )

File: torchaudio/models/_wavernn.py (hunk 5 above)
Comment (vincentqb) [reply torchaudio/models/_wavernn.py:239]: I agree that the example doesn't the relationship between x and mels. How about the following?
```
>>> waveform = torchaudio.load(file)  # shape: batch x channel x time
>>> specgram = MelSpectrogram(waveform)   # shape: batch x freq x time`
>>> output = upsamplenetwork(waveform, specgram)  # shape: ...
```

File: torchaudio/models/_wavernn.py [hunk 20]
```diff
@@ -1,105 +1,325 @@
...
+        n_res_block: the number of ResBlock in stack (default=10)
+        n_rnn: the dimension of RNN layer (default=512)
+        n_fc: the dimension of fully connected layer (default=512)
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
+        n_freq: the number of bins in a spectrogram (default=128)
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        mode: the type of input waveform in ['RAW', 'MOL'] (default='RAW')
+
+    Examples::
```
vincentqb, line 224: nit: no needs for "::", see [example](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L382)

torchaudio/models/_wavernn.py
vincentqb: In fact, can you only specify the parameters that are required and omitted optional ones in this list? 
```
>>> upsamplenetwork = _waveRNN(
>>>     upsample_scales=[5,5,8],
>>>     n_bits=9,
>>>     sample_rate=24000,
>>>     hop_length=200,
>>> )
```

torchaudio/models/_wavernn.py
jimchen90: Nice suggestion! fixed.

torchaudio/models/_wavernn.py
jimchen90: Fixed, thanks.

torchaudio/models/_wavernn.py
jimchen90: I used the math block here inspired by the format in [transformer](https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/transformer.py#L85). But I agree that the code block is a good idea used here which is clear and matches other places. I will update this part.

torchaudio/models/_wavernn.py
jimchen90: Fixed.

torchaudio/models/_wavernn.py
jimchen90: Fixed.

torchaudio/models/_wavernn.py
jimchen90: `dtype` is added.

torchaudio/models/_wavernn.py
jimchen90: The description has been updated. Thanks for the comment and link information.

torchaudio/models/_wavernn.py
jimchen90: Nice suggestions, updated.

test/test_models.py
jimchen90: Test comments added.

test/test_models.py
jimchen90: Updated.

torchaudio/models/_wavernn.py
jimchen90: When I removed it, I was thinking that this description repeats because the forward function itself means passing the input through this function. I am okay with it and this sentence has been added in each function.

torchaudio/models/_wavernn.py
jimchen90: I agree, the description of outcome has been added.

torchaudio/models/_wavernn.py
jimchen90: The output shape is added.

torchaudio/models/_wavernn.py
jimchen90: fixed.

torchaudio/models/_wavernn.py
jimchen90: fixed.

torchaudio/models/_wavernn.py
jimchen90: I have added the whole block and omitted optional variables in the list.

torchaudio/models/_wavernn.py
jimchen90: fixed.

test/test_models.py
vincentqb: Thanks for also opening #751 to update this changes

test/test_models.py
jimchen90: Test comments added.

I updated this change to each stack PR (#751, #724, #735 )

torchaudio/models/_wavernn.py
```diff
@@ -1,105 +1,313 @@
...
-        P is the number of output sequence, T is the length of input sequence.
+        Return:
+            Tensor shape: (n_batch, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
         """
 
-        return self.melresnet_model(x)
+        batch_size = waveform.size(0)
+        h1 = torch.zeros(1, batch_size, self.n_rnn, dtype=waveform.dtype, device=waveform.device)
+        h2 = torch.zeros(1, batch_size, self.n_rnn, dtype=waveform.dtype, device=waveform.device)
+        mels, aux = self.upsample(specgram)
```
jimchen90, line 288: Need to add transpose here because the transpose operation in upsampling part is removed.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
         upsampling_output = upsampling_output.squeeze(1)[:, :, self.indent:-self.indent]
 
         return upsampling_output, resnet_output
+
+
+class _WaveRNN(nn.Module):
+    r"""WaveRNN model based on "Efficient Neural Audio Synthesis".
+
+    The paper link is https://arxiv.org/pdf/1802.08435.pdf. The input channels of waveform
```
vincentqb, line 201: nit: can we follow [this](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L512) example for the link?

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        n_res_block: the number of ResBlock in stack (default=10)
+        n_rnn: the dimension of RNN layer (default=512)
+        n_fc: the dimension of fully connected layer (default=512)
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
+        n_freq: the number of bins in a spectrogram (default=128)
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        mode: the mode of waveform in ['waveform', 'mol'] (default='waveform')
+
+    Examples
```
vincentqb, line 218: nit: this is one example?

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
         upsampling_output = upsampling_output.squeeze(1)[:, :, self.indent:-self.indent]
 
         return upsampling_output, resnet_output
+
+
+class _WaveRNN(nn.Module):
+    r"""WaveRNN model based on "Efficient Neural Audio Synthesis".
+
+    The paper link is https://arxiv.org/pdf/1802.08435.pdf. The input channels of waveform
+    and spectrogram have to be 1. The product of upsample_scales must equal hop_length.
```
vincentqb, line 202: nit: The product of ``upsample_scales`` must equal ``hop_length``.
(with double backticks)

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        mels = mels.transpose(1, 2)
+        aux = aux.transpose(1, 2)
+
+        aux_idx = [self.n_aux * i for i in range(5)]
+        a1 = aux[:, :, aux_idx[0]:aux_idx[1]]
+        a2 = aux[:, :, aux_idx[1]:aux_idx[2]]
+        a3 = aux[:, :, aux_idx[2]:aux_idx[3]]
+        a4 = aux[:, :, aux_idx[3]:aux_idx[4]]
+
+        x = torch.cat([waveform.unsqueeze(-1), mels, a1], dim=2)
```
vincentqb, line 300: nit: `waveform.unsqueeze(-1)` adds a 1 dimension at the end, and `dim=2` means to concatenate on the last dimension, right? if so, then let's use `dim=-1` just for clarity.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+
+    def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
+        r"""Pass the input through the _WaveRNN model.
+
+        Args:
+            waveform: the input waveform to the _WaveRNN layer (n_batch, (n_time - kernel_size + 1) * hop_length)
+            specgram: the input spectrogram to the _WaveRNN layer (n_batch, n_freq, n_time)
+
+        Return:
+            Tensor shape: (n_batch, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
```
vincentqb, line 284: can you remind me why we put the classification dimension at the end?

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        res = x
+        x = torch.cat([x, a2], dim=2)
+        x, _ = self.rnn2(x, h2)
+
+        x = x + res
+        x = torch.cat([x, a3], dim=2)
+        x = self.relu1(self.fc1(x))
+
+        x = torch.cat([x, a4], dim=2)
+        x = self.relu2(self.fc2(x))
```
vincentqb, line 315: to have the same shape for all, let's do:
```
        x = torch.cat([x, a4], dim=2)
        x = self.fc2(x)
        x = self.relu2(x)
```
same above

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        x = x + res
+        res = x
+        x = torch.cat([x, a2], dim=2)
+        x, _ = self.rnn2(x, h2)
+
+        x = x + res
+        x = torch.cat([x, a3], dim=2)
+        x = self.relu1(self.fc1(x))
+
+        x = torch.cat([x, a4], dim=2)
```
vincentqb, line 314: nit: `dim=-1` and above

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
+        n_freq: the number of bins in a spectrogram (default=128)
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        mode: the mode of waveform in ['waveform', 'mol'] (default='waveform')
+
+    Examples
+        >>> wavernn = _waveRNN(upsample_scales=[5,5,8], n_bits=9, sample_rate=24000, hop_length=200)
+        >>> waveform, sample_rate = torchaudio.load(file)  # waveform shape:
+        >>> (n_batch, n_channel, (n_time - kernel_size + 1) * hop_length)
```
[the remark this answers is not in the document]
jimchen90: Good suggestion. Updated.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        >>> specgram = MelSpectrogram(sample_rate)(waveform)  # shape: (n_batch, n_channel, n_freq, n_time)
+        >>> output = wavernn(waveform.squeeze(1), specgram.squeeze(1))  # shape:
+        >>> (n_batch, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
```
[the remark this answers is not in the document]
jimchen90: Updated.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+
+    Args:
+        upsample_scales: the list of upsample scales
+        n_bits: the bits of output waveform
+        sample_rate: the rate of audio dimensions (samples per second)
+        hop_length: the number of samples between the starts of consecutive frames
+        n_res_block: the number of ResBlock in stack (default=10)
+        n_rnn: the dimension of RNN layer (default=512)
+        n_fc: the dimension of fully connected layer (default=512)
+        kernel_size: the number of kernel size in the first Conv1d layer (default=5)
```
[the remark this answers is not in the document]
jimchen90: Since other blocks #751 #724 have the same format (use `=`). I plan to open a separate pull request to change all of them later. Now I will keep them same here.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,125 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+            specgram: the input spectrogram to the _WaveRNN layer (n_batch, n_freq, n_time)
+
+        Return:
+            Tensor shape: (n_batch, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
+        """
+
+        batch_size = waveform.size(0)
+        h1 = torch.zeros(1, batch_size, self.n_rnn, dtype=waveform.dtype, device=waveform.device)
+        h2 = torch.zeros(1, batch_size, self.n_rnn, dtype=waveform.dtype, device=waveform.device)
+        mels, aux = self.upsample(specgram)
```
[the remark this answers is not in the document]
jimchen90: Yes, `specgram` looks better. I have updated them.

torchaudio/models/_wavernn.py
jimchen90: Updated.

torchaudio/models/_wavernn.py
jimchen90: Thanks. Updated.

torchaudio/models/_wavernn.py
jimchen90: Yes. I have changed it to `dim = -1`. Same for below.

torchaudio/models/_wavernn.py
jimchen90: updated.

torchaudio/models/_wavernn.py
jimchen90: updated.

torchaudio/models/_wavernn.py
jimchen90: Other example is added.

jimchen90:
Failed tests relate to #766

torchaudio/models/_wavernn.py
vincentqb: I just meant there was a grammar typo "example" not "examples" :)

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,131 @@ def forward(self, specgram: Tensor) -> Tensor:
         upsampling_output = upsampling_output.squeeze(1)[:, :, self.indent:-self.indent]
 
         return upsampling_output, resnet_output
+
+
+class _WaveRNN(nn.Module):
+    r"""WaveRNN model based on "Efficient Neural Audio Synthesis".
+
+    The paper link is `<https://arxiv.org/pdf/1802.08435.pdf>`_. The input channels of waveform
```
vincentqb, line 201: nit:  For more info see the paper `Efficient Neural Audio Synthesis <https://arxiv.org/pdf/1802.08435.pdf>`_

vincentqb: Given the feedback in [comment](https://github.com/pytorch/audio/pull/749#issuecomment-653630756), I suggest we make it more explicit that this version is not the original.

```
"""
WaveRNN model based on the implementation from `fatchord <https://github.com/fatchord/WaveRNN>`_ .

The original implementation was introduced in `"Efficient Neural Audio Synthesis" <https://arxiv.org/pdf/1802.08435.pdf>`_.

The input channels of waveform and spectrogram have to be 1. The product of `upsample_scales` must equal `hop_length`.
"""
```

If it doesn't make the first line too long, I would merge the first and second paragraph. Thoughts?

vincentqb: Another option is to add a flag in WaveRNN init, say `version` that can only take the value `fatchord` for now, but could eventually also take the value `deepmind` later if/when implemented. Thoughts?

vincentqb: (Thanks for the feedback in [comment](https://github.com/pytorch/audio/pull/749#issuecomment-653630756) @PetrochukM :))

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,132 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        n_hidden: the number of hidden dimensions (default=128)
+        n_output: the number of output dimensions (default=128)
+        mode: the mode of waveform in ['waveform', 'mol'] (default='waveform')
+
+    Example
+        >>> wavernn = _waveRNN(upsample_scales=[5,5,8], n_bits=9, sample_rate=24000, hop_length=200)
+        >>> waveform, sample_rate = torchaudio.load(file)
+        >>> # waveform shape: (n_batch, n_channel, (n_time - kernel_size + 1) * hop_length)
+        >>> specgram = MelSpectrogram(sample_rate)(waveform)  # shape: (n_batch, n_channel, n_freq, n_time)
+        >>> output = wavernn(waveform.squeeze(1), specgram.squeeze(1))
```
vincentqb, line 225: let's squeeze/unsqueeze in the forward pass so that we expose the standard torchaudio convention here too. we'll need an assert that there's only one channel.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,132 @@ def forward(self, specgram: Tensor) -> Tensor:
...
 
         return upsampling_output, resnet_output
+
+
+class _WaveRNN(nn.Module):
+    r"""WaveRNN model based on the implementation from `fatchord <https://github.com/fatchord/WaveRNN>`_.
+
+    The original implementation was introduced in
+    `"Efficient Neural Audio Synthesis" <https://arxiv.org/pdf/1802.08435.pdf>`_.
+    The input channels of waveform and spectrogram have to be 1. The product of
```
vincentqb, line 203: This is currently not quite true, since the model assumes there is no channel dimension. However, I prefer leaving this assumption as it is and changing the forward pass around it.

jimchen90: I added the channel dimension in the waveform and spectrogram (also the input in `test_model.py` file). So the `unsqueeze(1)` and `squeeze(1)` operations are added in the forward function. A channel dimension assertion is also added before the `unsqueeze(1)` operation.

torchaudio/models/_wavernn.py
```diff
@@ -320,5 +324,6 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
         x = torch.cat([x, a4], dim=-1)
         x = self.fc2(x)
         x = self.relu2(x)
+        x = self.fc3(x).unsqueeze(1)
```
vincentqb, line 327: nit: let's make that a new paragraph :)
```python
        x = torch.cat([x, a4], dim=-1)
        x = self.fc2(x)
        x = self.relu2(x)
        x = self.fc3(x)

        return x.unsqueeze(1)
```

torchaudio/models/_wavernn.py
```diff
@@ -279,13 +279,17 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+            specgram: the input spectrogram to the _WaveRNN layer (n_batch, 1, n_freq, n_time)
 
         Return:
-            Tensor shape: (n_batch, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
+            Tensor shape: (n_batch, 1, (n_time - kernel_size + 1) * hop_length, 2 ** n_bits)
         """
 
+        assert waveform.size(1) == 1, 'Require the input channel of waveform is 1'
+        assert specgram.size(1) == 1, 'Require the input channel of specgram is 1'
+        waveform, specgram = waveform.squeeze(1), specgram.squeeze(1)
```
vincentqb, line 291: nit:
```
        # remove channel dimension until the end
        waveform, specgram = waveform.squeeze(1), specgram.squeeze(1)
```

jimchen90: The comment has been added.

torchaudio/models/_wavernn.py
jimchen90: The comment has been added.

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,139 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+                 mode: str = 'waveform') -> None:
+        super().__init__()
+
+        self.mode = mode
+        self.kernel_size = kernel_size
+
+        if self.mode == 'waveform':
+            self.n_classes = 2 ** n_bits
+        elif self.mode == 'mol':
+            self.n_classes = 30
```
vincentqb, line 250: Let's replace `mode` and `n_bits` parameters simply by `n_classes`.

cc [comment](https://github.com/pytorch/audio/pull/749#discussion_r456144384)

torchaudio/models/_wavernn.py
```diff
@@ -192,3 +193,139 @@ def forward(self, specgram: Tensor) -> Tensor:
...
+        self.hop_length = hop_length
+        self.sample_rate = sample_rate
+
+        total_scale = 1
+        for upsample_scale in upsample_scales:
+            total_scale *= upsample_scale
+        if total_scale != self.hop_length:
+            raise ValueError(f"Expected: total_scale == hop_length, but found {total_scale} != {hop_length}")
+
+        self.upsample = _UpsampleNetwork(upsample_scales, n_res_block, n_freq, n_hidden, n_output, kernel_size)
```
vincentqb, line 265: Let's change to `n_hidden_resblock` and `n_output_upsample`. Thoughts?

jimchen90: I updated the name as `n_hidden_resblock` and `n_output_melresnet`.
because the `n_output` is used in melresenet block and melresnet block is one part of upsample block, so I use `n_output_melresnet`. Any suggestion?

torchaudio/models/_wavernn.py
jimchen90: Yes, I changed the `mode` to `loss` and replaced `n_bits` by `n_classes` in #797 .

PR #742, opened 2020-06-23
Refactor Cache bust mechanism and bust on daily basis

mthrok:
This PR refactors cache generation mechanism by introducing dedicated command
and bust cache on daily basis.

At this moment, Windows unittest job for 3.6 and 3.7 are broken because of
broken scipy but the environment is cached this persists at least until the next week.
As we have nightly build, we do not need to keep cache for one week.

Note: the SciPy distribution seems to be broken, so this PR does not fix the ongoing Windows unittest job failure.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,12 @@ executors:
       shell: bash.exe
 
 commands:
+  generate_cache_key:
+    description: "Generates a cache key file that changes daily"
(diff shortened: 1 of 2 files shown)


peterjc123:
@mthrok To fix that for Windows, just pin scipy to 1.4.1.

mthrok:
Thanks!

PR #744, opened 2020-06-23
Unittest function in test

jimchen90:
This is related to #727. 
Compared to this pull request and other test files (e.g. [here](https://github.com/pytorch/audio/blob/master/test/test_datasets.py)), I remove the `unittest.main` function in the end and `import unittest` function in the first line in #727 .

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -1,3 +1,5 @@
+import unittest
+
 import torch
 from torchaudio.models import Wav2Letter, _MelResNet
 
@@ -51,3 +53,7 @@ class TestMelResNet(common_utils.TorchaudioTestCase):
         out = model(x)
 
         assert out.size() == (batch_size, output_dims, num_features - pad * 2)
+
+
+if __name__ == "__main__":
+    unittest.main()


mthrok:
What value does this PR bring?

jimchen90:
> Can you provide description of what value this PR brings?

I want to separate this change with #727 and mention this difference in model test file when compare to other test files. It is for reference and it doesn't need to be merged.

mthrok:
> it doesn't need to be merged.

if so can you close it?

vincentqb:
Thanks for following up on [comment](https://github.com/pytorch/audio/pull/727#discussion_r442361565) and opening this @jimchen90 :) it'd be nice to understand the pros and cons of merging this or not, since we may want to apply this to all test files.

* If we decide to not add this, then some test files have it, and some don't. What should we do looking forward?

* If we decide to add this, then all the test files have it, and we would like to maintain consistency, and add that to other tests moving forward. One advantage is that users can run `python test.py` to run the tests locally.

* If we decide to remove this, then users who want to run this using the command line need to use `python -m pytest test.py`. Other way?

I'm ok with either last way, but we should make a conscious decision :) @mthrok thoughts?

mthrok:
Thanks for following up on [comment](https://github.com/pytorch/audio/pull/727#discussion_r442361565) and opening this @jimchen90 :) it'd be nice to understand the pros and cons of merging this or not, since we may want to apply this to all test files.

* If we decide to add this, then all the test files should have it to maintain consistency. We'd then add that to other tests moving forward. One advantage is that users can run `python test.py` to run the tests locally.  This does not require pytest, and, at this time, we prefer not depending on `pytest` due to internal systems.

* If we decide to remove this, then we should remove this from all files eventually, and not add it moving forward. Users who want to run this using the command line need to use `python -m pytest test.py`. Other way?

I'm ok to add it or not, but we should make an explicit decision so we can use that same decision moving forward :) @jimchen90 @mthrok thoughts?

vincentqb:
> * Firstly, none of our CI system requires this.
> * Secondly, it is not clear why it has to be `unittest.main` that will be invoked when ran as regular python script. Why not another test runner suite?
> * Thirdly, there is an ongoing discussion on adopting `pytest` in PyTorch testing framework. If we add this code, we have to change that in the future.

Cool, I agree with (1) and (3). Let's not add it to files moving forward, and we can remove 
```python
if __name__ == "__main__":
    unittest.main()
```
from existing files. If pytorch decides to go with a particular framework, we'll consider it at that time. Glad we documented our decision on this :)

PR #746, opened 2020-06-24
Cherry picking changes for release improvements

seemethere:
These were changes that were made in the process of releasing `v0.6.0-rc1`


seemethere:
It’s actually just to add the test channels to the install priorities. 

This is so we can install RCs when we designate a PYTORCH_VERSION

Issue #740, opened 2020-06-23
Very slow speed in complex_norm() function.


lbjcom:
## 🐛 Bug

When I call `spectrogram()` in torchaudio 0.5.0 my training script runs very slow. I downgraded to pytorch 1.4 and torchaudio 0.4.0 then the script runs fast. So I compared the difference betweeen 0.4.0 and 0.5.0 and found that the complex_norm() is the root cause of the slow speed: 

https://github.com/pytorch/audio/blob/bc1df4882b967d55d8544e572ef16abeee9f45b5/torchaudio/functional.py#L171

If I install torchaudio 0.5.0 and change the line 171 to the 0.4.0's one: 

```python
- spec_f = complex_norm(spec_f, power=power)
+ spec_f = spec_f.pow(power).sum(-1)  # get power of "complex" tensor
```

and the speed is same as 0.4.0's.

For example in my training script with torachaudio 0.5.0:
```
--------- TRAINING - Epoch:     1/ 1000 ------------
| Batch:  2461/ 2461, 100.00%                      |
| Loss:   3.253154@lr=1.000000e-03                 |
| Speed:   69.64 files/sec, Elapsed Time: 00:58:53 |
----------------------------------------------------
```

The speed is about 70 files/sec. But with the 0.4.0's code:
```
--------- TRAINING - Epoch:     1/ 1000 ------------
| Batch:  2461/ 2461, 100.00%                      |
| Loss:   3.275024@lr=1.000000e-03                 |
| Speed:  408.14 files/sec, Elapsed Time: 00:10:02 |
----------------------------------------------------
```

As you can see the speed of 0.4.0's is about 5.7x faster than 0.5.0's. I can not share the script because the code is from my company. But I think that the speed difference comes from the process that call's the complex_norm() function. For example [voxceleb_trainer](https://github.com/clovaai/voxceleb_trainer) calls `spectrogram()` in main process and there's no speed reduction while my script calls spectrogram() in a forked process which is forked by pytorch's DataLoader.

## To Reproduce

1. Install torchaudio 0.5.0
1. run the script.
1. get slow speed
1. change the `spec_f = complex_norm(spec_f, power=power)` to `spec_f = spec_f.pow(power).sum(-1)`
1. run the script again.
1. get normal speed

## Expected behavior

- the speed of my script becomes normal when I use either torchaudio 0.4.0 or 0.5.0.

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)?
    - pip
[...]

PR #747, opened 2020-06-24
rollback torch.norm() in spectrogram()

lbjcom:
rollback torch.norm() in spectrogram() to v0.4.0 because torch.norm() is very slow in CPU mode as mentioned in following issues:
- https://github.com/pytorch/audio/issues/740
- https://github.com/pytorch/audio/issues/455

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -169,7 +169,6 @@ def spectrogram(
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
         spec_f = complex_norm(spec_f, power=power)
-        
 
     return spec_f
 
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -169,6 +169,9 @@ def spectrogram(
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
         spec_f = complex_norm(spec_f, power=power)
+        # Replace by torch.norm once issue is fixed
+        # https://github.com/pytorch/pytorch/issues/34279
+        spec_f = spec_f.pow(2.).sum(-1).pow(0.5 * power)
 
     return spec_f
 
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -168,10 +168,8 @@ def spectrogram(
     if normalized:
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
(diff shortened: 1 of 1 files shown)


torchaudio/functional.py
```diff
@@ -169,6 +169,9 @@ def spectrogram(
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
         spec_f = complex_norm(spec_f, power=power)
```
mthrok, line 171:
```
# spec_f = complex_norm(spec_f, power=power)
```
Did you forget to comment out the original?

lbjcom: Thanks! I just updated it.

vincentqb:
Thanks for submitting this! It is indeed enough to address the original concern. However, as mentioned [here](https://github.com/pytorch/audio/issues/740#issuecomment-648899787), this affects any function invoking `complex_norm` (since the latter uses `torch.norm`). I would prefer fixing `complex_norm` [here](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L490) instead of `spectrogram`. Is there a reason why you modified `spectrogram` instead?

lbjcom:
> Thanks for submitting this! It is indeed enough to address the original concern. However, as mentioned [here](https://github.com/pytorch/audio/issues/740#issuecomment-648899787), this affects any function invoking `complex_norm` (since the latter uses `torch.norm`). I would prefer fixing `complex_norm` [here](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L490) instead of `spectrogram`. Is there a reason why you modified `spectrogram` instead?

That's a good point! The reason I changed the `spectrogram()` is that the difference between v0.4.0 and v0.5.0 is from that function:
image

But there's no change in `complex_norm()` between v0.4.0 and v0.5.0. So I changed `spectrogram()` since I don't know all history of torchaudio. If there's no problem, as you mentioned, fixing `complex_norm()` is the right solution 😄

vincentqb:
Thanks for being careful :) In this case though, I do believe changing `complex_norm` is the better solution: not only will it fix your problem, but it should also make `complex_norm` faster :)

lbjcom:
> Thanks for being careful! In this case though, changing `complex_norm` is the better solution: not only will it fix your problem, but it will also make `complex_norm` faster :)

I updated the code as you mentioned  and found out that there's no problem in my training script. Thanks!

PR #748, opened 2020-06-24
[BC Breaking] Replace sox_effects init/list/shutdown with TS binding

mthrok:
This PR replaces sox effects `init/list/shutdown` functions with Torch Script binding. (SoxEffectsChain is not changed yet.)
I also moved error checking to C++, so the function signature is changed -> BC breaking.
The previous implementation returned `sox_errot_t` enum to Python, which is practically useless so I think this BC breaking does not have an impact.

### DIFF
--- a/torchaudio/csrc/register.cpp
+++ b/torchaudio/csrc/register.cpp
@@ -1,6 +1,7 @@
 #ifndef TORCHAUDIO_REGISTER_H
 #define TORCHAUDIO_REGISTER_H
 
+#include <torchaudio/csrc/sox_effects.h>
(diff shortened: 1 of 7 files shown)


vincentqb:
This means that `sox_effects` would be torchscriptable, right? If so, could you add a test for this?

mthrok:
> This means that `sox_effects` would be torchscriptable, right? If so, could you add a test for this?

Nope. the main `sox_effects` are more complicated so it's not included in this PR, and there is not really way to test init/shutdown/list in the current framework.

mthrok:
thanks

Issue #80, opened 2019-02-10
Which library is torchaudio consistent against?


PCerles:
Hi, I'm currently updating my torch codebase from using librosa to torchaudio for transforms, to take advantage of the (much) faster stft torch implementation on the GPU. However, running into several occasions where the output from Spectrogram vs. librosa.core._spectrogram, MelSpectrogram vs. librosa.melspectrogram have different results. Does this repo ensure consistency with another python audio library for those transformations? I think it would be good to have consistency with another widely used library.

faroit:
@PCerles great effort and this is really relevant, I agree. The problem is that even among other stft implementations,  its hard to get consistent results.  E.g. see [librosa vs. scipy](https://gist.github.com/bmcfee/746e572232be36f3bd462749fb1796da)

PCerles:
Thanks! The torch.stft implementation is consistent with librosa defaults if you use the same window type, I tested that first.
```
window = torch.hann_window(n_fft)
torch.stft(sound, hop_length = hop_length, n_fft=n_fft, window=window)
==
librosa.stft(sound, hop_length, n_fft)
```
I was more interested in the function signatures being consistent and any transforms being visible/configurable (e.g. the spectrogram normalization), but didn't know if these functions were modeled off some library. Personally, I think it makes sense to mirror librosa functionality for functions like MelSpectrogram, since librosa seems to be the most popular python audio library. What do you think? I'm sure the current torchaudio default implementations also make sense for some applications.

soumith:
totally open to a PR for this.
cc; @dhpollack

dhpollack:
First off, I'm open to a PR as well.

Secondly, the MEL function converts the spectrogram to the DB scale because I felt that was a bit more intuitive than the log scale. There is also the choice of a power spectrogram vs the amplitude spectrogram. I played with a lot of that but in the end, I wasn't sure it was worth mucking things up with more options. 

I'm a believer in the paradox of choice theory that too many choices can be a bad thing. I also sort of think librosa has too many options especially for the typical use case of torchaudio. I'm not totally against being consistent with librosa but it's not perfect.

One quick example is librosa by default converts audio to a sample rate of 22k instead of just reading the audios actual sample rate. That's behavior that I don't think should be emulated. Again, if the people want this, I'm open to it.

PCerles:
Alright thanks guys, PR'd here: 
https://github.com/pytorch/audio/pull/83

PR #749, opened 2020-06-24
Add wavernn example pipeline

jimchen90:
This is a reference example using WaveRNN model to train on LJSpeech. The structure will be inspired by #632  and [WaveRNN](https://github.com/fatchord/WaveRNN).

There are at least a few more things to do:

- [x] Add bg_iterator and README.
- [x] ~~Add torchaudio transforms on mel-spectrogram.~~

Related to #446

Stack:

~~Add MelResNet Block #705, #751~~
~~Add Upsampling Block #724~~
~~Add WaveRNN Model #735~~
**Add example pipeline with WaveRNN #749**

cc @cpuhrsch @zhangguanheng66 
[internal](https://fb.quip.com/YM2zAHoRU6Tg)


examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,78 @@
...
+        file = self.files[index]
+        x, sample_rate = torchaudio.load(file)
+        mel = self.transforms(x)
+
+        bits = 16 if self.mode == 'MOL' else self.n_bits
+
+        x = (x + 1.) * (2 ** bits - 1) / 2
+        x = torch.clamp(x, min=0, max=2 ** bits - 1)
+
+        return mel.squeeze(0), x.int().squeeze(0)
```
vincentqb, line 32: This converts representation of a waveform from [-1, 1] to 16-bit integer representation. For instance, this is done in [load_wav](https://github.com/pytorch/audio/blob/master/torchaudio/backend/sox_backend.py#L68) already. Since this is an important step and can be generalized, let's make this into a function within torchaudio. One point of discussion is whether we add that directly in WaveRNN.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,489 @@
...
+    return sum(p.numel() for p in model.parameters() if p.requires_grad)
+
+
+def train_one_epoch(model, criterion, optimizer, data_loader, device, pbar=None):
+
+    model.train()
+
+    sums = defaultdict(lambda: 0.0)
+
+    for i, (x, m, y) in enumerate(data_loader):
```
vincentqb, line 249: nit: don't forget to remove i once done debugging :)

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,85 @@
...
+    return train_dataset, test_dataset
+
+
+def collate_factory(args):
+
+    def raw_collate(batch):
+
+        pad = (args.kernel_size - 1) // 2
+        seq_len = args.hop_length * args.seq_len_factor
+        mel_win = args.seq_len_factor + 2 * pad
```
vincentqb, line 61: pad is not defined within the function

vincentqb:
btw, can you add a README.md to discuss the pipeline?

vincentqb:
It'd be nice to get a baseline by comparing the error you get here to the output obtained by Griffin-Lim, say, and in other norms too L^1, L^2 for instance.

examples/pipeline_wavernn/loss_mol.py
```diff
@@ -0,0 +1,81 @@
+import torch
+import torch.nn.functional as F
+
+# Adapted from wavenet vocoder:
+# https://github.com/r9y9/wavenet_vocoder/blob/master/wavenet_vocoder/mixture.py
+# Explain mol loss:
+# https://github.com/Rayhane-mamah/Tacotron-2/issues/155
```
vincentqb, line 7: nit: please move this in the relevant docstring

examples/pipeline_wavernn/loss_mol.py
```diff
@@ -0,0 +1,81 @@
...
+    axis = len(x.size()) - 1
+    m, _ = torch.max(x, dim=axis)
+    m2, _ = torch.max(x, dim=axis, keepdim=True)
+    return m + torch.log(torch.sum(torch.exp(x - m2), dim=axis))
+
+
+def LossFn_Mol(y_hat, y, num_classes=65536, log_scale_min=None, reduce=True):
+    """ calculate the loss of mol mode"""
+
+    min_value = -32.23619130191664  # = float(np.log(1e-14))
```
vincentqb, line 24: nit: how about `torch.log(1e-14).item()`?

examples/pipeline_wavernn/loss_mol.py
```diff
@@ -0,0 +1,81 @@
...
+    cdf_delta = cdf_plus - cdf_min
+
+    mid_in = inv_stdv * centered_y
+    # log probability in the center of the bin, to be used in extreme cases
+    # (not actually used in our code)
+    log_pdf_mid = mid_in - log_scales - 2. * F.softplus(mid_in)
+
+    inner_inner_cond = (cdf_delta > 1e-5).float()
+
+    tmp = 10.397192449493701  # = np.log((num_classes - 1) / 2)
```
vincentqb, line 66: nit: same about `torch.log` or even just `math.log` since these are just scalars. same for above.

vincentqb: also, what if you are changing the number of classes?

examples/pipeline_wavernn/loss_mol.py
```diff
@@ -0,0 +1,81 @@
+import torch
+import torch.nn.functional as F
+
+# Adapted from wavenet vocoder:
+# https://github.com/r9y9/wavenet_vocoder/blob/master/wavenet_vocoder/mixture.py
+# Explain mol loss:
+# https://github.com/Rayhane-mamah/Tacotron-2/issues/155
+
+# Remove numpy dependency
```
vincentqb, line 9: lint?

examples/pipeline_wavernn/loss_mol.py
```diff
@@ -0,0 +1,81 @@
...
+def log_sum_exp(x):
+    """ numerically stable log_sum_exp implementation that prevents overflow """
+
+    axis = len(x.size()) - 1
+    m, _ = torch.max(x, dim=axis)
+    m2, _ = torch.max(x, dim=axis, keepdim=True)
+    return m + torch.log(torch.sum(torch.exp(x - m2), dim=axis))
+
+
+def LossFn_Mol(y_hat, y, num_classes=65536, log_scale_min=None, reduce=True):
```
vincentqb, line 21: let's follow a standard naming, like `mol_loss`?

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,59 @@
...
+        return torch.from_numpy(waveform)
+
+    def stft(self, y):
+        return librosa.stft(
+            y=y,
+            n_fft=self.n_fft, hop_length=self.hop_length, win_length=self.win_length)
+
+    def linear_to_mel(self, spectrogram):
+        return librosa.feature.melspectrogram(
+            S=spectrogram, sr=self.sample_rate, n_fft=self.n_fft, n_mels=self.num_mels, fmin=self.fmin)
```
vincentqb, line 35: nit: can you explicitly invoke `slaney` (even though it's the default) since we depend on this here?

#593

PetrochukM:
Since this is not the original WaveRNN model, I'd recommend renaming it as "FatchordWaveRNN" or something similar.

examples/pipeline_wavernn/loss_mol.py
jimchen90: The docstring has been updated.

examples/pipeline_wavernn/loss_mol.py
jimchen90: Good suggestion. It has been updated.

examples/pipeline_wavernn/loss_mol.py
jimchen90: Updated.

examples/pipeline_wavernn/loss_mol.py
jimchen90: This part has been moved to the docstring.

examples/pipeline_wavernn/loss_mol.py
jimchen90: I change the file name to `mol_loss` and class name to `Mol_Loss`.

examples/pipeline_wavernn/transform.py
jimchen90: Yes, the transform of mel-spectrogram uses the `norm='slaney'`, `htk=False` (the default in librosa) here. 

Related to #593 #80

examples/pipeline_wavernn/wavernn.py
jimchen90: This part has been updated by `bg_iterator`.

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,35 @@
+This is an example pipeline for WaveRNN vocoder using the WaveRNN model trained with LJSPEECH. WaveRNN and LJSPEECH are available in torchaudio.
```
vincentqb, line 1: You can add more references here.

jimchen90: References are added.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+            waveform = float_2_int(waveform, 16)
+
+        return waveform, specgram
+
+    def __len__(self):
+        return len(self.files)
+
+
+# From float waveform [-1, 1] to integer label [0, 2 ** bits - 1]
+def float_2_int(waveform, bits):
```
vincentqb, line 46: Please add a standard docstring describing what this is doing, instead of the comment above.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+    assert abs(waveform).max() <= 1.0
+    waveform = (waveform + 1.) * (2**bits - 1) / 2
+    return torch.clamp(waveform, 0, 2**bits - 1).int()
+
+
+# From integer label [0, 2 ** bits - 1] to float waveform [-1, 1]
+def int_2_float(waveform, bits):
```
vincentqb, line 53: same as above

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+        self.mode = mode
+
+    def __getitem__(self, index):
+
+        file = self.files[index]
+
+    # use torchaudio transform to get waveform and specgram
+        # waveform, sample_rate = torchaudio.load(file)
+        # specgram = self.transforms(x)
+        # return waveform.squeeze(0), mel.squeeze(0)
```
vincentqb, line 25: remove?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # use librosa transform to get waveform and specgram
```
vincentqb, line 27: nit: indentation

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,114 @@
...
+        self.files = files
+        self.mulaw = mulaw
+        self.n_bits = n_bits
+        self.mode = mode
+
+    def __getitem__(self, index):
+
+        file = self.files[index]
+
+    # use torchaudio transform to get waveform and specgram
```
vincentqb, line 22: nit: indentation

examples/pipeline_wavernn/mol_loss.py
```diff
@@ -0,0 +1,89 @@
...
+def log_sum_exp(x):
+    """ numerically stable log_sum_exp implementation that prevents overflow """
+
+    axis = len(x.size()) - 1
+    m, _ = torch.max(x, dim=axis)
+    m2, _ = torch.max(x, dim=axis, keepdim=True)
+    return m + torch.log(torch.sum(torch.exp(x - m2), dim=axis))
+
+
+def Mol_Loss(y_hat, y, num_classes=65536, log_scale_min=None, reduce=True):
```
vincentqb, line 14: nit: one word?

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,59 @@
+import librosa
+import numpy as np
```
vincentqb, line 2: can you add a comment somewhere in module linking the issue to remove the need for librosa? #593

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,59 @@
...
+        self.n_fft = n_fft
+        self.hop_length = hop_length
+        self.win_length = win_length
+        self.num_mels = num_mels
+        self.fmin = fmin
+        self.min_level_db = min_level_db
+
+    def load(self, path):
+        waveform = librosa.load(path, sr=self.sample_rate)[0]
+        return torch.from_numpy(waveform)
```
vincentqb, line 26: this can be replaced by torchaudio

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def stft(self, y):
+        return librosa.stft(
+            y=y,
+            n_fft=self.n_fft, hop_length=self.hop_length, win_length=self.win_length)
```
vincentqb, line 31: can you run lint like black on this file to reformat this?

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,59 @@
...
+        return np.clip((S - self.min_level_db) / - self.min_level_db, 0, 1)
+
+    def denormalize(self, S):
+        return (np.clip(S, 0, 1) * -self.min_level_db) + self.min_level_db
+
+    def amp_to_db(self, x):
+        return 20 * np.log10(np.maximum(1e-5, x))
+
+    def db_to_amp(self, x):
+        return np.power(10.0, x * 0.05)
```
vincentqb, line 47: Why is numpy used here? We can replace all of this by pytorch tensor operations.

examples/pipeline_wavernn/transform.py
vincentqb, line 35: This may be the only function that is actually needed here. The rest can be done in pytorch, no?

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+import torchaudio
+from transform import Transform
+from datasets import datasets_ljspeech, collate_factory
+from typing import List
+from torchaudio.models import _WaveRNN
+from torch.utils.data import DataLoader
+from torchaudio.datasets.utils import bg_iterator
+from torch.optim.lr_scheduler import ExponentialLR, ReduceLROnPlateau
+from torch.optim import Adam
+from mol_loss import Mol_Loss
```
vincentqb, line 20: let's call this file "losses"

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        help="if used, model is jitted"
+    )
+    parser.add_argument(
+        '--resume',
+        default='',
+        type=str,
+        metavar='PATH',
+        help='path to latest checkpoint'
+    )
+    # the product of `upsample_scales` must equal `hop_length`
```
vincentqb, line 156: intentional?

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        help="the path of audio files",
+    )
+
+    args = parser.parse_args()
+    return args
+
+
+def signal_handler(a, b):
+    global SIGNAL_RECEIVED
+    print("Signal received", a, datetime.now().strftime("%y%m%d.%H%M%S"), flush=True)
```
vincentqb, line 273: let's use logging for this.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        default="/private/home/jimchen90/datasets/LJSpeech-1.1/wavs/",
+        type=str,
+        help="the path of audio files",
+    )
+
+    args = parser.parse_args()
+    return args
+
+
+def signal_handler(a, b):
```
vincentqb, line 271: this is needed to run on cluster, but elsewhere. we won't include this in this pipeline.

examples/pipeline_wavernn/wavernn.py
vincentqb: we'll just remove this function.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+
+        if SIGNAL_RECEIVED:
+            return
+
+    avg_loss = sums["loss"] / len(data_loader)
+
+    metric = MetricLogger("train_epoch")
+    metric("epoch", epoch)
+    metric("loss", avg_loss)
+    if "gradient" in sums:
```
vincentqb, line 342: let's remove the if condition. written as it is the code would simply log zeros.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+#        "hop_length": args.hop_length,
+#        "f_min": args.f_min,
+#        "win_length": args.win_length
+#    }
+
+#    transforms = torch.nn.Sequential(
+#        torchaudio.transforms.MelSpectrogram(
+#            sample_rate=args.sample_rate, **melkwargs
+#        ),
+#    )
```
vincentqb, line 422: intentional?

vincentqb: please add a comment why this is needed.

nit: align comment with rest of code

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+
+    melkwargs = {
+        "n_fft": 2048,
+        "num_mels": args.n_freq,
+        "hop_length": args.hop_length,
+        "fmin": args.f_min,
+        "win_length": args.win_length,
+        "sample_rate": args.sample_rate,
+        "min_level_db": args.min_level_db
+    }
```
vincentqb, line 434: this should be the same with librosa or pytorch. please create an interface for librosa that mimics pytorch.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+                     hop_length=args.hop_length,
+                     n_res_block=args.n_res_block,
+                     n_rnn=args.n_rnn,
+                     n_fc=args.n_fc,
+                     kernel_size=args.kernel_size,
+                     n_freq=args.n_freq,
+                     n_hidden=args.n_hidden,
+                     n_output=args.n_output,
+                     mode=args.mode,
+                     )
```
vincentqb, line 477: I don't remember what's the standard for `black`. Have you ran `black` on this file?

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        **loader_training_params,
+    )
+    loader_test = DataLoader(
+        test_dataset,
+        batch_size=args.batch_size,
+        collate_fn=collate_fn,
+        **loader_validation_params,
+    )
+
+    # model
```
vincentqb, line 464: Can you elaborate? I'm not sure what this comment is adding.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        )
+
+        if SIGNAL_RECEIVED:
+            save_checkpoint({
+                'epoch': epoch,
+                'state_dict': model.state_dict(),
+                'best_loss': best_loss,
+                'optimizer': optimizer.state_dict(),
+                'scheduler': scheduler.state_dict(),
+            }, False, args.checkpoint)
```
vincentqb, line 551: this will be removed with removing signal_handler

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+
+        optimizer.step()
+
+        metric("iteration", sums["iteration"])
+        metric("time", time() - start2)
+        metric.print()
+        sums["iteration"] += 1
+
+        if SIGNAL_RECEIVED:
+            return
```
vincentqb, line 335: this will be removed with removing signal_handler

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+            else:
+                raise ValueError(
+                    f"Expected mode: `waveform` or `mol`, but found {mode}"
+                )
+
+            loss = criterion(output, target)
+            sums["loss"] += loss.item()
+
+            if SIGNAL_RECEIVED:
+                break
```
vincentqb, line 383: this will be removed with removing signal_handler

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+    }
+
+    optimizer = Adam(model.parameters(), **optimizer_params)
+
+    if args.scheduler == "exponential":
+        scheduler = ExponentialLR(optimizer, gamma=args.gamma)
+    elif args.scheduler == "reduceonplateau":
+        scheduler = ReduceLROnPlateau(optimizer, patience=10, threshold=1e-3)
+
+    criterion = nn.CrossEntropyLoss() if args.mode == 'waveform' else Mol_Loss
```
vincentqb, line 503: I prefer this to raising errors in this pipeline: default to `MelLoss` if mode is not specified in the rest of your code. Others are free to disagree on this :)

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+            output = output.transpose(1, 2)
+            target = target.long()
+
+        elif mode == 'mol':
+            target = target.unsqueeze(-1)
+
+        else:
+            raise ValueError(
+                f"Expected mode: `waveform` or `mol`, but found {mode}"
+            )
```
vincentqb, line 310: I'd suggest using mol as the default value.
```python
        if mode == 'waveform':
            output = output.transpose(1, 2)
            target = target.long()

        else:
            # use mol mode
            target = target.unsqueeze(-1)
```

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+                output = output.transpose(1, 2)
+                target = target.long()
+
+            elif mode == 'mol':
+                target = target.unsqueeze(-1)
+
+            else:
+                raise ValueError(
+                    f"Expected mode: `waveform` or `mol`, but found {mode}"
+                )
```
vincentqb, line 377: same

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+from torchaudio.datasets.utils import bg_iterator
+from torch.optim.lr_scheduler import ExponentialLR, ReduceLROnPlateau
+from torch.optim import Adam
+from mol_loss import Mol_Loss
+from utils import MetricLogger, count_parameters, save_checkpoint
+
+SIGNAL_RECEIVED = False
+
+
+def parse_args():
```
vincentqb, line 26: Are the defaults your best ones so far?

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+from typing import List
+from torchaudio.models import _WaveRNN
+from torch.utils.data import DataLoader
+from torchaudio.datasets.utils import bg_iterator
+from torch.optim.lr_scheduler import ExponentialLR, ReduceLROnPlateau
+from torch.optim import Adam
+from mol_loss import Mol_Loss
+from utils import MetricLogger, count_parameters, save_checkpoint
+
+SIGNAL_RECEIVED = False
```
vincentqb, line 23: this will be removed

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+    )
+    parser.add_argument(
+        "--test-samples",
+        default=50,
+        type=float,
+        help="the number of waveforms for testing",
+    )
+    parser.add_argument(
+        "--file-path",
+        default="/private/home/jimchen90/datasets/LJSpeech-1.1/wavs/",
```
vincentqb, line 262: Let's use the current value as default.

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        "eps": args.eps,
+        "weight_decay": args.weight_decay,
+    }
+
+    optimizer = Adam(model.parameters(), **optimizer_params)
+
+    if args.scheduler == "exponential":
+        scheduler = ExponentialLR(optimizer, gamma=args.gamma)
+    elif args.scheduler == "reduceonplateau":
+        scheduler = ReduceLROnPlateau(optimizer, patience=10, threshold=1e-3)
```
vincentqb, line 501: Have you tried your code with this scheduler? It will generate an error, since `scheduler.step` needs a parameter :)

examples/pipeline_wavernn/wavernn.py
```diff
@@ -0,0 +1,586 @@
...
+        default=8,
+        type=int,
+        help="the bits of output waveform",
+    )
+    parser.add_argument(
+        "--sample-rate",
+        default=22050,
+        type=int,
+        help="the rate of audio dimensions (samples per second)",
+    )
```
vincentqb, line 174: is that not read from the audio file?

examples/pipeline_wavernn/wavernn.py
vincentqb, line 20: can you run `isort` to sort the imports in all the files?

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,36 @@
+This is an example pipeline for WaveRNN vocoder using the WaveRNN model trained with LJSpeech. WaveRNN model is based on the implementation from https://github.com/fatchord/WaveRNN. The original implementation was
```
vincentqb, line 1: nit: from [this repository](https://github.com/fatchord/WaveRNN).

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Usage
+
+More information about each command line parameters is available with the `--help` option. An example can be invoked as follows.
+```
+python main.py \
+    --batch-size 128 \
+    --learning-rate 1e-4 \
+    --n-freq 80 \
+    --mode 'waveform' \
+    --n_bits 8 \
```
vincentqb, line 35: `n_bits` doesn't follow the other convention `--n-bits`. please check other command line too.

examples/pipeline_wavernn/datasets.py
jimchen90: Good suggestion.
- I added the description in the function docstring.
- The name is changed to `waveform_to_label`.
- I put this function in the `utils` file.

examples/pipeline_wavernn/datasets.py
jimchen90: This part has been removed.

examples/pipeline_wavernn/mol_loss.py
jimchen90: Now I use `MoLLoss` as the function name and `losses` as the file name.

examples/pipeline_wavernn/transform.py
jimchen90: I add a comment in [module](https://github.com/pytorch/audio/pull/749/files#diff-f426c1a173746281ffa37227de895169R47)  and link #593 .

examples/pipeline_wavernn/transform.py
jimchen90: Yes, now only this function is needed. The rest has been updated with pytorch.

examples/pipeline_wavernn/wavernn.py
jimchen90: All this part has been removed.

examples/pipeline_wavernn/transform.py
jimchen90: This part has been replaced by pytorch operations.

examples/pipeline_wavernn/wavernn.py
jimchen90: if condition is removed.

examples/pipeline_wavernn/wavernn.py
jimchen90: Yes. I ran `black'.

examples/pipeline_wavernn/wavernn.py
jimchen90: I have removed this comment since it is not needed.

examples/pipeline_wavernn/wavernn.py
jimchen90: Good point. I have updated it.

examples/pipeline_wavernn/wavernn.py
jimchen90: Updated.

examples/pipeline_wavernn/wavernn.py
jimchen90: I didn't try this scheduler. Now this part is removed.

examples/pipeline_wavernn/wavernn.py
jimchen90: Yes, I ran `isort` to sort the imports.

examples/pipeline_wavernn/README.md
jimchen90: updated.

examples/pipeline_wavernn/README.md
jimchen90: Thanks. I have updated this and checked other lines.

examples/pipeline_wavernn/wavernn.py
jimchen90: Yes, now the arguments are the best for training.

examples/pipeline_wavernn/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    assert abs(waveform).max() <= 1.0
+    waveform = (waveform + 1.0) * (2 ** bits - 1) / 2
+    return torch.clamp(waveform, 0, 2 ** bits - 1).int()
+
+
+def label_to_waveform(label, bits):
+    r"""Transform label [0, 2 ** bits - 1] to waveform [-1, 1]
+    """
+
+    return 2 * label / (2 ** bits - 1.0) - 1.0
```
vincentqb, line 98: Please move those functions in a separate file called `functionals.py`.

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,36 @@
+This is an example pipeline for WaveRNN vocoder using the WaveRNN model trained on LJSpeech. WaveRNN model is based on the implementation from [this repository](https://github.com/fatchord/WaveRNN). The original implementation was
+introduced in "Efficient Neural Audio Synthesis". WaveRNN and LJSpeech are available in torchaudio.
```
vincentqb, line 2: As a separate pull request, we could consider adding the plots for training and validation errors in this readme for the pre-trained weight. Either in #776 or another one.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+        self.args = args
+
+    def __getitem__(self, index):
+
+        file = self.files[index]
+        args = self.args
+        n_fft = 2048
+        waveform, sample_rate = torchaudio.load(file)
+        specgram = self.transforms(waveform)
+        # Will be replaced by torchaudio as described in https://github.com/pytorch/audio/pull/593
```
vincentqb, line 47:
```python
        # TODO Replace by torchaudio, once https://github.com/pytorch/audio/pull/593 is resolved
```

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,403 @@
...
+from torchaudio.datasets.utils import bg_iterator
+from torchaudio.models._wavernn import _WaveRNN
+
+from datasets import collate_factory, datasets_ljspeech
+from losses import MoLLoss
+from utils import MetricLogger, count_parameters, save_checkpoint
+
+
+def parse_args():
+    parser = argparse.ArgumentParser()
```
vincentqb, line 24: for later: if you train on another dataset, you could add a command line argument below to select the dataset to experiment with.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+def datasets_ljspeech(args, transforms):
+
+    root = args.file_path
+    wavefiles = [os.path.join(root, file) for file in os.listdir(root)]
+
+    random.seed(args.seed)
+    random.shuffle(wavefiles)
+
+    train_files = wavefiles[: -args.test_samples]
+    test_files = wavefiles[-args.test_samples:]
```
vincentqb, line 74: * I'd say it'd be more natural to have a percentage than a number here, since this would also translate more easily to other dataset. Thoughts?
* I'm curious to hear what others would think about the choice of word to go along "train": "test",  "validation", "development", "evaluation"?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+        return len(self.files)
+
+
+def datasets_ljspeech(args, transforms):
+
+    root = args.file_path
+    wavefiles = [os.path.join(root, file) for file in os.listdir(root)]
+
+    random.seed(args.seed)
+    random.shuffle(wavefiles)
```
vincentqb, line 71: I'd say it'd be more natural to have a percentage than a number here, since this would also translate more easily to other dataset. Thoughts?

examples/pipeline_wavernn/datasets.py
vincentqb, line 74: I'm curious to hear what others would think about the choice of word to go along "train": "test",  "validation", "development", "evaluation"?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+        return waveform, specgram
+
+    def __len__(self):
+        return len(self.files)
+
+
+def datasets_ljspeech(args, transforms):
+
+    root = args.file_path
+    wavefiles = [os.path.join(root, file) for file in os.listdir(root)]
```
vincentqb, line 68: We should organize the code in a way that doesn't require walking the files, but only use the utilities already available in torchaudio. Maybe we need a new one for this :) Thoughts?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+                else waveform_to_label(waveform, args.n_bits)
+            )
+
+        return waveform, specgram
+
+    def __len__(self):
+        return len(self.files)
+
+
+def datasets_ljspeech(args, transforms):
```
vincentqb, line 65: The name of the function should reflect what is happening in it.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,125 @@
...
+    wavefiles = [os.path.join(root, file) for file in os.listdir(root)]
+
+    random.seed(args.seed)
+    random.shuffle(wavefiles)
+
+    train_files = wavefiles[: -args.test_samples]
+    test_files = wavefiles[-args.test_samples:]
+
+    train_dataset = ProcessedLJSPEECH(train_files, transforms, args)
+    test_dataset = ProcessedLJSPEECH(test_files, transforms, args)
```
vincentqb, line 77: One suggestion of a workflow would be
- Introduce a split dataset functionality that randomly assigns a percentage to train/validation/test.
- Redefine `ProcessedLJSpeech` to only apply transforms to `LJSpeech`.
- Use `MapMemoryCache`.

examples/pipeline_wavernn/wavernn.py
vincentqb: did you mean to add this to the help of the command line?

examples/pipeline_wavernn/utils.py
jimchen90: I move them to `functional.py`. I noticed that `functional.py` is used in [torchaudio](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py). Do you think `functionals` is better than `functional`?

examples/pipeline_wavernn/README.md
jimchen90: Good suggestion, I will have a separate pull request to add the plots.

examples/pipeline_wavernn/datasets.py
jimchen90: Good point. I use the argument `test_ratio` instead of `test_files`.

examples/pipeline_wavernn/datasets.py
jimchen90: From a [pytorch example](https://github.com/pytorch/examples/blob/master/imagenet/main.py#L312),  it uses `train` and `validate` as the function name. The names of dataset and dataloader are `train_dataset/loader`, `val_dataset/loader`. So I update the names in this way. Do you have other suggestion?

examples/pipeline_wavernn/datasets.py
jimchen90: Now I use the `_walker` from LJSPEECH from torchaudio to get the files instead of walking the files.

examples/pipeline_wavernn/datasets.py
jimchen90: I use the name `gen_datasets_ljspeech` that means generate datasets from LJSpeech. Do you have other suggestion?

examples/pipeline_wavernn/datasets.py
jimchen90: 1.  I add a function named `split_data` to split the dataset with a `val_ratio`.
2. In `ProcessedLJSpeech`, a `process_datapoint` function is used to only apply transforms to LJSpeech.
3. Use `MapMemoryCache` in `gen_datasets_ljspeech` function.

examples/pipeline_wavernn/datasets.py
jimchen90: The comment has been updated.

examples/pipeline_wavernn/datasets.py
vincentqb: We shouldn't recommend that users access `_walker` since it is a private function as noted by the underscore. How about using a dictionary to map random subset of indices to original indices, and wrapping this into a class? Anyone else has other suggestions? @cpuhrsch @zhangguanheng66

examples/pipeline_wavernn/utils.py
vincentqb: Let's mirror the notation used in torchaudio, so `functional.py`. Also, it may be confusing to use `functionals.py` when we already have `functional.py`.

examples/pipeline_wavernn/datasets.py
vincentqb: I would go for even more explicit like `split_process_ljspeech`.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+class ProcessedLJSPEECH(LJSPEECH):
+    def __init__(self, dataset, transforms, args):
+        self.dataset = dataset
+        self.transforms = transforms
+        self.args = args
+
+    def __getitem__(self, index):
+        filename = self.dataset[index][0]
+        folder = "LJSpeech-1.1/wavs/"
+        file = os.path.join(self.args.file_path, folder, filename + ".wav")
```
vincentqb, line 47: The user should not have to manage paths here.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return self.process_datapoint(file)
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def process_datapoint(self, file):
+        args = self.args
+        n_fft = 2048
+        waveform, sample_rate = torchaudio.load(file)
```
vincentqb, line 57: Isn't dataset loading the file for you?

examples/pipeline_wavernn/transform.py
jimchen90: This part has been removed.

examples/pipeline_wavernn/transform.py
jimchen90: Yes, I have reformatted this.

examples/pipeline_wavernn/wavernn.py
jimchen90: Updated

examples/pipeline_wavernn/wavernn.py
jimchen90: Since this has been mentioned in the WaveRNN model, I remove this comment.

examples/pipeline_wavernn/wavernn.py
jimchen90: This part has been updated.

examples/pipeline_wavernn/wavernn.py
jimchen90: Do you mean:
```
if args.mode == 'waveform':
    criterion = nn.CrossEntropyLoss()
elif args.mode == 'mol':
    criterion = MoLLoss()
else: 
    raise ValueError(f"Expected mode: `waveform` or `mol`, but found {arg.mode}")
```

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+        file = os.path.join(self.args.file_path, folder, filename + ".wav")
+
+        return self.process_datapoint(file)
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def process_datapoint(self, file):
+        args = self.args
+        n_fft = 2048
```
vincentqb, line 56: Why is this hard coded?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+        waveform, sample_rate = torchaudio.load(file)
+        specgram = self.transforms(waveform)
+        # TODO Replace by torchaudio, once https://github.com/pytorch/audio/pull/593 is resolved
+        specgram = linear_to_mel(specgram, sample_rate, n_fft, args.n_freq, args.f_min)
+        specgram = specgram_normalize(specgram, args.min_level_db)
```
vincentqb, line 61: Shouldn't these be part of `self.transforms`?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,142 @@
...
+
+    train_dataset, val_dataset = split_data(data, args.val_ratio, args.seed)
+
+    train_dataset = ProcessedLJSPEECH(train_dataset, transforms, args)
+    val_dataset = ProcessedLJSPEECH(val_dataset, transforms, args)
+
+    train_dataset = MapMemoryCache(train_dataset)
+    val_dataset = MapMemoryCache(val_dataset)
+
+    return train_dataset, val_dataset
```
vincentqb, line 99: Maybe something like this instead? Thoughts? I haven't tested :)

```python
def gen_datasets_ljspeech(
    args, transforms,
):

    data = LJSPEECH(root=args.file_path, download=False)
    train_dataset, val_dataset = split_data(data, args.val_ratio, args.seed)

    train_dataset = Processed(train_dataset, transforms)
    val_dataset = Processed(val_dataset, transforms)

    train_dataset = MapMemoryCache(train_dataset)
    val_dataset = MapMemoryCache(val_dataset)
    return train_dataset, val_dataset

```

```python
class Splitted(torch.utils.data.Dataset):
    def __init__(self, dataset, indices):
        self._dataset = dataset
        self._indices = indices
        super().__init__()

    def __getitem__(self, key):
        return self._dataset[self._indices[key]]

def split_data(data, val_ratio, seed):

    random.seed(seed)
    train_indices = random.sample(range(len(data)), k = int(val_ratio * len(data)))
    val_indices = [i for i in range(len(data)) if i not in train_indices]  # does not need to be shuffled

    return Splitted(data, train_indices), Splitted(data, val_indices)
```

```python
class Processed(torch.utils.data.Dataset):
    def __init__(self, dataset, transforms):
        self._dataset = dataset
        self.transforms = transforms
        super().__init__()

    def __getitem__(self, key):
        item = self._dataset[key]
        return self.process_datapoint(item)

    def process_datapoint(self, item):
        # ...
        return transformed, target

    def __len__(self):
        return len(self._dataset)
```

vincentqb: See also [this](https://discuss.pytorch.org/t/torch-utils-data-dataset-random-split/32209/4)

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,403 @@
...
+    torch.cuda.empty_cache()
+
+    melkwargs = {
+        "n_fft": 2048,
+        "power": 1,
+        "hop_length": args.hop_length,
+        "win_length": args.win_length,
+    }
+
+    transforms = torch.nn.Sequential(torchaudio.transforms.Spectrogram(**melkwargs))
```
vincentqb, line 279: You don't need `Sequential` for one transform.

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,403 @@
...
+
+    devices = ["cuda" if torch.cuda.is_available() else "cpu"]
+
+    logging.info("Start time: {}".format(str(datetime.now())))
+
+    # Empty CUDA cache
+    torch.cuda.empty_cache()
+
+    melkwargs = {
+        "n_fft": 2048,
```
vincentqb, line 273: I see this value hard coded in two places.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,113 @@
...
+        return len(self.dataset)
+
+
+class Processed(torch.utils.data.Dataset):
+    def __init__(self, dataset, transforms):
+        self.dataset = dataset
+        self.transforms = transforms
+
+    def __getitem__(self, key):
+        item = self.dataset[key][0]
```
vincentqb, line 39: nit: let's move the index to `process_datapoint`

examples/pipeline_wavernn/datasets.py
vincentqb: Thanks for updating this :)

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,420 @@
...
+
+    optimizer = Adam(model.parameters(), **optimizer_params)
+
+    criterion = nn.CrossEntropyLoss() if args.mode == "waveform" else MoLLoss
+
+    best_loss = 10.0
+
+    load_checkpoint = args.checkpoint and os.path.isfile(args.checkpoint)
+
+    if load_checkpoint:
```
vincentqb, line 361: since you are not using distributed setup, you can move the condition back in the if statement

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def process_datapoint(self, waveform):
+        specgram = self.transforms(waveform)
+        return waveform.squeeze(0), specgram
+
+
+def gen_datasets_ljspeech(args, transforms):
```
vincentqb, line 50: nit: how about `ljspeech_process_split` or `split_process_ljspeech` as a name?

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,113 @@
...
+        if args.mode == "waveform":
+
+            if args.mulaw:
+                waveform = mulaw_encode(waveform, 2 ** args.n_bits)
+                target = mulaw_encode(target, 2 ** args.n_bits)
+
+                waveform = label_to_waveform(waveform, args.n_bits)
+
+            else:
+                target = waveform_to_label(target, args.n_bits)
```
vincentqb, line 109: nit: should this be part of the transforms sequential?

vincentqb: because of the second mode doesn't transform waveform, let's keep things as they are

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,420 @@
...
+
+
+def main(args):
+
+    devices = ["cuda" if torch.cuda.is_available() else "cpu"]
+
+    logging.info("Start time: {}".format(str(datetime.now())))
+
+    # Empty CUDA cache
+    torch.cuda.empty_cache()
```
vincentqb, line 277: nit: I'm not sure this is necessary

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,420 @@
...
+    melkwargs = {
+        "n_fft": args.n_fft,
+        "power": 1,
+        "hop_length": args.hop_length,
+        "win_length": args.win_length,
+    }
+
+    transforms = torch.nn.Sequential(
+        torchaudio.transforms.Spectrogram(**melkwargs),
+        # TODO Replace by torchaudio, once https://github.com/pytorch/audio/pull/593 is resolved
```
vincentqb, line 288: can you move this comment to the definition of `linear_to_mel`?

examples/pipeline_wavernn/functional.py
```diff
@@ -0,0 +1,30 @@
+import torch
+
+
+def mulaw_encode(waveform, mu):
```
vincentqb, line 4: Is [this](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L510) the same?

examples/pipeline_wavernn/functional.py
```diff
@@ -0,0 +1,30 @@
...
+def waveform_to_label(waveform, bits):
+    r"""Transform waveform [-1, 1] to label [0, 2 ** bits - 1]
+    """
+
+    assert abs(waveform).max() <= 1.0
+    waveform = (waveform + 1.0) * (2 ** bits - 1) / 2
+    return torch.clamp(waveform, 0, 2 ** bits - 1).int()
+
+
+def label_to_waveform(label, bits):
```
vincentqb, line 26: Is this the same as [mulaw decoding](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L537)?

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,35 @@
+import librosa
+import torch
+import torch.nn as nn
+
+
+class linear_to_mel(nn.Module):
```
vincentqb, line 6: nit: for `nn.Module` we use the convention `LinearToMel`

examples/pipeline_wavernn/transform.py
```diff
@@ -0,0 +1,35 @@
...
+            S=specgram.squeeze(0).numpy(),
+            sr=self.sample_rate,
+            n_fft=self.n_fft,
+            n_mels=self.n_mels,
+            fmin=self.fmin,
+        )
+        return torch.from_numpy(specgram)
+
+
+class specgram_normalize(nn.Module):
```
vincentqb, line 25: nit: `NormalizeSpectrogram`

examples/pipeline_wavernn/functional.py
```diff
@@ -0,0 +1,30 @@
+import torch
```
vincentqb, line 1: For the purpose of this example, it may just be simpler to have a single file `processing.py` or `signal_processing.py` or `transforms.py` or `transformations.py` combining both `functional.py` and `transform.py`.

examples/pipeline_wavernn/losses.py
```diff
@@ -0,0 +1,90 @@
...
+    r""" Numerically stable log_sum_exp implementation that prevents overflow
+    """
+
+    axis = len(x.size()) - 1
+    m, _ = torch.max(x, dim=axis)
+    m2, _ = torch.max(x, dim=axis, keepdim=True)
+    return m + torch.log(torch.sum(torch.exp(x - m2), dim=axis))
+
+
+def MoLLoss(y_hat, y, num_classes=65536, log_scale_min=None, reduce=True):
```
vincentqb, line 15: nit: `mol_loss` given that this is not a class

examples/pipeline_wavernn/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    r"""Logger for model metrics
+    """
+
+    def __init__(self, group, print_freq=1):
+        self.print_freq = print_freq
+        self._iter = 0
+        self.data = defaultdict(lambda: deque(maxlen=self.print_freq))
+        self.data["group"].append(group)
+
+    def __call__(self, key, value):
```
vincentqb, line 19: let's replace this by `__setitem__` so that we can do `metric["epoch"] = epoch` instead :)

vincentqb: We could add `__getitem__` as
```python
def __getitem__(self, key):
  return self.data[key][-1]
```
to get the last value. But let's not since we are not using it in the code.

examples/pipeline_wavernn/utils.py
```diff
@@ -0,0 +1,61 @@
...
+        self.data[key].append(value)
+
+    def _get_last(self):
+        return {k: v[-1] for k, v in self.data.items()}
+
+    def __str__(self):
+        return str(self._get_last())
+
+    def print(self):
```
vincentqb, line 28: since we free `__call__`, let's replace this by
```python
def __call__(self):
  ...
```

examples/pipeline_wavernn/utils.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, group, print_freq=1):
+        self.print_freq = print_freq
+        self._iter = 0
+        self.data = defaultdict(lambda: deque(maxlen=self.print_freq))
+        self.data["group"].append(group)
+
+    def __call__(self, key, value):
+        self.data[key].append(value)
+
+    def _get_last(self):
```
vincentqb, line 22: Thinking out loud: Instead of `_get_last` we could add support for `metric_logger[:]` via `__getitem__` to return this. We could then add the snippet I mention [here](https://github.com/pytorch/audio/pull/749/files#r455140233).

vincentqb:
```python
class X:
  def __getitem__(self, key):
    if isinstance(key, slice) and key.start is None and key.stop is None and key.step is None:
      return {k: v[-1] for k, v in self.data.items()}
  return self.data[key][-1]
```
this means we could do:
```python
    def __str__(self):
        return str(self[:])
```

vincentqb: however, this returns a dictionary, not the list of values. maybe this wouldn't be natural :)

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,420 @@
...
+            )
+
+    logging.info(f"End time: {datetime.now()}")
+
+
+if __name__ == "__main__":
+
+    logging.basicConfig(level=logging.INFO)
+    args = parse_args()
+    main(args)
```
vincentqb, line 420: you could add a final function after `main` that takes the save the final weights as you have done [here](https://github.com/jimchen90/TestRep/blob/master/wavernn/load_weights.ipynb).

vincentqb: let's not do this in the example pipeline to keep it minimal

examples/pipeline_wavernn/functional.py
jimchen90: I test them in notebook and they are same. I replaced it.

examples/pipeline_wavernn/datasets.py
jimchen90: I prefer the second one. Updated.

examples/pipeline_wavernn/datasets.py
jimchen90: Updated.

examples/pipeline_wavernn/main.py
jimchen90: I removed this part since I didn't test it specifically.

examples/pipeline_wavernn/functional.py
jimchen90: It is different. It is the inverse operation of `waveform_to_label`.

examples/pipeline_wavernn/functional.py
jimchen90: Good suggestion. I have combined them in `processing.py`

examples/pipeline_wavernn/losses.py
jimchen90: Yes, I have updated it.

examples/pipeline_wavernn/main.py
jimchen90: Do you mean:
`if args.checkpoint and os.path.isfile(args.checkpoint):`
instead of:
`load_checkpoint = args.checkpoint and os.path.isfile(args.checkpoint)`
`if load_checkpoint:`

examples/pipeline_wavernn/main.py
jimchen90: Yes, i moved it.

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    logging.info(f"Number of parameters: {n}")
+
+    # Optimizer
+    optimizer_params = {
+        "lr": args.learning_rate,
+    }
+
+    optimizer = Adam(model.parameters(), **optimizer_params)
+
+    criterion = nn.CrossEntropyLoss() if args.mode == "waveform" else mol_loss
```
vincentqb, line 348: Given this syntax, let's make `MoLLoss` an `nn.Module` like `CrossEntropyLoss`

examples/pipeline_wavernn/processing.py
```diff
@@ -0,0 +1,52 @@
...
+def waveform_to_label(waveform, bits):
+    r"""Transform waveform [-1, 1] to label [0, 2 ** bits - 1]
+    """
+
+    assert abs(waveform).max() <= 1.0
+    waveform = (waveform + 1.0) * (2 ** bits - 1) / 2
+    return torch.clamp(waveform, 0, 2 ** bits - 1).int()
+
+
+def label_to_waveform(label, bits):
```
vincentqb, line 48: same

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    )
+    parser.add_argument(
+        "--n-bits", default=8, type=int, help="the bits of output waveform",
+    )
+    parser.add_argument(
+        "--sample-rate",
+        default=22050,
+        type=int,
+        help="the rate of audio dimensions (samples per second)",
+    )
```
vincentqb, line 88: The dataset knows the sample rate, so is there a reason why you are not leveraging this?

examples/pipeline_wavernn/processing.py
```diff
@@ -0,0 +1,52 @@
...
+    def __init__(self, min_level_db):
+        super().__init__()
+        self.min_level_db = min_level_db
+
+    def forward(self, specgram):
+        specgram = 20 * torch.log10(torch.clamp(specgram, min=1e-5))
+        return torch.clamp((self.min_level_db - specgram) / self.min_level_db, min=0, max=1)
+
+
+def waveform_to_label(waveform, bits):
```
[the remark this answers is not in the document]
jimchen90: I changed the function to `encode_waveform_into_bits`.

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,36 @@
+This is an example pipeline for WaveRNN vocoder using the WaveRNN model trained on LJSpeech. WaveRNN model is based on the implementation from [this repository](https://github.com/fatchord/WaveRNN). The original implementation was
```
vincentqb, line 1: nit: 
> This is an example vocoder pipeline using the WaveRNN model trained with LJSpeech.

examples/pipeline_wavernn/README.md
```diff
@@ -0,0 +1,36 @@
...
+    --batch-size 256 \
+    --learning-rate 1e-4 \
+    --n-freq 80 \
+    --mode 'waveform' \
+    --n-bits 8 \
+```
+
+### Output
+
+The information reported at each iteration and epoch (e.g. loss) is printed to standard output in the form of one json per line. Here is an example python function to parse the standard output.
```
vincentqb, line 18: nit:
> ... Here is an example python function to parse the output if redirected to a file.

examples/pipeline_wavernn/losses.py
```diff
@@ -0,0 +1,98 @@
...
+        log_probs = cond * log_cdf_plus + (1.0 - cond) * inner_out
+
+        log_probs = log_probs + F.log_softmax(logit_probs, -1)
+
+        if self.reduce:
+            return -torch.mean(self.log_sum_exp(log_probs))
+        else:
+            return -self.log_sum_exp(log_probs).unsqueeze(-1)
+
+    def log_sum_exp(self, x):
```
vincentqb, line 91: Since this does not depend on `self`, was there a reason why this is a method instead of a staticmethod? In any case, let's move this to a private function `_ log_sum_exp` outside of the class for readability.

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    parser.add_argument("--clip-grad", metavar="NORM", type=float, default=4.0)
+    parser.add_argument(
+        "--mulaw",
+        default=True,
+        action="store_true",
+        help="if used, waveform is mulaw encoded",
+    )
+    parser.add_argument(
+        "--jit", default=False, action="store_true", help="if used, model is jitted"
+    )
```
vincentqb, line 73: Have you tested your code with jit?

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    parser.add_argument(
+        "--kernel-size",
+        default=5,
+        type=int,
+        help="the number of kernel size in the first Conv1d layer",
+    )
+    parser.add_argument(
+        "--n-freq", default=80, type=int, help="the number of bins in a spectrogram",
+    )
```
vincentqb, line 130: nit:
> the number of spectrogram bins to use

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    parser.add_argument(
+        "--n-hidden", default=128, type=int, help="the number of hidden dimensions",
+    )
+    parser.add_argument(
+        "--n-output", default=128, type=int, help="the number of output dimensions",
+    )
```
vincentqb, line 136: output to what?

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    parser.add_argument(
+        "--n-fft", default=2048, type=int, help="the number of Fourier bins",
+    )
+    parser.add_argument(
+        "--mode",
+        default="waveform",
+        choices=["waveform", "mol"],
+        type=str,
+        help="the mode of waveform",
```
vincentqb, line 145: Having waveform in the help may be confusing since `waveform` is one of the option. Please rephrase.

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+    )
+    parser.add_argument(
+        "--seq-len-factor",
+        default=5,
+        type=int,
+        help="seq_length = hop_length * seq_len_factor",
```
vincentqb, line 151: This is a neat description, but `seq_length` is not defined. Please add its definition here.

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+        metric["iteration"] = sums["iteration"]
+        metric["time"] = time() - start2
+        metric()
+        sums["iteration"] += 1
+
+    avg_loss = sums["loss"] / len(data_loader)
+
+    metric = MetricLogger("train_epoch")
+    metric["epoch"] = epoch
+    metric["loss"] = avg_loss
```
vincentqb, line 222: nit: you can remove `avg_loss` and replace this by
```
    metric["loss"] = sums["loss"] / len(data_loader)
```

vincentqb: ... or we can just mimic "evaluate" and leave this as it is.

examples/pipeline_wavernn/main.py
vincentqb: This essentially decides what loss function to use, right? Either `CrossEntropyLoss` or `discretized mixture of logistic distributions loss`. This would be clearer as a setting.

vincentqb: How about calling this `--loss`, and then having choices: `CrossEntropy` and `MoL` instead?

vincentqb:
> There are at least a few more things to do:
> 
> * [x]  Add bg_iterator and README.
> * [ ]  Add torchaudio transforms on mel-spectrogram.

What is the second item?

examples/pipeline_wavernn/main.py
jimchen90: Yes, that is neat. I changed it.

examples/pipeline_wavernn/main.py
jimchen90: I like the `--loss` name. But `mode` is also a parameter of wavernn model. In wavernn model, each mode has a different output channel (`n_classes = 2 ** n_bits` or `n_classes = 30`).

examples/pipeline_wavernn/main.py
jimchen90: I rephrase it as `the output dimension of upsample network in WaveRNN model`

examples/pipeline_wavernn/main.py
jimchen90: Updated. Thanks

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,411 @@
...
+        default=1100,
+        type=int,
+        help="the number of samples between the starts of consecutive frames",
+    )
+    parser.add_argument(
+        "--f-min",
+        default=40.0,
+        type=float,
+        help="the number of samples between the starts of consecutive frames",
+    )
```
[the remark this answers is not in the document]
jimchen90: since `f-min` is used in torchaudio [transforms](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L214). I prefer this name as the variable name in the transforms. Right?

examples/pipeline_wavernn/main.py
jimchen90: No, I haven't test it.

examples/pipeline_wavernn/losses.py
jimchen90: That is good point. I moved it.

examples/pipeline_wavernn/README.md
jimchen90: Updated, thanks.

examples/pipeline_wavernn/README.md
jimchen90: Updated, thanks.

examples/pipeline_wavernn/main.py
jimchen90: I use the `sample_rate` here because it is also an input variable in WaveRNN model and the `LinearToMel` class. Since `sample_rate` is not needed in the model. I plan to open a separate pull request to remove the `sample_rate` in the model and in the pipeline. Suggestion?

jimchen90:
> > There are at least a few more things to do:
> > 
> > * [x]   Add bg_iterator and README.
> > * [ ]   Add torchaudio transforms on mel-spectrogram.
> 
> What is the second item?

This is to replaced the LinearToMel function. I remove it since we don't need it now.

examples/pipeline_wavernn/main.py
jimchen90: I agree. I will rephrased it as 
`the length factor of input waveform, the length of input waveform sep_length = hop_length * seq_len_factor`

vincentqb: Let's simply and make more precise:
> the length of each waveform to process per batch = hop_length * seq_len_factor

examples/pipeline_wavernn/main.py
vincentqb: Let's change to `--loss` and then please open a separate pull request to address [comment](https://github.com/pytorch/audio/pull/735/files#r456477181) beforehand.

examples/pipeline_wavernn/main.py
vincentqb: Let's also change the parameter to `--n-output-upsample`?

cc [comment](https://github.com/pytorch/audio/pull/735/files#r456481205)

examples/pipeline_wavernn/main.py
```diff
@@ -0,0 +1,402 @@
...
+        "--kernel-size",
+        default=5,
+        type=int,
+        help="the number of kernel size in the first Conv1d layer",
+    )
+    parser.add_argument(
+        "--n-freq", default=80, type=int, help="the number of spectrogram bins to use",
+    )
+    parser.add_argument(
+        "--n-hidden", default=128, type=int, help="the number of hidden dimensions",
```
vincentqb, line 123: cc [comment](https://github.com/pytorch/audio/pull/735/files#r456481205)

examples/pipeline_wavernn/main.py
vincentqb: Based on offline discussion, it looks like you've ran the code with jit, and so I would be ok leaving it in. I would in fact like that you leave it in to make sure the model is actually jitable.

examples/pipeline_wavernn/losses.py
vincentqb: Can you add the underscore? This is not a loss function so we do not want to expose it directly in the file.

examples/pipeline_wavernn/main.py
vincentqb: Sounds good, but in this case, let's not add a command line argument to the model for this. Can you just hard code `sample_rate=0` when creating the model?

Please open a pull request to remove `sample_rate` from the model.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,115 @@
...
+    def __len__(self):
+        return len(self.dataset)
+
+    def process_datapoint(self, waveform):
+        specgram = self.transforms(waveform[0])
+        return waveform[0].squeeze(0), specgram
+
+
+def split_process_ljspeech(args, transforms):
+    data = LJSPEECH(root=args.file_path, download=False)
```
vincentqb, line 52: You can get the `sample_rate` from here by simply looking at `data[0][1]` (please double check the shape of LJSpeech data point here). You can then output the sample_rate from the function too.

examples/pipeline_wavernn/datasets.py
```diff
@@ -0,0 +1,115 @@
...
+        self.transforms = transforms
+
+    def __getitem__(self, key):
+        item = self.dataset[key]
+        return self.process_datapoint(item)
+
+    def __len__(self):
+        return len(self.dataset)
+
+    def process_datapoint(self, waveform):
```
vincentqb, line 46: nit: change variable name `waveform` to `item`

examples/pipeline_wavernn/datasets.py
vincentqb: The issue though is that, as you pointed out, we need the `sample_rate` to defined `LinearToMel` transform.

vincentqb: Alright, let's leave the `sample_rate` argument in :)

examples/pipeline_wavernn/processing.py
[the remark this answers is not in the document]
vincentqb: Let's use `normalized_waveform_to_bits` as you suggested offline.

examples/pipeline_wavernn/losses.py
```diff
@@ -0,0 +1,101 @@
...
+    def __init__(self, num_classes=65536, log_scale_min=None, reduce=True):
+        super(MoLLoss, self).__init__()
+        self.num_classes = num_classes
+        self.log_scale_min = log_scale_min
+        self.reduce = reduce
+
+    def forward(self, y_hat, y):
+
+        if self.log_scale_min is None:
+            self.log_scale_min = torch.log(torch.as_tensor(1e-14)).item()
```
vincentqb, line 33: nit: this doesn't really need torch. Would `math.log(1e-14)` do the same?

examples/pipeline_wavernn/losses.py
```diff
@@ -0,0 +1,101 @@
...
+        mid_in = inv_stdv * centered_y
+        # log probability in the center of the bin, to be used in extreme cases
+        log_pdf_mid = mid_in - log_scales - 2.0 * F.softplus(mid_in)
+
+        inner_inner_cond = (cdf_delta > 1e-5).float()
+
+        inner_inner_out = inner_inner_cond * torch.log(
+            torch.clamp(cdf_delta, min=1e-12)
+        ) + (1.0 - inner_inner_cond) * (
+            log_pdf_mid - torch.log(torch.as_tensor((self.num_classes - 1) / 2)).item()
```
vincentqb, line 77: same: this does not need torch

examples/pipeline_wavernn/datasets.py
jimchen90: Yes, it has been changed.

examples/pipeline_wavernn/main.py
jimchen90: It has been changed to:
```
parser.add_argument(
        "--n-hidden-resblock", default=128, type=int, help="the number of hidden dimensions of resblock",
    )
```

examples/pipeline_wavernn/processing.py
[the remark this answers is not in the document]
jimchen90: It has been changed.

examples/pipeline_wavernn/losses.py
jimchen90: yes, I tested in notebook and replaced it.

examples/pipeline_wavernn/main.py
vincentqb: or rename to something else, appropriate with the description

examples/pipeline_wavernn/main.py
jimchen90: It has been updated and a separate PR is #797 .

examples/pipeline_wavernn/main.py
jimchen90: With the [comment](https://github.com/pytorch/audio/pull/797#discussion_r456626653), I changed its description.

examples/pipeline_wavernn/main.py
jimchen90: I noticed that I had the `jit` code in previous commits. The code rans correctly but the default value of `jit` is false. So I didn't run the `jit` test thus I remove this part.

examples/pipeline_wavernn/losses.py
jimchen90: Yes, the underscore has been added.

examples/pipeline_wavernn/losses.py
jimchen90: it has been replaced.

examples/pipeline_wavernn/datasets.py
jimchen90: This function has been added as `normalized_waveform_to_bits` function in `processing.py`.

examples/pipeline_wavernn/main.py
vincentqb: Let's change `--n-output-melresnet` and `--n-hidden-melresnet` with `the number of hidden dimensions of resblock in melresnet`

Issue #674, opened 2020-06-01
Torchaudio Compilation Error


jimchen90:
## 🐛 Bug

When trying to compile torchaudio from source after  `python setup.py install` within anaconda,  it fails with an error that `configure: error: cannot find flac`  (see below for full information).

## To Reproduce

1. Compile PyTorch from source in conda
2. Compile Torchaudio from source in conda

See code as below:

```
module load cuda/10.1
conda create -n audio-master python=3.7
conda activate audio-master
# install cuda with needed version
conda install -c pytorch magma-cuda101

conda install numpy ninja pyyaml mkl mkl-include setuptools cmake cffi typing
git clone --recursive https://github.com/pytorch/pytorch
cd pytorch
git submodule sync
git submodule update --init --recursive
git submodule sync
git submodule update --init --recursive

python setup.py clean --all
export CMAKE_PREFIX_PATH=${CONDA_PREFIX:-"$(dirname $(which conda))/../"}
python setup.py install

cd ..
python -c "import torch; print(torch.__version__); print(torch.cuda.is_available());"

conda install -c conda-forge sox
git clone https://github.com/pytorch/audio
cd audio

python setup.py clean --all              
python setup.py install

cd ..
python -c "import torchaudio;"
```
Error message:

After python `setup.py install `for torchaudio,  it shows:

```
Building sox
configure: error: in `/private/home/jimchen90/audio/third_party/tmp/sox-14.4.2':
configure: error: cannot find flac
See `config.log' for more details
subprocess.CalledProcessError: Command '['/private/home/jimchen90/audio/build_tools/setup_helpers/build_third_party.sh']' returned non-zero exit status 1.
```

## Expected behavior

Torchaudio can be imported outside the source folder.

## Environment

* PyTorch Version: 1.6.0a0+78acc9d
* OS: Linux
* How you installed PyTorch: Conda
* Build command you used : As described above
* Python version: 3.7
* CUDA/cuDNN version: CUDA 10.1

## Additional context

With the help of Moto @mthrok , torchaudio can be imported after adding an external sox file in the audio directory by:

```
conda uninstall sox
rm -rf third_party
touch .use_external_sox
python setup.py clean develop
```
The result of `python -m torch.utils.collect_env` 
```
Collecting environment information...
PyTorch version: 1.6.0a0+78acc9d
Is debug build: No
CUDA used to build PyTorch: 10.1

OS: Ubuntu 18.04.3 LTS
[...]

PR #750, opened 2020-06-26
Add ogg/vorbis support to binary distribution

mthrok:
This PR adds ogg/vorbis support to binary build.

@seemethere Can we slip this into the next release?

Solves #674


vincentqb:
> This PR adds ogg/vorbis support to binary build.

Should we have a test for this?

> @seemethere Can we slip this into the next release?

As far as I know, the release is still a few weeks away :)

mthrok:
> Should we have a test for this?
> 

Test is already there.

vincentqb:
> > Should we have a test for this?
> 
> Test is already there.

[Right](https://github.com/pytorch/audio/blob/master/test/sox_io_backend/test_load.py#L100) :)

mthrok:
Thanks!

PR #753, opened 2020-06-30
Use cmake for third party

mthrok:
This PR replaces the custom bash script for building codecs and libsox with CMake.
CMake gives better caching, secure fetching with much smaller numbers of lines of code.


mthrok:
> This doesn't mean we have to build with CUDA correct?
> 

Nope, this has nothing to do with CUDA, and `torchaudio` does not have any custom CUDA code.

mthrok:
Thanks!

PR #758, opened 2020-07-01
Use complex tensors in phase_vocoder

anjali411:
This PR updates the `phase_vocoder` to use complex tensors. A local boolean flag `USE_COMPLEX` is added in `torch.phase_vocoder` to detect if the input is complex or not. If the input is complex, tensors with complex dtypes will be used in the implementation as well as a complex dtype tensor will be returned. 

Test Plan: It adds JIT tests and new tests for complex dtype in separate files.

Documentation: reflects both the old API behavior which used `(..., complex=2)` dimension real tensors and new API behavior which uses complex dtype tensors.

Deprecation Warnings will be added at the end (before the release) when all functions support complex tensors.

In a follow up PR, [`torch.polar`](https://github.com/pytorch/pytorch/pull/39617) should be added to construct complex tensors using abs and angle in `torch.functional.phase_vocoder`.


torchaudio/functional.py
```diff
@@ -532,72 +531,70 @@ def phase_vocoder(
...
-    complex_specgrams_stretch = torch.stack([real_stretch, imag_stretch], dim=-1)
+    complex_specgrams_stretch = torch.view_as_complex(torch.stack([real_stretch, imag_stretch], dim=-1))
 
     # unpack batch
-    complex_specgrams_stretch = complex_specgrams_stretch.reshape(shape[:-3] + complex_specgrams_stretch.shape[1:])
+    complex_specgrams_stretch = complex_specgrams_stretch.reshape(shape[:-2] + complex_specgrams_stretch.shape[1:])
 
     return complex_specgrams_stretch
 
-
```
vincentqb, line 600: nit: lint

torchaudio/functional.py
```diff
@@ -532,72 +531,70 @@ def phase_vocoder(
...
     Returns:
-        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate), complex=2)`
+        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate))`
 
     Example
         >>> freq, hop_length = 1025, 512
-        >>> # (channel, freq, time, complex=2)
-        >>> complex_specgrams = torch.randn(2, freq, 300, 2)
+        >>> # (channel, freq, time)
+        >>> complex_specgrams = torch.randn(2, freq, 300, dtype=torch.cfloat)
```
vincentqb, line 544: this is neat!

torchaudio/functional.py
```diff
@@ -522,7 +522,6 @@ def magphase(
     phase = angle(complex_tensor)
     return mag, phase
 
-
```
vincentqb, line 525: nit: lint

vincentqb:
Are you planning to change all the implementations of complex numbers in torchaudio? We could add this to the release if it's ready and all the references are changed. Anything to be aware of? In particular, we are not testing autograd, and we have not made any commitments yet. Would this support autograd?

anjali411:
> Are you planning to change all the implementations of complex numbers in torchaudio? We could add this to the release if it's ready and all the references are changed. Anything to be aware of? In particular, we are not testing autograd, and we have not made any commitments yet. Would this support autograd?

synced offline:
1. We don't plan to include these changes in the upcoming release.
2. We should return a complex tensor when the input is a complex tensor, and a real tensor when input is a real tensor
3. There should be a deprecation warning if the user inputs a real tensor (..., 2)
4. Ensure jit consistency tests run fine.
5. Add tests for autograd?

anjali411:
@vincentqb @mthrok The CI for this PR looks green (not sure how the `codecov` report is generated). 

Now on further look at the `torchaudio` repo, it seems like almost all functions in `functional.py` will have to be modified to have them correctly work with complex numbers. These are some possible reasons why the functions would have to be modified:

1. The function expects as input or returns a real representation (..., 2) instead of a complex tensor. In this case, the implementation will also have to be modified accordingly.
2. The function doesn't have any (..., 2) assumptions but uses functions like `torch.clamp` whose behavior is not intuitively defined for complex numbers and is currently disable in torch.
3. The functions are using internally using workaround functions for `angle`, `norm` etc.

Now the functions in `functional.py` are called at different places in the code and so there would a lot more sites where modifications might have to be made.

_TorchScript_
1. 
2. `TorchScript consistency tests` first explicitly convert the dtype of the tensors to float/double, but this behavior will no longer be correct to test since after above modifications some functions will expect as input complex tensors.

_Question for you:_ 

Should we add a warning in this release that `torchaudio` will switch to complex tensors in the next release so that we don't have to add these `if-else` blocks at every site:
```
if input.is_complex():
    do ...
else:
    do ...
```

_Things to figure out:_

Autograd requirements for torchaudio to be able to adapt complex numbers.

vincentqb:
> 2. The function doesn't have any (..., 2) assumptions but uses functions like `torch.clamp` whose behavior is not intuitively defined for complex numbers and is currently disable in torch.

How about truncating the norm of a complex number in polar coordinate? i.e. (1) convert to polar coordinate, (2) clamp the norm, (3) convert back to cartesian coordinate. This has the advantage of reducing to clamp [-x, x] for real numbers? I guess this wouldn't work as well for asymmetric interval clamp. Other cases to thing about?

torchaudio/functional.py
```diff
@@ -532,68 +532,67 @@ def phase_vocoder(
...
+
     phase_acc = torch.cumsum(phase, -1)
 
     mag = alphas * norm_1 + (1 - alphas) * norm_0
 
     real_stretch = mag * torch.cos(phase_acc)
     imag_stretch = mag * torch.sin(phase_acc)
 
-    complex_specgrams_stretch = torch.stack([real_stretch, imag_stretch], dim=-1)
+    complex_specgrams_stretch = torch.view_as_complex(torch.stack([real_stretch, imag_stretch], dim=-1))
```
gchanan, line 592: this seems pretty clunky, is there not a better way of doing it?

anjali411: agreed! After this PR: https://github.com/pytorch/pytorch/pull/39617, we can just rewrite this as `torch.complex_polar(mag, phase_acc)` instead of:

```
real_stretch = mag * torch.cos(phase_acc)
imag_stretch = mag * torch.sin(phase_acc)
complex_specgrams_stretch = torch.view_as_complex(torch.stack([real_stretch, imag_stretch], dim=-1))
```

torchaudio/functional.py
```diff
@@ -458,68 +458,67 @@ def phase_vocoder(
     factor of ``rate``.
 
     Args:
-        complex_specgrams (Tensor): Dimension of `(..., freq, time, complex=2)`
```
vincentqb, line 461: How about something like this?
```
        complex_specgrams (Tensor): Either a real tensor of dimension of `(..., freq, time, complex=2)`
            or a tensor of dimension `(..., freq, time)` with complex dtype.
```
We were using "complex tensor" to mean `(..., complex=2)`. This is now ambiguous. What expression do you recommend to refer to a tensor of complex dtype? "tensor with a complex dtype"?

torchaudio/functional.py
```diff
@@ -458,68 +458,67 @@ def phase_vocoder(
...
+        complex_specgrams (Tensor): Dimension of `(..., freq, time)`
         rate (float): Speed-up factor
         phase_advance (Tensor): Expected phase advance in each bin. Dimension of (freq, 1)
 
     Returns:
-        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate), complex=2)`
+        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate))`
```
vincentqb, line 466: based on the comment above, this would become:
```
        Tensor: Complex Specgrams Stretch, represented either as a real tensor with dimension of
            `(..., freq, ceil(time/rate), complex=2)` or a tensor of dimension `(..., freq, time)` with complex dtype.
```
thoughts?

torchaudio/functional.py
```diff
@@ -458,68 +458,67 @@ def phase_vocoder(
...
-        >>> complex_specgrams = torch.randn(2, freq, 300, 2)
+        >>> # (channel, freq, time)
+        >>> complex_specgrams = torch.randn(2, freq, 300, dtype=torch.cfloat)
         >>> rate = 1.3 # Speed up by 30%
         >>> phase_advance = torch.linspace(
         >>>    0, math.pi * hop_length, freq)[..., None]
         >>> x = phase_vocoder(complex_specgrams, rate, phase_advance)
         >>> x.shape # with 231 == ceil(300 / 1.3)
-        torch.Size([2, 1025, 231, 2])
+        torch.Size([2, 1025, 231])
```
vincentqb, line 477: the example doesn't need to be changed. we could add a second example, or a comment next to each `, 2`. other ideas?

torchaudio/functional.py
anjali411: yeah tensor with a complex dtype sounds good. However this "or" way of documenting could be problematic in case where the function takes more than one complex tensors. Perhaps in those cases, we can add a note. 

I think it might be nicer to add a separate example with complex dtype tensors so that it's also clear that the returned output would also be complex (if applicable) etc., especially since we are planning to switch to using complex dtype tensors in the release after the upcoming release

vincentqb:
For deprecations to switch between (real, 2) and complex-dtype, the options are:

1. Detect dtype automatically, and warn if user uses real tensors (and maybe have a local keyword argument to suppress the warning).
1. Have a local keyword argument to switch the behavior, with default None that raises a warning.
1. Have a global flag to switch the behavior, with default None that raises a warning.
1. Have separate functionals/transforms with complex dtype.

I'm leaning for 1 without a warning until we are ready for deprecation. At that point, we'll just tell the user that to suppress the warning they need to convert to complex dtype. Thoughts?

vincentqb:
For testing, you can copy existing test into a new test file prefixed with `complex_` to make it clear the test is almost the same but for complex.

anjali411:
> For deprecations to switch between (real, 2) and complex-dtype, the options are:
> 
> 1. Detect dtype automatically, and warn if user uses real tensors (and maybe have a local keyword argument to suppress the warning).
> 2. Have a local keyword argument to switch the behavior, with default None that raises a warning.
> 3. Have a global flag to switch the behavior, with default None that raises a warning.
> 4. Have separate functionals/transforms with complex dtype.
> 
> I'm leaning for 1 without a warning until we are ready for deprecation. At that point, we'll just tell the user that to suppress the warning they need to convert to complex dtype. Thoughts?

Yeah I would also agree that having local flags for each function makes sense since there will be jit issues with a global flag. And once fully migrated, we should generate warning everytime a user uses the (..., complex=2) real tensor.

torchaudio/functional.py
anjali411: I think we should have an example with tensors of complex dtype so that users know how to deal with complex tensors. This is an option:
```
(Old API) Example:
....

(New API) Example:
...
``` what do you think?

File: test/torchaudio_unittest/torchscript_consistency_impl.py [hunk 8]
```diff
@@ -529,6 +529,28 @@ def func(tensor):
         self._assert_consistency(func, waveform)
 
 
+class TransformsWithComplexDtypes(common_utils.TestBaseMixin):
```
mthrok, line 532: This class structure looks exactly same as `Transforms` class bellow. Any reason you needed this to be a separate test class?

torchaudio/functional.py
```diff
@@ -458,14 +458,17 @@ def phase_vocoder(
...
         phase_advance (Tensor): Expected phase advance in each bin. Dimension of (freq, 1)
 
     Returns:
-        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate), complex=2)`
+        Tensor: Complex Specgrams Stretch with either a real dtype and dimension of
+                `(..., freq, ceil(time/rate), complex=2)` or
+                a complex dtype and dimension of `(..., freq, ceil(time/rate))`.
 
-    Example
+    (Old API) Example
```
mthrok, line 471: `Example` block gets a special treatment in documentation, so adding the differentiation after `Example`, like `Example - foo` gives better rendering.

Screen Shot 2020-08-06 at 16 05 08

vs 

Screen Shot 2020-08-06 at 16 05 17

anjali411: that's a good point! updated

test/torchaudio_unittest/torchscript_consistency_impl.py
anjali411: yeah because the tests in this class would only run for complex dtypes (call site https://github.com/pytorch/audio/pull/758/files#diff-b4e9b4cf7b36e86d0bc844f4dd0cfba4R32) whereas tests in the class below run for real dtypes.

test/torchaudio_unittest/complex_librosa_compatibility_test.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # == Test shape
+    expected_size = list(complex_specgrams.size())
+    expected_size[-1] = int(np.ceil(expected_size[-1] / rate))
+
+    assert complex_specgrams.dim() == complex_specgrams_stretch.dim()
+    assert complex_specgrams_stretch.size() == torch.Size(expected_size)
+
+    # == Test values
+    index = [0] * (complex_specgrams.dim() - 2) + [slice(None)] * 2
```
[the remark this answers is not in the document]
anjali411: yeah fair! updated

torchaudio/functional.py
```diff
@@ -475,51 +478,83 @@ def phase_vocoder(
...
+        >>> # (channel, freq, time)
+        >>> complex_specgrams = torch.randn(2, freq, 300, dtype=torch.cfloat)
+        >>> rate = 1.3 # Speed up by 30%
+        >>> phase_advance = torch.linspace(
+        >>>    0, math.pi * hop_length, freq)[..., None]
+        >>> x = phase_vocoder(complex_specgrams, rate, phase_advance)
+        >>> x.shape # with 231 == ceil(300 / 1.3)
+        torch.Size([2, 1025, 231])
+    """
+    USE_COMPLEX = complex_specgrams.is_complex()
```
ezyang, line 493: I probably wouldn't actually all caps this, typically that means you have a global constant but that's not what this variable is.

anjali411: yeah makes sense! updated

test/torchaudio_unittest/torchscript_consistency_impl.py
```diff
@@ -585,7 +607,7 @@ def test_TimeStretch(self):
         n_freq = 400
         hop_length = 512
         fixed_rate = 1.3
-        tensor = torch.rand((10, 2, n_freq, 10, 2))
+        tensor = torch.rand((10, 2, n_freq, 10, 2), dtype=torch.double)
```
mthrok, line 610: Is this change related or necessary?
My understanding is that `_assert_consistency` method will change the dtype/device to appropriate ones so this change has no effect.

test/torchaudio_unittest/torchscript_consistency_impl.py
mthrok: I see. In that case, can we extract a class with `_assert_consistency` method, so that there is no duplicated method between `Transform` class and `TransformsWithComplexDtypes` ?

```python
class TransformsMixin
    def _assert_consistency(self, transform, tensor):
        ...

class TransformsWithComplexDtypes(TransformsMixin, common_utils.TestBaseMixin):
    ...

class Transforms(TransformsMixin, common_utils.TestBaseMixin):
    ...
```

test/torchaudio_unittest/complex_batch_consistency_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+from torchaudio_unittest import common_utils
+
+
+class TestTransforms(common_utils.TorchaudioTestCase):
+    backend = 'default'
+
+    def test_batch_TimeStretch(self):
+        test_filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
+        waveform, _ = common_utils.load_wav(test_filepath)  # (2, 278756), 44100
```
mthrok, line 18: BTW does the content of the audio matter here?
We are trying to avoid external audio files in new tests.
We have some utilities for generating synthetic data outlined [here](https://github.com/pytorch/audio/tree/master/test/torchaudio_unittest#waveforms-for-testing-purposes).
Sorry I should have thought of this in the first place.

mthrok: Also, if you decided to use synthetic data generated on-the-fly, please remove `backend='default'` above, so that torch audio's io function is used accidentally.

test/torchaudio_unittest/complex_batch_consistency_test.py
```diff
@@ -0,0 +1,49 @@
...
+
+import torch
+import torchaudio
+import torchaudio.functional as F
+
+from torchaudio_unittest import common_utils
+
+
+class TestTransforms(common_utils.TorchaudioTestCase):
+    backend = 'default'
```
mthrok, line 14: Since you got rid of `torchaudio.load`, can you remove this too?

torchaudio/functional.py
vincentqb: I do find this "or" way of discussing this a little cumbersome, and I agree this will get long if many tensors are involved. We could add a note in each, and just define the args/returns with complex dtype. We can still keep the example for clarity.

```python
"""
    We are migrating to complex-dtype tensors. For backward compatibility reason,
    this function still supports the legacy convention of ending with a dimension of 2
    to represent a complex tensor.

    Args:
        complex_specgrams (Tensor): A tensor of dimension `(..., freq, time)` with complex dtype.
        rate (float): Speed-up factor
        phase_advance (Tensor): Expected phase advance in each bin. Dimension of (freq, 1)
    Returns:
        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate))`
            with a complex dtype.

    Example

    Example - Legacy
"""
```
Thoughts?

torchaudio/functional.py
vincentqb: Good suggestion :)

torchaudio/functional.py
vincentqb: P.S. Good suggestion [below](https://github.com/pytorch/audio/pull/758/files#r466610938) for example naming

mthrok:
> For testing, you can copy existing test into a new test file prefixed with `complex_` to make it clear the test is almost the same but for complex.

@vincentqb @anjali411 

I do not see an advantage or necessity to put these test suites in a separate files, unless these complex types are not supported in fbcode (which is). Especially, because this is not a new function, module or test category. Yet the test suite class is a new due to the different type so putting them in the same existing class files make more sense.

anjali411:
> > For testing, you can copy existing test into a new test file prefixed with `complex_` to make it clear the test is almost the same but for complex.
> 
> @vincentqb @anjali411
> 
> I do not see an advantage or necessity to put these test suites in a separate files, unless these complex types are not supported in fbcode (which is). Especially, because this is not a new function, module or test category. Yet the test suite class is a new due to the different type, which is good enough, so putting them in the same existing class files make more sense.

I am fine either way. We can also add a new class in the same file! @mthrok will the tests in the newly added files not run in fbcode?

mthrok:
> I am fine either way. We can also add a new class in the same file! @mthrok will the tests in the newly added files not run in fbcode?

@anjali411 They do, if we add the definition for the new file. But to me, it makes sense to separate them if fbcode cannot run complex type (but I bet it can because pytorch's always the latest version). Adding the tests to existing file would just start running these tests automatically.

torchaudio/functional.py
```diff
@@ -458,14 +458,17 @@ def phase_vocoder(
...
         phase_advance (Tensor): Expected phase advance in each bin. Dimension of (freq, 1)
 
     Returns:
-        Tensor: Complex Specgrams Stretch with dimension of `(..., freq, ceil(time/rate), complex=2)`
+        Tensor: Complex Specgrams Stretch with either a real dtype and dimension of
+                `(..., freq, ceil(time/rate), complex=2)` or
+                a complex dtype and dimension of `(..., freq, ceil(time/rate))`.
 
-    Example
+    Example - old API
```
ezyang, line 471: From the discussion above, it is not clear which API is "old" and which one is "new". More direct to say "with real tensor input" and "with complex tensor input"

torchaudio/functional.py
```diff
@@ -475,51 +478,83 @@ def phase_vocoder(
...
-    # (new_bins, freq, 2)
-    complex_specgrams_0 = complex_specgrams.index_select(-2, time_steps.long())
-    complex_specgrams_1 = complex_specgrams.index_select(-2, (time_steps + 1).long())
-
-    angle_0 = angle(complex_specgrams_0)
-    angle_1 = angle(complex_specgrams_1)
-
-    norm_0 = torch.norm(complex_specgrams_0, p=2, dim=-1)
-    norm_1 = torch.norm(complex_specgrams_1, p=2, dim=-1)
+    if use_complex:
```
ezyang, line 495: As an alternative to duplicating all the logic here, you could have instead taken the real tensor, viewed it as complex, and then used the complex codepath (viewing it back as real in the end). Something to consider?

test/torchaudio_unittest/torchscript_consistency_impl.py
anjali411: oh okay removed it!

torchaudio/functional.py
anjali411: I think that's a possibility however the goal is to be able to remove the code  in `if not use_complex` branch after a deprecation cycle and just use the code in the other branch (which has similar logic, however there are some substantial differences, e.g., padding logic).

torchaudio/functional.py
anjali411: yeah I agree that's more clear! updated

test/torchaudio_unittest/complex_batch_consistency_test.py
anjali411: I am not sure if it does. I just used the same logic as being used in the existing `test_batch_TimeStretch`.
btw I do see `steam-train-whistle-daniel_simon.wav` listed [here](https://github.com/pytorch/audio/tree/master/test/torchaudio_unittest#waveforms-for-testing-purposes) which is the what's being used in this test.

torchaudio/functional.py
ezyang: well, in my suggestion, you'd delete the real code immediately :) Anyway, this is NBD

test/torchaudio_unittest/complex_batch_consistency_test.py
mthrok: I see. Could you then use synthetic data?

```
waveform = common_utils.get_whitenoise(n_channels=2)
```

should do. I will clean up and do the same for the original tests you referenced.

test/torchaudio_unittest/torchscript_consistency_impl.py
mthrok: I still see `dtype=torch.double` ...

torchaudio/functional.py
anjali411: oh sorry I thought you meant the other way round! yeah that sounds reasonable to me. cc. @mthrok thoughts?

anjali411: on more thought, current lack of autograd support and testing for complex, might introduce bc breaking changes.

mthrok: > oh sorry I misread your comment and thought you meant the other way round! yeah that sounds reasonable to me. cc. @mthrok thoughts?
> on more thought, current lack of autograd support and testing for complex, might introduce bc breaking changes.

I am in favor for not duplicating the logic, however if that introduces BC breaking on real value tensor input, then I think we can wait until the autograd support arrives.

ezyang: If it's R2R with complex insides, the choice of JAX/TF convention doesn't matter, you'll always get the same gradients in the end.

vincentqb:
@anjali411 --`TimeStretch` wraps `phase_vocoder` as an `torch.nn.Module`. Have you tried to run the backward pass with `TimeStretch` the following?

* before this pull request with a real tensor
* after  this pull request with a real tensor
* after this pull request with a complex tensor

This could make for an interesting new test: not to check correctness of the result, but just whether autograd runs without errors.

mthrok:
To be replaced by #1410

PR #760, opened 2020-07-01
Add Torchscript sox effects

mthrok:
This PR add new sox effects functions `torchaudio.sox_effects.apply_effects_tensor` and `torchaudio.sox_effects.apply_effects_file`, which applies sox effects to Tensor object and file object respectively.

This new functions are written from scratch to take advantage of Torchscript, and tested on various effects by comparing against the results from `sox` command. The existing `torchaudio.sox_effects.SoxEffectsChain` does not have this and it does not work correctly on certain format (see #771 )

Also this PR adds `torchaudio.utils.sox_utils` module, which allows users to set verbosity, multi-threading option, buffer size and get the list of supported effects/formats.

Off topic: @cpuhrsch I was adding some tests on data loader with `num_workers > 1`. This test example applies speed perturbation, which alters the length of input Tensor. It would be interesting if we can combine `nestedtensor` there.
https://github.com/mthrok/audio/blob/sox-effects-chain/test/sox_effect/test_dataset.py#L17-L63


cpuhrsch:
Is it possible to split this up a bit more into smaller PRs or would that result in unreasonable amount of work?

mthrok:
> Is it possible to split this up a bit more into smaller PRs or would that result in unreasonable amount of work?

@cpuhrsch I spliced the PR into logical commits. I can still make them separate PRs if necessary, but this way, we do not need to worry about merge conflict between them. Let me know if you feel separate PRs is better.

The commits are basically about
1. 4491e57 Add sox utilities. Add functions that can control verbosity, multithreading, buffer size of sox effects, list format/effect functions.
2. 8fa6af1 Make init/shutdown_sox_effects functions thread safe by adding lock/guard mechanism
3. c4083b4 Add `sox_effects` implementation.
4. b2ba4dc Add tests for `sox_effects` implementation.

vincentqb:
btw, could you add a example code snippet to the description to showcase how to use the new interface?

mthrok:
> btw, could you add a example code snippet to the description to showcase how to use the new interface? with old-new syntax comparison?

@vincentqb  Updated the docstring and added examples e5c716d.

I will add migration message in the next PR in which deprecation messages are added to the existing `SoxEffectsChain` class. I do not think migration step should belong to the docsrtings of the new functions.

vincentqb:
> > btw, could you add a example code snippet to the description to showcase how to use the new interface? with old-new syntax comparison?
> 
> @vincentqb Updated the docstring and added examples [c4dafdb](https://github.com/pytorch/audio/commit/c4dafdb480974fdc65eb704ef983cb39ba036bfb).

Cool!

> I will add migration message in the next PR in which deprecation messages are added to the existing `SoxEffectsChain` class. I do not think migration step should belong to the docsrtings of the new functions.

That's even better. I initially meant to simply add instructions in the description of this PR as a first step.

vincentqb:
cc @eugene-kharitonov for interest in applying sox effects to tensors directly without using files :)

vincentqb:
> This PR add new sox effects functions `torchaudio.sox_effects.apply_effects_tensor` and `torchaudio.sox_effects.apply_effects_file`, which applies sox effects to Tensor object and file object respectively.

Is there value in having an overloaded function that switched between the two depending on input?

torchaudio/utils/sox_utils.py
```diff
@@ -0,0 +1,84 @@
...
+            4-6: increasing levels of debug messages
+
+    See Also:
+        http://sox.sourceforge.net/sox.html
+    """
+    torch.ops.torchaudio.sox_utils_set_verbosity(verbosity)
+
+
+@_mod_utils.requires_module('torchaudio._torchaudio')
+def set_buffer_size(buffer_size: int):
```
vincentqb, line 41: When I see functions like these, I understand that there are global parameters behind the scene to manage the `sox_effects_chain`. Is that the case? Is there a way here to have many chains with different such settings?

torchaudio/csrc/sox_effects.cpp
```diff
@@ -1,41 +1,48 @@
...
-using namespace torch::indexing;
-
 namespace torchaudio {
 namespace sox_effects {
 
 namespace {
 
 enum SoxEffectsResourceState { NotInitialized, Initialized, ShutDown };
 SoxEffectsResourceState SOX_RESOURCE_STATE = NotInitialized;
+std::mutex SOX_RESOUCE_STATE_MUTEX;
```
vincentqb, line 15: So the code is thread through this global lock, right?

torchaudio/csrc/sox_io.cpp
```diff
@@ -125,14 +125,12 @@ void save_audio_file(
     const c10::intrusive_ptr<TensorSignal>& signal,
     const double compression) {
   const auto tensor = signal->getTensor();
-  const auto sample_rate = signal->getSampleRate();
```
vincentqb, line 128: nit: what do people do to get sample rate then?

mthrok:
> > This PR add new sox effects functions `torchaudio.sox_effects.apply_effects_tensor` and `torchaudio.sox_effects.apply_effects_file`, which applies sox effects to Tensor object and file object respectively.
> 
> Is there value in having an overloaded function that switched between the two depending on input?

No.

torchaudio/utils/sox_utils.py
mthrok: > When I see functions like these, I understand that there are global parameters behind the scene to manage the `sox_effects_chain`. Is that the case? 

Yes

> Is there a way here to have many chains with different such settings?

Nope

torchaudio/csrc/sox_effects.cpp
mthrok: > So the code is thread safe through this global lock, right?

Yes.

mthrok:
Thanks!

PR #761, opened 2020-07-01
Get rid of typedefs/SignalInfo and replace AudioMetaData

mthrok:
This PR removes `typedefs.[h|cpp]`, moves `SignalInfo` from `torchaudio` namespace to `torchaudio::sox_io`, removes dummy module from Python interface, then replace `info` return type with `AudioMetaData`.

`torchaudio::sox_io::SignalInfo` struct was expected be data exchange structure between Python and C++ for all `info`, `load` and `save` function of `sox_io` backend. However, as I simplify the function signature of `save` and `load`, they end up not needing such structure, and now `info` function is the only function that use `torchaudio::sox_io::SignalInfo` struct.

On Python API level, we can replace this struct with simple Python class while keeping Torchscript-ability. This removes the necessity to introduce dummy module because we no longer need to annotate `info` function with `torchaudio::sox_io::SignalInfo`.


torchaudio/csrc/register.cpp
```diff
@@ -4,21 +4,10 @@
...
 
 namespace torchaudio {
 namespace {
 
-////////////////////////////////////////////////////////////////////////////////
-// typedefs.h
-////////////////////////////////////////////////////////////////////////////////
-static auto registerSignalInfo =
-    torch::class_<SignalInfo>("torchaudio", "SignalInfo")
-        .def(torch::init<int64_t, int64_t, int64_t>())
```
vincentqb, line 17: nit: can you elaborate why this line is no longer needed?

mthrok: It is still used and moved to `sox_io.h`. Originally I intended this struct to be somewhat in larger scope, but now it is only used by `get_info` func so I moved it to `torchaudio::sox_io` namespace.

test/sox_io_backend/test_info.py
```diff
@@ -32,9 +32,9 @@ def test_wav(self, dtype, sample_rate, num_channels):
         data = get_wav_data(dtype, num_channels, normalize=False, num_frames=duration * sample_rate)
         save_wav(path, data, sample_rate)
         info = sox_io_backend.info(path)
-        assert info.get_sample_rate() == sample_rate
-        assert info.get_num_frames() == sample_rate * duration
-        assert info.get_num_channels() == num_channels
+        assert info.sample_rate == sample_rate
+        assert info.num_frames == sample_rate * duration
+        assert info.num_channels == num_channels
```
vincentqb, line 37: BC breaking?

vincentqb, line 37: The functions are still in there so the change is not BC breaking, right? Is there a reason why you are updating the tests then?

mthrok: > The functions are still in there so the change is not BC breaking, right?

No. These `get_*` functions are removed so this is BC breaking.

vincentqb: If a user tries to use them, what will happen? I don't see any instructions to the user on what to do?

mthrok:
> Let's make sure we have the warnings in place to let the users know what to do if they encounter errors for missing operations.

That approach is good for stable released feature, but considering the product lifecycle of "sox_io" backed, which is still under heavy development, I do not think that really fits.

vincentqb:
> > Let's make sure we have the warnings in place to let the users know what to do if they encounter errors for missing operations.
> 
> That approach is good for stable released feature, but considering the product lifecycle of "sox_io" backed, which is still under heavy development, I do not think that really fits.

What I meant [above](https://github.com/pytorch/audio/pull/761/files#r449272233) when I asked if something was BC-breaking was whether this was changing the existing "sox" (not "sox_io") interface. "sox_io" has not been part of a release yet, so we don't need to add warnings when we are changing the interface.

* If this is not changing "sox" (and only changing "sox_io"), then there is no need for warnings and we can proceed.
* If this is changing "sox", then this is BC-breaking of a feature that was already in a torchaudio release, and we do need a warning.

Can you clarify the [response](https://github.com/pytorch/audio/pull/761/files#r450234351)?

mthrok:
> What I meant [above](https://github.com/pytorch/audio/pull/761/files#r449272233) when I asked if something was BC-breaking was whether this was changing the existing "sox" (not "sox_io") interface. "sox_io" has not been part of a release yet, so we don't need to add warnings when we are changing the interface.

This changes the function signature of `sox_io` backend and has nothing to do with `sox` backend.

mthrok:
thanks

PR #762, opened 2020-07-01
Replace `torchaudio.load_wav` with `common_utils.load_wav`

mthrok:
This PR replaces `torchaudio.load_wav` with `common_utils.load_wav`, which is a wrapper around `scipy.io.wavfile.read`.

This allows each test to not depend on torchaudio's I/O implementation.


mthrok:
thanks

PR #772, opened 2020-07-11
Convert CommonVoice tartar training file from mp3 to wav (8kHz), and remove unused file

engineerchuan:
Used 'sox -i <mp3> <wav>' with default 14.4.2 options

soxi output for the wave file I created:
```
Input File     : 'assets/CommonVoice/cv-corpus-4-2019-12-10/tt/clips/common_voice_tt_00000000.wav'
Channels       : 1
Sample Rate    : 44100
Precision      : 16-bit
Duration       : 00:00:05.00 = 220500 samples = 375 CDDA sectors
File Size      : 441k
Bit Rate       : 706k
Sample Encoding: 16-bit Signed Integer PCM
```

I wasn't sure about whether we need to change Line 111 of torchaudio/datasets/commonvoice.py.

```
class COMMONVOICE(Dataset):
    """
    Create a Dataset for CommonVoice. Each item is a tuple of the form:
    (waveform, sample_rate, dictionary)
    where dictionary is a dictionary built from the tsv file with the following keys:
    client_id, path, sentence, up_votes, down_votes, age, gender, accent.
    """

    _ext_txt = ".txt"
    _ext_audio = ".mp3"
    _folder_audio = "clips"
```


mthrok:
Hi @engineerchuan 

Thanks for working on this.

Looks like `waves_yesno/0_1_0_1_0_1_1_0.wav` and one more file (not sure which one but the one used for `test_gtzan`) are actually used by test, which I overlooked. Could you put them ?

> I wasn't sure about whether we need to change Line 111 of `torchaudio/datasets/commonvoice.py`.

We should be only changing test, so `torchaudio/datasets/commonvoice.py` should stay the same. Did you have an issue that way?

mthrok:
Also, can you try reduce the sampling rate of the new wav file (`assets/CommonVoice/cv-corpus-4-2019-12-10/tt/clips/common_voice_tt_00000000.wav`.) to 8000 Hz? The quality of the audio should not matter to tests and we would like to keep them as small as possible.

mthrok:
Followup: Remove https://github.com/pytorch/audio/blob/master/test/test_datasets.py#L51 and https://github.com/pytorch/audio/blob/master/test/test_datasets.py#L49 to see if this test works on Windows.

@engineerchuan If you have time, would you like to try ^ ?

engineerchuan:
> Followup: Remove https://github.com/pytorch/audio/blob/master/test/test_datasets.py#L51 and https://github.com/pytorch/audio/blob/master/test/test_datasets.py#L49 to see if this test works on Windows.
> 
> @engineerchuan If you have time, would you like to try ^ ?

Yes I will try.

Issue #771, opened 2020-07-08
🐛 Bug: SoxEffectsChain returns wrong result for float32 wav


mthrok:
## 🐛 Bug

**UPDATE**: `SoxEffectsChain` has been marked as deprecated. Please use the new `apply_effects_tensor` or `apply_effects_file` functions.

`SoxEffectsChain` returns wrong result for 32 bit floating point WAV format.

**TOP** Input wav
**MIDDLE** Output from `SoxEffectsChain` with `lowpass -1 300`
**BOTTOM** Output from `sox` command with `lowpass -1 300`

<img width="1663" alt="Screen Shot 2020-07-08 at 13 58 21" src="https://user-images.githubusercontent.com/855818/86953969-3bd43300-c123-11ea-9a1d-60bf8e33034a.png">

## To Reproduce

Run the following script from the top to generate input and result. Run `sox` command to get refertence.

Other bit depths (`int32`, `int16`, `uint8`) seem to work.

<Details>

```python
import torch
import torchaudio
from test.common_utils import save_wav, get_whitenoise


for dtype in ['float32', 'int32', 'int16', 'uint8']:
    sample_rate = 8000
    input_path = f'data/input_{dtype}.wav'
    output_path = f'data/result_{dtype}.wav'

    data = get_whitenoise(sample_rate=sample_rate, dtype=dtype)
    save_wav(input_path, data, sample_rate)

    sec = torchaudio.sox_effects.SoxEffectsChain()
    sec.append_effect_to_chain("lowpass", ["-1", "300"])
    sec.set_input_file(input_path)
    t, sr = sec.sox_build_flow_effects()
    save_wav(output_path, t, sr)
```

</Details>

## Expected behavior

`SoxEffectsChain` should return correct result on float32 wav.

## Environment

```
$ python -m torch.utils.collect_env
Collecting environment information...
PyTorch version: 1.7.0a0+49e12d8
Is debug build: No
CUDA used to build PyTorch: 10.1

OS: Ubuntu 18.04.3 LTS
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
CMake version: version 3.10.2

Python version: 3.8
Is CUDA available: Yes
CUDA runtime version: 10.1.243
GPU models and configuration:
GPU 0: Quadro GP100
GPU 1: Quadro GP100

Nvidia driver version: 418.116.00
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5

[...]

PR #773, opened 2020-07-11
Issue 764: Convert mp3 to wav

engineerchuan:
Convert mp3 to wav or on the fly generation.

### DIFF
--- a/test/test_sox_compatibility.py
+++ b/test/test_sox_compatibility.py
@@ -22,7 +22,7 @@ class TestFunctionalFiltering(common_utils.TempDirMixin, common_utils.Torchaudio
 
         self.NOISE_SAMPLE_RATE = 44100  # N.B. 44.1 kHz required by SoX deemph effect
         noise_waveform_as_int = common_utils.get_whitenoise(
-            sample_rate=self.NOISE_SAMPLE_RATE, duration=5, dtype=torch.int16,
+            sample_rate=self.NOISE_SAMPLE_RATE, duration=5, dtype=torch.int16, scale_factor=0.9,
         )
         self.noise_filepath = self.get_temp_path("whitenoise.wav")
         common_utils.save_wav(
@@ -250,7 +250,7 @@ class TestFunctionalFiltering(common_utils.TempDirMixin, common_utils.Torchaudio
 
         output_waveform = F.bass_biquad(self.noise_waveform, self.NOISE_SAMPLE_RATE, gain, central_freq, q)
 
-        self.assertEqual(output_waveform, sox_output_waveform, atol=1.5e-3, rtol=1e-4)
+        self.assertEqual(output_waveform, sox_output_waveform, atol=1.5e-4, rtol=1e-5)
 
     def test_deemph(self):
         """
--- a/test/test_sox_compatibility.py
+++ b/test/test_sox_compatibility.py
@@ -9,18 +9,37 @@ from . import common_utils
 
 
 @common_utils.skipIfNoSoxBackend
-class TestFunctionalFiltering(common_utils.TorchaudioTestCase, common_utils.TempDirMixin):
-    backend = 'sox'
+class TestFunctionalFiltering(
+    common_utils.TempDirMixin, common_utils.TorchaudioTestCase
+):
+    backend = "sox"
 
     def setUp(self):
-        super().setUp()
+        common_utils.TempDirMixin.setUp(self)
+        common_utils.TorchaudioTestCase.setUp(self)
+
+        NOISE_SAMPLE_RATE = 44100
+        INT16_MAX = 32767
+        noise_waveform = common_utils.get_whitenoise(
+            sample_rate=NOISE_SAMPLE_RATE, duration=5
+        )
+        # convert to int16, scipy wav writer automatically selects wav format based on tensor dtype
+        noise_waveform_as_int = torch.as_tensor(
+            INT16_MAX * noise_waveform, dtype=torch.int16
+        )
+        self.noise_filepath = self.get_temp_path("whitenoise.wav")
+        common_utils.save_wav(
+            self.noise_filepath, noise_waveform_as_int, NOISE_SAMPLE_RATE
+        )
 
     def test_gain(self):
-        test_filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
+        test_filepath = common_utils.get_asset_path(
+            "steam-train-whistle-daniel_simon.wav"
+        )
         waveform, _ = torchaudio.load(test_filepath)
 
         waveform_gain = F.gain(waveform, 3)
-        self.assertTrue(waveform_gain.abs().max().item(), 1.)
+        self.assertTrue(waveform_gain.abs().max().item(), 1.0)
 
         E = torchaudio.sox_effects.SoxEffectsChain()
         E.set_input_file(test_filepath)
@@ -30,7 +49,9 @@ class TestFunctionalFiltering(common_utils.TorchaudioTestCase, common_utils.Temp
         self.assertEqual(waveform_gain, sox_gain_waveform, atol=1e-04, rtol=1e-5)
 
     def test_dither(self):
-        test_filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
+        test_filepath = common_utils.get_asset_path(
+            "steam-train-whistle-daniel_simon.wav"
+        )
         waveform, _ = torchaudio.load(test_filepath)
 
         waveform_dithered = F.dither(waveform)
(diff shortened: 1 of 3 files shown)


engineerchuan:
I tried to avoid converting to `int16` tensor which saves as 16 bit PCM wave.  I tried to keep it as `float32` tensor which should save as 32 bit.  However I ran into the following error:

```
formats: raw can't encode Signed Integer PCM to 25-bit
```

The temp file that was written out had these properties:

```
$ soxi /tmp/test.wav

Input File     : '/tmp/test.wav'
Channels       : 1
Sample Rate    : 44100
Precision      : 25-bit
Duration       : 00:00:05.00 = 220500 samples = 375 CDDA sectors
File Size      : 882k
Bit Rate       : [1.41M](url)
Sample Encoding: 32-bit Floating Point PCM
```

engineerchuan:
@mthrok  I'm not sure what's wrong with the binary macos conda.  It seems to be related to some specification mismatch that my code should not have affected.

test/test_sox_compatibility.py
```diff
@@ -9,9 +9,27 @@
...
+            sample_rate=NOISE_SAMPLE_RATE, duration=5
+        )
+        # convert to int16, scipy wav writer automatically selects wav format based on tensor dtype
+        noise_waveform_as_int = torch.as_tensor(
+            INT16_MAX * noise_waveform, dtype=torch.int16
+        )
+        self.noise_filepath = self.get_temp_path("whitenoise.wav")
+        common_utils.save_wav(
+            self.noise_filepath, noise_waveform_as_int, NOISE_SAMPLE_RATE
+        )
```
mthrok, line 31: As far as I understand, the current implementation runs test this way

```
1. Generate a white noise
 |
 v
2. Save the noise to file --------> 3. Run sox effect from the file
 |                                            |
 | 4.1. Load data from the file               |
 v                                            v
4.2. Run functional on loaded data -------> compare    
```

This process has a duplicated data preparation at `1.` and `4.1.`. The result of data loaded from the file at `4.1.` is already available at `1`, so it's redundant. And `4.1.` requires a loading function, which is `torchaudio.load` currently. This has a problem because the test now depends on both `torchaudio.load` and functionals being tested. When test is implemented this way and the test fails it is not immediately clear which caused a problem, and the worst case, both implementation are wrong but somehow the test passes so that we do not notice the actual issue. To avoid that we should avoid using `torchaudio.load` in this test suite. We could use `common_utils.load_wav` but simplifying the process to generate synthetic data allows us to do it without it.

```
1. Generate a white noise ---------------> 2.1. Save the noise to file
|                                            |
v                                            v
3. Run functional the generated data.         2.2. Run sox effect from the file
|                                            |
-----------------> compare <------------------
```

test/test_sox_compatibility.py
```diff
@@ -251,29 +253,26 @@ def test_bass(self):
...
+        E.set_input_file(self.noise_filepath)
         E.append_effect_to_chain("bass", [gain, central_freq, str(q) + 'q'])
         sox_output_waveform, sr = E.sox_build_flow_effects()
 
-        waveform, sample_rate = torchaudio.load(noise_filepath, normalization=True)
+        waveform, sample_rate = torchaudio.load(self.noise_filepath, normalization=True)
         output_waveform = F.bass_biquad(waveform, sample_rate, gain, central_freq, q)
 
-        self.assertEqual(output_waveform, sox_output_waveform, atol=1.5e-4, rtol=1e-5)
+        self.assertEqual(output_waveform, sox_output_waveform, atol=1e-3, rtol=1e-4)
```
mthrok, line 264: This is dues to edge effect, and the trick to avoid reducing tolerance here is to scale the intensity of the generated whitenoise (say `x0.9`).

test/test_sox_effects.py
```diff
@@ -11,7 +11,7 @@
 class Test_SoxEffectsChain(common_utils.TorchaudioTestCase):
     backend = 'sox'
 
-    test_filepath = common_utils.get_asset_path("steam-train-whistle-daniel_simon.mp3")
+    test_filepath = common_utils.get_asset_path("steam-train-whistle-daniel_simon.wav")
```
mthrok, line 14: In this test suite, there are couple of usage of `torchaudio.load`, but now the test file is `wav` format, can you replace them with `common_utils.load_wav`??

test/test_sox_compatibility.py
```diff
@@ -9,9 +9,27 @@
...
 @common_utils.skipIfNoSoxBackend
-class TestFunctionalFiltering(common_utils.TorchaudioTestCase):
+class TestFunctionalFiltering(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
     backend = 'sox'
 
+    def setUp(self):
+        common_utils.TempDirMixin.setUp(self)
+        common_utils.TorchaudioTestCase.setUp(self)
+
+        NOISE_SAMPLE_RATE = 44100
```
mthrok, line 19: Can you use 8000 Hz? though 44100 Hz is common in audio, many speech-related tasks only use 16k or 8k Hz and 8k Hz will reduce test run time.

mthrok:
> I tried to avoid converting to `int16` tensor which saves as 16 bit PCM wave. I tried to keep it as `float32` tensor which should save as 32 bit. However I ran into the following error:
> 
> ```
> formats: raw can't encode Signed Integer PCM to 25-bit
> ```
> 
> The temp file that was written out had these properties:
> 
> ```
> $ soxi /tmp/test.wav
> 
> Input File     : '/tmp/test.wav'
> Channels       : 1
> Sample Rate    : 44100
> Precision      : 25-bit
> Duration       : 00:00:05.00 = 220500 samples = 375 CDDA sectors
> File Size      : 882k
> Bit Rate       : [1.41M](url)
> Sample Encoding: 32-bit Floating Point PCM
> ```

I am not sure where you got `formats: raw can't encode Signed Integer PCM to 25-bit` error, but the current sox effects does not work correctly for floating point PCM ([example](https://github.com/pytorch/audio/issues/771)), so generating white noise as 16bit signed integer PCM is correct (and probably the only) working way. Once https://github.com/pytorch/audio/pull/760 has landed then this can be cleaned up.

mthrok:
> @mthrok I'm not sure what's wrong with the binary macos conda. It seems to be related to some specification mismatch that my code should not have affected.

Yes, that happens when PyTorch's binary build fails. You can ignore it.

mthrok:
Thanks!

test/test_sox_compatibility.py
```diff
@@ -9,9 +9,27 @@
 
 
 @common_utils.skipIfNoSoxBackend
-class TestFunctionalFiltering(common_utils.TorchaudioTestCase):
+class TestFunctionalFiltering(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
     backend = 'sox'
 
+    def setUp(self):
+        common_utils.TempDirMixin.setUp(self)
+        common_utils.TorchaudioTestCase.setUp(self)
```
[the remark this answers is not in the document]
engineerchuan: So with the current order, wouldn't `TorchaudioTestCase.setUp()` never get called? Which means that `TestBaseMixin.setUp()` would never get called? which means `set_audio_backend(self.backend)` never gets called?

What if in the future we add more functionality that we want to rely on with other mixins?  Wouldn't we want to have this robust to that?  Thoughts?

test/test_sox_compatibility.py
engineerchuan: Sox `deemph` seems to only work on 44.1kHz  or 48kHz

`deemph: sample rate must be 44100 (audio-CD) or 48000 (DAT)`

test/test_sox_compatibility.py
engineerchuan: Tried this was `bass` and still got a few errors exceeding 1e-3.
```
E   AssertionError: False is not true : Tensors failed to compare as equal! With rtol=0.0001 and atol=0.001, found 136 element(s) (out of 220500) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.0013254880905151367 (0.8861521482467651 vs. 0.88482666015625), which occurred at index (0, 47224).
```

test/test_sox_compatibility.py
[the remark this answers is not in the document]
mthrok: with `class TestFo(TempDirMixin, TorchaudioTestCase)`, the following happens.

1. Test framework tries to invoke `setUp` on `TestFoo`.

Case 1.  If `TestFoo` has `setUp` method defined
2.1. TestFoo.setUp is called.
2.2. If `TestFoo.setUp` has `super().setUp()`, then `setUp` on super class is called.
2.3. It keeps resolving super class methods as long as `super().setUp()` is defined.  (supposed to stop at `unittest.TestCase`).

Case 2. If `TestFoo` does not have `setUp` method defined
2. According to MRO rule, it invokes `TempDirMixin.setUp`.
3 Inside of `TempDirMixin.setUp`, `super().setUp()` is executed.
4. `super()` finds `TorchaudioTestCase` as super class so `TorchaudioTestCase.setUp` is invoked.
5.. 3. & 4. is repeated as long as `super().setUp()` is defined in `setUp` method of any class (supposed to stop at `unittest.TestCase`).


Suppose all the `setUp` methods have  call to `super().setUp()`

For case 1 the following methods are called.
`TestFoo.setUp()` -> `TempDirMixin.setUp()` -> `TorchaudioTestCase.setUp()` -> `torch.testing._internal.common_utils.TestCase.setUp()` -> some other classes -> `unites.TestCase`.

For case 2. the following methods are called.
`TempDirMixin.setUp()` -> `TorchaudioTestCase.setUp()` -> `torch.testing._internal.common_utils.TestCase.setUp()` -> some other classes -> `unites.TestCase`.

`MRO` defines where your method starts, but the chain of superclass calls depend on the implementation. (`super().setUp()`).

And the design of MixIn here is that they add features to base `TorchaudioTestCase`, so it makes sense to start the call chain from `Mixin`. (Because `TorchaudioTestCase` class is not aware of `Mixin`.)

When you have multiple MixIn, and you implement the test class as follow;

`class MyTest(Mixin2, Mixin1, TorchaudioTestCase)`

Then the inheritance looks like this;

```
TorchaudioTestCase
↑ super()
Mixin1
↑super()
Mixin2
```

and the MRO starts from `Mixin2`.

test/test_sox_compatibility.py
engineerchuan: Sure, the only caveat is you'll have to generate two versions of the same signal, a `float32` and `int16` version, or we do the conversion outside the `get_whitenoise` function.  Thoughts?

test/test_sox_compatibility.py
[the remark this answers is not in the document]
mthrok: It is confusing but when you do multiple inheritance in Python, you need to put most parent class on the right side. I often get confused on this one and had to google every time. That will give you a right inheritance order, and with that, `super()` should work as expected.

engineerchuan:
@mthrok Check out the latest.  I think this approach captures what you want but it introduces quite a number of instabilities and we would need to relax atol thresholds for 4-5 tests.  Why do you think that is?  Do you think there's some issue with the float -> int quantization that differs?

mthrok:
Thanks!

PR #777, opened 2020-07-13
Add macOS CPU unittest

mthrok:


peterjc123:
@mthrok So actually you are sharing the common script for unit tests between MacOS and Linux, right?

mthrok:
> @mthrok So actually you are sharing the common script for unit tests between MacOS and Linux, right?

Yeah, I thought of splitting them since I asked you to split Windows from Linux, but I felt that it's more beneficial to keep *nix systems same. (this is a vague line and subjective, splitting macOS and Linux feels like later when we need to apply changes to Linux, we might forget to apply the same changes to macOS) However, my goal is to add address sanitizer on macOS unit test, so depending on how it is done, I might split Linux and macOS later.

mthrok:
> That's okay for me. But I'm just a bit worried that the same thing could happen to Windows, too.

Yes, it's already happening. In https://github.com/pytorch/audio/pull/664 I forgot to remove style check from windows job, so I am doing it in https://github.com/pytorch/audio/pull/778 haha.

mthrok:
Thanks!

PR #779, opened 2020-07-13
Merge sox effect and sox_io C++ implementation

mthrok:
This PR is built on top of #760 , and it merges `sox_io` backend `load`/`save` functions with `sox_effects` on C++ side. No BC breaking as it's C++ internal.

~The number of frames to process at a time (`frames_per_chunk` parameter of `save` function) can be adjusted via `torchaudio.utils.sox_utils.set_buffer_size`, so removed from the function signature.~ Addressed in #780 separately

### DIFF
--- a/torchaudio/csrc/sox_effects_chain.cpp
+++ b/torchaudio/csrc/sox_effects_chain.cpp
@@ -46,6 +46,9 @@ struct TensorInputPriv {
 struct TensorOutputPriv {
   std::vector<sox_sample_t>* buffer;
 };
+struct FileOutputPriv {
+  sox_format_t* sf;
+};
 
 /// Callback function to feed Tensor data to SoxEffectChain.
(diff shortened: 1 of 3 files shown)


vincentqb:
> merges `sox_io` backend `load`/`save` functions with `sox_effects` on C++ side

Can you elaborate? I believe this is the change we talked about where `frames_per_chunk` can be detected?

mthrok:
> > merges `sox_io` backend `load`/`save` functions with `sox_effects` on C++ side
> 
> Can you elaborate? I believe this is the change we talked about where `frames_per_chunk` can be detected?

This change gets rid of custom implementation for load/save and it uses sox effects chain implementation for them. `frames_per_chunk` is controlled by sox utils' buffer size.

mthrok:
thanks

PR #780, opened 2020-07-13
Remove frames_per_chunk argument from save

mthrok:
In #779, we plan to remove `frames_per_chunk` parameter from `save` function, but it will take some time before we can land #779, so we go ahead and remove the parameter first to reduce the conflict caused by interface change.


mthrok:
Thanks!

PR #781, opened 2020-07-14
Do not use SoxEffectsChain in sox compatibility test

mthrok:
This PR replaces `torchaudio.sox_effects.SoxEffectsChain` in `test_sox_compatibility` with bare `sox` command.

The parity of `torchaudio.sox_effects.SoxEffectsChain` against `sox` command is not tested and it has known issues https://github.com/pytorch/audio/issues/771, therefore it is not appropriate to use this class for testing other functions.

### DIFF
--- a/test/test_sox_compatibility.py
+++ b/test/test_sox_compatibility.py
@@ -1,492 +1,294 @@
 import unittest
 
 import torch
-import torchaudio
 import torchaudio.functional as F
-import torchaudio.transforms as T
 
-from . import common_utils
+from .common_utils import (
+    skipIfNoSoxBackend,
+    skipIfNoExec,
+    TempDirMixin,
+    TorchaudioTestCase,
+    get_asset_path,
+    sox_utils,
+    load_wav,
+    save_wav,
+    get_whitenoise,
+)
 
 
-@common_utils.skipIfNoSoxBackend
-class TestFunctionalFiltering(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
-    backend = 'sox'
+@skipIfNoSoxBackend
+@skipIfNoExec('sox')
+class TestFunctionalFiltering(TempDirMixin, TorchaudioTestCase):
+    def run_sox_effect(self, input_file, effect):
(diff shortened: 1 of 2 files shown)


test/test_sox_compatibility.py
```diff
@@ -1,510 +1,304 @@
...
-
-        self.assertEqual(waveform_dithered, sox_dither_waveform, atol=1e-04, rtol=1e-5)
-        E.clear_chain()
-
-        E.append_effect_to_chain("dither", ["-s"])
-        sox_dither_waveform_ns = E.sox_build_flow_effects()[0]
-
-        self.assertEqual(waveform_dithered_noiseshaped, sox_dither_waveform_ns, atol=1e-02, rtol=1e-5)
-
-    def test_vctk_transform_pipeline(self):
```
mthrok, line 67: This test has been removed because it does not test anything new.

`channels` and `rate` effects are no effect because the input file is already one channel with 16k Hz, then the remaining effect is `gain -h gain -rh dither -s`, but applying gain twice does not make sense (`sox` command returns error) and `dither` is tested separately.

mthrok:
The tests failing on macOS has nothing to do with this PR, and is addressed at #782 .
(This PR touches `test/test_sox_compatibility.py`, not the failing `test/test_sox_effects.py`)

test/test_sox_compatibility.py
```diff
@@ -1,510 +1,304 @@
...
+    def test_dither_noise(self):
+        path = get_asset_path('steam-train-whistle-daniel_simon.wav')
+        data, _ = load_wav(path)
+        result = F.dither(data, noise_shaping=True)
+        self.assert_sox_effect(result, path, ['dither', '-s'], atol=1.5e-4)
 
     def test_lowpass(self):
-        """
-        Test biquad lowpass filter, compare to SoX implementation
-        """
```
vincentqb, line 91: Why is the description removed?

mthrok: Because all the tests in the class have the same description. It's redundant and does not carry new useful information.

mthrok:
Thanks

PR #782, opened 2020-07-14
Skip lowpass test on macOS

mthrok:
`test/test_sox_effects.py::Test_SoxEffectsChain::test_lowpass_speed` has some issue on our macOS CI, even though there was no issue at #777 .

While we figure out the fix, we disable this test for macOS.


mthrok:
Merging this as it affects all the PR.

PR #687, opened 2020-06-04
Migrate kaldi spectrogram

bhargavkathivarapu:
This PR migrates the kaldi spectrogram test (#597 ) from test/test_compliance_kaldi.py to test/kaldi_compatibility_impl.py

PR #681, opened 2020-06-03
Migrate kaldi mfcc

bhargavkathivarapu:
This PR migrates the kaldi mfcc test (#597 ) from test/test_compliance_kaldi.py to test/kaldi_compatibility_impl.py

PR #783, opened 2020-07-14
Issue 764: Switch Pitch Detection Test to use On the Fly Generation instead of file.

engineerchuan:
1. Switch Pitch Detection to on the fly generation.
2. Refactor the integer encoding that whitenoise and sinusoid uses to be shared. 
3. Refactor `test_compliance_kaldi.py`.  Cannot get rid of `kaldi_file_8000.wav` because we compare the output to corresponding `ark` files.


engineerchuan:
Maybe there is no point in trying to use on the fly generation if we have to keep `kaldi_file_8000.wav` around because we are keeping cached `ark` files around.

mthrok:
> Maybe there is no point in trying to use on the fly generation if we have to keep `kaldi_file_8000.wav` around because we are keeping cached `ark` files around.

Yes, the problem is that the code used to generate those `ark` files are not checked-in so we cannot make modifications to test. If we can recover the code used for those `ark` files, we can switch the test to completely on-the-fly data generation, which is done for other kaldi compatible tests. #672 #681 #687 #690 and #699.

test/functional_cpu_test.py
```diff
@@ -300,19 +300,23 @@ def test_linearity_of_istft4(self):
 
 class TestDetectPitchFrequency(common_utils.TorchaudioTestCase):
     def test_pitch(self):
```
mthrok, line 302: Since this test is now parameterized by frequency (100 and 440) can you use `parameterized.expand`? that will simplify the test implementation.

test/functional_cpu_test.py
```diff
@@ -300,19 +300,23 @@ def test_linearity_of_istft4(self):
...
-        test_filepath_100 = common_utils.get_asset_path("100Hz_44100Hz_16bit_05sec.wav")
-        test_filepath_440 = common_utils.get_asset_path("440Hz_44100Hz_16bit_05sec.wav")
+        SAMPLE_RATE = 44100
```
mthrok, line 303: Why is this variable capitalized?

engineerchuan: I was thinking this is a constant per https://www.python.org/dev/peps/pep-0008/#id48?  What case would you prefer?

engineerchuan:
> > Maybe there is no point in trying to use on the fly generation if we have to keep `kaldi_file_8000.wav` around because we are keeping cached `ark` files around.
> 
> Yes, the problem is that the code used to generate those `ark` files are not checked-in so we cannot make modifications to test. If we can recover the code used for those `ark` files, we can switch the test to completely on-the-fly data generation, which is done for other kaldi compatible tests. #672 #681 #687 #690 and #699.
> 
> This situation is very bad because resample is used in other places too resample function has to be tested thoroughly but the data stored as `ark` provides very limited number of use cases.

Let's address this in a follow up.

test/test_batch_consistency.py
```diff
@@ -48,15 +48,12 @@ def test_griffinlim(self):
         )
 
     def test_detect_pitch_frequency(self):
-        filenames = [
-            'steam-train-whistle-daniel_simon.wav',  # 2ch 44100Hz
-            # Files from https://www.mediacollege.com/audio/tone/download/
-            '100Hz_44100Hz_16bit_05sec.wav',  # 1ch
-            '440Hz_44100Hz_16bit_05sec.wav',  # 1ch
+        waveform_and_sample_rates = [
+            torchaudio.load(common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')),  # 2ch 44100Hz
```
mthrok, line 52: Can you remove `steam-train-whistle-daniel_simon.wav` case?
I think we can cover pretty much the same thing with synthetic data.
We need to test the combinations of `sample_rate = 8000, 16000, 44100`, `n_channels = 1, 2`, and some frequency (100 and 440 should do). 

You can combine them using `parameterized.expand` and `itertools.product`. See https://github.com/pytorch/audio/blob/master/test/sox_io_backend/test_info.py#L24-L29

test/common_utils/data_utils.py
```diff
@@ -13,6 +13,28 @@ def get_asset_path(*paths):
     return os.path.join(_TEST_DIR_PATH, 'assets', *paths)
 
 
+def convert_tensor_encoding(
+    tensor: torch.tensor,
+    dtype: torch.dtype,
+):
```
mthrok, line 19: Nice!

test/common_utils/test_case_utils.py
```diff
@@ -69,6 +69,9 @@ def skipIfNoModule(module, display_name=None):
     return unittest.skipIf(not is_module_available(module), f'"{display_name}" is not available')
 
 
+def name_func(func, _, params):
+    return f'{func.__name__}_{"_".join(str(arg) for arg in params.args)}'
+
```
mthrok, line 74: I prefer not touching `name_func` from `sox_io_backend` module, because some tests have very different way of parameterization and this `name_func` is not as generally applicable as it seems. For example, in #760 I had to use different `name_func` https://github.com/pytorch/audio/pull/760/files#diff-033abc42147aa7f1b5d3b3c8bedb2116R1. 
but if you want to move this function, please put it under `test/common_utils/parameterized_utils.py`, and then maybe rename it to something like `simple_name_func`, to be a little bit less generic.

test/test_sox_effects.py
```diff
@@ -43,8 +43,8 @@ def test_lowpass_speed(self):
         E.append_effect_to_chain("speed", speed)
         E.append_effect_to_chain("rate", si.rate)
         x, sr = E.sox_build_flow_effects()
-        # check if effects worked
-        self.assertEqual(x.size(1), int((si.length / si.channels) / speed))
+        # check if effects worked, add small tolerance for rounding effects
+        self.assertEqual(x.size(1), int((si.length / si.channels) / speed), atol=1, rtol=1e-8)
```
mthrok, line 47: Do you think this resolve #782 ?

test/common_utils/test_case_utils.py
mthrok: You can also use `lambda` in test code. 

`name_func=lambda f, _, p: f'{f.__name__}_{"_".join(str(arg) for arg in p.args)}'`

test/test_sox_effects.py
engineerchuan: Yes, I think so.

test/common_utils/test_case_utils.py
```diff
@@ -69,6 +69,9 @@ def skipIfNoModule(module, display_name=None):
     return unittest.skipIf(not is_module_available(module), f'"{display_name}" is not available')
 
 
+def name_func(func, _, params):
```
mthrok, line 72: Now that this function is not used by any test, can you remove it??

mthrok:
Thanks!

PR #787, opened 2020-07-16
Add deprecation warnings to SoxEffect and SoxEffectsChain

mthrok:


torchaudio/sox_effects/sox_effects.py
```diff
@@ -179,19 +179,29 @@ def apply_effects_file(
 
 
 @_mod_utils.requires_module('torchaudio._torchaudio')
+@_mod_utils.deprecated('Please migrate to `apply_effects_file` or `apply_effects_tensor`.')
```
vincentqb, line 182: nit: is it clear in which modules these functions live from the message?

torchaudio/_internal/module_utils.py
```diff
@@ -48,7 +48,7 @@ def decorator(func):
         def wrapped(*args, **kwargs):
             message = (
                 f'{func.__module__}.{func.__name__} has been deprecated '
-                f'and will be removed from {"future" if version is None else version} release.'
+                f'and will be removed from {"future" if version is None else version} release. '
```
vincentqb, line 51: nice catch :)

torchaudio/sox_effects/sox_effects.py
mthrok: That information is automatically encoded. See https://github.com/pytorch/audio/pull/787/files#diff-3ce6065b377e608566858654a3b2c2adL50

mthrok:
Thanks

Issue #786, opened 2020-07-15
Issues with documentation generations


mthrok:
`make html` command issues the following warnings;

```
/torchaudio/torchaudio/models/wav2letter.py:docstring of torchaudio.models.Wav2Letter:4: WARNING: Inline interpreted text or phrase reference start-string without end-string.
o/torchaudio/torchaudio/transforms.py:docstring of torchaudio.transforms.GriffinLim:4: WARNING: Footnote [1] is not referenced.
/torchaudio/torchaudio/transforms.py:docstring of torchaudio.transforms.GriffinLim:8: WARNING: Footnote [2] is not referenced.
/torchaudio/torchaudio/transforms.py:docstring of torchaudio.transforms.GriffinLim:13: WARNING: Footnote [3] is not referenced.
/torchaudio/docs/source/models.rst: WARNING: document isn't included in any toctree
```

Issue #638, opened 2020-05-13
ComputeDeltas equation is not rendered


keunwoochoi:
https://pytorch.org/audio/functional.html#compute-deltas

A closing parenthesis `}` is missing, but I'm not entirely sure what it is supposed to be so instead of making a PR I'm reporting this.

mthrok:
Thanks for reporting. I will take a look at it.

vincentqb:
The bracket is missing for the upper bound in the final sum :) Feel free to open a PR for this!
```
\[d_t = \frac{\sum_{n=1}^{\text{N}} n (c_{t+n} - c_{t-n})}{2 \sum_{n=1}^{\text{N}} n^2} \]
```

mthrok:
It's unbalanced curly bracket. Will be fixed in #788.

mthrok:
Fixed by #788 

https://pytorch.org/audio/0.7/functional.html#compute-deltas

Issue #549, opened 2020-04-14
Backend functions not in documentation


vincentqb:
Even though the documentation was updated for [0.4.0](https://github.com/pytorch/audio/releases/tag/v0.4.0), functions such as `set_audio_backend` do not appear in the [doc](pytorch.org/audio).

vincentqb:
Closed since `torchaudio.set_audio_backend` is [here](https://pytorch.org/audio/stable/torchaudio.html#torchaudio.set_audio_backend).

PR #788, opened 2020-07-16
Update documentation and fix docstrings

mthrok:
- Addresses #549 #638 #786 
- Add `torchaudio` top level module doc
- Separate `torchaudio` top level module doc from `index.html`
- Add `backend` module doc.
- Remove `-> None` from function signature as it adds noise to documentation
- Changed function argument name of `torchaudio.backend.sox_io_backend.save` from `tensor` to `src`, so that it matches with the reset of backends.
- Tweak bunch of docstrings

### DIFF
--- a/docs/source/backend.rst
+++ b/docs/source/backend.rst
@@ -7,7 +7,7 @@ torchaudio.backend
 To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
 
 .. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the signatures.
+   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
 
 .. note::
    Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
--- a/docs/source/backend.rst
+++ b/docs/source/backend.rst
@@ -20,7 +20,7 @@ There are currently three implementations available.
 
 ``sox`` backend is the original backend which is built on ``libsox``. This module is currently default but is known to have number of issues, such as wrong handling of WAV files other than 16-bit signed integer. Users are encouraged to use ``sox_io`` backend. This backend requires C++ extension module and is not available on Windows system.
 
-``sox_io`` backend is the new backend which is built on ``libsox`` and bound to Python with ``Torchscript``. This module is thoroughly tested and addresses all the known issues ``sox`` backend has. Function calls to this backend can be Torchscriptable. This backend requires C++ extension module and is not available on Windows system.
+``sox_io`` backend is the new backend which is built on ``libsox`` and bound to Python with ``Torchscript``. This module addresses all the known issues ``sox`` backend has. Function calls to this backend can be Torchscriptable. This backend requires C++ extension module and is not available on Windows system.
 
 ``soundfile`` backend is built on ``PySoundFile``. You need to install ``PySoundFile`` separately.
 
--- a/docs/source/backend.rst
+++ b/docs/source/backend.rst
@@ -1,0 +1,138 @@
+.. _backend:
+
+torchaudio.backend
+==================
+
+:mod:`torchaudio.backend` module provides implemenrations for audio file I/O, using different backend libraries
+To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
+
+.. warning::
+   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the signatures.
+
+.. note::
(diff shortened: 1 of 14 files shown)


docs/source/backend.rst
```diff
@@ -7,7 +7,7 @@ torchaudio.backend
 To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
 
 .. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the signatures.
+   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
```
vincentqb, line 10: Actually,  let's simplify the statement:
```
``sox`` backend is currently the default for backward compatibility reason. It is recommended to migrate to ``sox_io`` backend, and use its new interface, which will become the default in a future version. See [here](link) for more information about how to migrate.
```

mthrok: It's better to do that after we figure out the plan for deprecation of `sox` backend.

vincentqb: My point is mostly to remove the idea of "issues", "interface refinement". How about this then?
```
``sox`` backend is the default for backward compatibility reason, but it is recommended to use ``sox_io`` backend instead. Note, however, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
```

Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the signatures.

mthrok: My point here is to discourage users to use `sox` backend and raise the awareness of the danger of using it. If we are not upfront of issues, users would not take efforts to migrate because it takes engineering efforts to do so.

vincentqb: Can you elaborate on "dangers"?

mthrok: As it's summarized in https://github.com/pytorch/audio/pull/726

`sox` backend does not load wav files other than 16-bit signed integer, `save` function does not preserve data, `info` returns the number of samples instead of the number of frames. This affects every aspect of ML application, from training to inference. There is always chance to introduce wrong data when using `load` and if one works on generative models, `save` function does not correctly save the data.

Ideally we should fix it but the inconsistent design and band aides placed in a wrong place to fix the issue(load_wav), it's very difficult to do that now without breaking BC.

vincentqb: I understand though I don't think this is the correct wording. Since I do not want to block this change here for this, I'll go ahead and approve this pull request anyway.

docs/source/backend.rst
```diff
@@ -0,0 +1,138 @@
...
+
+There are currently three implementations available.
+
+    * :ref:`sox<sox_backend>`
+    * :ref:`sox_io<sox_io_backend>`
+    * :ref:`soundfile<soundfile_backend>`
+
+``sox`` backend is the original backend which is built on ``libsox``. This module is currently default but is known to have number of issues, such as wrong handling of WAV files other than 16-bit signed integer. Users are encouraged to use ``sox_io`` backend. This backend requires C++ extension module and is not available on Windows system.
+
+``sox_io`` backend is the new backend which is built on ``libsox`` and bound to Python with ``Torchscript``. This module is thoroughly tested and addresses all the known issues ``sox`` backend has. Function calls to this backend can be Torchscriptable. This backend requires C++ extension module and is not available on Windows system.
```
vincentqb, line 23: nit: Let's have a more neutral tone here:
> This module ~~is thoroughly tested and~~ addresses all the known issues ``sox`` backend has.

torchaudio/backend/utils.py
```diff
@@ -29,12 +33,13 @@ def list_audio_backends() -> List[str]:
     return backends
 
 
-def set_audio_backend(backend: Optional[str]) -> None:
+def set_audio_backend(backend: Optional[str]):
```
vincentqb, line 36: nit: why are you removing the typing?

torchaudio/models/wav2letter.py
```diff
@@ -1,14 +1,13 @@
...
 
 __all__ = ["Wav2Letter"]
 
 
 class Wav2Letter(nn.Module):
-    r"""Wav2Letter model architecture from the `"Wav2Letter: an End-to-End ConvNet-based Speech Recognition System"
-     <https://arxiv.org/abs/1609.03193>`_ paper.
+    r"""Wav2Letter model architecture from the `Wav2Letter an End-to-End ConvNet-based Speech Recognition System`_.
+
+    .. _Wav2Letter an End-to-End ConvNet-based Speech Recognition System: https://arxiv.org/abs/1609.03193
```
vincentqb, line 10: nit: How will this be rendered?

torchaudio/sox_effects/sox_effects.py
```diff
@@ -114,6 +119,40 @@ def apply_effects_tensor(
...
+        >>> effects = [
+        ...     ["lowpass", "-1", "300"],  # apply single-pole lowpass filter
+        ...     ["rate", "8000"],  # change sample rate to 8000
+        ... ]
+        >>> trans = SoxEffectTensorTransform(effects, input_sample_rate)
+        >>>
+        >>> # Dump it to file and load
+        >>> path = 'sox_effect.zip'
+        >>> torch.jit.script(trans).save(path)
+        >>> trans = torch.jit.load(path)
```
vincentqb, line 150: nit: `transform`

docs/source/backend.rst
vincentqb: Thanks for the clarification. I understand what you mean though I don't think this is the correct wording. Since I do not want to block this change here for this, I'll approve the pull request.

torchaudio/backend/utils.py
mthrok: This typing is redundant and messed up the documentation rendering.

torchaudio/models/wav2letter.py
mthrok: Screen Shot 2020-07-20 at 17 59 32

mthrok:
thanks!

Issue #764, opened 2020-07-02
Remove insignificant test assets


mthrok:
@astaff had introduced guideline for test assets in https://github.com/pytorch/audio/pull/759 and we can get rid of the following existing assets.

 - [ ] `100Hz_44100Hz_16bit_05sec.wav` sine wave, should be replaced by on-the-fly generation.
 - [ ] `440Hz_44100Hz_16bit_05sec.wav` sine wave, should be replaced by on-the-fly generation.
 - [ ] `CommonVoice/cv-corpus-4-2019-12-10/tt/clips/common_voice_tt_00000000.mp3` whitenoise, should be converted to wav so that test does not require mp3 decoder.
 - [ ] `dtmf_30s_stereo.mp3` not used.
 - [ ]`genres/noise/noise.0000.wav` not used.
 - [ ] `kaldi_file.wav` sine wave only contains 20 samples and I do not think this is appropriate for test.
 - [ ] `kaldi_file_8000.wav` sine wave, should prefer on-the-fly generation.
 - [ ] `sinewave.wav` sine wave, should prefer on-the-fly generation.
 - [ ] `steam-train-whistle-daniel_simon.mp3` should be replaced by `steam-train-whistle-daniel_simon.wav`
 - [ ] `test.wav` file generated during `test_io.py` accidentally checked in
 - [ ] `waves_yesno/0_1_0_1_0_1_1_0.wav` not used
 - [ ] `whitenoise_1min.mp3` should be replaced by on-the-fly generation.
 - [ ] `whitenoise.mp3` should be replaced by on-the-fly generation.
 - [ ] `whitenoise.wav` should be replaced by on-the-fly generation.

## General Direction for replacing assets with on-the-fly generation

1. Create Tensor
```
common_utils.get_sinusoid
common_utils.get_whitenoise
```

2. Get temporary file path

```
self.get_temp_path('foo.wav')
# suppose this class is composed of `common_utils.TempDirMixin`
```

3. Save wav file

```
common_utils.save_wav(path, data)
```

4. Load wav file

```
common_utils.load_wav(path)
```

engineerchuan:
I would like to work on this.  Would we prefer doing this in one PR or okay to split across a 2-3 PR?

mthrok:
Hi @engineerchuan 

Thanks for signing up.

It is preferable to split PRs into ones simple enough to review and each PR works on one single aspect. For example

- deleting all the unused files is simple enough to be in one PR.
[...]

PR #790, opened 2020-07-16
Add LibriTTS dataset

jimchen90:
Add the LibriTTS dataset. 

The output `utterance_id` is the full name of audio files.
(for example `utterance_id` = '84_121123_000007_000001')

Related to #550  

[Notebook](https://github.com/jimchen90/TestRep/blob/master/libritts_test.ipynb)


test/test_datasets.py
```diff
@@ -45,6 +46,10 @@ def test_cmuarctic(self):
         data = CMUARCTIC(self.path)
         data[0]
 
+    def test_libritts(self):
```
mthrok, line 49: Can you make a separate test case? I see putting everything on the same `TestCase` is a bad pattern.

Also, since the quality of the audio file does not really matter, can you switch to generate all the test asset on-the-fly? (both audio and text) See the direction in https://github.com/pytorch/audio/issues/764.

jimchen90: Thanks for the suggestion. I tried to make a separate test case and generate the audio and text on-the-fly. It is based on #792 .

torchaudio/datasets/libritts.py
```diff
@@ -0,0 +1,128 @@
...
+                extract_archive(archive)
+
+        walker = walk_files(
+            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
+        )
+        self._walker = list(walker)
+
+    def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, int, int, str]:
+        fileid = self._walker[n]
+        return load_libritts_item(fileid, self._path, self._ext_audio, self._ext_original_txt, self._ext_normalized_txt)
```
vincentqb, line 125: nit: have you ran `black` on this file?

test/test_datasets.py
```diff
@@ -94,6 +95,7 @@ def setUpClass(cls):
         for label in cls.labels:
             filename = f'{"_".join(str(l) for l in label)}.wav'
             path = os.path.join(base_dir, filename)
+
```
vincentqb, line 98: nit: why is this changing?

torchaudio/datasets/libritts.py
```diff
@@ -0,0 +1,128 @@
...
+        int(speaker_id),
+        int(chapter_id),
+        utterance_id,
+    )
+
+
+class LIBRITTS(Dataset):
+    """
+    Create a Dataset for LibriTTS. Each item is a tuple of the form:
+    waveform, sample_rate, original_utterance, normalized_utterance, speaker_id, chapter_id, utterance_id
```
vincentqb, line 75: Since the waveform is normalized, there is no bits data to share in the data point. This looks good to me.

torchaudio/datasets/libritts.py
```diff
@@ -0,0 +1,131 @@
...
+        self._path = os.path.join(root, folder_in_archive)
+
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _CHECKSUMS.get(url, None)
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        walker = walk_files(
```
mthrok, line 115: Please do not use `walk_files` this returns files in unpredicted order. https://github.com/pytorch/audio/issues/794

torchaudio/datasets/libritts.py
jimchen90: I ran `black` to reformat it now.

test/test_datasets.py
jimchen90: I have changed it back.

torchaudio/datasets/libritts.py
jimchen90: Thanks for the information, do you have any suggestion instead of using `walk_files`?

test/test_datasets.py
```diff
@@ -110,5 +111,67 @@ def test_yesno(self):
...
+            normalized_text_filename = f'{"_".join(str(u) for u in utterance_id)}.normalized.txt'
+            path_normalized = os.path.join(file_dir, normalized_text_filename)
+            f = open(path_normalized, 'w')
+            f.write(cls.normalized_text)
+            f.close()
+
+    def test_libritts(self):
+        dataset = LIBRITTS(self.root_dir)
+        samples = list(dataset)
+        samples.sort(key=lambda s: s[4])
```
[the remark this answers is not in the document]
jimchen90: Thank you for the information. I will think about it. Do you have any suggest function instead of using `walk_files`

torchaudio/datasets/libritts.py
vincentqb: Good point. Since we are using `walk_files` in all other datasets, I'm ok with moving forward as it is for this pull request, and leaving the migration for a later time.

test/test_datasets.py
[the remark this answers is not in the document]
vincentqb: Good catch. Let's at least add a FIXME comment linking to #792. Is there something else in this snippet that should be flagged by #792?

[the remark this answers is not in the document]
mthrok: > Thank you for the information. I will think about it. Do you have any suggest function instead of using `walk_files`?

FYI: In https://github.com/pytorch/audio/pull/791 GTZAN dataset got rid of `walk_files` with very simple implementation to add pattern matching for file name patterns. If LibriTTS also has patterns in dataset, it should be checking the patterns, too. (I believe such pattern matching option should be available on utility function side, but that's separate topic)

test/test_datasets.py
```diff
@@ -110,5 +111,67 @@ def test_yesno(self):
...
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        base_dir = os.path.join(cls.root_dir, 'LibriTTS', 'train-clean-100')
+        for utterance_id in cls.utterance_ids:
+            filename = f'{"_".join(str(u) for u in utterance_id)}.wav'
+            file_dir = os.path.join(base_dir, str(utterance_id[0]), str(utterance_id[1]))
+            os.makedirs(file_dir, exist_ok=True)
+            path = os.path.join(file_dir, filename)
+
+            data = get_whitenoise(sample_rate=8000, duration=6, n_channels=1, dtype='int16')
```
mthrok, line 136: I just realized that I forgot to add `seed` to `get_whitenoise` function call in #792 and all the data generated are identical. Can you add `seed`?

```
for i, utterance_id in enumerate(cls.utterance_ids):
    ...
    data = get_whitenosie(..., seed=i)
```

test/test_datasets.py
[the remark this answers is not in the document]
vincentqb: > FYI: In #791 GTZAN dataset got rid of `walk_files` with very simple implementation to add pattern matching for file name patterns. If LibriTTS also has patterns in dataset, it should be checking the patterns, too. (I believe such pattern matching option should be available on utility function side, but that's separate topic)

Thanks for pointing this out :) However, I wouldn't replicate this particular solution here, until we have more general abstractions to do so. One of the strength of the dataset implementations that we currently have is how simple it is to replicate and extend to other cases.

test/test_datasets.py
jimchen90: The seed has been added.

test/test_datasets.py
```diff
@@ -110,5 +111,67 @@ def test_yesno(self):
...
+            f.close()
+
+    def test_libritts(self):
+        dataset = LIBRITTS(self.root_dir)
+        samples = list(dataset)
+        samples.sort(key=lambda s: s[4])
+
+        for i, (waveform,
+                sample_rate,
+                original_utterance,
```
zhizhengwu, line 159: original_utterance -> original_text

test/test_datasets.py
```diff
@@ -110,5 +111,67 @@ def test_yesno(self):
...
+                normalized_utterance,
```
zhizhengwu, line 160: normalized_utterance -> normalized_text

jimchen90: This name has been changed.

test/test_datasets.py
jimchen90: This name has been changed.

test/test_datasets.py
```diff
@@ -110,5 +111,67 @@ def test_yesno(self):
...
+        cls.root_dir = cls.get_base_temp_dir()
+        base_dir = os.path.join(cls.root_dir, 'LibriTTS', 'train-clean-100')
+        for i, utterance_id in enumerate(cls.utterance_ids):
+            filename = f'{"_".join(str(u) for u in utterance_id)}.wav'
+            file_dir = os.path.join(base_dir, str(utterance_id[0]), str(utterance_id[1]))
+            os.makedirs(file_dir, exist_ok=True)
+            path = os.path.join(file_dir, filename)
+
+            data = get_whitenoise(sample_rate=8000, duration=6, n_channels=1, dtype='int16', seed=i)
+            save_wav(path, data, 8000)
```
mthrok, line 137: @jimchen90 I am checking LIbrispeech `train-clean-100` but I do not see a file with 8kHz. Most of them seems 16000 Hz. Can you confirm?

jimchen90: You are right, I checked the published papers of [LibriSpeech](http://www.danielpovey.com/files/2015_icassp_librispeech.pdf) and [LibriTTS](https://arxiv.org/pdf/1904.02882.pdf) linked from their website. They are 16 kHz for LibriSpeech and 24 kHz for LibriTTS.

In abstract of LibriSpeech paper, it mentions ' The LibriSpeech corpus is derived from audiobooks that are part of the LibriVox project, and contains 1000 hours of speech sampled at 16 kHz.'

In abstract of LibriTTS paper, it mentions 'The released corpus consists of 585 hours of speech data at 24kHz sampling rate from 2,456 speakers and the corresponding texts.'

I will open a pull request to update this from 8000 to 24K.

mthrok:
Okay, I just moved the test script in #817. so make sure that you branch
from the latest upstream commits.
Also, please check the number of channels too.
You can also shorten the duration too (does not really matter though). 6
was picked for YESNO dataset as that's the expected length there.

moto 🛵

PR #792, opened 2020-07-16
Generate YESNO dataset on-the-fly for test

mthrok:
Part of #764


test/common_utils/backend_utils.py
```diff
@@ -29,8 +29,8 @@ def supports_mp3(backend):
 def set_audio_backend(backend):
     """Allow additional backend value, 'default'"""
     if backend == 'default':
-        if 'sox' in BACKENDS:
-            be = 'sox'
```
vincentqb, line 33: note: this is removing "sox" from "set_audio_backend"

vincentqb:
cc @jimchen90 for visibility

test/common_utils/backend_utils.py
mthrok: no, it's changing the default backend to be `sox_io` when it's available.

PR #797, opened 2020-07-17
Update variable names in wavernn model

jimchen90:
This is to update the variable names in [WaveRNN model](https://github.com/pytorch/audio/pull/735).

Related to #735

### DIFF
--- a/test/test_models.py
+++ b/test/test_models.py
@@ -42,17 +42,17 @@ class TestMelResNet(common_utils.TorchaudioTestCase):
         n_batch = 2
         n_time = 200
         n_freq = 100
-        n_output = 128
+        n_output_melresnet = 128
         n_res_block = 10
-        n_hidden = 128
+        n_hidden_resblock = 128
         kernel_size = 5
 
-        model = _MelResNet(n_res_block, n_freq, n_hidden, n_output, kernel_size)
+        model = _MelResNet(n_res_block, n_freq, n_hidden_resblock, n_output_melresnet, kernel_size)
 
         x = torch.rand(n_batch, n_freq, n_time)
         out = model(x)
 
-        assert out.size() == (n_batch, n_output, n_time - kernel_size + 1)
+        assert out.size() == (n_batch, n_output_melresnet, n_time - kernel_size + 1)
 
 
 class TestUpsampleNetwork(common_utils.TorchaudioTestCase):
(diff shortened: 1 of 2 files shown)


test/test_models.py
```diff
@@ -65,75 +65,80 @@ def test_waveform(self):
...
         x = torch.rand(n_batch, 1, hop_length * (n_time - kernel_size + 1))
         mels = torch.rand(n_batch, 1, n_freq, n_time)
         out = model(x, mels)
 
-        assert out.size() == (n_batch, 1, hop_length * (n_time - kernel_size + 1), 2 ** n_bits)
+        assert out.size() == (n_batch, 1, hop_length * (n_time - kernel_size + 1), n_classes)
 
     def test_mol(self):
-        """Validate the output dimensions of a _WaveRNN model in mol mode.
+        """Validate the output dimensions of a _WaveRNN model in mol loss.
```
vincentqb, line 122: nit: "with mo loss"

torchaudio/models/_wavernn.py
```diff
@@ -205,54 +205,54 @@ class _WaveRNN(nn.Module):
...
-        if self.mode == 'waveform':
-            self.n_classes = 2 ** n_bits
-        elif self.mode == 'mol':
+        if self.loss == 'crossentropy':
+            self.n_classes = n_classes
+        elif self.loss == 'mol':
             self.n_classes = 30
         else:
-            raise ValueError(f"Expected mode: `waveform` or `mol`, but found {self.mode}")
+            raise ValueError(f"Expected loss: `crossentropy` or `mol`, but found {self.loss}")
```
vincentqb, line 252: We discussed offline that we would remove the "loss/mode" parameter for WaveRNN. Let's do it as part of this PR.

torchaudio/models/_wavernn.py
```diff
@@ -50,8 +50,8 @@ class _MelResNet(nn.Module):
     Args:
         n_res_block: the number of ResBlock in stack (default=10)
         n_freq: the number of bins in a spectrogram (default=128)
-        n_hidden: the number of hidden dimensions (default=128)
-        n_output: the number of output dimensions (default=128)
+        n_hidden_resblock: the number of hidden dimensions of resblock (default=128)
+        n_output_melresnet: the number of output dimensions of melresnet (default=128)
```
vincentqb, line 54: These are clear as they are for a MelResBlock. Let's keep `n_hidden` and `n_output` here, but add "of resblock" and "of melresnet" in the description as you have done here.

torchaudio/models/_wavernn.py
jimchen90: Yes, I have removed it. I also removed the `sample_rate` parameter since it is not needed.

torchaudio/models/_wavernn.py
jimchen90: Yes, I keep them as before.

test/test_models.py
jimchen90: This test has been removed.

PR #806, opened 2020-07-22
Add smoke tests to sox_io and sox_effects

mthrok:
Currently all the tests in `sox_io_backend` and `sox_effects` (for new SoX effects implementation) requires additional `sox`, and this prevents running test in environment where `sox` command is not available even though `torchaudio` extension is available (such as fb internal). This PR adds smoke tests for these modules, which just runs functions to see if they do not crash.


mthrok:
thanks

PR #811, opened 2020-07-22
[BC Breaking] Split `list_formats()` for read and write

mthrok:
Turned out that `libsox` in fbcode only has reader (`libmad`) and not writer (`libmp3lame`), therefor the smoke test added in #806 cannot run as is.
I tried to add skip condition using `sox_utils.list_formats` utility but this function does not care if a format handler has writer or not. (cf. `libsox` does not implement writer for `opus` too)

From the user point of view, and debugging the installation, this is confusing and not very convenient, so this PR splits `torchaudio.utils.sox_utils.list_formats()` into `torchaudio.utils.sox_utils.list_read_formats()` and `torchaudio.utils.sox_utils.list_write_formats()`.

### DIFF
--- a/test/sox_io_backend/smoke_test.py
+++ b/test/sox_io_backend/smoke_test.py
@@ -1,6 +1,9 @@
 import itertools
+import unittest
 
+from torchaudio.utils import sox_utils
 from torchaudio.backend import sox_io_backend
+from torchaudio._internal.module_utils import is_module_available
 from parameterized import parameterized
 
 from ..common_utils import (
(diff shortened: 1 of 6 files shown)


vincentqb:
> From the user point of view, and debugging the installation, this is confusing and not very convenient, so this PR splits `torchaudio.utils.sox_utils.list_formats()` into `torchaudio.utils.sox_utils.list_read_formats()` and `torchaudio.utils.sox_utils.list_write_formats()`.

Was `list_formats` returning the union or intersection of `list_read_formats` and `list_write_formats`?

mthrok:
> > From the user point of view, and debugging the installation, this is confusing and not very convenient, so this PR splits `torchaudio.utils.sox_utils.list_formats()` into `torchaudio.utils.sox_utils.list_read_formats()` and `torchaudio.utils.sox_utils.list_write_formats()`.
> 
> Was `list_formats` returning the union or intersection of `list_read_formats` and `list_write_formats`?

Union, and that's why `list_formats` cannot be used for skipping smoke test. It cannot tell when only one of reader/writer is missing.

mthrok:
thanks

mthrok:
> We could maintain `list_formats` to avoid BC-breaking, but this would add confusion. LGTM.

Yeah, and I do not think anyone would be using `list_formats` in their ML application anyways, so I think it's fine.

PR #812, opened 2020-07-22
Add VCTK_092 dataset

Abhi011999:
- Updated dataset URL's
- Updated the zip file checksum
- Fixed URL path parsing for the updated link
- Updated audio folder and audio extension names

The last URL returned a 404 HTTP not found error as the dataset was moved to a new link.

Verified the changes by loading the dataset into dataloader and looping over every file.
Tested by training some epochs.


torchaudio/datasets/vctk.py
```diff
@@ -26,10 +26,10 @@ def load_vctk_item(fileid: str,
                    folder_audio: str,
                    folder_txt: str,
                    downsample: bool = False) -> Tuple[Tensor, int, str, str, str]:
-    speaker_id, utterance_id = fileid.split("_")
+    speaker_id, utterance_id, _ = fileid.split("_")
 
     # Read text
-    file_txt = os.path.join(path, folder_txt, speaker_id, fileid + ext_txt)
```
mthrok, line 32: Slicing string like this is harder to read what is changed. Is there a simple way to do split string? If there is not, can you add a comment?

torchaudio/datasets/vctk.py
```diff
@@ -11,11 +11,11 @@
     walk_files
 )
 
-URL = "http://homepages.inf.ed.ac.uk/jyamagis/release/VCTK-Corpus.tar.gz"
+URL = "https://datashare.is.ed.ac.uk/bitstream/handle/10283/3443/VCTK-Corpus-0.92.zip"
```
mthrok, line 14: From https://datashare.is.ed.ac.uk/handle/10283/3443, there is also a download link to http://datashare.is.ed.ac.uk/download/DS_10283_3443.zip . Do you know the difference?

I am downloading it now, but it takes a while.

torchaudio/datasets/vctk.py
```diff
@@ -53,15 +53,14 @@ class VCTK(Dataset):
...
     Folder `p315` will be ignored due to the non-existent corresponding text files.
     For more information about the dataset visit: https://datashare.is.ed.ac.uk/handle/10283/3443
     """
 
     _folder_txt = "txt"
-    _folder_audio = "wav48"
+    _folder_audio = "wav48_silence_trimmed"
     _ext_txt = ".txt"
-    _ext_audio = ".wav"
+    _ext_audio = ".flac"
```
mthrok, line 63: Assuming that this is correct. I wonder how we should approach backward compatibility.

If we just update this, people who have already downloaded the previous format of the dataset would face error. Ideally we do not want to disturb and force those people to re-download it. Do you have a thought?

torchaudio/datasets/vctk.py
Abhi011999: Added appropriate comments

torchaudio/datasets/vctk.py
Abhi011999: They are both the same, `http://datashare.is.ed.ac.uk/download/DS_10283_3443.zip` contains an extra License and readme files.

torchaudio/datasets/vctk.py
Abhi011999: Added backward compatibility for older versions of datasets. Although it's not the best and clean implementation, it will work in 99.9% scenarios.

Abhi011999:
Hi @mthrok ! Thanks for reviewing this.

I made some changes. Hope these help.

Abhi011999:
Fixed everything. @mthrok you can squash & merge when ready.

mthrok:
Hi @Abhi011999 

Sorry for the late reply. We were discussing internal how we should approach this problem.
We decided to keep the original VCTK as is (but raise an error when `download=True` with the message to use a new version.), then add the new version, (say `VCTK_V2`).

Could you move your changes to `VCTK_V2`? You can either copy-paste the original `VCTK` or subclass `VCTK` (if that works and is simpler)

Abhi011999:
@mthrok An optional boolean argument can be added in the existing `VCTK` class that identifies the old and the new dataset with the default value set to `False`. That can work?

mthrok:
> @mthrok An optional boolean argument can be added in the existing `VCTK` class that identifies the old and the new dataset with the default value set to `False`. That can work?

It would, but we do not know if there will be another dataset change like this (though chances are small),
and boolean is not future proof.
Also, the previous version is now generally inaccessible, yet downloading the new one takes a bit time.
So rather than mixing the code for dataset that are only useful for those who has cached data and the code 
that general public can use, we concluded it's cleaner to separate them.
Maybe we should call the new dataset `VCTK_092`.

Abhi011999:
> > @mthrok An optional boolean argument can be added in the existing `VCTK` class that identifies the old and the new dataset with the default value set to `False`. That can work?
> 
> It would, but we do not know if there will be another dataset change like this (though chances are small),
> and boolean is not future proof.
> Also, the previous version is now generally inaccessible, yet downloading the new one takes a bit time.
> So rather than mixing the code for dataset that are only useful for those who has cached data and the code
> that general public can use, we concluded it's cleaner to separate them.
> Maybe we should call the new dataset `VCTK_092`.

That'll work! I will get to work now. 👍

Abhi011999:
@mthrok So the extra comments are not necessary or should I just leave them?

mthrok:
> @mthrok So the extra comments are not necessary or should I just leave them?

Which one do you mean by "the extra comments" ??


Also, can you put the new class `VCTK_092` into the same module, `vctk.py`?
That way, you do not need to duplicate `load_vctk_item` function, right?

mthrok:
Also I am seeing CI job errors that should have been resolved by https://github.com/pytorch/audio/pull/850, can you rebase your work?

Abhi011999:
> That way, you do not need to duplicate `load_vctk_item` function, right?

No, some lines are different for the new version.

> Which one do you mean by "the extra comments" ??

I mean the comments I put when I made changes to the original `vctk.py`.

> can you rebase your work?

will do, just want a final confirmation.

mthrok:
> > That way, you do not need to duplicate `load_vctk_item` function, right?
> 
> No, some lines are different for the new version.

I see. In that case, could you make it to a method?
That way it's easier to allow users to customize the loading behavior if necessary.

> > Which one do you mean by "the extra comments" ??
> 
> I mean the comments I put when I made changes to the original `vctk.py`.

I can no longer see the original commit. Could you describe it?

Also can you put the new class in the same `vctk.py` module?
I think adding a new submodule for a new version of existing dataset might get 
unmanageable at some point if there are many updates.

Once mode thing, can you change the original VCTK class so that it will raise 
an error if `download=True` with an error message that explains the dataset is
no longer available and users should use the new class you are adding.

torchaudio/datasets/vctk_092.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def __init__(self,
+                 root: str,
+                 url: str = URL,
+                 folder_in_archive: str = FOLDER_IN_ARCHIVE,
+                 download: bool = False,
+                 downsample: bool = False,
+                 transform: Any = None,
+                 target_transform: Any = None) -> None:
+
```
mthrok, line 83: Could you remove `downsample`, `transform` and `target_transform` parameters? Since we are adding a new class, we should take advantage of the opportunity not to carry out some legacy customs.

Abhi011999: can I add a new param - `mic_id` as there are recordings using a new mic from `0.92` version?

mthrok: I think you are referring this one

> 1: Audio files recorded via DPA 4035 (mic1) have very low frequency noises and therefore we added audio files recorded via MKH 800 (mic2)

I think that's reasonable to add the parameter. What is your thought on the behavior of `mic_id` ?
Assuming `mic_id` is integer type, I imagine giving either `1` or `2` would only retrieve the corresponding audio. Then I wonder if there is a use-case where one wants to randomize or get both of them. What do you think?

torchaudio/datasets/vctk_092.py
```diff
@@ -0,0 +1,150 @@
...
+                "`target_transform=True` to suppress this warning."
+            )
+
+        self.downsample = downsample
+        self.transform = transform
+        self.target_transform = target_transform
+
+        archive = os.path.basename(url)
+
+        # url contains un-wanted '?sequence=2&isAllowed=y'
```
mthrok, line 104: > # url contains un-wanted '?sequence=2&isAllowed=y'
Does this still apply?

Abhi011999: yes and we need to slice it

torchaudio/datasets/vctk_092.py
Abhi011999: It depends on several research use cases, if someone wants to do some research on only old audio with the low-frequency noises, it can be helpful. Besides users can conveniently ignore `mic_id`, currently, it's being ignored by default when slicing the filename in load item method.

Abhi011999: and one more thing, should I remove the deprecated params from the old `VCTK` class and `load_vctk_item` method too?

mthrok: > Besides users can conveniently ignore mic_id, currently, it's being ignored by default when slicing the filename in load item method.

My understanding is that microphone ID was ignored in the original `VCTK` implementation because there was only one ID there. Now that with 0.92 update, there are two microphones, and we would like to give users easy control and intuitive behavior. Randomizing microphone ID would require another parameter for random seed for the sake of reproducibility, and I think we can revisit that later.

Therefore, I think, for the sake of simplicity here, we can just allow `mic_id:int=1 or 2`, which fixes the micro phone ID. What do you think?

torchaudio/datasets/vctk_092.py
mthrok: Where does it come from?

I mean the default value is `https://datashare.is.ed.ac.uk/bitstream/handle/10283/3443/VCTK-Corpus-0.92.zip`, which does not contain any URL option.

mthrok: Oh I see, this is the case for custom URL provided by a user.

If this concerns only for the case where user provides such URL with options, can you update the comment
> custom `url` may contain options like '?sequence=2&isAllowed=y'

torchaudio/datasets/vctk_092.py
mthrok: > and one more thing, should I remove the deprecated params from the old `VCTK` class and `load_vctk_item` method too?

Let's keep the changes to the original `VCTK` minimum. So I suggest the only change we are going to make to the original `VCTK` class in the context of this PR will be raising error when `download=True`.

Abhi011999: > and one more thing, should I remove the deprecated params from the old `VCTK` class and `load_vctk_item` method too?

@mthrok

torchaudio/datasets/vctk_092.py
Abhi011999: will do

torchaudio/datasets/vctk_092.py
Abhi011999: okay, will push the changes after testing

Abhi011999:
I am a little out of knowledge here but why are the tests failing exactly?

mthrok:
> I am a little out of knowledge here but why are the tests failing exactly?
> 
> EDIT: Is my branch `patch-1` in conflict here?

It's not conflict but there was a breaking change in PyTorch core and we added the change for that in #850 .
I see that your branch is based off of #00cc000, which does not have this fix. So if you fetch the latest master branch of `pytorch/audio` and rebase your changes on it, then the tests should pass.

Abhi011999:
I see random tests failing and passing. Seems like I have to pass the "torch" to you now @mthrok 👀

mthrok:
> I see random tests failing and passing. Seems like I have to pass the "torch" to you now @mthrok 👀

lol nice one. Yeah, unfortunately we are hit by another upstream issue. https://github.com/pytorch/pytorch/pull/43050
CI should be working fine tomorrow.

Meanwhile there are couple of improvements on the current code base. If you have time, could you give some thoughts? This will make the implementation of `VCTK_092` much better.

1. Making `load_vctk_092_item` function into a method of `VCTK_092`.
    Putting `load_vctk_092_item` as a plain function makes it difficult to customize the behavior on user code. If a user wants to change a bit of behavior, the user has to re-write the whole class. 
2. Make the return value typed.
    In general, when returning multiple objects from a function, it a good idea to introduce micro protocol with `namedtuple`. This way the access to the individual item gets cleaner and it becomes future-prof (adding a new item to be returned is backward compatible.) Check out [my example here](https://github.com/mthrok/audio/blob/wsj0-mix-dataset/torchaudio/datasets/wsj0mix.py#L10).
3. Wrap `torchaudio.load` as another method of `VCTK_092`.
    Similar to 1, this will allow user to customize the loading behavior, like disabling normalization or swapping channels. See example [here](https://github.com/mthrok/audio/blob/c103e39ee4a811f52d26d4f30a0e04e1d52f81ce/torchaudio/datasets/wsj0mix.py#L22-L23)
4. Get rid of `walk_files` function.
    Once the root directory is located, we have the idea of expected directory structure, so there is no need to traverse directories as if it's unknown. `walk_files` function can do harm by traversing unrelated directories, but it does not bring any benefit.

    ```
    speaker_ids = sorted(d for d in os.listdir(txt_dir) if os.path.isdir(os.path.join(txt_dir, d)))
    for speaker_id in speaker_ids:
        utterance_dir = os.path.join(txt_dir, speaker_id)
        speech_dir = os.path.join(audio_dir, speaker_id)
        for utterance_file in sorted(f for f in os.listdir(utterance_dir) if f.endswith("txt")):
            utterance_id = os.path.splitext(utterance_file)[0]
            utterance_path = os.path.join(utterance_dir, utterance_id)
            audio_path = os.path.join(audio_dir, f'{utterance_id}_{mic_id}.{self.audio_ext}'
    ```
    
    This way, you do not need attributes like `_except_folder`. and the directory traversal is minimum.

Abhi011999:
@mthrok 

So I created a test script for the new implementation based on your above approach **without** `walk_files` and guess what?

Half of the `mic1` audio files have been removed from `p362`, but there `occurrence` is present in the `txt` folder. Below is the full list with IDs -
https://pastebin.com/p02Fyby7

My take here is that after the total path is calculated of every possible audio file, it performs a validation check if the file exists or not using `os.path.exists` method and then add it to a list.
What do you think?

Note - Everything else has been taken care of, only this remains.

mthrok:
> @mthrok
> 
> So I created a test script for the new implementation based on your above approach **without** `walk_files` and guess what?
> 
> Half of the `mic1` audio files have been removed from `p362`, but there `occurrence` is present in the `txt` folder. Below is the full list with IDs -
> https://pastebin.com/p02Fyby7
> 
> This is not even mentioned in the logs.
> 
> My take here is that after the total path is calculated of every possible audio file, it performs a validation check if the file exists or not using `os.path.exists` method and then add it to a list.
> What do you think?
> 
> Note - Everything else has been taken care of, only this remains.

Hi @Abhi011999 

That's an interesting finding, and thanks for reporting. I think agree with your approach. We can skip if the corresponding audio file for a given text is not present. There is nothing much we can do about the missing files, but maybe it's a good idea to add a note about it in the docstring.

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +166,67 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+    _folder_txt = "txt"
+    _folder_audio = "wav48_silence_trimmed"
+    _ext_txt = ".txt"
+    _ext_audio = ".flac"
+    _except_folder = "p315"
+
+    def __init__(self,
+                 root: str,
+                 url: str = URL,
+                 folder_in_archive: str = FOLDER_IN_ARCHIVE_092,
```
mthrok, line 188: Can you also remove this `folder_in_archive` from the signature? I do not see a reason why this has to be a variable or the benefit of it. It's only used one place and if this is something different it is no longer a VCTK 0.92 dataset.

Abhi011999:
Hi, @mthrok. As I have updated with the required changes, I also made some performance comparisons with and without the `walk_files` implementation and here are some insights -

- With `walk_files`
  - `__init__()` takes 0.5 seconds approx. to walk over files.
  - Retrieving one sample takes about 0.05 seconds approx.
  - Iterating over the entire dataset takes about 18 min 26 secs.
  - It can wrongly walk over any file in any nested directory with the same extensions.
  - We need to explicitly specify which folders or files to omit.

- Without `walk_files`
  - `__init__()` takes 7 seconds approx. to walk over files.
  - Retrieving one sample takes about 0.048 seconds approx.
  - Iterating over the entire dataset takes about 18 min 26 secs.
  - This implementation is future proof for additional upcoming updates.

The conclusion that we can draw from the above stats is that we can keep the new implementation without `walk_files` as its future proof and further relevant changes can be made easily in accordance with the upcoming updates for the dataset.

That's my take. Would love to hear what you think.

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +140,99 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+
+        if not os.path.isdir(self._path):
+            raise RuntimeError(
+                "Dataset not found. Please use `download=True` to download it."
+            )
+
+        # Extracting speaker IDs from the folder structure
+        self._speaker_ids = sorted(d for d in os.listdir(self._txt_dir)
+                                   if os.path.isdir(os.path.join(self._txt_dir, d)))
+        self._audio_ids = []
```
mthrok, line 188: I think storing `[(speaker_id, utterance_id), ...]` is enough.

If the dataset instance has `mic_id` attribute, then we can build the necessary path on With both `(speaker_id, utterance_id)`.

I think `load_sample(self, speaker_id, utterance_id, mic_id)` will give a more intuitive UX.

mthrok:
> Hi, @mthrok. As I have updated with the required changes, I also made some performance comparisons with and without the `walk_files` implementation and here are some insights -
> 
> * With `walk_files`
>   
>   * `__init__()` takes 0.5 seconds approx. to walk over files.
>   * Retrieving one sample takes about 0.05 seconds approx.
>   * Iterating over the entire dataset takes about 18 min 26 secs.
>   * It can wrongly walk over any file in any nested directory with the same extensions.
>   * We need to explicitly specify which folders or files to omit.
> * Without `walk_files`
>   
>   * `__init__()` takes 7 seconds approx. to walk over files.
>   * Retrieving one sample takes about 0.048 seconds approx.
>   * Iterating over the entire dataset takes about 18 min 26 secs.
>   * This implementation is future proof for additional upcoming updates.
> 
> The conclusion that we can draw from the above stats is that we can keep the new implementation without `walk_files` as its future proof and further relevant changes can be made easily in accordance with the upcoming updates for the dataset.
> 
> That's my take. Would love to hear what you think.

Thanks for doing the benchmark. The only concern of the new implementation is 

> * `__init__()` takes 7 seconds approx. to walk over files.

I bet this is coming from the `os.path.exists` calls. We can employ our knowledges of the dataset to reduce this. We know two facts;
 - `p315` is missing all the transcripts.
 - `p362` is missing a bunch of audio files.

I previously suggested to look at transcript directory because transcripts are common to microphone ID, but this ended up force us to check the existing of audio files in `p362`.
1. Opposite way is to look at audio directories (except `p315`) first, build the corresponding transcript path, because that way we know that as long as audio file is found, the corresponding transcript exist so we do not need to perform file existence check.
2. Since `p362` is the only speaker without speech files, we can perform the file existence check only on this directory. We know that all the other speakers/utterance combinations have valid mapping from transcript to audios, so we do not need to perform file existence check on them.

I would push for `2.` because that way changes from the current implementation is smaller and also treating exceptional directory as exception (i.e. `if speaker_id == "p362":...`) will make the code express the logic we are designing better.

What do you think?

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +140,99 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+    Folder `p315` will be ignored due to the non-existent corresponding text files.
+    For more information about the dataset visit: https://datashare.is.ed.ac.uk/handle/10283/3443
+    """
+
+    _folder_txt = "txt"
+    _folder_audio = "wav48_silence_trimmed"
+    _ext_txt = ".txt"
+    _ext_audio = ".flac"
+    _mic_1 = "mic1"
+    _mic_2 = "mic2"
```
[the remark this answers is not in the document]
Abhi011999: Will change it.
About the `mic_id`, I will hardcode them as we know the structure of the dataset.

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +140,99 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+
+    def __init__(self,
+                 root: str,
+                 url: str = URL,
+                 download: bool = False) -> None:
+
+        # Custom url may contain options like '?sequence=2&isAllowed=y'
+        # Slicing it off to extract only 'VCTK-Corpus-0.92.zip'
+        archive = os.path.join(root, os.path.basename(url)[:20])
```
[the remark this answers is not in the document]
Abhi011999: I was thinking the same, will change it.

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +140,99 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+
+        self._path = os.path.join(root, FOLDER_IN_ARCHIVE_092)
```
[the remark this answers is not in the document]
Abhi011999: okay

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +140,99 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+        different parameters required by the user.
+        """
+        for speaker_id in self._speaker_ids:
+            utterance_dir = os.path.join(self._txt_dir, speaker_id)
+            for utterance_file in sorted(f for f in os.listdir(utterance_dir) if f.endswith(self._ext_txt)):
+                utterance_id = os.path.splitext(utterance_file)[0]
+                audio_path_mic1 = os.path.join(self._audio_dir, speaker_id,
+                                               f'{utterance_id}_{self._mic_1}{self._ext_audio}')
+                audio_path_mic2 = os.path.join(self._audio_dir, speaker_id,
+                                               f'{utterance_id}_{self._mic_2}{self._ext_audio}')
```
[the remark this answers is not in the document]
Abhi011999: Okay, will do the change.

torchaudio/datasets/vctk.py
Abhi011999: Explain it again pls I didn't get it.

Abhi011999:
I think it will work, I will do the suggested changes, run the benchmark, and then push.

Abhi011999:
Hi, @mthrok.

`mic_id` optional param is added with the default value of `mic2` as its the new and better microphone.

Performance is good now -
`__init__()` takes 0.33 seconds approx now for both `mic2` and `mic1`.

I think this can't get any more straight forward than this. Let me know what you think.

torchaudio/datasets/vctk.py
mthrok: What I meant is that currently, `self.sample_ids` is a list of string `<SPEAKER>_<UTTERANCE_KEY>_<MIC>` (here, I separate speaker and utterance for the purpose of clarity. I am aware that the variable `utterance_id` in the implementation includes `speaker_id` too), but now that microphone ID can be stored as `self.mic_id = mic_id`, we do not need to store `mic_id` inside of `self.sample_ids`.

Plus, rather than concatenating variables into one string (as `f{utterance_id}_{mic_id}` in the code) then splitting it again in `load_sample` method, storing the sample IDs as a tuple of `(<SPEAKER>, <UTTERANCE>)`, and passing it alongside with `mic_id`, to `load_sample` method will give even better separation of logic, in my opinion.

so what I am suggesting is something like

In `__init__`
```
    self.mic_id = mic_id
    ...
                 if speaker_id == 'p280' and mic_id == 'mic2':
                     break
                 if speaker_id == 'p362' and not os.path.isfile(audio_path_mic):  # <- Also I think flipping this if clause to use break will improve the readability of logic
                     break
                 self._sample_ids.append(utterance_id.split('_'))
```

In `__get_item__`
```
speaker_id, utterance_id = self.sample_ids[n]
self.load_item(speaker_id, utterance_id, self.mic_id)
```

This way the method signature of `load_item` express it better that 3 IDs are necessary and sufficient to locate one data sample from VCTK dataset.

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +139,93 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+                                              f'{utterance_id}_{mic_id}.flac')
+                if speaker_id == 'p280' and mic_id == 'mic2':
+                    break
+                elif speaker_id == 'p362':
+                    if os.path.isfile(audio_path_mic):
+                        self._sample_ids.append(f'{utterance_id}_{mic_id}')
+                else:
+                    self._sample_ids.append(f'{utterance_id}_{mic_id}')
+
+    def load_text(self, file_path) -> str:
```
mthrok, line 204: Could you append this method name with underscore `_` so as to tell this is not expected to be used publicly??

torchaudio/datasets/vctk.py
```diff
@@ -136,3 +139,93 @@ def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
...
+        with open(file_path) as file_path:
+            return file_path.readlines()[0]
+
+    def load_audio(self, file_path) -> Tuple[Tensor, int]:
```
mthrok, line 208: Could you append this method name with underscore `_` so as to tell this is not expected to be used publicly??

mthrok:
> Hi, @mthrok.
> 
> `mic_id` optional param is added with the default value of `mic2` as its the new and better microphone.
> 
> Performance is good now -
> `__init__()` takes 0.33 seconds approx now for both `mic2` and `mic1`.
> 
> Also `mic2` audio files are missing in `p280` so added an exception for that too.
> 
> I think this can't get any more straight forward than this. Let me know what you think.

Hi @Abhi011999 

Thanks for sticking with all the review process and prompt replies. Now the code is in super good shape, and I really enjoy working with you.
The current code is really great. I responded to your comments. Let me know what you think, (and also let me know if you cannot spend any more time on this. I understand you have your thing too, and I will wrap it up in that case.)

torchaudio/datasets/vctk.py
Abhi011999: Done

torchaudio/datasets/vctk.py
Abhi011999: Done

torchaudio/datasets/vctk.py
Abhi011999: It will be `continue` in second `if` condition not `break` as we are not trying to skip the entire speaker if only one file doesn't exist. All else has been taken care of.

Abhi011999:
> Thanks for sticking with all the review process and prompt replies.

My pleasure.

> I really enjoy working with you.

It was really nice learning things from you.

> Let me know what you think, (and also let me know if you cannot spend any more time on this. I understand you have your thing too, and I will wrap it up in that case.)

I have to say that this is the first time I am working so in-depth with a PR and a reviewer. It's an awesome experience learning from the community and simultaneously trying to improve it. In the future, I am looking forward to donating more of my time to this awesome repository.

mthrok:
> EDIT: I also see you are formatting all the code with `black` as in #854, should I format it that way or leave it?

If you can only apply to the new class (`VCTK_092`) and leave the `VCTK` class as-is, then that will be nice.

mthrok:
Oh I forgot to mention but you may want to add the new class to https://github.com/pytorch/audio/blob/master/docs/source/datasets.rst

PR #814, opened 2020-07-22
Make directory traversal order alphabetical and breadth-first

### LINKED ISSUE #794: Dataset traversal cannot be deterministic
[Many dataset implementations use `walk_files` function](https://github.com/pytorch/audio/search?q=walk_files&unscoped_q=walk_files) which uses `os.walk` which does not give us choice to make directory traversal in deterministic order.

This makes it harder to test dataset and causes issue on reproducibility. Dataset should have an option to be deterministic.

mthrok:
Resolves #794


mthrok:
thanks

Issue #800, opened 2020-07-19
🚀 Feature Request: Loading audio data from BytesIO or memory


antimora:
## 🚀 Feature
The load API does not support loading audio bytes from the memory. It would a great addition to be able to load file like object, e.g. BytesIO. This is would be similar to SoundFile's read function (https://github.com/bastibe/SoundFile/blob/master/soundfile.py#L170)

## Motivation
This addition will support a use case for reading audio as blobs directly from DB instead writing the files locally first.

## Pitch
Without this feature, torchaudio.load is not useful for users who load files from DB and would love to use torchaudio for all audio operations. 

## Alternatives
SoundFile supports loading from bytes but currently does not support MP3 files. CommonVoice's audio files are saved in MP3, which requires to convert to FLAC or WAV before training.
```
waveform, samplerate = sf.read(file=io.BytesIO(audio_bytes), dtype='float32')
```

mthrok:
Hi @antimora  (and @faroit ) cc @vincentqb 

Thanks for bring this up. I gave some thoughts to this and did quick research. Let me start with stating where we stand.

- `torchaudio` does not implement encode/decode, instead it uses external libraries for this.

We use `libsox` or `PySoundFile` for this.

- We cannot pass file object to `libsox`.

I do not see a way to pass file object (`FILE*`) to `libsox` or extract `FILE*` from file-like object on Python 3. I looked into a way to pass data as stream, but [this](https://stackoverflow.com/a/54875668) says that it's not portable on macOS or Windows.

- `PySoundFile` (or `libsndfile`) does not support MP3.

There is nothing much we can do about it.

Following these, I looked at some Python solution for MP3, and it seems that [miniaudio](https://pypi.org/project/miniaudio/) supports in-memory MP3 decoding.

I did not use this library, but if this library provides MP3 decoding, then one should be able to convert it to (via NumPy array) torch Tensor type.

As of adding such functionality to `torchaudio`, maybe we can add it as a new backend, though I cannot promise. We will need to discuss interface/design if we do so.

However at this moment, we, `torchaudio` team does not have much bandwidth to actively work on this.

If someone can confirm that `miniaudio` approach mentioned above works, then we can discuss the possibility to add it to `torchaudio`.

What do you think?

antimora:
@mthrok , thank you very much for you detailed quick response. This is very helpful.

I agree with you regarding the challenges and limitations of currently used back-ends.

[...]

PR #816, opened 2020-07-23
Add sample_rate conversion option to `sox_io_backend.load`

mthrok:
This PR adds `sample_rate` option to `sox_io_backend.load` function which performs resampling when loading audio data from file.

Inspired by [this comment](https://github.com/pytorch/audio/issues/800#issuecomment-661661511)

~On Python interface this change is backward compatible, but Torchscript registration schema is changed, (see `register.cpp`), therefore previously dumped Torchscript object that includes call to `sox_io_backend.load` will stop working. -> BC breaking.~
~Added backward compatibility. TODO: test -> https://github.com/pytorch/audio/pull/838~

### DIFF
--- a/torchaudio/csrc/sox_io.cpp
+++ b/torchaudio/csrc/sox_io.cpp
@@ -52,6 +52,15 @@ c10::intrusive_ptr<TensorSignal> load_audio_file(
     const int64_t frame_offset,
     const int64_t num_frames,
     const bool normalize,
+    const bool channels_first) {
+  return load_audio_file_v1(path, frame_offset, num_frames, channels_first, -1);
+}
+
+c10::intrusive_ptr<TensorSignal> load_audio_file_v1(
+    const std::string& path,
+    const int64_t frame_offset,
+    const int64_t num_frames,
+    const bool normalize,
     const bool channels_first,
     const int64_t sample_rate) {
   if (frame_offset < 0) {
--- a/torchaudio/csrc/sox_io.cpp
+++ b/torchaudio/csrc/sox_io.cpp
@@ -52,15 +52,6 @@ c10::intrusive_ptr<TensorSignal> load_audio_file(
     const int64_t frame_offset,
     const int64_t num_frames,
     const bool normalize,
-    const bool channels_first) {
-  return load_audio_file_v1(path, frame_offset, num_frames, channels_first, -1);
-}
-
-c10::intrusive_ptr<TensorSignal> load_audio_file_v1(
-    const std::string& path,
-    const int64_t frame_offset,
-    const int64_t num_frames,
-    const bool normalize,
     const bool channels_first,
     const int64_t sample_rate) {
   if (frame_offset < 0) {
--- a/torchaudio/csrc/sox_io.cpp
+++ b/torchaudio/csrc/sox_io.cpp
@@ -53,6 +53,16 @@ c10::intrusive_ptr<TensorSignal> load_audio_file(
     const int64_t num_frames,
     const bool normalize,
     const bool channels_first) {
+  return load_audio_file_v1(path, frame_offset, num_frames, channels_first, -1);
+}
+
+c10::intrusive_ptr<TensorSignal> load_audio_file_v1(
+    const std::string& path,
+    const int64_t frame_offset,
+    const int64_t num_frames,
+    const bool normalize,
+    const bool channels_first,
+    const int64_t sample_rate) {
   if (frame_offset < 0) {
     throw std::runtime_error(
         "Invalid argument: frame_offset must be non-negative.");
@@ -61,8 +71,16 @@ c10::intrusive_ptr<TensorSignal> load_audio_file(
     throw std::runtime_error(
         "Invalid argument: num_frames must be -1 or greater than 0.");
   }
+  if (sample_rate == 0 || sample_rate < -1) {
+    throw std::runtime_error(
+        "Invalid argument: sample_rate must be -1 or greater than 0.");
+  }
 
   std::vector<std::vector<std::string>> effects;
+  if (sample_rate != -1) {
+    effects.emplace_back(
+        std::vector<std::string>{"rate", std::to_string(sample_rate)});
+  }
   if (num_frames != -1) {
     std::ostringstream offset, frames;
     offset << frame_offset << "s";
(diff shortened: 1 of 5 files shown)


torchaudio/csrc/sox_io.cpp
```diff
@@ -61,8 +71,16 @@ c10::intrusive_ptr<TensorSignal> load_audio_file(
...
   }
+  if (sample_rate == 0 || sample_rate < -1) {
+    throw std::runtime_error(
+        "Invalid argument: sample_rate must be -1 or greater than 0.");
+  }
 
   std::vector<std::vector<std::string>> effects;
+  if (sample_rate != -1) {
+    effects.emplace_back(
+        std::vector<std::string>{"rate", std::to_string(sample_rate)});
```
vincentqb, line 82: So this is adding to the sox_effects chain a resampling transform, right? I feel the user could want many other transform, and that would add many flags. Could it work instead to point out that sox_effects_chain can combine those operations already?

mthrok: The underlying implementation is same (which is why we can add sample rate conversion to load function), but the intended use of frontend functions are different. If users want to use filtering they can certainly do so by using sox effects. This is for convenience, and not for compatibility. Changing sampling rate is such a popular use case, like using the same dataset to train ASR models for different environments (realtime vs batch or on-device vs server etc...). So having convenient and fast way to change sampling rate is valuable.

vincentqb: Users also have `torchaudio.transforms.Resample` :) Are any of these two (`Resample` and `sox_effects`) measurably slower than this implementation?

I'm not keen on adding parameters like these to load functions when there are already simple ways of doing so. On top of adding parameters to the signature, this also does add layers of backward-compatibility etc to the codebase.

How about other backends like soundfile?

mthrok: > Users also have `torchaudio.transforms.Resample` :) Are any of these two (`Resample` and `sox_effects`) measurably slower than this implementation?

The difference is 100x

<details><summary>

</summary></details>


> 
> I'm not keen on adding parameters like these to load functions when there are already simple ways of doing so. On top of adding parameters to the signature, this also does add layers of backward-compatibility etc to the codebase.
> 
> How about other backends like soundfile?

torchaudio/csrc/sox_io.h
```diff
@@ -23,13 +23,23 @@ struct SignalInfo : torch::CustomClassHolder {
...
+// ver. 0
 c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> load_audio_file(
     const std::string& path,
     const int64_t frame_offset = 0,
     const int64_t num_frames = -1,
     const bool normalize = true,
     const bool channels_first = true);
 
+// ver. 1 sample_rate is added
+c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> load_audio_file_v1(
```
vincentqb, line 35: cc @mruberry -- in case you would like to comment on how to deprecate torchscript APIs

mruberry: It looks like you're adding a new argument with a default value. In that case you should be able to load older serialized Torchscript without issue, and I don't think you need to the old signature.

mthrok: @mruberry It seems not the case. I removed the BC code and BC test fails. https://app.circleci.com/pipelines/github/pytorch/audio/3270/workflows/0f877fa3-6256-4ef5-8d84-124684d17a97/jobs/93108

mruberry: > @mruberry It seems not the case. I removed the BC code and BC test fails. https://app.circleci.com/pipelines/github/pytorch/audio/3270/workflows/0f877fa3-6256-4ef5-8d84-124684d17a97/jobs/93108

Sorry, @mthrok, I feel like I misled you but that's what I was told. Empirically it seems to not be the case and I'll have to go back and investigate. I suppose you can keep the old signature and remap it into the new one like you originally were if you want to continue being able to load older serialized Torchscript.

torchaudio/csrc/sox_io.cpp
vincentqb: One more consideration: could someone want to use a specific algorithm for resampling, instead of the default one?

> When training models for different purposes, it would really beneficial to load audio while changing sampling rate and extracting a portion like one will do with STM files, without pre-converting audio files and storing on the file system.

We can already do that with in-memory sox effects, right?

torchaudio/csrc/sox_io.h
```diff
@@ -28,7 +28,8 @@ c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> load_audio_file(
     const int64_t frame_offset = 0,
     const int64_t num_frames = -1,
     const bool normalize = true,
-    const bool channels_first = true);
+    const bool channels_first = true,
+    const int64_t sample_rate = -1);
```
vincentqb, line 32: Where is the value of -1 documented?

torchaudio/csrc/sox_io.h
mthrok: > Sorry, @mthrok, I feel like I misled you but that's what I was told. Empirically it seems to not be the case and I'll have to go back and investigate. I suppose you can keep the old signature and remap it into the new one like you originally were if you want to continue being able to load older serialized Torchscript.

Do you have an estimate of when it would work without code for backward-compatibility?
I am wondering if I should wait if it's soon. That way we can avoid adding a new schema.

torchaudio/csrc/sox_io.h
mthrok: It is omitted because I judged that the code is self-explanatory.

torchaudio/csrc/sox_io.cpp
mthrok: > One more consideration: could someone want to use a specific algorithm for resampling, instead of the default one? How would the behavior of the resampling be guaranteed across different backends?

I do not think it is necessary to make that guarantee into a specification, because each backend has different limitations. If we make that guarantee into specification (except loading the data as is), we limit ourselves to only take the smallest subset of functionalities of all the backends.
Also, so far backends are for the purpose of platform-portability, and it is not meant to be swapped very often for the same installation.
If users can use the Resample transform for that purpose. I think, leveraging the backend's specific capability to provide more efficient way of perform it has added value.

> We can already do that with in-memory sox effects, right?

Yes, but this way it's more efficient when one wants to trim the audio, because sox will discard the un-needed portion.

PR #819, opened 2020-07-23
Make GTZAN dataset sorted and use on-the-fly data in GTZAN test

mthrok:
1. Change GTZAN to return samples in order sorted by file paths.
2. Change test to use on-the-fly data.

Relates to #764 and #791


mthrok:
thanks!

mthrok:
> LGTM, but I'd say the long term solution is to align GTZAN's implementation with the other ones also, and fix all at the same time.

Agreed. We need to find a generic way to perform pattern matching on paths regardless of platform.

PR #820, opened 2020-07-23
Update sample rate of libritts test

jimchen90:
This is to update the sample rate in libritts_test file.

In [LibriTTS](https://arxiv.org/pdf/1904.02882.pdf):
channel = 1
sample rate = 24 kHz
I use an arbitrary duration value in the test.

Related to #790

### DIFF
--- a/test/datasets/libritts_test.py
+++ b/test/datasets/libritts_test.py
@@ -33,8 +33,8 @@ class TestLibriTTS(TempDirMixin, TorchaudioTestCase):
             os.makedirs(file_dir, exist_ok=True)
             path = os.path.join(file_dir, filename)
 
-            data = get_whitenoise(sample_rate=8000, duration=6, n_channels=1, dtype='int16', seed=i)
-            save_wav(path, data, 8000)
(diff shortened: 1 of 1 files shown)


test/datasets/libritts_test.py
```diff
@@ -33,8 +33,8 @@ def setUpClass(cls):
             os.makedirs(file_dir, exist_ok=True)
             path = os.path.join(file_dir, filename)
 
-            data = get_whitenoise(sample_rate=8000, duration=6, n_channels=1, dtype='int16', seed=i)
-            save_wav(path, data, 8000)
+            data = get_whitenoise(sample_rate=24000, duration=2, n_channels=1, dtype='int16', seed=i)
+            save_wav(path, data, 24000)
```
vincentqb, line 37: @mthrok, @jimchen90 -- since we now use all synthetic data and dataset, we didn't catch this error. how do you suggest preventing this in the future?

mthrok: There is nothing to prevent here. Since the dataset implementation allows to load arbitral file in the given directory, there is no guarantee that all the data users added have some fixed properties.

Setting this to `24000` here is rather cosmetic, although if it's not the value presented in the paper, it feels wrong.

PR #822, opened 2020-07-23
Avoid "-" symbol in test name

mthrok:
Buck is not happy when test name contains "-" symbol, and it fails to run these tests.

### DIFF
--- a/test/sox_io_backend/smoke_test.py
+++ b/test/sox_io_backend/smoke_test.py
@@ -62,7 +62,7 @@ class SmokeTest(TempDirMixin, TorchaudioTestCase):
         [8000, 16000],
         [1, 2],
         [-4.2, -0.2, 0, 0.2, 96, 128, 160, 192, 224, 256, 320],
-    )), name_func=name_func)
+    )))
     @skipIfNoMP3
     def test_mp3(self, sample_rate, num_channels, bit_rate):
         """Run smoke test on mp3 format"""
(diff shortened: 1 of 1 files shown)


test/sox_io_backend/smoke_test.py
```diff
@@ -62,7 +62,7 @@ def test_wav(self, dtype, sample_rate, num_channels):
         [8000, 16000],
         [1, 2],
         [-4.2, -0.2, 0, 0.2, 96, 128, 160, 192, 224, 256, 320],
-    )), name_func=name_func)
+    )))
```
vincentqb, line 65: Can you clarify how this is working around the issue?

mthrok: `Parameterized` generates simple test names like `test_foo_1`, `test_foo_2`, `test_foo_3` ...

but since I generate a whole lot of tests, this default naming convention is not very helpful when some of them fail on CI. `name_func` here changes the naming convention to simply stringify parameters and concatenate them as test name, which is why "-" symbol is included in the test name.

The code change in PR gives up the custom naming function and use the default naming convention.

vincentqb: Good to know. Are you intentionally leaving some of the custom naming conventions in the other functions?

mthrok: Yes, they work well in places other than buck.

mthrok:
thanks

Issue #821, opened 2020-07-23
Improve tests for Dataset


mthrok:
`torchaudio` had minimalistic test for dataset implementations. See [here](https://github.com/pytorch/audio/blob/master/test/datasets/datasets_test.py)

Recently we have improved our test utilities and now we can generate synthetic data which emulates a subset of dataset. See examples [YesNo](https://github.com/pytorch/audio/blob/master/test/datasets/yesno_test.py) and [GTZAN](https://github.com/pytorch/audio/blob/master/test/datasets/gtzan_test.py) 

We would like to do the same for the remaining datasets

 - [ ] VCTK
 - [ ] LibriSpeech
 - [ ] LJSpeech
 - [ ] SpeechCommands
 - [ ] CMUArctic
 - [ ] CommonVoice

General Direction
1. Check the dataset of interest and pick subset of files, (check their sampling rate and number of channels)
2. Follow the approach of existing test module, create a new test module `test/datasets/XXX_test.py` and define your test class.
3. Generate pseudo dataset in `setUpClass` method. Create a list of expected data.
4. Traverse the directory with Dataset implementation
5. Check that files are traversed in the expected order, then loaded data match.
6. Check that Dataset traversed the expected number of files.

esqu1:
I'll take a stab at the `SpeechCommands` dataset 😄

edwardhdlu:
I'll take a look at LibriSpeech

lawrencechen98:
I'll check out the LJSpeech dataset

suraj813:
I'll have a go at CMUArctic

YazhiGao:
I'll take a look at CommonVoice dataset.

JianwuXu:
I'm taking the VCTK dataset

mthrok:
@JianwuXu Thanks, please be aware that VCTK dataset has been moved on upstream, and we are discussing the update to the Dataset implementation. https://github.com/pytorch/audio/pull/812

This is strange, but when we use mocked datasets in test, the loose specification of VCTK should work with the previous implementation, you can write the test for the new dataset found at https://datashare.is.ed.ac.uk/bitstream/handle/10283/3443/VCTK-Corpus-0.92.zip. Let me know if you are not sure. See other PRs for references too.

JianwuXu:
@mthrok the vctk.py code is expected to change?

Abhi011999:
@JianwuXu for updated dataset, yes.

JianwuXu:
[...]

PR #824, opened 2020-07-23
Add test for Speech Commands dataset

esqu1:
Adds a test for the `speechcommands` dataset and checks to see that the dataset loads the samples in the correct order.

Relates to #821.

# To test
`pytest test/datasets/speechcommands_test.py`


test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,70 @@
...
+    @classmethod
+    def setUp(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        dataset_dir = os.path.join(
+            cls.root_dir, speechcommands.FOLDER_IN_ARCHIVE, speechcommands.URL
+        )
+        os.makedirs(dataset_dir, exist_ok=True)
+        sample_rate = 16000  # 16kHz sample rate
+        for i in range(30):
+            label = "label_{:02d}".format(i)
```
mthrok, line 29: Directory structure must mimic that of the real dataset. That is, you need to make directory with names like the following; 

<details>

```
bed
bird
cat
dog
down
eight
five
follow
forward
four
go
happy
house
learn
left
marvin
nine
no
off
on
one
right
seven
sheila
six
stop
three
tree
two
up
visual
wow
yes
zero
```

</details>

these are the supposed label values returned by dataset.

also note that Dataset will walk directories in alphabetical order so you need to add samples in the expected order too.

test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,70 @@
...
+
+                for utterance in range(3):
+                    filename = f"{speaker}{speechcommands.HASH_DIVIDER}{utterance}.wav"
+                    file_path = os.path.join(path, filename)
+                    data = get_whitenoise(
+                        sample_rate=sample_rate,
+                        duration=0.01,
+                        n_channels=1,
+                        dtype="int16",
+                        seed=0,
```
mthrok, line 44: Please set unique `seed` value for each data. otherwise the exact same data are generated and we cannot differentiate them.

mthrok:
SpeechCommand test in `test/datasets/dataests_test.py` must be removed too.

test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,70 @@
...
+                    save_wav(file_path, data, sample_rate)
+                    sample = (
+                        normalize_wav(data),
+                        sample_rate,
+                        label,
+                        speaker,
+                        utterance,
+                    )
+                    cls.samples.append(sample)
+
```
mthrok, line 55: Can you put back the `print` statement here? It does not affect test  behavior but it is helpful when debugging. `pytest -s ` will show the directory structure constructed.

test/datasets/datasets_test.py
```diff
@@ -1,19 +1,16 @@
+from torchaudio.datasets.cmuarctic import CMUARCTIC
```
mthrok, line 1: Please keep changes minimum and revert changes related to formatting. Formatting changes update 'git blame' and it will make it hard to reason about the changes later.

test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,73 @@
...
+        dataset_dir = os.path.join(
+            cls.root_dir, speechcommands.FOLDER_IN_ARCHIVE, speechcommands.URL
+        )
+        os.makedirs(dataset_dir, exist_ok=True)
+        sample_rate = 16000  # 16kHz sample rate
+        random.seed(0)
+        for label in speechcommands.LABELS:
+            path = os.path.join(dataset_dir, label)
+            os.makedirs(path, exist_ok=True)
+            for j in range(100):
```
mthrok, line 33: Though nothing is wrong but 100 is too many and makes the test slow without much benefit. Let's change it to 2.

torchaudio/datasets/speechcommands.py
```diff
@@ -2,13 +2,10 @@
 from typing import Tuple
```
mthrok, line 2: Like mentioned above, please keep the changes minimum.

test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,73 @@
...
+
+    @classmethod
+    def setUp(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        dataset_dir = os.path.join(
+            cls.root_dir, speechcommands.FOLDER_IN_ARCHIVE, speechcommands.URL
+        )
+        os.makedirs(dataset_dir, exist_ok=True)
+        sample_rate = 16000  # 16kHz sample rate
+        random.seed(0)
```
mthrok, line 29: Please do not use random in test. It adds complexity without benefit. See my comment https://github.com/pytorch/audio/pull/827#discussion_r460899418 for why.

mthrok: Simply use integer value as seed and increment it.

test/datasets/speechcommands_test.py
```diff
@@ -0,0 +1,74 @@
...
```
mthrok, line 29: `random` is no longer used, and leaving it here will be confusing later. Can you remove it?

test/datasets/datasets_test.py
```diff
@@ -28,10 +28,6 @@ def test_ljspeech(self):
         data = LJSPEECH(self.path)
         data[0]
 
-    def test_speechcommands(self):
-        data = SPEECHCOMMANDS(self.path)
-        data[0]
```
mthrok, line 33: Can you also remove the corresponding `import` statement?

torchaudio/datasets/speechcommands.py
```diff
@@ -20,6 +20,42 @@
...
+    "stop",
+    "three",
+    "tree",
+    "two",
+    "up",
+    "visual",
+    "wow",
+    "yes",
+    "zero",
+]
```
mthrok, line 58: I gave some thoughts of this, but can you put this in test? Logically, this should belong here but since it's not used in the implementation, it might be confusing later.

mthrok:
Looks good. Can you resolve the conflict?

PR #826, opened 2020-07-24
Add unit test for LJSpeech dataset

lawrencechen98:
Add unit test for LJSpeech dataset by generating synthetic data which emulates a subset of dataset, as defined by #821


test/datasets/datasets_test.py
```diff
@@ -1,19 +1,15 @@
+from torchaudio.datasets.cmuarctic import CMUARCTIC
```
mthrok, line 1: Please keep the changes to minimum, and do not change lines for the sake of formatting. (it's okay to format the lines you are touching)

Running formatter is nice but it will disturb `git blame` later and it becomes hard to reason about changes in the future.

test/datasets/ljspeech_test.py
```diff
@@ -0,0 +1,80 @@
...
+        sample_rate = 22050
+
+        with open(metadata_path, mode="w", newline='') as metadata_file:
+            metadata_writer = csv.writer(
+                metadata_file, delimiter="|", quoting=csv.QUOTE_NONE
+            )
+            for i, (transcript, normalized_transcript) in enumerate(
+                zip(cls.transcripts, cls.normalized_transcripts)
+            ):
+                fileid = "LJ001-000" + str(i)
```
mthrok, line 56: Can you do `f'LJ001-{i:04d}'`? That way, if someone adds more 6 new transcript examples, they still work.

test/datasets/ljspeech_test.py
```diff
@@ -0,0 +1,80 @@
...
+                metadata_writer.writerow([fileid, transcript, normalized_transcript])
+                filename = fileid + ".wav"
+                path = os.path.join(archive_dir, filename)
+                data = get_whitenoise(
+                    sample_rate=sample_rate, duration=10, n_channels=1, dtype="int16", seed=i
```
mthrok, line 61: 10 seconds is redundant, please keep it short like one second.

test/datasets/ljspeech_test.py
```diff
@@ -0,0 +1,80 @@
...
+
+    root_dir = None
+    data = []
+    transcripts = [
+        "Test transcript 1",
+        "Test transcript 2",
+        "Test transcript 3",
+        (
+            "The examination and testimony of the experts enabled the "
+            "Commission to conclude that 5 shots may have been fired,"
```
mthrok, line 26: this might sound nit-picky but I do not find this transcript. In fact, the transcript I found in the actual dataset is same as normalized one.

```
$grep "The examination and testimony of the experts enabled the" LJSpeech-1.1/metadata.csv
LJ001-0031|In 1465 Sweynheim and Pannartz began printing in the monastery of Subiaco near Rome,|In fourteen sixty-five Sweynheim and Pannartz began printing in the monastery of Subiaco near Rome,
```

Since we are kind of using the real sample from the dataset, can we stick with the actual sample?
How about this one? (`1465` -> `fourteen sixty-five`)
```
LJ001-0031|In 1465 Sweynheim and Pannartz began printing in the monastery of Subiaco near Rome,|In fourteen sixty-five Sweynheim and Pannartz began printing in the monastery of Subiaco near Rome,
```

torchaudio/datasets/ljspeech.py
```diff
@@ -63,7 +63,7 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        with open(self._metadata_path, "r") as metadata:
+        with open(self._metadata_path, "r", newline='') as metadata:
```
mthrok, line 66: I assume this change is required for Windows. Is that correct?

test/datasets/ljspeech_test.py
lawrencechen98: Interesting, I took the example off the LJSpeech dataset website, but I didn't cross check it with the actual dataset, so there was probably some inconsistencies. I will use the example you provided. Thanks!

torchaudio/datasets/ljspeech.py
lawrencechen98: Yes, there were issues with the Windows binary checks that are run, and it appears to be a newline issue. This solution fixes the problem, and is the recommended way of opening a CSV file according to Python docs.

PR #827, opened 2020-07-25
Add test for CommonVoice dataset

YazhiGao:
Add dataset test for CommonVoice dataset as part of #821 .
```
pytest test/datasets/commonvoice_test.py
```
@mthrok for review and will sync with internal task once reviewed.

- [x]  test passed.
- [x]  fix possible review issues.
- [x]  CI passed.


torchaudio/datasets/commonvoice.py
```diff
@@ -13,7 +13,8 @@
 # test.tsv
 # train.tsv
 # validated.tsv
-
+TSVS = ["dev.tsv", "invalidated.tsv", "other.tsv", "test.tsv", "train.tsv", "validated.tsv"]
+DEFAULT_TSV = "train.tsv"
```
mthrok, line 17: Since these changes are not used in anywhere, can you revert?

test/datasets/commonvoice_test.py
```diff
@@ -0,0 +1,80 @@
...
+    _headers = [u"client_ids", u"path", u"sentence", u"up_votes", u"down_votes", u"age", u"gender", u"accent"]
+    _client_ids = [u"0", u"1"]
+    _sentence = [u"test", u"pytorch"]
+    _up_votes = [u"0", u"1"]
+    _down_votes = [u"0", u"1"]
+    _age = [u"0", u"1"]
+    _gender = [u"M", u"W", u"U"]
+    _accent = [u"Y", u"N"]
+    _folder_audio = "clips"
+    sample_rate = 22050
```
mthrok, line 29: I checked couple of randomly chosen files and the sample rates are all 48k Hz. Can you confirm?

```
soxi en/clips/common_voice_en_18885791.mp3

Input File     : 'en/clips/common_voice_en_18885791.mp3'
Channels       : 1
Sample Rate    : 48000
Precision      : 16-bit
Duration       : 00:00:03.26 = 156672 samples ~ 244.8 CDDA sectors
File Size      : 26.2k
Bit Rate       : 64.1k
Sample Encoding: MPEG audio (layer I, II or III)
```

test/datasets/commonvoice_test.py
```diff
@@ -0,0 +1,80 @@
...
+        tsv_filename = os.path.join(base_dir, commonvoice.DEFAULT_TSV)
+        with open(tsv_filename, "w", newline='') as tsv:
+            writer = csv.writer(tsv, delimiter='\t')
+            writer.writerow(cls._headers)
+            for i in range(100):
+                audio_filename = f'voice_{i:05d}.wav'
+
+                # Generate and store text data
+                text_data = [
+                    random.choice(cls._client_ids),
```
mthrok, line 48: Adding randomness to test is tricky, because in case of failure, you need to log everything that's required to reproduce the failure. In this case, random seed should do but we cannot really guarantee always that's the case.

Then the question is "Are we benefitting from adding randomness here?" or "Is randomness adding wider coverage of test cases?", and I do not think so, because after all, the generated samples are homogeneous in how they are processed by Dataset class. What matters more here is to have multiple of samples (which this code covers), because that way, we can see the ordering that Dataset traverses. (Which is the ultimate motivation to add this test. See https://github.com/pytorch/audio/issues/794 for some context.)

Instead I suggest to simply hard code some contents (preferably on module level in the hope for the improved readability)

```python
# on module level
train_csv_headers = ["client_ids", "path", "sentence", "up_votes", "down_votes", "age", "gender", "accent"]
# Note: extension is changed to wav for the sake of test
# Note: the first content is missing values for `age`, `gender` and `accent` as in the original data.
train_csv_contents = [
        ("9d16c5d980247861130e0480e2719f448be73d86a496c36d01a477cbdecd8cfd1399403d7a77bf458d211a70711b2da0845c1afdc5f1f694a2bbbccd162a34bf", "common_voice_en_18885784.wav", "He was accorded a State funeral, and was buried in Drayton and Toowoomba Cemetery.", "2", "0", "", "", "")
    ("c82eb9291328620f06025a1f8112b909099e447e485e99236cb87df008650250e79fea5ca772061fb6a370830847b9c44d2078cfa0f601b1f8cf6b98c479a7fc", "common_voice_en_556542.wav", "Once more into the breach", "2", "0", "thirties", "male", "us"),
    ("f74d880c5ad4c5917f314a604d3fc4805159d255796fb9f8defca35333ecc002bdf53dc463503c12674ea840b21b4a507b7cd0b3c636434f9a06bd81538de4f9", "common_voice_en_18607573.wav", "Caddy, show Miss Clare and Miss Summerson their rooms.", "2", "0", "twenties", "male", "canada"),
]

# in setUpClass
writer.writerow(train_csv_headers)
for i, content in enumerate(train_csv_contents):
    writer.write(content)
    audio_filename = content[1]

    data = get_whitenoise(sample_rate=cls.sample_rate, duration=1, n_channels=1, seed=i, dtype='float32')
```

test/datasets/commonvoice_test.py
```diff
@@ -0,0 +1,80 @@
...
+                    random.choice(cls._gender),
+                    random.choice(cls._accent)
+                ]
+                writer.writerow(text_data)
+
+                # Generate and store audio
+                audio_base_path = os.path.join(base_dir, cls._folder_audio)
+                os.makedirs(audio_base_path, exist_ok=True)
+                audio_path = os.path.join(audio_base_path, audio_filename)
+                data = get_whitenoise(sample_rate=cls.sample_rate, duration=6, n_channels=1, dtype='float32')
```
mthrok, line 64: `duration=6` is specific for `YesNo` dataset. You can set it to something smaller, say 1.
Also please set the `seed` value, otherwise all the generated data have exactly same shape. See the above comment for this.

test/datasets/datasets_test.py
```diff
@@ -35,26 +35,3 @@ def test_speechcommands(self):
...
-
-class TestCommonVoice(TorchaudioTestCase):
-    backend = 'default'
-    path = get_asset_path()
-
-    def test_commonvoice(self):
-        data = COMMONVOICE(self.path, url="tatar")
-        data[0]
-
-    def test_commonvoice_diskcache(self):
```
mthrok, line 48: Can you migrate these two tests that use `disckcache_iterator` and `bg_iterator` to `test/datasets/utils_test.py`?

`class TestIterator(TorchaudioTestCase)` then add as methods `test_disckcache_iterator` and `test_bg_iterator`. The contents of these tests can stay same.

test/datasets/commonvoice_test.py
YazhiGao: agreed

YazhiGao:
@mthrok for re-review.

test/datasets/commonvoice_test.py
```diff
@@ -0,0 +1,72 @@
...
+        ["f74d880c5ad4c5917f314a604d3fc4805159d255796fb9f8defca35333ecc002bdf53dc463503c12674ea840b21b4a507b7c",
+            "common_voice_en_18607573.wav",
+            "Caddy, show Miss Clare and Miss Summerson their rooms.", "2", "0", "twenties", "male", "canada"],
+    ]
+    _folder_audio = "clips"
+    sample_rate = 48000
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = os.path.join(cls.get_base_temp_dir(), 'waves_commonvoice')
```
mthrok, line 38: What is this `'waves_commonvoice'`? If it's at root level, it does not matter for Dataset, so it's redundant.

test/datasets/utils_test.py
```diff
@@ -45,3 +47,26 @@ def test_walk_files(self):
...
         assert n_ites == len(self.expected)
+
+
+class TestIterator(TorchaudioTestCase):
+    backend = 'default'
+    path = get_asset_path()
+
+    def test_commonvoice(self):
+        data = COMMONVOICE(self.path, url="tatar")
+        data[0]
```
mthrok, line 58: You can remove this `test_commonvoice` as the test suite you adding covers more than what this method can do.

test/datasets/datasets_test.py
```diff
@@ -35,26 +35,3 @@ def test_speechcommands(self):
     def test_cmuarctic(self):
         data = CMUARCTIC(self.path)
         data[0]
-
-
-class TestCommonVoice(TorchaudioTestCase):
```
mthrok, line 40: Can you remove the corresponding `import` statement too?

YazhiGao:
plz review @mthrok

YazhiGao:
> Looks good. Can you rebase onto the latest master to resolve the conflict??

done. feel free to merge after CI completion

YazhiGao:
@mthrok  ready for merge

mthrok:
Thanks!

PR #829, opened 2020-07-27
Added test for CMUArctic Dataset

suraj813:
PR linked to #821 
I have used the same dummy utterance for all the emulated samples. 

Requested review: @mthrok


test/datasets/cmuarctic_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        sample_rate = 16000
+        seed = 42
+        utterance = "This is a test utterance."
+
+        base_dir = os.path.join(cls.root_dir, "ARCTIC", "cmu_us_" + cls.URL + "_arctic")
+        # Contains utterance ID & sentence prompts
```
mthrok, line 29: This kind of comments, ones not adding more information than what code expresses are not necessary. You can remove them.

test/datasets/cmuarctic_test.py
```diff
@@ -0,0 +1,68 @@
...
+        os.makedirs(txt_dir, exist_ok=True)
+        txt_file = os.path.join(txt_dir, "txt.done.data")
+        # Contains the audio files
+        audio_dir = os.path.join(base_dir, "wav")
+        os.makedirs(audio_dir, exist_ok=True)
+
+        with open(txt_file, "w") as txt:
+            for i in range(10):
+                # Write audio file
+                utterance_id = f"arctic_a{i:04d}"
```
mthrok, line 40: Can you also add some of `f"arctic_b{i:04d}"` patterns?

test/datasets/cmuarctic_test.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def test_cmuarctic(self):
+        dataset = cmuarctic.CMUARCTIC(self.root_dir)
+        for i, (waveform, sample_rate, utterance, utterance_id) in enumerate(dataset):
+            expected_sample = self.samples[i]
+            assert sample_rate == expected_sample[1]
+            assert utterance == expected_sample[2]
+            assert utterance_id == expected_sample[3]
+            self.assertEqual(expected_sample[0], waveform, atol=5e-5, rtol=1e-8)
+        assert (i + 1) == len(self.samples)
```
mthrok, line 68: Please do not use temporary variable outside of the loop in which the variable was defined and meant to be used. This works, but this is easy to miss for the other developers who work on this code later. Define a dedicated variable for this.

test/datasets/datasets_test.py
```diff
@@ -32,10 +32,6 @@ def test_speechcommands(self):
         data = SPEECHCOMMANDS(self.path)
         data[0]
 
-    def test_cmuarctic(self):
```
mthrok, line 35: Can you remove the corresponding asset and `import` statement??

test/datasets/cmuarctic_test.py
```diff
@@ -0,0 +1,67 @@
...
+            for c in ["a", "b"]:
+                for i in range(5):
+                    utterance_id = f"arctic_{c}{i:04d}"
+                    path = os.path.join(audio_dir, f"{utterance_id}.wav")
+                    data = get_whitenoise(
+                        sample_rate=sample_rate,
+                        duration=3,
+                        n_channels=1,
+                        dtype="int16",
+                        seed=seed,
```
mthrok, line 44: Please use different `seed` value for each generated samples, otherwise all the loaded Tensors have the exactly same shape and value, and that will lose the point of comparing loaded Tensor object.

mthrok:
Could you rebase onto the latest master?

test/datasets/cmuarctic_test.py
```diff
@@ -0,0 +1,66 @@
...
+            for c in ["a", "b"]:
+                for i in range(5):
+                    utterance_id = f"arctic_{c}{i:04d}"
+                    path = os.path.join(audio_dir, f"{utterance_id}.wav")
+                    data = get_whitenoise(
+                        sample_rate=sample_rate,
+                        duration=3,
+                        n_channels=1,
+                        dtype="int16",
+                        seed=i + ord(c),
```
mthrok, line 43: This does not produce unique values for all the generated data. Simply initialize the seed value outside of the most outer loop and increment it. See https://github.com/pytorch/audio/blob/master/test/datasets/gtzan_test.py#L27-L44

suraj813: Thanks @mthrok, sorry I overlooked this.

PR #830, opened 2020-07-28
Update documentation and fix docstrings (#788)

vincentqb:
- Addresses #549 #638 #786
- Add `torchaudio` top level module doc
- Separate `torchaudio` top level module doc from `index.html`
- Add `backend` module doc.
- Remove `-> None` from function signature as it adds noise to documentation
- Changed function argument name of `torchaudio.backend.sox_io_backend.save` from `tensor` to `src`, so that it matches with the reset of backends.
- Tweak bunch of docstrings

### DIFF
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -7,6 +7,12 @@ from torchaudio._internal import (
 
 
 class AudioMetaData:
+    """Data class to be returned by :py:func:`~torchaudio.backend.sox_io_backend.info`.
+
+    :ivar int sample_rate: Sample rate
+    :ivar int num_frames: The number of frames
+    :ivar int num_channels: The number of channels
+    """
     def __init__(self, sample_rate: int, num_frames: int, num_channels: int):
         self.sample_rate = sample_rate
         self.num_frames = num_frames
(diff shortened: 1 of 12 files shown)


torchaudio/backend/sox_io_backend.py
```diff
@@ -83,37 +112,49 @@ def load(
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def save(
         filepath: str,
-        tensor: torch.Tensor,
```
mthrok, line 86: Probably we want to revert these signature changes now that 0.6.0 is built.

torchaudio/backend/sox_io_backend.py
```diff
@@ -127,8 +168,22 @@ def save(
             compression = 3.
         else:
             raise RuntimeError(f'Unsupported file type: "{ext}"')
-    signal = torch.classes.torchaudio.TensorSignal(tensor, sample_rate, channels_first)
```
mthrok, line 130: Probably we want to revert these signature changes now that 0.6.0 is built.

torchaudio/backend/sox_io_backend.py
```diff
@@ -127,8 +168,22 @@ def save(
...
+    signal = torch.classes.torchaudio.TensorSignal(src, sample_rate, channels_first)
     torch.ops.torchaudio.sox_io_save_audio_file(filepath, signal, compression)
 
 
-load_wav = load
```
mthrok, line 134: Probably we want to revert these signature changes now that 0.6.0 is built.

torchaudio/backend/sox_io_backend.py
vincentqb: good catch, do you want to go ahead and just edit the change them directly?

vincentqb:
For reference, how to build docs for [website](https://pytorch.org/audio/):
```
conda create -n torchaudiodocs
conda activate torchaudiodocs

conda install sphinx-build
conda install sphinx
conda install -c pytorch pytorch
pip install sphinxcontrib-katex

# Install latest pytorch_sphinx_theme in conda environment

# Get a fresh copy of torchaudio
git clone https://github.com/pytorch/audio audio-docs
git checkout origin/release/0.6
cd audio-docs/docs
make html
git checkout origin/gh-pages
cd ..
cp -rp docs/build/html/* ./
git add .
git commit -m"update docs for 0.6.0"
git push origin gh-pages
```

PR #842, opened 2020-07-30
Set zip_safe = False to disable egg installation

mthrok:
Currently `setuptools` assumes that torhcaudio is zip safe and performs
egg (zip) installation when `python setup.py install` (or `pip install .`).

When torchaudio load extension module written for Torchscript, the corresponding
loading function requires the `so` file to be present as an actual file, and this
does not work well with egg installation.

This PR fixes this by setting `zip_safe=False` in setup.py and disable egg
installation, so that `python setup.py install` installs uncompressed files
with regular directory structure.

The following is the error you get when you try to use the package installed with
`python setup.py install`, before this change.

```
    import torchaudio
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 656, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 626, in _load_backward_compatible
  File "/torchaudio/envs/master-py3.6/lib/python3.6/site-packages/torchaudio-0.7.0a0+d6831c6-py3.6-linux-x86_64.egg/torchaudio/__init__.py", line 1, in <module>
  File "<frozen importlib._bootstrap>", line 971, in _find_and_load
  File "<frozen importlib._bootstrap>", line 955, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 656, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 626, in _load_backward_compatible
  File "/torchaudio/envs/master-py3.6/lib/python3.6/site-packages/torchaudio-0.7.0a0+d6831c6-py3.6-linux-x86_64.egg/torchaudio/extension/__init__.py", line 5, in <module>
  File "/torchaudio/envs/master-py3.6/lib/python3.6/site-packages/torchaudio-0.7.0a0+d6831c6-py3.6-linux-x86_64.egg/torchaudio/extension/extension.py", line 12, in _init_extension
  File "/torchaudio/envs/master-py3.6/lib/python3.6/site-packages/torchaudio-0.7.0a0+d6831c6-py3.6-linux-x86_64.egg/torchaudio/extension/extension.py", line 19, in _init_script_module
  File "/torchaudio/envs/master-py3.6/lib/python3.6/site-packages/torch/_classes.py", line 46, in load_library
    torch.ops.load_library(path)
[...]

PR #845, opened 2020-07-31
[CI] Run unit test with non-editable installation

mthrok:
We have been running unit test with editable installation. (i.e. `python setup.py develop`), with which we missed issues like #842. 

This PR makes installation in CI non-editable, and change test directory structure so that the source code will not shadow the installed version of `torchaudio`. With simple `pytest test`, [`pytest` modifies](https://docs.pytest.org/en/stable/pythonpath.html) and prepend checked out repository, which shadows the installed version.

To remedy this, the whole test suites has been moved from `./test` to `./test/torchaudio_unittest`. This adds nice module structure to our test code and we can do absolute import in each test module, which makes it possible again to run test with `python -m unittest torchaudio_unittest/XXX.py`

This change does not affect the regular development process (`python setup.py develop` && `pytest test`)


.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -15,8 +15,16 @@ python -m torch.utils.collect_env
 export TORCHAUDIO_TEST_FAIL_IF_NO_EXTENSION=1
 export PATH="${PWD}/third_party/install/bin/:${PATH}"
 
+declare -a common_args=(
+    '--import-mode' 'importlib'
```
mthrok, line 19: See https://docs.pytest.org/en/stable/pythonpath.html for the detail of the behavior change by `--import-mode`.

By default `pytest` modifies `sys.path` at runtime and prepend the path to the root directory of the repository, and it cause the source directory of `torchaudio` shadows the version of `torchaudio` built/installed instead. We want to test the installed version so we do not want `pytest` to modify `sys.path`.

.circleci/unittest/windows/scripts/run_test.sh
```diff
@@ -6,5 +6,5 @@ eval "$(./conda/Scripts/conda.exe 'shell.bash' 'hook')"
 conda activate ./env
 
 python -m torch.utils.collect_env
-pytest --cov=torchaudio --junitxml=test-results/junit.xml -v --durations 20 test
-flake8 torchaudio test
```
vincentqb, line 10: this is not tested separately, hence why this is removed here, right?

.circleci/unittest/linux/scripts/run_test.sh
```diff
@@ -15,8 +15,15 @@ python -m torch.utils.collect_env
...
+)
 if [ "${os}" == MacOSX ] ; then
-    pytest -q -n auto --dist=loadscope --cov=torchaudio --junitxml=test-results/junit.xml --durations 20 test
+    declare -a args=('-q' '-n' 'auto' '--dist=loadscope')
 else
-    pytest -v --cov=torchaudio --junitxml=test-results/junit.xml --durations 20 test
+    declare -a args=('-v')
 fi
+cd test
+pytest "${args[@]}" "${common_args[@]}" torchaudio_unittest
```
vincentqb, line 29: could the following command work without moving files from `test/` to `test/torchaudio_unittest/`?
```
pytest "${args[@]}" "${common_args[@]}" test
```

.circleci/unittest/windows/scripts/run_test.sh
mthrok: that's correct.

.circleci/unittest/linux/scripts/run_test.sh
mthrok: An alternative would be remove source `torchaudio` directory.
1. do `rm torchaudio` before running test - ugly
2. Move the source code to `src`. See [this](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure). - though this is clean and standard way, it involves the reorg of whole source code.

vincentqb: Nice, thanks for pointing these out! I'm ok with the approach suggested here -- though neither pytorch nor the other libraries do it.

cc @cpuhrsch @fmassa @zhangguanheng66

mthrok:
Thanks

PR #846, opened 2020-07-31
Fix import order to suppress the setuptools warning

mthrok:
setuptools warns that it should be imported before Distutils

```
site-packages/setuptools/distutils_patch.py:25: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,8 +3,8 @@ import os
 import shutil
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


mthrok:
thanks

PR #847, opened 2020-07-31
[CI] Make *nix unit test fail if C++ extension is not available

mthrok:
I am trying to fix the issue C++ extension availability when building/installing torchaudio from source. See https://github.com/pytorch/audio/pull/842 for the detail.

Currently our test suites automatically/silently skip tests on C++ extension if it is not available.
This is nice in local env, but in CI these tests should be enforced and reported as failure if C++ extension is not available.

This PR adds switch for making tests fail if C++ extension is not available.

This should make https://github.com/pytorch/audio/pull/845 fail so that we can debug what is wrong with the current setup.

### DIFF
--- a/.circleci/unittest/linux/scripts/run_test.sh
+++ b/.circleci/unittest/linux/scripts/run_test.sh
@@ -12,6 +12,7 @@ eval "$(./conda/bin/conda shell.bash hook)"
 conda activate ./env
 
 python -m torch.utils.collect_env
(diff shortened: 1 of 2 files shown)


mthrok:
Example of failure on CI

https://app.circleci.com/pipelines/github/pytorch/audio/3049/workflows/5b3042b8-eb77-414c-9e1e-b89e8faf6704/jobs/84548

mthrok:
thanks!

Issue #823, opened 2020-07-23
🚀 Feature Request: Audio mix and degrader


VictorBeraldo:
## 🚀 Feature

Add mixing feature to torchaudio with an desirable SNR
## Motivation
This feature wil be usefull for audio data augmentation by adding noise to the audio signal. 

## Alternatives
There are a lot of python packages that contains this features, but it will be very useful if we have it inside torchaudio

mthrok:
Hi @VictorBeraldo 

I love this idea and I was actually thinking about a new augmentation (degradation) module [here](https://github.com/pytorch/audio/issues/800#issuecomment-662005956) (although it turned `miniaudio` leaks memory, so we cannot do it right away).

Do you have an idea of interface or an inspiration from existing packages? It would help to discuss requirements, pros and cons.

VictorBeraldo:
Hello @mthrok!
I find two of them here: https://github.com/emilio-molina/audio_degrader and https://github.com/sleekEagle/audio_processing.
The first is more complete with a lot of complex features and the second uses what exactly what i want to have in torchaudio.

mthrok:
Hi @VictorBeraldo 

Thanks for the links. I skimmed through them and here is my thoughts
 - *audio_degrader*

mthrok:
Hi @VictorBeraldo

Thanks for the links. I skimmed through them and here is my thoughts

- **audio_degrader**
    - **sox**
        torchaudio has direct binding to `libsox` where you can run most of sox effects [directly on a Tensor](https://github.com/pytorch/audio/blob/577796ba112228a7965c3b1ffc277c692e7b8ce9/torchaudio/sox_effects/sox_effects.py#L61) or [on a file](https://github.com/pytorch/audio/blob/577796ba112228a7965c3b1ffc277c692e7b8ce9/torchaudio/sox_effects/sox_effects.py#L163) so it's already covered.
    - **ffmpeg**
       It seems that `MP3` codec is the unique functionality of this. (other things like speed change, filtering are covered). This is what I was thinking with doing with `miniaudio`.
    - **rubberband**
       I am not familiar with this library. Will take a look again later (but I have a feeling that `libsox` can do this too)
- **audio_processing**
   - Both functions look valuable and I believe it should not be difficult to add to `torchaudio`. Though we need to settle on the interface (and module name/location).

I guess we can start from functions to add noises to Tensor object.
Some parameters for this includes SNR and maybe the type of noises.

What do you think?

[...]

PR #848, opened 2020-08-03
[WIP]Mix noise with audio

dataprowl:
This PR relates to #823 

**Included in this:**

- Simple utility to mix any background noise with audio at a specified SNR (in dB).

- Function to mix audio with white noise  at a specified SNR (in dB).

- Function to mix audio with red noise  at a specified SNR (in dB).

**Assumptions**

- Both background noise and audio are sampled at the sampling rate. 

- Both background noise and audio are mono channel.

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -888,7 +888,7 @@ def treble_biquad(
         waveform (Tensor): audio waveform of dimension of `(..., time)`
         sample_rate (int): sampling rate of the waveform, e.g. 44100 (Hz)
         gain (float): desired gain at the boost (or attenuation) in dB.
-        central_freq (float, optional): central frequency (in Hz). (Default: ``3000``)
+        central_freq (float, optional): central frequency (in Hz). (Default: ``3000``)whi
         Q (float, optional): https://en.wikipedia.org/wiki/Q_factor (Default: ``0.707``).
 
     Returns:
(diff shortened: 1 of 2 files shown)


dataprowl:
> I think that the correct formule is snr = 10 ** (snr/10) instead 20.

@VictorBeraldo 

I agree! Regret for the mistake. Please see my change [here](https://github.com/pytorch/audio/pull/848/commits/e265ff3b4ec0a755463c38eeb0558133ea26219f#diff-292a2ef33b3ecf4c40cda1a4afe58760R2346). 

I hope this utility is similar to what you intend to integrate into torchaudio.

vincentqb:
The addition of white noise is a simple operation, and wouldn't in itself warrant a new abstraction. That being said, it could make sense in the context of other noise types. Before going too deep in this pull request, we need to agree on a reference to look at and test against. Thoughts?

torchaudio/functional.py
```diff
@@ -2314,3 +2321,93 @@ def vad(
...
+
+
+def add_background_noise(
+    waveform: Tensor,
+    noise: Tensor,
+    snr: float,
+    normalize: bool = True
+)-> Tensor:
+    r"""
+    Augment noise to waveform, with specified SNR value to generate degraded waveform.
```
vincentqb, line 2333: Based on this description, it is not clear to me what this function would do beyond "addition". What use case do you have in mind for such a user-facing function? Would this be better as a private function?

vincentqb: Naively, this function would consume the color of the noise that a user wants to add?

mthrok:
Closing as the work was initiated without specification. Please propose the specification in#823 first.

PR #849, opened 2020-08-03
Fix skipIfNoExtension

mthrok:
There was a flaw in the logic introduced in #847, and now that tests that require C++ extension are skipped all the time. This PR fixes it.


mthrok:
thanks!

PR #863, opened 2020-08-07
Add spectrogram normalization option

jimchen90:
This is to add the spectrogram normalization option in [WaveRNN pipeline](https://github.com/pytorch/audio/pull/749).

### DIFF
--- a/examples/pipeline_wavernn/processing.py
+++ b/examples/pipeline_wavernn/processing.py
@@ -31,15 +31,19 @@ class NormalizeDB(nn.Module):
     r"""Normalize the spectrogram with a minimum db value
     """
 
-    def __init__(self, min_level_db):
+    def __init__(self, min_level_db, normalization):
         super().__init__()
         self.min_level_db = min_level_db
+        self.normalization = normalization
 
     def forward(self, specgram):
-        specgram = 20 * torch.log10(torch.clamp(specgram, min=1e-5))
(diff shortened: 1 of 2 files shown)


examples/pipeline_wavernn/processing.py
```diff
@@ -31,15 +31,19 @@ class NormalizeDB(nn.Module):
...
-        return torch.clamp(
-            (self.min_level_db - specgram) / self.min_level_db, min=0, max=1
-        )
+        if self.normalization:
+            specgram = 20 * torch.log10(torch.clamp(specgram, min=1e-5))
+            return torch.clamp(
+                (self.min_level_db - specgram) / self.min_level_db, min=0, max=1
+            )
+        else:
+            return torch.log10(torch.clamp(specgram, min=1e-5))
```
vincentqb, line 46: Is the factor of 20 missing intentional?

If so, let's refactor this as:
```python
        specgram = torch.log10(torch.clamp(specgram, min=1e-5))
        if self.normalization:
            specgram = torch.clamp(
                (self.min_level_db - 20 * specgram) / self.min_level_db, min=0, max=1
            )
        return specgram
```

jimchen90: yes, we don't need the factor of 20 when `normalization` is false.
I agree, will update it.

PR #864, opened 2020-08-07
Update librosa/numba in CI

mthrok:
librosa 0.8.0 has been released which include fix for https://github.com/librosa/librosa/issues/1160 (https://github.com/librosa/librosa/pull/1164)


mthrok:
thanks

PR #881, opened 2020-08-14
[CI] Install conda package from workspace

mthrok:
When running smoke test, proper way to install conda package is to
install from workspace, instead of installing from archive.
Otherwise the dependencies are not properly installed

https://github.com/conda/conda/issues/1884

For this, we pass the whole `conda-bld` directory from
conda package build job to upload job then smoke test job.

`torchaudio` does not have a mandatory third-party python package dependency,
so this was not an issue.

See also https://github.com/pytorch/text/pull/803


mthrok:
thanks!

Issue #765, opened 2020-07-05
Add TED-LIUM 3 dataset 


jiwidi:
## 🚀 Feature
Hi! I saw torchaudio is missing the TED-LIUM dataset from https://arxiv.org/abs/1805.04699, https://www.openslr.org/51/ which is commonly used in ASR benchmarks and I would like to propose the addition of it to the list of avaliable datasets (And propose myself to code and contribute for it! :) )

## Motivation

TED-LIUM is a very common dataset in ASR benchmarking tasks and we should include it on the library.

## Pitch

Add TED-LIUM 3 dataset to the list of datasets

I would like to work in a PR to introduce it if it sounds good to the main contributors :)

vincentqb:
Sounds good to me, and you can open a PR for this! Please make sure to follow the current dataset code organization when doing so :)

jiwidi:
> Sounds good to me, and you can open a PR for this! Please make sure to follow the current dataset code organization when doing so :)

Nice! Will do :)

vincentqb:
Closed by #882 -- please feel free to re-open if this is not the case :)

PR #871, opened 2020-08-10
Add SPHERE support to "sox_io" backend

mthrok:
Relates to #868

PR #882, opened 2020-08-15
Add tedlium dataset (all 3 releases)

jiwidi:
Hi!

So this is my pull request for adding tedlium support (all 3 releases of tedlium) reference to https://github.com/pytorch/audio/issues/765.

The code is not yet finished (still need to add test, read the phonetic dictionaries of each release, proper formatting) but I wanted to create the PR to get some feedback on the way I read/iterate tedlium files and if the torchaudio manteiners are okay with it or have an idea on how to improve it. This way I make sure the core of the datasetloader is good to merge and I can start working on the test or change it before coding the tests. I will explain how I iterate through the dataset now.

Tedlium dataset (all releases) is composed of `.sph` and `.stm` files. These files are divided in the root folder in two different folders:
```
sph/
stm/
```

Each `.stm` file contains the transcripts for a FULL tedtalk and while the brother audio `.sph` file contains the audio for the FULL tedtalk as well. 

Since this dataset is used in speech recognition, people don't train in full tedtalk audios but rather split into sentences. Each sentence corresponds to a line in the `.stm` file. This means that a `.stm` file contains as many training samples as lines inside the file. 

The previous datasets that I found in torchaudio didn't encounter this case and treated every transcript file as a sample, so I had to make a small change to the `self._walker` used in the datasets. I created a new `self._extended_walker` list that contains as many occurrences of a `.stm` file with its line number as lines this file has, so for example if we had two stm files A (5 lines/5 samples) and B (3 lines/3 samples) `self._extended_walker` will look like this:
```
self._extended_walker = [[A,0],[A,1],[A,2],[A,3],[A,4],[B,0,][B,1]]
```

And every time we load an item `i` of the dataset we will index this `_extended_walker` where we retrieve the file containing the sample and the corresponding line, the line contains the transcript and start time,end time of the audio file corresponding to the audio file. With all of this information we can load the line transcript and corresponding subaudio from the full ted talk.

This will allow the dataset to work as any other torchaudio datasets(being able to run sequentially or shuffling it). Looks great right? Then why I am checking with the maintainers if this is a good idea?

Well, in order to create this `_extended_walker` list we have to read how many lines each of our `.stm` file has while initializing the dataset, causing a "not optimal I/O operation". The operation happens in 119-123:
```python
for file in self._walker:
    stm_path = os.path.join(self._path, self._folder_txt, file + self._ext_txt)
    with open(stm_path) as f:
        l = len(f.readlines())
        self._extended_walker += [(file, line) for line in range(l)]
```

Is fast enough to not represent any downside for the user
image
(release 3 contains the higher number of files)

I'm open for edits and suggestions on the implementation of the TEDLIUM dataset and please notice is still a work in progress.

## Things that still need work:
* `_getdict` function to return the phonems dictionary of the dataset
* Several test for the dataset and different releases.
* Upload my notebook where I play with the dataset object to show the functionality for reviewers
* Better styling and compliance with the rest of torchaudio code

Thanks

### DIFF
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -4,11 +4,11 @@ torchaudio.datasets
 All datasets are subclasses of :class:`torch.utils.data.Dataset`
 i.e, they have ``__getitem__`` and ``__len__`` methods implemented.
 Hence, they can all be passed to a :class:`torch.utils.data.DataLoader`
-which can load multiple samples parallelly using ``torch.multiprocessing`` workers. 
+which can load multiple samples parallelly using ``torch.multiprocessing`` workers.
 For example: ::
-    
+
     yesno_data = torchaudio.datasets.YESNO('.', download=True)
-    data_loader = torch.utils.data.DataLoader(yesno_data, 
+    data_loader = torch.utils.data.DataLoader(yesno_data,
                                               batch_size=1,
                                               shuffle=True,
                                               num_workers=args.nThreads)
@@ -22,7 +22,7 @@ All the datasets have almost similar API. They all have two common arguments:
 ``transform`` and  ``target_transform`` to transform the input and target respectively.
 
 
-.. currentmodule:: torchaudio.datasets 
+.. currentmodule:: torchaudio.datasets
 
 
 CMUARCTIC
@@ -80,6 +80,13 @@ SPEECHCOMMANDS
   :members: __getitem__
   :special-members:
 
+
+TEDLIUM
+~~~~~~~~~~~~~~
+
+.. autoclass:: TEDLIUM
+  :members: __getitem__
+  :special-members: get_phoneme_dict
 
 VCTK
 ~~~~
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -19,6 +19,16 @@ UTTERANCES = [
     "AaronHuey_2010X 1 AaronHuey_2010X 8.0 10.0 <o,f0,female> script5\n",
 ]
 
+PHONEME = [
+    "a AH",
+    "a(2) EY",
+    "aachen AA K AH N",
+    "aad AE D",
+    "aaden EY D AH N",
+    "aadmi AE D M IY",
+    "aae EY EY",
+]
+
 
 class TestTedlium(TempDirMixin, TorchaudioTestCase):
     backend = "default"
@@ -60,6 +70,11 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
             with open(trans_path, "w") as f:
                 f.write("".join(UTTERANCES))
 
+            dict_filename = f"{release}.dic"
+            dict_path = os.path.join(release_dir, dict_filename)
+            with open(dict_path, "w") as f:
+                f.write("\n".join(PHONEME))
+
             # Create a samples list to compare with
             cls.samples[release] = []
             for utterance in UTTERANCES:
@@ -92,6 +107,11 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
 
         assert num_samples == len(self.samples[release])
 
+        dataset._dict_path = os.path.join(dataset._path, f"{release}.dic")
+        phoneme_dict = dataset.phoneme_dict
+        phoenemes = [f"{key} {' '.join(value)}" for key, value in phoneme_dict.items()]
+        assert phoenemes == PHONEME
+
     def test_tedlium_release2(self):
         release = "release2"
         dataset = tedlium.TEDLIUM(self.root_dir, release=release)
@@ -107,6 +127,11 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
 
         assert num_samples == len(self.samples[release])
 
+        dataset._dict_path = os.path.join(dataset._path, f"{release}.dic")
+        phoneme_dict = dataset.phoneme_dict
+        phoenemes = [f"{key} {' '.join(value)}" for key, value in phoneme_dict.items()]
+        assert phoenemes == PHONEME
+
     def test_tedlium_release3(self):
         release = "release3"
         dataset = tedlium.TEDLIUM(self.root_dir, release=release)
@@ -121,4 +146,9 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
             num_samples += 1
 
         assert num_samples == len(self.samples[release])
+
+        dataset._dict_path = os.path.join(dataset._path, f"{release}.dic")
+        phoneme_dict = dataset.phoneme_dict
+        phoenemes = [f"{key} {' '.join(value)}" for key, value in phoneme_dict.items()]
+        assert phoenemes == PHONEME
 
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -151,4 +151,3 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
         phoneme_dict = dataset.phoneme_dict
         phoenemes = [f"{key} {' '.join(value)}" for key, value in phoneme_dict.items()]
         assert phoenemes == PHONEME
-
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -24,7 +24,7 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
     backend = "default"
 
     root_dir = None
-    samples = []
+    samples = {}
 
     @classmethod
     def setUpClass(cls):
@@ -33,8 +33,9 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
         os.makedirs(dataset_dir, exist_ok=True)
         sample_rate = 16000  # 16kHz
         seed = 0
-        data = get_whitenoise(sample_rate=sample_rate, duration=10.00, n_channels=1, dtype="float32", seed=seed)
+
         for release in ["release1", "release2", "release3"]:
+            data = get_whitenoise(sample_rate=sample_rate, duration=10.00, n_channels=1, dtype="float32", seed=seed)
             if release in ["release1", "release2"]:
                 release_dir = os.path.join(
                     dataset_dir,
@@ -42,7 +43,11 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
                     tedlium._RELEASE_CONFIGS[release]["subset"],
                 )
             else:
-                release_dir = os.path.join(dataset_dir, tedlium._RELEASE_CONFIGS[release]["folder_in_archive"])
+                release_dir = os.path.join(
+                    dataset_dir,
+                    tedlium._RELEASE_CONFIGS[release]["folder_in_archive"],
+                    tedlium._RELEASE_CONFIGS[release]["data_path"],
+                )
             os.makedirs(release_dir, exist_ok=True)
             os.makedirs(os.path.join(release_dir, "stm"), exist_ok=True)  # Subfolder for transcripts
             os.makedirs(os.path.join(release_dir, "sph"), exist_ok=True)  # Subfolder for audio files
@@ -55,39 +60,65 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
             with open(trans_path, "w") as f:
                 f.write("".join(UTTERANCES))
 
-        # Create a samples list to compare with
-        for i, utterance in enumerate(UTTERANCES):
-            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = utterance.split(" ", 6)
-            start_time = int(float(start_time)) * 16000
-            end_time = int(float(end_time)) * 16000
-            sample = (
-                data[:, start_time:end_time],
-                sample_rate,
-                transcript,
-                talk_id,
-                speaker_id,
-                identifier,
-            )
-            cls.samples.append(sample)
+            # Create a samples list to compare with
+            cls.samples[release] = []
+            for i, utterance in enumerate(UTTERANCES):
+                talk_id, _, speaker_id, start_time, end_time, identifier, transcript = utterance.split(" ", 6)
+                start_time = int(float(start_time)) * sample_rate
+                end_time = int(float(end_time)) * sample_rate
+                sample = (
+                    data[:, start_time:end_time],
+                    sample_rate,
+                    transcript,
+                    talk_id,
+                    speaker_id,
+                    identifier,
+                )
+                cls.samples[release].append(sample)
+            seed += 1
 
-    @classmethod
-    def tearDownClass(cls):
-        # In case of test failure
-        tedlium.TEDLIUM._ext_audio = ".flac"
-
-    def test_tedlium(self):
-        tedlium.TEDLIUM._ext_audio = ".wav"
-        dataset = tedlium.TEDLIUM(self.root_dir)
-        print(dataset._path)
+    def test_tedlium_release1(self):
+        release = "release1"
+        dataset = tedlium.TEDLIUM(self.root_dir, release=release)
         num_samples = 0
         for i, (data, sample_rate, transcript, talk_id, speaker_id, identifier) in enumerate(dataset):
-            self.assertEqual(data, self.samples[i][0], atol=5e-5, rtol=1e-8)
-            assert sample_rate == self.samples[i][1]
-            assert transcript == self.samples[i][2]
-            assert talk_id == self.samples[i][3]
-            assert speaker_id == self.samples[i][4]
-            assert identifier == self.samples[i][5]
+            self.assertEqual(data, self.samples[release][i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.samples[release][i][1]
+            assert transcript == self.samples[release][i][2]
+            assert talk_id == self.samples[release][i][3]
+            assert speaker_id == self.samples[release][i][4]
+            assert identifier == self.samples[release][i][5]
             num_samples += 1
 
-        assert num_samples == len(self.samples)
-        tedlium.TEDLIUM._ext_audio = ".flac"
+        assert num_samples == len(self.samples[release])
+
+    def test_tedlium_release2(self):
+        release = "release2"
+        dataset = tedlium.TEDLIUM(self.root_dir, release=release)
+        num_samples = 0
+        for i, (data, sample_rate, transcript, talk_id, speaker_id, identifier) in enumerate(dataset):
+            self.assertEqual(data, self.samples[release][i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.samples[release][i][1]
+            assert transcript == self.samples[release][i][2]
+            assert talk_id == self.samples[release][i][3]
+            assert speaker_id == self.samples[release][i][4]
+            assert identifier == self.samples[release][i][5]
+            num_samples += 1
+
+        assert num_samples == len(self.samples[release])
+
+    def test_tedlium_release3(self):
+        release = "release3"
+        dataset = tedlium.TEDLIUM(self.root_dir, release=release)
+        num_samples = 0
+        for i, (data, sample_rate, transcript, talk_id, speaker_id, identifier) in enumerate(dataset):
+            self.assertEqual(data, self.samples[release][i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.samples[release][i][1]
+            assert transcript == self.samples[release][i][2]
+            assert talk_id == self.samples[release][i][3]
+            assert speaker_id == self.samples[release][i][4]
+            assert identifier == self.samples[release][i][5]
+            num_samples += 1
+
+        assert num_samples == len(self.samples[release])
+
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -1,0 +1,93 @@
+import os
+
+from torchaudio.datasets import tedlium
+
+from torchaudio_unittest.common_utils import (
+    TempDirMixin,
+    TorchaudioTestCase,
+    get_whitenoise,
+    save_wav,
+    normalize_wav,
+)
+
+# Used to generate a unique utterance for each dummy audio file
+UTTERANCES = [
+    "AaronHuey_2010X 1 AaronHuey_2010X 0.0 2.0 <o,f0,female> script1\n",
+    "AaronHuey_2010X 1 AaronHuey_2010X 2.0 4.0 <o,f0,female> script2\n",
+    "AaronHuey_2010X 1 AaronHuey_2010X 4.0 6.0 <o,f0,female> script3\n",
+    "AaronHuey_2010X 1 AaronHuey_2010X 6.0 8.0 <o,f0,female> script4\n",
+    "AaronHuey_2010X 1 AaronHuey_2010X 8.0 10.0 <o,f0,female> script5\n",
+]
+
+
+class TestTedlium(TempDirMixin, TorchaudioTestCase):
+    backend = "default"
+
+    root_dir = None
+    samples = []
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        cls.root_dir = dataset_dir = os.path.join(cls.root_dir, "tedlium")
+        os.makedirs(dataset_dir, exist_ok=True)
+        sample_rate = 16000  # 16kHz
+        seed = 0
+        data = get_whitenoise(sample_rate=sample_rate, duration=10.00, n_channels=1, dtype="float32", seed=seed)
+        for release in ["release1", "release2", "release3"]:
+            if release in ["release1", "release2"]:
+                release_dir = os.path.join(
+                    dataset_dir,
+                    tedlium._RELEASE_CONFIGS[release]["folder_in_archive"],
+                    tedlium._RELEASE_CONFIGS[release]["subset"],
+                )
+            else:
+                release_dir = os.path.join(dataset_dir, tedlium._RELEASE_CONFIGS[release]["folder_in_archive"])
+            os.makedirs(release_dir, exist_ok=True)
+            os.makedirs(os.path.join(release_dir, "stm"), exist_ok=True)  # Subfolder for transcripts
+            os.makedirs(os.path.join(release_dir, "sph"), exist_ok=True)  # Subfolder for audio files
+            filename = f"{release}.sph"
+            path = os.path.join(os.path.join(release_dir, "sph"), filename)
+            save_wav(path, data, sample_rate)
+
+            trans_filename = f"{release}.stm"
+            trans_path = os.path.join(os.path.join(release_dir, "stm"), trans_filename)
+            with open(trans_path, "w") as f:
+                f.write("".join(UTTERANCES))
+
+        # Create a samples list to compare with
+        for i, utterance in enumerate(UTTERANCES):
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = utterance.split(" ", 6)
+            start_time = int(float(start_time)) * 16000
+            end_time = int(float(end_time)) * 16000
+            sample = (
+                data[:, start_time:end_time],
+                sample_rate,
+                transcript,
+                talk_id,
+                speaker_id,
+                identifier,
+            )
+            cls.samples.append(sample)
+
+    @classmethod
+    def tearDownClass(cls):
+        # In case of test failure
+        tedlium.TEDLIUM._ext_audio = ".flac"
+
+    def test_tedlium(self):
+        tedlium.TEDLIUM._ext_audio = ".wav"
+        dataset = tedlium.TEDLIUM(self.root_dir)
+        print(dataset._path)
+        num_samples = 0
+        for i, (data, sample_rate, transcript, talk_id, speaker_id, identifier) in enumerate(dataset):
+            self.assertEqual(data, self.samples[i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.samples[i][1]
+            assert transcript == self.samples[i][2]
+            assert talk_id == self.samples[i][3]
+            assert speaker_id == self.samples[i][4]
+            assert identifier == self.samples[i][5]
+            num_samples += 1
+
+        assert num_samples == len(self.samples)
+        tedlium.TEDLIUM._ext_audio = ".flac"
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -62,7 +62,7 @@ class TestTedlium(TempDirMixin, TorchaudioTestCase):
 
             # Create a samples list to compare with
             cls.samples[release] = []
-            for i, utterance in enumerate(UTTERANCES):
+            for utterance in UTTERANCES:
                 talk_id, _, speaker_id, start_time, end_time, identifier, transcript = utterance.split(" ", 6)
                 start_time = int(float(start_time)) * sample_rate
                 end_time = int(float(end_time)) * sample_rate
--- a/torchaudio/datasets/tedlium.py
+++ b/torchaudio/datasets/tedlium.py
@@ -57,7 +57,7 @@ class TEDLIUM(Dataset):
         Args:
             root (str): Path containing dataset or target path where its downloaded if needed
             release (str, optional): TEDLIUM identifier (release1,release2,release3). Defaults to RELEASE.
-            subset (str, optional): Subset of data(train,test,dev) supported for release 1 and 2. Defaults to Train/None.
+            subset (str, optional): Subset of data(train,test,dev) supported for release 1,2. Defaults to Train/None.
             download (bool, optional): Download dataset in case is not founded in root path. Defaults to False.
             audio_ext (str, optional): Overwrite audio extension when loading items. Defaults to ".sph".
 
--- a/torchaudio/datasets/tedlium.py
+++ b/torchaudio/datasets/tedlium.py
@@ -12,8 +12,6 @@ from torchaudio.datasets.utils import (
 from collections import namedtuple
 
 
-RELEASE = "release1"  # Default release
-
 _RELEASE_CONFIGS = {
     "release1": {
         "folder_in_archive": "TEDLIUM_release1",
@@ -21,6 +19,7 @@ _RELEASE_CONFIGS = {
         "checksum": "30301975fd8c5cac4040c261c0852f57cfa8adbbad2ce78e77e4986957445f27",
         "data_path": "",
         "subset": "train",
+        "dict": "TEDLIUM.150K.dic",
     },
     "release2": {
         "folder_in_archive": "TEDLIUM_release2",
@@ -28,6 +27,7 @@ _RELEASE_CONFIGS = {
         "checksum": "93281b5fcaaae5c88671c9d000b443cb3c7ea3499ad12010b3934ca41a7b9c58",
         "data_path": "",
         "subset": "train",
+        "dict": "TEDLIUM.152k.dic",
     },
     "release3": {
         "folder_in_archive": "TEDLIUM_release-3",
@@ -35,6 +35,7 @@ _RELEASE_CONFIGS = {
         "checksum": "ad1e454d14d1ad550bc2564c462d87c7a7ec83d4dc2b9210f22ab4973b9eccdb",
         "data_path": "data/",
         "subset": None,
+        "dict": "TEDLIUM.152k.dic",
     },
 }
 
@@ -50,7 +51,7 @@ class TEDLIUM(Dataset):
     """
 
     def __init__(
-        self, root: str, release: str = RELEASE, subset: str = None, download: bool = False, audio_ext=".sph"
+        self, root: str, release: str = "release1", subset: str = None, download: bool = False, audio_ext=".sph"
     ) -> None:
         """Constructor for TEDLIUM dataset
 
@@ -93,11 +94,24 @@ class TEDLIUM(Dataset):
                 extract_archive(archive)
 
         self._walker = []
+
+        # Create walker for all samples
         for file in walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True):
             stm_path = os.path.join(self._path, "stm", file + ".stm")
             with open(stm_path) as f:
                 l = len(f.readlines())
                 self._walker.extend((file, line) for line in range(l))
+        # Read phoneme dictionary
+        dict_path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["dict"])
+        self.phoneme_dict = {}
+        with open(dict_path, "rb") as f:
+            for line in f.readlines():
+                content = line.decode("utf-8").strip("\n").split(" ", 1)
+                if len(content) > 1:
+                    key, value = content[0], content[1]
+                    self.phoneme_dict[key] = value.strip()
+                    # # Some lines in release1 dont have a phoneme for a word so value will be out of index
+                    # # Need to find a better solution to read the dictionary
 
     def load_tedlium_item(self, fileid: str, line: int, path: str) -> Tedlium_item:
         """Loads a TEDLIUM dataset sample given a file name and corresponding sentence name
@@ -116,26 +130,27 @@ class TEDLIUM(Dataset):
             talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(" ", 6)
 
         wave_path = os.path.join(path, "sph", fileid)
-        # Calculate indexes for start time and endtime
-        start_time = int(float(start_time) * 16000)
-        end_time = int(float(end_time) * 16000)
-        waveform, sample_rate = self.load_audio(
-            wave_path + self._ext_audio, frame_offset=start_time, num_frames=end_time - start_time
+        waveform, sample_rate = self.__load_audio__(
+            wave_path + self._ext_audio, start_time=start_time, end_time=end_time
         )
+
         return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
 
-    def load_audio(self, path: str, frame_offset: int = 0, num_frames: int = -1) -> [Tensor, int]:
+    def __load_audio__(self, path: str, start_time: float, end_time: float, sample_rate: int = 16000) -> [Tensor, int]:
         """Default load function used in TEDLIUM dataset, you can overwrite this function to customize functionality
+        and load individual sentnces from a full ted audio talk file
 
         Args:
             path (str): Path to audio file
-            frame_offset (int, optional): Number of frames to use as offstet when loading audio. Defaults to 0
-            num_frames (int, optional): How many frames to load from the audio. Defaults to -1
+            start_time (int, optional): Time in seconds where the sample sentence stars
+            end_time (int, optional): Time in seconds where the sample sentence finishes
 
         Returns:
             [Tensor, int]: Audio tensor representation and sample rate
         """
-        return torchaudio.load(path, frame_offset=frame_offset, num_frames=num_frames)
+        start_time = int(float(start_time) * 16000)
+        end_time = int(float(end_time) * 16000)
+        return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
 
     def __getitem__(self, n: int) -> Tedlium_item:
         """TEDLIUM dataset custom function overwritting default loadbehaviour.
@@ -157,3 +172,11 @@ class TEDLIUM(Dataset):
             int: TEDLIUM dataset length
         """
         return len(self._walker)
+
+    def get_phoneme_dict(self):
+        """Returns the phoneme dictionary of a TEDLIUM release
+
+        Returns:
+            dictionary: Phoneme dictionary for the current tedlium release
+        """
+        return self.phoneme_dict
--- a/torchaudio/datasets/tedlium.py
+++ b/torchaudio/datasets/tedlium.py
@@ -105,7 +105,7 @@ class TEDLIUM(Dataset):
             identifier,
         )
 
-    def load_audio(self, path: str) -> Tensor:
+    def load_audio(self, path: str) -> [Tensor, int]:
         return torchaudio.load(path)
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
--- a/torchaudio/datasets/tedlium.py
+++ b/torchaudio/datasets/tedlium.py
@@ -8,7 +8,6 @@ from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
 )
-from collections import namedtuple
 
 
 _RELEASE_CONFIGS = {
@@ -41,14 +40,10 @@ _RELEASE_CONFIGS = {
     },
 }
 
-Tedlium_item = namedtuple(
-    "Tedlium_item", ["waveform", "sample_rate", "transcript", "talk_id", "speaker_id", "identifier"]
-)
-
 
 class TEDLIUM(Dataset):
     """
-    Create a Dataset for Tedlium. It supports releases 1,2 and 3, each item is a tuple of the form:
+    Create a Dataset for Tedlium. It supports releases 1,2 and 3, each item is a list containings:
     [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
 
     Constructor arguments:
@@ -122,8 +117,8 @@ class TEDLIUM(Dataset):
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        # Create walker for all samples
-        self._walker = []
+        # Create list for all samples
+        self._filelist = []
         stm_path = os.path.join(self._path, "stm")
         for file in sorted(os.listdir(stm_path)):
             if file.endswith(".stm"):
@@ -131,17 +126,11 @@ class TEDLIUM(Dataset):
                 with open(stm_path) as f:
                     l = len(f.readlines())
                     file = file.replace(".stm", "")
-                    self._walker.extend((file, line) for line in range(l))
+                    self._filelist.extend((file, line) for line in range(l))
+        # Create dict path for later read
+        self.dict_path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["dict"])
 
-        # Read phoneme dictionary
-        dict_path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["dict"])
-        self.phoneme_dict = {}
-        with open(dict_path, "r", encoding="utf-8") as f:
-            for line in f.readlines():
-                content = line.strip().split(maxsplit=1)
-                self.phoneme_dict[content[0]] = content[1:]  # content[1:] can be empty list
-
-    def _load_tedlium_item(self, fileid: str, line: int, path: str) -> Tedlium_item:
+    def _load_tedlium_item(self, fileid: str, line: int, path: str) -> Tuple[Tensor, int, str, int, int, int]:
         """Loads a TEDLIUM dataset sample given a file name and corresponding sentence name
 
         Args:
@@ -160,7 +149,7 @@ class TEDLIUM(Dataset):
         wave_path = os.path.join(path, "sph", fileid)
         waveform, sample_rate = self._load_audio(wave_path + self._ext_audio, start_time=start_time, end_time=end_time)
 
-        return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
+        return (waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
 
     def _load_audio(self, path: str, start_time: float, end_time: float, sample_rate: int = 16000) -> [Tensor, int]:
         """Default load function used in TEDLIUM dataset, you can overwrite this function to customize functionality
@@ -176,9 +165,11 @@ class TEDLIUM(Dataset):
         """
         start_time = int(float(start_time) * 16000)
         end_time = int(float(end_time) * 16000)
-        return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
+        if torchaudio.get_audio_backend() == "sox_io":
+            return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
+        return torchaudio.load(path)[:, start_time:end_time]
 
-    def __getitem__(self, n: int) -> Tedlium_item:
+    def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
         """TEDLIUM dataset custom function overwritting default loadbehaviour.
         Loads a TEDLIUM sample given a index N
 
@@ -188,7 +179,7 @@ class TEDLIUM(Dataset):
         Returns:
             Tedlium_item: A namedTuple containing [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
         """
-        fileid, line = self._walker[n]
+        fileid, line = self._filelist[n]
         return self._load_tedlium_item(fileid, line, self._path)
 
     def __len__(self) -> int:
@@ -197,7 +188,7 @@ class TEDLIUM(Dataset):
         Returns:
             int: TEDLIUM dataset length
         """
-        return len(self._walker)
+        return len(self._filelist)
 
     def get_phoneme_dict(self):
         """Returns the phoneme dictionary of a TEDLIUM release
@@ -205,4 +196,11 @@ class TEDLIUM(Dataset):
         Returns:
             dictionary: Phoneme dictionary for the current tedlium release
         """
+        # Read phoneme dictionary
+        if not hasattr(self, "phoneme_dict"):
+            self.phoneme_dict = {}
+            with open(self.dict_path, "r", encoding="utf-8") as f:
+                for line in f.readlines():
+                    content = line.strip().split(maxsplit=1)
+                    self.phoneme_dict[content[0]] = content[1:]  # content[1:] can be empty list
         return self.phoneme_dict
--- a/torchaudio/datasets/tedlium.py
+++ b/torchaudio/datasets/tedlium.py
@@ -7,7 +7,6 @@ from torch.utils.data import Dataset
 from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
-    walk_files,
 )
 from collections import namedtuple
 
@@ -19,6 +18,7 @@ _RELEASE_CONFIGS = {
         "checksum": "30301975fd8c5cac4040c261c0852f57cfa8adbbad2ce78e77e4986957445f27",
         "data_path": "",
         "subset": "train",
+        "supported_subsets": ["train", "test", "dev"],
         "dict": "TEDLIUM.150K.dic",
     },
     "release2": {
@@ -27,6 +27,7 @@ _RELEASE_CONFIGS = {
         "checksum": "93281b5fcaaae5c88671c9d000b443cb3c7ea3499ad12010b3934ca41a7b9c58",
         "data_path": "",
         "subset": "train",
+        "supported_subsets": ["train", "test", "dev"],
         "dict": "TEDLIUM.152k.dic",
     },
     "release3": {
@@ -35,6 +36,7 @@ _RELEASE_CONFIGS = {
         "checksum": "ad1e454d14d1ad550bc2564c462d87c7a7ec83d4dc2b9210f22ab4973b9eccdb",
         "data_path": "data/",
         "subset": None,
+        "supported_subsets": [None],
         "dict": "TEDLIUM.152k.dic",
     },
 }
@@ -46,8 +48,27 @@ Tedlium_item = namedtuple(
 
 class TEDLIUM(Dataset):
     """
-    Create a Dataset for Tedlium. Each item is a tuple of the form:
+    Create a Dataset for Tedlium. It supports releases 1,2 and 3, each item is a tuple of the form:
     [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
+
+    Constructor arguments:
+
+    Args:
+        root (str): Path containing dataset or target path where its downloaded if needed
+        release (str, optional): TEDLIUM identifier (release1,release2,release3). Defaults to RELEASE.
+        subset (str, optional): train/dev/test for releases 1&2, None for release3. Defaults to Train/None
+        download (bool, optional): Download dataset in case is not founded in root path. Defaults to False.
+        audio_ext (str, optional): Overwrite audio extension when loading items. Defaults to ".sph".
+
+    Special functions:
+
+    _load_tedlium_item: Loads a TEDLIUM dataset sample given a file name and corresponding sentence name
+
+    _load_audio: Default load function used in TEDLIUM dataset, you can overwrite this function to customize
+                 functionality and load individual sentences from a full ted audio talk file
+
+    get_phoneme_dict: Returns the phoneme dictionary of a TEDLIUM release
+
     """
 
     def __init__(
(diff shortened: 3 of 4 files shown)


jiwidi:
Pinging @vincentqb

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+        if subset in ["train", "dev", "test"]:
+            self._path = os.path.join(self._path, subset + "/")
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _CHECKSUMS.get(url, None)
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        walker = walk_files(self._path, suffix=self._ext_txt, prefix=False, remove_suffix=True)
```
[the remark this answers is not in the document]
mthrok: I am downloading all TEDLIUM datasets now so that we can discuss the detail of how to traverse these directories. I will give my thoughts once I see the dataset strucutres.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+    _ext_audio = ".sph"
+    _folder_audio = "sph/"
+    _folder_txt = "stm/"
+
+    def __init__(
+        self,
+        root: str,
+        release: str = RELEASE,
+        subset: str = None,
+        folder_in_archive: str = _RELEASE_CONFIGS[RELEASE]["folder_in_archive"],
```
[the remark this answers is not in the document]
jiwidi: Yep, there is no reason why it would stay 👍

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
jiwidi: That would be perfect, lets have a discussion for it when you have the data.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+        "checksum": "685d27c39c53217383d7933cc405a07048004127",
+        "data_path": "data/",
+        "subset": None,
+    },
+}
+
+
+def load_tedlium_item(
+    fileid: str, line: int, path: str, ext_audio: str, ext_txt: str
+) -> Tuple[Tensor, int, str, int, int, int]:
```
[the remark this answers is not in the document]
jiwidi: Yes, makes sense

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+    transcript_path = os.path.join(path, "stm/", fileid)
+    with open(transcript_path + ext_txt) as f:
+        transcript = f.readlines()[line]
+        talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(
+            " ", 6
+        )
+
+    wave_path = os.path.join(path, "sph/", fileid)
+    waveform, sample_rate = torchaudio.load(wave_path + ext_audio)
```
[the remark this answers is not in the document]
jiwidi: Yes, makes sense

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+        "data_path": "data/",
+        "subset": None,
+    },
+}
+
+
+def load_tedlium_item(
+    fileid: str, line: int, path: str, ext_audio: str, ext_txt: str
+) -> Tuple[Tensor, int, str, int, int, int]:
+    transcript_path = os.path.join(path, "stm/", fileid)
```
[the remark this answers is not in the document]
jiwidi: Good catch

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+
+class TEDLIUM(Dataset):
+    """
+    Create a Dataset for Tedlium. Each item is a tuple of the form:
+    waveform, sample_rate, utterance, speaker_id, chapter_id, utterance_id
+    """
+
+    _ext_txt = ".stm"
+    _ext_audio = ".sph"
+    _folder_audio = "sph/"
```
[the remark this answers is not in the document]
jiwidi: I used them to mimic structure on the Librispeech dataset:https://github.com/pytorch/audio/blob/c692fe99ac353f0e2fc7b5ad7162f037a70a35fe/torchaudio/datasets/librispeech.py#L75-L76

I agree there is no benefit from having them as a variable rather than a attribute but there is the consistency also if we want to keep them. I have no problem replacing them with hardcoding but wanted to point this before making the change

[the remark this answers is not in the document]
jiwidi: Yes, makes sense

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class TEDLIUM(Dataset):
+    """
+    Create a Dataset for Tedlium. Each item is a tuple of the form:
+    waveform, sample_rate, utterance, speaker_id, chapter_id, utterance_id
+    """
+
+    _ext_txt = ".stm"
+    _ext_audio = ".sph"
```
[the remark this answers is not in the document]
jiwidi: Sure no problem

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+
+class TEDLIUM(Dataset):
+    """
+    Create a Dataset for Tedlium. Each item is a tuple of the form:
+    waveform, sample_rate, utterance, speaker_id, chapter_id, utterance_id
+    """
+
+    _ext_txt = ".stm"
```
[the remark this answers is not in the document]
jiwidi: Is used for the file walker as we look for files with this extension and I keept it as a variable for consistency with the Librispeech dataset
https://github.com/pytorch/audio/blob/c692fe99ac353f0e2fc7b5ad7162f037a70a35fe/torchaudio/datasets/librispeech.py#L75-L76

It will probably unused depending on what logic we decide on to iterate though the files. We can take a look at this when we are done with the logic ?

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,133 @@
...
+            )
+
+        basename = os.path.basename(url)
+        archive = os.path.join(root, basename)
+
+        basename = basename.split(".")[0]
+
+        self._path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["data_path"])
+        if subset in ["train", "dev", "test"]:
+            self._path = os.path.join(self._path, subset + "/")
```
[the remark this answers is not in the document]
jiwidi: Will fix it and the rest of suggestions you pointed out, a commit will follow :+1:

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
mthrok: > Sure no problem but when is the user going to want to change the ext_audio? Is `.sph` for all tedlium datasets, there is no point in changing the value of it, we have it as a variable here so its not hardcoded and can be easier to read.

Yeah, normally, there is no need to change the audio extension. One potential I think of is Windows users. I do not know much about Windows dev env, but say they cannot load SPHERE format. Once thing they can do is to recreate the dataset by converting them into `wav` format. In such case would be nice if they can reuse your implementation and just change the audio extension.

BTW I think `audio_ext` is better than `ext_audio` as a name of the function arguments.

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
mthrok: > Is used for the file walker as we look for files with this extension 

Even that's the case, I think it can be hardcoded at where the walker function is called, however

> It will probably unused depending on what logic we decide on to iterate though the files. We can take a look at this when we are done with the logic ?

Yeah, we can differ the change until then.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,116 @@
...
+        walker = walk_files(self._path, suffix=self._ext_txt, prefix=False, remove_suffix=True)
+        self._walker = list(walker)
+        self._extended_walker = []
+        for file in self._walker:
+            stm_path = os.path.join(self._path, "stm", file + self._ext_txt)
+            with open(stm_path) as f:
+                l = len(f.readlines())
+                self._extended_walker += [(file, line) for line in range(l)]
+
+    def load_tedlium_item(self, fileid: str, line: int, path: str) -> Tuple[Tensor, int, str, int, int, int]:
```
mthrok, line 87: Instead of returning bunch of variables, can you define a dedicated type with `namedtuple`?

In general, when returning multiple objects from a function, it a good idea to introduce micro protocol with `namedtuple`. This way the access to the individual item gets cleaner and it becomes future-prof (adding a new item to be returned is backward compatible.), and the code becomes more readable. Check out my example [here](https://github.com/mthrok/audio/blob/wsj0-mix-dataset/torchaudio/datasets/wsj0mix.py#L10).

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,116 @@
...
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import (
+    download_url,
+    extract_archive,
+    walk_files,
+)
+
+RELEASE = "release1"  # Default release
+
+_RELEASE_CONFIGS = {
```
mthrok, line 15: I just ran `sha256sum` and got different values. Which algorithm did you use?

```
$ sha256sum archives/tedlium/*
ad1e454d14d1ad550bc2564c462d87c7a7ec83d4dc2b9210f22ab4973b9eccdb  archives/tedlium/TEDLIUM_release-3.tgz
30301975fd8c5cac4040c261c0852f57cfa8adbbad2ce78e77e4986957445f27  archives/tedlium/TEDLIUM_release1.tar.gz
93281b5fcaaae5c88671c9d000b443cb3c7ea3499ad12010b3934ca41a7b9c58  archives/tedlium/TEDLIUM_release2.tar.gz
```

jiwidi: My bad here, I used sha1sum by mistake, I runned again and got your same hashes, will update them in the next commit

torchaudio/datasets/tedlium.py
jiwidi: Yes! Nice example, first time I see this practice :)

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
jiwidi: I moved _ext_audio initialization to the constructor on my coming new commit so I removed it as well. We can always go back.

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
jiwidi: Looks reasonable then, agree that is a better naming for paremeter

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+            transcript = f.readlines()[line]
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(" ", 6)
+
+        wave_path = os.path.join(path, "sph", fileid)
+        waveform, sample_rate = self.load_audio(wave_path + self._ext_audio)
+        # Calculate indexes for start time and endtime
+        start_time = int(float(start_time) * sample_rate)
+        end_time = int(float(end_time) * sample_rate)
+        waveform = waveform[:, start_time:end_time]
+        return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
```
vincentqb, line 126: Though I like having NamedTuple, this is a point of discussion on which domains have not yet resolved, e.g. #330

[internal](https://fb.quip.com/ZbB5AqwjiQtr#DHDACAQItBJ)

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+        walker = walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True)
+        self._walker = list(walker)
+        self._extended_walker = []
+        for file in self._walker:
+            stm_path = os.path.join(self._path, "stm", file + ".stm")
+            with open(stm_path) as f:
+                l = len(f.readlines())
+                self._extended_walker += [(file, line) for line in range(l)]
+
+    def load_tedlium_item(self, fileid: str, line: int, path: str) -> Tedlium_item:
```
vincentqb, line 104: Let's stick with the format used in other dataset where this is a function outside of the class.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+    def load_audio(self, path: str) -> [Tensor, int]:
+        """Default load function used in TEDLIUM dataset, you can overwrite this function to customize functionality
+
+        Args:
+            path (str): Path to audio file
+
+        Returns:
+            [Tensor, int]: Audio tensor representation and sample rate
+        """
+        return torchaudio.load(path)
```
vincentqb, line 137: This is not necessary as a method, we can simply move this inside the load item function as done in other datasets.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _RELEASE_CONFIGS[release]["checksum"]
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        walker = walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True)
+        self._walker = list(walker)
+        self._extended_walker = []
```
vincentqb, line 97: I'm not sure I understand why we need an extended walker in place of updating walker? It seems like this simply extracts information from a file. Could this be done in load item function?

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+            self._path = os.path.join(self._path, subset)
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _RELEASE_CONFIGS[release]["checksum"]
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        walker = walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True)
+        self._walker = list(walker)
```
vincentqb, line 96: in our current implementation, this would simply go as the last command manipulating walker (or extended walker here)

torchaudio/datasets/tedlium.py
vincentqb: Though I like the idea of NamedTuple, this is a point of discussion on which domains have not yet agreed, e.g. #330. I'm happy to revisit this discussion though.

[internal](https://fb.quip.com/ZbB5AqwjiQtr#DHDACAQItBJ)

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
vincentqb: @mthrok -- What's the advantage of doing this over having `torchaudio.load` inside the load item function? Is your suggestion to also eventually have a `load_text` method? One property of having a single load item function was that it factored the item loading as a pure function that was taking an identifier to return a data point.

torchaudio/datasets/tedlium.py
jiwidi: It does simple extract information for the files contained in the original walker. I create it because we need this information before calling the load item function so the dataset contains samples equals to total lines in all files (1 line is one sentence) instead of containing samples equals to files.

This dataset is a bit tricky because each file contains a full ted talk and no one in speech recognition (where this dataset is usually used) treats a full talk as a training sample. We split the file and audio file into multiple samples (sentences).

This means that once we instantiate the dataset we need to know how many samples we have(lines in all files) so the user can index any sample N at any time or shuffle it. If we were to implement this on the load function then this functionality will be way harder. See the original post where I explain with detail how the files are iterated.

I'm happy to discuss which method is the best method to achieve this functionality but I'm checking first if you understand why this extender walker is here first, maybe I didn't explain it correctly on the original post

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
jiwidi: > @mthrok -- What's the advantage of doing this over having `torchaudio.load` inside the load item function? Is your suggestion to also eventually have a `load_text` method? One property of having a single load item function was that it factored the item loading as a pure function that was taking an identifier to return a data point.

As far as I understood it he wants to give the user the option to overwrite the load function of the dataset to use optional parameters if he feels like.

So user could do:
```python
def custom_load(path):
    return torchaudio.load(path,normalize=False,channels_first=False)
dt = torchaudio.datasets.TEDLIUM()
dt.load_audio = custom_load
```
 
As `normalize` and `channels_first` are True by default and the `load_item` function doesnt change them the user has no ability to change them.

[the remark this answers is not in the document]
mthrok: Another example is Windows users who cannot use `torchaudio.load` because torchaudio on Windows does not support `sph` format so they need to provide something else in the custom load method.

mthrok:
@jiwidi 
> * `_getdict` function to return the phonems dictionary of the dataset

Do you mean the lexicon file such as `TEDLIUM.150K.dic` and `TEDLIUM.152K.dic`?
`dict` is very overloaded word here, so I just want a clarification on what you mean.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+        with open(transcript_path + ".stm") as f:
+            transcript = f.readlines()[line]
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(" ", 6)
+
+        wave_path = os.path.join(path, "sph", fileid)
+        waveform, sample_rate = self.load_audio(wave_path + self._ext_audio)
+        # Calculate indexes for start time and endtime
+        start_time = int(float(start_time) * sample_rate)
+        end_time = int(float(end_time) * sample_rate)
+        waveform = waveform[:, start_time:end_time]
```
mthrok, line 125: You can use `frame_offset`, and `num_frames` arguments to perform slicing directory.
This will not load the entire audio file, so it will be faster.

To do this, you need to know `sample_rate` beforehand.
You can either hardcode 16000 or do `torchaudio.info` to get the sample rate dynamically.
I think `torchaudio.info` can incur performance issue because it will open each file twice.
Hardcoding 16000 should be fine because all the TED-LIUM dataset is 16k[Hz].

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
vincentqb: > As `normalize` and `channels_first` are True by default and the `load_item` function doesnt change them the user has no ability to change them.

If this is indeed something desired, we could also consider exposing those parameters to the constructor instead. I wouldn't say this is enough alone to change the implementation through the addition of a load method.

In torchaudio, we went with the convention that waveforms are between -1 and 1 with batch first, see [readme](https://github.com/pytorch/audio/blob/master/README.md). This is in particular to avoid having to carry such parameters everywhere.

> Another example is Windows users who cannot use `torchaudio.load` because torchaudio on Windows does not support `sph` format so they need to provide something else in the custom load method.

This is an interesting point, and there's precedence for this for datasets using mp3. I see two paths for this:

* Offering a backend that supports `sph` in windows
* Decoupling `torchaudio.load` from `torchaudio.datasets`.

In the former, a user who decides to change the datasets would copy-paste the code and modify to their heart's content. In the latter, we need to decide how we would standardize this. Is any of torch{text,vision} doing something like this already? Thoughts?

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,158 @@
...
+                extract_archive(archive)
+
+        walker = walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True)
+        self._walker = list(walker)
+        self._extended_walker = []
+        for file in self._walker:
+            stm_path = os.path.join(self._path, "stm", file + ".stm")
+            with open(stm_path) as f:
+                l = len(f.readlines())
+                self._extended_walker += [(file, line) for line in range(l)]
```
mthrok, line 102: I checked the contents of STM files. For the case of `release-3`

```
$ wc -l TEDLIUM_release1/*/stm/*.stm | tail -1
   58863 total
$ wc -l TEDLIUM_release2/*/stm/*.stm | tail -1
   95033 total
$ wc -l TEDLIUM_release-3/data/stm/* | tail -1
  268263 total
```

So if loading the whole `release-3` dataset, `self._extended_walker` could take a couple of megabytes on memory.


Also this will open the number of files;

```
$ find  TEDLIUM_release1 -type f -name '*.stm' | wc -l
793
$ find  TEDLIUM_release2 -type f -name '*.stm' | wc -l
1514
$ find  TEDLIUM_release-3 -type f -name '*.stm' | wc -l
2370
```

For `release-3` dataset, this initialization process will go over 2370 files sequentially. How long does this take?

mthrok: oh is this what your report as 24 micros second??

torchaudio/datasets/tedlium.py
vincentqb: Thanks for adding details :) What I meant is more about the particular use of `_walker` and `_extended_walker`. I would have suggested something like:
```python
        self._walker = []
        for file in walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True):
            stm_path = os.path.join(self._path, "stm", file + ".stm")
            with open(stm_path) as f:
                l = len(f.readlines())
                self._walker.extend((file, line) for line in range(l))
```
or, to stick with generators,
```python
    def _generate_walker(...):
        for file in walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True):
            stm_path = os.path.join(self._path, "stm", file + ".stm")
            with open(stm_path) as f:
                for line in range(len(f.readlines())):
                    yield (file, line)

    self._walker = list(_generate_walker(...))
```

Also, why just indexing and not storing the line directly since the file is already being read? How big is the total text?
```python
    def _generate_walker(...):
        for file in walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True):
            stm_path = os.path.join(self._path, "stm", file + ".stm")
            with open(stm_path) as f:
                for line in f.readlines():
                    yield (file, line)

    self._walker = list(_generate_walker(...))
```

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
jiwidi: I would say offering support for sph in windows would be better, less hassling and moving around for the user. At the end, the users who use this custom dataset loaders are the ones looking for a working solution and easy to implement, if they want to customize loading or whatever they will probably rewrite a dataset on their own.

torchaudio/datasets/tedlium.py
jiwidi: Oh I see what you mean now, yes it will be cleaner to stick to the _walker name for the resultant walker.

I went ahead and checked the size of the walker list for release3 (the biggest)
```python
import numpy as np
np.array(test_dataset._extended_walker).nbytes
```
At its `316733528` byte,~316 megabytes` storing the full line and `20449080 ~20 megabytes`. I would say is a considerable size but it does remove a I/O operation each time we load the item. Does any other dataset take this much space in memory?

jiwidi:
> @jiwidi
> 
> > * `_getdict` function to return the phonems dictionary of the dataset
> 
> Do you mean the lexicon file such as `TEDLIUM.150K.dic` and `TEDLIUM.152K.dic`?
> `dict` is very overloaded word here, so I just want a clarification on what you mean.

Yes, that `.dic` file contains the phoneme dictionary of the dataset and is used sometimes. I want to add a function to the class to load and return it

torchaudio/datasets/tedlium.py
jiwidi: I think we should go with a hardcoded sample rate then, will update in a following commit

torchaudio/datasets/tedlium.py
jiwidi: Yes, I report the time of initiating the dataset and per se going over all files counting lines. I think that is quite fast and we shouldn't worry about it.

About the size it occupies in memory, I did a quick test for a comment from @vincentqb just on top of this one. I checked how much would the walker occupy if we were to store the full line and not the line identifier as we already loop it, saving some I/O when loading each item.

I went ahead and checked the size of the walker list for release3 (the biggest)
```python
import numpy as np
np.array(test_dataset._extended_walker).nbytes
```
At its `316733528 bytes,~316 megabytes` storing the full line and `20449080 ~20 megabytes` with just the line identifier. I would say is a considerable size but it does remove a I/O operation each time we load the item. Does any other dataset take this much space in memory?

torchaudio/datasets/tedlium.py
jiwidi: I found some inconsistency between sox and sox_io load functions:

https://github.com/pytorch/audio/blob/c692fe99ac353f0e2fc7b5ad7162f037a70a35fe/torchaudio/backend/sox_backend.py#L20-L28

https://github.com/pytorch/audio/blob/c692fe99ac353f0e2fc7b5ad7162f037a70a35fe/torchaudio/backend/sox_io_backend.py#L37-L43

`offset` vs `frame_offset`

Is this intended? This will cause some trouble when changing the backend if we make use of the parameter here.

mthrok: This is intended. "sox" backend is going to be deprecated in the next release and "sox_io" backend will become the default backend, then "sox" backend will be removed.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,159 @@
...
+        wave_path = os.path.join(path, "sph", fileid)
+        # Calculate indexes for start time and endtime
+        start_time = int(float(start_time) * 16000)
+        end_time = int(float(end_time) * 16000)
+        waveform, sample_rate = self.load_audio(
+            wave_path + self._ext_audio, frame_offset=start_time, num_frames=end_time - start_time
+        )
+        return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
+
+    def load_audio(self, path: str, frame_offset: int = 0, num_frames: int = -1) -> [Tensor, int]:
```
mthrok, line 127: I would do

`def load_audio(self, path:str, start_time:float, end_time:float, sample_rate:int = 16000)`

that way, we can delegate the responsibility of frame calculation into `load_audio` and `load_tedlium_item`'s responsibility becomes more focused on finding the right item/path.

mthrok: Also could you prefix the function name with underscore `_load_audio` to let users know it's not for public use?

torchaudio/datasets/tedlium.py
mthrok: > Yes, I report the time of initiating the dataset and per se going over all files counting lines. I think that is quite fast and we shouldn't worry about it.

I did a quick benchmark on my end and 

```python
import time

import torchaudio

torchaudio.set_audio_backend('sox_io')

n_rep = 100
t0 = time.monotonic()
for _ in range(n_rep):
    ds = torchaudio.datasets.TEDLIUM("../dataset/tedlium", release="release3", download=False)
t1 = time.monotonic()
print(len(ds._walker))
print((t1 - t0) / n_rep)
```

</details>

```
$python foo.py
268263
0.21009648161940275
```

I did not get as fast as micro seconds, but it was only 0.2 seconds so I agree that speed is not an issue here.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,159 @@
...
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import (
+    download_url,
+    extract_archive,
+    walk_files,
+)
+from collections import namedtuple
+
+
+RELEASE = "release1"  # Default release
```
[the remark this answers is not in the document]
jiwidi: done in the last commit, :+1:

torchaudio/datasets/tedlium.py
jiwidi: done in the last commit, :+1:

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,182 @@
...
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(" ", 6)
+
+        wave_path = os.path.join(path, "sph", fileid)
+        waveform, sample_rate = self.__load_audio__(
+            wave_path + self._ext_audio, start_time=start_time, end_time=end_time
+        )
+
+        return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
+
+    def __load_audio__(self, path: str, start_time: float, end_time: float, sample_rate: int = 16000) -> [Tensor, int]:
```
mthrok, line 139: double underscore prefix and suffix are reserved for specific purposes in Python. Please use single underscore prefix (and no suffix)

> \_\_double_leading_and_trailing_underscore__: "magic" objects or attributes that live in user-controlled namespaces. E.g. \_\_init__, \_\_import__ or \_\_file__. Never invent such names; only use them as documented.

Reference: [PEP8](https://www.python.org/dev/peps/pep-0008/)
This page summarizes the naming convention for function and variable https://dbader.org/blog/meaning-of-underscores-in-python

torchaudio/datasets/tedlium.py
mthrok: Regarding the memory consumption, I talked with @cpuhrsch. 20 megabytes is not a big deal. However be aware that when using multiple data loader worker process, each process will consume the same amount of memory.

Also DataLoader has memory leakage issue https://github.com/pytorch/pytorch/issues/13246 .

torchaudio/datasets/tedlium.py
jiwidi: Oh thanks, I didnt knew!

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,182 @@
...
+            stm_path = os.path.join(self._path, "stm", file + ".stm")
+            with open(stm_path) as f:
+                l = len(f.readlines())
+                self._walker.extend((file, line) for line in range(l))
+        # Read phoneme dictionary
+        dict_path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["dict"])
+        self.phoneme_dict = {}
+        with open(dict_path, "rb") as f:
+            for line in f.readlines():
+                content = line.decode("utf-8").strip("\n").split(" ", 1)
```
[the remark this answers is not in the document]
jiwidi: Yeah makes more sense, I was trying to come with a error-proof solution like this but couldn't think of it.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,182 @@
...
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _RELEASE_CONFIGS[release]["checksum"]
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        self._walker = []
+
+        # Create walker for all samples
+        for file in walk_files(self._path, suffix=".stm", prefix=False, remove_suffix=True):
```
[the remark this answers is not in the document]
jiwidi: That is true, I used the walker since I saw it was used in other datasets but listing all stm files will do the job as all of them are supposed to be inside the same folder.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,182 @@
...
+class TEDLIUM(Dataset):
+    """
+    Create a Dataset for Tedlium. Each item is a tuple of the form:
+    [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
+    """
+
+    def __init__(
+        self, root: str, release: str = "release1", subset: str = None, download: bool = False, audio_ext=".sph"
+    ) -> None:
+        """Constructor for TEDLIUM dataset
```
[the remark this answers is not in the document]
jiwidi: I just checked with the other datasets, they have a similar docstring under the class description:
We have
```
    """
    Create a Dataset for Tedlium. Each item is a tuple of the form:
    [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
    """
```

Librispeech:
```
    """
    Create a Dataset for LibriSpeech. Each item is a tuple of the form:
    waveform, sample_rate, utterance, speaker_id, chapter_id, utterance_id
    """
```

Are you sure you want to move the big constructor docstring to class-level?

docs/source/datasets.rst
```diff
@@ -81,6 +81,13 @@ SPEECHCOMMANDS
   :special-members:
 
 
+TEDLIUM
+~~~~~~~~~~~~~~
+
+.. autoclass:: TEDLIUM
+  :members: __getitem__
+  :special-members: get_phoneme_dict
```
jiwidi, line 89: @mthrok What does special-members mean here? I interpretate it as extra functions to include in the docs? Thats why I included get_phoneme_dict

torchaudio/datasets/tedlium.py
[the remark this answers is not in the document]
mthrok: > Are you sure you want to move the big constructor docstring to class-level?

The fact that the other dataset lacks the documentation for the constructor is not user-friendly.
It is better if each Dataset explains what options a user can change.
I will update the other dataset documentation.

docs/source/datasets.rst
mthrok: These are Sphinx's directive. Check out [their documentation](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html)

 - `:members:` is where you list the members you want document
 - `:special-members:` is where you list the special methods like `__init__`, `__len__`, `__getitem__` etc...

I think the other documentations are wrong, (`__getitem__` should be under `:special-member:` but it will not show up either way because they don't have a docstring.)

I think you can just do `.. autoclass:: TEDLIUM` and the rest (`get_phoneme_dict`) will be handled.

You can build the documentation and check how the resulting documentation looks like.

```
cd docs
pip install -r requirements.txt
make html
# open ./build/html/index.html
```

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,182 @@
...
+
+        Args:
+            root (str): Path containing dataset or target path where its downloaded if needed
+            release (str, optional): TEDLIUM identifier (release1,release2,release3). Defaults to RELEASE.
+            subset (str, optional): Subset of data(train,test,dev) supported for release 1,2. Defaults to Train/None.
```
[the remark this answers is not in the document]
mthrok: Could you add value check and raise a `ValueError` if it's not the right one for the given release?

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,179 @@
+import os
+from typing import Tuple
+
+import torchaudio
+from torch import Tensor
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import (
+    download_url,
+    extract_archive,
+    walk_files,
```
mthrok, line 10: You can remove `walk_files` from import too.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,179 @@
...
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _RELEASE_CONFIGS[release]["checksum"]
+                    download_url(url, root, hash_value=checksum)
+                extract_archive(archive)
+
+        # Create walker for all samples
+        self._walker = []
+        stm_path = os.path.join(self._path, "stm")
+        for file in os.listdir(stm_path):
```
mthrok, line 99: Can you sort the file? Different OS returns files in different order.

```suggestion
        for file in sorted(os.listdir(stm_path)):
```

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Read phoneme dictionary
+        dict_path = os.path.join(root, folder_in_archive, _RELEASE_CONFIGS[release]["dict"])
+        self.phoneme_dict = {}
+        with open(dict_path, "r", encoding="utf-8") as f:
+            for line in f.readlines():
+                content = line.strip().split(maxsplit=1)
+                self.phoneme_dict[content[0]] = content[1:]  # content[1:] can be empty list
+
+    def load_tedlium_item(self, fileid: str, line: int, path: str) -> Tedlium_item:
```
mthrok, line 115: Can you also mark this function as private?

```suggestion
    def _load_tedlium_item(self, fileid: str, line: int, path: str) -> Tedlium_item:
```

jiwidi:
> The code looks much better now.
> 
> @jiwidi What else is left for this PR?

Test! If we are okay with the logic for iterating files I'll go ahead and create some test based on the last updates I saw you guys do on the other dataset test.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+        """
+        transcript_path = os.path.join(path, "stm", fileid)
+        with open(transcript_path + ".stm") as f:
+            transcript = f.readlines()[line]
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = transcript.split(" ", 6)
+
+        wave_path = os.path.join(path, "sph", fileid)
+        waveform, sample_rate = self._load_audio(wave_path + self._ext_audio, start_time=start_time, end_time=end_time)
+
+        return Tedlium_item(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)
```
vincentqb, line 163: I can't find the discussion about NamedTuple, so I'm leaving a new comment here :)

There has been prior discussion about changing the type of data points (and in particular using NamedTuple), see [internal document](https://fb.quip.com/vlWwA35cmq0t). Though there may be value in redefining what a data point is, I feel the discussion would delay this pull request, so I would avoid it for now.

jiwidi: I can't access the document but regardless I have no problem with changing it back to the previous version.

We can delay this NamedTuple to a PR that will change multiple Datasets/files and not do 1 by 1 and get inconsistent code. What do you think?

mthrok: I oppose to using tuple here. 6 heterogeneous items as tuple give very bad user experience and make code harder to read/maintain. Specifically named tuple is compatible with tuple, there is virtually no disadvantage of using named tuple here.

vincentqb: The main disadvantage of using namedtuple here is (1) inconsistent behavior with other datasets in audio and elsewhere, (2) would make the user leverage an API that could change in the near future, (3) will slow down this particular pull request.

The reason we did not standardize around NamedTuple in the past was that it didn't allow for easy introspection of keys, and each dataset would have its own DataPoint NamedTuple. Please see [internal document](https://fb.quip.com/vlWwA35cmq0t) for more discussions.

We are also using tuples with many items with other torchaudio datasets. That being said, we can revisit these design choices. Thoughts @zhangguanheng66 @fmassa @cpuhrsch ?

jiwidi:
Hi again! @vincentqb and @mthrok do I have a thumbs up in the way we iterate files? I would like to continue working on the test and finish the PR (I have more ideas to PR for audio :))

mthrok:
> Hi again! @vincentqb and @mthrok do I have a thumbs up in the way we iterate files? I would like to continue working on the test and finish the PR (I have more ideas to PR for audio :))

Hi @jiwidi 

Sorry for the delay. Yes, the way it files are visited looks fine for me. (although I would not call it `_walker`, it will be better to pick a name that describes what the list contains.) Feel free to ping me on how to write the test.

torchaudio/datasets/tedlium.py
vincentqb: > > (1) inconsistent behavior with other datasets in audio and elsewhere
> 
> Can you elaborate this? Each dataset returns different number of items, so it does not look like they can be swapped easily from the beginning. And using NamedTuple with consistent key name will make it possible to give consistent user experience.

I was really just referring to using NamedTuple in this particular pull request. Moving toward consistently using NamedTuple in torchaudio/text/vision is fine by me.

> > (2) would make the user leverage an API that could change in the near future
> 
> Not sure what this means. Can you elaborate?

Since we have not, as a team (torchaudio/text/vision), revisited the decision of using Tuple, using NamedTuple here exposes us to changing the API again would we settle on another option.

> > (3) will slow down this particular pull request.
> 
> It's already in NamedTuple so, it seems that changing it tuple seems to be slowing down the process.

But we should avoid using NamedTuple until we, as a team, have made a decision on the API for data point. The current decision had been Tuple, but I'm personally fine with revisiting this past decision.

> > The reason we did not standardize around NamedTuple in the past was that it didn't allow for easy introspection of keys
> 
> When is it necessary or useful to perform key introspection?
>
> > each dataset would have its own DataPoint NamedTuple.
> 
> That sounds the right approach. Because each dataset returns the different number of items of different types. They should be typed differently.

I don't feel too strongly about either points. I am really just listing reasons that we left in the [internal document](https://fb.quip.com/vlWwA35cmq0t). :)

jiwidi:
> > Hi again! @vincentqb and @mthrok do I have a thumbs up in the way we iterate files? I would like to continue working on the test and finish the PR (I have more ideas to PR for audio :))
> 
> Hi @jiwidi
> 
> Sorry for the delay. Yes, the way it files are visited looks fine for me. (although I would not call it `_walker`, it will be better to pick a name that describes what the list contains.) Feel free to ping me on how to write the test.

Mm I have been copying the https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/datasets/librispeech_test.py but for TEDLIUM, I have a doubt here, how will you save the whitenoise into '.sph' files? As far as I know is not supported yet(?)

jiwidi:
> > > Hi again! @vincentqb and @mthrok do I have a thumbs up in the way we iterate files? I would like to continue working on the test and finish the PR (I have more ideas to PR for audio :))
> > 
> > 
> > Hi @jiwidi
> > Sorry for the delay. Yes, the way it files are visited looks fine for me. (although I would not call it `_walker`, it will be better to pick a name that describes what the list contains.) Feel free to ping me on how to write the test.
> 
> Mm I have been copying the https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/datasets/librispeech_test.py but for TEDLIUM, I have a doubt here, how will you save the whitenoise into '.sph' files? As far as I know is not supported yet(?)

Friendly ping @mthrok :)

mthrok:
> > Mm I have been copying the https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/datasets/librispeech_test.py but for TEDLIUM, I have a doubt here, how will you save the whitenoise into '.sph' files? As far as I know is not supported yet(?)
> 
> Friendly ping @mthrok :)

Hi @jiwidi 

Thanks for singing me again and sorry for missing your reply in the first place.
There are two things to consider here.
1. Did you try saving audio data in `sph` format? With https://github.com/pytorch/audio/pull/871, I believe you can save audio with `sph` format. (If not, let me know.) however,
2. In the case of `librispeech_test.py` and the others, we emulate the dataset with wav format. This is because wav is easy to handle and does not require any third-party library to load (although we use `scipy` for simplicity, but we do so because we believe `scipy` does the correct job to load wav file we can reference.). In unit-test, it is ideal to only use the module being tested. In the case of dataset implementation, we want to only use dataset module as much as possible, and we do not want to use I/O module in the test. Otherwise when test fails, we do not know immediately if it is due to I/O module or dataset module. (of course, the dataset implementation itself uses `torchaudio.load`, so the perfect separation at test time is not achieved without monkey-patching `torchaudio.load` function, but I think that's too much)

So because of 2., it is preferable if we can test the dataset implementation with WAV format, even though the real dataset is in SPH. (and I believe this is also a valid use case for end users to convert the entire dataset into WAV.). For that purpose we make dataset implementation to be able to change the target extension. So can you write your test with `"audio_ext='.wav'"` when instantiating Dataset?
If you strongly thing this dataset also should be tested with `.sph` format, then you can put the whole test logic in a helper method and make the test change format too.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+        "data_path": "data/",
+        "subset": None,
+        "supported_subsets": [None],
+        "dict": "TEDLIUM.152k.dic",
+    },
+}
+
+Tedlium_item = namedtuple(
+    "Tedlium_item", ["waveform", "sample_rate", "transcript", "talk_id", "speaker_id", "identifier"]
+)
```
mthrok, line 46: @jiwidi 

Regarding the `namedtuple`, I talked with @cpuhrsch and I will conduct performance check. This one is tricky and could take some time, So I'm sorry to ask but can you revert it back to a regular `tuple` for the sake of merging this PR first? I feel so sorry the process is taking so long.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+            path (str): Path to audio file
+            start_time (int, optional): Time in seconds where the sample sentence stars
+            end_time (int, optional): Time in seconds where the sample sentence finishes
+
+        Returns:
+            [Tensor, int]: Audio tensor representation and sample rate
+        """
+        start_time = int(float(start_time) * 16000)
+        end_time = int(float(end_time) * 16000)
+        return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
```
mthrok, line 179: I just realized that this will only work with `sox_io` backend. Could you add fall-back for the other backend?

```
if torchaudio.get_audio_backen() == 'sox_io':
    return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
return torchaudio.load(path)[:, start_time:end_time] 
```

torchaudio/datasets/tedlium.py
jiwidi: No problem!

jiwidi:
> > > Mm I have been copying the https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/datasets/librispeech_test.py but for TEDLIUM, I have a doubt here, how will you save the whitenoise into '.sph' files? As far as I know is not supported yet(?)
> > 
> > 
> > Friendly ping @mthrok :)
> 
> Hi @jiwidi
> 
> Thanks for singing me again and sorry for missing your reply in the first place.
> There are two things to consider here.
> 
> 1. Did you try saving audio data in `sph` format? With #871, I believe you can save audio with `sph` format. (If not, let me know.) however,
> 2. In the case of `librispeech_test.py` and the others, we emulate the dataset with wav format. This is because wav is easy to handle and does not require any third-party library to load (although we use `scipy` for simplicity, but we do so because we believe `scipy` does the correct job to load wav file we can reference.). In unit-test, it is ideal to only use the module being tested. In the case of dataset implementation, we want to only use dataset module as much as possible, and we do not want to use I/O module in the test. Otherwise when test fails, we do not know immediately if it is due to I/O module or dataset module. (of course, the dataset implementation itself uses `torchaudio.load`, so the perfect separation at test time is not achieved without monkey-patching `torchaudio.load` function, but I think that's too much)
> 
> So because of 2., it is preferable if we can test the dataset implementation with WAV format, even though the real dataset is in SPH. (and I believe this is also a valid use case for end users to convert the entire dataset into WAV.). For that purpose we make dataset implementation to be able to change the target extension. So can you write your test with `"audio_ext='.wav'"` when instantiating Dataset?
> If you strongly thing this dataset also should be tested with `.sph` format, then you can put the whole test logic in a helper method and make the test change format too.

Dont worry, you've been very helpful. I will test it late today and get back if there are any problems, otherwise a commit will follow :)

I think I can make it work with wav files as well but since this is the only interaction of the library with sph files I see value in keeping the test with sph format as it will serve as a check for this functionality. If we check for sph support in other test then I have no preference with wav or sph.

mthrok:
> I think I can make it work with wav files as well but since this is the only interaction of the library with sph files I see value in keeping the test with sph format as it will serve as a check for this functionality. If we check for sph support in other test then I have no preference with wav or sph.

load/save/info functions on `.sph` format are tested separately. (check out #871)  

If you decide to test sph, then following is (only the gist of) my recommendation.

```python
def _generate_dataset(root_dir, audio_ext, dataset_version, <other parameters, such as dataset versions>) -> List[tuple]:
    # Generates dataset with the given extension
    for foo in whatever:
        ...
        # Switch saving function
        if audio_ext == "wav":
            save_wav(file_path, data, sample_rate)  # use scipy-based test utility for wav
        else:
            torchaudio.save(file_path, data, sample_rate)  # use sox-io backend
                                                            # assuming that the backend is properly set.
                                                            # See the bellow backend="sox_io" for this.
    # Returns expected data


# Note: Probably better to define different test class for different TEDLIUM versions.
class TestTedliumWav(TempDirMixin, TorchaudioTestCase):
    backend = "default"  # <- this test should work on all supported platforms

    @classmethod
    def setUp(cls):
        cls.root_dir = cls.get_base_temp_dir()
        _generate_dataset(cls.root_dir, "wav", ...)

    # define the test
    def testFoo(self):
        ...


class TestTedliumSph(TempDirMixin, TorchaudioTestCase):
    backend = "sox_io"  # <- this test will run only on the platforms where "sox_io" backend is available.

    @classmethod
    def setUp(cls):
        cls.root_dir = cls.get_base_temp_dir()
        _generate_dataset(cls.root_dir, "sph", ...)

    # define the test
    def testFoo(self):
        ...
```


> edit: saving sph files work perfectly with `torchaudio.save()` sorry I didnt even checked it and thought it was only for wavs

glad it works :)

jiwidi:
> > I think I can make it work with wav files as well but since this is the only interaction of the library with sph files I see value in keeping the test with sph format as it will serve as a check for this functionality. If we check for sph support in other test then I have no preference with wav or sph.
> 
> load/save/info functions on `.sph` format are tested separately. (check out #871)
> 
> If you decide to test sph, then following is (only the gist of) my recommendation.
> 
> ```python
> def _generate_dataset(root_dir, audio_ext, dataset_version, <other parameters, such as dataset versions>) -> List[tuple]:
>     # Generates dataset with the given extension
>     for foo in whatever:
>         ...
>         # Switch saving function
>         if audio_ext == "wav":
>             save_wav(file_path, data, sample_rate)  # use scipy-based test utility for wav
>         else:
>             torchaudio.save(file_path, data, sample_rate)  # use sox-io backend
>                                                             # assuming that the backend is properly set.
>                                                             # See the bellow backend="sox_io" for this.
>     # Returns expected data
> 
> 
> # Note: Probably better to define different test class for different TEDLIUM versions.
> class TestTedliumWav(TempDirMixin, TorchaudioTestCase):
>     backend = "default"  # <- this test should work on all supported platforms
> 
>     @classmethod
>     def setUp(cls):
>         cls.root_dir = cls.get_base_temp_dir()
>         _generate_dataset(cls.root_dir, "wav", ...)
> 
>     # define the test
>     def testFoo(self):
>         ...
> 
> 
> class TestTedliumSph(TempDirMixin, TorchaudioTestCase):
>     backend = "sox_io"  # <- this test will run only on the platforms where "sox_io" backend is available.
> 
>     @classmethod
>     def setUp(cls):
>         cls.root_dir = cls.get_base_temp_dir()
>         _generate_dataset(cls.root_dir, "sph", ...)
> 
>     # define the test
>     def testFoo(self):
>         ...
> ```
> 
> > edit: saving sph files work perfectly with `torchaudio.save()` sorry I didnt even checked it and thought it was only for wavs
> 
> glad it works :)

I just pushed the test for tedlium. I saw the sph load is already tested elsewhere as you mentioned so I kept the test with `.wav` files for simplicity and standard among other test.

For this new commit I also changed the phoneme dictionary to be only read if the user actually calls the `get_phoneme_dict` and this also makes it easier when testing. 

The testing is very similar to the librispeech one, I try to recreate the 3 releases folder structure in a temporary folder and put some white noise there. Instantiate tedlium datasets and check all the data is well retrieved when iterating them.

jiwidi:
@mthrok is it normal this many unittest fail due to failure to download/connectivity issues?

Test run on my side, but we have to assure they also run in the CI pipeline.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def get_phoneme_dict(self):
+        """Returns the phoneme dictionary of a TEDLIUM release
+
+        Returns:
+            dictionary: Phoneme dictionary for the current tedlium release
+        """
+        # Read phoneme dictionary
+        if not hasattr(self, "phoneme_dict"):
+            self.phoneme_dict = {}
```
mthrok, line 201: This design pattern might leave users questions like "Which one should I use, `phoneme_dict` or `get_phoneme_dict`?", "What are the differences?"

The following techniques are often used to improve this;
1. Use internal attribute name like `_phoneme_dict`.
    This does not prevent users to access the underlying object directly, but accessing an attribute prefixed with underscore clearly sends a message that user code is not doing the right thing. so it's okay.
2. Use `property` decorator.
    Lazy-initialization often fits well in `property` attribute.
    Something like this;

    ```python
    @property
    def phoneme_dict(self):
        if self._phoneme_dict is None:
            self._phoneme_dict = dict()
            # Fill the dictionary
        return self._phoneme_dict.copy()  # see the bellow for `copy`
    ```

Also instead of checking if it's initialized with `hasattr`, it's more straight forward to initialize the actual attribute object in constructor as `None` such as `self._phone_dict = None` and perform the check as `if self._phone_dict is None:`

Another concern is returning dictionary reference. The user code might accidentally modify the dictionary which can lead to a very subtle bug. Maybe it's better to return a copy of the dictionary. as `return self._phoneme_dict.copy()` (however this is still shallow copy so the value points to the original lists and user code can still modify the lists. maybe changing it to tuple might prevent such unintended modification, like `self.phoneme_dict[content[0]] = tuple(content[1:])`)

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -0,0 +1,93 @@
...
+                talk_id,
+                speaker_id,
+                identifier,
+            )
+            cls.samples.append(sample)
+
+    @classmethod
+    def tearDownClass(cls):
+        # In case of test failure
+        tedlium.TEDLIUM._ext_audio = ".flac"
```
mthrok, line 76: This hack is required in the case of `librispeech_test`, because the dataset does not receive any audio extension in the constructor. `TEDLIUM` simply accepts "audio_ext", so you do not need to do this (neither `tedlium.TEDLIUM._ext_audio = ".wav"`) you can just do `dataset = tedium.TEDLIUM(self.root_dir, audio_ext=".wav")`

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def test_tedlium(self):
+        tedlium.TEDLIUM._ext_audio = ".wav"
+        dataset = tedlium.TEDLIUM(self.root_dir)
```
mthrok, line 80: Since the mocked datasets are generated for all the releases, how about testing all the releases, (as separate test case (method))?

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -0,0 +1,93 @@
...
+            trans_filename = f"{release}.stm"
+            trans_path = os.path.join(os.path.join(release_dir, "stm"), trans_filename)
+            with open(trans_path, "w") as f:
+                f.write("".join(UTTERANCES))
+
+        # Create a samples list to compare with
+        for i, utterance in enumerate(UTTERANCES):
+            talk_id, _, speaker_id, start_time, end_time, identifier, transcript = utterance.split(" ", 6)
+            start_time = int(float(start_time)) * 16000
+            end_time = int(float(end_time)) * 16000
```
mthrok, line 62: Since there is a variable `sample_rate = 16000`, defined above, why not use it here?
If something has to be changed in the future, it will be easier that way.

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -0,0 +1,93 @@
...
+    samples = []
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        cls.root_dir = dataset_dir = os.path.join(cls.root_dir, "tedlium")
+        os.makedirs(dataset_dir, exist_ok=True)
+        sample_rate = 16000  # 16kHz
+        seed = 0
+        data = get_whitenoise(sample_rate=sample_rate, duration=10.00, n_channels=1, dtype="float32", seed=seed)
```
mthrok, line 36: Mocked audio data has to be generated for each files being mocked, with different seed value, otherwise all the files generated are exactly same (for the sake or test reproducibility, test helper functions are implemented deterministic by default, that is with fixed seed value) and we do not know if the dataset implementation is traversing the files in expected order.

mthrok: oh I get it now. it's one file with different segments.
Can we have multiple audio files per release?
Testing for cases with multiple objects often reveal a bug, so it's a good practice.
Here, we would like to make sure that `os.listdir` function (which returns items in OS-specific way) does not change the result across OSs. (though our test is broken for non-linux environment, which I have to fix soon.)

mthrok:
> @mthrok is it normal this many unittest fail due to failure to download/connectivity issues?
> 
> Test run on my side, but we have to assure they also run in the CI pipeline.

Hi @jiwidi 

Yes, it's broken for Windows and macOS and I have to fix it at some point. These failures have nothing to do with your work.

torchaudio/datasets/tedlium.py
jiwidi: Really good points, a commit with all the suggestions will follow :+1:

test/torchaudio_unittest/datasets/tedlium_test.py
jiwidi: good catch

test/torchaudio_unittest/datasets/tedlium_test.py
jiwidi: Just pushed a new version with the previous feedback and including a separate test for every release. More test never hurt and as you say we could have miss some failures only testing for the 1st release.

torchaudio/datasets/tedlium.py
vincentqb: The current sox backend also support reading file segment. Not soundfile? If the datasets needs to be aware of the backend, then we should see how we could evolve the backends also so that this does not need to happen. Thoughts?

jiwidi: `torchaudio.load(path)[:, start_time:end_time] ` will still work for both backends as far as I understand. But @mthrok pointed out is more efficient to use the `torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)` if sox_io is enabled.

We either:

* Add support for other backends with same parameter name so the optimized loading can be used with all (this could be better for the long term, unifying 

* Use the slicing operator and support all backends right away.

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -0,0 +1,124 @@
...
+            save_wav(path, data, sample_rate)
+
+            trans_filename = f"{release}.stm"
+            trans_path = os.path.join(os.path.join(release_dir, "stm"), trans_filename)
+            with open(trans_path, "w") as f:
+                f.write("".join(UTTERANCES))
+
+            # Create a samples list to compare with
+            cls.samples[release] = []
+            for i, utterance in enumerate(UTTERANCES):
```
NikeNano, line 65:
```suggestion
            for  utterance in UTTERANCES:
```

I don't think the enumerate is used.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+    def __len__(self) -> int:
+        """DTEDLIUM dataset custom function overwritting len default behaviour.
+
+        Returns:
+            int: TEDLIUM dataset length
+        """
+        return len(self._filelist)
+
+    @property
+    def get_phoneme_dict(self):
```
mthrok, line 195: `@property` is accessed like an attribute and it is invoked without parenthesis, so `get_phoneme_dict` is strange. Just `phoneme_dict` makes more sense.

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+        Returns:
+            dictionary: Phoneme dictionary for the current tedlium release
+        """
+        # Read phoneme dictionary
+        if not self._phoneme_dict:
+            self._phoneme_dict = {}
+            with open(self.dict_path, "r", encoding="utf-8") as f:
+                for line in f.readlines():
+                    content = line.strip().split(maxsplit=1)
+                    self._phoneme_dict[content[0]] = content[1:]  # content[1:] can be empty list
```
mthrok, line 207: What do you think of changing the turning this to tuple to prevent the modification on user code?
Does it affect user experience?

```suggestion
                    self._phoneme_dict[content[0]] = tuple(content[1:])  # content[1:] can be empty list
```

torchaudio/datasets/tedlium.py
mthrok: > The current sox backend also support reading file segment. Not soundfile? If the datasets needs to be aware of the backend, then we should see how we could evolve the backends also so that this does not need to happen. Thoughts?

We have already planned to resolve this in 0.9.0 release. See #903

torchaudio/datasets/tedlium.py
jiwidi: good idea

test/torchaudio_unittest/datasets/tedlium_test.py
jiwidi: good catch!

jiwidi:
@mthrok new commit with the PR comments suggestion, what do you think of the state of this PR? Doesnt seem much left to do (?)

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+
+        Returns:
+            dictionary: Phoneme dictionary for the current tedlium release
+        """
+        # Read phoneme dictionary
+        if not self._phoneme_dict:
+            self._phoneme_dict = {}
+            with open(self.dict_path, "r", encoding="utf-8") as f:
+                for line in f.readlines():
+                    content = line.strip().split(maxsplit=1)
```
mthrok, line 206: Was `maxsplit=1` always here?

I thought this is splitting into the list of phonemes, but with `maxsplit=1` it is a list of one string which has all the phonemes.

with `maxsplit=1`, it's 

```python
content = ['dani', 'D AA N IY']
self._phenome_dict = {'dani': ('D AA N IY', ), ...}
```

whereas without `maxsplit`, it's

```python
content = ['dani', 'D', 'AA', 'N', 'IY']`,
self._phenome_dict = {'dani': ('D', 'AA', 'N', 'IY'), ...}
```

Which one do you intend? I thought it's the later.

mthrok (commented):
Hi @jiwidi 

I had one more question about the form of `phoneme_dict`. Please check the comment.
It's also a good idea to add test for it but since this PR has got super long, so it's okay. Other than that, I think we are ready to merge. Thanks for sticking with the such a rigorous review. I enjoyed our interactions. :)

torchaudio/datasets/tedlium.py
```diff
@@ -0,0 +1,208 @@
...
+        Args:
+            path (str): Path to audio file
+            start_time (int, optional): Time in seconds where the sample sentence stars
+            end_time (int, optional): Time in seconds where the sample sentence finishes
+
+        Returns:
+            [Tensor, int]: Audio tensor representation and sample rate
+        """
+        start_time = int(float(start_time) * 16000)
+        end_time = int(float(end_time) * 16000)
```
mthrok, line 168: Can you use `sample_rate` variable here instead of hard coded `16000`?

torchaudio/datasets/tedlium.py
jiwidi: It was always there and the intent is to have the first option you mention:

```
content = ['dani', 'D AA N IY']
self._phenome_dict = {'dani': ('D AA N IY', ), ...}
```

I'm used to seeing phoneme as a one string only instead of a list as second option, but this doesn't mean maybe other people expect the second option. But at least this is an easy change from the user perspective.

jiwidi:
@mthrok I just pushed a new commit where I include test for the dictionary for every release as well. More test never hurt :) 

It also changes the value of the dictionary with the second option you described above and fixes wrong naming in a private variable.

I also enjoyed the interaction here very much, very good coding practices review you left me here :100:

mthrok (approved):
@jiwidi 

Thanks for your work.

> fixes wrong naming in a private variable.

Grad you caught this, I did not notice this :)

> I also enjoyed the interaction here very much, very good coding practices reviews you left me here 💯

Thanks for the nice words. Let me know in future if there is something I can help you. ;)

mthrok:
@jiwidi 

Sorry, I just noticed that tedlium dataset test fails on style test. Do you have time to fix it? I tried to modify the PR but I cannot push to your branch. If you are done with it, let me know, I will merge it first and create a follow-up PR>

jiwidi:
> @jiwidi
> 
> Sorry, I just noticed that tedlium dataset test fails on style test. Do you have time to fix it? I tried to modify the PR but I cannot push to your branch. If you are done with it, let me know, I will merge it first and create a follow-up PR>

Just pushed a new commit that should fix it

mthrok:
@jiwidi 

Thank you so much!

PR #885, opened 2020-08-23
Fix incorrect extension parsing

tejasvi:
Existing implementation doesn't work for FLAC format.


mthrok:
I will follow up with the failing MP3 test. Thanks for the contribution!

vincentqb:
linking to #886

PR #888, opened 2020-08-24
fix resample type hint

faroit:
As indicated in the docstrings, the sample rates would need to be `float`. I also discovered that using `int` can result in an error

```
  File "/Users/faro/repositories/open-unmix-pytorch-dev/env/lib/python3.7/site-packages/torch/nn/modules/module.py", line 722, in _call_impl
    result = self.forward(*input, **kwargs)
  File "/Users/faro/repositories/open-unmix-pytorch-dev/env/lib/python3.7/site-packages/torchaudio/transforms.py", line 596, in forward
    waveform = kaldi.resample_waveform(waveform, self.orig_freq, self.new_freq)
  File "/Users/faro/repositories/open-unmix-pytorch-dev/env/lib/python3.7/site-packages/torchaudio/compliance/kaldi.py", line 930, in resample_waveform
    window_width, lowpass_cutoff, lowpass_filter_width, device, dtype)
  File "/Users/faro/repositories/open-unmix-pytorch-dev/env/lib/python3.7/site-packages/torchaudio/compliance/kaldi.py", line 807, in _get_LR_indices_and_weights
    assert lowpass_cutoff < min(orig_freq, new_freq) / 2
RuntimeError: Integer division of tensors using div or / is no longer supported, and in a future release div will perform true division as in Python 3. Use true_divide or floor_divide (// in Python) instead.

```

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -567,8 +567,8 @@ class Resample(torch.nn.Module):
     """
 
     def __init__(self,
-                 orig_freq: int = 16000,
-                 new_freq: int = 16000,
+                 orig_freq: float = 16000,
+                 new_freq: float = 16000,
                  resampling_method: str = 'sinc_interpolation') -> None:
         super(Resample, self).__init__()
         self.orig_freq = orig_freq


vincentqb:
The convention we've been using so far as been `int` as you can see for instance in MFCC transform. To maintain the same behavior, we should replace the division by `floor_divide`. Do you want to open a pull request for this?

faroit:
> The convention we've been using so far as been int as you can see for instance in MFCC transform

I actually used `int` for decades but recently learned that float is more appropriate and mathematically correct. It is also often used in resamplers e.g, in [sox](https://github.com/WavX/soxr/blob/master/src/soxr.h#L101).

> To maintain the same behavior, we should replace the division by floor_divide. Do you want to open a pull request for this?

yes, but in any case, either the [docstring](https://github.com/pytorch/audio/blob/d38aab425a9e4e6522d94b92d5f3efa6069e81f8/torchaudio/transforms.py#L570) or the type hint of thins function would need to changed as well.

vincentqb:
Thanks for the feedback. I'd be happy to update the code to floats, #891. However, this may interfere with C++ extensions when saving.

The current behavior of this particular assertion (in torchaudio) is to do integer division, so I'd expect to use `floor_division`. That being said, this code is ported from Kaldi: do you know if the original intent was floor or true division?

mthrok:
This has been resolved in #1499 (part of #1487)

PR #893, opened 2020-08-28
Add conv-tasnet model

mthrok:
See https://github.com/pytorch/audio/issues/897


examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,250 @@
...
+        for layer in self.conv_layers:
+            residual, skip = layer(feats)
+            if residual is not None:  # the last conv layer does not produce residual
+                feats = feats + residual
+            output = output + skip
+        output = self.output_layer(output)
+        return output.view(batch_size, self.num_speakers, self.input_dim, -1)
+
+
+class ConvTasNet(torch.nn.Module):
```
vincentqb, line 125: Paper reference?

mthrok: done

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
+"""Implements Conv-TasNet with building blocks of it."""
+
+from typing import Tuple, Optional
+
+import torch
+
+
+class ConvBlock(torch.nn.Module):
```
vincentqb, line 8: not in [torchaudiomodels](https://github.com/pytorch/audio/tree/master/torchaudio/models)?

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        self.decoder = torch.nn.ConvTranspose1d(
+            in_channels=enc_num_feats,
+            out_channels=1,
+            kernel_size=enc_kernel_size,
+            stride=self.enc_stride,
+            padding=self.enc_stride,
+            bias=False,
+        )
+
+    def _pad_input(self, input: torch.Tensor) -> Tuple[torch.Tensor, int]:
```
vincentqb, line 249: This seems more general and reusable than just for this model?

mthrok: I do not think the assumption of `padding == filter_width // 2` to be generic enough.
We can do that kind of refactoring as needed once we have another model that requires the same kind of padding.

examples/source_separation/conv_tasnet/model.py
mthrok: I will handle that in the later PR.

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        kernel_size: int,
+        padding: int,
+        dilation: int = 1,
+        causal: bool = False,
+        no_residual: bool = False,
+    ):
+        super().__init__()
+
+        if causal:
+            raise NotImplementedError("causal=True is not implemented")
```
vincentqb, line 39: Why having this in the first place?

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        num_sources: int,
+        kernel_size: int,
+        num_feats: int,
+        num_hidden: int,
+        num_layers: int,
+        num_stacks: int,
+        causal: bool = False,
+    ):
+        if causal:
+            raise NotImplementedError("causal=True is not implemented")
```
vincentqb, line 115: same

examples/source_separation/conv_tasnet/model.py
vincentqb: Is there a way to factor this code as a static-like method then so as to decouple it from the rest?

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        msk_num_feats: int = 128,
+        msk_num_hidden_feats: int = 512,
+        msk_num_layers: int = 8,
+        msk_num_stacks: int = 3,
+        causal: bool = False,
+    ):
+        super().__init__()
+
+        if causal:
+            raise NotImplementedError("causal=True is not implemented")
```
vincentqb, line 216: same

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        input_dim: int,
+        num_sources: int,
+        kernel_size: int,
+        num_feats: int,
+        num_hidden: int,
+        num_layers: int,
+        num_stacks: int,
+        causal: bool = False,
+    ):
+        if causal:
```
cpuhrsch, line 114: Do you want to remove the flag and just add this to the docs?

examples/source_separation/conv_tasnet/model.py
vincentqb: you can just move this into the doc

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+                        padding=multi,
+                        causal=causal,
+                        # The last ConvBlock does not need residual
+                        no_residual=(l == (num_layers - 1) and s == (num_stacks - 1)),
+                    )
+                )
+                self.receptive_field += (
+                    kernel_size if s == 0 and l == 0 else (kernel_size - 1) * multi
+                )
+        self.output_layer = torch.nn.Sequential(
```
cpuhrsch, line 148: minor nit: You could turn this into control flow in the forward function instead of using the declarative nn.Sequential container (which encodes a very particular kind of control flow).

examples/source_separation/conv_tasnet/model.py
```diff
@@ -0,0 +1,325 @@
...
+        hidden_channels: int,
+        kernel_size: int,
+        padding: int,
+        dilation: int = 1,
+        causal: bool = False,
+        no_residual: bool = False,
+    ):
+        super().__init__()
+
+        if causal:
```
cpuhrsch, line 38: Do you want to remove this flag and add to documentation it's not supported?

examples/source_separation/conv_tasnet/model.py
cpuhrsch, line 249: I'd change the name of this to avoid confusion (e.g. someone might confuse it with a function such as https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_sequence.html )

vincentqb:
Can you also add the tests you mentioned?

examples/source_separation/conv_tasnet/model.py
mthrok: Can you help come up with the name?
Padding is what is curried out in this method, yet it's still difficult to summarize the why the padding is curried out in a few words. Also this is an internal method, so when someone reads this I assume they have some context of the model and they can assume this is related to convolution operation and encoder/decoder operation.

mthrok: Why static? This method uses the model parameters. It should be tied to the model instance and has a very specific purpose. What is your justification of decoupling it?

examples/source_separation/conv_tasnet/model.py
mthrok: I see your point, and will consider that in the future.

mthrok: It's just changing this now and I will have to retrain the model before the model release. (well, it might be worth doing it) so I am a bit hesitant to do.

mthrok: a (logically weak) excuse is that since this is a reference implementation, so we do not intend to change it, but we cannot be sure about the future.

mthrok: actually, I realized that there are still some improvements I can apply. I will apply your suggestion and run the training again.

examples/source_separation/conv_tasnet/model.py
mthrok: Changed it to `_align_num_frames_with_strides`

examples/source_separation/conv_tasnet/model.py
mthrok: done

mthrok:
Superseded by #920

PR #894, opened 2020-08-28
Add si_snr metrics

mthrok:
#897


examples/source_separation/conv_tasnet/metrics.py
```diff
@@ -0,0 +1,116 @@
...
+        """
+        assert estimate.shape == target.shape
+        assert reduction in ["mean", "sum"]
+
+        num_speakers = target.shape[1]
+
+        min_metric = None
+        min_metric_value = float("inf")
+        for idx in permutations(range(num_speakers)):
+            metric = self.loss_func(estimate, target[:, idx, :])
```
mthrok, line 106: This applies the shuffling to all the samples in batch. Each samples should be shuffled separately.

examples/source_separation/utils/metrics.py
```diff
@@ -0,0 +1,164 @@
...
+        mix (torch.Tensor): Mixed souce signals, from which the setimated signals were generated.
+            Shape: [batch, speakers == 1, time frame]
+
+    Returns:
+        torch.Tensor: Improved SDR. Shape: [batch, ]
+
+    References:
+        - Conv-TasNet: Surpassing Ideal Time--Frequency Magnitude Masking for Speech Separation
+          Luo, Yi and Mesgarani, Nima
+          https://arxiv.org/abs/1809.07454
```
faroit, line 160: I think the first reference should be https://arxiv.org/abs/1811.02508

faroit (changes requested):

examples/source_separation/utils/metrics.py
mthrok: Thanks for the suggestion.

mthrok: @faroit 

I was looking at it, and just realized that SDR is very common metric that may be not novel here.
What do you think I should site? In this PR, I have the following functions;

1. `SDR` which implementation was taken from [here](https://github.com/naplab/Conv-TasNet/blob/e66d82a8f956a69749ec8a4ae382217faa097c5c/utility/sdr.py#L34-L56), which is a part of reference implementation of conv-tasnet provided by the paper authors.
2. `PIT`: reference is the paper https://arxiv.org/abs/1703.06284, which I think is correct.
3. SDR-PIT: I sited the conv-tasnet paper, but is there a better source?
4. `sdri`, which contains the improvement `sdr`, which as you suggested is refered in https://arxiv.org/abs/1811.02508.

examples/source_separation/utils/metrics.py
```diff
@@ -0,0 +1,164 @@
...
+    mix_pow = (estimate * reference).mean(axis=2, keepdim=True)
+    scale = mix_pow / (reference_pow + epsilon)
+
+    reference = scale * reference
+    error = estimate - reference
+
+    reference_pow = reference.pow(2).mean(axis=2)
+    error_pow = error.pow(2).mean(axis=2)
+
+    return 10 * torch.log10(reference_pow) - 10 * torch.log10(error_pow)
```
mthrok, line 45: @faroit Could you help me understand, is this SDR or SI-SDR?

examples/source_separation/test/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+_METRICS_REFERENCE = os.path.join(_THIS_DIR, "metrics", "reference.py")
+
+
+sys.path.append(os.path.join(_THIS_DIR, "..", "..", "..", "test"))
+
+
+def _download_metrics_reference():
+    import requests
+
+    url = "https://raw.githubusercontent.com/naplab/Conv-TasNet/e66d82a8f956a69749ec8a4ae382217faa097c5c/utility/sdr.py"
```
cpuhrsch, line 14: I'd move this into the third_party folder as a subfolder named e.g. Conv-TasNet or whatever you prefer. Just make sure you provide proper reference

mthrok: Moved the test code to the unit test and pulled in the reference implementation as `test/torchaudio_unittest/example/souce_sepration/sdr_reference.py.`

test/torchaudio_unittest/example/__init__.py
```diff
@@ -0,0 +1,8 @@
+import os
+import sys
+
+
+sys.path.append(
```
cpuhrsch, line 5: Are there alternatives to this?

cpuhrsch (approved):
I think this is for now, assuming there will be a follow-up pull request that resolves how we might be able to avoid the runtime modification of sys.path. One option would be to deploy a minimal set of testing tools and move these tests into the example folder. That also delineates which modules are in the package and which are still within in the examples. The tests within the example folder could then we run on a weekly basis.

test/torchaudio_unittest/example/__init__.py
mthrok: From the discussion, so as to truly get rid of this, we need to make the test utilities installable package, (it can be a part of torchaudio library). Putting the tests in example directory still requires the similar path modification so that it can access the test utilities defined in `test/torchaudio_unittest`.

The other pros and cons mentioned was that
 - putting example test in `test/torchaudio_unittest`
    - easy to run it in CI
 - putting example tests in `example`
    - better locality

mthrok:
thanks.

examples/source_separation/utils/metrics.py
faroit: I think things are a bit messed up with respect to SDR namings (also pointed out in https://arxiv.org/abs/1811.02508)

1. The SDR as implemented [here](https://github.com/naplab/Conv-TasNet/blob/e66d82a8f956a69749ec8a4ae382217faa097c5c/utility/sdr.py#L34-L56) is scale-invariant, so this would should be called `SI-SDR` citing https://arxiv.org/abs/1811.02508
2. agreed
3. I am not sure if there was another paper before, so probably correct. @mpariente ?
4. yes

I would recommend to separate the actual loss function (SDR, SISDR, SDSDR) from the property of being permutation invariant or not (PIT vs non-pit). E.g. like in asteroids [pit wrapper](https://github.com/mpariente/asteroid/blob/master/asteroid/losses/pit_wrapper.py)

examples/source_separation/utils/metrics.py
faroit: yes this is SI-SDR. Normal SDR is basically log-scaled MSE

faroit:
@mthrok are a bit late. Maybe you can open another PR to update if necessary. Feel free to ping me again

examples/source_separation/utils/metrics.py
mthrok: @faroit Thanks for the reply. Fortunately these metrics went into `examples` directly. So we do can fix this without being worried about backward compatibility. I will open another PR to fix this and ping you.
We are working on the next release (expected to happen at the end of October), so I will do it after that.

PR #896, opened 2020-08-28
Add conv-tasnet training script

mthrok:


examples/source_separation/train.py
```diff
@@ -0,0 +1,169 @@
...
+        _LG.info("Launching worker %s: `%s`", i, " ".join(command))
+        worker = subprocess.Popen(command)
+        workers.append(worker)
+
+    num_failed = 0
+    for worker in workers:
+        worker.wait()
+        if worker.returncode != 0:
+            num_failed += 1
+    sys.exit(num_failed)
```
fmassa, line 148: I believe this could be accomplished with `torch.multiprocessing.spawn`? Here is [one example on how to do it](https://github.com/pytorch/examples/blob/4b119d735b802453479d739bf823f3f7d8d5d422/imagenet/main.py#L103-L112)

examples/source_separation/train.py
```diff
@@ -0,0 +1,169 @@
...
+        dist_utils.setup_distributed(
+            world_size=args.num_workers,
+            rank=args.worker_id,
+            local_rank=args.device_id,
+            init_method=args.sync_protocol,
+        )
+        if torch.cuda.is_available():
+            torch.cuda.set_device(args.device_id)
+            _LG.info("CUDA device set to %s", args.device_id)
+        _run_training(args.rest)
```
fmassa, line 124: I had a better experience (much faster execution for some reason) when using `torch.distributed.launch` instead of spawning the jobs with `torch.multiprocessing.spawn`. This might be up to users' preference, but I generally let slurm spawn the processes for me, I only need to handle single process in my scripts.

examples/source_separation/utils/dist_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    if dist.get_world_size() < 2:
+        return
+    rank = dist.get_rank()
+    if rank == 0:
+        _LG.info("[Parameter Sync]: Saving parameters to a temp file...")
+        torch.save({f"{i}": m.state_dict() for i, m in enumerate(modules)}, path)
+    dist.barrier()
+    if rank != 0:
+        _LG.info("[Parameter Sync]:    Loading parameters...")
+        data = torch.load(path)
```
fmassa, line 67: You might want to load this on cpu via `map_location`, otherwise the rank 0 will have the same copy of the parameters 8 times

examples/source_separation/utils/dist_utils.py
```diff
@@ -0,0 +1,74 @@
...
+    os.environ["LOCAL_RANK"] = str(local_rank)
+
+
+def save_on_master(obj, path):
+    if dist.get_rank() == 0:
+        _LG.info("Saving %s", path)
+        torch.save(obj, path)
+
+
+def synchronize_params(path, *modules):
```
fmassa, line 57: If you don't run this function very often it's fine to do it as is, but it is possible to avoid serializing the model to a file and instead use distributed communications to perform the broadcasting for you. Probably something involving using pickle to serialize the dict.

Here is an example of what we did for a slightly different function (all_gather) https://github.com/pytorch/vision/blob/497744b9d510ff2df756f479ee5a19fce0d579b6/references/detection/utils.py#L75-L115

examples/source_separation/utils/dist_utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def _set_env_vars(world_size, rank, local_rank):
+    for key, default in [("MASTER_ADDR", "127.0.0.1"), ("MASTER_PORT", "29500")]:
+        if key not in os.environ:
+            os.environ[key] = default
+
+    os.environ["WORLD_SIZE"] = str(world_size)
+    os.environ["RANK"] = str(rank)
+    os.environ["LOCAL_RANK"] = str(local_rank)
```
fmassa, line 48: If you were to let slurm spawn the processes for you (which can be handy in multi-node training) here is the recipe I  have that works fine on both slurm and with `torch.distributed.launch` https://github.com/facebookresearch/detr/blob/5e66b4cd15b2b182da347103dd16578d28b49d69/util/misc.py#L406-L428

examples/source_separation/conv_tasnet/trainer.py
```diff
@@ -0,0 +1,77 @@
...
+        self.criterion = criterion
+        self.grad_clip = grad_clip
+        self.device = device
+
+    def train_one_epoch(self):
+        num_batches = len(self.train_loader)
+        last_log_time = 0
+        last_log_progress = 0
+        t0 = time.monotonic()
+        for i, (input, reference) in enumerate(self.train_loader, start=1):
```
fmassa, line 28: nit: I found it useful to have a generator a la `tqdm` to handle logging / timing for me, see https://github.com/facebookresearch/detr/blob/5e66b4cd15b2b182da347103dd16578d28b49d69/util/misc.py#L193-L245
although it might just be an overkill and might not be needed here

examples/source_separation/conv_tasnet/train.py
```diff
@@ -0,0 +1,175 @@
...
+        trainer.train_one_epoch()
+        loss = trainer.evaluate()
+        lr_scheduler.step(loss)
+
+        save_path = args.save_dir / f"epoch_{epoch}.pt"
+        dist_utils.save_on_master(
+            {
+                "model": model.module.state_dict(),
+                "optimizer": optimizer.state_dict(),
+                "num_speakers": args.num_speakers,
```
fmassa, line 170: I've found it useful to also serialize the `args` together with the model. It allows for simpler `resume`ing of the model, as we can load all the default args etc

fmassa (commented):
I had a quick looks and overall the structure looks pretty good!

I've made some comments about a few things I've done differently in my past scripts, but they are by no means meant to be the "ground-truth" on how to do things, and I'd be happy to hear your thoughts as well.

examples/source_separation/train.py
mthrok: Yes, that is right. Adopting `torch.multiprocessing.spawn` makes it harder to understand the code so I stripped it.

examples/source_separation/train.py
mthrok: When running from SLURM, this script does not spawn any subprocess. The main process serves as the training worker, so  neither of `torch.distributed.launch` or `torch.multiprocessing.spawn` is used here.

examples/source_separation/utils/dist_utils.py
mthrok: Can you give me more details?

Rank 0 worker saves the parameter to a temporary file then wait for all the other workers to load it before it deletes the temporary file.
So, my understanding is that rank 0 should not have the copy of parameters. (Other workers might)

mthrok: oh reading https://pytorch.org/tutorials/intermediate/ddp_tutorial.html#save-and-load-checkpoints again, I understand what you mean. I will add `map_location`.

examples/source_separation/utils/dist_utils.py
mthrok: This function is dedicated for the parameter initialization on training from scratch and based on the example. 

https://pytorch.org/tutorials/intermediate/ddp_tutorial.html#save-and-load-checkpoints

I think the `gather_all` utility is a bit too much for this purpose.

examples/source_separation/utils/dist_utils.py
mthrok: Yeah, I studied the same script from torchvision/reference/segmentation while I was studying the distributed training, and I came to the conclusion that not including SLURM-specific code makes the implementation simpler yet usable for SLARM. Check out the README bellow.

https://github.com/mthrok/audio/tree/conv-tasnet/examples/source_separation#slurm

examples/source_separation/conv_tasnet/trainer.py
mthrok: I have an exact opposite idea. `tqdm` messes up the log when running in background, and adding more code do disable it goes against of writing a simple example. Also the interactive progress bar is not useful in the first place to me as I never launch training process in foreground except debugging.

examples/source_separation/conv_tasnet/train.py
mthrok: I think saving `args` as a record can be a good idea, but making the training script governed by the dumped value makes the training script less flexible. The dumped file could outlive the original script. What if I want to run training with different parameters? What if I want to change the signatures? One good thing about this script not in library but in example directory is that this script does not have to consider backward compatibility (as long as the training script works on all the commits). So making resume behavior depend on `args` in file goes against it.

examples/source_separation/train.py
mthrok: Also, this entry point is intended only for single-node training mode. (for users who wants to JUST try)

For the real distributed training, please refer to the [README](https://github.com/mthrok/audio/tree/conv-tasnet/examples/source_separation).

examples/source_separation/utils/dist_utils.py
mthrok: Added `map_location`

examples/source_separation/.gitignore
```diff
@@ -0,0 +1 @@
+data
```
cpuhrsch, line 1: Isn't there a toplevel .gitignore? Why have a separate one?

examples/source_separation/README.md
```diff
@@ -0,0 +1,168 @@
...
+│┌──────────┐ ┌─────────┐││┌───────────┐ ┌─────────┐│     │┌─────────────┐ ┌─────────┐│
+││ Process  ├─┤ GPU:P-1 ││││ Process   ├─┤ GPU:P-1 ││     ││ Process     ├─┤ GPU:P-1 ││
+││ Rank:P-1 │ └─────────┘│││ Rank:2P-1 │ └─────────┘│     ││ Rank:NP-1   │ └─────────┘│
+│└──────────┘            ││└───────────┘            │     │└─────────────┘            │
+└────────────────────────┘└─────────────────────────┘     └───────────────────────────┘
+```
+
+</details>
+
+### SLURM
```
cpuhrsch, line 97: submitit is great for user experience: https://github.com/facebookresearch/detr/blob/master/run_with_submitit.py

cpuhrsch: E.g. can do something like
```
python run_with_submitit.py --nodes 4 --timeout 3000 --coco_path /datasets01/COCO/022719
```

cpuhrsch: See https://github.com/facebookincubator/submitit for submitit repo

examples/source_separation/conv_tasnet/train.py
```diff
@@ -0,0 +1,310 @@
...
+import torch
+import torchaudio
+
+import conv_tasnet
+from utils import dataset_utils, dist_utils
+
+_LG = dist_utils.getLogger(__name__)
+
+
+def _parse_args(args):
```
cpuhrsch, line 16: Split the CLI interface into a separate file. Treat the training loop as a library (engine or whatever). That way you're separating the user interface from the implementaiton.

examples/source_separation/README.md
```diff
@@ -0,0 +1,168 @@
...
+This directory contains reference implementations for source separations. For the detail of each model, please checkout the followings.
+
+- [Conv-TasNet](./conv_tasnet/README.md)
+
+## Usage
+
+### Overview
+
+To traing a model, you can use [`train.py`](./train.py). This script takes the form of
+`trin.py [parameters for distributed training] -- [parameters for model/training]`
```
cpuhrsch, line 12: nit: train

examples/source_separation/conv_tasnet/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Training
+
+The training takes about 5 mins per epoch with 8 V100 GPUs in a single node.
+
+## Evaluation
+
+The following is the evaluation result of training the model on WSJ0-2mix and WSJ0-3mix datasets.
+
+### wsj0-mix 2speakers
```
cpuhrsch, line 42: Do you want to upload a model to make it easier to reproduce these numbers?

examples/source_separation/README.md
mthrok: I looked at it, but it feels like it adds the more learning curve to go through this example.
The code is right now completely agnostic to SLURM yet runnable with SLURM.
I think that leaving the decision of the adaptation of submitit to users will keep our code simpler.
As long as the bare minimum of how to use in SLURM environment is explained in README, then it should do.

examples/source_separation/conv_tasnet/README.md
mthrok: Yeah, I think that will be great.

examples/source_separation/.gitignore
mthrok: data sounds too generic to put under the top-level `.gitignore`, however this is for my dev use and it was not meant to be checked in. I will remove it.

examples/source_separation/README.md
mthrok: fixed

examples/source_separation/README.md
```diff
@@ -0,0 +1,168 @@
...
+
+This directory contains reference implementations for source separations. For the detail of each model, please checkout the followings.
+
+- [Conv-TasNet](./conv_tasnet/README.md)
+
+## Usage
+
+### Overview
+
+To traing a model, you can use [`train.py`](./train.py). This script takes the form of
```
cpuhrsch, line 11: nit: train

cpuhrsch (approved):

mthrok:
Thanks!

Issue #903, opened 2020-09-10
[Announcement] Improving I/O for correct and consistent experience


mthrok:
**tl;dr: how to migrate to new backend/interface in `0.7`**

* If you are using `torchaudio` in Linux/macOS environments, please use `torchaudio.set_audio_backend("sox_io")` to adopt to the upcoming changes.

* If you are in Windows environment, please set `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False` and reload backend to use the new interface.

* Note that this ships with some bug-fixes for formats other than 16bit signed integer WAV, so you might experience some BC-breaking changes as described in the section below.

**News**
[UPDATE] 2021/03/06
 - All the migration works have been completed on master branch.

[UPDATE] 2021/02/12
 - Added `bits_per_sample` and `encoding` argument (replaced `dtype`) to `save` function.

[UPDATE] 2021/01/29
 - Added `encoding` to `AudioMetaData`

[UPDATE] 2021/01/22
 - Added `format` argument to `load`/`info`/`save` function.
 - `bits_per_sample` to `AudioMetaData`

[UPDATE] 2020/10/21
 - Added Description of `"soundfile"` backend legacy interface.

[UPDATE] 2020/09/18
 - Added migration guide for `"soundfile"` backend.
 - Moved the phase when `"soundfile"` backend signatures change from 0.9.0 to 0.8.0 so that they match with `"sox_io"` backend, which becomes default in 0.8.0.

[UPDATE] 2020/09/17
 - Added information on deprecation of native `libsox` structures such as `signalinfo_t` and `encoding_t`.

# Improving I/O for correct and consistent experience

This is an announcement for users that we are making backward-incompatible changes to I/O functions of `torchaudio` backends from 0.7.0 release throughout 0.9.0 release.

## What is affected?

- **Public APIs**
  - `torchaudio.load`
    - [Linux/macOS] By switching the default backend from `"sox"` backend to `"sox_io"` backend in 0.8.0, loading audio formats other than 16bit signed integer WAV returns the correct tensor.
    - [Linux/macOS/Windows] The signature of `"soundfile"` backend will be change in 0.8.0 to match that of `"sox_io"` backend.
  - `torchaudio.save`
    - [Linux/macOS] By switching to `"sox_io"` backend, saving audio files will no longer degrade the data. The supported format will be restricted to the tested formats only. (please refer to [the doc](https://pytorch.org/audio/backend.html#backend) for the supported formats.)
[...]

PR #906, opened 2020-09-10
Add expected BC-breaking change warning to soundfile

mthrok:
See #903 

Superseded by #922

### DIFF
--- a/torchaudio/backend/soundfile_backend.py
+++ b/torchaudio/backend/soundfile_backend.py
@@ -1,4 +1,5 @@
 import os
+import warnings
 from typing import Optional, Tuple
 
 import torch
@@ -34,8 +35,26 @@ def load(filepath: str,
          offset: int = 0,
          signalinfo: SignalInfo = None,
          encodinginfo: EncodingInfo = None,
-         filetype: Optional[str] = None) -> Tuple[Tensor, int]:
+         filetype: Optional[str] = None,
+         *,
+         normalize: bool = True,
+         frame_offset: int = 0) -> Tuple[Tensor, int]:
     r"""See torchaudio.load"""
+    warnings.warn(
+        'The return type of `load` function in "soundfile" backend is going to be changed in 0.8.0. '
+        'Please refer to https://github.com/pytorch/audio/issues/903 for the detail.'
+    )
+
+    if not normalization:
+        warnings.warn(
+            '"normalization" argument is deprecated and removed in 0.8.0. Please use "normalize".')
+    elif not normalize:
+        normalization = normalize
+    if offset != 0:
+        warnings.warn(
+            '"offset" argument is deprecated and removed in 0.8.0. Please use "frame_offset".')
+    elif frame_offset != 0:
+        offset = frame_offset
 
     assert out is None
     assert normalization
@@ -82,8 +101,28 @@ def load_wav(filepath, **kwargs):
 
 @_mod_utils.requires_module('soundfile')
 @common._impl_save
-def save(filepath: str, src: Tensor, sample_rate: int, precision: int = 16, channels_first: bool = True) -> None:
+def save(
+        filepath: str,
+        src: Tensor,
+        sample_rate: int,
+        precision: int = 16,
+        channels_first: bool = True,
+        *,
+        compression: Optional[float] = None,
+) -> None:
     r"""See torchaudio.save"""
+
+    if precision:
+        warnings.warn(
+            '"precision" argument is depricated and removed in 0.8.0. In 0.8.0, '
+            'convert the input Tensor to dtype of desired precision.'
+        )
+
+    if compression is not None:
+        warnings.warn(
+            '"soundfile" backend\'s `save` function  does not support changing compression level. '
+            '`compression` argument is silently ignore.'
+        )
 
     ch_idx, len_idx = (0, 1) if channels_first else (1, 0)
 
--- a/torchaudio/backend/utils.py
+++ b/torchaudio/backend/utils.py
@@ -59,8 +59,8 @@ def set_audio_backend(backend: Optional[str]):
         module = sox_io_backend
     elif backend == 'soundfile':
         warnings.warn(
-            '"soundfile" backend is planned to change the function signatures in 0.9.0. '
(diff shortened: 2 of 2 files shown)


torchaudio/backend/utils.py
```diff
@@ -53,6 +53,9 @@ def set_audio_backend(backend: Optional[str]):
     elif backend == 'sox_io':
         module = sox_io_backend
     elif backend == 'soundfile':
+        warnings.warn(
+            '"soundfile" backend is planned to change the function signatures in 0.9.0. '
+            'Please refer to https://github.com/pytorch/audio/issues/903')
```
vincentqb, line 58: When I see a warning like this, I think "Ok, how can I change my code to migrate and suppress the warning?". This warning may give the impression soundfile should not be used, which is not the case.

If we want to warn, then I'd say we should have a mechanism for the user to use the new interface right away.

vincentqb: (offline comment on [internal document](https://docs.google.com/document/d/1ekaLxE9BzSlZ3JcTZD90icfvtO3N6JKs5XmBuZigcDI/edit#))

vincentqb (approved):
As mentioned offline, I'd suggest instead:
- For `out=`, `signalinfo=`, `encodinginfo=`, `filetype`, in `load` and `precision` in `save`, I would simply raise a warning if different from default and then remove.
- For `normalization=` and `offset=`, I would simply create the new parameter, and tell user to switch, since it will be removed.

That being said, I'm also ok with the current warning, so I'll approve the PR so you can decide what to do and move forward :)

torchaudio/backend/soundfile_backend.py
```diff
@@ -82,9 +101,29 @@ def load_wav(filepath, **kwargs):
...
+    if precision:
+        warnings.warn(
+            '"precision" argument is depricated and removed in 0.8.0. In 0.8.0, '
+            'convert the input Tensor to dtype of desired precision.'
+        )
+
+    if compression is not None:
+        warnings.warn(
+            '"soundfile" backend\'s `save` function  does not support changing compression level. '
+            '`compression` argument is silently ignore.'
```
vincentqb, line 124: nit: silently ignored

vincentqb: will this parameter be removed?

torchaudio/backend/soundfile_backend.py
```diff
@@ -34,8 +35,26 @@ def load(filepath: str,
...
+    if not normalization:
+        warnings.warn(
+            '"normalization" argument is deprecated and removed in 0.8.0. Please use "normalize".')
+    elif not normalize:
+        normalization = normalize
+    if offset != 0:
+        warnings.warn(
+            '"offset" argument is deprecated and removed in 0.8.0. Please use "frame_offset".')
+    elif frame_offset != 0:
+        offset = frame_offset
```
vincentqb, line 57: can always be done?

vincentqb (approved):
LGTM

torchaudio/backend/soundfile_backend.py
```diff
@@ -82,9 +101,29 @@ def load_wav(filepath, **kwargs):
...
 
+    if precision:
+        warnings.warn(
+            '"precision" argument is depricated and removed in 0.8.0. In 0.8.0, '
+            'convert the input Tensor to dtype of desired precision.'
+        )
+
+    if compression is not None:
+        warnings.warn(
+            '"soundfile" backend\'s `save` function  does not support changing compression level. '
```
cpuhrsch, line 123: Do you have an issue that gives more detail for this?

torchaudio/backend/soundfile_backend.py
```diff
@@ -34,8 +35,26 @@ def load(filepath: str,
          offset: int = 0,
          signalinfo: SignalInfo = None,
          encodinginfo: EncodingInfo = None,
-         filetype: Optional[str] = None) -> Tuple[Tensor, int]:
+         filetype: Optional[str] = None,
+         *,
```
vincentqb, line 39: what is the `*`?

torchaudio/backend/utils.py
```diff
@@ -58,6 +58,9 @@ def set_audio_backend(backend: Optional[str]):
     elif backend == 'sox_io':
         module = sox_io_backend
     elif backend == 'soundfile':
+        warnings.warn(
+            '"soundfile" backend is planned to change the function signatures in 0.8.0. '
+            'Please refer to https://github.com/pytorch/audio/issues/903 for the migration step.')
```
vincentqb, line 63: the real issue behind this particular comment is the return type? then it could be specialized to return type?

torchaudio/backend/soundfile_backend.py
```diff
@@ -82,9 +101,29 @@ def load_wav(filepath, **kwargs):
...
+        precision: int = 16,
+        channels_first: bool = True,
+        *,
+        compression: Optional[float] = None,
+) -> None:
     r"""See torchaudio.save"""
 
+    if precision:
+        warnings.warn(
+            '"precision" argument is depricated and removed in 0.8.0. In 0.8.0, '
```
vincentqb, line 117: nit: deprecated

torchaudio/backend/soundfile_backend.py
```diff
@@ -34,8 +35,26 @@ def load(filepath: str,
...
-         filetype: Optional[str] = None) -> Tuple[Tensor, int]:
+         filetype: Optional[str] = None,
+         *,
+         normalize: bool = True,
+         frame_offset: int = 0) -> Tuple[Tensor, int]:
     r"""See torchaudio.load"""
+    warnings.warn(
+        'The return type of `load` function in "soundfile" backend is going to be changed in 0.8.0. '
+        'Please refer to https://github.com/pytorch/audio/issues/903 for the detail.'
+    )
```
vincentqb, line 46: that's fine to me. next best thing to creating a new backend?

torchaudio/backend/utils.py
vincentqb: how about a parameter that controls the return type?

Issue #902, opened 2020-09-09
Backend inconsistency with torch.load


timlod:
## 🐛 Bug

`sox` and `soundfile` backends use the parameter `normalization`, whereas `sox_io` uses `normalize`.
Code defaulting on one and falling back on another therefore will throw errors due to a wrong keyword argument.
Similarly, `sox_io` uses `frame_offset` instead of `offset`.

## To Reproduce

1. `torchaudio.set_audio_backend("sox_io")`
2. `torchaudio.load(..., normalization=False)`

## Expected behavior

Consistent keyword naming across backends.

## Environment

torchaudio 0.6.0 

## Additional context

I saw some typos in the new documentation and may make a pull request for the lot if I get the time soon.
While I actually prefer `normalize`, `normalization` is probably preferred for backwards compatibility.

timlod:
I also found that `sox_io` does not currently support `pathlib` Paths for loading, where other backends do.

timlod:
And I reread the warning note and find that it does advise the changed signature :)
I would still suggest accepting `pathlib` Paths for the `filepath` argument.

mthrok:
Hi @timlod 

Thanks for the report. I am also feeing the need to make "sox_io" backend compatible with `"pathlib.Path"` type. If you would like to make a PR, let me know, else I will work on it some time.

timlod:
I'll have a look on Sunday, and will probably submit a PR then (I don't think it should be a lot of work?)!

mthrok:
@timlod 

> I'll have a look on Sunday, and will probably submit a PR then (I don't think it should be a lot of work?)!

Thanks.
Yes, I think it's just changing the type annotation and adding `path=str(path)` to convert it to string internally before passing it to C++ code.

PR #907, opened 2020-09-13
Add pathlib.Path support to sox_io backend

timlod:
Allows for passing of `pathlib.Path` `filepath` to sox_io functions.
See #902.

For type union docstrings I used (str/pathlib.Path), as I found this pattern was used in the `kaldi_io` methods (for str/FileDescriptor). 
There may be a better pattern though?

### DIFF
--- a/torchaudio/backend/common.py
+++ b/torchaudio/backend/common.py
@@ -125,8 +125,7 @@ Returns:
 _SAVE_DOCSTRING = r"""Saves a Tensor on file as an audio file
 
 Args:
-    filepath: Path to audio file. This function also handles ``pathlib.Path`` objects, but is
-             annotated as ``str`` for TorchScript compiler compatibility.
+    filepath: Path to audio file.
     src: An input 2D tensor of shape `[C x L]` or `[L x C]` where L is
         the number of audio frames, C is the number of channels
     sample_rate: An integer which is the sample rate of the
--- a/torchaudio/backend/common.py
+++ b/torchaudio/backend/common.py
@@ -60,7 +60,8 @@ class EncodingInfo:
 _LOAD_DOCSTRING = r"""Loads an audio file from disk into a tensor
 
 Args:
-    filepath: Path to audio file
+    filepath: Path to audio file. This function also handles `pathlib.Path` objects, but is annotated
+             as ``str`` for TorchScript compiler compatibility.
 
     out: An optional output tensor to use instead of creating one. (Default: ``None``)
 
@@ -112,7 +113,8 @@ It assumes that the wav file uses 16 bit per sample that needs normalization by
 shifting the input right by 16 bits.
 
 Args:
-    filepath: Path to audio file
+    filepath: Path to audio file. This function also handles `pathlib.Path` objects, but is annotated
+             as ``str`` for TorchScript compiler compatibility.
 
 Returns:
     (Tensor, int): An output tensor of size `[C x L]` or `[L x C]` where L is the number
@@ -123,7 +125,8 @@ Returns:
 _SAVE_DOCSTRING = r"""Saves a Tensor on file as an audio file
 
 Args:
-    filepath: Path to audio file
+    filepath: Path to audio file. This function also handles ``pathlib.Path`` objects, but is
+             annotated as ``str`` for TorchScript compiler compatibility.
     src: An input 2D tensor of shape `[C x L]` or `[L x C]` where L is
         the number of audio frames, C is the number of channels
     sample_rate: An integer which is the sample rate of the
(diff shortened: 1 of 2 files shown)


vincentqb (changes requested):
torchscript tests are failing, and indicate that the union type is not supported by torchscript, see also [here](https://discuss.pytorch.org/t/will-torch-script-support-typing-union/61720/2?u=vincentqb).

For soundfile, even though functions convert to string and make the code compatible with path, only the string type is in the signature, [here](https://github.com/pytorch/audio/blob/master/torchaudio/backend/soundfile_backend.py#L29). I would be ok having a similar unofficial support for path as is done in soundfile.

torchaudio/backend/sox_io_backend.py
```diff
@@ -20,22 +21,24 @@ def __init__(self, sample_rate: int, num_frames: int, num_channels: int):
 
 
 @_mod_utils.requires_module('torchaudio._torchaudio')
-def info(filepath: str) -> AudioMetaData:
+def info(filepath: Union[str, Path]) -> AudioMetaData:
```
mthrok, line 24: Could you keep the signature as `str` for torchscript compatibility? (same for the other functions too.)

```suggestion
def info(filepath: str) -> AudioMetaData:
```

torchaudio/backend/sox_io_backend.py
```diff
@@ -20,22 +21,24 @@ def __init__(self, sample_rate: int, num_frames: int, num_channels: int):
...
     """Get signal information of an audio file.
 
     Args:
-        filepath (str): Path to audio file
+        filepath (str/pathlib.Path): Path to audio file
```
mthrok, line 28: Could you add note on the `pathlib.Path` compatibility, like the following one? (same for the other functions too.)

```suggestion
        filepath (str or pathlib.Path): Path to audio file. Note that the function is annotated only as ``str`` for the reason TorchScript compiler compatibility, this function can also handle `pathlib.Path`.
```

mthrok (commented):
Hi @timlod 

Thanks for the PR. The signature part is a tricky one, which had slipped from my mind when I was replying in #902.

The "sox_io" backend is designed to be compatible with TorchScript compiler (so that the function works in C++ application too), so types that are specific to Python (in this case, they are `Union`, `Path`) cannot be added. Also, because of the way the loading function (in C++) is bound to Python code, we cannot pass file-like object, even though that's common practice in Python.

I checked the intermediate representation that TorchScript compiler generates and `filepath = str(filepath)` has no effect in TorchScript so this change is good. So just keeping the signature as is should do.

torchaudio/backend/sox_io_backend.py
vincentqb: How about the following?
```
This function also handles `pathlib.Path` outside of TorchScript.
```

Since we'd be making an official statement here, should we also update soundfile too? And should we add some tests to guarantee this?

timlod: That all makes sense - good to know, thanks!
I'll update this tonight with the suggestions you added.

timlod: @vincentqb `soundfile_backend` (and the old `sox_backend`) use the docstring from `common.py`, so I added the note regarding pathlib there too now.

torchaudio/backend/common.py
```diff
@@ -60,7 +60,8 @@ def __init__(self,
 _LOAD_DOCSTRING = r"""Loads an audio file from disk into a tensor
 
 Args:
-    filepath: Path to audio file
+    filepath: Path to audio file. This function also handles `pathlib.Path` objects, but is annotated
+             as ``str`` for TorchScript compiler compatibility.
```
mthrok, line 64: Actually, the other backend is not TorchScript compatible. So can you evert this?
I think the change to other backends should be separated from this PR. (note: this is not asking you to work on it.)

torchaudio/backend/sox_io_backend.py
```diff
@@ -1,4 +1,5 @@
-from typing import Tuple, Optional
+from typing import Tuple, Optional, Union
+from pathlib import Path
```
mthrok, line 2: Do we still need these changes?

mthrok (commented):
Hi @timlod 

Thanks for updating the PR. Changes related to the other backends can be omitted from this PR.

torchaudio/backend/sox_io_backend.py
timlod: good catch! current update reflects these comments.

torchaudio/backend/common.py
mthrok: @timlod  Is this accidentally resolved?

If you want to change the cdocsrting of other backends, then can you make it appear not for TorchScript compiler compatibility but for the consistency between "sox_io" backend which cannot be annotated as `pathlib.Path` due to TorchScript compiler compatibility?

timlod: @mthrok the most recent commit removes that change - am I missing something? I was not planning on changing the other backends (I could do that if you'd like me to, but as you mention above that should go in another PR)

mthrok: Oh, I see you reverted change on `_SAVE_DOCSTRING` but the same has to happen to the other docstrings too.
I think `git checkout b6a61c3f -- torchaudio/backend/common.py` will fix this.

timlod: Oops - not quite sure what went wrong there, but I fixed it now. Thanks for your patience!

mthrok (approved):
Looks good, thanks!

PR #909, opened 2020-09-14
packaging: Use cpu directory by default

seemethere:
Default to using the `cpu` directory since the base one is not reliable

similar to https://github.com/pytorch/text/pull/981


seemethere:
I don't believe the failures in the tests are related to the contents of this PR

mthrok (approved):
Looks good!

Issue #908, opened 2020-09-14
more efficient resample module


small-yellow-duck:
The resampling function in Kaldi that pytorchaudio is currently using has some inefficient for loops and padding steps. I've put together efficient module and evaluated the performance in this notebook:
https://www.kaggle.com/smallyellowduck/fast-audio-resampling-layer-in-pytorch (code is in the the notebook)

edit to make two separate comparisons of the resampling time without the file load time:
Comparison 1: 'kaiser_best' settings in librosa vs 'kaiser_best' setting in the efficient pytorch resampler (should be the same setup)

    librosa: 51 s
    efficient pytorch resampler: 9 s

Comparison 2: default setting in torchaudio vs window='hann', num_zeros=6 in the efficient pytorch resampler (should be the same set-up)

    torchaudio: 10 s
    efficient pytorch resampler: 1 s

The performance improvement is most substantial when the input sample rate and output sample rate are not whole number multiple of each other.

I think it would be good for torchaudio to switch to the more efficient resample module. 

Before making a PR, perhaps other people have feedback about what the API for the module should look like? I have largely tried to follow the api for the resample method in librosa. Any other additional comments? @vincentqb 

```
    def __init__(self,
                 input_sr, output_sr, dtype,
                 num_zeros = 64, cutoff_ratio = 0.95, filter='kaiser', beta=14.0):
        super().__init__()  # init the base class
        """
        This creates an object that can apply a symmetric FIR filter
        based on torch.nn.functional.conv1d.

        Args:
          input_sr:  The input sampling rate, AS AN INTEGER..
          output_sr:  The output sampling rate, AS AN INTEGER.
          dtype:  The torch dtype to use for computations
          num_zeros: The number of zeros per side in the (sinc*hanning-window)
              filter function.  More is more accurate, but 64 is already quite a lot. 
          cutoff_ratio: The filter rolloff point as a fraction of the Nyquist freq.
          filter: one of ['kaiser', 'kaiser_best', 'kaiser_fast', 'hann']
          beta: parameter for 'kaiser' filter
```

mthrok:
Hi @small-yellow-duck 

Thanks for the suggestion and the code. This is very cool, and I am very happy to replacing the existing resampling code which we are having difficult time testing.

I looked at the test code of the implementation and it seems it's parity with `librosa`. Do the additional filter types also give the same results as `librosa`? (This is a question out of curiosity, not the requirement for adding it to torchaudio.)

[...]

Issue #452, opened 2020-03-03
`hann_window` does not obey COLA; therefore, it cannot be used with `istft`


PetrochukM:
## 🐛 Bug

> In order to enable inversion of an STFT via the inverse STFT in istft, it is sufficient that the signal windowing obeys the constraint of “Constant OverLap Add” (COLA). This ensures that every point in the input data is equally weighted, thereby avoiding aliasing and allowing full reconstruction.
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.check_COLA.html

Learn more: https://ccrma.stanford.edu/~jos/sasp/Mathematical_Definition_STFT.html#19930

## To Reproduce

```python
>>> from scipy import signal
>>> import torch
>>> signal.check_COLA(signal.hann(1024, sym=False), 1024, 1024 - 256)
True
>>> signal.check_COLA(torch.hann_window(1024).numpy(), 1024, 1024 - 256)
False
```

## Expected behavior

The expected behavior is that the hann window is implemented such that it is useable with `istft`.

PetrochukM:
While it looks like the windows are mostly similar after visualization, it does look like the PyTorch window is off `scipy` on the third decimal... 

```python
from scipy import signal
import torch

window = 1024
torch_hann_window = torch.hann_window(window)
scipy_signal_window = signal.hann(1024, sym=False)
print('scipy_signal_window', scipy_signal_window[0], scipy_signal_window[-1])
print('torch_hann_window', torch_hann_window[0], torch_hann_window[-1])
```

```
scipy_signal_window 0.0 9.412358699445456e-06
torch_hann_window tensor(0.) tensor(9.4175e-06)
```

The `scipy` window is the one Librosa uses under the hood: https://librosa.github.io/librosa/_modules/librosa/filters.html#get_window

vincentqb:
What do you get if you use that window in `istft`? :)

PetrochukM:
Perfect reconstruction is not possible and aliasing occurs. 

Sources:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.check_COLA.html
https://www.dsprelated.com/freebooks/sasp/Constant_Overlap_Add_COLA_Cases.html
https://www.dsprelated.com/freebooks/sasp/Dual_Constant_Overlap_Add.html

vincentqb:
I meant: does torchaudio silently computes the wrong quantity, gives an error, returns NaNs, etc?

PetrochukM:
[...]

PR #911, opened 2020-09-15
Faster module for resampling time series.

small-yellow-duck:
The proposed Resampler routine is more efficient than the existing Resample
module for resampling time series signals. Speed improvements are obtained by splitting
the signal into blocks where there are 'input_sr' input samples and 'output_sr' output
samples. Each block is treated with a convolution mapping 'input_sr' input channels to
'output_sr' output channels per time step. The existing Resample module uses a for loop
to iterate over the first index where each filter is applied, but this implementation is
fully convolutional.

The module is based on

https://github.com/danpovey/filtering/blob/master/lilfilter/resampler.py

with improvements to include additional filter types and input parameters that align
with the librosa api.

https://github.com/pytorch/audio/issues/908

- [x] inputs where the sequence length is not a whole number multiple of `input_sr` should be padded with zeros
- [x] adapt module to deal with multi-channel input
- [ ] adapt to deal with non-integer sample rates?

Should tests try to reproduce signals resampled with [scipy.signal.resample](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.resample.html) ?

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -649,17 +649,14 @@ class Resampler(torch.nn.Module):
               quite a lot. The kernel size is 2*num_zeros + 1.
           cutoff_ratio: The filter rolloff point as a fraction of the
              Nyquist frequency.
-          filter: one of [ 'hann'] 
-               ('kaiser', 'kaiser_best', 'kaiser_fast' can be options when 
-               pytorch includes scipy.special.i0)
+          filter: one of [ 'hann', 'kaiser', 'kaiser_best', 'kaiser_fast'] 
           beta: parameter for 'kaiser' filter
 
-        You can think of this algorithm as dividing up the signals
-        (input,output) into blocks where there are `input_sr` input
-        samples and `output_sr` output samples.  Then we treat it
-        using convolutional code, imagining there are `input_sr`
-        input channels and `output_sr` output channels per time step.
-
+        You can think of this algorithm as dividing up the signal
+        into blocks where there are `input_sr` input
+        samples and `output_sr` output samples.  Then we treat each
+        block of `input_sr` elements with a convolution which maps to `output_sr`
+        elements. 
         """
         assert isinstance(input_sr, int) and isinstance(output_sr, int)
         if input_sr == output_sr:
@@ -677,12 +674,12 @@ class Resampler(torch.nn.Module):
         d = gcd(input_sr, output_sr)
         input_sr, output_sr = input_sr // d, output_sr // d
 
+        #TODO: does  dtype need to be explicitly provided or can we just use float32?
         assert dtype in [torch.float32, torch.float64]
         assert num_zeros > 3  # a reasonable bare minimum
 
-        assert filter in ['hann']
-        # ['kaiser', 'kaiser_best', 'kaiser_fast'] can be supported when pytorch
-        # includes scipy.special.io
+        assert filter in ['hann', 'kaiser', 'kaiser_best', 'kaiser_fast']
+
         # settings in https://github.com/bmcfee/resampy/blob/master/resampy/filters.py#L41
         if filter == 'kaiser_best':
             num_zeros = 64
@@ -694,6 +691,9 @@ class Resampler(torch.nn.Module):
             beta = 8.555504641634386
             cutoff_ratio = 0.85
             filter = 'kaiser'
+
+        if not torch.is_tensor(beta):
+            beta = torch.tensor(beta)
 
         # Define one 'block' of samples `input_sr` input samples
         # and `output_sr` output samples.  We can divide up
@@ -746,18 +746,24 @@ class Resampler(torch.nn.Module):
             The heaviside function returns (a > 0 ? 1 : 0).
             """
             #return np.heaviside(1 - np.abs(a), 0.0) * (0.5 + 0.5 * np.cos(a * np.pi))
-            pi = torch.acos(torch.zeros(1)).item() * 2
-            return torch.heaviside(1 - torch.abs(a), 0.0) * (0.5 + 0.5 * torch.cos(a * pi))
+            pi = 2.0*torch.acos(torch.zeros(1)).item()
+            return torch.heaviside(1 - torch.abs(a), torch.tensor(0.0)) * (0.5 + 0.5 * torch.cos(a * pi))
 
+            #t = 0.5 - 0.5*(1 - torch.abs(a) <= 0.0).float() + 0.5*(1 - torch.abs(a) > 0.0).float()
+            #return t * (0.5 + 0.5 * torch.cos(a * pi))
 
         def kaiser_window(a, beta):
-            #TODO: the kaiser window can be added once scipy.special.i0 has been implement in torch
+            '''
+            kaiser_window returns the Kaiser window on [-1,1] and zero otherwise
+            '''
             #w = scipy.special.i0(beta * np.sqrt(np.clip(1 - ((a - 0.0) / 1.0) ** 2.0, 0.0, 1.0))) / special.i0(beta)
             #return np.heaviside(1 - np.abs(a), 0.0) * w
-            return None
+            w = torch.i0(beta * torch.sqrt(torch.clamp(1.0 - torch.pow(a, 2), 0.0, 1.0))) / torch.i0(beta)
+            return torch.heaviside(1 - torch.abs(a), torch.tensor(0.0)) * w
 
 
         def sinc(x):
+            #stop using this definition once torch.sinc is a supported function
             return torch.exp(-torch.lgamma(1 + x) - torch.lgamma(1 - x))
 
             # The weights below are a sinc function times a Hann-window function.
@@ -790,7 +796,6 @@ class Resampler(torch.nn.Module):
         if output_sr == 1:
             self.resample_type = 'integer_downsample'
             self.padding = input_sr * blocks_per_side
-            weights = torch.tensor(weights, dtype=dtype, requires_grad=False)
             self.weights = weights.transpose(1, 2).contiguous().view(1, 1, input_sr * kernel_width)
 
         elif input_sr == 1:
@@ -799,13 +804,12 @@ class Resampler(torch.nn.Module):
             # output_sr had been swapped.
             self.resample_type = 'integer_upsample'
             self.padding = output_sr * blocks_per_side
-            weights = torch.tensor(weights, dtype=dtype, requires_grad=False)
             self.weights = weights.flip(2).transpose(0, 2).contiguous().view(1, 1, output_sr * kernel_width)
         else:
             self.resample_type = 'general'
             self.reshaped = False
             self.padding = blocks_per_side
-            self.weights = torch.tensor(weights, dtype=dtype, requires_grad=False)
+            self.weights = weights #torch.tensor(weights, dtype=dtype, requires_grad=False)
 
         self.weights = torch.nn.Parameter(self.weights, requires_grad=False)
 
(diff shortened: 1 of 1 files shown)


torchaudio/transforms.py
```diff
@@ -9,6 +9,9 @@
 from torchaudio import functional as F
 from torchaudio.compliance import kaldi
 
+import numpy as np
+from scipy import special
```
mthrok, line 13: Hi @small-yellow-duck 

Thanks for opening a PR. I will try this out to think how to test. I have not looked into the detail yet but `numpy` and `scipy` are not a mandatory dependencies of `torchaudio`.

We have a mechanism to differ loading the optional dependencies but it would be nice if we can implement without them. Quick googling tells that `torch` has `heaviside` function, so removing `numpy` appears plausible. We need to think what to do for `scipy` dependency.

small-yellow-duck: `scipy.special.i0` is used to compute the kaiser window for `kaiser`, `kaiser_best` and `kaiser_fast`. If scipy dependency is not desirable, then one option would be to just allow `hann` window mode. (the original goal of including `kaiser` windows was to try to duplicate the behaviour of librosa, but I don't think the results of librosa and this proposed routine are identical to numeric precision, so there is likely some other detail about how the librosa filters were generated that I haven't managed to duplicate). If the routine only needs to support `hann` windows, then I agree that it should be possible to remove numpy and scipy as dependencies.

I also see that there is an issue suggesting that pytorch add the modified bessel functions, although there doesn't seem to be a PR to merge the code into the pytorch codebase: https://github.com/pytorch/pytorch/issues/7815 . If the code to calculate the equivalent of `scipy.special.i0` is added to pytorch, then it should also be possible to include `kaiser` windows.

small-yellow-duck: I realized that sinc can be written in terms of the gamma function - and pytorch has a routine to compute the log of the gamma function (`torch.lgamma`)! So I think it should be possible to excise numpy as a dependency as well.... (see output 10 in https://mathworld.wolfram.com/SincFunction.html)

small-yellow-duck:
Any comments to the following changes? @mthrok 

- Removing numpy and scipy as dependencies means that the resampler will no longer support `kaiser` windows. `kaiser` windows can be supported if there is a pytorch definition for `scipy.special.i0`.

- In order to remove the numpy dependency, the sinc function is now defined as torch.exp(-torch.lgamma(1-x) - torch.lgamma(1+x)). This is not a good solution because the `exp` introduces numerical errors. A solution is to have a pytorch definition based on a taylor series expansion of `sinc`.

- The `general` mode in forward no longer crops out the portion of the signal that is longer than `(seq_len//input_sr) * input_sr`. (YAY)

I am still thinking about how to support non-integer sample rates. 
Further modifications to support multi-channel signals is desirable.

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        using convolutional code, imagining there are `input_sr`
+        input channels and `output_sr` output channels per time step.
+
+        """
+        assert isinstance(input_sr, int) and isinstance(output_sr, int)
+        if input_sr == output_sr:
+            self.resample_type = 'trivial'
+            return
+
+        def gcd(a, b):
```
mthrok, line 669: Kind of surprise to me but `torch` has gcd. https://pytorch.org/docs/master/generated/torch.gcd.html#torch-gcd

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+
+
+        def kaiser_window(a, beta):
+            #TODO: the kaiser window can be added once scipy.special.i0 has been implement in torch
+            #w = scipy.special.i0(beta * np.sqrt(np.clip(1 - ((a - 0.0) / 1.0) ** 2.0, 0.0, 1.0))) / special.i0(beta)
+            #return np.heaviside(1 - np.abs(a), 0.0) * w
+            return None
+
+
+        def sinc(x):
```
mthrok, line 760: Looks like `sinc` function is under development.

mthrok (commented):
Hi @small-yellow-duck 

Sorry for the late reply. I took a look at your PR and it was surprisingly straightforward to follow. I have not got to the point of thinking of testing phase, but I made couple of comments.

Specifically, `torch` seems to already have `kaiser_window` and `hann_window`. Did you look at them? `sinc` is also under development https://github.com/pytorch/pytorch/pull/44713 so it might become available rather soon.

On the interface design aspect, since this module has four distinctive operation modes (`trivial`, `integer_downsample`, `integer_upsample` and `general`), it might be a better to split the resampling class into 4 different and introduce a factory function. cc @cpuhrsch

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        """
+        Resample the data
+
+        Args:
+         input: a torch.Tensor with the same dtype as was passed to the
+           constructor.
+         There must be 2 axes, interpreted as (minibatch_size, sequence_length)...
+         the minibatch_size may in practice be the number of channels.
+
+         TODO: make default input dim (minibatch_size, channels, seq_len)?
```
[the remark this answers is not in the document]
small-yellow-duck: I also like the flexibility of having 2d or 3d inputs.

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+
+            # num_blocks = seq_len // self.input_sr
+            # if num_blocks == 0:
+            #     #if signal is shorter than the filter, then pad signal with zeros.
+            #     zeros = torch.zeros(data.size()[0:-1]+(self.input_sr - seq_len,), dtype=data.dtype)
+            #     data = torch.cat([data, zeros], dim=-1)
+            #     num_blocks = 1
+
+            # data = data[:, 0:(num_blocks*self.input_sr)]  # Truncate input
+            data = data[:, 0:(num_blocks * self.input_sr)].view(minibatch_size, num_blocks, self.input_sr)
```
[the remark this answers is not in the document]
small-yellow-duck: Nope! Let me tidy up this piece of code a bit more...

torchaudio/transforms.py
small-yellow-duck: `gcd` is also in the standard python library. I'm fine to switch to the pytorch `gcd` but is there some underlying reason why it would compelling to do so?

torchaudio/transforms.py
small-yellow-duck: Oh good. I'll leave the definition of the sinc function as it is for now, but the new function should be used when it's available.

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        # sign that the kernel_width term needs to be will depend on whether it's
+        # convolution or correlation, and the logic is tricky.. I will just find
+        # which sign works.
+
+        times = (
+            torch.arange(output_sr, dtype=dtype).reshape((output_sr, 1, 1)) / output_sr -
+            torch.arange(input_sr, dtype=dtype).reshape((1, input_sr, 1)) / input_sr -
+            (torch.arange(kernel_width, dtype=dtype).reshape((1, 1, kernel_width)) - blocks_per_side))
+
+        def hann_window(a):
```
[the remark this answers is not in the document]
small-yellow-duck: `torch.hann_window` doesn't have quite the same functionality as the `hann_window` function in the resample routine: the pytorch `hann_window` computes the filter for a fixed window size (the only input is the window size) without the ability to scale the period of the filter (ie to change the argument of `cos(2*pi*n/N)` to `cos(2*pi*n*a/N)`. Compare this to the behaviour of the `hann_window` function in the resampler, which takes the 3d `times` array and computes the value of the filter for each element.

torchaudio/transforms.py
```diff
@@ -610,6 +610,282 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+            This is applied elementwise to a, which should be a NumPy array.
+
+            The heaviside function returns (a > 0 ? 1 : 0).
+            """
+            #return np.heaviside(1 - np.abs(a), 0.0) * (0.5 + 0.5 * np.cos(a * np.pi))
+            pi = torch.acos(torch.zeros(1)).item() * 2
+            return torch.heaviside(1 - torch.abs(a), 0.0) * (0.5 + 0.5 * torch.cos(a * pi))
+
+
+        def kaiser_window(a, beta):
```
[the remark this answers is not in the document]
small-yellow-duck: same issue as with `hann_window` above... but the good news is that pytorch now has the `i0` function, so I can rewrite the `resampler` `kaiser_window` routine with that. https://pytorch.org/docs/master/generated/torch.i0.html#torch.i0

small-yellow-duck:
Your idea is to have a base class `resample` that does the weight generation and which `resample_trivial`, `resample_integer_upsample`, `resample_integer_downsample` and `resample_general` can inherit from? I am pretty agnostic on whether to split the resampler into `resample_trivial`, `resample_integer_upsample`, `resample_integer_downsample` and `resample_general`: I'm happy to do it if people think it makes the code easier to understand. 

What are your thoughts on what the goal of tests should be? This issue has some discussion about trying to duplicate the behaviour of audio processing tools in scipy with pytorch.
https://github.com/pytorch/audio/issues/452

Another issue is that the resampler `hann_window` function needs to have a value for pi, but pytorch doesn't have a definition for pi. I've used a suggested workaround (`pi = 2.0*torch.acos(torch.zeros(1)).item()`, but there is also an open issue to make pytorch include pi: https://github.com/pytorch/pytorch/issues/19172

small-yellow-duck:
The latest PR now has the ability to accept 2d or 3d inputs.

- Is there a compelling reason to switch from using `gcd` to using `torch.gcd`?
- Is there a better workaround for a definition of `pi`?
- Is it desirable to split the `resample` class into `resample_trivial`, `resample_integer_upsample`, `resample_integer_downsample` and `resample_general` and have each class inherit the weight initializations from `resample`?
- What should the goals of unit tests be? (with the existing definition of sinc and pi, we can't expect numeric agreement with other packages...) Perhaps one option is to downsample a signal and then upsample it again? (and vice versa?)
@mthrok 

I think there is also a minor issue with `general` mode padding the input with 0's so that the `seq_len` is an integer multiple of `input_sr`. When the kernel width is small, this is pretty safe to do, but if the kernel width is large, then the convolution will involve a lot of the padded zeros and the output of the convolution is not going to be scaled correctly (the output will be too small). It should be possible to see this by comparing a signal padded with zeros at the end (`resample(torch.cat(data, zeros), dim=-1)[:, 0:new_seq_len]`) to the same signal padded with zeros at the begining (`resample(torch.cat(zeros, data), dim=-1)[:, -new_seq_len:]`). Maybe there is a smart way to to construct an appropriately normalized kernel to take care of the last segment of the input `data[:, :, (seq_len//self.input_sr)*self.input_sr:]`, or the case where the `seq_len` is shorter than `self.input_sr`. (However, this kernel would have to be constructed during the `.forward` call as the size depends on the data `seq_len`.)

Finally, I think it would be good for somebody to really scrutinize the definition I have for `kaiser_window` to make sure that it is correct. :-)

small-yellow-duck:
Any comments on the latest revision or the questions in the previous post? @mthrok

mthrok:
Hi @small-yellow-duck 

Sorry for the late reply. I do not forget you and this PR, but I got caught up with some works related to the upcoming release. I will try my best to get back to you before the end of the week.

mthrok:
Hi @small-yellow-duck

Sorry for the delayed response. I am still working on release-related task, please give me some more time before I can give feedback.

small-yellow-duck:
No worries, thanks for keeping this in PR mind. :-)

mthrok:
@small-yellow-duck 

Sorry for the late reply. torchaudio 0.7.0 is released and I was reviewing the project ideas. I posted an RFC #1000 regarding the faster I/O, which I have been working, then realized that in this work we can provide a resampling algorithm faster than the current Python implementation. Therefore I would like to hold this PR. I am so sorry for the efforts you have made based on the interaction here.

small-yellow-duck:
I expect most people want to resample data as they load it and I agree that faster resampling with C++ on the cpu is the functionality most worth focusing on.

faroit:
Also maybe its worth, comparing to https://github.com/adefossez/julius that was just released

mthrok:
Closing the issue as it is mostly addressed by #1487. 
Please comment in #1487 if we are missing something.
Thank you for the contribution @small-yellow-duck

PR #914, opened 2020-09-18
Issue warning when a Mel filter bank bin is empty

toddstep:
The default parameters for creating Mel filterbanks can result in some mel features always being zero.
This PR is to warn that this is occurring.

This resembles the following issue:
https://github.com/librosa/librosa/issues/478

The unit tests can be run as follow:
```
pytest -k TestCreateFBMatrix test/torchaudio_unittest/functional/functional_cpu_test.py
```

### DIFF
--- a/torchaudio/functional.py
+++ b/torchaudio/functional.py
@@ -315,9 +315,9 @@ def create_fb_matrix(
 
     if (fb.max(dim=0).values == 0.).any():
         warnings.warn(
-            "At least one mel filterbank has all zero values."
-            "The value for `n_mels` ({}) may be set too high."
-            "Or, the value for `n_freqs` ({}) may be set too low.".format(n_mels, n_freqs)
+            "At least one mel filterbank has all zero values. "
+            f"The value for `n_mels` ({n_mels}) may be set too high. "
+            f"Or, the value for `n_freqs` ({n_freqs}) may be set too low."
         )
 
     return fb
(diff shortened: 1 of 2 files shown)


torchaudio/functional.py
```diff
@@ -313,6 +313,13 @@ def create_fb_matrix(
         enorm = 2.0 / (f_pts[2:n_mels + 2] - f_pts[:n_mels])
         fb *= enorm.unsqueeze(0)
 
+    if (fb.max(dim=0).values == 0.).any():
+        warnings.warn(
+            "At least one mel filterbank has all zero values."
```
mthrok, line 318: I think you need a space after the period, otherwise the constructed string would not have a space between two sentences.

```suggestion
            "At least one mel filterbank has all zero values. "
```

torchaudio/functional.py
```diff
@@ -313,6 +313,13 @@ def create_fb_matrix(
...
+            "The value for `n_mels` ({}) may be set too high."
```
mthrok, line 319:
```suggestion
            f"The value for `n_mels` ({n_mels}) may be set too high. "
```

torchaudio/functional.py
```diff
@@ -313,6 +313,13 @@ def create_fb_matrix(
...
+            "Or, the value for `n_freqs` ({}) may be set too low.".format(n_mels, n_freqs)
```
mthrok, line 320:
```suggestion
            f"Or, the value for `n_freqs` ({n_freqs}) may be set too low."
```

mthrok (commented):
Hi @toddstep 

Thanks for the contribution. It looks good. I made some small suggestions.

toddstep:
Thank you for the review, @mthrok .
I have incorporated your suggestions into the PR.

mthrok (approved):
Looks good. Thanks!

PR #915, opened 2020-09-22
[CI] Disable xdist in macOS unittest in CI

mthrok:
This reverts commit 1ecbc249b5eed3a2e0e765ac0046e2bf9c231884.

Reason: The macOS CI jobs are stuck with xdist recently.

**Note** some macOS CI unittest jobs are still failing at environment setup which is not caused by this PR and is going to be fixed by a separate PR.


vincentqb (approved):
LGTM

Issue #897, opened 2020-08-28
Adding Conv-TasNet


mthrok:
This issue is for tracking the PRs for Conv-TasNet training pipeline.

**PRs for prototype**
- [x] ~#893~ -> #920 Model
- [x] ~#894~ Metrics
- [x] ~#895~ wsj0mix dataset
- [x] ~#896~ training script

**Follow up**
- [ ] Update citation https://github.com/pytorch/audio/pull/894#discussion_r500926586
- [ ] Add pretrained model
    - [ ] Retrain the model
    - [ ] Upload pre-trained model
    - [ ] Add code for pre-trained model
- [ ] Move C++ runtime to master branch: See [my branch](https://github.com/mthrok/audio/tree/libtorchaudio-cpp-example)
- [ ] Move metrics/loss to library.

**Others**
- [ ] Change loss to SI-SNR and re-train.
- [ ] notebook-tize
- [ ] Add google colab

mthrok:
FYI: Initial training result

Best evaluation SI_SNR: 11.22 (Epoch: 54)

Some separation result

<img width="1680" alt="Screen Shot 2020-08-28 at 12 41 43" src="https://user-images.githubusercontent.com/855818/91592028-e4d12a00-e92b-11ea-9bc9-bf9dc6129bdd.png">

<details><summary>log</summary>

```
2020-08-27 08:53:24,322: [worker  0] ----------------------------------------------------------------------
2020-08-27 08:53:24,322: [worker  0] Epoch: 1
2020-08-27 08:53:24,322: [worker  0] Learning rate: 0.001
2020-08-27 08:53:24,322: [worker  0] ----------------------------------------------------------------------
2020-08-27 08:57:20,850: [worker  0] Eval   Loss: -1.437e+00
2020-08-27 08:57:20,850: [worker  0] Eval  Speed: 165.30 [samples/sec]
2020-08-27 08:57:21,148: [worker  0] ----------------------------------------------------------------------
2020-08-27 08:57:21,148: [worker  0] Epoch: 2
2020-08-27 08:57:21,148: [worker  0] Learning rate: 0.001
2020-08-27 08:57:21,148: [worker  0] ----------------------------------------------------------------------
2020-08-27 09:00:31,881: [worker  0] Eval   Loss: -2.280e+00
2020-08-27 09:00:31,881: [worker  0] Eval  Speed: 254.57 [samples/sec]
[...]

PR #920, opened 2020-09-25
Add ConvTasNet model

mthrok:
Replacing #893 to commit to master branch directly.
See #897 for the overview of ConvTasNet work.


torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+
+    References:
+        - Conv-TasNet: Surpassing Ideal Time--Frequency Magnitude Masking for Speech Separation
+          Luo, Yi and Mesgarani, Nima
+          https://arxiv.org/abs/1809.07454
+    """
+
+    def __init__(
+        self,
+        io_channels: int,
```
vincentqb, line 33: name mismatch doc?

torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+            torch.nn.GroupNorm(num_groups=1, num_channels=hidden_channels, eps=1e-08),
+        )
+
+        self.res_out = (
+            None
+            if no_residual
+            else torch.nn.Conv1d(
+                in_channels=hidden_channels, out_channels=io_channels, kernel_size=1
+            )
+        )
```
vincentqb, line 66: nit: echo if-block style of how res_out is called?

torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+    Generates masks for separation.
+
+    Args:
+        input_dim (int): Input feature dimension, <N>.
+        num_sources (int): The number of sources to separate.
+        kernel_size (int): The convolution kernel size of conv blocks, <P>.
+        num_featrs (int): Input/output feature dimenstion of conv blocks, <B, Sc>.
+        num_hidden (int): Intermediate feature dimention of conv blocks, <H>
+        num_layers (int): The number of conv blocks in one stack, <X>.
+        num_stacks (int): The number of conv block stacks, <R>.
```
vincentqb, line 95: nit: num_* or n_*? [readme](https://github.com/pytorch/audio/blob/master/README.md) points to `n_*`

torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+
+    Args:
+        num_sources (int): The number of sources to split.
+        enc_kernel_size (int): The convolution kernel size of the encoder/decoder, <L>.
+        enc_num_feats (int): The feature dimensions passed to mask generator, <N>.
+        msk_kernel_size (int): The convolution kernel size of the mask generator, <P>.
+        msk_num_feats (int): The input/output feature dimension of conv block in the mask generator, <B, Sc>.
+        msk_num_hidden_feats (int): The internal feature dimension of conv block of the mask generator, <H>.
+        msk_num_layers (int): The number of layers in one conv block of the mask generator, <X>.
+        msk_num_stacks (int): The numbr of conv blocks of the mask generator, <R>.
```
vincentqb, line 187: nit: `msk_feats` instead?

torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+        self.decoder = torch.nn.ConvTranspose1d(
+            in_channels=enc_num_feats,
+            out_channels=1,
+            kernel_size=enc_kernel_size,
+            stride=self.enc_stride,
+            padding=self.enc_stride,
+            bias=False,
+        )
+
+    def _align_num_frames_with_strides(
```
vincentqb, line 244: nit: `align_frames_with_strides` ?

torchaudio/models/conv_tasnet.py
```diff
@@ -0,0 +1,323 @@
...
+        self, input: torch.Tensor
+    ) -> Tuple[torch.Tensor, int]:
+        """Pad input Tensor so that the end of the input tensor corresponds with
```
vincentqb, line 247: the high level goal of the function is to have overlap between tensors in a batch?

vincentqb (commented):
i prefer how the code looked with sequential? but let's go with @cpuhrsch recommendation of avoiding using sequential as currently done here.

vincentqb (approved):
The test allows us to see that the model can be ran (forward pass) and get the expected shape. The pipeline is providing the correctness of the model introduced here. LGTM. @cpuhrsch suggests adding deterministic value in the model and checking the output gives what it's expected. I would be ok adding this now or as a follow-up.

One comment is the choice of parameter names `num_*`: on one hand it fits the existing ones, but disagree with readme (e.g. `n_*`). In particular the name `msk_num_*` which seems strange. Thoughts?

test/torchaudio_unittest/models_test.py
```diff
@@ -115,3 +125,58 @@ def test_waveform(self):
...
+            _ConvTasNetParams(512, 40, 256, 512, 3, 7, 2),
+            _ConvTasNetParams(512, 40, 256, 512, 3, 7, 2),
+            _ConvTasNetParams(512, 40, 128, 512, 3, 6, 4),
+            _ConvTasNetParams(512, 40, 128, 512, 3, 4, 6),
+            _ConvTasNetParams(512, 40, 128, 512, 3, 8, 3),
+            _ConvTasNetParams(512, 32, 128, 512, 3, 8, 3),
+            _ConvTasNetParams(512, 16, 128, 512, 3, 8, 3),
+        ],
+    )))
+    def test_paper_configuration(self, num_sources, model_params):
```
cpuhrsch, line 164: Another test to add here is feed parameters some deterministic values (GPUs are not deterministic, so it'll be CPU)
Super low tech:
```
def gen_deterministic(shape):
    numel = prod(shape)
    return torch.arange(t.numel().mod(20) / 10).reshape(shape).double)
```
```
for p in model.parameters():
    p.data.copy_(gen_deterministic(p.data.shape))
tensor = gen_deterministic((batch_size, 1, num_frames))
output = model(tensor)
output.sum() == <some reference number>
```

cpuhrsch (approved):

torchaudio/models/conv_tasnet.py
mthrok: Ternary operator expresses the intention of execution flow better. I-else block in forward is necessary for torchscript compatibility

torchaudio/models/conv_tasnet.py
mthrok: looks like torch uses `num_`.
Changing this now requires to retrain model or model surgery.
since it's labeled nit I will keep it as is.

torchaudio/models/conv_tasnet.py
mthrok: `msk_feats` is not descriptive enough. could mean type definition or anything about feats,

mthrok:
> i prefer how the code looked with sequential? but let's go with @cpuhrsch recommendation of avoiding using sequential as currently done here.

The point is not about preference, but about the flexibility. Not using Sequential container will allow users to change just the controls flow in forward method. This is better because if you use pertained weight, the structure of the model is encoded in the binary file. In this case if you want to change the execution flow in `forward`, you need to perform some kind of model surgery over the binary file. It is not possible but cumbersome. Not using Sequential container removes this burden.

PR #922, opened 2020-09-28
Add soundfile compatibility backend 

mthrok:
As a part of the "sox" backend sunset plan (#903), we add a "soundfile" backend that is compatible with the "sox_io" backend.

This PR contains;
 - The implementation of the new "soundfile" backend.
 - The function and global flag to switch the behavior of "soundfile" backend.
    No new public backend is added. We provide a switch to change the interface/behavior of "soundfile" backend.
 - Test for the new backend.

The default behavior of "soundfile" backend is not changed. The users who want to opt-in the new "soundfile" interface can do so by `torchaudio.backend.utils.use_soundfile_legacy_interface(False)` before changing the backend to "soundfile".

In 0.8.0 release, the "soundfile" backend will use this interface by default, and users can still use the legacy one with `torchaudio.backend.utils.use_soundfile_legacy_interface(True)`. In 0.9.0, the legacy interface is removed and `use_soundfile_legacy_interface` function becomes no-op, (and eventually removed from the code base.)

TODO: 
- [ ] Update #903

### DIFF
--- a/torchaudio/backend/utils.py
+++ b/torchaudio/backend/utils.py
@@ -16,7 +16,6 @@ __all__ = [
     'list_audio_backends',
     'get_audio_backend',
     'set_audio_backend',
-    'use_soundfile_legacy_interface',
 ]
 
 
@@ -33,16 +32,6 @@ def list_audio_backends() -> List[str]:
         backends.append('sox')
         backends.append('sox_io')
     return backends
-
-
-_USE_SOUNDFILE_LEGACY_INTERFACE = True
-
-
-def use_soundfile_legacy_interface(value: bool):
-    """Switch soundfile backend interface.
-    """
-    global _USE_SOUNDFILE_LEGACY_INTERFACE
-    _USE_SOUNDFILE_LEGACY_INTERFACE = value
 
 
 def set_audio_backend(backend: Optional[str]):
@@ -70,12 +59,12 @@ def set_audio_backend(backend: Optional[str]):
     elif backend == 'sox_io':
         module = sox_io_backend
     elif backend == 'soundfile':
-        if _USE_SOUNDFILE_LEGACY_INTERFACE:
+        if torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE:
             warnings.warn(
                 'The interface of "soundfile" backend is planned to change in 0.8.0 to '
                 'match that of "sox_io" backend and the current interface will be removed in 0.9.0. '
                 'To use the new interface, do '
-                '`torchaudio.backend.utils.use_soundfile_legacy_interface(False)` '
+                '`torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False` '
                 'before setting the backend to "soundfile". '
                 'Please refer to https://github.com/pytorch/audio/issues/903 for the detail.'
             )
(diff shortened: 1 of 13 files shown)


torchaudio/backend/utils.py
```diff
@@ -33,6 +35,16 @@ def list_audio_backends() -> List[str]:
     return backends
 
 
+_USE_SOUNDFILE_LEGACY_INTERFACE = True
```
cpuhrsch, line 38: Alternatively (and probably saver) you can do
```
torchaudio._USE_SOUNDFILE_LEGACY_INTERFACE=True
```
and create a _USE_SOUNDFILE_LEGACY_INTERFACE in the ```torchaudio/__init__.py``` file.

cpuhrsch: What about someone doing

```
def fn1():
    import torchaudio
    torchaudio.backend.utils.use_soundfile_legacy_interface(True)
    def _fn():
         torchaudio.set_backend('soundfile')
         torchaudio.<do_something>

def fn2(
    import torchaudio
    torchaudio.backend.utils.use_soundfile_legacy_interface(False)

fn = fn1()
fn2()
fn() # Is this using the legacy or non-legacy version?
```

Maybe this is fixed if instead of using the global, you create a variable in ```torchaudio/__init__.py```?

torchaudio/backend/utils.py
```diff
@@ -33,6 +35,16 @@ def list_audio_backends() -> List[str]:
...
+
+
+def use_soundfile_legacy_interface(value: bool):
+    """Switch soundfile backend interface.
+    """
+    global _USE_SOUNDFILE_LEGACY_INTERFACE
```
cpuhrsch, line 44: I'm not even sure you need to use the ```global``` keyword here

cpuhrsch: You might also want a getter and a setter to help send warnings when this variable is being changed a lot

cpuhrsch (approved):

torchaudio/backend/utils.py
mthrok: Okay, so it seems that accessing the variable directly has more advantage (like users can check the current configuration), so I will move to it.

mthrok: Moved "USE_SOUNDFILE_LEGACY_INTERFACE" to `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE` and got rid of the helper function.

vincentqb:
> In 0.8.0 release, the "soundfile" backend will use this interface by default, and users can still use the legacy one with `torchaudio.backend.utils.use_soundfile_legacy_interface(True)`. In 0.9.0, the legacy interface is removed and `use_soundfile_legacy_interface` function becomes no-op, (and eventually removed from the code base.)

I'll edit the description to reflect the correct interface:
```python
torchaudio.backend.utils.use_soundfile_legacy_interface = False
```

mthrok:
> > In 0.8.0 release, the "soundfile" backend will use this interface by default, and users can still use the legacy one with `torchaudio.backend.utils.use_soundfile_legacy_interface(True)`. In 0.9.0, the legacy interface is removed and `use_soundfile_legacy_interface` function becomes no-op, (and eventually removed from the code base.)
> 
> I'll edit the description to reflect the correct interface:
> 
> ```python
> torchaudio.backend.utils.use_soundfile_legacy_interface = False
> ```

It's `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE`.

PR #925, opened 2020-09-29
Update index.rst

brianjo:
This PR adds links to the PyTorch libraries to the left navigation of the site.


brianjo:
Actually, I'm going to mock this up first.

PR #927, opened 2020-09-29
levenshtein distance

vincentqb:
This PR takes the levenshtein distance from #632 and moves it to torchaudio. This also adds the docstring from vincentqb/audio#3.

cc [notebook](https://github.com/vincentqb/audio-tutorial/blob/master/Levenshtein.ipynb)


test/torchaudio_unittest/metrics.py
```diff
@@ -0,0 +1,20 @@
+from parameterized import parameterized
+from torchaudio.metrics import levenshtein_distance
+from torchaudio_unittest import common_utils
+
+
+class TestLevenshteinDistance(common_utils.TorchaudioTestCase):
+    @parameterized.expand(
```
cpuhrsch, line 7: I'd add edge cases such as ["abc", "", 3] or ["", "", 0] as well

torchaudio/metrics.py
```diff
@@ -1,9 +1,17 @@
 from typing import List, Union
 
 
-def levenshtein_distance(r: Union[str, List[str]], h: Union[str, List[str]]):
+def levenshtein_distance(r: Union[str, List[str]], h: Union[str, List[str]]) -> int:
```
cpuhrsch, line 4: From a minimalism perspective, I can see that List[str] arguments can be useful, but aside from syntactic sugar, what else do they offer users?

test/torchaudio_unittest/metrics.py
```diff
@@ -0,0 +1,20 @@
...
+        [
+            ["abc", "abc", 0],
+            ["aaa", "aba", 1],
+            ["aba", "aaa", 1],
+            ["aa", "aaa", 1],
+            ["aaa", "aa", 1],
+            ["abc", "bcd", 2],
+            [["hello", "world"], ["hello", "world", "!"], 1],
```
cpuhrsch, line 15: What would [["hello", "world"], "world"] do?

torchaudio/metrics.py
cpuhrsch: Also, can you mix List[str] and str?

torchaudio/metrics.py
```diff
@@ -1,9 +1,17 @@
...
     """
     Calculate the Levenshtein distance between two lists or strings.
+
+    The function computes an edit distance allowing deletion, insertion and substitution.
+    The result is an integer. Users may want to normalize by the length of the reference.
```
cpuhrsch, line 9: This should imo also include an explanation of what happens when the user passes a list of strings and how that is different.

torchaudio/metrics.py
vincentqb: The function runs with any two arbitrary sequences. I updated this. That being said, a user would only really using this for two Sequence[T] for the same T.

test/torchaudio_unittest/metrics.py
vincentqb: It would said that ["hello", "world"] is not the same as "world", so the edit distance would be 1 replacement. This is a particular case of [comment](https://github.com/pytorch/audio/pull/927#discussion_r497630708).

vincentqb: We could decide to detect such cases and compute the edit distance between the word, and each of the word, and return a list of edit distance instead.

This then lead to the question: what API do we expect a metric to offer, and how would this one align?

torchaudio/metrics.py
vincentqb: Agree, and updated. Doc may need to be updated further following discussion from [comment](https://github.com/pytorch/audio/pull/927#discussion_r497641380).

torchaudio/metrics.py
vincentqb: relate to [comment](https://github.com/pytorch/audio/pull/927#discussion_r497641380).

test/torchaudio_unittest/metrics.py
vincentqb: If we go that path, then we'd treat as special cases:
- `[Sequence[T]], Sequence[T]]]`
- `[Sequence[T], List[Sequence[T]]]`
- `[List[Sequence[T]], Sequence[T]]`
- `[List[Sequence[T]], List[Sequence[T]]]`

The last could be a little ambiguous. batch compare 1-1, compare all possible pairs, or edit distance between the two lists? the first seems most likely. Thoughts?

The two current use cases are 
- two strings
- two sentences (list of strings)

torchaudio/metrics.py
```diff
@@ -0,0 +1,40 @@
+from collections.abc import Sequence
```
mthrok, line 1: Does this really deserve to be in a new namespace, for user-side API perspective?
I could also see that this can live inside of `torchaudio.functionals`.

mthrok: Splitting the implementation to the dedicated module makes sense. `functionals.py` grew too large in my opinion, but we can also import functions defined in this module in `functionals.py` so that users can access it via `torchaudio.functionals`

vincentqb: That's a very good question. We'll need to follow-up on where this should go. `functionals.py` does seem like a good candidate.

test/torchaudio_unittest/metrics.py
cpuhrsch: I'd stick to "two strings" and add the other behavior later on, if necessary. I know you already use it in your example, but how much worse does the code become?

vincentqb: I'm not sure I follow your suggestiong: are you saying the function only handles a reference and hypothesis string: `levenshtein_distance("Hello", "ello")`?

How do you suggest comparing the edit distance between the sentences "Hello World!" and "Bonjour  World!"? Right now, the sentence is first split into by spaces, and the distance function is then applied on the two lists of words: `levenshtein_distance(["Hello", "World!"], ["Bonjour", "World!"])`.

vincentqb:
Closed by #1601

Issue #926, opened 2020-09-29
Add CONTRIB.md


vincentqb:
We should add a CONTRIB.md file to empower users in contributing more.

cc [comment](https://github.com/pytorch/audio/pull/566#pullrequestreview-397433345), T57210445

vincentqb:
It should

- point to [the README for tests](https://github.com/pytorch/audio/tree/master/test/torchaudio_unittest), 
- remind to test for correctness against librosa/sox/scipy/kaldi/etc, torchscriptability, batchable
- add documentation

Thoughts?

bhargavkathivarapu:
@vincentqb , I would like to help with this . How do you want to organize this ?
I am thinking of below organization. thoughts ?
1. Basic setups & installations
2. Some resources to understand pytorch tensors 
3. A generic workflow from raise issue , code change , code linting , submitting PR , merging 
4. Then some different ways of contributing and details,checklist specific to each type of contribution 
    - Contribute to Datasets
    - Contribute to Transforms
    - Contribute to Functional 
    - Contribute to Examples
    - Contribute to documentation 
    - Other contributions

Issue #924, opened 2020-09-29
A Law encoding


dvisockas:
## 🚀 Feature
I would like to propose adding of [A-Law](https://en.wikipedia.org/wiki/A-law_algorithm) to `torchaudio.transforms`

## Motivation

A-law is a companding algorithm similar to mu-law (that is already in `torchaudio`. It is used in European 8-bit PCM systems. By convention, A-law is used for an international connection if at least one country uses it. With these transforms `torchaudio` would 
enable even wider modelling capabilities of telephony audio. 

## Pitch

I imagine the API could be almost identical to `MuLaw` with one caveat - the compression parameter _A_ would be an additional parameter. So at the end there would be two additional transforms `ALawEncoding` and `ALawDecoding`. 

## Additional context

I have already written the code and would be happy to open PR.

vincentqb:
Yes, this would be a good addition :)

To verify correctness, do you have a reference implementation to test and compare against? I'm adding a few pointers to make sure we cover all that needs to be covered [here](https://github.com/pytorch/audio/issues/926#issuecomment-700978174) for a new transform to be added. If you have any questions about them before opening a pull request, please feel free to ask clarification :) Does that help?

dvisockas:
Haven't found a reference implementation, but will look for it :)

dvisockas:
Does the TODO in PR #930 look good?

vincentqb:
> Does the TODO in PR #930 look good?

I've updated the [comment](https://github.com/pytorch/audio/issues/926#issuecomment-700978174) to make it clearer what's entailed :)

PR #930, opened 2020-09-30
A Law

dvisockas:
TODO:

- [ ] Point to the README for tests
- [ ] Test for correctness against librosa/sox/scipy/kaldi/etc, torchscriptability, batchable
- [ ] Add documentation


vincentqb:
Thanks for working on this :)

The main thing we need is a way of knowing that the algorithm is correct (now and in the future). Do you have a reference you used, e.g. a paper, a book, or a software? Looking at [wikipedia](https://en.wikipedia.org/wiki/A-law_algorithm) as you pointed [here](https://github.com/pytorch/audio/issues/924#issuecomment-701237059), I see two graphs that could be used for testing. A first test would consist of a python implementation for the graph "Plot of F(x) for A-Law for A = 87.6", and then compare with pytorch. Maybe the other graph can also be turned into a second test.

I believe the first test would be enough, along with a link to wikipedia, which does provide the equation.

vincentqb:
As mentioned [here](https://github.com/pytorch/audio/issues/926#issuecomment-700978174), let me clarify the things that need to be done
- [x] open an issue first to get feedback before coding :)
- [ ] test for correctness against librosa/sox/scipy/kaldi/etc
- [ ] test for torchscriptability on CPU and GPU, if possible
- [ ] test the operation is batchable
- [ ] add documentation
- [x] avoid additional dependencies in the library (e.g. numpy, scipy, etc)

Does [the readme for tests](https://github.com/pytorch/audio/blob/master/test/README.md) give you enough clarity?

dvisockas:
For the implementation, I used wikipedia formula as a reference. Now I found that [python has alaw in its stdlib](https://docs.python.org/3/library/audioop.html). 

The question is then should I just run a sample wave through `audioop.lin2alaw` and torchaudio and test for the results to match and then do the same for decoding in `transforms_test.py`?

How can I test "for torchscriptability on CPU and GPU" and "the operation is batchable"?

Regarding the documentation - I added an A Law entry in the `transforms.rst`, will that be enough or there are additional things I have to do?

vincentqb:
Thanks again for working on this :)

> For the implementation, I used wikipedia formula as a reference. Now I found that [python has alaw in its stdlib](https://docs.python.org/3/library/audioop.html).
> 
> The question is then should I just run a sample wave through `audioop.lin2alaw` and torchaudio and test for the results to match and then do the same for decoding in `transforms_test.py`?

Good catch! Yes, that is a great way to test for correctness. :)

> How can I test "for torchscriptability on CPU and GPU" and "the operation is batchable"?

For torchscript-ability, all you have to do is add a test like [this one](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/torchscript_consistency_impl.py#L507). The test will then be ran automatically on CPU and GPU in CircleCI :) It checks that the python version and the jitted version of a transform give the same result.

For batch-ability, all you need to do is add a test like [this one](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/batch_consistency_test.py#L260). The test simply checks that "running 3 times the same transformation" versus "running one time on a batch tensor with 3 copies" give the same results.

> Regarding the documentation - I added an A Law entry in the `transforms.rst`, will that be enough or there are additional things I have to do?

Adding an entry for the functional in [here](https://github.com/pytorch/audio/blob/master/docs/source/functional.rst) and for the transform in [here](https://github.com/pytorch/audio/blob/master/docs/source/transforms.rst) as you point out.

torchaudio/transforms.py
```diff
@@ -567,6 +569,61 @@ def forward(self, x_mu: Tensor) -> Tensor:
...
+    def forward(self, x_mu: Tensor) -> Tensor:
+        r"""
+        Args:
+            x_a (Tensor): An A-law encoded signal which needs to be decoded.
+
+        Returns:
+            Tensor: The signal decoded.
+        """
+        return F.a_law_decoding(x_a, self.quantization_channels, self.compression_param)
+
```
vincentqb, line 626: Is `x_mu` equal to `x_a`? How about just using `x`? this seems to lead to [error](https://app.circleci.com/pipelines/github/pytorch/audio/3546/workflows/7f02a96c-2ad6-426c-b809-41fc6e12c192/jobs/104322)

test/torchaudio_unittest/torchscript_consistency_impl.py
```diff
@@ -137,6 +137,22 @@ def func(tensor):
...
+            qc = 256
+            return F.a_law_encoding(tensor, qc)
+
+        waveform = common_utils.get_whitenoise()
+        self._assert_consistency(func, waveform)
+
+    def test_a_law_decoding(self):
+        def func(tensor):
+            qc = 256
+            return F.a_law_decoding(tensor, qc)
```
vincentqb, line 151: nit: the definition requires 3 parameters

vincentqb:
> Added documentation, tested for torchscriptability anr batchablity, yet the tests seem to be failing. Is there a way to inspect why?

There should be a "details" button next to unittest section in the checks list (I believe this is public :) ). On the following page, we can select a system and click. This should lead to [here](https://app.circleci.com/pipelines/github/pytorch/audio/3546/workflows/7f02a96c-2ad6-426c-b809-41fc6e12c192/jobs/104322) for instance. Does that help?

test/torchaudio_unittest/transforms_test.py
```diff
@@ -43,6 +43,20 @@ def test_mu_law_companding(self):
...
 
+    def test_a_law_companding(self):
+        waveform = self.waveform.clone()
+        if not waveform.is_floating_point():
+            waveform = waveform.to(torch.get_default_dtype())
+        waveform /= torch.abs(waveform).max()
+
+        self.assertTrue(waveform.min() >= -1. and waveform.max() <= 1.)
+
+        waveform_a = transforms.ALawEncoding(quantization_channels)(waveform)
```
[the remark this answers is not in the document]
dvisockas: Fixed this

test/torchaudio_unittest/torchscript_consistency_impl.py
```diff
@@ -137,6 +137,22 @@ def func(tensor):
         tensor = torch.rand((1, 10))
         self._assert_consistency(func, tensor)
 
+    def test_a_law_encoding(self):
+        def func(tensor):
+            qc = 256
+            return F.a_law_encoding(tensor, qc)
```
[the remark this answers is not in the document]
dvisockas: Fixed :)

test/torchaudio_unittest/torchscript_consistency_impl.py
dvisockas: Fixed :)

torchaudio/transforms.py
dvisockas: I named the input `x_a` since it is an `x` encoded with `a` law. In `MuLaw` it is called `x_mu` so I wanted to name it consistently

vincentqb:
btw the errors below are fixed on master (#934), so rebasing will fix them :)
```
self = <torchaudio.datasets.tedlium.TEDLIUM object at 0x000002028AD91B50>
path = 'C:\\Users\\circleci\\AppData\\Local\\Temp\\tmpobcfx03l\\tedlium\\TEDLIUM_release2\\train\\sph\\release2.sph'
start_time = 0, end_time = 32000, sample_rate = 16000

    def _load_audio(self, path: str, start_time: float, end_time: float, sample_rate: int = 16000) -> [Tensor, int]:
        """Default load function used in TEDLIUM dataset, you can overwrite this function to customize functionality
        and load individual sentences from a full ted audio talk file.
    
        Args:
            path (str): Path to audio file
            start_time (int, optional): Time in seconds where the sample sentence stars
            end_time (int, optional): Time in seconds where the sample sentence finishes
    
        Returns:
            [Tensor, int]: Audio tensor representation and sample rate
        """
        start_time = int(float(start_time) * sample_rate)
        end_time = int(float(end_time) * sample_rate)
        if torchaudio.get_audio_backend() == "sox_io":
            return torchaudio.load(path, frame_offset=start_time, num_frames=end_time - start_time)
>       return torchaudio.load(path)[:, start_time:end_time]
E       TypeError: tuple indices must be integers or slices, not tuple

..\env\lib\site-packages\torchaudio-0.7.0a0+8896b41-py3.8.egg\torchaudio\datasets\tedlium.py:171: TypeError
```

vincentqb:
I'm sure you've noticed :) but for reference: there's some lint errors 
```
torchaudio/functional.py:409:1: E302 expected 2 blank lines, found 1
torchaudio/transforms.py:573:1: E302 expected 2 blank lines, found 1
flake8 failed. Check the format of Python files.
```
and the new test is failing 
```
self = <torchaudio_unittest.transforms_test.Tester testMethod=test_a_law_companding>

    def test_a_law_companding(self):
    
        quantization_channels = 256
        compression_param = 83.7
    
        waveform = self.waveform.clone()
        if not waveform.is_floating_point():
            waveform = waveform.to(torch.get_default_dtype())
        waveform /= torch.abs(waveform).max()
    
        self.assertTrue(waveform.min() >= -1. and waveform.max() <= 1.)
    
        waveform_a = transforms.ALawEncoding(quantization_channels, compression_param)(waveform)
        self.assertTrue(waveform_a.min() >= 0. and waveform_a.max() <= quantization_channels)
    
        waveform_exp = transforms.ALawDecoding(quantization_channels, compression_param)(waveform_a)
        self.assertTrue(waveform_exp.min() >= -1. and waveform_exp.max() <= 1.)
    
        segment_length = 1
        small_int_waveform = waveform.to(torch.uint8)
        waveform_bytes = bytearray(small_int_waveform[0, :])
    
        encoded = audioop.lin2alaw(waveform_bytes, segment_length)
        torch_encoded = transforms.ALawEncoding(quantization_channels, compression_param)(small_int_waveform)
>       self.assertEqual(torch.tensor(list(encoded)), torch_encoded[0])

torchaudio_unittest/transforms_test.py:71: 
../env/lib/python3.8/site-packages/torch/testing/_internal/common_utils.py:1139: in assertEqual
    self.assertTrue(result, msg=msg)
E   AssertionError: False is not true : Tensors failed to compare as equal! Found 64000 different element(s) (out of 64000), with the greatest difference of 295 (90 vs. 385) occuring at index 200.
```

dvisockas:
Yes, I noticed. Just wanted to push some changes to have an ability to work on other computer.

Regarding the tests, `audioop` gives different results when encoding with `ALaw`. The same is with `MuLaw`. Was `MuLaw` tested for correctness?

vincentqb:
> Yes, I noticed. Just wanted to push some changes to have an ability to work on other computer.
> 
> Regarding the tests, `audioop` gives different results when encoding with `ALaw`. The same is with `MuLaw`. Was `MuLaw` tested for correctness?

No, this was not against audioop. In fact, if you notice some difference, could you open an issue with a code snippet so we can look into it? Thanks!

arlofaria-cartesia:
> > Yes, I noticed. Just wanted to push some changes to have an ability to work on other computer.
> > Regarding the tests, `audioop` gives different results when encoding with `ALaw`. The same is with `MuLaw`. Was `MuLaw` tested for correctness?
> 
> No, this was not against audioop. In fact, if you notice some difference, could you open an issue with a code snippet so we can look into it? Thanks!

I have also noticed this discrepancy.  Here's a test:

```python
import audioop  # NOTE: will be removed from Python stdlib at v3.13!

import torchaudio

# Load as normalized fp32.
waveform, _ = torchaudio.load("https://mod9.io/hi.wav", normalize=True)
ulaw_encoded_torchaudio = torchaudio.functional.mu_law_encoding(waveform, quantization_channels=256)

# Load as int16.
waveform, _ = torchaudio.load("https://mod9.io/hi.wav", normalize=False)
ulaw_encoded_audioop = audioop.lin2ulaw(waveform.numpy().tobytes(), 2)

print("torchaudio", ulaw_encoded_torchaudio[0, :16].tolist())
print("audioop   ", list(ulaw_encoded_audioop)[:16])
```

This prints:

```
torchaudio [147, 148, 143, 138, 137, 136, 132, 129, 128, 128, 127, 124, 123, 123, 123, 121]
audioop    [236, 236, 239, 245, 247, 248, 251, 254, 254, 254, 126, 124, 123, 123, 123, 121]
```

I think the difference might be because TorchAudio is using the Wikipedia definition of the mu-law algorithm, but other audio tools tend to use the reference C implementation of the G711, e.g.:
https://github.com/escrichov/G711/blob/master/g711.c#L202-L263

PR #931, opened 2020-10-01
Update docstrings/documentations of all the datasets

mthrok:
Currently the documentation page does not list constructor arguments or overwrote `__getitem__` methods. This PR fixes it.

(Note: the annotation is broken form the beginning and I cannot fix it.)

Example: 
Screen Shot 2020-10-01 at 18 10 45

### DIFF
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -29,82 +29,85 @@ CMUARCTIC
 ~~~~~~~~~
 
 .. autoclass:: CMUARCTIC
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 COMMONVOICE
 ~~~~~~~~~~~
 
 .. autoclass:: COMMONVOICE
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 GTZAN
 ~~~~~
 
 .. autoclass:: GTZAN
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 LIBRISPEECH
 ~~~~~~~~~~~
 
 .. autoclass:: LIBRISPEECH
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 LIBRITTS
 ~~~~~~~~
 
 .. autoclass:: LIBRITTS
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 LJSPEECH
 ~~~~~~~~
 
 .. autoclass:: LJSPEECH
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 SPEECHCOMMANDS
 ~~~~~~~~~~~~~~
 
 .. autoclass:: SPEECHCOMMANDS
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 TEDLIUM
 ~~~~~~~~~~~~~~
 
 .. autoclass:: TEDLIUM
-  :members: __getitem__
-  :special-members: get_phoneme_dict
+  :members:
+  :special-members: __getitem__
+
 
 VCTK
 ~~~~
 
 .. autoclass:: VCTK
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
 
 
 VCTK_092
 ~~~~~~~~
 
 .. autoclass:: VCTK_092
+  :members:
+  :special-members: __getitem__
 
 
 YESNO
 ~~~~~
 
 .. autoclass:: YESNO
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -96,7 +96,7 @@ class CMUARCTIC(Dataset):
                  root: str,
                  url: str = URL,
                  folder_in_archive: str = FOLDER_IN_ARCHIVE,
-                 download: bool = False):
+                 download: bool = False) -> None:
 
         if url in [
             "aew",
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -76,9 +76,15 @@ def load_cmuarctic_item(line: str,
 
 
 class CMUARCTIC(Dataset):
-    """
-    Create a Dataset for CMU_arctic. Each item is a tuple of the form:
-    waveform, sample_rate, utterance, utterance_id
+    """Create a Dataset for CMU_ARCTIC.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional):
+            Type of the dataset to dowload. This is **NOT** the actual URL. (default: ``"aew"``)
+        folder_in_archive (str, optional):
+            The top-level directory of the dataset. (default: ``"ARCTIC"``)
+        download (bool, optional): Download dataset if it is not found at root path. (default: ``False``).
     """
 
     _file_text = "txt.done.data"
@@ -90,7 +96,7 @@ class CMUARCTIC(Dataset):
                  root: str,
                  url: str = URL,
                  folder_in_archive: str = FOLDER_IN_ARCHIVE,
-                 download: bool = False) -> None:
+                 download: bool = False):
 
         if url in [
             "aew",
@@ -143,6 +149,14 @@ class CMUARCTIC(Dataset):
             self._walker = list(walker)
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str]:
+        """Load the n-th sample from the dataset.
+
+        Args:
+            n (int): The index of the sample to be loaded
+
+        Returns:
+            tuple: ``(waveform, sample_rate, utterance, utterance_id)``
+        """
         line = self._walker[n]
         return load_cmuarctic_item(line, self._path, self._folder_audio, self._ext_audio)
 
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -84,7 +84,8 @@ class CMUARCTIC(Dataset):
             Type of the dataset to dowload. This is **NOT** the actual URL. (default: ``"aew"``)
         folder_in_archive (str, optional):
             The top-level directory of the dataset. (default: ``"ARCTIC"``)
-        download (bool, optional): Download dataset if it is not found at root path. (default: ``False``).
+        download (bool, optional):
+            Whether to download the dataset if it is not found at root path. (default: ``False``).
     """
 
     _file_text = "txt.done.data"
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -82,6 +82,9 @@ class CMUARCTIC(Dataset):
         root (str): Path to the directory where the dataset is found or downloaded.
         url (str, optional):
             Type of the dataset to dowload. This is **NOT** the actual URL. (default: ``"aew"``)
+            The valid choice is ``"aew"``, ``"ahw"``, ``"aup"``, ``"awb"``, ``"axb"``, ``"bdl"``,
+            ``"clb"``, ``"eey"``, ``"fem"``, ``"gka"``, ``"jmk"``, ``"ksp"``, ``"ljm"``, ``"lnh"``,
+            ``"rms"``, ``"rxr"``, ``"slp"`` or ``"slt"``.
         folder_in_archive (str, optional):
             The top-level directory of the dataset. (default: ``"ARCTIC"``)
         download (bool, optional):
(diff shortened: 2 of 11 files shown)


torchaudio/datasets/cmuarctic.py
```diff
@@ -90,7 +96,7 @@ def __init__(self,
                  root: str,
                  url: str = URL,
                  folder_in_archive: str = FOLDER_IN_ARCHIVE,
-                 download: bool = False) -> None:
```
mruberry, line 93: Interesting: why remove the typing annotation?

torchaudio/datasets/cmuarctic.py
```diff
@@ -76,9 +76,15 @@ def load_cmuarctic_item(line: str,
...
-    waveform, sample_rate, utterance, utterance_id
+    """Create a Dataset for CMU_ARCTIC.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional):
+            Type of the dataset to dowload. This is **NOT** the actual URL. (default: ``"aew"``)
+        folder_in_archive (str, optional):
+            The top-level directory of the dataset. (default: ``"ARCTIC"``)
+        download (bool, optional): Download dataset if it is not found at root path. (default: ``False``).
```
mruberry, line 87: nit: "Whether to download the dataset if ..."

torchaudio/datasets/cmuarctic.py
mthrok: This adds `-> None` to documentation, and I do not think it adds any value.

torchaudio/datasets/cmuarctic.py
```diff
@@ -76,9 +76,15 @@ def load_cmuarctic_item(line: str,
 
 
 class CMUARCTIC(Dataset):
-    """
-    Create a Dataset for CMU_arctic. Each item is a tuple of the form:
```
mruberry, line 80: Do you really want to remove the information about the structure of these datasets?

mthrok: It's moved to `__getitem__` method.

torchaudio/datasets/commonvoice.py
```diff
@@ -100,11 +100,17 @@ def load_commonvoice_item(line: List[str],
...
-    (waveform, sample_rate, dictionary)
-    where dictionary is a dictionary built from the tsv file with the following keys:
-    client_id, path, sentence, up_votes, down_votes, age, gender, accent.
+    """Create a Dataset for CommonVoice.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        tsv (str, optional): The name of the tsv file used to construct the metadata.
+            (default: ``"train.tsv"``)
+        url (str, optional): The name of the language to select the URL when downloading the dataset.
```
mruberry, line 109: Is there a better way to describe this? Maybe "the language of the dataset  to download?"

mthrok: Ya, that sounds better.

torchaudio/datasets/gtzan.py
```diff
@@ -998,12 +998,20 @@ def load_gtzan_item(fileid: str, path: str, ext_audio: str) -> Tuple[Tensor, str
...
 
-    Please see http://marsyas.info/downloads/datasets.html
-    if you are planning to use this dataset to publish results.
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional): The URL to download the dataset from.
+            (default: ``"http://opihi.cs.uvic.ca/sound/genres.tar.gz"``)
+        folder_in_archive (str, optional): The top-level directory of the dataset.
+        download (bool, optional): Download dataset if it is not found at root path. (default: ``False``).
+        subset (str, optional): Subset of the dataset. Either ``"training"``, ``"validation"`` or
```
mruberry, line 1013: Maybe: "Which subset of the dataset to use"? 

What happens when subset is None? I assume the full dataset is used. Maybe something like, "Which subset of the dataset to use. Either None, training, validation, or testing. If None the entire dataset is used. (default: None)." ?

torchaudio/datasets/cmuarctic.py
```diff
@@ -76,9 +76,15 @@ def load_cmuarctic_item(line: str,
...
-    waveform, sample_rate, utterance, utterance_id
+    """Create a Dataset for CMU_ARCTIC.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional):
+            Type of the dataset to dowload. This is **NOT** the actual URL. (default: ``"aew"``)
```
mruberry, line 84: How does a reader know which values they can use and what they do?

torchaudio/datasets/tedlium.py
```diff
@@ -140,7 +116,7 @@ def _load_tedlium_item(self, fileid: str, line: int, path: str) -> Tuple[Tensor,
             path (str): Dataset root path
 
         Returns:
-            Tedlium_item: A namedTuple containing [waveform, sample_rate, transcript, talk_id, speaker_id, identifier]
+            tuple: ``(waveform, sample_rate, transcript, talk_id, speaker_id, identifier)``
```
mruberry, line 119: Nice fix here.

torchaudio/datasets/vctk.py
```diff
@@ -252,6 +272,14 @@ def _load_sample(self, speaker_id: str, utterance_id: str, mic_id: str) -> Sampl
         return Sample(waveform, sample_rate, utterance, speaker_id, utterance_id)
 
     def __getitem__(self, n: int) -> Sample:
+        """Load the n-th sample from the dataset.
```
mruberry, line 275: Interesting that the tuple description is kept on the class docstring for this dataset.

torchaudio/datasets/cmuarctic.py
mthrok: Yeah, so that's the problem. I see the list of valid values like the followings and I wonder, what is the best way to add all of them???????

```
        if url in [
            "aew",
            "ahw",
            "aup",
            "awb",
            "axb",
            "bdl",
            "clb",
            "eey",
            "fem",
            "gka",
            "jmk",
            "ksp",
            "ljm",
	    "lnh",
            "rms",
            "rxr",
            "slp",
            "slt"
	]:
```

torchaudio/datasets/cmuarctic.py
mruberry: We actually had an issue recently in PyTorch to add these "-> None" typing annotations to our inits. Check out:

https://github.com/pytorch/pytorch/issues/45497

You might save yourself a similar issue by leaving them in.

torchaudio/datasets/vctk.py
mthrok: Oh this is because we intend to change NamedTuple to a normal tuple for consistency reason before the release. (so very soon) So this way I do not have to change the docstrings twice.

torchaudio/datasets/cmuarctic.py
mthrok: oh good to know. Okay I will revert it.

torchaudio/datasets/gtzan.py
mthrok: Yes, your assumption is correct. Updated the description.

torchaudio/datasets/cmuarctic.py
mthrok: okay added them as is.

torchaudio/datasets/vctk.py
mthrok: oh I miss red your comment. it's an oversight. I will remove it from there.

torchaudio/datasets/cmuarctic.py
mthrok: I think describing it in `__getitem__` has better locality, which is better for the developer experience. If something is changed on the implementation side, they only need to change the corresponding docstring and there are less chances that there are discrepancy between the docstring description and the actual implementation, compared to the case where the same information are repeated in class description.
Of course, we are having the current docathon to catch such discrepancies, so one can argue that it's okay to have such discrepancies, but it will confuse users on master branch still.

torchaudio/datasets/cmuarctic.py
```diff
@@ -76,9 +76,19 @@ def load_cmuarctic_item(line: str,
...
+            The valid choice is ``"aew"``, ``"ahw"``, ``"aup"``, ``"awb"``, ``"axb"``, ``"bdl"``,
```
mruberry, line 85: Nit: "The valid choice is" -> "Allowed url values are ..."

torchaudio/datasets/speechcommands.py
```diff
@@ -36,9 +36,16 @@ def load_speechcommands_item(filepath: str, path: str) -> Tuple[Tensor, int, str
...
 class SPEECHCOMMANDS(Dataset):
-    """
-    Create a Dataset for Speech Commands. Each item is a tuple of the form:
-    waveform, sample_rate, label, speaker_id, utterance_number
+    """Create a Dataset for Speech Commands.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional): Type of the dataset to dowload. This is **NOT** the actual URL.
+            (default: ``"speech_commands_v0.02"``)
```
mruberry, line 44: This could also use a more complete enumeration of the options (and possibly what each means). That can be a separate PR, too.

mruberry (approved):
This changes look like a significant improvement.

In the future it may be interesting to enumerate all supported inputs for each param along with a brief description of what they do.

docs/source/datasets.rst
```diff
@@ -29,82 +29,85 @@ CMUARCTIC
 ~~~~~~~~~
 
 .. autoclass:: CMUARCTIC
-  :members: __getitem__
-  :special-members:
+  :members:
+  :special-members: __getitem__
```
vincentqb, line 33: nit: out of curiosity, does this change the output?

torchaudio/datasets/libritts.py
```diff
@@ -65,9 +65,16 @@ def load_libritts_item(
...
 
 class LIBRITTS(Dataset):
-    """
-    Create a Dataset for LibriTTS. Each item is a tuple of the form:
-    waveform, sample_rate, original_text, normalized_text, speaker_id, chapter_id, utterance_id
+    """Create a Dataset for LibriTTS.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional): Type of the dataset to dowload. This is **NOT** the actual URL.
```
vincentqb, line 72: same about code path for url

torchaudio/datasets/speechcommands.py
```diff
@@ -36,9 +36,16 @@ def load_speechcommands_item(filepath: str, path: str) -> Tuple[Tensor, int, str
...
 
 class SPEECHCOMMANDS(Dataset):
-    """
-    Create a Dataset for Speech Commands. Each item is a tuple of the form:
-    waveform, sample_rate, label, speaker_id, utterance_number
+    """Create a Dataset for Speech Commands.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional): Type of the dataset to dowload. This is **NOT** the actual URL.
```
vincentqb, line 43: same about url

torchaudio/datasets/commonvoice.py
```diff
@@ -100,11 +100,27 @@ def load_commonvoice_item(line: List[str],
...
-    (waveform, sample_rate, dictionary)
-    where dictionary is a dictionary built from the tsv file with the following keys:
-    client_id, path, sentence, up_votes, down_votes, age, gender, accent.
+    """Create a Dataset for CommonVoice.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        tsv (str, optional): The name of the tsv file used to construct the metadata.
+            (default: ``"train.tsv"``)
+        url (str, optional): The language of the dataset to download.
```
vincentqb, line 109: same about url

torchaudio/datasets/cmuarctic.py
```diff
@@ -76,9 +76,19 @@ def load_cmuarctic_item(line: str,
...
+            ``"clb"``, ``"eey"``, ``"fem"``, ``"gka"``, ``"jmk"``, ``"ksp"``, ``"ljm"``, ``"lnh"``,
+            ``"rms"``, ``"rxr"``, ``"slp"`` or ``"slt"``.
```
vincentqb, line 87: same about url

torchaudio/datasets/speechcommands.py
mthrok: Updated as part of this PR.

torchaudio/datasets/librispeech.py
```diff
@@ -67,9 +67,16 @@ def load_librispeech_item(fileid: str,
...
 
 class LIBRISPEECH(Dataset):
-    """
-    Create a Dataset for LibriSpeech. Each item is a tuple of the form:
-    waveform, sample_rate, utterance, speaker_id, chapter_id, utterance_id
+    """Create a Dataset for LibriSpeech.
+
+    Args:
+        root (str): Path to the directory where the dataset is found or downloaded.
+        url (str, optional): Type of the dataset to dowload. This is **NOT** the actual URL.
```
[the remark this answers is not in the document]
mthrok: oh you are right. updated.

vincentqb (approved):
thanks for updating this, lgtm overall :)

PR #935, opened 2020-10-02
[doc] Update backend docstring/documentation

mthrok:
Update documentation for backend. This includes,

- Documentation for the new interface of `"soundfile"` backend.
- The timeline for the change of default backend and deprecation of `"sox"` backend and `"soundfile"` backend with the legacy interface.
- Deprecation warnings to the classes removed as part of backend removals.


docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
+``"soundfile"`` backend is built on ``PySoundFile``. You need to install ``PySoundFile`` separately. The current interface of this backend (legacy interface) will be changed in ``0.8.0`` (new interface) to match the ``"sox_io"`` backend.
 
-``sox`` backend is the original backend which is built on ``libsox``. This module is currently default but is known to have number of issues, such as wrong handling of WAV files other than 16-bit signed integer. Users are encouraged to use ``sox_io`` backend. This backend requires C++ extension module and is not available on Windows system.
+.. note::
+   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
 
-``sox_io`` backend is the new backend which is built on ``libsox`` and bound to Python with ``Torchscript``. This module addresses all the known issues ``sox`` backend has. Function calls to this backend can be Torchscriptable. This backend requires C++ extension module and is not available on Windows system.
+Changes in default backend and Deprecation
```
vincentqb, line 28: I'd keep that in the issue instead of adding to the doc here. 
* Adding it here means we now have two sources of truth? #903
* The issue is easier to close when the transition is over?

mthrok: I think putting this on documentation makes more sense (specifically on master branch documentation). The target audience of documentation is all the users, which include both the consumer and developers of other libraries. GitHub issues are typically targeted to developers and it focuses on development of the library.

> Adding it here means we now have two sources of truth? #903

I will remove the duplicated part from the issue and redirect users to documentation for the migration guideline.

> The issue is easier to close when the transition is over?

The issue can be used for contents that change, such as the list of TODO and related PRs.

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
+Overview
+~~~~~~~~
+
 :mod:`torchaudio.backend` module provides implementations for audio file I/O, using different backend libraries.
 To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
 
-.. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
+There are currently four implementations available.
```
cpuhrsch, line 12: I'd mention which one is the default and whether users actually need to worry about this in most cases

cpuhrsch: Also write out why you might care about reading this section and give an overview of its content.

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
-.. note::
-   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
+    * ``"sox"`` (deprecated)
+    * ``"sox_io"``
+    * ``"soundfile"``, legacy interface (deprecated)
+    * ``"soundfile"``, new interface
+
+``"sox"`` backend is the original backend which is built on ``libsox``. This backend requires C++ extension module and is not available on Windows system. This backend is currently the default backend when available, but is being deprecated. Please use ``"sox_io"`` backend.
```
cpuhrsch, line 19: nit: "This backend requires C++ extension module and is not available on Windows system." -> "This backend requires 
 the C++ extension module and is not available on Windows."

cpuhrsch: "This backend is currently the default backend when available, but is being deprecated" -> "This backend is currently the default when available, but is being deprecated"

cpuhrsch: "Please use ``"sox_io"`` backend." -> "Please use the ``"sox_io"`` backend."

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
-There are currently three implementations available.
+``"sox_io"`` backend is the new backend which is built on ``libsox``, and bound with ``TorchScript``. This backend requires C++ extension module and is not available on Windows system. Function calls to this backend are TorchScript-able. This backend will become the default backend in ``"0.8.0"`` release.
```
cpuhrsch, line 21: You use the word backend 4 times here

cpuhrsch: "and bound with ``TorchScript``" doesn't mean much to most users I'd say. Instead write that this implies.

cpuhrsch: "This backend requires C++ extension module" - Is this something I'd need to install or worry about in most cases? Or is it only something to worry about on Windows? If I'm on Windows, what can I do? (add this to the intro). You probably want to have a platform-specific suggestion or guidance in the intro (might simply be Windows vs. not-Windows) and also again whether the user needs to worry or more specifically when the user needs to worry (e.g. mp3 support).

Bonus points for mentioning we're working on resolving some of these issues, e.g. mp3 work in progress for Windows etc., and a link to the corresponding issue

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
-    * :ref:`sox<sox_backend>`
-    * :ref:`sox_io<sox_io_backend>`
-    * :ref:`soundfile<soundfile_backend>`
+``"soundfile"`` backend is built on ``PySoundFile``. You need to install ``PySoundFile`` separately. The current interface of this backend (legacy interface) will be changed in ``0.8.0`` (new interface) to match the ``"sox_io"`` backend.
```
cpuhrsch, line 23: "You need to install ``PySoundFile`` separately." - how do I get it?

cpuhrsch: "The current interface of this backend (legacy interface) will be changed in ``0.8.0`` (new interface) to match the ``"sox_io"`` backend." - More detail in the corresponding issue? Which one is it?

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
-``sox`` backend is the original backend which is built on ``libsox``. This module is currently default but is known to have number of issues, such as wrong handling of WAV files other than 16-bit signed integer. Users are encouraged to use ``sox_io`` backend. This backend requires C++ extension module and is not available on Windows system.
+.. note::
+   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
```
cpuhrsch, line 26: Ideally we'd even have a tutorial or some kind of example somewhere that illustrates this behavior.

cpuhrsch: nit: "torhcaudio" -> "torchaudio"

docs/source/backend.rst
cpuhrsch, line 28: nit: "Deprecation" is not a [proper noun](https://en.wikipedia.org/wiki/Proper_noun_and_common_noun), so it should be lower cases

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-``soundfile`` backend is built on ``PySoundFile``. You need to install ``PySoundFile`` separately.
+.. warning::
+   * The default backend for Linux/macOS will be changed from ``"sox"`` to ``"sox_io"`` in ``0.8.0`` release.
+   * ``"sox"``  backend will be removed in ``0.9.0`` release.
+   * The function signatures of ``"soundfile"`` backends are changed in ``0.8.0`` to match ``"sox_io"`` backend.
+   * To opt-in to the new signature of ``"soundfile"`` backend, do ``torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False`` **before** switching to ``"soundfile"``.
+
+   The following table summarizes the timeline for the changes and deprecations.
```
cpuhrsch, line 37: This is great! Want to make this part of an issue as well?

cpuhrsch: Do we want to write a migration guide tutorial?

docs/source/backend.rst
cpuhrsch: We probably want to redo https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
-.. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
+There are currently four implementations available.
 
-.. note::
-   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
+    * ``"sox"`` (deprecated)
+    * ``"sox_io"``
+    * ``"soundfile"``, legacy interface (deprecated)
+    * ``"soundfile"``, new interface
```
mruberry, line 17: Maybe mention that this will be the default in 0.8 if the C++ extension isn't available?

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 
 .. _sox_backend:
 
-Sox Backend
-~~~~~~~~~~~
+Sox Backend (Deprecated)
+~~~~~~~~~~~~~~~~~~~~~~~~
 
-``sox`` backend is available on ``torchaudio`` installation with C++ extension. It is currently not available on Windows system.
+``"sox"`` backend is available on ``torchaudio`` installation with C++ extension. It is currently not available on Windows system.
```
mruberry, line 81: The sox backend... installations with the C++ extension. It is not available on Windows.

docs/source/backend.rst
```diff
@@ -78,19 +122,19 @@ Sox IO Backend
 
 ``sox_io`` backend is available on ``torchaudio`` installation with C++ extension. It is currently not available on Windows system.
 
-This new backend is recommended over ``sox`` backend. You can switch from another backend to ``sox_io`` backend with the following;
+This new backend is recommended over ``sox`` backend, and will become the default backend in ``0.8.0`` release when C++ extension is present.
```
mruberry, line 125: "when the C++ extension is available" to be consistent with previous phrasing

docs/source/backend.rst
```diff
@@ -78,19 +122,19 @@ Sox IO Backend
...
+
+You can switch from another backend to ``sox_io`` backend with the following;
```
mruberry, line 127: to the sox_io backend

docs/source/backend.rst
```diff
@@ -78,19 +122,19 @@ Sox IO Backend
...
 
 .. code::
 
    torchaudio.set_audio_backend("sox_io")
 
-The function call to this backend can be Torchsript-able. You can apply :func:`torch.jit.script` and dump the object to file, then call it from C++ application.
+The function calls to this backend are TorchSript-able. You can apply :func:`torch.jit.script` and dump the object toa file, then call it from C++ application.
```
mruberry, line 133: Probably not for this PR, but this might be a good opportunity to elaborate with an example, or point the reader to another reference.

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
 
 Soundfile Backend
 ~~~~~~~~~~~~~~~~~
 
 ``soundfile`` backend is available when ``PySoundFile`` is installed. This backend works on ``torchaudio`` installation without C++ extension. (i.e. Windows)
 
-You can switch from another backend to ``soundfile`` backend with the following;
+``"soundfile"`` backend has two interfaces, legacy and new.
+
+* In ``0.7.0`` release, the legacy interface is used by default when switching to ``"soundfile"`` backend.
```
mruberry, line 162: In the 0.7.0 release... (analogous changes below)

mruberry: "to the soundfile backend"

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
+^^^^
 
 .. autofunction:: torchaudio.backend.soundfile_backend.save
+
+.. _soundfile_backend:
+
+New Interface
+-------------
+
+``"soundfile"`` backend with new interface will become the default interface in ``0.8.0``.
```
mruberry, line 211: The soundfile...

torchaudio/backend/_soundfile_backend.py
```diff
@@ -210,6 +210,10 @@ def load_wav(
 ) -> Tuple[torch.Tensor, int]:
     """Load wave file.
 
+    Warnings:
+        This functions is being deprecated and will be removed in ``0.9.0``.
```
mruberry, line 214: "This function is deprecated and will be removed in the 0.9.0 release. Use load(..., normalize=False), instead."

torchaudio/backend/common.py
```diff
@@ -1,8 +1,13 @@
 from typing import Any, Optional
 
+from torchaudio._internal import module_utils as _mod_utils
+
 
 class AudioMetaData:
-    """Data class to be returned by :py:func:`~torchaudio.info`.
+    """Data class to be returned by ``torchaudio.info`` function.
```
mruberry, line 7: This is a little misleading because AudioMetaData isn't actually a Python dataclass, right? 

This could just say "Return type of torchaudio.info."

torchaudio/backend/common.py
```diff
@@ -72,6 +89,9 @@ def __init__(self,
 
 _LOAD_DOCSTRING = r"""Loads an audio file from disk into a tensor
 
+Warnings:
+    This function is deprecated. Please migrate to the new backend.
```
mruberry, line 93: "... is deprecated and will be removed in a future release."

"new backend" should be replaced with a more specific recommendation.

torchaudio/backend/common.py
```diff
@@ -124,6 +144,9 @@ def __init__(self,
 It assumes that the wav file uses 16 bit per sample that needs normalization by
 shifting the input right by 16 bits.
 
+Warnings:
+    This function is deprecated. Please migrate to the new backend and use ``load(.., normalize=False)``
```
mruberry, line 148: "... is deprecated and will be removed in a future release."

"New backend" should also be more specific here.

torchaudio/backend/common.py
```diff
@@ -135,6 +158,9 @@ def __init__(self,
 
 _SAVE_DOCSTRING = r"""Saves a Tensor on file as an audio file
 
+Warnings:
+    This function is deprecated. Please migrate to the new backend.
```
mruberry, line 162: Analogous comments here as above.

torchaudio/backend/sox_io_backend.py
```diff
@@ -14,11 +14,11 @@ def info(filepath: str) -> AudioMetaData:
 
     Args:
         filepath (str or pathlib.Path):
-        Path to audio file. This function also handles ``pathlib.Path`` objects, but is annotated as
-        ``str`` for TorchScript compiler compatibility.
+            Path to audio file. This function also handles ``pathlib.Path`` objects,
+            but is annotated as ``str`` for TorchScript compiler compatibility.
```
mruberry, line 18: "for Torchscript compatibility"

torchaudio/backend/sox_io_backend.py
```diff
@@ -14,11 +14,11 @@ def info(filepath: str) -> AudioMetaData:
...
 
     Returns:
-        AudioMetaData: meta data of the given audio.
+        AudioMetaData: Meta data of the given audio.
```
mruberry, line 21: "Metadata" (one word)

torchaudio/backend/sox_io_backend.py
```diff
@@ -184,6 +184,9 @@ def load_wav(
 ) -> Tuple[torch.Tensor, int]:
     """Load wave file.
 
+    Warnings:
+        This functions is being deprecated and will be removed in ``0.9.0``.
```
mruberry, line 188: "This function is deprecated..."

docs/source/backend.rst
vincentqb, line 17: it may be confusing to the user to see to backend with the same name. can you add "soundfile with some option = True" ?

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
 
-.. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
+There are currently four implementations available.
 
-.. note::
-   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
+    * ``"sox"`` (deprecated)
+    * ``"sox_io"``
```
vincentqb, line 15: move to first place, and change order of paragraph below.

docs/source/backend.rst
vincentqb, line 21: this will be easier to read if moved before "sox" ?

docs/source/backend.rst
vincentqb: > This is great! Want to make this part of an issue as well?

[previous response](https://github.com/pytorch/audio/pull/935#discussion_r499009228)

docs/source/backend.rst
vincentqb, line 133:
```suggestion
The function calls to this backend are TorchSript-able. You can apply :func:`torch.jit.script` and dump the object to a file, then call it from C++ application.
```

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
+* In ``0.7.0`` release, the legacy interface is used by default when switching to ``"soundfile"`` backend.
+* In ``0.8.0`` release, the new interface will become the default.
+* In ``0.9.0`` release, the legacy interface will be removed.
+
+To change the interface, set ``torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE`` flag **before** switching the backend.
+
+.. code::
+
+   torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = True
+   torchaudio.set_audio_backend("soundfile")  # The legacy interface
```
vincentqb, line 171:
```suggestion
   torchaudio.set_audio_backend("soundfile")  # uses legacy interface
```

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
+
+   torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False
+   torchaudio.set_audio_backend("soundfile")  # The new interface
```
vincentqb, line 174:
```suggestion
   torchaudio.set_audio_backend("soundfile")  # uses new interface
```

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
+
+``"soundfile"`` backend with legacy interface is currently the default backend if C++ is not available, however this interface is deprecated and will be removed in ``0.9.0`` release.
+
+To switch to this backend/interface, set ``torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE`` flag **before** switching the backend.
 
 .. code::
 
-   torchaudio.set_audio_backend("soundfile")
+   torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = True
+   torchaudio.set_audio_backend("soundfile")  # The legacy interface
```
vincentqb, line 186:
```suggestion
   torchaudio.set_audio_backend("soundfile")  # use legacy interface
```

docs/source/backend.rst
```diff
@@ -106,33 +150,87 @@ save
...
+-------------
+
+``"soundfile"`` backend with new interface will become the default interface in ``0.8.0``.
+
+To switch to this backend/interface, set ``torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE`` flag **before** switching the backend.
+
+.. code::
+
+   torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False
+   torchaudio.set_audio_backend("soundfile")  # The new interface
```
vincentqb, line 218:
```suggestion
   torchaudio.set_audio_backend("soundfile")  # use new interface
```

torchaudio/backend/common.py
```diff
@@ -149,6 +175,9 @@ def __init__(self,
 
 _INFO_DOCSTRING = r"""Gets metadata from an audio file without loading the signal.
 
+Warnings:
+    This function is deprecated. Please migrate to the new backend.
```
vincentqb, line 179: [same](https://github.com/pytorch/audio/pull/935/files#r500526562)

docs/source/backend.rst
mthrok: Okay, so the description here was basically duplications from description of each backend. So I removed the description here and made the list into links to the backends. That way, the list here will only give the necessary information (default/deprecated).

So the `Overall` section will talk about stuff things across backends.

docs/source/backend.rst
mthrok: no, the default one should come first.

docs/source/backend.rst
mthrok: This is the documentation, so the human-readable format is better here. also technically `option = True` is not enough because you can change the flag without switching the backend.

docs/source/backend.rst
```diff
@@ -3,44 +3,88 @@
...
 :mod:`torchaudio.backend` module provides implementations for audio file I/O, using different backend libraries.
 To switch backend, use :py:func:`torchaudio.set_audio_backend`. To check the current backend use :py:func:`torchaudio.get_audio_backend`.
 
-.. warning::
-   Although ``sox`` backend is default for backward compatibility reason, it has a number of issues, therefore it is highly recommended to use ``sox_io`` backend instead. Note, however, that due to the interface refinement, functions defined in ``sox`` backend and those defined in ``sox_io`` backend do not have the same signatures.
+There are currently four implementations available.
 
-.. note::
-   Instead of calling functions in :mod:`torchaudio.backend` directly, please use ``torchaudio.info``, ``torhcaudio.load``, ``torchaudio.load_wav`` and ``torchaudio.save`` with proper backend set with :func:`torchaudio.get_audio_backend`.
+    * ``"sox"`` (deprecated)
```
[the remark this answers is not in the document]
mthrok: adopted the suggestion

docs/source/backend.rst
mthrok: Addressed all the suggestions. I am hesitant to mention the new I/O work as it's very early phase and we do not know when it will be available.

torchaudio/backend/_soundfile_backend.py
```diff
@@ -210,6 +210,10 @@ def load_wav(
 ) -> Tuple[torch.Tensor, int]:
     """Load wave file.
 
+    Warnings:
```
cpuhrsch, line 213: I think this is unnecessary

torchaudio/backend/common.py
```diff
@@ -72,6 +89,10 @@ def __init__(self,
 
 _LOAD_DOCSTRING = r"""Loads an audio file from disk into a tensor
 
+Warnings:
```
cpuhrsch, line 92: I think this is unnecessary, because it's duplicated with the warning statement and because we don't do this in any other library.

torchaudio/backend/common.py
```diff
@@ -135,6 +161,10 @@ def __init__(self,
 
 _SAVE_DOCSTRING = r"""Saves a Tensor on file as an audio file
 
+Warnings:
```
cpuhrsch, line 164: I think this is unnecessary, because it's duplicated with the warning statement and because we don't do this in any other library.

torchaudio/backend/common.py
```diff
@@ -149,6 +179,10 @@ def __init__(self,
 
 _INFO_DOCSTRING = r"""Gets metadata from an audio file without loading the signal.
 
+Warnings:
```
cpuhrsch, line 182: I think this is unnecessary, because it's duplicated with the warning statement and because we don't do this in any other library.

torchaudio/backend/sox_io_backend.py
```diff
@@ -184,6 +184,9 @@ def load_wav(
 ) -> Tuple[torch.Tensor, int]:
     """Load wave file.
 
+    Warnings:
```
cpuhrsch, line 187: I think this is unnecessary, because it's duplicated with the warning statement and because we don't do this in any other library.

torchaudio/backend/sox_io_backend.py
```diff
@@ -14,11 +14,11 @@ def info(filepath: str) -> AudioMetaData:
 
     Args:
         filepath (str or pathlib.Path):
-        Path to audio file. This function also handles ``pathlib.Path`` objects, but is annotated as
-        ``str`` for TorchScript compiler compatibility.
+            Path to audio file. This function also handles ``pathlib.Path`` objects,
+            but is annotated as ``str`` for TorchScript compatibility.
```
vincentqb, line 18: I would remove the mention to pathlib.

cpuhrsch (approved):

torchaudio/backend/sox_io_backend.py
vincentqb, line 18: looks like pathlib is the new standard, so could extend to where it also supported elsewhere

vincentqb (approved):
LGTM, let's land this :)

PR #934, opened 2020-10-02
fix tedlium load_audio

vincentqb:
Fixes #928

PR #937, opened 2020-10-06
Fix Windows unit tests

peterjc123:
@mthrok @vincentqb


peterjc123:
@vincentqb Do you have any idea about the test failures?

mthrok:
@peterjc123 

Thanks for the fix. The test failures are being addressed in #934 .

mthrok (approved):

PR #939, opened 2020-10-06
Add pysoundfile to readme

vincentqb:
Add missing dependency to readme when using soundfile backend, hinted [here](https://github.com/pytorch/audio/pull/935/files#r500431382).


cpuhrsch:
This can be addressed by Moto's backend documentation in a central place.

PR #948, opened 2020-10-12
allow deprecation decorator to be used on a class

mattip:
Since `SoxEffectsChain` is a class, it needs a different decorator to emit a deprecation warning. This one will warn every time an instance is initialized (via `__init__`).

### DIFF
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -1,7 +1,7 @@
 import warnings
 import importlib.util
 from typing import Optional
-from functools import wraps
+from functools import wraps, WRAPPER_ASSIGNMENTS
 
 
 def is_module_available(*modules: str) -> bool:
@@ -43,14 +43,34 @@ def deprecated(direction: str, version: Optional[str] = None):
     Args:
         direction: Migration steps to be given to users.
     """
-    def decorator(func):
-        @wraps(func)
-        def wrapped(*args, **kwargs):
-            message = (
-                f'{func.__module__}.{func.__name__} has been deprecated '
-                f'and will be removed from {"future" if version is None else version} release. '
-                f'{direction}')
-            warnings.warn(message, stacklevel=2)
-            return func(*args, **kwargs)
+    def decorator(obj):
+        if isinstance(obj, type):
+            # Update the class to emit a warning when used
+            class Wrapper(obj):
+                def __init__(self, *args, **kwargs):
+                    message = (
+                        f'{obj.__module__}.{obj.__name__} has been deprecated '
+                        f'and will be removed from '
+                        f'{"future" if version is None else version} release. '
+                        f'{direction}')
+                    warnings.warn(message, stacklevel=1)
+                    self._wrapped = wrapped
+                    for attr in WRAPPER_ASSIGNMENTS:
+                        setattr(self, attr, getattr(wrapped, attr))
+                    super(self).__init__(*args, **kwargs)
+
+                def __repr__(self):
+                    return repr(self._wrapped)
+
+            return Wrapper
+        else:
+            @wraps(obj)
+            def wrapped(*args, **kwargs):
+                message = (
+                    f'{obj.__module__}.{obj.__name__} has been deprecated '
+                    f'and will be removed from {"future" if version is None else version} release. '
+                    f'{direction}')
+                warnings.warn(message, stacklevel=2)
+                return obj(*args, **kwargs)
         return wrapped
     return decorator


torchaudio/_internal/module_utils.py
```diff
@@ -43,14 +43,34 @@ def deprecated(direction: str, version: Optional[str] = None):
...
+            # Update the class to emit a warning when used
+            class Wrapper(obj):
+                def __init__(self, *args, **kwargs):
+                    message = (
+                        f'{obj.__module__}.{obj.__name__} has been deprecated '
+                        f'and will be removed from '
+                        f'{"future" if version is None else version} release. '
+                        f'{direction}')
+                    warnings.warn(message, stacklevel=1)
+                    self._wrapped = wrapped
```
mthrok, line 57: Looks like `wrapperd` is not defined yet. Could you fix this?

`NameError: free variable 'wrapped' referenced before assignment in enclosing scope`

You can check the error here; https://app.circleci.com/pipelines/github/pytorch/audio/3543/workflows/2b475e36-74ce-4a39-afe1-b316e234886d/jobs/104175

mthrok (commented):
Hi @mattip 

Thanks for the PR. I totally missed this, and apologize for flooding your training log.
Overall direction looks good but there is a fatal error in the current fix could you fix it?
Also this is very timely as we are working on the next release, I am so grateful that you brought up this.

Also here is a suggestion for improved readability/maintainability. Since the `deprecated` function is getting long, could you extract the main components into helper functions?

```python
def depricated(...):
    message = ...
    def decorator(obj):
        if isinstance(obj, type):
            return _decorate_class(obj, message)
        return _decorate_func(obj, message)

def _decorate_class(obj, message):
    ...

def _decorate_func(obj, message):
    ...
```

mthrok:
@mattip 

Can you provide a snippet that cause multiple warnings?
I just ran the following code and the warnings were issued only once.

```python
# foo.py
import torchaudio

for _ in range(10):
    sfx = torchaudio.sox_effects.SoxEffectsChain()
    eff = torchaudio.sox_effects.SoxEffect()
```

```
$ python foo.py
/scratch/moto/torchaudio/torchaudio/backend/utils.py:53: UserWarning: "sox" backend is being deprecated. The default backend will be changed to "sox_io" backend in 0.8.0 and "sox" backend will be removed in 0.9.0. Please migrate to "sox_io" backend. Please refer to https://github.com/pytorch/audio/issues/903 for the detail.
  warnings.warn(
foo.py:5: UserWarning: torchaudio.sox_effects.sox_effects.SoxEffectsChain has been deprecated and will be removed from future release. Please migrate to `apply_effects_file` or `apply_effects_tensor`.
  sfx = torchaudio.sox_effects.SoxEffectsChain()
foo.py:6: UserWarning: torchaudio.sox_effects.sox_effects.SoxEffect has been deprecated and will be removed from future release. Please migrate to `apply_effects_file` or `apply_effects_tensor`.
  eff = torchaudio.sox_effects.SoxEffect()
```

mattip:
`python -Walways foo.py`

mthrok:
So using `-Walways` seems to make warnings triggered all the time, regardless of the object wrapped being function or classes. Could you help me understand how your suggestion would make it different for classes?

For example, the following snippet issues the same warnings. I think it's no different for classes too.

```
# foo.py
import torchaudio

for _ in range(3):
    torchaudio.set_audio_backend('soundfile')
```

```
$ python -Walways foo.py
/scratch/moto/torchaudio/torchaudio/backend/utils.py:53: UserWarning: "sox" backend is being deprecated. The default backend will be changed to "sox_io" backend in 0.8.0 and "sox" backend will be removed in 0.9.0. Please migrate to "sox_io" backend. Please refer to https://github.com/pytorch/audio/issues/903 for the detail.
  warnings.warn(
/scratch/moto/torchaudio/torchaudio/backend/utils.py:63: UserWarning: The interface of "soundfile" backend is planned to change in 0.8.0 to match that of "sox_io" backend and the current interface will be removed in 0.9.0. To use the new interface, do `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False` before setting the backend to "soundfile". Please refer to https://github.com/pytorch/audio/issues/903 for the detail.
  warnings.warn(
/scratch/moto/torchaudio/torchaudio/backend/utils.py:63: UserWarning: The interface of "soundfile" backend is planned to change in 0.8.0 to match that of "sox_io" backend and the current interface will be removed in 0.9.0. To use the new interface, do `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False` before setting the backend to "soundfile". Please refer to https://github.com/pytorch/audio/issues/903 for the detail.
  warnings.warn(
/scratch/moto/torchaudio/torchaudio/backend/utils.py:63: UserWarning: The interface of "soundfile" backend is planned to change in 0.8.0 to match that of "sox_io" backend and the current interface will be removed in 0.9.0. To use the new interface, do `torchaudio.USE_SOUNDFILE_LEGACY_INTERFACE = False` before setting the backend to "soundfile". Please refer to https://github.com/pytorch/audio/issues/903 for the detail.
  warnings.warn(
```

mattip:
I discovered the problem when building docs. The object returned before this PR was converted from a class to a function, so building documentation for it failed. There is still a problem with `SoxEffect` it is [documented as a class](https://github.com/pytorch/audio/blame/master/docs/source/sox_effects.rst#L45) but it is a function. Since @brianjo is working on a PR to improved documentation, I commented there.

I think there are subtle implications of wrapping a class and returning a function, maybe it breaks typing?

mattip:
I guess my statement in the top PR comment is confusing
> This one will warn every time an instance is initialized (via `__init__`).

It was meant as a statement of fact, not as a judgement call right/wrong. I think it is the right thing to do. It is like the warning emitted in the deprecated functions: emitted each time the function is called and filtered by default to only show to the user once.

mthrok:
Oh I see. It's about the documentation. If the decorator is causing the trouble, we can just use plain `warnings.warn` call inside of `__init__`.

For `SoxEffect` issue, I think it is intended. The real structure is defined inside of C++ and bound via PyBind11, so the `SoxEffect` is a kind of surrogate (not sure if factory function is more appropriate term here). The thing is, though, that structure is not necessary or any use/value from external user perspective. It's only consumed in `SoxEffectChain`. I think it's okay to remove it from documentation too, if we do not have an alternative.

mattip:
> we can just use plain warnings.warn call inside of `__init__`

Indeed, that would be simpler than this PR

mthrok:
> Indeed, that would be simpler than this PR

Would you proceed in that way? or let me know how I can assist you.

mattip:
OK. Closing this, thanks for the review.

PR #955, opened 2020-10-13
Wav2letter clean

vincentqb:
Extends pytorch#632
Closes vincentqb/audio#2


examples/pipeline_wav2letter/datasets.py
```diff
@@ -1,7 +1,61 @@
+import itertools
+from typing import List
+
 import torch
+from torch import Tensor
 from torchaudio.datasets import LIBRISPEECH
 
 
+def pad_sequence(sequences, padding_value=0.0):
```
vincentqb, line 9: note to self: how fast this runs against something like this?
```
def pad_sequence(batch):
    # Make all tensor in a batch the same length by padding with zeros
    batch = [item.t() for item in batch]
    batch = torch.nn.utils.rnn.pad_sequence(batch, batch_first=True, padding_value=0.)
    return batch.permute(0, 2, 1)
```

PR #956, opened 2020-10-14
move deprecation from class to class.__init__

mattip:
Adding the deprecation decorator turns a class into a function. Move it to the `__init__` method instead. This requires specifying the class name, otherwise the warning prints the `__init__` method. 

Because `@wraps` messes with the local namespace, hide the `name` on a "private" attribute.

### DIFF
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -37,17 +37,26 @@ def requires_module(*modules: str):
     return decorator
 
 
-def deprecated(direction: str, version: Optional[str] = None):
+def deprecated(direction: str,
+               version: Optional[str] = None,
+               name: Optional[str] = None,
+              ):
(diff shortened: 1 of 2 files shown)


mattip:
replaces gh-948

mthrok:
Thanks for the update. Instead of modifying the `deprecated` decorator, can we use plain `warnings.warn` function call inside of `SoxEffectChain.__init__()` ?

The reason is that `SoxEffectChain` and `SoxEffect` are going to be removed from master branch after the 0.7.0 release, which I plan to do in a couple of weeks, so it's not worth modifying the `deprecated` decorator for that purpose.

mattip:
> SoxEffectChain and SoxEffect are going to be removed from master branch 

OK, cool. It's fine to say this is not needed, it was meant to be a little tweak and grew to be too complicated. I will close this.

PR #959, opened 2020-10-15
Make removal version explicit for sox deprecations

mthrok:
Make the sox deprecations explicit about the future version, at which they are removed.

@mattip I was working on updating deprecation decorators and took the opportunity to solve the issue you were facing. Not sure how doc works are organized, but when we build versioned docs and if we are building it from `release/0.7`, then it will be reflected.

### DIFF
--- a/torchaudio/sox_effects/sox_effects.py
+++ b/torchaudio/sox_effects/sox_effects.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import torch
@@ -275,7 +276,6 @@ def SoxEffect():
     return _torchaudio.SoxEffect()
 
 
-@_mod_utils.deprecated('Please migrate to `apply_effects_file` or `apply_effects_tensor`.', "0.8.0")
 class SoxEffectsChain(object):
     r"""SoX effects chain class.
 
@@ -339,6 +339,11 @@ class SoxEffectsChain(object):
                  out_siginfo: Any = None,
                  out_encinfo: Any = None,
                  filetype: str = "raw") -> None:
+        warnings.warn(
+            'torchaudio.sox_effects.SoxEffectsChain has been deprecated and '
+            'will be removed from 0.8.0 release. '
+            'Please migrate to `apply_effects_file` or `apply_effects_tensor`.'
+        )
         self.input_file: Optional[str] = None
         self.chain: List[str] = []
         self.MAX_EFFECT_OPTS = 20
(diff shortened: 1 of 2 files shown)


torchaudio/sox_effects/sox_effects.py
```diff
@@ -339,6 +339,11 @@ def __init__(self,
                  out_siginfo: Any = None,
                  out_encinfo: Any = None,
                  filetype: str = "raw") -> None:
+        warnings.warn(
+            'torchaudio.sox_effects.SoxEffectsChain has been deprecated and '
+            'will be removed from 0.8.0 release. '
+            'Please migrate to `apply_effects_file` or `apply_effects_tensor`.'
+        )
```
mattip, line 346: maybe make it filterable?
```suggestion
            'Please migrate to `apply_effects_file` or `apply_effects_tensor`.',
            DeprecationWarning
        )
```

mthrok: Thanks for suggestion. Although I agree with you and that is better, but somehow domain libraries have rules to use `UserWarning` for deprecation.

mattip: Ok, sorry for the noise (again)

mthrok: No worries, I appreciate your feedback. :)
I think we should sit together to sync up on your documentation work soon.

mattip: OK. Maybe we can bet @brianjo involved.

vincentqb: > Thanks for suggestion. Although I agree with you and that is better, but somehow domain libraries have rules to use `UserWarning` for deprecation.

This is also standard with pytorch. The issue with DeprecationWarning is that it is hidden by default.

vincentqb (approved):
This LGTM as is. Any reason why it's marked as draft?

torchaudio/sox_effects/sox_effects.py
```diff
@@ -260,7 +261,7 @@ def apply_effects_file(
 
 
 @_mod_utils.requires_module('torchaudio._torchaudio')
-@_mod_utils.deprecated('Please migrate to `apply_effects_file` or `apply_effects_tensor`.')
+@_mod_utils.deprecated('Please migrate to `apply_effects_file` or `apply_effects_tensor`.', "0.8.0")
 def SoxEffect():
     r"""Create an object for passing sox effect information between python and c++
```
vincentqb, line 266: nit: no need to change the docstring though we did discuss removing warnings from the docstring to avoid duplicated signal

Issue #950, opened 2020-10-13
Add pathlib.Path support all io


vincentqb:
Since pathlib is the standard python library for handling paths since python 3.4, let's add explicit support for pathlib in i/o operations. 

- [x] `sox_io`
- [ ] ~~`sox` (won't update)~~
- [x] "new" `soundfile` 
- [ ] ~~legacy `sox` (won't update)~~
- [ ] ~~`apply_effects_file`, [here](https://github.com/pytorch/audio/blob/master/torchaudio/sox_effects/sox_effects.py#L163), jit does not support union, see [comment](https://github.com/pytorch/audio/pull/1048#issuecomment-730450782)~~

There are also datasets who consume paths as string in their constructor, which could be extended to pathlib. (This means adding a conversion at the beginning of each function to convert path to string so the rest can proceed as expected.)

- [x] cmuarctic
- [x] commonvoice
- [x] gtzan
- [x] librispeech
- [x] libritts
- [x] ljspeech
- [x] speechcommands
- [x] tedlium
- [x] vctk
- [x] yesno
- [ ] ~~see dataset utils consuming a path [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/utils.py)~~

cc offline discussion with @mthrok and @cpuhrsch about #907

mthrok:
Can you provide the road map for this?

Note that `"sox_io"` backend and the new interface of `"soundfile"` backend support `path lib.Path` already, so no need to change the implementation.

Since we are getting rid of `"sox"` backend and the legacy interface of `"soundfile"` backend, I would not put them on a high priority, rather I would leave them as is.

vincentqb:
> Note that `"sox_io"` backend and the new interface of `"soundfile"` backend support `path lib.Path` already, so no need to change the implementation.
> 
> Since we are getting rid of `"sox"` backend and the legacy interface of `"soundfile"` backend, I would not put them on a high priority, rather I would leave them as is.

Agree.

> Can you share the list of action items for this?

The datasets, and `apply_effects_file` are the other places were files are being used. That's all I can think of. Some/All may already support, we should just double check :) Any others?

mthrok:
Can you clarify on what **actions** are needed for each component?

`apply_effects_file` makes sense and I was not aware so thanks for listing up, however, it is not immediately clear to me why dataset implementations are on the list. They are not responsible for I/O, nor they receive path from client code.

vincentqb:
[...]

Issue #852, opened 2020-08-04
Re-align GTZAN with other datasets


vincentqb:
The update done in #791 removed `walk_files` from GTZAN, and added pattern recognition. We'll need to re-align GTZAN with the other dataset.

~I suggest we simply add a filter after `walk_files` instead of updating `walk_files`, as done in [SPEECHCOMMANDS](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/speechcommands.py#L74).~ As suggested by @mthrok below, let's instead iterate through the folder structure.

mmxgn:
Hi,

That PR addressed the comment on #764 that with `walk_files` we could introduce extra files.

```
2. `GTZAN` dataset does not have the list of files that corresponds to the entire dataset. Since `GTZAN` has a fixed number of samples, it should not be traversing all the audio files under the given directory. This could introduce extra files or does not react to missing files.
   @mmxgn Do you have a thought on 2.?
```

I could still change it to use `walk_files` and do the pattern recognition with `filter` if that's ok?

vincentqb:
> I could still change it to use `walk_files` and do the pattern recognition with `filter` if that's ok?

That would be great, thanks!

mthrok:
I do not think it's a good idea to keep using `walk_files`. The utility is solving the problem that Dataset does not have. Once the root directory of each dataset is located, then there is no need to do directory traversal recursively. It's better to have a new utility function that does exactly what is needed by Dataset requirements, that is iterate the audio files in the list of subdirectories in dataset.

The gist is;

```
for subdir in subdirs:
    for audio_file in subdir:
        yield audio_file
```

Keep using `walk_files` forces Dataset classes to define the problem in loose manner (unnecessary recursion).

vincentqb:
I'm ok with your suggestion. I updated the description to reflect it. That being said, I'd leave this issue open until all the datasets use the same structure :)

mthrok:
An approach better than `walk_files` is to delegate the responsibility of finding directories where audio files are found to Dataset, and the responsibility to traverse audio files in the directories in fixed order to a utility function.

vincentqb:
> An approach better than `walk_files` is to delegate the responsibility of finding directories where audio files are found to Dataset, and the responsibility to traverse audio files in the directories in fixed order to a utility function. That way, it is clear who is responsible for what.

This is similar to having a deterministic `walk_files` to detect audio files and a filter to preserve only files matching a certain pattern, no?

[...]

PR #966, opened 2020-10-16
Add SpeechCommands train/valid/test split

vincentqb:
Add SpeechCommands train/valid/test split. This follows the convention established in [gtzan](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/gtzan.py#L1014).

### DIFF
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -112,8 +112,8 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
         assert num_samples == len(self.samples)
 
-    def testSpeechCommandsSplit(self):
-        dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, split="testing")
+    def testSpeechCommandsSubset(self):
+        dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="testing")
         print(dataset._path)
 
         num_samples = 0
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -107,7 +107,22 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
     def testSpeechCommands(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir)
-        print(dataset._path)
+
+        num_samples = 0
+        for i, (data, sample_rate, label, speaker_id, utterance_number) in enumerate(
+            dataset
+        ):
+            self.assertEqual(data, self.samples[i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.samples[i][1]
+            assert label == self.samples[i][2]
+            assert speaker_id == self.samples[i][3]
+            assert utterance_number == self.samples[i][4]
+            num_samples += 1
+
+        assert num_samples == len(self.samples)
+
+    def testSpeechCommandsNone(self):
+        dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset=None)
 
         num_samples = 0
         for i, (data, sample_rate, label, speaker_id, utterance_number) in enumerate(
@@ -169,3 +184,11 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
             num_samples += 1
 
         assert num_samples == len(self.test_samples)
+
+    def testSpeechCommandsSum(self):
+        dataset_all = speechcommands.SPEECHCOMMANDS(self.root_dir)
+        dataset_train = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="training")
+        dataset_valid = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="validation")
+        dataset_test = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="testing")
+
+        assert len(dataset_train) + len(dataset_valid) + len(dataset_test) == len(dataset_all)
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -96,15 +96,14 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
                             utterance,
                         )
                         cls.samples.append(sample)
-                        label_filename = os.path.join(label, filename)
-                        if 2 <= j < 4:
-                            valid.write(f'{label_filename}\n')
-                            cls.valid_samples.append(sample)
-                        elif 4 <= j < 6:
-                            test.write(f'{label_filename}\n')
-                            cls.test_samples.append(sample)
-                        else:
+                        if j < 2:
                             cls.train_samples.append(sample)
+                        elif j < 4:
+                            valid.write(f'{label}/{filename}\n')
+                            cls.valid_samples.append(sample)
+                        elif j < 6:
+                            test.write(f'{label}/{filename}\n')
+                            cls.test_samples.append(sample)
 
     def testSpeechCommands(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir)
@@ -141,7 +140,6 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
     def testSpeechCommandsSubsetValid(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="validation")
-        print(dataset._path)
 
         num_samples = 0
         for i, (data, sample_rate, label, speaker_id, utterance_number) in enumerate(
@@ -156,9 +154,8 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
         assert num_samples == len(self.valid_samples)
 
-    def testSpeechCommandsSubset(self):
+    def testSpeechCommandsSubsetTest(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="testing")
-        print(dataset._path)
 
         num_samples = 0
         for i, (data, sample_rate, label, speaker_id, utterance_number) in enumerate(
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -53,11 +53,12 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
     root_dir = None
     samples = []
+    train_samples = []
     valid_samples = []
     test_samples = []
 
     @classmethod
-    def setUp(cls):
+    def setUpClass(cls):
         cls.root_dir = cls.get_base_temp_dir()
         dataset_dir = os.path.join(
             cls.root_dir, speechcommands.FOLDER_IN_ARCHIVE, speechcommands.URL
@@ -102,6 +103,8 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
                         elif 4 <= j < 6:
                             test.write(f'{label_filename}\n')
                             cls.test_samples.append(sample)
+                        else:
+                            cls.train_samples.append(sample)
 
     def testSpeechCommands(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir)
@@ -119,6 +122,22 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
             num_samples += 1
 
         assert num_samples == len(self.samples)
+
+    def testSpeechCommandsSubsetTrain(self):
+        dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="training")
+
+        num_samples = 0
+        for i, (data, sample_rate, label, speaker_id, utterance_number) in enumerate(
+            dataset
+        ):
+            self.assertEqual(data, self.train_samples[i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate == self.train_samples[i][1]
+            assert label == self.train_samples[i][2]
+            assert speaker_id == self.train_samples[i][3]
+            assert utterance_number == self.train_samples[i][4]
+            num_samples += 1
+
+        assert num_samples == len(self.train_samples)
 
     def testSpeechCommandsSubsetValid(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="validation")
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -53,6 +53,8 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
     root_dir = None
     samples = []
+    valid_samples = []
+    test_samples = []
 
     @classmethod
     def setUp(cls):
@@ -63,12 +65,13 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
         os.makedirs(dataset_dir, exist_ok=True)
         sample_rate = 16000  # 16kHz sample rate
         seed = 0
-        txt_file = os.path.join(dataset_dir, "testing_list.txt")
-        with open(txt_file, "w") as txt:
+        valid_file = os.path.join(dataset_dir, "validation_list.txt")
+        test_file = os.path.join(dataset_dir, "testing_list.txt")
+        with open(valid_file, "w") as valid, open(test_file, "w") as test:
             for label in LABELS:
                 path = os.path.join(dataset_dir, label)
                 os.makedirs(path, exist_ok=True)
-                for j in range(2):
+                for j in range(6):
                     # generate hash ID for speaker
                     speaker = "{:08x}".format(j)
 
@@ -93,7 +96,12 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
                         )
                         cls.samples.append(sample)
                         label_filename = os.path.join(label, filename)
-                        txt.write(f'{label_filename}\n')
+                        if 2 <= j < 4:
+                            valid.write(f'{label_filename}\n')
+                            cls.valid_samples.append(sample)
+                        elif 4 <= j < 6:
+                            test.write(f'{label_filename}\n')
+                            cls.test_samples.append(sample)
 
     def testSpeechCommands(self):
         dataset = speechcommands.SPEECHCOMMANDS(self.root_dir)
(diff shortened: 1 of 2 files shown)


vincentqb:
Merging this before release would simplify the [tutorial]()https://colab.research.google.com/drive/1yIiDkm4ixjNr6dg6oowwfxeDPf1NlXHP?authuser=1#scrollTo=aGauAPh_l9A).

test/torchaudio_unittest/datasets/speechcommands_test.py
```diff
@@ -53,43 +53,58 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
 
     root_dir = None
     samples = []
+    train_samples = []
+    valid_samples = []
+    test_samples = []
 
     @classmethod
-    def setUp(cls):
+    def setUpClass(cls):
```
vincentqb, line 61: setUp is ran at every method call, and so setUpClass should be used instead as is done for the other dataset.

cpuhrsch (approved):

torchaudio/datasets/speechcommands.py
```diff
@@ -81,6 +92,26 @@ def __init__(self,
...
+        if subset == "validation":
+            walker = validation_list
+        elif subset == "testing":
+            walker = testing_list
+        elif subset == "training":
+            walker = filter(
+                lambda w: not (w in validation_list or w in testing_list), walker
+            )
+
         self._walker = list(walker)
```
mthrok, line 115: `list` is no longer necessary and should be removed.

torchaudio/datasets/speechcommands.py
```diff
@@ -81,6 +92,26 @@ def __init__(self,
...
+
+        if subset in ["training", "validation"]:
+            filepath = os.path.join(self._path, "validation_list.txt")
+            with open(filepath) as f:
+                validation_list = [os.path.join(self._path, l.strip()) for l in f.readlines()]
+
+        if subset in ["training", "testing"]:
+            filepath = os.path.join(self._path, "testing_list.txt")
+            with open(filepath) as f:
+                testing_list = [os.path.join(self._path, l.strip()) for l in f.readlines()]
```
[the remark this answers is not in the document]
mthrok: Also note that the call to `readlines` here is redundant. You can simply to `[os.path.join(self._path, l.strip()) for l in f]`

torchaudio/datasets/speechcommands.py
```diff
@@ -48,13 +48,24 @@ class SPEECHCOMMANDS(Dataset):
             The top-level directory of the dataset. (default: ``"SpeechCommands"``)
         download (bool, optional):
             Whether to download the dataset if it is not found at root path. (default: ``False``).
+        subset (Optional[str]):
+            Select a subset of the dataset [None, "training", "validation", "testing"].
+            (default: ``None``)
```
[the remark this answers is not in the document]
vincentqb: As you point out, `training_list.txt` doesn't exist, it's defined to be what's not in validation or testing. The other two are defined by the dataset. Is that what you meant?

torchaudio/datasets/speechcommands.py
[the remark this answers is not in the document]
vincentqb: sure, I'm ok with your suggestion

torchaudio/datasets/speechcommands.py
```diff
@@ -81,6 +92,26 @@ def __init__(self,
 
         walker = walk_files(self._path, suffix=".wav", prefix=True)
```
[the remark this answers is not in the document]
vincentqb: Fair point

torchaudio/datasets/speechcommands.py
```diff
@@ -81,6 +92,26 @@ def __init__(self,
...
         walker = filter(lambda w: HASH_DIVIDER in w and EXCEPT_FOLDER not in w, walker)
```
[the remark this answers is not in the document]
vincentqb: That's technically outside the scope of this PR, but I've updated it anyway since we are moving away from generators.

torchaudio/datasets/speechcommands.py
vincentqb: same

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +92,22 @@ def __init__(self,
...
+                return [os.path.join(self._path, l.strip()) for l in f]
+
+        if subset == "validation":
+            self._walker = load_list(VALIDATION_LIST)
+        elif subset == "testing":
+            self._walker = load_list(TESTING_LIST)
+        elif subset == "training":
+            excludes = load_list(VALIDATION_LIST) + load_list(TESTING_LIST)
+            walker = walk_files(self._path, suffix=".wav", prefix=True)
+            self._walker = [w for w in walker if HASH_DIVIDER in w and EXCEPT_FOLDER not in w and w not in excludes]
```
mthrok, line 107: Looking way better than the previous. One more push.

Rather than parsing the all the files present in the file system first then filter only the interested, it's better to only parse the needed one.

Also, repeating the combination of `walk_files` and filtering `HASH_DIVIDER` in two clauses looks overwhelming. You can combine them into a much shorter code like the bellow.

```python
elif subset == "testing":
    ...
else:
    paths = [str(p) for p in Path(self._path).glob('*/*_no_hash_*.wav')]
        # no need to check EXCEPT_FOLDER because
        # none of the wav file in the directory match with `_no_hash_`
    if subset == "training":
        excludes = set(_load_lists("validation_list.txt", "testing_list.txt"))
        paths = [p for p in paths if p not in excludes]
    self._walker = paths

File: torchaudio/datasets/speechcommands.py (hunk 2 above)
Comment (vincentqb) [reply torchaudio/datasets/speechcommands.py]: With the change mentioned [above](https://github.com/pytorch/audio/pull/966/#discussion_r512198533), this is indeed no longer needed. Removed.

File: torchaudio/datasets/speechcommands.py [hunk 8]
```diff
@@ -79,9 +92,22 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum, hash_type="md5")
                 extract_archive(archive, self._path)
 
-        walker = walk_files(self._path, suffix=".wav", prefix=True)
-        walker = filter(lambda w: HASH_DIVIDER in w and EXCEPT_FOLDER not in w, walker)
-        self._walker = list(walker)
+        def load_list(filename):
+            filepath = os.path.join(self._path, filename)
+            with open(filepath) as fileobj:
+                return [os.path.join(self._path, line.strip()) for line in fileobj]
```
mthrok, line 98: Please prefer a plain function defined on module level especially when there is no local variable needs to be captured.

Also, since you are defining a helper function for your own sake, you can adopt a signature like

```
def _load_lists(*filename)
```

and it will make `excludes = load_lists(VALIDATION_LIST) + load_list(TESTING_LIST)` into as simple as `excludes = _load_lists("validation_list.txt", "testing_list.txt")`. (continue to the next comment)

mthrok (commented):
Definitely better than the previous code. Made some comments to push it to the next level.

test/torchaudio_unittest/datasets/speechcommands_test.py
```diff
@@ -53,43 +53,59 @@ class TestSpeechCommands(TempDirMixin, TorchaudioTestCase):
...
+                        )
+                        cls.samples.append(sample)
+                        if j < 2:
+                            cls.train_samples.append(sample)
+                        elif j < 4:
+                            label_filename = os.path.join(label, filename)
+                            valid.write(f'{label_filename}\n')
+                            cls.valid_samples.append(sample)
+                        elif j < 6:
+                            label_filename = os.path.join(label, filename)
```
mthrok, line 106: The contents are of `"testing_list.txt"` and `"validation_list.txt"` are in Linux format. 

```
$ head speech_commands_v0.01/testing_list.txt
bed/0c40e715_nohash_0.wav
bed/0ea0e2f4_nohash_0.wav
bed/0ea0e2f4_nohash_1.wav
```

Here, using `os.path.join` make it to the valid path on Windows system (something like `bed\0c40e715_nohash_0.wav`), so this is not correct. Something like the following is the correct way to mimic the data.

```suggestion
                            label_filename = f'{label}/{filename}'
```

Same goes to the case where `j < 4`

vincentqb: good catch

mthrok: I think it's more straightforward to use `List[pathlib.Path]` objects for `self._walker`, rather than converting them to `List[str]`.

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +92,22 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum, hash_type="md5")
                 extract_archive(archive, self._path)
 
-        walker = walk_files(self._path, suffix=".wav", prefix=True)
-        walker = filter(lambda w: HASH_DIVIDER in w and EXCEPT_FOLDER not in w, walker)
-        self._walker = list(walker)
+        def load_list(filename):
+            filepath = os.path.join(self._path, filename)
+            with open(filepath) as f:
+                return [os.path.join(self._path, l.strip()) for l in f]
```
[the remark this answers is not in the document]
vincentqb: > Please prefer a plain function defined on module level especially when there is no local variable needs to be captured.

`load_list` uses `self._path` in its current form :)

> Also, since you are defining a helper function for your own sake, you can adopt a signature like
> 
> ```python
> def load_lists(*filename)
> ```
> 
> and it will make `excludes = load_list(VALIDATION_LIST) + load_list(TESTING_LIST)` into as simple as `excludes = load_list("validation_list.txt", "testing_list.txt")`. (continue to the next comment)

Either way looks reasonable to me, but I've made the update since you have a preference :)

torchaudio/datasets/speechcommands.py
vincentqb: I agree that parsing and then filtering is not optimal. Changing `os_walk` is out of scope for this PR though, and could change the order of the files from #814. I'm ok with replacing `os_walk` with `glob` in a next PR due to #950.

torchaudio/datasets/speechcommands.py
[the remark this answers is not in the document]
mthrok: > `load_list` uses `self._path` in its current form :)

Oops my bad.

torchaudio/datasets/speechcommands.py
mthrok: I encourage you to take this opportunity ro get rid of `walk_files` because, if you try to stick with `walk_files`, I do not see any pro at a glance. You will have to more works to settle with sub-optimal solution, which is not as compact yet inefficient.

Also #950 is about the support of `Path` object, so I do not see the logical connection between it and `walk_files` (which returns list of string).

Regarding, the order it returns, we have a unit test which can verify the proper ordering of files returned, so you do not need to worry about aligning the order with `walk_files` function. Instead, you can just focus on writing the correct code. However, this is a valid point and probably you will need to insert `sort` function in the snippet above.

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +90,25 @@ def __init__(self,
...
 
-        walker = walk_files(self._path, suffix=".wav", prefix=True)
-        walker = filter(lambda w: HASH_DIVIDER in w and EXCEPT_FOLDER not in w, walker)
-        self._walker = list(walker)
+        def _load_list(*filenames):
+            output = []
+            for filename in filenames:
+                filepath = os.path.join(self._path, filename)
+                with open(filepath) as fileobj:
+                    output += [os.path.join(self._path, line.strip()) for line in fileobj]
```
mthrok, line 98: I'm surprised our Windows CI job does not raise FileNotFound error with this implementation. If my understanding is correct, as previously noted, `os.path.join(self._path, line.strip())` would produce strings that look like `speech_commands\speech_commands_v0.01/testing_list.txt`, which is an invalid path on Windows.

test/torchaudio_unittest/datasets/speechcommands_test.py
```diff
@@ -107,3 +121,51 @@ def testSpeechCommands(self):
             num_samples += 1
 
         assert num_samples == len(self.samples)
+
+    def testSpeechCommandsSubsetTrain(self):
+        dataset = speechcommands.SPEECHCOMMANDS(self.root_dir, subset="training")
```
mthrok, line 126: Can you add a test case for `subset=None`?

torchaudio/datasets/speechcommands.py
vincentqb: Yeah, that's what I see too. It seems like `os.path.normpath` would be useful here.

test/torchaudio_unittest/datasets/speechcommands_test.py
vincentqb: Yeah, that will be more convenient to do that in the future

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +90,30 @@ def __init__(self,
...
+            self._walker = _load_list("validation_list.txt")
+        elif subset == "testing":
+            self._walker = _load_list("testing_list.txt")
+        elif subset == "training":
+            excludes = set(_load_list("validation_list.txt", "testing_list.txt"))
+            walker = walk_files(self._path, suffix=".wav", prefix=True)
+            self._walker = [
+                w for w in walker
+                if HASH_DIVIDER in w
+                and EXCEPT_FOLDER not in w
```
mthrok, line 111: `if HASH_DIVIDER in w and EXCEPT_FOLDER not in w` is redundant because files in background noises do not have `_nohash_` in filenames.

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +90,30 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum, hash_type="md5")
                 extract_archive(archive, self._path)
 
-        walker = walk_files(self._path, suffix=".wav", prefix=True)
-        walker = filter(lambda w: HASH_DIVIDER in w and EXCEPT_FOLDER not in w, walker)
-        self._walker = list(walker)
+        def _load_list(*filenames):
```
mthrok, line 93: Can you move this to outside for the purpose of readability?

something like `def _load_list(root_dir, *filenames)`

torchaudio/datasets/speechcommands.py
```diff
@@ -79,9 +90,30 @@ def __init__(self,
...
+                filepath = os.path.join(self._path, filename)
+                with open(filepath) as fileobj:
+                    output += [os.path.normpath(os.path.join(self._path, line.strip())) for line in fileobj]
+            return output
+
+        if subset == "validation":
+            self._walker = _load_list("validation_list.txt")
+        elif subset == "testing":
+            self._walker = _load_list("testing_list.txt")
+        elif subset == "training":
```
mthrok, line 105: I see there is a behavior inconsistency between `"training"/None` and `"validation"/"testing"`.

If certain valid files are removed from the dataset, then this dataset implementation will keep working for `"training"/None` case with less files, but for `"validation"/"testing"` it will raise an exception.

What's your take on that?

torchaudio/datasets/speechcommands.py
vincentqb: I'll avoid changing the logic here to avoid breaking any codes, say if a user moved a file in background_noise folder/etc. I can look into it for a later PR.

torchaudio/datasets/speechcommands.py
vincentqb: "testing", "validation", "training" are defined by the dataset by the training/validation files. It is technically undefined by the dataset outside of that. Given how the three are defined by the dataset in those files, as a user, I'd expect changes to those file to propagate. I'd say it'd be fair to add a quick note in the docstring.

vincentqb: Added.

mthrok: > It is technically undefined by the dataset outside of that.

I agree with that, however IIRC, one of the goal of torchaudio's dataset implementation is to make it easy to modify the dataset. Something along the line of point 3&4 of https://github.com/pytorch/audio/issues/852#issuecomment-670590773 . With this rule, we have to take the extra step to think through what kind of modification is valid/invalid and what is the expected behavior, and put it in the implementation.

I am trying to raise the awareness of it.

mthrok: And of course, if we are just getting rid of the easy modification of the dataset, we do no need think about the any modification to dataset.

torchaudio/datasets/speechcommands.py
mthrok: thanks this looks much readable now

torchaudio/datasets/speechcommands.py
```diff
@@ -22,6 +22,15 @@
 }
 
 
+def _load_list(root, *filenames):
+    output = []
+    for filename in filenames:
+        filepath = os.path.join(root, filename)
+        with open(filepath) as fileobj:
+            output += [os.path.normpath(os.path.join(root, line.strip())) for line in fileobj]
```
mthrok, line 30: Can you add a comment on why `normapth` is necessary here?
Even though [`normpath`'s official document](https://docs.python.org/3/library/os.path.html#os.path.normpath) says `On Windows, it converts forward slashes to backward slashes. To normalize case, use normcase().`. For many of us, who are linux users, It is not well known. (I learned the only today.)

torchaudio/datasets/speechcommands.py
vincentqb: Actually, SpeechCommands does explain how those two files are generated, and how to generalize their approach, see [README](https://github.com/sankalp2610/Speech_Command_Recognition).

vincentqb:
Merging based on offline discussion

torchaudio/datasets/speechcommands.py
mthrok: @vincentqb Now that #950 is closed, can you followup replacing `walk_files` with `glob` with something like the following?

> I agree that parsing and then filtering is not optimal. Changing `os_walk` is out of scope for this PR though, and could change the order of the files from #814. I'm ok with replacing `os_walk` with `glob` in a next PR due to #950.

> ```python
> elif subset == "testing":
>     ...
> else:
>     paths = [str(p) for p in Path(self._path).glob('*/*_no_hash_*.wav')]
>         # no need to check EXCEPT_FOLDER because
>         # none of the wav file in the directory match with `_no_hash_`
>     if subset == "training":
>         excludes = set(_load_lists("validation_list.txt", "testing_list.txt"))
>         paths = [p for p in paths if p not in excludes]
>     self._walker = paths
> ```

vincentqb: (about #940: #950 discussed using the Path library which arrived in 3.4. now that we use it, i'm ok with using other functions from it.)

vincentqb: good idea: i've opened an issue to request help from the community to remove `walk_files`: #1051.

PR #970, opened 2020-10-19
Skip failing tests on macOS Python 3.6/3.7

mthrok:
Some tests are failing on very specific environments and we believe it's related to PyTorch/Python internal. See https://github.com/pytorch/pytorch/issues/46409 for the detail. For the time being, we disable the corresponding test.


cpuhrsch (approved):

mthrok:
thanks

PR #972, opened 2020-10-20
Fix doc rendering

mthrok:
Currently there is no version specification for `sphinx` in docs and `pip install -r requirements.txt` would install the version that produces the corrupted rendering for function signatures. This issue is fixed by using `2.4.4`, which is [the same version as PyTorch](https://github.com/pytorch/pytorch/blob/master/docs/requirements.txt#L1)

*Note*: This commit should also go to master branch.

Before

Screen Shot 2020-10-20 at 16 13 47

After

Screen Shot 2020-10-20 at 16 00 45

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx
+sphinx==2.4.4
 -e git+git://github.com/pytorch/pytorch_sphinx_theme.git#egg=pytorch_sphinx_theme
 sphinxcontrib.katex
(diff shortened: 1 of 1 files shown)


astaff:
LGTM :-)

vincentqb (approved):
LGTM, thx for catching this.

vincentqb:
As you pointed out offline, this might need to go in master too, though we'll need to keep track of the pinned version later.

PR #987, opened 2020-10-27
Fix CI unittest job setup

mthrok:
This PR resolves macOS unit test jobs then align the Windows script for the consistency.

The job is fixed by changing the order of installations to
1. Create env
1. Install build tools
1. (Cache the env)
1. Install PyTorch
1. Install torchaudio
1. Install test tools

Once this is merged, we can cherry-pick #970 to skip the tests that are stuck.


vincentqb (approved):
LGTM and tests have passed. Any reason why marked as draft?

I agree with cherry picking #970 after this.

PR #306, opened 2019-10-09
Nightly build [DO NOT DELETE OR MERGE]

ezyang:
See pytorch/pytorch#26817 for what this does

CircleCI: https://app.circleci.com/pipelines/github/pytorch/audio?branch=nightly
FB only: Chronos job page: https://fburl.com/chronos/3ft16107 (click "Run Now" to manually trigger a nightly)

PR #1006, opened 2020-11-04
build docs and push to branch in CI

mattip:
This creates a job to build docs and push them. The basic idea is to reuse the install part of the ~unittest~ smoke_test jobs, and then do
```
pushd doc
pip install -r requriements.txt
build html
popd
```

Then a separate step commits the built documentation, checks out the gh-pages branch, copies the rendered documentation into the correct directory, commits it, and pushes the result to the branch.

Things I don't understand about the config.yml, I hope I did it right
~- how do the unittest_linux_cpu jobs know not to run if the build job fails? There is no `requires` clause~
- git push hopefully will JustWork when a PR is merged, although it will fail when done on a fork (like when people submit a PR). Should I limit the job to only run on nightly to avoid the problem?

### DIFF
--- a/.circleci/build_docs/commit_docs.sh
+++ b/.circleci/build_docs/commit_docs.sh
@@ -3,20 +3,25 @@
 set -ex
 
 
-if [ "$1" == "" ]; then
-    echo call as $0 "<target branch>", where branch should be "master" or "1.7" or so
+if [ "$2" == "" ]; then
+    echo call as $0 "<src>" "<target branch>"
+    echo where src is the built documentation and
+    echo branch should be "master" or "1.7" or so
     exit 1
 fi
 
+scr=$1
+target=$2
+
 set -ex
-echo "committing docs to ${target}"
+echo "committing docs from ${stc} to ${target}"
 
 git checkout gh-pages
 rm -rf docs/$target/*
-cp -r doc/build/html/* docs/$target
+cp -r ${src}/build/html/* docs/$target
 if [ $target == "master" ]; then
     rm -rf docs/_static/*
-    cp -r doc/build/html/_static/* docs/_static
+    cp -r ${src}/build/html/_static/* docs/_static
 fi
 git add docs || true
 git config user.email "soumith+bot@pytorch.org"
--- a/.circleci/build_docs/commit_docs.sh
+++ b/.circleci/build_docs/commit_docs.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -ex
+
+
+if [ "$1" == "" ]; then
+    echo call as $0 "<target branch>", where branch should be "master" or "1.7" or so
+    exit 1
+fi
+
+set -ex
+echo "committing docs to ${target}"
+
+git checkout gh-pages
+rm -rf docs/$target/*
+cp -r doc/build/html/* docs/$target
+if [ $target == "master" ]; then
+    rm -rf docs/_static/*
+    cp -r doc/build/html/_static/* docs/_static
+fi
+git add docs || true
+git config user.email "soumith+bot@pytorch.org"
+git config user.name "pytorchbot"
+# If there aren't changes, don't make a commit; push is no-op
+git commit -m "auto-generating sphinx docs" || true
+git push -u origin gh-pages
+
+
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -650,59 +650,27 @@ jobs:
             .circleci/torchscript_bc_test/setup_master_envs.sh
             .circleci/torchscript_bc_test/validate_objects.sh
   build_docs:
-    <<: *binary_common
-    docker:
-      - image: "pytorch/torchaudio_unittest_base:manylinux"
-    resource_class: medium
+    <<: *smoke_test_common
     steps:
-    - checkout
-    - attach_workspace:
-        at: third_party
-    - designate_upload_channel
-    - generate_cache_key
-    - restore_cache:
-
-        keys:
-          - env-v3-linux-{{ arch }}-py<< parameters.python_version >>-{{ checksum ".circleci/unittest/linux/scripts/setup_env.sh" }}-{{ checksum "third_party/CMakeLists.txt" }}-{{ checksum ".cachekey" }}
-
-    - run:
-        name: Setup
-        command: .circleci/unittest/linux/scripts/setup_env.sh
-    - save_cache:
-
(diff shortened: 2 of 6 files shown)


mattip:
ping @brianjo

mattip:
hmm, I wanted to use the built wheel which should be in `./dist` but instead seem to be building the project. @seemethere any ideas what I am doing wrong?

mthrok:
> hmm, I wanted to use the built wheel which should be in `./dist` but instead seem to be building the project. @seemethere any ideas what I am doing wrong?

The script used in the preparation steps (`.circleci/unittest/linux/scripts/install.sh` and `.circleci/unittest/linux/scripts/setup_env.sh`) are intended for installing torchaudio from source.

If the binary package is preferable, you can set up things like `smoke_test` do, which depends on the binary build process.

https://github.com/pytorch/audio/blob/b7c17f80b8f8f5cd25a167dc1589ada4c5b0c7cc/.circleci/config.yml.in#L275-L292

mthrok:
> * how do the unittest_linux_cpu jobs know not to run if the build job fails? There is no `requires` clause

`unittest` workflow and `binary` workflow are completely independent. This is because `binary` workflow can fail due to external reasons like network issue, but most often we still want to run unit tests for the PR we are working on.

> * git push hopefully will JustWork when a PR is merged, although it will fail when done on a fork (like when people submit a PR). Should I limit the job to only run on nightly to avoid the problem?

@seemethere Does the CI job executor have a push privilege?

mthrok (approved):
Looks good to me, although I might be missing some detail. Let's see how things work out on nightly build.

.circleci/build_docs/commit_docs.sh
```diff
@@ -0,0 +1,28 @@
...
+if [ $target == "master" ]; then
+    rm -rf docs/_static/*
+    cp -r doc/build/html/_static/* docs/_static
+fi
+git add docs || true
+git config user.email "soumith+bot@pytorch.org"
+git config user.name "pytorchbot"
+# If there aren't changes, don't make a commit; push is no-op
+git commit -m "auto-generating sphinx docs" || true
+git push -u origin gh-pages
```
seemethere, line 26: We should separate out the building and the pushing of docs into two separate jobs so that we can test that docs build without having to push them

seemethere (changes requested):
Looks mostly good to me, just one thing

.circleci/build_docs/commit_docs.sh
mattip: @seemethere What distinguishes between the two jobs? How would you trigger `build_docs` alone but always trigger the `push_docs` job on a merge or a tag?

mthrok: I gave some thoughts on this.

With the current way, `build_doc` happens on the `nightly` branch.
The `nightly` branch is managed by cronjob and the jobs are triggered by the PR https://github.com/pytorch/audio/pull/306.
With this approach, the hosted documentation will be updated once a day. I think this is fine.

But at the same time, it will be also nice if normal PR jobs include doc build process too, so that we can check how the documentation will be changed. (so the job builds the documentation html and uploads them to somewhere in CCI's artifact, which is browsable)

One solution I think of is to remove the `only` clause from the configuration YAML file (so the doc build process runs on all the PR and commits) then do the branch checking manually in the bash script.

Something like

```
if [ $CIRCLE_BRANCH = "master" ]; then
    push
fi
```

The advantage of this approach is it's easy to add release branch.

@seemethere @mattip What do you think?

mattip: Building the docs on every PR seems like a good thing, then we could check for sphinx warnings and stop the merge until the PR fixes docs. As it stands the `build_docs` job depends on the `nightly_binary_linux_conda_py3.8`. What is the name of the python 3.8 build job on every PR that should be required instead, so the docs do not have to rebuild the package as well?

mattip: I guess it would be `binary_linux_wheel_py3.8`

mthrok: > I guess it would be `binary_linux_wheel_py3.8`

Yes, that's correct.

seemethere (approved):

mattip:
@seemethere `build_docs` is failing, it is not "continuing" from where the `binary_linux_wheel_3.8` is finishing. What do I need to preserve to not have to rebuild from 0?

.circleci/config.yml
```diff
@@ -649,6 +649,43 @@ jobs:
...
             .circleci/torchscript_bc_test/validate_objects.sh
+  build_docs:
+    <<: *binary_common
+    docker:
+      - image: "pytorch/manylinux-cuda100"
+    resource_class: 2xlarge+
+    steps:
+      - attach_workspace:
+          at: ~/workspace
+      - run:
```
mthrok, line 660: @mattip Looks like `- checkout` step is missing, so the repository is not fetched, thus `packaging` directory is not found.

Also for the completeness, can you add `- designate_upload_channel` too?

mattip:
OK, CI passes. There are [warnings when building the docs](https://app.circleci.com/pipelines/github/pytorch/audio/3836/workflows/d98cf3cb-b352-46be-b336-2039d3c369a3/jobs/114816), but that is not new: brianjo/audio#1.

mthrok:
Looks good. Let's see how the build process goes. Thanks!

.circleci/config.yml
```diff
@@ -734,6 +774,22 @@ workflows:
...
+          python_version: '3.8'
+          requires:
+          - binary_linux_wheel_py3.8
+      - upload_docs:
+          filters:
+            branches:
+              only:
+              - nightly
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*-rc[0-9]+/
```
mthrok, line 788: Sorry for making comment after the merge, but this means that release documentation is only pushed on release tag, right?
How about changing this to `release/*` branch so that even after the release, we can fix the documentation by pushing doc change to `release/*` branch?
@seemethere @brianjo @mattip

mattip: The way I see it is like this:
- gh-pages/master is built every night and documents the latest and greatest HEAD of the development
- gh-pages/0.8 and all the numbered directories document the latest release; what end users would download to their machines if they install from conda or pip and specified a version.
- stable points to the latest numbered directory, what users would get if they conda or pip install and do not specify a version.

Bugs in the release would indeed be committed to the release/* branch via PRs, these would trigger the `build_docs` build and you could examine the artifacts\*. But the new documentation would only become "official" when the release is tagged and the fix is available to the general public as a major/minor/patch release, so a tag event is when the contents of the numbered directory should change.

You could fix bugs in the documentation that are discovered after release by building documentation locally and pushing the built documentation to the correct directory (using the steps in `.circleci/build_docs/commit_docs.sh` or calling the script to do it for you).

\* I should improve the CI process and make the artifacts available for viewing after the `build_docs` step.

PR #1009, opened 2020-11-05
[D] Update spectrogram to use complex

mthrok:
This PR is for getting signal on complex tensor adaptation, similar to #864

The internal changes should go into #1013, as it resolves the issue with backward pass of the Spectrogram. Then this PR will be about adding `return_complex` arguments for the migration.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -42,7 +42,8 @@ def spectrogram(
         normalized: bool,
         center: bool = True,
         pad_mode: str = "reflect",
-        onesided: bool = True
+        onesided: bool = True,
+        return_complex: bool = False,
 ) -> Tensor:
     r"""Create a spectrogram or a batch of spectrograms from a raw audio signal.
     The spectrogram can be either magnitude-only or complex.
@@ -65,12 +66,22 @@ def spectrogram(
             :attr:`center` is ``True``. Default: ``"reflect"``
         onesided (bool, optional): controls whether to return half of results to
             avoid redundancy. Default: ``True``
+        return_complex (bool, optional):
+            ``return_complex = True``, this function returns the resulting Tensor in
+            complex dtype, otherwise it returns the resulting Tensor in real dtype with extra
+            dimension for real and imaginary parts. (see ``torch.view_as_real``).
+            When ``power`` is provided, the value must be False, as the resulting
+            Tensor represents real-valued power.
 
     Returns:
         Tensor: Dimension (..., freq, time), freq is
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is not None and return_complex:
+        raise ValueError(
+            'When `power` is provided, the return value is real-valued. '
+            'Therefore, `return_complex` must be False.')
 
     if pad > 0:
         # TODO add "with torch.no_grad():" back when JIT supports it
@@ -103,7 +114,9 @@ def spectrogram(
         if power == 1.0:
             return spec_f.abs()
         return spec_f.abs().pow(power)
-    return torch.view_as_real(spec_f)
+    if not return_complex:
+        return torch.view_as_real(spec_f)
+    return spec_f
 
 
 def griffinlim(
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -54,6 +54,12 @@ class Spectrogram(torch.nn.Module):
             :attr:`center` is ``True``. Default: ``"reflect"``
         onesided (bool, optional): controls whether to return half of results to
             avoid redundancy Default: ``True``
+        return_complex (bool, optional):
+            ``return_complex = True``, this function returns the resulting Tensor in
+            complex dtype, otherwise it returns the resulting Tensor in real dtype with extra
+            dimension for real and imaginary parts. (see ``torch.view_as_real``).
+            When ``power`` is provided, the value must be False, as the resulting
+            Tensor represents real-valued power.
     """
     __constants__ = ['n_fft', 'win_length', 'hop_length', 'pad', 'power', 'normalized']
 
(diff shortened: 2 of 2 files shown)


torchaudio/functional.py
```diff
@@ -75,6 +76,8 @@ def spectrogram(
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is not None and return_complex:
+        raise ValueError('When `power` is provided, `return_complex` must be False.')
```
anjali411, line 80: why is that?

torchaudio/functional.py
```diff
@@ -86,16 +89,21 @@ def spectrogram(
...
     # unpack batch
-    spec_f = spec_f.reshape(shape[:-1] + spec_f.shape[-3:])
+    spec_f = spec_f.reshape(shape[:-1] + spec_f.shape[-2:])
 
     if normalized:
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
-        spec_f = complex_norm(spec_f, power=power)
+        spec_f = spec_f.abs().pow(power)
+    elif not return_complex:
```
anjali411, line 105: shouldn't it be a separate if statement?

torchaudio/functional.py
mthrok: In the original implementation, when the `power` argument is provided, the returned signal changes from pseudo-complex Tensor to plain real Tensor. (as the spectrogram is passed to `complex_norm` function which is equivalent of `abs` of complex type tensor).

We could still return complex type with all the imaginary parts equal zero, but I thought there is not much point of that. I am open to changing it though.

torchaudio/functional.py
mthrok: As I mentioned in the previous reply, when `power` is provided, this function returns the power of spectrogram, which is real value. So `return_complex` only applies when `power` is not applied (the function is returning complex-valued tensor).

mthrok:
@anjali411 

Do you have a suggestion for the tests to be added with respect to the complex dtype?

torchaudio/functional.py
anjali411: `torch.abs` returns a floating point tensor though so I don't see any reason why we should change it

anjali411: However for performance reasons we should add a special case for power 1 (not call pow), and pow 2 (check if torch.view_as_real(x).pow(2).sum(-1) is performance wise better that `x.abs().pow(2)`).

mthrok: Okay, thanks.

> However for performance reasons we should add a special case for power 1 (not call pow), and pow 2 (check if torch.view_as_real(x).pow(2).sum(-1) is performance wise better that `x.abs().pow(2)`).

I will look into this.

mthrok: I got the following result, `torch.view_as_real(x).pow(2).sum(-1)` is slightly faster on CUDA.

| Method                                            | CPU                                   | GPU                              |
|---------------------------------|------------------------------------------|------------------------------------------|
| `x.abs().pow(2)`                                          | 100 loops, best of 5: 3.17 msec per loop | 100 loops, best of 5: 48.4 usec per loop |
| `torch.view_as_real(x).pow(2).sum(-1)` | 100 loops, best of 5: 28.8 msec per loop | 100 loops, best of 5: 41.7 usec per loop |

<details> <summary>script</summary>

```
OMP_NUM_THREADS=1 numactl --membind 0 --cpubind 0 python -m timeit -n 100 -r 5 -s """
import torch;
import torch.fft;
x = torch.zeros(1, 32*44100, dtype=torch.cfloat)
""" """
x.abs().pow(2);
"""

OMP_NUM_THREADS=1 numactl --membind 0 --cpubind 0 python -m timeit -n 100 -r 5 -s """
import torch;
import torch.fft;
x = torch.zeros(1, 32*44100, dtype=torch.cfloat)
""" """
torch.view_as_real(x).pow(2).sum(-1)
"""

OMP_NUM_THREADS=1 numactl --membind 0 --cpubind 0 python -m timeit -n 100 -r 5 -s """
import torch;
import torch.fft;
x = torch.zeros(1, 32*44100, dtype=torch.cfloat, device='cuda')
""" """
x.abs().pow(2);
"""

OMP_NUM_THREADS=1 numactl --membind 0 --cpubind 0 python -m timeit -n 100 -r 5 -s """
import torch;
import torch.fft;
x = torch.zeros(1, 32*44100, dtype=torch.cfloat, device='cuda')
""" """
torch.view_as_real(x).pow(2).sum(-1)
"""
```

</details>

mthrok: @anjali411 Updated the code based on the result.

torchaudio/functional.py
```diff
@@ -75,6 +76,8 @@ def spectrogram(
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is not None and return_complex:
```
anjali411, line 79: based on the discussion below, we should get rid of this

mthrok: Okay. So the suggestion is that if power is provided, we return real-type tensor regardless of the `return_complex` value?

torchaudio/functional/functional.py
```diff
@@ -91,17 +94,30 @@ def spectrogram(
...
     if power is not None:
-        spec_f = complex_norm(spec_f, power=power)
-
+        if power == 1.0:
+            return spec_f.abs()
+        if power == 2.0 and spec_f.is_cuda:
+            return torch.view_as_real(spec_f).pow(power).sum(-1)
+        return spec_f.abs().pow(power)
+    if not return_complex:
+        return torch.view_as_real(spec_f)
```
vincentqb, line 120: [same](https://github.com/pytorch/audio/pull/1013/files#r520215197)

PR #1012, opened 2020-11-06
SpeechCommands: fix for the test set

yan12125:
In the test set of SpeechCommands v2 [1], some files have the form
"xxx.wav.wav".

[1] https://storage.googleapis.com/download.tensorflow.org/data/speech_commands_test_set_v0.02.tar.gz

### DIFF
--- a/torchaudio/datasets/speechcommands.py
+++ b/torchaudio/datasets/speechcommands.py
@@ -34,7 +34,15 @@ def _load_list(root, *filenames):
 def load_speechcommands_item(filepath: str, path: str) -> Tuple[Tensor, int, str, str, int]:
     relpath = os.path.relpath(filepath, path)
     label, filename = os.path.split(relpath)
+    # Besides the officially supported split method for datasets defined by "validation_list.txt"
+    # and "testing_list.txt" over "speech_commands_v0.0x.tar.gz" archives, an alternative split
+    # method referred to in paragraph 2-3 of Section 7.1, references 13 and 14 of the original
+    # paper, and the checksums file from the tensorflow_datasets package [1] is also supported.
+    # Some filenames in those "speech_commands_test_set_v0.0x.tar.gz" archives have the form
+    # "xxx.wav.wav", so file extensions twice needs to be stripped twice.
+    # [1] https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/url_checksums/speech_commands.txt
     speaker, _ = os.path.splitext(filename)
+    speaker, _ = os.path.splitext(speaker)
 
     speaker_id, utterance_number = speaker.split(HASH_DIVIDER)
     utterance_number = int(utterance_number)
@@ -60,7 +68,10 @@ class SPEECHCOMMANDS(Dataset):
         subset (Optional[str]):
             Select a subset of the dataset [None, "training", "validation", "testing"]. None means
             the whole dataset. "validation" and "testing" are defined in "validation_list.txt" and
-            "testing_list.txt", respectively, and "training" is the rest. (default: ``None``)
+            "testing_list.txt", respectively, and "training" is the rest. Details for the files
+            "validation_list.txt" and "testing_list.txt" are explained in the README of the dataset
+            and in the introduction of Section 7 of the original paper and its reference 12. The
+            original paper can be found here: https://arxiv.org/pdf/1804.03209.pdf (default: ``None``)
     """
 
     def __init__(self,
--- a/torchaudio/datasets/speechcommands.py
+++ b/torchaudio/datasets/speechcommands.py
@@ -71,7 +71,7 @@ class SPEECHCOMMANDS(Dataset):
             "testing_list.txt", respectively, and "training" is the rest. Details for the files
             "validation_list.txt" and "testing_list.txt" are explained in the README of the dataset
             and in the introduction of Section 7 of the original paper and its reference 12. The
-            original paper can be found here: https://arxiv.org/pdf/1804.03209.pdf (default: ``None``)
+            original paper can be found `here <https://arxiv.org/pdf/1804.03209.pdf>`_. (Default: ``None``)
     """
 
     def __init__(self,


yan12125:
As a side note, TensorFlow has embedded the checksum of the archive in problem at https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/url_checksums/speech_commands.txt. I don't think Google will fix it.

mthrok:
Hi @yan12125 

Thanks for bringing this up. My observations
 - `speech_commands_test_set_v0.02.tar.gz` has files with `.wav.wav` suffix
 - though, this archive is technically not the one intended by the implementation. 
 - yet this is very relevant
 - the change seems not harmful.

We will discuss this change within the team (and the treatment of `speech_commands_test_set_v0.02.tar.gz`) and get back to you.

cpuhrsch:
Two follow-up questions

According to the [tensorflow documentation of SpeechCommands] tehre are 4,890 test entries. As a first order of investigation, do we match the number of entries mentioned in this documentation?

Secondly, is there documentation to be found on why there is this separate test archive?

yan12125:
> do we match the number of entries mentioned in this documentation?

Hmm, Google's dataset is somewhat confusing. `testing_list.txt` in http://download.tensorflow.org/data/speech_commands_v0.02.tar.gz contains 11005 lines, while `speech_commands_test_set_v0.02.tar.gz` has 4890 `*.wav` files. Also, the `tensorflow_datasets` package uses the "test" archive with 4890 files as the test split [1]. Seems the latter is the actual test set.

> is there documentation to be found on why there is this separate test archive?

I didn't find an official claim. But I guess it's related to differences between the "test" archive and `testing_list.txt` in the archive currently used by torchaudio.

[1] https://github.com/tensorflow/datasets/blob/v4.1.0/tensorflow_datasets/audio/speech_commands.py#L86-L111

vincentqb:
Mmmm, let me summarize the information we have:
- 4890 entries listed in [tf doc](https://www.tensorflow.org/datasets/catalog/speech_commands) for v0.0.2, which matches the length available through tensorflow (see [colab](https://colab.research.google.com/drive/1fq17mPHS1v7o11fJ3BqAZfW8XYSwil79?usp=sharing)).
- 2824 `**/*_nohash_*.wav` files in `speech_commands_test_set_v0.01`
- 4482 `**/*_nohash_*.wav` files in `speech_commands_test_set_v0.02`
- Section 7 of [original paper](https://arxiv.org/pdf/1804.03209.pdf) explicitly refers to `testing_lists.txt`.
   - `testing_list.txt` in v0.0.1 lists 6835 entries (out of 906248 `*_nohash_*.wav` files), which matches the length of `torchaudio.datasets.SPEECHCOMMANDS("./", url="speech_commands_v0.01", download=True, split="test")`.
   - `testing_list.txt` in v0.0.2 lists 11005 entries (out of 1481774 `*_nohash_*.wav` files), which matches the length of `torchaudio.datasets.SPEECHCOMMANDS("./", url="speech_commands_v0.02", download=True, split="test")`.
- The README for v0.0.1 ([here](https://github.com/sankalp2610/Speech_Command_Recognition), and v0.0.2) has an algorithm for generating the testing list.

vincentqb:
> UPDATE: another fact - in the referenced paper [2], the author Pete Warden also uses the "test" archive for experiments in Section 7.

you mean `testing_list.txt`, is that correct?

yan12125:
> 4482 `**/*_nohash_*.wav` files in `speech_commands_test_set_v0.02`

There are 408 `_silence_` wav files in `speech_commands_test_set_v0.02`. Apparently 4890 includes both 4482 `_nohash_` and 408 `_silence_` wav files.

> you mean `testing_list.txt`, is that correct?

Well, I was referring paragraph 2 of Section 7.1 in https://arxiv.org/pdf/1804.03209.pdf.

> I’ve uploaded a standard set of test files[13] to makeit easier to reproduce this metric.

And reference item 13 is the link http://download.tensorflow.org/data/speech_commands_test_set_v0.02.tar.gz. On the other hand, `testing_list.txt` for speech commands v2 is only available from http://download.tensorflow.org/data/speech_commands_v0.02.tar.gz

yan12125:
So, the "test set" split by torchaudio is not the actual test set? If so, I will try to change more codes so that torchaudio splits the correct test set.

vincentqb:
To be clear torchaudio is correct in the sense that it follows precisely:
- what is described in the [original paper](https://arxiv.org/pdf/1804.03209.pdf) which recommends `validation_list.txt` and `testing_list.txt` in the introduction of Section 7
- what is described in the readme (which also recommends `validation_list.txt` and `testing_list.txt`) and the algorithm described therein. This algorithm is also presented in reference [12] of the original paper. I confirm that using this algorithm `which_set(filename, 10, 10)` does generate the two files.

```python
import re
import hashlib
import torchaudio


MAX_NUM_WAVS_PER_CLASS = 2**27 - 1  # ~134M

def which_set(filename, validation_percentage, testing_percentage):
  """Determines which data partition the file should belong to.

  We want to keep files in the same training, validation, or testing sets even
  if new ones are added over time. This makes it less likely that testing
  samples will accidentally be reused in training when long runs are restarted
  for example. To keep this stability, a hash of the filename is taken and used
  to determine which set it should belong to. This determination only depends on
  the name and the set proportions, so it won't change as other files are added.

  It's also useful to associate particular files as related (for example words
  spoken by the same person), so anything after '_nohash_' in a filename is
  ignored for set determination. This ensures that 'bobby_nohash_0.wav' and
  'bobby_nohash_1.wav' are always in the same set, for example.

  Args:
    filename: File path of the data sample.
    validation_percentage: How much of the data set to use for validation.
    testing_percentage: How much of the data set to use for testing.

  Returns:
    String, one of 'training', 'validation', or 'testing'.
  """
  base_name = os.path.basename(filename)
  # We want to ignore anything after '_nohash_' in the file name when
  # deciding which set to put a wav in, so the data set creator has a way of
  # grouping wavs that are close variations of each other.
  hash_name = re.sub(r'_nohash_.*$', '', base_name)
  # This looks a bit magical, but we need to decide whether this file should
  # go into the training, testing, or validation sets, and we want to keep
  # existing files in the same set even if more files are subsequently
  # added.
  # To do that, we need a stable way of deciding based on just the file name
  # itself, so we do a hash of that and then use that to generate a
  # probability value that we use to assign it.
  hash_name_hashed = hashlib.sha1(hash_name.encode('utf-8')).hexdigest()
  percentage_hash = ((int(hash_name_hashed, 16) %
                      (MAX_NUM_WAVS_PER_CLASS + 1)) *
                     (100.0 / MAX_NUM_WAVS_PER_CLASS))
  if percentage_hash < validation_percentage:
    result = 'validation'
  elif percentage_hash < (testing_percentage + validation_percentage):
    result = 'testing'
  else:
    result = 'training'
  return result


class SubsetSC(torchaudio.datasets.SPEECHCOMMANDS):
    def __init__(self, subset: str = None):
        super().__init__("./", download=True)

        def load_list(filename):
            filepath = os.path.join(self._path, filename)
            with open(filepath) as fileobj:
                return [os.path.join(self._path, line.strip()) for line in fileobj]

        if subset == "validation":
            self._walker = load_list("validation_list.txt")
        elif subset == "testing":
            self._walker = load_list("testing_list.txt")
        elif subset == "training":
            excludes = load_list("validation_list.txt") + load_list("testing_list.txt")
            excludes = set(excludes)
            self._walker = [w for w in self._walker if w not in excludes]

            
# Create training and testing split of the data. We do not use validation in this tutorial.
all_set = SPEECHCOMMANDS("./", download=True)
train_set = SubsetSC("training")
test_set = SubsetSC("testing")
valid_set = SubsetSC("validation")

valid_list = set(filename for filename in all_set._walker if which_set(filename, 10, 10) == "validation")
test_list = set(filename for filename in all_set._walker if which_set(filename, 10, 10) == "testing")

# Check that the three splits add up to the whole dataset
print(len(train_set) + len(valid_set) + len(test_set) == len(all_set))

# Check that validation_list.txt and testing_list.txt and the corresponding lists generated by which_set give the same set
print(valid_list == set(valid_set._walker))
print(test_list == set(test_set._walker))
```

Let's tentatively call this Split Method 1.

The reason there is some confusion is that, as you point out, there is a second split (a.k.a Split Method 2) referred to in paragraph 2-3 of Section 7.1 and references 13 and 14 of the original paper. This is also what is referred to in the [checksums](https://github.com/tensorflow/datasets/blob/master/tensorflow_datasets/url_checksums/speech_commands.txt), and what appears to be used in tensorflow.

Split Method 1 and 2 are different, but both are documented in the original paper. To help clarify this to the user, I recommend we had the following to the docstring:

> The subset parameter uses the training/validation/testing split described by the files ``validation_list.txt`` and ``testing_list.txt`` which are explained in the README of the dataset and in Section 7 of the original paper and its reference 12. The original paper can be found here: https://arxiv.org/pdf/1804.03209.pdf

Going further, we could decide to offer an option to chose which Split Method to use. Thoughts?

vincentqb (changes requested):
Following the discussion above, can you show an example to clarify what this fix is addressing? I do not find files with `.wav.wav`.

```
❯ ls **/*.wav.wav
ls: **/*.wav.wav: No such file or directory
❯ cat testing_list.txt validation_list.txt|grep '.wav.wav'
❯ 
```

yan12125:
Thank you very much for pointing out the original `which_set` function! Yep those two distinct split methods in the same paper is quite confusing.

> To help clarify this to the user, I recommend we add the following to the docstring:

Thanks that looks clear.

> we could decide to offer an option to chose which Split Method to use.

I vote for also supporting Split Method 2 in torchaudio for various reasons, two of which are consistency with the "test set" generated by `tensorflow_datasets` as in the Colab notebook mentioned above (https://github.com/pytorch/audio/pull/1012#issuecomment-724995542), and less amount of data to download (108MB for `speech_commands_test_set_v0.02.tar.gz` versus 2.3GB for `speech_commands_v0.02.tar.gz`).

> Following the discussion above, can you show an example to clarify what this fix is addressing?

My fix is for Split Method 2. Files in question are extracted from `speech_commands_test_set_v0.02.tar.gz`:
```
$ ls **/*.wav.wav | head
speech_commands_test_set_v0.02/_unknown_/backward_0c540988_nohash_0.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_0cb74144_nohash_2.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_5170b77f_nohash_4.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_87070229_nohash_2.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_8fe67225_nohash_4.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_a6f2fd71_nohash_3.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_bfd26d6b_nohash_2.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_f2e59fea_nohash_4.wav.wav
speech_commands_test_set_v0.02/_unknown_/backward_f9643d42_nohash_3.wav.wav
speech_commands_test_set_v0.02/_unknown_/bed_0ea0e2f4_nohash_1.wav.wav

$ ls **/*.wav.wav | wc -l
408
```

yan12125:
Sorry for the delay. I add some explanations to the docstring and comments. Hopefully it is clear enough now.

vincentqb (approved):
This looks good to me! I've just turned the URL into a link before merging. 

Also, thanks a lot for bringing up that discussion and for updating the documentation :) I agree that it would be a good idea to expose split method 2 in some way.

yan12125:
Thanks a lot for kind reviews and great suggestions!

PR #1013, opened 2020-11-06
Fix nan gradient by using native complex abs op

### LINKED ISSUE #993: torchaudio.transforms.Spectrogram with power=1 can easily NaN
## 🐛 Bug

If you use STFT through Spectrogram, with power=1, you can easily get a NaN. If your spectrogram square magnitude underflows to zero, you end up taking backprop of sqrt(0), which leads to a NaN. 

## To Reproduce

1. Use torchaudio.transforms.Spectrogram, set power=1, do a backward pass.

## Expected behavior

There should be a configurable epsilon to clamp the input to sqrt, otherwise users have to use torch.stft directly or use other workarounds (such as power=2).

mthrok:
Resolves #993

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -1,6 +1,7 @@
 """Test defintion common to CPU and CUDA"""
 import torch
 import torchaudio.functional as F
+from parameterized import parameterized
 
 from torchaudio_unittest import common_utils
 
@@ -29,3 +30,25 @@ class Lfilter(common_utils.TestBaseMixin):
         assert output_signal.max() <= 1
         output_signal = F.lfilter(input_signal, a_coeffs, b_coeffs, clamp=False)
         assert output_signal.max() > 1
+
+
+class Spectrogram(common_utils.TestBaseMixin):
+    @parameterized.expand([(0., ), (1., ), (2., ), (3., )])
+    def test_grad_at_zero(self, power):
+        """The gradient of power spectrogram should not be nan but zero near x=0
+
+        https://github.com/pytorch/audio/issues/993
+        """
+        x = torch.zeros(1, 22050, requires_grad=True)
+        spec = F.spectrogram(
+            x,
+            pad=0,
+            window=None,
+            n_fft=2048,
+            hop_length=None,
+            win_length=None,
+            power=power,
+            normalized=False,
+        )
+        spec.sum().backward()
+        assert not x.grad.isnan().sum()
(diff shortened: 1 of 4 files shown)


test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -29,3 +29,24 @@ def test_clamp(self):
...
+        """
+        x = torch.zeros(1, 22050, requires_grad=True)
+        spec = F.spectrogram(
+            x,
+            pad=0,
+            window=None,
+            n_fft=2048,
+            hop_length=None,
+            win_length=None,
+            power=1,
```
mthrok, line 48: @anjali411 Should I also ensure this when power is other than 1?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -29,3 +29,24 @@ def test_clamp(self):
         assert output_signal.max() <= 1
         output_signal = F.lfilter(input_signal, a_coeffs, b_coeffs, clamp=False)
         assert output_signal.max() > 1
+
+
+class Spectrogram(common_utils.TestBaseMixin):
+    def test_grad_at_zero(self):
+        """The gradient of power spectrogram should not be non but zero near x=0
```
gibiansky, line 36: typo 'non->nan'

mthrok: thanks. fixed.

torchaudio/functional/functional.py
```diff
@@ -91,18 +91,30 @@ def spectrogram(
...
     if normalized:
         spec_f /= window.pow(2.).sum().sqrt()
     if power is not None:
-        spec_f = complex_norm(spec_f, power=power)
-
-    return spec_f
+        if power == 1.0:
+            return spec_f.abs()
+        if power == 2.0 and spec_f.is_cuda:
+            return torch.view_as_real(spec_f).pow(power).sum(-1)
```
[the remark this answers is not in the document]
vincentqb: I'm assuming [this](https://github.com/pytorch/audio/pull/1009#discussion_r518965072) is the reason. It's not clear the gain is significant. I would add a comment here.

torchaudio/functional/functional.py
```diff
@@ -91,18 +91,30 @@ def spectrogram(
...
+        return spec_f.abs().pow(power)
+    return torch.view_as_real(spec_f)
```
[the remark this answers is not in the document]
anjali411: yeah I agree we should move this logic to `complex_norm`

anjali411 (approved):

vincentqb (approved):
LGTM, but I would move the logic to compute the complex norm out of spectrogram, and move to the `complex_norm` functional. This would potentially benefit other locations, and keep the semantic cleaner.

PR #1015, opened 2020-11-07
YesNo Dataset Pathlib change

bhargavkathivarapu:
Relates to #950 .

Conversion added in the `__init__` function to convert Path object to string.

For adding a unit test should I refactor [existing one](https://github.com/pytorch/audio/blob/5e54c770b41bbdb7b228fe511b364f3f2aa96a88/test/torchaudio_unittest/datasets/yesno_test.py#L39) like below 
```python

def test_yesno_util(dataset):
   ...  comparison code

def test_yesno_str():
  dataset = yesno.YESNO(self.root_dir)
  test_yesno_util(dataset)


def test_yesno_path():
  dataset = yesno.YESNO(Path(self.root_dir))
  test_yesno_util(dataset)

```
@vincentqb could you review these changes

### DIFF
--- a/test/torchaudio_unittest/datasets/yesno_test.py
+++ b/test/torchaudio_unittest/datasets/yesno_test.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 from torchaudio.datasets import yesno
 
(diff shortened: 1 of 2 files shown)


vincentqb:
LGTM. Can you add tests? Building on what you suggested, I'd suggest to update the methods as such:
```python
def _test_yesno(self, dataset):
   ...  comparison code

def test_yesno_str(self):
  dataset = yesno.YESNO(self.root_dir)
  test_yesno_util(dataset)


def test_yesno_path(self):
  dataset = yesno.YESNO(Path(self.root_dir))
  test_yesno_util(dataset)
```

bhargavkathivarapu:
@vincentqb  Added the unit test

vincentqb (approved):
LGTM, thanks!

PR #1018, opened 2020-11-10
no longer download CommonVoice directly

vincentqb:
CommonVoice requests a user agreement before downloading, see [here](https://commonvoice.mozilla.org/en/datasets).
- raise an error if url is not None
- raise an error if download is True, and a warning if False

This should be cherry picked for 0.7.1.


cpuhrsch:
I suggest we follow the [conventions for imagenet](https://github.com/pytorch/vision/blob/master/torchvision/datasets/imagenet.py).

cpuhrsch (approved):

vincentqb:
cc @seemethere for 0.7.1 release

PR #1024, opened 2020-11-13
Workaround stuck job by using 'spawn' context

mthrok:
Looks like the jobs being stuck is related to the process initialization context.
Using 'spawn' context on DataLoader make it work for macOS.

Relates to https://github.com/pytorch/pytorch/issues/46409

TODO: set the context only on the environments known to have issues.


mthrok:
Note: Probably we should add note on documentation. (even for 0.7.0 release)

PR #1033, opened 2020-11-16
Add nightly build_docs, upload_docs jobs

mattip:
In PR gh-1108 I added support for building docs on every pull request. The intention was that the upload_docs job would run nightly to update the master version of the documentation once a day, but the job definitions were not sufficient to drive that. So this PR adds nightly_build_docs and nightly_upload_docs jobs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -780,21 +780,18 @@ workflows:
           name: binary_windows_conda_py3.8
           python_version: '3.8'
       - build_docs:
-          name: build_docs
+          name: always_build_docs
           python_version: '3.8'
           requires:
           - binary_linux_wheel_py3.8
       - upload_docs:
           filters:
-            branches:
-              only:
-              - nightly
             tags:
               only: /v[0-9]+(\.[0-9]+)*-rc[0-9]+/
           name: upload_docs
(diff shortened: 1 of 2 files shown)


vincentqb:
Oops, the job is failing?

mthrok:
Hi @mattip 

Thanks for looking into this.
Regarding the current nightly job, the `build_docs` step happens in [`build` workflow](https://app.circleci.com/pipelines/github/pytorch/audio/3871/workflows/7c57f69d-8a96-4e21-9438-45afb065ae00 ), which is same as when they are run on PRs. (They are basically the duplicated workflows. `build` is the basic one and `nightly` workflow uploads the built packages to conda.)

And if you look into [the failing job](https://app.circleci.com/pipelines/github/pytorch/audio/3871/workflows/7c57f69d-8a96-4e21-9438-45afb065ae00/jobs/116371), there is an issue with the script itself.  (`cp` failing.)

I think another approach here is to have the script fixed and nightly build leave too. That way you have less definitions for build doc job. What do you think?

mattip:
Ahh, thanks for the explaination. I will close this and open a new PR to just fix the script.

PR #1039, opened 2020-11-17
Add pathlib support for SPEECHCOMMANDS

bhargavkathivarapu:
Relates to #950 

Made some changes in speechcommand tests

In the existing one both `testSpeechCommands` and `testSpeechCommandsNone` are equivalent as the default argument for `subset` is **None**. So removed the `testSpeechCommandsNone` test

- Factored the common comparison in all tests into `_testSpeechCommands`
- Existing  `testSpeechCommands` is now `testSpeechCommands_str`
- Added new `testSpeechCommands_path`
- train,valid,test functions changed to use the common comparison code

@vincentqb could review these changes.

Do we need separate Path testing for train,valid,test subset ? I think it is not needed as we are already testing the Path functionality using whole dataset .

### DIFF
--- a/test/torchaudio_unittest/datasets/speechcommands_test.py
+++ b/test/torchaudio_unittest/datasets/speechcommands_test.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 
 from torchaudio.datasets import speechcommands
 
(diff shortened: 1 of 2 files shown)


vincentqb (approved):
Thanks for doing this, LGTM! And thanks for factoring the tests :)

Issue #1040, opened 2020-11-17
libsox does not detect mp3 when extension is missing


orlando-labs:
## 🐛 Bug

Sox IO backend doesn't allow to load file with explicit filetype while doesn't offer auto detection.

## To Reproduce

```
>>> torchaudio.set_audio_backend("sox")
>>> torchaudio.load("/path/to/extensionless/file", filetype='mp3')
(tensor([[0., 0., 0.,  ..., 0., 0., 0.]]), 16000)
>>> torchaudio.set_audio_backend("sox_io")
>>> torchaudio.load("/path/to/extensionless/file", filetype='mp3')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: load() got an unexpected keyword argument 'filetype'
>>> torchaudio.load("/path/to/extensionless/file")
formats: can't determine type of file `/path/to/extensionless/file'
```

## Expected behavior

Expected an autodetect feature or allowance to pass an explicit format argument 

## Environment

torch/torchaudio build from source. master of v0.7.0 affected

mthrok:
Hi @orlando-labs

Thanks for the feedback. I will look into it.

mthrok:
Looks like this happens on specific formats, like `mp3`, but not on `wav` or `flac`.

mthrok:
Hi @orlando-labs 

Do you use other formats than `mp3` that I should be looking into?

mthrok:
Investigation Note:

On torchaudio side, error happens [here](https://github.com/pytorch/audio/blob/f3b9208f8ab970ce8bcf16b6137ade282053c4b1/torchaudio/csrc/sox_utils.cpp#L97), which was called [here](https://github.com/pytorch/audio/blob/f3b9208f8ab970ce8bcf16b6137ade282053c4b1/torchaudio/csrc/sox_effects.cpp#L103)

This happens because [`sox_open_read`](https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/formats.c#L608) had some internal error and returns null pointer.

The actual implementation of `sox_open_read` is found in [`open_read`](https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/formats.c#L460) and the `filetype` member is set at 

https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/formats.c#L559

[...]

PR #1042, opened 2020-11-17
Add support for loading mp3 files without extension to "sox_io" backend

mthrok:
When it comes to `mp3` format detection, `libsox` only depends on the file extension. (See #1040)
This PR adds header-based format detection to `mp3`.

Relates to #1040 
~Should go to `0.7.1` release.~ Will ask users to test with nightly build.

### DIFF
--- a/test/torchaudio_unittest/common_utils/sox_utils.py
+++ b/test/torchaudio_unittest/common_utils/sox_utils.py
@@ -1,4 +1,5 @@
 import subprocess
+import warnings
 
 
 def get_encoding(dtype):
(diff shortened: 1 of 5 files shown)


cpuhrsch (approved):

vincentqb (approved):
Ideally, this should be upstreamed and we should see why this isn't in sox in the first place, but in the meanwhile this LGTM.

Hades32:
Did you ever try to upstream this? I just hit this issue and somehow this PR/issue are the only ones mentioning it...

mthrok:
@Hades32 No it is not upstreamed. Feel free to do so.

PR #1043, opened 2020-11-17
Use c10:optional in optional arguments

mthrok:
This replaces how optional arguments are processed in C++ code by using `c10::optional`.

### DIFF
--- a/torchaudio/csrc/sox_effects.cpp
+++ b/torchaudio/csrc/sox_effects.cpp
@@ -91,8 +91,8 @@ c10::intrusive_ptr<TensorSignal> apply_effects_tensor(
 c10::intrusive_ptr<TensorSignal> apply_effects_file(
     const std::string path,
     std::vector<std::vector<std::string>> effects,
-    const bool normalize,
-    const bool channels_first) {
+    c10::optional<bool>& normalize,
+    c10::optional<bool>& channels_first) {
   // Open input file
   SoxFormat sf(sox_open_read(
       path.c_str(),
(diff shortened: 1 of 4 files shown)


cpuhrsch (approved):

torchaudio/csrc/sox_effects.cpp
```diff
@@ -121,16 +121,17 @@ c10::intrusive_ptr<TensorSignal> apply_effects_file(
...
+  bool channels_first_ = channels_first.value_or(true);
   auto tensor = convert_to_tensor(
       /*buffer=*/out_buffer.data(),
       /*num_samples=*/out_buffer.size(),
       /*num_channels=*/chain.getOutputNumChannels(),
       dtype,
-      normalize,
-      channels_first);
+      normalize.value_or(true),
+      channels_first_);
```
cpuhrsch, line 131: I think you could even do
```
channels_first ? *channels_first : true
```

torchaudio/csrc/sox_effects.h
```diff
@@ -18,8 +18,8 @@ c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> apply_effects_tensor(
 c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> apply_effects_file(
     const std::string path,
     std::vector<std::vector<std::string>> effects,
-    const bool normalize = true,
-    const bool channels_first = true);
+    c10::optional<bool>& normalize,
```
cpuhrsch, line 21: I don't think passing by reference here actually makes a difference

torchaudio/csrc/sox_effects.h
```diff
@@ -18,8 +18,8 @@ c10::intrusive_ptr<torchaudio::sox_utils::TensorSignal> apply_effects_tensor(
...
+    c10::optional<bool>& channels_first);
```
vincentqb, line 22: i see, default already set by python

vincentqb (approved):
LGTM

torchaudio/csrc/sox_effects.h
mthrok: I read these and I thought it might be, so I took the safer option.

https://fb.workplace.com/groups/pytorch.dev/permalink/768545893723893/
https://fb.workplace.com/groups/pytorch.dev/permalink/717862835458866/

torchaudio/csrc/sox_effects.cpp
mthrok: Yeah, it's just the resulting value is used twice, so I had to cache it other wise we define default value twice.

PR #1044, opened 2020-11-17
Add format option to soundfile backend

mthrok:
Part of #1042 . This PR adds `format` option to the load function of soundfile backend. The soundfile backend does format detection correctly, so it does not need to be actually used.


mthrok:
Closing as I changed the strategy to resolve #1040. See #1042

PR #1045, opened 2020-11-18
Add pathlib support for TEDLIUM

bhargavkathivarapu:
Relates to #950 

- Factored the common comparison in all tests into `_test_tedlium`
- `path` test only for release1


vincentqb (approved):
Thanks again, LGTM!

PR #1048, opened 2020-11-19
Add Pathlib support for 'apply_effects_file'

bhargavkathivarapu:
Relates to #950 

For unit test copied the test for , I copied the existing one and passed Path object .
Didn't do any factoring as it fetching parameters of test case

### DIFF
--- a/test/torchaudio_unittest/sox_effect/sox_effect_test.py
+++ b/test/torchaudio_unittest/sox_effect/sox_effect_test.py
@@ -1,4 +1,5 @@
 import itertools
+from pathlib import Path
 
 from torchaudio import sox_effects
 from parameterized import parameterized
@@ -104,8 +105,8 @@ class TestSoxEffectsFile(TempDirMixin, PytorchTestCase):
         load_params("sox_effect_test_args.json"),
         name_func=lambda f, i, p: f'{f.__name__}_{i}_{p.args[0]["effects"][0][0]}',
     )
-    def test_apply_effects(self, args):
-        """`apply_effects_file` should return identical data as sox command"""
+    def test_apply_effects_str(self, args):
+        """`apply_effects_file` should return identical data as sox command when file path is given as a string"""
         dtype = 'int32'
         channels_first = True
         effects = args['effects']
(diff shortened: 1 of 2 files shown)


bhargavkathivarapu:
@vincentqb , The newly added testcases `test_apply_effects_str` , `test_apply_effects_path` under sox_effect_test.py ran successfully . But exisiting torchscript tests failed for test_apply_effect_file , from error message it seems Union Type is causing the issue

vincentqb:
Thanks for catching this: unfortunately, jit does not support Union type, pytorch/pytorch#41412. This means we won't be able to support path also with this function here until pytorch/pytorch#41412 is resolved.

Thanks for opening this pull request, we can keep it open in case pytorch/pytorch#41412 gets resolved and we would then be able to merge it even though this may take some time.

vincentqb (changes requested):
I'll mark this as "changed needed" until pytorch/pytorch#41412 is resolved, thanks for opening this and noticing the issue :)

test/torchaudio_unittest/sox_effect/sox_effect_test.py
```diff
@@ -127,6 +128,33 @@ def test_apply_effects(self, args):
         assert sr == expected_sr
         self.assertEqual(found, expected)
 
+    @parameterized.expand(
+        load_params("sox_effect_test_args.json"),
+        name_func=lambda f, i, p: f'{f.__name__}_{i}_{p.args[0]["effects"][0][0]}',
+    )
+    def test_apply_effects_path(self, args):
```
mthrok, line 135: This is overkill for testing the `Pathlib.Path` compatibility. Can you just pick up one test case, so that the focus of the test is intuitively clear that it's about `Pathlib.Path`.

torchaudio/sox_effects/sox_effects.py
```diff
@@ -153,7 +155,7 @@ def apply_effects_tensor(
 
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def apply_effects_file(
-        path: str,
+        path: Union[str, Path],
```
mthrok, line 158: You can remove the Union and just use `str`, but explaining why in the docstring which is how it is done in sox_io.

vincentqb: I second this idea :)

test/torchaudio_unittest/sox_effect/sox_effect_test.py
bhargavkathivarapu: @mthrok  changed the pathlib unit test by hard coding a specific test case argument

torchaudio/sox_effects/sox_effects.py
bhargavkathivarapu: @vincentqb changed type as per sox_io.
Also used `str()` like sox_io to convert Path to string, as` os.fspath()` is not supported by torchscript.
Now all checks passed.

vincentqb (approved):
LGTM, thanks!

torchaudio/sox_effects/sox_effects.py
```diff
@@ -1,4 +1,5 @@
-from typing import List, Tuple
+from typing import List, Tuple, Union
```
mthrok, line 1: @bhargavkathivarapu Can you remove unused `Union`?

vincentqb: #1052

PR #1052, opened 2020-11-20
remove unused imports in sox_effects

vincentqb:
cc #1048

(Out of scope for this PR: the lint test should pick unused dependencies where relevant and possible.)

### DIFF
--- a/torchaudio/sox_effects/sox_effects.py
+++ b/torchaudio/sox_effects/sox_effects.py
@@ -1,5 +1,4 @@
-from typing import List, Tuple, Union
-from pathlib import Path
+from typing import List, Tuple
 
 import torch
 


mthrok (approved):
Looks good, as long as tests pass

mthrok:
To enable linting for unused imports, `F401` has to be removed from `tox.ini`. 

But this is not perfect as it will also enable the warnings to `__init__.py`.
To workaround, 
1. one has to define `__all__` in `__init__.py`, or
2. add `#noqa: F401` to each import

I think these workaround requires a little bit of work at the beginning but is well justified and enforces a good practice.

BTW I do not understand why we have both `.flake8` and `tox.ini`.

vincentqb:
Quick partial notes:
- [text](https://github.com/pytorch/text) also has both files, but not pytorch and vision.
- [vision](https://github.com/pytorch/vision/blob/master/tox.ini#L3) has F401 too.
- Ideally, we should have a consensus between domains and pytorch.

PR #1026, opened 2020-11-13
Temporarily Disable OpenMP support for libsox

mthrok:
Sox effect functions and subprocess cause issues like #1021. 
In this PR, I added the tests that mimic this issue. This test properly fails in my development env (see the log bellow), however, it somehow does not fail in our CI. Furthermore, all the tests in `test/torchaudio_unittest/sox_effects/dataset_test.py` can fail with segmentation fault on my local env, but it does not happen in our CI. My environment is very similar to the CI settings (Docker/Ubuntu + Anaconda), and I can reproduce the issue on my working env but I have not figured out the key difference between my local env and CI.

What are common to these failure cases include;

* Run `sox_effects` in subprocess.
* Subprocesses are launched with `fork` method.

What we know is that;

* With `fork` method, setting `OMP_NUM_THREDAS=1` does not resolve the segmentation fault issue.
* Launching subprocesses with `spawn` would circumvent the segmentation fault.
  * but this incurs the initialization cost and slow down the data loading processes.  https://github.com/lhotse-speech/lhotse/issues/126
  * Setting `OMP_NUM_THREADS=1` helps the slowdown https://github.com/k2-fsa/snowfall/issues/18 but this affects all the operations of the application code (including PyTorch)
* Disabling OpenMP when compiling `libsox` resolves the issue. (`fork` works too)
  * This will remove `-fopenmp` flag when compiling `libsox`.

The above suggests that there is some sort of interference OpenMP of PyTorch (or MKL) and `libsox`.
This PR disables OpenMP support of `libsox` so that, at least users won't have an issue using these functionality while we investigate the issue further.

<details><summary>error report</summary>

```
============================================================================================================ test session starts =============================================================================================================
platform linux -- Python 3.8.3, pytest-5.4.3, py-1.9.0, pluggy-0.13.1
rootdir: /scratch/moto/torchaudio
plugins: hypothesis-5.18.0
collected 3 items

test/torchaudio_unittest/sox_effect/dataset_test.py FFFatal Python error: Segmentation fault

Current thread 0x00007f55ba56a740 (most recent call first):
  File "/scratch/moto/torchaudio/torchaudio/sox_effects/sox_effects.py", line 150 in apply_effects_tensor
  File "/scratch/moto/torchaudio/test/torchaudio_unittest/sox_effect/dataset_test.py", line 134 in speed
  File "/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/concurrent/futures/process.py", line 239 in _process_worker
[...]

Issue #1021, opened 2020-11-12
SoX effect "rate" crashing or hanging in multiprocessing


pzelasko:
## 🐛 Bug

This time I'm pretty sure it's a bug :P 

When running torchaudio speed + rate SoX effect chain inside of a `ProcessPoolExecutor` on the CLSP grid, the subprocess experiences segmentation fault inside the `apply_effects_tensor` function. I managed to make the subprocess wait and attached gdb to it, and this is the native stack trace I got:

```
Program received signal SIGSEGV, Segmentation fault.
0x00007fb3862087d6 in __kmp_acquire_ticket_lock () from /home/pzelasko/miniconda3/envs/lhotse/lib/python3.7/site-packages/mkl/../../../libiomp5.so
(gdb) bt
#0  0x00007fb3862087d6 in __kmp_acquire_ticket_lock () from /home/pzelasko/miniconda3/envs/lhotse/lib/python3.7/site-packages/mkl/../../../libiomp5.so
#1  0x00007fb3861dad4a in __kmpc_set_lock () from /home/pzelasko/miniconda3/envs/lhotse/lib/python3.7/site-packages/mkl/../../../libiomp5.so
#2  0x00007fb30e244746 in update_fft_cache (len=len@entry=2048) at /opt/conda/conda-bld/torchaudio_1603752092839/work/third_party/src/libsox/src/effects_i_dsp.c:190
#3  0x00007fb30e245187 in lsx_safe_rdft (len=2048, type=1, d=0x55f3b1e30700) at /opt/conda/conda-bld/torchaudio_1603752092839/work/third_party/src/libsox/src/effects_i_dsp.c:218
#4  0x00007fb30e254be5 in dft_stage_init (instance=instance@entry=0, Fp=0.91362772738460019, Fs=Fs@entry=1, Fn=2, att=att@entry=132.45319809215172, phase=phase@entry=50, stage=0x55f3b1ed5730, L=L@entry=2, M=1)
    at /opt/conda/conda-bld/torchaudio_1603752092839/work/third_party/src/libsox/src/rate.c:239
#5  0x00007fb30e2568b6 in rate_init (noSmallIntOpt=<optimized out>, max_coefs_size=400, interpolator=-1, use_hi_prec_clock=sox_false, maintain_3dB_pt=<optimized out>, rolloff=rolloff_small, anti_aliasing_pc=<optimized out>,
[...]

PR #1054, opened 2020-11-20
Cherry-pick: Temporarily Disable OpenMP support for libsox (#1026)

mthrok:
Currently `libsox` on Linux is compiled with GPU OpenMP and it interferes with the version PyTorch uses (Intel in case of binary distribution). This PR disables OpenMP support for `libsox`, while we investigate the way to use the same OpenMP as PyTorch's version.

### DIFF
--- a/test/torchaudio_unittest/sox_effect/dataset_test.py
+++ b/test/torchaudio_unittest/sox_effect/dataset_test.py
@@ -2,6 +2,7 @@ import sys
 import platform
 from unittest import skipIf
 from typing import List, Tuple
+from concurrent.futures import ProcessPoolExecutor
 
 import numpy as np
 import torch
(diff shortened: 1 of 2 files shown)


test/torchaudio_unittest/sox_effect/dataset_test.py
```diff
@@ -122,3 +123,36 @@ def test_apply_effects_tensor(self):
...
+            data = get_whitenoise(n_channels=1, sample_rate=sample_rate, duration=1, dtype='float')
+            save_wav(path, data, sample_rate)
+            self.flist.append(path)
+
+    def test_executor(self):
+        """Test that apply_effects_tensor with speed + rate does not crush
+
+        https://github.com/pytorch/audio/issues/1021
+        """
+        executor = ProcessPoolExecutor(1)
```
cpuhrsch, line 155: Why do you do this?

mthrok: This is the reproduction of tests reported in https://github.com/pytorch/audio/issues/1021

cpuhrsch (approved):

mthrok:
Thanks!

PR #1062, opened 2020-12-01
Fix multiple grammar and spelling mistakes in README

ProGamerGov:
The only assumption I made was that when referring to dimensions, "channel" was able to be more than one. Let me know if that is not the case, and then I can change it back.

Edit: 

I only changed the README, so the test failures do not appear related to anything I did.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ the audio domain. By supporting PyTorch, torchaudio follows the same philosophy
 of providing strong GPU acceleration, having a focus on trainable features through
 the autograd system, and having consistent style (tensor names and dimension names).
 Therefore, it is primarily a machine learning library and not a general signal
-processing library. The benefits of PyTorch is be seen in torchaudio through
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Looks good. Thanks!

README.md
```diff
@@ -195,16 +195,16 @@ Conventions
 
 With torchaudio being a machine learning library and built on top of PyTorch,
 torchaudio is standardized around the following naming conventions. Tensors are
-assumed to have channel as the first dimension and time as the last
```
vincentqb, line 198: nit: the choice of singular for channel was intentional :)

`channel` was used as "the name of the `channel` dimension" in a way similar to "the `mel` dimension", "the dimension of `freq`", `time`. Plural was avoided in order to avoid conjugating all other ones :)

It would have been a good idea to use single or double quotes instead to emphasize that this is the intended name.

As mentioned below, the number of channel would then be `n_channel`, in a way similar to `n_freq` or `n_mel`.

PR #1063, opened 2020-12-02
Add ComputeKaldiPitch prototype

mthrok:
Superseded by https://github.com/pytorch/audio/pull/1243

This is the initial step to add Kaldi's pitch feature.

## Approach

Kaldi's pitch feature is somewhat convoluted and it involves resampling and Viterbi algorithm, which is not easy to reproduce correctly in another language (i.e. Python). Therefore I took an approach to reuse Kaldi's original implementation as much as possible.

In this approach, I re-implemented the minimum interface of the family of Kaldi's matrix library, (such as `kaldi::VectorBase`,  `kaldi::MatrixBase`, etc...) with `torch::Tensor` class, and pulled the source files required to compile `kaldi/src/feat/pitch-functions.cc` with minimum modifications. (Commenting out some `#include` statements and correcting the type definitions. This is done in my fork of kaldi.)

The reason I used `torch::Tensor` class is that, this way, we do not need to be worried about the availability of a BLAS library, and the existing build process should just work. (with the exception of the addition of `git submodule` initialization) Once we find out a reliable way to detect BLAS library that PyTorch is linked against, then we should be able to switch to use bare Vector/Matrix classes.

The resulting `torchaudio.functional.compute_kaldi_pitch` function produces numerically identical result as the `compute-kaldi-pitch-feats` binary, but it is very slow. (x60) One major reason seems to be element-wise access, which Kaldi uses a lot but is not efficient in PyTorch.

There are three major things to consider forward;

## 1. Speed

The possible way to improve the speed is;
1. Modify the higher level implementation like `ComputeKaldiPitch` function to apply operation in vectorized manner.
2. Use the stripped version of the original implementation of Kaldi's Vector/Matrix classes and compile them with the same BLAS library that PyTorch uses. In Conda environment this is kind of easy, but we need to figure out the correct way to do so cross-platform (Linux/Windows/macOS, conda/pip)

### Batch
Currently `torchaudio.functional.compute_kaldi_pitch` is a simple wrapper around `kaldi::ComputeKaldiPitch` and can only process one waveform at a time. We need to extend the function so that it can handle batched samples. A naive approach will be to parallelize the operation over batch dimension.

### Return values
Currently the returned Tensor is 2D with `[time, pitch then NCCF]`. Providing an easy way to get pitch or NCCF without manually doing index slicing is preferred. One possible way is named Tensor.

## 3. Implementation detail

There are miscellaneous implementation details to be polished. Examples are;
- I found that bare `KALDI_ASSERT` hangs, so they have to be replaced with `TORCH_CHECK` or `TORCH_INTERNAL_ASSERT`
- Better way to maintain modified Kaldi. We should maintain a fork of Kaldi in pytorch org.


vincentqb:
To consider: would `git subtree` allow for a more appropriate way of handling the history and modification of kaldi code than `git submodule` ? e.g. [here](https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories)

mthrok:
Proposed improved approach;
 - find out the way to submoduling only needed (unmodified files) so that we do not need to keep original source code that we do not need to maintain the copy of unmodified files.
 - Only keep the modified source code.

PR #1067, opened 2020-12-04
[0.7.1] Update documentation

mthrok:
* Bump version to 0.7.1
* Checkout changes from the latest


vincentqb (approved):
LGTM.

docs/source/functional.rst
```diff
@@ -8,145 +8,176 @@ torchaudio.functional
 
 Functions to perform common audio operations.
 
-:hidden:`spectrogram`
```
vincentqb, line 11: the functionals aren't moved in 0.7, right?

docs/source/functional.rst
```diff
@@ -8,145 +8,176 @@ torchaudio.functional
...
-~~~~~~~~~~~~~~~~~~~~~
+:hidden:`Filtering`
```
vincentqb, line 11: the functionals aren't moved in 0.7 but this subsection isn't visible, right?

mthrok: It is visible. `hidden` here is a command to tell doxygen which class it should use to render this section.
Filtering subsection is for the improved user experience and has nothing to do with where the location these functions are defined.

vincentqb: Right, this seems reasonable

Issue #1051, opened 2020-11-19
Use glob in datasets instead of walk_files


vincentqb:
Replace walk_files+filter in datasets by globing, 
- [x] librispeech
- [x] libritts
- [x] speechcommands
- [x] ~~tedlium~~
- [x] vctk
- [x] yesno

Example (noted in [discussion](https://github.com/pytorch/audio/pull/1069#discussion_r542765428)):
```
walker = sorted(str(p) for p in Path(self._path).glob('*/*_no_hash_*.wav'))
```

- [x] Once this is done, we can deprecate `torchaudio.datasets.utils.walk_files`.

cc #910

faroit:
Shouldn't the globs be sorted for reproducibility reasons?

vincentqb:
Yes, tests should catch if the order is not consistent, see also [comment](https://github.com/pytorch/audio/pull/966#discussion_r512372561).

vincentqb:
Closed by #1109

PR #1069, opened 2020-12-05
Using Path and glob instead of walk_files

krishnakalyan3:
Hello @vincentqb,

I wanted to take an initial stab at this. I have made a minor modification for 

- [x] yesno
- [x] librispeech
- [x] libritts
- [x] speechcommands

Looks like `tedlium`, `ljspeech` and `vctk` (VCTK_092) don't have `walk_files`

cc https://github.com/pytorch/audio/issues/1051

### DIFF
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -125,7 +125,7 @@ class LIBRISPEECH(Dataset):
                 extract_archive(archive)
 
         walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)])
-        self._walker.sort()
+        self._walker = walker.sort()
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
         """Load the n-th sample from the dataset.
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -125,7 +125,7 @@ class LIBRISPEECH(Dataset):
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*'+self._ext_audio)])
+        walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)])
         self._walker = list(walker)
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -124,9 +124,7 @@ class LIBRISPEECH(Dataset):
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = [str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)]
-        walker.sort()
-        self._walker = walker
+        self._walker = sorted(str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio))
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
         """Load the n-th sample from the dataset.
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -8,7 +8,6 @@ from torch.utils.data import Dataset
 from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
-    walk_files,
 )
 
 URL = "train-clean-100"
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -124,7 +124,7 @@ class LIBRISPEECH(Dataset):
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)])
+        walker = [str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)]
         self._walker = walker.sort()
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -125,9 +125,7 @@ class LIBRISPEECH(Dataset):
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
-        )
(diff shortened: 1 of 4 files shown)


krishnakalyan3:
To test 

```
import torchaudio
yesno_data = torchaudio.datasets.YESNO('./', download=True)
yesno_data._walker
```

krishnakalyan3:
```
yesno_data._walker

Before:
['0_0_0_0_1_1_1_1',
'0_0_0_1_0_0_0_1',
'0_0_0_1_0_1_1_0',
'0_0_1_0_0_0_1_0',
'0_0_1_0_0_1_1_0', ...]

After:
['0_0_0_0_1_1_1_1',
 '0_0_0_1_0_0_0_1',
 '0_0_0_1_0_1_1_0',
 '0_0_1_0_0_0_1_0',
 '0_0_1_0_0_1_1_0', ...]
```

torchaudio/datasets/librispeech.py
```diff
@@ -125,9 +125,7 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
-        )
+        walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*'+self._ext_audio)])
```
vincentqb, line 128: nit: missing whitespace around arithmetic operator :)

torchaudio/datasets/yesno.py
```diff
@@ -85,9 +84,7 @@ def __init__(self,
                 "Dataset not found. Please use `download=True` to download it."
             )
 
-        walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
-        )
+        walker = sorted([str(p.stem) for p in Path(self._path).glob('*.wav')])
```
vincentqb, line 87: nit: `+ self._ext_audio` instead of `.wav`

vincentqb:
> Looks like `tedlium` does not contain `walk_files/filter` and I am assuming that `ljspeech` should also be added to the list?.

ljspeech also doesn't use `walk_files`, see [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/ljspeech.py#L78).

vincentqb:
Thanks for looking into this! gave minor comments, but LGTM so far :)

krishnakalyan3:
@vincentqb looks like `vctk` does not have `walk_files`.

torchaudio/datasets/librispeech.py
```diff
@@ -125,9 +125,7 @@ def __init__(self,
                     download_url(url, root, hash_value=checksum)
                 extract_archive(archive)
 
-        walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
-        )
+        walker = sorted([str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio)])
         self._walker = list(walker)
```
mthrok, line 129: Could you get rid of `list(walker)` as it is redundant?

I would do

```python
self._walker =  [str(p.stem) for p in Path(self._path).glob(f'*/*/*{self._ext_audio}')]
self._walker.sort()
```

mthrok (commented):
Mostly looks good. Could you also remove the definition of `walk_files`?

torchaudio/datasets/librispeech.py
vincentqb: Good point. It seems like using iterator+sorted may be faster? Is this a fair comparison?

```
In [7]: %timeit sorted(range(1000, 0, -1))                                                                                                                                                                                                          
14.3 µs ± 29.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [8]: def func(): 
   ...:     a = list(range(1000, 0, -1)) 
   ...:     a.sort() 
   ...:                                                                                                                                                                                                                                             

In [9]: %timeit func()                                                                                                                                                                                                                              
15 µs ± 201 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

vincentqb: I've updated the PR with sorted(generator). Once the tests pass, I'll merge this PR as it is. Thanks @krishnakalyan3 for the quick follow-up :)

vincentqb:
> Looks like `tedlium`, `ljspeech` and `vctk` (VCTK_092) don't have `walk_files`
> 
> Fixes: #1051

VCTK does have `walk_files`, so let's update VCTK and remove `walk_files` as a separate PR. Thanks again @krishnakalyan3  for the pull request :)

vincentqb (approved):
Thanks again for the work! We can do VCTK and removing `walk_files` as follow-up.

torchaudio/datasets/librispeech.py
mthrok: List comprehension is not a generator, please revert it.

mthrok: ah, okay, you changed the list comprehension to a generator.

mthrok: The next time, please consult it with me before doing so.

mthrok: I cannot find a reference that says `sorted` is in-place. can you give the reference for that?

vincentqb: See [implementation of sorted](https://github.com/python/cpython/blob/226a012d1cd61f42ecd3056c554922f359a1a35d/Python/bltinmodule.c#L2218). Regardless, the [test](https://github.com/pytorch/audio/pull/1069#discussion_r542804733) above points to that implementation to be faster anyway.

vincentqb:
LGTM! Merging. Thanks again @krishnakalyan3!

krishnakalyan3:
Thanks for all the awesome feedback gentlemen. 

The `VCTK` corpus which uses `walk_files` is deprecated in favor of `VCTK_092`. At this moment even if I use `VCTK` to forces me to use `VCTK_092` dataset. Hence, I did not modify as I was not sure that with the deprecation `walk_files` removal will be handeled.

mthrok:
> Thanks for all the awesome feedback gentlemen.
> 
> The `VCTK` corpus which uses `walk_files` is deprecated in favor of `VCTK_092`. At this moment even if I use `VCTK` to forces me to use `VCTK_092` dataset. Hence, I was not sure what to do.

@krishnakalyan3 

It was not explained in the original issue, but the fundamental reason of these changes you contributed is removing `walk_files`. The use of `walk_files` makes it ambiguous who is responsible to locate the files. (Dataset class? or utility?), and in fact just glob-ing everything is not the right problem being solved in implementing Dataset, because if you have a specific dataset you consider to access, then the directory structure and file locations are determined. Each Dataset implementation should be glob-ing the right set of files it requires.

Unfortunately, the original VCTK dataset is no longer publicly available and I do not have a copy so I do not know the expected structure of directory, but we can still move the glob logic to `VCTK` dataset so that `VCTK` dataset is responsible in locating the files. Would you like to work on it? If you do not have a time, I will file a separate issue for that.

krishnakalyan3:
@mthrok thanks for the explanation. I will work on the PR.

PR #1070, opened 2020-12-07
[0.7.1] Revert CommonVoice changes and disallow download=True

mthrok:
Split into #1074 #1076 #1076 

---

This PR
* Reverts #1065, which contains BC-braking changes, and undocumented default version change, which makes it unsuitable for minor release.
* Adds minimal changes to disallow `download=True` and direct users for manual download.
* Adds test for the changed behavior with archive extraction.

When review, please look at the commits except the first revert commit and ignore the revert. Click [here](https://github.com/pytorch/audio/pull/1070/files/6b00962da3dcec0ad584883381c46af2527debf4..093e5c66307d84182dfe1c8e6e61ba1d904f8b23).

**Note** Changes to the default version of the dataset `cv-corpus-4-2019-12-10` to `cv-corpus-5.1-2020-06-22` is not documented anywhere in the original PR, so I kept it reverted. I do not think we should make such subtle change on minor release, but I am open to changing it too.

Note 2: Also changed the upstream PyTorch version to point the `1.7.1rc`, which should have been added to `release/0.7` but was missing somehow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,12 +32,8 @@ commands:
       - run:
           name: adding UPLOAD_CHANNEL to BASH_ENV
           command: |
-            our_upload_channel=nightly
-            # On tags upload to test instead
-            if [[ -n "${CIRCLE_TAG}" ]] || [[ ${CIRCLE_BRANCH} =~ release/* ]]; then
-              our_upload_channel=test
-            fi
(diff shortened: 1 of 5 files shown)


mthrok:
Unrelated tests are failing, and somehow it is ignoring the config for `release/0.7` and instead it's using upstream master (1.8.0).

The tests for CommonVoice are passing, though

PR #1075, opened 2020-12-08
Revert "[Cherry-pick for 0.7.1] no longer download CommonVoice direct…

mthrok:
…ly (#1065)"

This reverts commit 1ec49fe6ce565510736d6a613083e0565570315b.

PR #1076, opened 2020-12-08
[0.7.1] Disallow download=True in CommonVoice

mthrok:
After https://github.com/pytorch/audio/pull/1075 is merged, this PR disallows `download=True`.

### DIFF
--- a/torchaudio/datasets/commonvoice.py
+++ b/torchaudio/datasets/commonvoice.py
@@ -179,6 +179,15 @@ class COMMONVOICE(Dataset):
             "romansh sursilvan": "rm-sursilv"
         }
 
+        if download:
+            raise RuntimeError(
+                "Common Voice dataset requires user agreement on the usage term, "
+                "and torchaudio no longer provides the download feature. "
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/commonvoice.py
```diff
@@ -196,23 +203,12 @@ def __init__(self,
...
-                    with open(filepath, "rb") as file_obj:
-                        if not validate_file(file_obj, checksum, "sha256"):
-                            raise RuntimeError(
-                                f"The hash of {filepath} does not match. Delete the file manually and retry."
-                            )
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _CHECKSUMS.get(url, None)
+                    download_url(url, root, hash_value=checksum)
```
vincentqb, line 210: This will never run, so the file needs to be extracted by the user.

mthrok: Updated the error message to `extract`.

vincentqb (approved):
LGTM

cpuhrsch (approved):

PR #1078, opened 2020-12-09
Update theme for 0.7 docs

mattip:
Update the stable (0.7) docs for the newer theme, which reformats the top navbar

### DIFF
--- a/0.7/_modules/index.html
+++ b/0.7/_modules/index.html
@@ -69,21 +69,7 @@
           </li>
 
           <li>
-            <div class="ecosystem-dropdown">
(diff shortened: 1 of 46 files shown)


mthrok (approved):
token approval.

mthrok:
At a glance, the font sizes are different. Is something changed functionally?

Screen Shot 2020-12-09 at 09 55 28

Screen Shot 2020-12-09 at 09 55 36

brianjo:
The theme is pretty dynamic, so we make little updates to it all the time. Should be fine.

mthrok:
> The theme is pretty dynamic, so we make little updates to it all the time. Should be fine.

Thanks for the clarification.

PR #1080, opened 2020-12-09
Disallow download=True in CommonVoice

mthrok:
After #1079, Cherry-pick #1076 into master branch.

See #1077 for the agenda.

Issue #1077, opened 2020-12-09
Update on CommonVoice


mthrok:
For master branch;

- [x] ~#1079~ Revert #1018
- [x] ~#1080~ Disallow `download=True` (cherry-pick ~#1076~)
- [x] #1082 Remove execution paths for `download=True` and clean up the code
- ~#1083~ Superseded by updated #1082 
    - ~Deprecate `download` and `url` argument~
    - ~Introduce `language` argument.~
    ~The introduction of `language` argument should come with the team's agreement.~
- ~Update the default version.~ Version parameter is deprecated and no longer used.
    ~This has to be a separate commit as this deserves a dedicated line in 0.8 release note.~

For release 0.7.1 branch;
- [x] ~#1075~ Revert #1065
- [x] ~#1076~ Disallow `download=True`

Issue #910, opened 2020-09-15
Better structure dataset implementations


vincentqb:
The suggestion is to adjust all the datasets (e.g. [speechcommands](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/speechcommands.py)) to follow some changes done in [tedlium](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py), see [example below](https://github.com/pytorch/audio/issues/910#issuecomment-708643874).

- [ ] Add a `_parse_filesystem` method to extract a list of "data point identifiers" in a pre-determined order to replace the generic `walk_files`, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L120) and #791.
- [ ] Move `_load_item` as method, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L134).
- [ ] Replace class attributes by constructor arguments, e.g. [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/librispeech.py#L85).
- [ ] Remove non-standard attributes, e.g. [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/gtzan.py#L1031)? or add attributes, e.g. [here](https://github.com/pytorch/tutorials/pull/1204#discussion_r518786423)?

Relates to #852, GTZAN #791, tedlium #882.
cc @mthrok @cpuhrsch

mthrok:
> * [ ]  Replace `walk` by an explicit iterator over the files, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L120)

Note iterator here is NOT a Python generator expression. None of the datasets require iterable, and requiring it to be itrable adds no benefit yet constraints.

> * [ ]  Rename `_walker` to `_filelist`, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L121).

What problem does this solve? I do not see the benefit of explicitly aligning the name of internal attributes.

> * [ ]  Move `_load_item` as method to main class, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L134).

The important thing about this is that the arguments to `_load_item` are sufficient information to locate a sample in the dataset.

> * [ ]  ~Add a (static?) method for loading audio, as [here](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/tedlium.py#L155).~

[...]

PR #1082, opened 2020-12-09
[BC-Breaking] Remove execution path for download=True and clean up

mthrok:
After #1080, this PR 
* removes downloading feature and logics 
* [BC-breaking] expects the dataset directory to be provided by users as `root` argument
* then deprecate the related constructor arguments

API-wise, it is not BC-braking, but users have to change the root directory they give to `root` argument.

|              |  `root` argument | Directory where data set is expected |
|--------|------------------|------------------------------------------------------|
| Before: |`<root_dir>`         | `<root_dir>/CommonVoice/<VERSION>/<LANGUAGE>` |
| After: |`<root_dir>`         | `<root_dir>` |

If you were using, `root="/data", url="english", version="cv-corpus-4-2019-12-10"`, then you need to change the `root` to `/data/CommonVoice/cv-corpus-4-2019-12-10/en`, (and you can should the `url` and `version` argument)

See #1077 for the agenda.
Closes #1083 #1077

### DIFF
--- a/torchaudio/datasets/commonvoice.py
+++ b/torchaudio/datasets/commonvoice.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import List, Dict, Tuple, Union
 
 import torchaudio
-from torchaudio.datasets.utils import unicode_csv_reader
+from torchaudio.datasets.utils import unicode_csv_reader, extract_archive
 from torch import Tensor
 from torch.utils.data import Dataset
 
@@ -136,6 +136,19 @@ class COMMONVOICE(Dataset):
 
         self._path = os.path.join(root, folder_in_archive)
         self._tsv = os.path.join(root, folder_in_archive, tsv)
+
+        if not os.path.isdir(self._path):
+            archive = os.path.join(root, f'{lang_code}.tar')
+            if not os.path.isfile(archive):
+                raise RuntimeError(
+                    f'Neither the data set directory "{folder_in_archive}", '
+                    f'or the archive "{lang_code}.tar" is found.')
+            extract_archive(archive)
+            if not os.path.isdir(self._path):
+                raise RuntimeError(
+                    f'The data set directory "{folder_in_archive}" is not found after '
+                    f'extracting "{lang_code}.tar" is found. '
+                    'Is the `version` matching to the archive?')
 
         with open(self._tsv, "r") as tsv:
             walker = unicode_csv_reader(tsv, delimiter="\t")
--- a/torchaudio/datasets/commonvoice.py
+++ b/torchaudio/datasets/commonvoice.py
@@ -3,7 +3,7 @@ from pathlib import Path
 from typing import List, Dict, Tuple, Union
 
 import torchaudio
-from torchaudio.datasets.utils import download_url, extract_archive, unicode_csv_reader
+from torchaudio.datasets.utils import unicode_csv_reader
 from torch import Tensor
 from torch.utils.data import Dataset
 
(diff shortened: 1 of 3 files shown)


torchaudio/datasets/commonvoice.py
```diff
@@ -104,120 +115,26 @@ class COMMONVOICE(Dataset):
...
-            if os.path.isfile(archive):
-                checksum = _CHECKSUMS.get(url, None)
-                if checksum:
-                    filepath = os.path.basename(url)
-                    with open(filepath, "rb") as file_obj:
-                        if not validate_file(file_obj, checksum, "sha256"):
-                            raise RuntimeError(
-                                f"The hash of {filepath} does not match. Delete the file manually and retry."
-                            )
-                extract_archive(archive)
```
vincentqb, line 213: Let's keep archive extraction, e.g. [here](https://github.com/pytorch/audio/pull/1018/files#diff-99487a0acfcf271732ec97aea1a9df354d52b6e55f05e0bf311080f12341e2bcR203-R212)

torchaudio/datasets/commonvoice.py
```diff
@@ -136,84 +119,22 @@ def __init__(self,
...
+        lang_code = _LANG_CODE[url]
+        folder_in_archive = os.path.join(folder_in_archive, version, lang_code)
 
         self._path = os.path.join(root, folder_in_archive)
-
-        if download:
-            if not os.path.isdir(self._path):
-                if not os.path.isfile(archive):
-                    checksum = _CHECKSUMS.get(url, None)
-                    download_url(url, root, hash_value=checksum)
```
cpuhrsch: While we can't download the dataset due to legal reason, we can still extract it.

If the user has not extracted the dataset yet, we can do so.

If we're about to extract and the user has provided a checksum (from the common voice website) we can also verify it.

mthrok: It turned out that dataset archives have different directory structure for different version.

* Version 4 - `cv-corpus-4-2019-12-10`

```
$ tar -tvf tmp/en.tar.gz
2019-12-09 17:47 .DS_Store
2019-12-10 08:09 clips/
2019-12-10 05:41 dev.tsv
2019-12-10 05:41 invalidated.tsv
2019-12-10 05:41 other.tsv
2019-12-10 05:41 test.tsv
2019-12-10 05:41 train.tsv
2019-12-10 05:41 validated.tsv
...
```

* Version 5.1 - `cv-corpus-5.1-2020-06-22`
```
$ tar -tvf en.tar.gz
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/
2020-07-11 20:40 cv-corpus-5.1-2020-06-22/en/clips/
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/dev.tsv
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/invalidated.tsv
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/other.tsv
2020-07-11 20:41 cv-corpus-5.1-2020-06-22/en/reported.tsv
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/test.tsv
2020-07-11 23:59 cv-corpus-5.1-2020-06-22/en/train.tsv
...
```

Also, because of this (and the fact that `folder_in_archive` argument which is not used properly), the previous implementation (download + extract) does not work. https://github.com/pytorch/audio/issues/1088 (I suspect it ever worked.)

The thing is that the current implementation expects the files to be presented at `<ROOT_DIR>/CommonVoice/<VERSION>/<LANG_CODE>`. But the previous code behavior extracts the archive at either `<ROOT_DIR>/` or `<ROOT_DIR>/<VERSION>/<LANG_CODE>`.

I was thinking to resort to the original behavior when handling archive, but since there was not working behavior in the previous implementation when it comes to download + extract, I again propose not to try to extract the archive and ask users to extract the archive and provide the directory where the data are found. (and deprecate `folder_in_archive`) argument in addition to `download` and `url` arguments.

mthrok: offline discussion: change the `root` to expect the exact directory users extract the archive.

vincentqb (approved):
LGTM. In fact, we could consider this simplified interface for other datasets too, cc #910.

torchaudio/datasets/commonvoice.py
```diff
@@ -131,93 +50,36 @@ class COMMONVOICE(Dataset):
...
+        deprecated = [
+            ('url', url),
+            ('folder_in_archive', folder_in_archive),
+            ('version', version),
+            ('download', download)
+        ]
+        for name, val in deprecated:
+            if val is not None:
+                warnings.warn(
+                    f"`{name}` argument is no longer used and deprecated. "
```
cpuhrsch, line 73: As a follow-up, can we improve the message here? What should someone do who is currently already using this dataset?

torchaudio/datasets/commonvoice.py
```diff
@@ -131,93 +50,36 @@ class COMMONVOICE(Dataset):
...
-        if url not in languages:
-            raise ValueError(f"`url` must be one of available languages: {languages.keys()}")
-
-        if url in languages:
-            ext_archive = ".tar.gz"
-            language = languages[url]
-
-            base_url = "https://voice-prod-bundler-ee1969a6ce8178826482b88e843c335139bd3fb4.s3.amazonaws.com"
-            url = os.path.join(base_url, version, language + ext_archive)
+                "Please download the dataset and extract it manually.")
```
cpuhrsch, line 62: As a follow-up, we can provide a bit more guidance (a link or such) on how to go about this.

cpuhrsch (approved):

torchaudio/datasets/commonvoice.py
mthrok: I thought `no longer used` + `Please remove it from the function call` would suffice. What do you suggest?

PR #1083, opened 2020-12-09
 Deprecate download/url args and add language arg in CommonVoice

mthrok:
Aftert #1080, this PR deprecates `download` and `url` arguments in CommonVoice and add `language` for replacement of `url` argument.

See #1077 for the agenda.


mthrok:
Superseded by #1082

PR #1022, opened 2020-11-12
[DOC] Add missing modules and minor fixes

krishnakalyan3:
- [ ] Issues in functional.py -> https://github.com/pytorch/audio/pull/1022#issuecomment-726293803
- [x] Invalid Link  ->  Scott Lehman, Effects Explained
- [x] griffinlim reference -> https://github.com/pytorch/audio/pull/1022#issuecomment-726985780
- [x] DB_to_amplitude -> https://github.com/pytorch/audio/issues/1002#issuecomment-721445590
- [x] Issue in datasets.rst (griffinlim, DB_to_amplitude) -> https://github.com/pytorch/audio/issues/1002#issuecomment-721365993
- [x] Issue in sox_effects.py -> https://github.com/pytorch/audio/issues/1002#issuecomment-721364845
- [x] Issue in sox_utils.py -> The list is not properly [rendered](https://pytorch.org/audio/0.7/utils.html#torchaudio.utils.sox_utils.set_verbosity) 

Fixes for #1002

PR #1086, opened 2020-12-12
Remove redundant dataset utilities 

### LINKED ISSUE #1081: Remove redundant dataset utilities
Some utilities in `torchaudio.datasets.utils` are no longer needed and we should get rid of them.

* [`unicode_csv_reader`](https://github.com/pytorch/audio/blob/a2085b85f04d693222297bf614a5e5b34de04986/torchaudio/datasets/utils.py#L21-L48)

* [`makedir_exist_ok`](https://github.com/pytorch/audio/blob/a2085b85f04d693222297bf614a5e5b34de04986/torchaudio/datasets/utils.py#L51-L61) (also mentioned in #477)

krishnakalyan3:
I have also addressed your comments on torch.random https://github.com/pytorch/audio/pull/1022#issuecomment-738500681. 
Fixes #1081

### DIFF
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -1,4 +1,5 @@
 import os
+import csv
 from pathlib import Path
 from typing import Tuple, Union
 
@@ -8,7 +9,6 @@ from torch.utils.data import Dataset
 from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
-    unicode_csv_reader,
 )
 
 URL = "aew"
@@ -154,7 +154,7 @@ class CMUARCTIC(Dataset):
         self._text = os.path.join(self._path, self._folder_text, self._file_text)
 
         with open(self._text, "r") as text:
-            walker = unicode_csv_reader(text, delimiter="\n")
+            walker = csv.reader(text, delimiter="\n")
             self._walker = list(walker)
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str]:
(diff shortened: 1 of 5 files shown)


mthrok:
Thanks for working on this but you need to get rid of all the internal use.
Please run tests by `pytest -vs test/torchaudio_unittest/datasets/`

krishnakalyan3:
Hi, @mthrok looks like I am missing some context. Can you please let me know why `unicode_csv_reader` should be removed?.
If removed what should I be replacing it with?.

mthrok:
These utilities were added (for internal use) to ease the difference between Python 2 and Python 3.

But now that Python 2 support is officially dropped, we do not need these utilities. Instead we can use the standard module as-is.
In Python 3 `os.makedirs` function has `exist_ok` argument, and csv module can handle unicode natively without the hack of changing the `field_limit_size`.

So we just need to replace the usages of these functions with standard module. Then we can remove the utility functions themselves.

krishnakalyan3:
@mthrok can you please review this.

mthrok (approved):
Looks good. Thank you so much!

krishnakalyan3:
@mthrok can you please recommend some other (easy - medium complexity) issues that I could work on?. 

Cheers,
Krishna

mthrok:
@krishnakalyan3 Sure. Thanks for offering :) Let me think of an issue. Meanwhile, if you can join PyTorch's slack channel, we can discuss collaboration there too. Communication will be faster that way. https://pytorch.org/resources/

mthrok:
@krishnakalyan3 

How about https://github.com/pytorch/audio/issues/910#issuecomment-708643874 ?
I consider this as *not easy* in terms of complexity because there are somewhat open-ended perspectives in the specification, so we need to iterate on the design.

Issue #1057, opened 2020-11-26
Torchaudio resampling could be faster and simpler


adefossez:
## 🐛 Bug

I have been implementing a few DSP algorithms lately, including sinc based resampling.
The following implementation could simplify the code in torchaudio and is also faster (single call to conv1d, instead of multiple calls to conv1d and conv_transpose1d). I'll be happy to open a pull requests if you are interested.

https://github.com/adefossez/julius/blob/main/julius/resample.py

## To Reproduce

1. Install modules `pip install -U torch torchaudio julius`
2. Run the benchmark code hereafter

```python
from torchaudio.compliance import kaldi
import torch
import julius
x = th.randn(1, 44100 * 10)
rolloff = 0.99  # lowpass filter freq used by torchaudio
zeros = 6  # use the same number of zero crossing as torchaudio
for from_sr, to_sr in [(5, 7), (7, 5)]:
    print("comparing for", from_sr, to_sr)
    %timeit kaldi.resample_waveform(x, from_sr, to_sr)
    %timeit julius.resample_frac(x, from_sr, to_sr, rolloff=rolloff, zeros=zeros)
    yt = kaldi.resample_waveform(x, from_sr, to_sr)
    yj = julius.resample_frac(x, from_sr, to_sr, rolloff=rolloff, zeros=zeros)
    print(torch.norm(yt - yj))
```

## Expected behavior

Would expect torchaudio to be as fast as julius, results are

```bash
comparing for 5 7
32.4 ms ± 309 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
4.46 ms ± 44.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
tensor(8.7085e-05)
comparing for 7 5
12.9 ms ± 323 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
2.83 ms ± 43.9 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
tensor(5.1333e-05)
```

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)? pip
 - If you are building from source, which commit is it?
 - What does `torchaudio.__version__` print? 0.7.0a0+ac17b64

PyTorch version: 1.7.0
Is debug build: True
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 10.15.7 (x86_64)
GCC version: Could not collect
Clang version: 11.0.0
CMake version: version 3.18.4

[...]

Issue #891, opened 2020-08-26
float sample_rate


vincentqb:
torchaudio currently treats sample_rate as an integer. In #890, it was discuss that float would be more appropriate. 

- Librairies we use for saving/loading may not support float sample_rate.
- C++ extensions currently use int.
- torchaudio signatures use int.

Thoughts?

faroit:
ping @f0k @bmcfee @justinsalamon @nils-werner to get some more dsp guys on board here...

bmcfee:
:100: for floating point.  There's no good reason to require integer sampling rates, and any interoperability requirements (eg C++ extensions) should have an explicit cast so it's clear what's going on.

psmaragdis:
Floating point.  How would you decimate 44,100 by a factor of 8 otherwise?!

justinsalamon:
Looks like there's no need for further comments :)

faroit:
> Looks like there's no need for further comments on my part :)

any comment by @psmaragdis should indeed taken seriously :-D

Here is a list of how some popular audio packages handle sample rate types

* __[soxr](https://github.com/WavX/soxr/blob/master/src/soxr.h#L101):__ `double`
* __[libsndfile](https://github.com/erikd/libsndfile/search?p=3&q=samplerate&unscoped_q=samplerate):__ not consistent, mostly int
* __[ffmpeg](https://github.com/FFmpeg/FFmpeg/search?p=5&q=samplerate&unscoped_q=samplerate):__ not consistent
* __[tensorflow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/lib/wav/wav_io.cc):__ int32
* __[tfio](https://github.com/tensorflow/io/blob/master/tensorflow_io/core/ops/audio_ops.cc):__ int64/int32
* __[aubio](https://github.com/aubio/aubio/blob/05ed7f562616cee876d2e4cd397cdb88f600942f/src/io/sink.c#L37):__ int
* __[portaudio](https://github.com/audacity/audacity/blob/8b87e18c6d752f4f86c556515fd3be4ca80cfd97/lib-src/portaudio-v19/qa/loopback/src/test_audio_analyzer.c#L63)__: double
* __[libpd](https://github.com/audacity/audacity/blob/8b87e18c6d752f4f86c556515fd3be4ca80cfd97/lib-src/portaudio-v19/qa/loopback/src/test_audio_analyzer.c#L63):__ inconsistent, mostly int
[...]

PR #1087, opened 2020-12-14
Simpler and faster resampling

adefossez:
This is the implementation for https://github.com/pytorch/audio/issues/1057
Implementation is taken from [julius/resample.py](https://github.com/adefossez/julius/blob/main/julius/resample.py) and adapted for consistency with the previous implementation. See the issues linked above for a comparison in speed.
Output should be exactly the same as I changed the default parameters to match those of the previous implementation.

### DIFF
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -772,16 +772,16 @@ def _get_sinc_resample_kernel(orig_freq: int, new_freq: int, lowpass_filter_widt
     #    y[j] = sum_i x[i] sinc(pi * orig_freq * (i / orig_freq - j / new_freq))
 
     # We see here that y[j] is the convolution of x[i] with a specific filter, for which
-    # we take an FIR approximation, stopping when we see at least `zeros` zeros crossing.
+    # we take an FIR approximation, stopping when we see at least `lowpass_filter_width` zeros crossing.
     # But y[j+1] is going to have a different set of weights and so on, until y[j + new_freq].
     # Indeed:
     # y[j + new_freq] = sum_i x[i] sinc(pi * orig_freq * ((i / orig_freq - (j + new_freq) / new_freq))
-    #               = sum_i x[i] sinc(pi * orig_freq * ((i - orig_freq) / orig_freq - j / new_freq))
-    #               = sum_i x[i + orig_freq] sinc(pi * orig_freq * (i / orig_freq - j / new_freq))
+    #                 = sum_i x[i] sinc(pi * orig_freq * ((i - orig_freq) / orig_freq - j / new_freq))
+    #                 = sum_i x[i + orig_freq] sinc(pi * orig_freq * (i / orig_freq - j / new_freq))
     # so y[j+new_freq] uses the same filter as y[j], but on a shifted version of x by `orig_freq`.
     # This will explain the F.conv1d after, with a stride of orig_freq.
     width = math.ceil(lowpass_filter_width * orig_freq / base_freq)
-    # If old_sr is still big after GCD reduction, most filters will be very unbalanced, i.e.,
+    # If orig_freq is still big after GCD reduction, most filters will be very unbalanced, i.e.,
     # they will have a lot of almost zero values to the left or to the right...
     # There is probably a way to evaluate those filters more efficiently, but this is kept for
     # future work.
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -852,5 +852,5 @@ def resample_waveform(waveform: Tensor,
     waveform = F.pad(waveform, (width, width + orig_freq))
     resampled = F.conv1d(waveform[:, None], kernel, stride=orig_freq)
     resampled = resampled.transpose(1, 2).reshape(num_wavs, -1)
-    target_length = int(new_freq * length / orig_freq)
+    target_length = int(math.ceil(new_freq * length / orig_freq))
     return resampled[..., :target_length]
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -754,14 +754,17 @@ def mfcc(
 
 
 def _get_sinc_resample_kernel(orig_freq: int, new_freq: int, lowpass_filter_width: int,
-                              rolloff: float, device: torch.device, dtype: torch.dtype):
+                              lowpass_cutoff_ratio: float,
+                              device: torch.device, dtype: torch.dtype):
+    assert lowpass_filter_width > 0
+    assert 1 >= lowpass_cutoff_ratio > 0
     kernels = []
     base_freq = min(orig_freq, new_freq)
     # rolloff will perform antialiasing filtering by removing the highest frequencies.
     # At first I thought I only needed this when downsampling, but when upsampling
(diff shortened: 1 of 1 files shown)


torchaudio/compliance/kaldi.py
```diff
@@ -752,147 +753,60 @@ def mfcc(
...
+    scale = base_freq / orig_freq
+    return torch.stack(kernels).view(new_freq, 1, -1).mul_(scale), width
 
 
 def resample_waveform(waveform: Tensor,
                       orig_freq: float,
                       new_freq: float,
-                      lowpass_filter_width: int = 6) -> Tensor:
+                      lowpass_filter_width: int = 6,
+                      rolloff: float = 0.99) -> Tensor:
```
mthrok, line 809: In my understanding, `rolloff` was the constant `0.99` in the original implementation.
If so, why not keep it that way? What's the necessity to export it as public API?

torchaudio/compliance/kaldi.py
```diff
@@ -904,80 +818,39 @@ def resample_waveform(waveform: Tensor,
 
     Args:
         waveform (Tensor): The input signal of size (c, n)
-        orig_freq (float): The original frequency of the signal
-        new_freq (float): The desired frequency
+        orig_freq (int): The original frequency of the signal
+        new_freq (int): The desired frequency
```
mthrok, line 822: If we are changing the docstring, I also suggest to change the annotation to `int`. This is kind of BC-breaking but it is an enforcement as it was expected in the original implementation too.

mthrok (commented):
Thanks for working on this. I made some comments.

torchaudio/compliance/kaldi.py
adefossez: I was thinking it could be useful to power users (rolloff being always a trade off between limiting aliasing while not losing too much signal, and there is not perfect value). It is not so different from allowing the end user to change `lowpass_filter_width`, and it does not break backward compatibility. But if you prefer I can remove it.

torchaudio/compliance/kaldi.py
adefossez: Yes, I forgot to do it but intended too.

torchaudio/compliance/kaldi.py
adefossez: I have renamed the parameter to `lowpass_cutoff_ratio` for clarity, if you agree on keeping it in the api :)

mthrok: I have somewhat mixing feeling about adding the new parameter. But I am not DSP expert so I cannot judge. If you go strong on adding it, I do not have a strong counter argument other than keeping API same. However, the way you described, `I was thinking it could be useful to power users` sounds it's still the assumption, it would be nice if we had some evidence.

Is there a library that expose similar parameter? I know that `lowpass_filter_width` is originated from Kaldi.

mthrok:
Looking at the CI jobs, the tests which were feeding `float` sampling rate are now failing.

Can I suggest to split this PR into the three PRs?
1. The speed up part
2. Changing sampling rate from float to int
3. Adding parameter for `lowpass_cutoff_ratio `

Then, the speed up part should be merged easily, and for the others I can bring them up in our team meeting to get more feedback.

adefossez:
I'm wondering whether I should keep the assertion that the sample rates are indeed integers (even if typed as float). People might be using the current implementation with true floats, and in that case my code would throw an exception, while the previous one would not, and the change would not be backward compatible.

In the original code, a mixture of integer and float values was used. I don't think this is completely accurate as for instance resampling from `2` to `2.9999` is going to be quite different from resampling from `2` to `3`. I could try to emulate this in the new implementation but I need a bit more time to make sure I am doing the same thing as the previous implementation.

I would need to add some form of unit tests because non integer change of sample rate are not covered in the tests. I could add a saved file from the current implementation as a test resource, but I'm afraid this would scare future contributors from trying to change that bit, while in fact the current implementation is doing something a bit weird and somewhat incorrect.

vincentqb:
> I'm wondering whether I should keep the assertion that the sample rates are indeed integers (even if typed as float). People might be using the current implementation with true floats, and in that case my code would throw an exception, while the previous one would not, and the change would not be backward compatible.
> 
> In the original code, a mixture of integer and float values was used. I don't think this is completely accurate as for instance resampling from `2` to `2.9999` is going to be quite different from resampling from `2` to `3`. I could try to emulate this in the new implementation but I need a bit more time to make sure I am doing the same thing as the previous implementation.
> 
> I would need to add some form of unit tests because non integer change of sample rate are not covered in the tests. I could add a saved file from the current implementation as a test resource, but I'm afraid this would scare future contributors from trying to change that bit, while in fact the current implementation is doing something a bit weird and somewhat incorrect.

As mentioned in #891, we do wish to change sample rate to floats in torchaudio. Parts of the code that assume integer sample rate should check that condition to avoid silent correctness issue later. 

If parts of this PR already support float for sample rate, then either we start with integer only and then extend in a new PR with tests later, or we could do it here (though this may make for an even longer PR).

adefossez:
@vincentqb  for some reason I thought the old code was doing something weird if the sample rate were floats, which I was worried about, but actually it was equivalent to just converting those with `int` first, which is what the new code does too.
So I've removes here everything change but the change to the algorithm itself, and this is functionally equivalent to the previous implementation.

adefossez:
Ping @vincentqb @mthrok :) let me know if you are good with the current version, it is completely equivalent to the previous one.

torchaudio/compliance/kaldi.py
```diff
@@ -752,141 +753,54 @@ def mfcc(
...
+def _get_sinc_resample_kernel(orig_freq: int, new_freq: int, lowpass_filter_width: int,
+                              device: torch.device, dtype: torch.dtype):
+    assert lowpass_filter_width > 0
+    kernels = []
+    base_freq = min(orig_freq, new_freq)
+    # This will perform antialiasing filtering by removing the highest frequencies.
+    # At first I thought I only needed this when downsampling, but when upsampling
+    # you will get edge artifacts without this, as the edge is equivalent to zero padding,
+    # which will add high freq artifacts.
+    base_freq *= 0.99
```
vincentqb, line 765: nit: I'm curious how this constant was chosen, and the description before it doesn't provide that information. Can you elaborate?

vincentqb (approved):
This LGTM, thanks!

torchaudio/compliance/kaldi.py
adefossez: It was chosen by the previous implementation so I did not try to change it. The previous implementation did not provide a rational on how this was chosen.
In my own implementation for julius I actually used the same one as resampy which is a bit lower (0.945 I think). Overall this constant controls the trade off between keeping high frequencies and removing aliasing. I don't think there is a perfect value. I would say 0.99 is probably not as conservative as the one used in resampy, and can likely introduce some aliasing, but it would be rather subtle.

vincentqb: Ah, I missed that the parameter already was being used. Thanks for making it stand out more :)

mthrok:
@adefossez Sorry for not getting back to you sooner. Let me come back to this the next week.

mthrok (approved):

mthrok:
@adefossez 

Sorry for not getting back to you sooner. The PR looks good. Thanks!

PR #1049, opened 2020-11-19
add org-member context and generate a .netrc file for git push

mattip:
Follow what is done in pytorch/pytorch to succeed at `git push`

PR #1041, opened 2020-11-17
Fix doc commit script 2

mattip:
The fix in #1034 got a little further in the [document commit job](https://app.circleci.com/pipelines/github/pytorch/audio/3893/workflows/23a3e829-9592-499a-8445-acf0884f0179/jobs/117250), but there was another mistake: the documentation lives in the top-level directory not inside `docs`. This PR should fix it. Also fix for first-time creation of the directories if they do not exist.

PR #1034, opened 2020-11-16
fix typos in commit_docs.sh script

mattip:
fix some typos in the script that prevented it from working

Issue #1031, opened 2020-11-15
CI: Nightly build is not running build_docs


mattip:
PR gh-1006 added `build_docs` and `upload_docs` jobs, the former is meant to run on every PR, merge, and nightly run, and the latter on nightly and tags. But the `build_docs` does not [run on nightly](https://app.circleci.com/pipelines/github/pytorch/audio/3868/workflows/b30f6af1-3252-4eaf-b81c-a8d6b230c43b) since it [requires `binary_linux_wheel_py3.8`](https://github.com/pytorch/audio/pull/1006/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R781) and not one of the nightly jobs. One alternative would be to have `build_docs_nightly` and `upload_docs_nightly` that duplicate the current ones but use `requires: nightly_binary_linux_wheel_py3.8`

mthrok:
Closing as this is addressed in  #1033 and #1034.

PR #1091, opened 2020-12-15
remove checkout step from upload

mattip:
Continuation of gh-1049, gh-1041, gh-1034, gh-1031, gh-1006 to automate the docs build and push the result to the gh-pages branch. We cannot use the checkout step since it will force push-via-ssh, which will fail. Instead re-use the git checkout from the `build_docs` step.

Also remove the "designate_upload_channel" step which is for uploading wheels, which is not necessary here.


mthrok (approved):
Thanks! let's see how this works.

mattip:
I think it will run in about 20 hours

PR #1092, opened 2020-12-16
DOC: move 0.7->0.7.0, rebuild master

mattip:
Two things combined, if it is too complicated to review I can redo them as two separate PRs:
- move 0.7 documentation to 0.7.0 in prerapration for 0.7.1 doc creation once there is a v0.7.1 tag. Maybe that should be v0.7.0 instead (with a 'v')?
- rebuild documentation ~on~ off master, to see what the 0.7.1 docs will look like.


mattip:
hmm. "build error" - I guess that is expected on this PR to a branch.

mthrok (approved):
Did spot check, looks good. Thanks!

PR #1096, opened 2020-12-17
Update calls to torch.stft to have return_complex=True

### LINKED ISSUE #1095: torch.stft calls need to be updated to set the return_complex kwarg
torch.stft (https://pytorch.org/docs/master/generated/torch.stft.html?highlight=stft#torch.stft) has deprecated not setting the return_complex kwarg, and we want to make passing return_complex required before the PyTorch 1.8 release. To avoid torchaudio breaking with PyTorch nightlies or PyTorch 1.8 it needs to update its torch.stft calls to pass the return_complex argument.

mthrok:
Resolves #1095

### DIFF
--- a/test/torchaudio_unittest/batch_consistency_test.py
+++ b/test/torchaudio_unittest/batch_consistency_test.py
@@ -236,6 +236,7 @@ class TestTransforms(common_utils.TorchaudioTestCase):
             'pad_mode': 'reflect',
             'normalized': True,
             'onesided': True,
+            'return_complex': False,
         }
         rate = 2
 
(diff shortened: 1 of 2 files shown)


mruberry (approved):
Thank you, @mthrok!

peterbell10:
Since the plan is to deprecate `return_complex=False`, you may want to go to `torch.view_as_real(torch.stft(..., return_complex=True))`.

mruberry:
That's a good future proofing idea, @peterbell10, but I suggest we just go with the simpler and smaller change for the moment and revisit how torchaudio wants to handle complex numbers for the 1.9 release. Maybe torchaudio will want to switch to using complex tensors.

peterbell10:
Not sure I understand. `view_as_real` is the exact same data format as `return_complex=False`, so it doesn't require any further changes. It just avoids the warning.

mruberry:
Oh you're right, thank you. I forgot we're also putting a warning on this. It's been a long day...

Sorry @mthrok, I was mistaken. We should take @peterbell10's recommendation to use `view_as_real(torch.stft(..., return_complex=True))` because this will avoid a warning, since we're also deprecating `return_complex=False` for the 1.8 release.

mthrok:
@peterbell10 @mruberry Update the PR to use `return_complex=True` + `view_as_real`. Review please.

mruberry (approved):
This looks great. Thank you again @mthrok and @peterbell10!

peterbell10 (approved):

PR #1093, opened 2020-12-16
remove buggy debug cruft

mattip:
The nightly task [ran but failed](https://app.circleci.com/pipelines/github/pytorch/audio/4135/workflows/4d0ea223-9e2e-47c4-95b3-3140a60c9f8d/jobs/127799). The debug line to show the contents of `doc` should have been `docs` instead. It did, however confirm the existance of the `docs` directory, so the PR removes the debug cruft so the script hopefully will run to completion.

<details>
<summary>Log output</summary>

```
#!/bin/bash -eo pipefail
# Don't use "checkout" step since it uses ssh, which cannot git push
# https://circleci.com/docs/2.0/configuration-reference/#checkout
set -ex
echo $PWD
ls ~/workspace
ls ~/workspace/doc
tag=${CIRCLE_TAG:1:5}
target=${tag:-master}
.circleci/build_docs/commit_docs.sh ~/workspace $target

+ echo /root/project
/root/project
+ ls /root/workspace
archives	    README.md
build_tools	    requirements.txt
CODE_OF_CONDUCT.md  setup.py
docs		    test
examples	    third_party
LICENSE		    torchaudio
mypy.ini	    torchaudio-0.8.0.dev20201216-cp38-cp38-linux_x86_64.whl
packaging	    tox.ini
+ ls /root/workspace/doc
ls: cannot access /root/workspace/doc: No such file or directory

Exited with code exit status 2

CircleCI received exit code 2
```

PR #1098, opened 2020-12-17
script is not in PWD

mattip:
Continuation of gh-1093: the script is located in the repo, which is in `~/workspace`.

Is there a way to trigger the nightly build after merging this so the cycles of trial and error can move faster?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -698,10 +698,9 @@ jobs:
             # Don't use "checkout" step since it uses ssh, which cannot git push
             # https://circleci.com/docs/2.0/configuration-reference/#checkout
             set -ex
-            echo $PWD
             tag=${CIRCLE_TAG:1:5}
             target=${tag:-master}
-            .circleci/build_docs/commit_docs.sh ~/workspace $target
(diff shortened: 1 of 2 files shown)


mthrok (approved):

mthrok:
The trigger is in Facebook's internal system so I can run it manually.
I cannot change the configuration so the next time, leave a comment that you would like to run nightly job, then when I merge your PR, I will also run the nightly build.

mthrok:
Triggered the nightly job manually

https://app.circleci.com/pipelines/github/pytorch/audio?branch=nightly

mthrok:
Looks like ssh needs to be told that it's okay to access github.com.

```
+ git push -u origin gh-pages
The authenticity of host 'github.com (140.82.114.3)' can't be established.
RSA key fingerprint is SHA256:nThbg6kXUpJWGl7E1IGOCspRomTxdCARLviKw6E5SY8.
RSA key fingerprint is MD5:16:27:ac:a5:76:28:2d:36:63:1b:56:4d:eb:df:a6:48.
Are you sure you want to continue connecting (yes/no)? 
```

Something like this should help https://stackoverflow.com/a/28527476

PR #1099, opened 2020-12-17
add warp transducer

vincentqb:
This pull request introduces `rnnt_loss` and `RNNTLoss` in `torchaudio.prototype.transducer` using [HawkAaron's warp-transducer](https://github.com/HawkAaron/warp-transducer).
- The python interface remains currently the same as the original.
- Update `cmdclass` to use custom clean [here](https://github.com/pytorch/audio/blob/master/setup.py#L85).
- The library is compiled statically, e.g. vincentqb/warp-transducer#2.
- Add a unit test for CPU.

#### Import the warp-transducer repository

Suggested options to load the repository with torchbind replacing pybind:
- (current, load as a submodule with bindings and CMakeList copied) Load original repository and bind outside of repository.
- Load original repository and [patch](https://patch-diff.githubusercontent.com/raw/vincentqb/warp-transducer/pull/1.patch).
- Load a fork like vincentqb/warp-transducer#1 as a submodule.
- `git subtree` as suggested in [comment](https://github.com/pytorch/audio/pull/1063#issuecomment-743346592).
- Submit an upstream patch

#### CUDA

This pull request can optionally build with GPU support. Environment variables need to be set before installing, as mentioned [here](https://github.com/HawkAaron/warp-transducer/issues/15#issuecomment-467668750) and [colab notebook](https://colab.research.google.com/drive/1jtCthsRthKINFMDhOM_hJUnK_DQSOeot?usp=sharing).
```bash
export CUDA_HOME=/usr/local/cuda
export CUDA_TOOLKIT_ROOT_DIR=$CUDA_HOME
export LD_LIBRARY_PATH="$CUDA_HOME/extras/CUPTI/lib64:$LD_LIBRARY_PATH"
export LIBRARY_PATH=$CUDA_HOME/lib64:$LIBRARY_PATH
export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
export CFLAGS="-I$CUDA_HOME/include $CFLAGS"
BUILD_CUDA_WT=ON python setup.py install
```

There is also a a unit test available for GPU.

cc @astaff, [internal](https://fb.quip.com/tzrDA3DHqlY8)


build_tools/setup_helpers/extension.py
```diff
@@ -116,27 +120,86 @@ def _build_third_party():
...
+    print("path for object:", rel_warp_rnnt_path)
+    return CppExtension(
+        name='_warp_transducer',
+        sources=[os.path.realpath(base_path / 'pytorch_binding' / 'src' / 'binding.cpp')],
+        # sources=[os.path.realpath(_TP_BASE_DIR / 'binding.cpp')],
+        include_dirs=include_dirs,
+        extra_objects=extra_objects,
+        library_dirs=[os.path.realpath(warp_rnnt_path)],
+        libraries=librairies,
+        extra_link_args=['-Wl,-rpath,' + rel_warp_rnnt_path],
```
vincentqb, line 182: The linux/macos builds are not finding the shared library, see [here](https://app.circleci.com/pipelines/github/pytorch/audio/4185/workflows/53e9a945-8248-4283-a0f9-290f10ebd0fa/jobs/129878
):
```
OSError: libwarprnnt.so: cannot open shared object file: No such file or directory
```
though I do find it (libwarprnnt.so) when sshing in the machine:
```
./opt/conda/conda-bld/torchaudio_1608332828661/work_moved_torchaudio-0.8.0.dev20201218-py38_linux-64/third_party/warp_transducer/build/libwarprnnt.so
```
The suggestion I saw does say to specify `rpath`, see [doc](https://docs.conda.io/projects/conda-build/en/latest/resources/use-shared-libraries.html) or [here](https://github.com/conda-forge/cmake-feedstock/issues/21#issuecomment-256090118). I haven't had success with this though. Thoughts?

hey, quick question: i'm trying to build a second C++ extension for torchaudio, and I'm running into an issue linking for conda, but not for the rest. when i ssh in the machine, i find the library (libwarprnnt.so). https://app.circleci.com/pipelines/github/pytorch/audio/4185/workflows/53e9a945-8248-4283-a0f9-290f10ebd0fa/jobs/129878 -- thoughts?

seemethere: cc @malfet for this as well

.circleci/unittest/linux/scripts/install.sh
```diff
@@ -38,6 +38,7 @@ conda install -y -c "pytorch-${UPLOAD_CHANNEL}" pytorch ${cudatoolkit}
 
 # 2. Install torchaudio
 printf "* Installing torchaudio\n"
+git submodule update --init --recursive
```
vincentqb, line 41: I'd prefer initializing and updating the submodules via the `git clone` command directly. Pointer on best way to do this?

build_tools/setup_helpers/extension.py
cpuhrsch: Are you sure the binary is actually in the right location upon creation of the conda package? That is, is libwarprnnt.so being moved / included in the move into the folder used to build the conda package?

mthrok: So the `libwarprnnt.so` is found in `third_party/warp_transducer/submodule/build/libwarprnnt.so`.

This explains why `conda` package cannot find it. Because this is not part of the package data.
Currently in torchaudio, when binding libsox, we do static linking, meaning we build libsox then statically link it to C++ extension (_torchaudio.so). So this problem does not happen.

The solution is either
1. link libwarprnnt statically 
2. think of a way (and a convention to be followed in the future) where to put dynamic library and how to add it to library look up path at runtime.

mthrok: If `libwarprnnt` allows static build, then it's easy to resolve

.circleci/unittest/linux/scripts/install.sh
mthrok: You need to provide custom cloning command in place of pre-defined `checkout` command in config file.
Though I do not see benefit of maintaining custom command over the maintaining cost if it's about preference.

build_tools/setup_helpers/extension.py
vincentqb: yeah, I'll just build statically

setup.py
```diff
@@ -83,7 +84,8 @@ def run(self):
     packages=find_packages(exclude=["build*", "test*", "torchaudio.csrc*", "third_party*", "build_tools*"]),
     ext_modules=setup_helpers.get_ext_modules(),
     cmdclass={
-        'build_ext': setup_helpers.BuildExtension.with_options(no_python_abi_suffix=True)
+        'build_ext': setup_helpers.BuildExtension.with_options(no_python_abi_suffix=True),
+        'clean': clean,
```
cpuhrsch, line 88: Why is this needed?

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -75,3 +75,21 @@ def skipIfNoExtension(test_item):
...
+
+
+skipIfNoTransducer = unittest.skipIf(
+    not is_module_available('_warp_transducer'),
+    '"_warp_transducer" is not available',
+)
+
+try:
+    torch.ops.warprnnt_pytorch_warp_rnnt.gpu_rnnt
+    _CUDA_TRANSDUCER = True
```
cpuhrsch, line 87: I'd make this switch dependent on whether CUDA [is available](https://pytorch.org/docs/master/cuda.html#torch.cuda.is_available) at all.

PR #1101, opened 2020-12-17
VCTK Using Path and glob instead of walk_files 

krishnakalyan3:
Remove walk_files from VCTK dataset

Fixes:
#1051
#1069

### DIFF
--- a/torchaudio/datasets/vctk.py
+++ b/torchaudio/datasets/vctk.py
@@ -9,7 +9,6 @@ from torch.utils.data import Dataset
 from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
-    walk_files
 )
 
 URL = "https://datashare.is.ed.ac.uk/bitstream/handle/10283/3443/VCTK-Corpus-0.92.zip"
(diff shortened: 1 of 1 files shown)


krishnakalyan3:
This PR was not tested as this dataset is no longer available.

torchaudio/datasets/vctk.py
```diff
@@ -123,9 +122,7 @@ def __init__(self,
                 "with `download=True` to donwload the latest version."
             )
 
-        walker = walk_files(
-            self._path, suffix=self._ext_audio, prefix=False, remove_suffix=True
-        )
+        walker = sorted(str(p.stem) for p in Path(self._path).glob('*/*/*' + self._ext_audio))
```
mthrok, line 125: Can you change the pattern to `'**/*' + self._ext_audio`?

That should match the behavior of the original implementation.

mthrok (commented):
Looks good. Please do the honor to delete the implementation of `walk_files`.

mthrok:
Thanks! @krishnakalyan3

PR #1104, opened 2020-12-18
Add format override to load and related I/O functions

### LINKED ISSUE #1040: libsox does not detect mp3 when extension is missing

mthrok:
This PR adds `format` argument to the following functions;

* `sox_io_backend.info`
* `sox_io_backend.load`
* `sox_effects.apply_effects_file`
* `_soundfile.load` (not used as [`PySoundfile` does not support format hinting]([[src](https://github.com/bastibe/python-soundfile/blob/e4a176255c462713cd25d47b7c855b015d749e7f/soundfile.py#L1433-L1436)]))

The background of this change is that there are certain MP3 files that libsox cannot open without format hinting. (See #1040)

With this change, #1042 is no longer necessary so 69d716c9 is reverted too.

Resolves #1040


torchaudio/backend/sox_io_backend.py
```diff
@@ -9,20 +9,27 @@
...
+        format: Optional[str] = None,
+) -> AudioMetaData:
     """Get signal information of an audio file.
 
     Args:
         filepath (str or pathlib.Path):
             Path to audio file. This function also handles ``pathlib.Path`` objects,
             but is annotated as ``str`` for TorchScript compatibility.
+        format (str, optional):
+            Hint libsox about the audio format. This argument helps libsox detect the audio
```
mthrok, line 23: Verify if this is fallback or overwrite, and update the docstring.

torchaudio/sox_effects/sox_effects.py
```diff
@@ -180,6 +181,10 @@ def apply_effects_file(
             than integer WAV type.
         channels_first (bool): When True, the returned Tensor has dimension ``[channel, time]``.
             Otherwise, the returned Tensor's dimension is ``[time, channel]``.
+        format (str, optional):
+            Hint libsox about the audio format. This argument helps libsox detect the audio
```
vincentqb, line 185: Clarify what "hint" means here?

torchaudio/backend/_soundfile_backend.py
```diff
@@ -99,6 +100,8 @@ def load(
         channels_first (bool):
             When True, the returned Tensor has dimension ``[channel, time]``.
             Otherwise, the returned Tensor's dimension is ``[time, channel]``.
+        format (str, optional):
+            Not used. PySoundFile does not accept format hint.
```
cpuhrsch, line 104: This makes me wonder whether it's time to introduce a top-level dispatch instead of using function pointers. Concretely right now we're doing

```torch.load = torchaudio.backend.sox_io_backend.load```

instead we could do

```
# torchaudio/__init__.py
def load(path):
    format = detect_format(path)
    [select backend]
    return backend.load(path, format)
```

vincentqb (changes requested):
LGTM

vincentqb (approved):
LGTM

cpuhrsch (approved):
As a follow-up I suggest we introduce the top-level dispatch and abstract away the format detection.

torchaudio/backend/sox_io_backend.py
mthrok: Okay, it's overwrite. https://github.com/dmkrepo/libsox/blob/master/src/formats.c#L475-L481

Issue #754, opened 2020-07-01
Add support for file-like object


liuyuyuil:
## 🚀 Feature

Add support to directly read file-like object for torchaudio.load , e.g. BytesIO.

mthrok:
Hi @liuyuyuil 

What type of data do you expect to be in the file-like object?
Is it raw samples or decoded/compressed?

liuyuyuil:
> Hi @liuyuyuil
> 
> What type of data do you expect to be in the file-like object?
> Is it raw samples or decoded/compressed?

It is just the same as current `torchaudio.load` function supports. Right now, the first parameter of `load `function is `filepath`, which can only be the audio file path from the disk, currently I want to load audio from cloud-storage, such as S3 on AWS and OSS on AlibabaCloud, I have to download my audio files from remote into temporary files before loading.

Can we access the audio file like this:
```
from io import BytesIO
fp = bucket.get_object('xxx/xxx.wav')
d, sr = torchaudio.load(BytesIO(fp.read()), **)
```

mthrok:
> It is just the same as current torchaudio.load function supports

That's very broad. In such case, loading audio data involves detection of audio format then decoding process, which we do not implement by ourselves but rely on widely used external C++ libraries (such as `libsox` and `libsndfile`). We cannot pass `BytesIO` directly to them.

If you know that your files are wav format, then you can use [`scipy.io.wavfile.read`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.read.html) or [`soundfile.read`](https://pysoundfile.readthedocs.io/en/latest/#soundfile.read) to get NumPy representation of data from `BytesIO`, and convert it to Tensor type with `torch.from_numpy`.

mthrok:
I have found a way to pass file-like object to libsox. #1108

mthrok:
Support for file-like object has been added in #1158

PR #1108, opened 2020-12-19
Support file-like object in info

mthrok:
This PR adds support for file-like object to `info` function of `sox_io` backend and `soundfile` backends.

See also #1115

### DIFF
--- a/torchaudio/csrc/sox/io.cpp
+++ b/torchaudio/csrc/sox/io.cpp
@@ -36,7 +36,7 @@ int64_t SignalInfo::getBitsPerSample() const {
   return bits_per_sample;
 }
 
-c10::intrusive_ptr<SignalInfo> get_info(
+c10::intrusive_ptr<SignalInfo> get_info_file(
     const std::string& path,
     c10::optional<std::string>& format) {
   SoxFormat sf(sox_open_read(
@@ -148,6 +148,56 @@ void save_audio_file(
 }
 
 #ifdef TORCH_API_INCLUDE_EXTENSION_H
+
+std::tuple<int64_t, int64_t, int64_t, int64_t> get_info_fileobj(
+    py::object fileobj,
+    c10::optional<std::string>& format) {
+  // Prepare in-memory file object
+  // When libsox opens a file, it also reads the header.
+  // When opening a file there are two functions that might touch FILE* (and the
+  // underlying buffer).
+  // * `auto_detect_format`
+  //   https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/formats.c#L43
+  // * `startread` handler of detected format.
+  //   https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/formats.c#L574
+  // To see the handler of a particular format, go to
+  //   https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/<FORMAT>.c
+  // For example, voribs can be found
+  //   https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/vorbis.c#L97-L158
+  //
+  // `auto_detect_format` function only requires 256 bytes, but format-dependant
+  // `startread` handler might require more data. In case of vorbis, the size of
+  // header is unbounded, but typically 4kB maximum.
+  //
+  // "The header size is unbounded, although for streaming a rule-of-thumb of
+  // 4kB or less is recommended (and Xiph.Org's Vorbis encoder follows this
+  // suggestion)."
+  //
+  // See:
+  // https://xiph.org/vorbis/doc/Vorbis_I_spec.html
+  auto capacity = 4096;
+  std::string buffer(capacity, '\0');
+  auto* buf = const_cast<char*>(buffer.data());
+  auto num_read = read_fileobj(&fileobj, capacity, buf);
+  // If the file is shorter than 256, then libsox cannot read the header.
+  auto buf_size = (num_read > 256) ? num_read : 256;
+
+  SoxFormat sf(sox_open_mem_read(
+      buf,
+      buf_size,
+      /*signal=*/nullptr,
+      /*encoding=*/nullptr,
+      /*filetype=*/format.has_value() ? format.value().c_str() : nullptr));
+
+  // In case of streamed data, length can be 0
+  validate_input_file(sf, /*check_length=*/false);
+
+  return std::make_tuple(
+      static_cast<int64_t>(sf->signal.rate),
+      static_cast<int64_t>(sf->signal.channels),
+      static_cast<int64_t>(sf->signal.length / sf->signal.channels),
+      static_cast<int64_t>(sf->encoding.bits_per_sample));
+}
 
 std::tuple<torch::Tensor, int64_t> load_audio_fileobj(
     py::object fileobj,
--- a/test/torchaudio_unittest/soundfile_backend/info_test.py
+++ b/test/torchaudio_unittest/soundfile_backend/info_test.py
@@ -1,5 +1,6 @@
 from unittest.mock import patch
 import warnings
+import tarfile
 
 import torch
 from torchaudio.backend import _soundfile_backend as soundfile_backend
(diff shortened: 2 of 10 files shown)


cpuhrsch:
In case you don't like file-like object, file-object is actually defined to be a synonym in the python glosassry. Same with stream. https://docs.python.org/3/glossary.html#term-file-object

cpuhrsch:
I'm a fan. Using pybind11 in eager mode and torchbind for deployment only really is opening a lot of doors. What's the error behavior when someone is using a pybind11 feature that torchbind does not support? For now we've been using this dual registration approach for performance reasons since pybind11 has lower eagermode overhead.

vincentqb:
- I'm assuming pybind is used since torchbind does not support bytes object, is that correct?
- Are you planning to expose this to `torchaudio.load`? There is no test there.
- Is this ready for review? It is currently marked as draft.

vincentqb:
> I'm a fan. Using pybind11 in eager mode and torchbind for deployment only really is opening a lot of doors. What's the error behavior when someone is using a pybind11 feature that torchbind does not support? For now we've been using this dual registration approach for performance reasons since pybind11 has lower eagermode overhead.

Ah, so it's for performance reason then? I know we observe performance difference in some context, but did we measure a significant difference here? There's a note in the description about being ok with low performance in the description for part of the PR.

torchaudio/csrc/sox_utils.cpp
```diff
@@ -99,9 +99,6 @@ void validate_input_file(const SoxFormat& sf) {
   if (sf->encoding.encoding == SOX_ENCODING_UNKNOWN) {
     throw std::runtime_error("Error loading audio file: unknown encoding.");
   }
-  if (sf->signal.length == 0) {
-    throw std::runtime_error("Error reading audio file: unkown length.");
-  }
```
vincentqb, line 104: This error is removed in preparation of streaming, or because the length of the signal when visiting the bytes path is zero?

cpuhrsch:
> > I'm a fan. Using pybind11 in eager mode and torchbind for deployment only really is opening a lot of doors. What's the error behavior when someone is using a pybind11 feature that torchbind does not support? For now we've been using this dual registration approach for performance reasons since pybind11 has lower eagermode overhead.
> 
> Ah, so pybind is picked for performance reason then? I know we observe performance difference in some context, but did we measure a significant difference here? There's a note in the description about being ok with low performance in the description for part of the PR.

It's been picked for performance reasons in the context of torchtext where the dual pybind11/torchbind approach originated.

Using it for the purposes of getting eagermode-only features is new.

mthrok:
@cpuhrsch 

> In case you don't like file-like object, file-object is actually defined to be a synonym in the python glosassry. Same with stream. https://docs.python.org/3/glossary.html#term-file-object

well, actually, I prefer the notion of file-`"like"`. Because in my personal experience, more often it's not a file but a something else like data streamed over the network. (still, I guess one can claim it's a file object on raw level Linux point of view)

> I'm a fan. Using pybind11 in eager mode and torchbind for deployment only really is opening a lot of doors. What's the error behavior when someone is using a pybind11 feature that torchbind does not support? For now we've been using this dual registration approach for performance reasons since pybind11 has lower eagermode overhead.

So the behavior is,
* if the function is being scripted, then it directly tries to call function bound via TorchScript. In this case, passing anything other than string object is a failure, because of TS limitation.
* if the function is not being scripted, it inspects the input argument, then based on the type, it uses either TS binding (for `str`) or PyBind11 version (`bytes`). And in this case, if the input type is file-like object, then perform `read()` to fetch the data as `bytes` then pass it to PyBind11 version.

mthrok:
> Using it for the purposes of getting eagermode-only features is new.

And it's a long-waited feature

#800 #754

mthrok:
@vincentqb 

> * I'm assuming pybind is used since torchbind does not support bytes object, is that correct?

yes

> * Are you planning to expose this to `torchaudio.load`? There is no test there.

It is tested. Checkout the tests added.

torchaudio/csrc/sox_utils.cpp
mthrok: It's the later. For container formats, which does not have a length information in header, these values become zero.

vincentqb:
> It is tested. Checkout the tests added.

I meant: there are tests for the load functions within both sox_io and soundfile backends, but not one calling directly the `torchaudio.load` function. Did I miss it? If we're confident this covers what we need and will cover what is needed of a new backend in the future, then what is there already in the pull request would be enough.

mthrok:
> > > * Are you planning to expose this to `torchaudio.load`? There is no test there.
> > 
> > 
> > It is tested. Checkout the tests added.
> 
> I meant: there are tests for the load functions within both sox_io and soundfile backends, but not one calling directly the `torchaudio.load` function. Did I miss it? If we're confident this covers what we need and will cover what is needed of a new backend in the future, then what is there already in the pull request would be enough.

Check out how the backend tests are written. `torchaudio.load` is not the right way to test.

cpuhrsch:
> So the behavior is,
> 
> * if the function is being scripted, then it directly tries to call function bound via TorchScript. In this case, passing anything other than string object is a failure, because of TS limitation.
> * if the function is not being scripted, it inspects the input argument, then based on the type, it uses either TS binding (for `str`) or PyBind11 version (`bytes`). And in this case, if the input type is file-like object, then perform `read()` to fetch the data as `bytes` then pass it to PyBind11 version.

Makes sense. We could either as a follow-up or using this as a vehicle look into how we can improve the error message, if necessary, when someone tries to convert a model into TorchScript that calls these functions using bytes.

torchaudio/csrc/sox_utils.cpp
vincentqb: thx for clarifying!

mthrok:
> > So the behavior is,
> > 
> > * if the function is being scripted, then it directly tries to call function bound via TorchScript. In this case, passing anything other than string object is a failure, because of TS limitation.
> > * if the function is not being scripted, it inspects the input argument, then based on the type, it uses either TS binding (for `str`) or PyBind11 version (`bytes`). And in this case, if the input type is file-like object, then perform `read()` to fetch the data as `bytes` then pass it to PyBind11 version.
> 
> Makes sense. We could either as a follow-up or using this as a vehicle look into how we can improve the error message, if necessary, when someone tries to convert a model into TorchScript that calls these functions using bytes.


Well, once the function is scripted, we cannot do much about it, because TorchScript runtime won't run the function if the input type is invalid. However it seems that the error message is clear if the reason for the failure is wrong type.

<details><summary> foo.py </summary>

```python
import torch
import torchaudio

path = 'test/torchaudio_unittest/assets/sinewave.wav'


# eager execution - works
with open(path, 'rb') as file_:
    torchaudio.load(file_)


load = torch.jit.script(torchaudio.load)

# jit execution with str - works
load(path)

# jit execution with file-like obj - does not work.
with open(path, 'rb') as file_:
    load(file_)
```

</details>


```
  File "foo.py", line 10, in <module>
    load(file_)
RuntimeError: load() Expected a value of type 'str' for argument 'filepath' but instead found type 'BufferedReader'.
Position: 0
Value: <_io.BufferedReader name='test/torchaudio_unittest/assets/sinewave.wav'>
Declaration: load(str filepath, int frame_offset=0, int num_frames=-1, bool normalize=True, bool channels_first=True, str? format=None) -> ((Tensor, int))
Cast error details: Unable to cast Python instance to C++ type (compile in debug mode for details)
```

mthrok:
@cpuhrsch Update the docstring to mention the difference between scripted/eager mode.

igorgad:
Hi @mthrok, I wonder if the parameters `num_frames` and `offset` could be used with network streams and if it would download only the requested frames?
Thanks

mthrok:
> Hi @mthrok, I wonder if the parameters `num_frames` and `offset` could be used with network streams and if it would download only the requested frames?
> Thanks

Hi @igorgad 

Thanks for the question, the answer is (unfortunately) no. The goal of this PR is to extend the support to Python's file object, which is the protocol `requests` happen to implement, thus it works for data transferred over the network. There is no specific optimization or operation carried out for online data transfer.

Then the question is, can we (relatively easily) extend the support for such efficient network data transfer? I think the answer is no, because of the limitation `libsox` poses. In my understanding, `libsox` works on `FILE` object pointer in  C level. (which is abstracted away from user API of `libsox`) You can use some URL with `sox` command but that's just calling `wget` in subprocess and piping the data to `stdin` of the process `sox` command is running. There is no network utility `libsox` utilizes, and plugging such library into `libsox` is not trivial.

Maybe `ffmpeg` has that kind of capability, but currently `torchaudio` does not bind `ffmpeg`, though it's in my wish list. (but we do not have an action plan for it)

If you want that feature, maybe you can file an issue with feature request and provide your workflow?  (total length of the audio file and the protocol you are using to fetch the data) Though I cannot guarantee that we can full fill your request, but it will certainly help us understand your demand.

igorgad:
Yes. Indeed I think that plugging a network utility into libsox isn't the best option. I investigated the use of FFmpeg to load files from GCS buckets, but it doesn't work unless you set a public URL.

I'm currently investigating options to scale the training of our models on a managed cloud infrastructure, and one of the options is the Ai-Platform running custom containers.  The problem is that we cannot mount NFS shares with our data on the containers due to the lack of the privileged run mode. The dataset is also big to copy from a remote path at the beginning of each experiment. 

This way, and inspired by [this](https://github.com/pytorch/audio/issues/800#issuecomment-661661511) implementation, would you consider a `cloud_io` backend (probably with [fsspec](https://filesystem-spec.readthedocs.io/en/latest/api.html?highlight=read#fsspec.spec.AbstractFileSystem.read_block)) using `miniaudio` a good solution to access files directly and efficiently from buckets? Of course, it requires some tinkering to translate `audio_frames` into file offsets but I think that's relatively easy for wav files.

Thanks

mthrok:
> Yes. Indeed I think that plugging a network utility into libsox isn't the best option. I investigated the use of FFmpeg to load files from GCS buckets, but it doesn't work unless you set a public URL.
> 
> I'm currently investigating options to scale the training of our models on a managed cloud infrastructure, and one of the options is the Ai-Platform running custom containers. The problem is that we cannot mount NFS shares with our data on the containers due to the lack of the privileged run mode. The dataset is also big to copy from a remote path at the beginning of each experiment.
> 
> This way, and inspired by [this](https://github.com/pytorch/audio/issues/800#issuecomment-661661511) implementation, would you consider a `cloud_io` backend (probably with [fsspec](https://filesystem-spec.readthedocs.io/en/latest/api.html?highlight=read#fsspec.spec.AbstractFileSystem.read_block)) using `miniaudio` a good solution to access files directly and efficiently from buckets? Of course, it requires some tinkering to translate `audio_frames` into file offsets but I think that's relatively easy for wav files.
> 
> Thanks

@igorgad 

Your question made me realized that I did not test on those frame slicing. While I was adding test, I looked into the read behavior, and turned out that `libsox` is smart enough to stop reading once it read enough samples for `num_frames` and `frame_offset` options. I think it still needs to read until the beginning of the frames.

Regarding `cloud_io` backend, I once had a conversation with @cpuhrsch about adding specific I/O functions for different data source like database, S3 but we do not have a concrete plan for adding it yet. We are considering getting rid of the notion of backend and moving to format-based automatic backend selection. From there we can add I/O functions that are specific to format or data source.

Giving some thoughts on this, I think the process can be broken down into the following steps;
1. Query a header and detect the format
2. Compute the byte position using the format information from 1.
3. Fetch the corresponding part of the data
4. Perform decoding

Then here are the questions that came up to me;
1. How can we perform efficient data transfer?
Looking at AWS S3 documentation, they accept `range` header so it is possible to download only the related part, but I guess it has to initiate a new connection. I do not have experience in other cloud storage but if the cloud vendor accepts `range` header or some equivalent, this should be possible.
2. How can we compute bytes range?
I think there are two potential questions to be answer
    1. Can you always get a header?
I am not an expert in audio formats, but I think there are formats that do not have a global header, in this case, it's hard to get the understanding of the whole file from the beginning part of the file.
    1. Can we always get the precise byte range?
I think there are formats that support variable bit rate. In this case, it will be difficult to know the byte range from header.

As you mention, if one knows that the format is WAV, then solving 2 will be easy, but solving it for general case will be difficult. Also with cloud provider, one has to think of a way to pass access tokens to the object that is making the network access. So I think, for now, to solve your problem, having your own custom implementation is the easiest solution.

torchaudio/csrc/sox/io.cpp
```diff
@@ -149,6 +149,56 @@ void save_audio_file(
...
+  // header is unbounded, but typically 4kB maximum.
+  //
+  // "The header size is unbounded, although for streaming a rule-of-thumb of
+  // 4kB or less is recommended (and Xiph.Org's Vorbis encoder follows this
+  // suggestion)."
+  //
+  // See:
+  // https://xiph.org/vorbis/doc/Vorbis_I_spec.html
+  auto capacity = 4096;
+  std::string buffer(capacity, '\0');
```
datumbox, line 179: Same as other PR. Perhaps there is a better type buffer than std:string here?

datumbox (approved):
LGTM! I left only a non-blocking comment.

torchaudio/csrc/sox/io.cpp
mthrok: https://github.com/pytorch/audio/pull/1181#discussion_r565473577

PR #1110, opened 2020-12-19
Fix `sox_effects` test path in docs

jcaw:
Docs seem to have an old path to the `sox_effects` tests. PR updates them.


mthrok (approved):
Good catch. Thanks for the contribution.

PR #1113, opened 2020-12-20
Fix `amplitude_to_DB` clamping behaviour on batches

### LINKED ISSUE #994: transforms.AmplitudeToDB does not handle cut-off correctly for multi-channel or batched data
## 🐛 Bug

From my understanding (based on e.g. #337), all transforms should be able to operate on tensors with dimensions `(batch, channels, ...)` or `(channels, ...)`, with `...` being dependent on the type of data being processed, e.g.`time` for waveform data and `freq, time` for spectrograms. In this way, we can pass multiple chunks of data (waveforms, spectrograms…) at once and expect to get the same results as if we would pass them one by one.

However, this is not the case `transforms.AmplitudeToDB`: As easily traceable in the [source code of the corresponding functional](https://pytorch.org/audio/_modules/torchaudio/functional.html#amplitude_to_DB), this transform blindly operates on the passed tensor without taking its dimensionality and the related semantics into account in any way.

This becomes a problem in the calculation of the cut-off. The purpose of this step is to clamp low dB values to a minimum value some fixed amount of decibels below the maximum value _in the respective spectrogram_. However, `amplitude_to_db` uses the single _global_ maximum in the passed tensor to calculate the cut-off for _all_ contained spectrograms. Thus, when passing batched data, the results for one spectrogram are dependent on all the other spectrograms in the same batch, which to my understanding is not the correct behavior. My conclusion is that `AmplitudeToDB` silently outputs wrong (in the sense of `transforms` general interface contract) data for batched or multi-channel data, which I would consider really dangerous for applications.

Ideally, this should be fixed directly in `functional.amplitude_to_DB`, so that we can also pass batched data there.

## Environment
TorchAudio 0.7.0

jcaw:
`amplitude_to_DB` currently clamps per-batch, but it should clamp per-item. I've modified it to clamp per-item (when a batch is provided) and I've modified the MFCC transform to take advantage of this new behavior. Tests for `amplitude_to_DB` are included but I've added no new tests for the modified MFCC transform.

This is just an initial draft. In #994 @vincentqb specified that it should always expect a tensor of shape `(..., freq, time)`, so this implementation assumes the input is a spectrogram and determines whether it's a batch based on the number of dimensions (it assumes a batch when there are more than three dimensions, i.e. more than `(channel, freq, time)`). I'm not sure if this is the most sensible solution. It restricts the inputs to spectrograms and only allows batches with 4 (or more) dimensions. A batch with shape `(item, freq, time)` would be treated as a single item. This does seem contradictory if the specified input shape is `(..., freq, time)`.

I also have a (rough) branch [here](https://github.com/jcaw/audio/tree/amplitude_to_db_batch_fix__flag) which takes a `batch` flag to differentiate. Alternatively, batchwise conversions could be pulled into a separate method. (Perhaps `amplitude_to_DB_iid`?)

Closes #994


jcaw:
(The tests in here are a little rough since the infrastructure is due to be replaced)

jcaw:
The torchscript compilation tripped me up but I'm having trouble getting something that works. [This line](https://github.com/pytorch/audio/pull/1113/commits/d3ad3cf7104e3c467b8c3e4a3d990e53fb01607a#diff-f69839488ce70c39d8c3db8c87790ac963364faf8aefcb38ff5b026ee9dca8c7R257) doesn't want to compile - . 

Is there a way to either add multiple singleton dimensions to the right like this, or to somehow broadcast from the opposite direction to normal (aligning to the left, instead of the right)?

I've got it working locally doing this, but I don't like it:

```python
        if x_db.dim() > 3:
            flat_shape = x_db.size(0), -1
            db_floors = x_db.reshape(flat_shape).amax(dim=1) - top_db
            x_db = torch.max(x_db.view(flat_shape), db_floors.unsqueeze(1)).view(x_db.shape)

```

(Is there a document I can reference for full contribution guidelines?)

torchaudio/functional/functional.py
```diff
@@ -248,7 +251,13 @@ def amplitude_to_DB(
...
 
     if top_db is not None:
-        x_db = x_db.clamp(min=x_db.max().item() - top_db)
+        dims = x_db.dim()
+        if dims > 3:
+            db_floors = x_db.reshape(x_db.size(0), -1).amax(dim=1) - top_db
+            broadcast_shape = (...,) + (None,)*(dims-1)
+            x_db = torch.max(x_db, db_floors[broadcast_shape])
+        else:
+            x_db = x_db.clamp(min=x_db.max().item() - top_db)
```
vincentqb, line 260: My suggestion is to replace directly clamp by `min and max` per batch which would simply this code.

torchaudio/functional/functional.py
```diff
@@ -248,7 +251,13 @@ def amplitude_to_DB(
     x_db -= multiplier * db_multiplier
 
     if top_db is not None:
-        x_db = x_db.clamp(min=x_db.max().item() - top_db)
+        dims = x_db.dim()
+        if dims > 3:
```
vincentqb, line 255: Here's an [example](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L256) on how to pack and unpack a tensor from (..., channel, freq, time) to (batch, channel, freq, time) and back again to (..., channel, freq, time) in a torchscriptable manner.

vincentqb (commented):
About the shape: in practice, saying `(..., freq, time)` means we expect `(freq, time)`, `(channel, freq, time)`, `(batch, freq, time)`, `(batch, channel, freq, time)`, etc. The ambiguity is in the 3 dimensions as you pointed out: `(batch, freq, time)` vs `(channel, freq, time)`.

My suggestion is to support the original behavior for  `(freq, time)`, `(channel, freq, time)` and add documentation. We then extend `(batch, channel, freq, time)` or `(..., channel, freq, time)` to do clamping "per `(channel, freq, time)`. Therefore, we do not support directly the case `(batch, freq, time)`, though someone (outside the function call) could easily `unsqueeze` to add a 1 channel.

> The torchscript compilation tripped me up but I'm having trouble getting something that's compatible. [This line](https://github.com/pytorch/audio/pull/1113/commits/d3ad3cf7104e3c467b8c3e4a3d990e53fb01607a#diff-f69839488ce70c39d8c3db8c87790ac963364faf8aefcb38ff5b026ee9dca8c7R257) doesn't want to compile - it's giving me `RuntimeError: Cannot emit expr for: (dots)`. The `Ellipsis` constant also fails.
> 
> Is there a way to either add multiple singleton dimensions to the right like this, or to somehow broadcast from the opposite direction to normal (aligning to the left, instead of the right)?

Do you mean to add a dimension like `(batch, 1, freq, time)` with `unsequeeze`? 

> (Is there also a document I can reference for full contribution guidelines, so I can run the CI locally?)

The tests detailed [here](https://github.com/pytorch/audio/tree/master/test) can be run using pytest. Is that what you meant?

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,47 +70,104 @@ def test_pitch(self, frequency):
...
-        x = torch.rand(1000)
+class Testamplitude_to_DB(common_utils.TorchaudioTestCase):
+    AMPLITUDE_MULT = 20.
+    POWER_MULT = 10.
+    AMIN = 1e-10
+    REF = 1.0
+    # FIXME: So the DB multiplier is 0?
+    DB_MULT = math.log10(max(AMIN, REF))
+
+    def _make_spectrogram(self, items, channels, wave_samples, scale=1.):
```
mthrok, line 81: This method does not seem to reference the class or an instance. Please make it a plain function. (if you end up keeping this)

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,47 +70,104 @@ def test_pitch(self, frequency):
...
 
-class TestDB_to_amplitude(common_utils.TorchaudioTestCase):
-    def test_DB_to_amplitude(self):
-        # Make some noise
-        x = torch.rand(1000)
+class Testamplitude_to_DB(common_utils.TorchaudioTestCase):
+    AMPLITUDE_MULT = 20.
+    POWER_MULT = 10.
+    AMIN = 1e-10
+    REF = 1.0
```
mthrok, line 77: Even if the same parameters are used, please define them in each test case separately, so that test cases are independent.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,47 +70,104 @@ def test_pitch(self, frequency):
...
+        db = F.amplitude_to_DB(spec, self.POWER_MULT, self.AMIN, self.DB_MULT, top_db=None)
+        x2 = F.DB_to_amplitude(db, self.REF, 1.)
 
         torch.testing.assert_allclose(x2, spec, atol=5e-5, rtol=1e-5)
 
+    def test_amplitude_to_DB(self):
+        spec = self._make_spectrogram(2, 2, 1000, scale=0.5)
+        # Ensure it works on both batches and items
+        self._ensure_reversible(spec)
+        self._ensure_reversible(spec[0])
```
mthrok, line 105: Please separate these tests so that the later test will be executed even when the first fails.

test/torchaudio_unittest/functional/functional_cpu_test.py
mthrok: Also looking at the individual test case, the resulting spectrograms are modified. That makes me wonder if it is worth introducing such helper function.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,47 +70,104 @@ def test_pitch(self, frequency):
...
 
+    def _ensure_reversible(self, spec):
+        """Check `amplitude_to_db` returns the original when reversed."""
         # Spectrogram amplitude -> DB -> amplitude
-        db = F.amplitude_to_DB(spec, multiplier, amin, db_multiplier, top_db=None)
-        x2 = F.DB_to_amplitude(db, ref, power)
+        db = F.amplitude_to_DB(spec, self.AMPLITUDE_MULT, self.AMIN, self.DB_MULT, top_db=None)
+        x2 = F.DB_to_amplitude(db, self.REF, 0.5)
 
         torch.testing.assert_allclose(x2, spec, atol=5e-5, rtol=1e-5)
```
mthrok, line 93: Please replace `torch.testing.assert_allclose` with `self.assertEqual`. The use of `assert_allclose` is discouraged now.

test/torchaudio_unittest/functional/functional_cpu_test.py
jcaw: Would you like me to do this for `test_top_db` too?

```python
        decibels = F.amplitude_to_DB(spec[0], self.AMPLITUDE_MULT, self.AMIN,
                                     self.DB_MULT, top_db=top_db)
        # The actual db floor will be just below 6.0206 - use 6.0205 to deal
        # with rounding error.
        above_top = decibels >= 6.0205
        assert above_top.all(), decibels

        # And check it works with batch dimension
        decibels_batch = F.amplitude_to_DB(spec, self.AMPLITUDE_MULT, self.AMIN,
                                           self.DB_MULT, top_db=top_db)
        above_top_batch = decibels_batch >= 6.0205
        assert above_top_batch.all(), decibels_batch
```

jcaw:
Awesome, thanks. I'll implement these suggestions now.

> Do you mean to add a dimension like (batch, 1, freq, time) with unsequeeze?

More like a version of `unsqueeze` that can add an arbitrary number of dimensions (rather than just 1), but with the rewrite I think this is unnecessary. Although, I'm still curious how to do this in a way that will compile to torchscript. I've run into the problem before.

> The tests detailed [here](https://github.com/pytorch/audio/tree/master/test) can be run using pytest. Is that what you meant?

I was isolating the `amplitude_to_DB` tests so it wasn't checking torchscript compilation. I assumed that was tested elsewhere, but including the torchscript consistency tests does the trick.

test/torchaudio_unittest/functional/functional_cpu_test.py
jcaw: I've removed this method. I'm just generating random tensors now - it does seem cleaner.

test/torchaudio_unittest/functional/functional_cpu_test.py
jcaw: I've gone ahead and split both.

torchaudio/functional/functional.py
```diff
@@ -227,14 +227,17 @@ def amplitude_to_DB(
         db_multiplier: float,
         top_db: Optional[float] = None
 ) -> Tensor:
-    r"""Turn a tensor from the power/amplitude scale to the decibel scale.
+    r"""Turn a spectrogram from the power/amplitude scale to the decibel scale.
 
     This output depends on the maximum value in the input tensor, and so
     may return different values for an audio clip split into snippets vs. a
     full clip.
```
vincentqb, line 234: Let's also update the text here.
```suggestion
    The output of each tensor in a batch depends on the maximum value of that tensor,
    and so may return different values for an audio clip split into snippets vs. a full clip.
```

torchaudio/functional/functional.py
```diff
@@ -227,14 +227,17 @@ def amplitude_to_DB(
...
 
     Args:
-        x (Tensor): Input tensor before being converted to decibel scale
+
+        x (Tensor): Input spectrogram(s) before being converted to decibel scale. Input should take
+          the form `(..., freq, time)`. Batched inputs must include a channel dimension and should
+          take the form `(batch, channel, freq, time)`.
```
vincentqb, line 240: nit: minor phrasing :)
```suggestion
        x (Tensor): Input spectrogram(s) before being converted to decibel scale. Input should take
          the form `(..., freq, time)`. Batched inputs should include a channel dimension and
          have the form `(batch, channel, freq, time)`.
```

vincentqb (approved):
LGTM with minor change to doc, thanks for working on this!

jcaw:
No worries!

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
-
-        torch.testing.assert_allclose(x2, torch.abs(x), atol=5e-5, rtol=1e-5)
+class Testamplitude_to_DB(common_utils.TorchaudioTestCase):
+    def _ensure_reversible(self, spec):
+        """Check `amplitude_to_db` returns the original when reversed."""
+        AMPLITUDE_MULT = 20.
+        POWER_MULT = 10.
+        AMIN = 1e-10
+        REF = 1.0
+        DB_MULT = math.log10(max(AMIN, REF))
```
mthrok, line 80: Since they are not global variables, can you lowercase them?
This might sound unimportant but, in the past, different contributors followed what's there and there were bunch of all capital parameters for not good reason.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
-        x2 = F.DB_to_amplitude(db, ref, power)
-
-        torch.testing.assert_allclose(x2, spec, atol=5e-5, rtol=1e-5)
+        db = F.amplitude_to_DB(spec, POWER_MULT, AMIN, DB_MULT, top_db=None)
+        x2 = F.DB_to_amplitude(db, REF, 1.)
+
+        self.assertEqual(x2, spec)
+
+    def test_amplitude_to_DB_batch(self):
+        spec = torch.rand([2, 2, 100, 100]) * 200
```
mthrok, line 95: Please set random seed every time `torch.rand` is called.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
+        AMPLITUDE_MULT = 20.
+        AMIN = 1e-10
+        REF = 1.0
+        DB_MULT = math.log10(max(AMIN, REF))
+        TOP_DB = 40.
+
+        decibels = F.amplitude_to_DB(spec, AMPLITUDE_MULT, AMIN,
+                                     DB_MULT, top_db=TOP_DB)
+        above_top = decibels >= 6.0205
+        assert above_top.all(), decibels
```
mthrok, line 116: Could you provide more descriptive message for the second argument? so that it is easy for a random maintainer later to tell what's wrong if it fails.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
+        self._check_top_db(spec[0])
+
+    def test_top_db_freq(self):
+        spec = torch.rand([1, 2, 100, 100]) * 200
+        # Predictability
+        spec[0, 0, 1] = 0
+        spec[0, 0, 0] = 200
+        self._check_top_db(spec[0][0])
+
+    def test_batched(self):
```
mthrok, line 139: I think `test_batched` and `test_per_spectrogram` are better suited in `batch_consistency_test`. And we should be applying the second type of test to all the functions.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
+        db = F.amplitude_to_DB(spec, POWER_MULT, AMIN, DB_MULT, top_db=None)
+        x2 = F.DB_to_amplitude(db, REF, 1.)
+
+        self.assertEqual(x2, spec)
+
+    def test_amplitude_to_DB_batch(self):
+        spec = torch.rand([2, 2, 100, 100]) * 200
+        self._ensure_reversible(spec)
+
+    def test_amplitude_to_DB_channels(self):
```
mthrok, line 98: Since the name of this test method is not descriptive, pleas add a description of each test so that it is easy for maintainers later to tell what is being tested.

I see that there is a description in a helper function but each test having doctoring allows to give overview with `pytest --collect-only`

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,124 @@ def test_pitch(self, frequency):
...
+                                          DB_MULT, top_db=TOP_DB)
+        itemwise_dbs = torch.stack([
+            F.amplitude_to_DB(item, AMPLITUDE_MULT, AMIN,
+                              DB_MULT, top_db=TOP_DB)
+            for item in spec
+        ])
+
+        self.assertEqual(batchwise_dbs, itemwise_dbs)
+
+    def test_per_spectrogram(self):
```
mthrok, line 165: I think this test is the fundamental response to #994. Can you refer that in docstring?

Example;

https://github.com/pytorch/audio/blob/422edb183f25ab91eb967381d328384e48c7a4a0/test/torchaudio_unittest/sox_io_backend/load_test.py#L359-L368

mthrok (changes requested):
Changes to the functional module looks good, but tests need more context for the maintainability.

test/torchaudio_unittest/functional/functional_cpu_test.py
jcaw: Not at all, that sounds sensible.

test/torchaudio_unittest/functional/functional_cpu_test.py
jcaw: I've moved these verbatim to `batch_consistency_test.py` - is that OK?

mthrok (commented):
Hi @jcaw 

Thanks for working on this.
The tests in `batch_consistency` looks good.
I had a couple of question regarding the tests in `Testamplitude_to_DB`.
Specifically `#Predictability` part, I am having difficulty understanding it.

If you do not have time to address the comments, let me know.
I do not want to drag you around too much, so I will move on and address them later.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,89 @@ def test_pitch(self, frequency):
...
+            )
+        )
 
-        torch.testing.assert_allclose(x2, spec, atol=5e-5, rtol=1e-5)
+    def test_top_db_batch(self):
+        torch.manual_seed(0)
+        spec = torch.rand([1, 2, 100, 100]) * 200
+        # Predictability
+        spec[0, 0, 1] = 0
+        spec[0, 0, 0] = 200
```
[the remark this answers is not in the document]
jcaw: The decibel cutoff is derived from the smallest value in the spectrogram, so in order to hard-code the cutoff, the smallest value needs to be predictable. There's also a (tiny) chance that no values large enough to be clamped are generated, so I manually set the max.

I default to using a random tensor when possible because it adds entropy to the inputs. It's just a rule of thumb, I can certainly change it.

I didn't really like this structure either, it is confusing. I've rewritten it to scale each spectrogram separately to strictly match the range (0, 200), which seems more sensible. Let me know what you think.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -70,46 +70,89 @@ def test_pitch(self, frequency):
...
-
         # Spectrogram power -> DB -> power
-        db = F.amplitude_to_DB(spec, multiplier, amin, db_multiplier, top_db=None)
-        x2 = F.DB_to_amplitude(db, ref, power)
+        db = F.amplitude_to_DB(spec, power_mult, amin, db_mult, top_db=None)
+        x2 = F.DB_to_amplitude(db, ref, 1.)
+
+        self.assertEqual(x2, spec)
+
+    def test_amplitude_to_DB_batch(self):
```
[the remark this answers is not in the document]
jcaw: I don't think it's nitpicky - you're right. I think I was sticking with the original naming convention for this test case, but the other cases use the naming scheme you described anyway.

mthrok (approved):
Hi @jcaw 

Thanks for working on this. The change looks good and the tests are now very comprehensive.
There is a conflict with latest master, so please resolve it (or let me know if you are busy, then I will take over).

test/torchaudio_unittest/functional/functional_cpu_test.py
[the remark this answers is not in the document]
mthrok: Thanks, the comments you added nicely explain the intention well. I think this is good.

jcaw:
Would you like me to merge or rebase?

mthrok:
> Would you like me to merge or rebase?

(I assume you mean `merge commit` as I do not think you can press the merge button of the PR.) Either works fine so long as the conflict is resolved. (so that I can click the merge button)
But since you have a bunch of commits on your branch, if you are going to rebase, you will need to squash the commits otherwise resolving the conflict will be difficult.

vincentqb:
The tests failing are not related to this pull request. I'll go ahead and merge the pull request. Thank you for the work @jcaw!

jcaw:
No worries! Happy to help.

PR #1114, opened 2020-12-21
Support bytes and file-like object in info

mthrok:
This PR is built on top of #1108, and it adds `bytes` and `file-like` object support to `info` function of "sox_io" backend and "soundfile (new interface)" backend.


mthrok:
Squashed to #1108

PR #1116, opened 2020-12-21
Support bytes and file-like obj in sox_effects

mthrok:
After #1114, this PR adds `bytes` and `file-like` object support to `sox_effect.apply_effects_file`.


mthrok:
Squashed into #1108

PR #1123, opened 2020-12-23
Add file-like obj handler for apply_effects

mthrok:
After #1116, this PR adds native support file-like object to load and sox_effects.


mthrok:
Squashed into #1108

Issue #1130, opened 2020-12-28
Improve Dataset utility test


mthrok:
There are couple of improvements we can make for dataset utility test.

https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/utils_test.py#L12-L36 

1. Use temporary directory for cache
    `test_disckcache_iterator` creates cache directory at working directory which does not get cleaned up. To improve this, one should use temporary directory for the `location` argument in `diskcache_iterator`. 
    To do this,
    1. Subclass the test suite with [`TempDirMixin`](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/common_utils/case_utils.py#L14-L40)
    2. Use `self.get_temp_path()` to get a temporary directory.
    3. Pass the temporary directory to `location` argument.
    
2. Make utility test not depend on an Dataset class
    CommonVoice dataset is another library code that can be broken anytime. The fact that utility test depends on both utility and another component from the library makes it confusing what is the cause when this test fails. Therefore CommonVoice should not be used in this test suite.
    The workaround is to define a simple dataset for this test suite. Something as simple as the following would do
    ```python
    class Dataset(torch.utils.data.Dataset):
        def __getitem__(self, n):
            return torch.randn(32, 2, 256)
    ```

AzizCode92:
I can work on this.

PR #1132, opened 2020-12-29
Improve Dataset utility test

AzizCode92:
Addresses this issue #1130


AzizCode92:
Hi @mthrok, can you please review my PR.

thank you

test/torchaudio_unittest/datasets/utils_test.py
```diff
@@ -1,36 +1,40 @@
...
+        return iter(range(32))
 
+
+class TestIterator(TorchaudioTestCase, TempDirMixin):
     @classmethod
-    def tearDownClass(cls):
-        COMMONVOICE._ext_audio = original_ext_audio
+    def setUpClass(cls):
+        cls.location = cls.get_base_temp_dir()
+        cls.dataset = Dataset()
```
mthrok, line 26: Since there is no requirement to share these attributes across the test cases, they should be defined and used in each test case.

```python
def test_diskcache_iterator(self)
    data = dataset_utils.diskcache_iterator(Dataset(), location=self.get_temp_path())
    ...

def test_bg_iterator(self):
    data = dataset_utils.bg_iterator(Dataset())
    ...
```

test/torchaudio_unittest/datasets/utils_test.py
```diff
@@ -1,36 +1,40 @@
...
+    TempDirMixin,
+    get_asset_path
 )
 
 from torchaudio.datasets import utils as dataset_utils
-from torchaudio.datasets.commonvoice import COMMONVOICE
 
-original_ext_audio = COMMONVOICE._ext_audio
 
+class Dataset(torch.utils.data.Dataset):
```
mthrok, line 11: The iterator protocol implemented here looks inconsistent between `__getitem__` and `__iter__`. i.e. index slicing will return `Tuple[Tensor, int]` while `__iter__` will return series of `int`s. Is `__iter__` method required?

I think something like the following would do
(I changed the returned Tensor to use the input argument `n`, and added variable `sample_rate` for the purpose of readability improvement.)

```python

def __getitem__(self, n):
    sample_rate = 8000
    waveform = n * torch.ones(2, 256)
    return waveform, sample_rate

def __len__(self) -> int:
    return 32
```

mthrok (commented):
@AzizCode92 Thanks for the contribution. Please see my comments and let me know what you think.

test/torchaudio_unittest/datasets/utils_test.py
AzizCode92: What I noticed is that without `__iter__` method inside the `Dataset` class, the test will have in infinite loop [here](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/torchaudio/datasets/utils.py#L258) and that's why we have timeout error in the CICD pipeline.

AzizCode92: Another option would also to use `IterableDataset`.

mthrok: > What I noticed is that without `__iter__` method inside the `Dataset` class, the test will have in infinite loop [here](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/torchaudio/datasets/utils.py#L258) and that's why we have timeout error in the CICD pipeline.

I see. When you saw it, did the Dataset have `__len__` implementation? The fact that CommonVoice has `__getitem__` and `__len__` but not `__iter__` makes me think that `__len__` would cap the maximum iteration.

AzizCode92: Yes, `__len__` method was there when we had the infinite loop.
Without the `__len__` method the `test_disckcache_iterator` test will fail due to an error:
`>       self._cache: List = [None] * len(dataset)
E       TypeError: object of type 'Dataset' has no len()`

mthrok: I see. In that case, reusing the implementation of `__len__` and `__getitem__` in `__iter__` will give a consistent behavior.

```python
def __iter__(self):
    for i in range(len(self)):
        yield self[i]
```

test/torchaudio_unittest/datasets/utils_test.py
```diff
@@ -1,36 +1,38 @@
+import torch
 from torchaudio_unittest.common_utils import (
     TorchaudioTestCase,
-    get_asset_path,
+    TempDirMixin,
+    get_asset_path
```
mthrok, line 5: Could you remove this `get_asset_path`? I think it is no longer used.
Also can you remove the files under `test/torchaudio_unittest/assets/CommonVoice/`?
Finally, we can get rid of it.

mthrok (approved):
Looks good. Thanks for working on this.

test/torchaudio_unittest/datasets/utils_test.py
AzizCode92: Done, thanks for the review.

mthrok:
Thanks!

PR #1134, opened 2020-12-30
refactor vctk unittest

AzizCode92:
- [x] Refactor [vctk](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/vctk_test.py#L39-L79)  unittest

Issue #1131, opened 2020-12-28
Improve Dataset test maintainability/readability


mthrok:
In #821, we improved the dataset tests to work on mocked files. In these changes, we applied the pattern to create mock dataset in `setUpClass`. As we keep improving the test for dataset, the `setUpClass` gets cluttered and became harder to grasp what's going on, so we should refactor the pattern.

In #1126 we extracted the dataset mocking part into a separate function. We can apply the same pattern to the other tests too. In the end, the initialization should look as simple as,

```python
@classmethod
def setUpClass(cls):
    cls.root_dir = cls.get_base_temp_dir()
    cls.data = get_mock_dataset(cls.root_dir)
```

with extracted helper function that creates mock data and returns the expected data.

```python
def get_mock_dataset(root_dir):
    ...
```

* [ ] [yesno_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/yesno_test.py#L31-L38)
* [x] [vctk_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/vctk_test.py#L39-L79)
* [ ] [tedlium_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/tedlium_test.py#L44-L95)
* [x] [speechcommand_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/speechcommands_test.py#L64-L107)
* [x] [ljspeech_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/ljspeech_test.py#L38-L59)
* [ ] [libritts_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/libritts_test.py#L30-L49)
* [ ] [librispeech_test](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/librispeech_test.py#L38-L88)
[...]

PR #1135, opened 2020-12-30
refactor tedlium unittest

AzizCode92:
- [x] refactor [tedlium](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/tedlium_test.py#L44-L95) unittest #1134 

#1131

### DIFF
--- a/test/torchaudio_unittest/datasets/tedlium_test.py
+++ b/test/torchaudio_unittest/datasets/tedlium_test.py
@@ -1,18 +1,15 @@
 import os
 import platform
-import unittest
 from pathlib import Path
 
-from torchaudio.datasets import tedlium
-
 from torchaudio_unittest.common_utils import (
-    TestBaseMixin,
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Looks good. Thanks!

test/torchaudio_unittest/datasets/tedlium_test.py
```diff
@@ -145,5 +151,6 @@ class TestTedliumSoundfileNew(Tedlium, TorchaudioTestCase):
     class TestTedliumSox(Tedlium, TorchaudioTestCase):
         backend = "sox"
 
+
```
mthrok, line 154: Looks like this is causing style test to fail, though I think this is okay. Can you fix this?

mthrok (approved):

PR #880, opened 2020-08-14
[Build] Set proper C++ ABI

mthrok:
With this change, `BUILD_TRANSDUCER=1 python setup.py build_ext` now sees `-D_GLIBCXX_USE_CXX11_ABI=` in the compilation command. (Note: sox is C-only so it is not relevant to sox build process)

```
[ 98%] Building CXX object transducer/CMakeFiles/warprnnt.dir/submodule/src/rnnt_entrypoint.cpp.o
cd /scratch/moto/torchaudio/build/temp.linux-x86_64-3.8/third_party/transducer && /home/moto/ccache/lib/c++  -DRNNT_DISABLE_OMP -I/scratch/moto/torchaudio/third_party/transducer/submodule/include  -D_GLIBCXX_USE_CXX11_ABI=0 -fPIC   -o CMakeFiles/warprnnt.dir/submodule/src/rnnt_entrypoint.cpp.o -c /scratch/moto/torchaudio/third_party/transducer/submodule/src/rnnt_entrypoint.cpp
```

See also: https://github.com/pytorch/text/pull/931

PR #1137, opened 2020-12-30
Add RNN Transducer Loss for CPU

vincentqb:
This pull request introduces `rnnt_loss` and `RNNTLoss` as a [prototype](https://pytorch.org/audio/stable/index.html) in `torchaudio.prototype.transducer` using [HawkAaron's warp-transducer](https://github.com/HawkAaron/warp-transducer).
- The python interface remains currently the same as the original.
- This has been tested as integrated within ESPNet [here](https://github.com/vincentqb/espnet/pull/1).

Follow-ups ([comment](https://github.com/pytorch/audio/pull/1137#issuecomment-753225422), [comment](https://github.com/pytorch/audio/pull/1137#issuecomment-754410503)):

- Extend guards for prototype
  - Guard prototype python files by omitting them from torchaudio, see also [comment](https://github.com/pytorch/audio/pull/1137#discussion_r551496192)
  - Guard building third party transducer even if not added as an extension
  - Enable building transducer in nightlies only, disable for release.
- Update building process
  - Pass along the `DEBUG` flag to cmake
  - Remove hardcoded O2/O3 optimization, see [comment](https://github.com/pytorch/audio/pull/1137#discussion_r551498022)
  - Build within same folders as libsox, [comment](https://github.com/pytorch/audio/pull/1137#discussion_r551499829) and [comment](https://github.com/pytorch/audio/pull/1137#discussion_r551556305)
  - Move libsox to a third_party subfolder as suggested [above](https://github.com/pytorch/audio/pull/1137#discussion_r550321378).
- Modernization
  - Migrate the [checks](https://github.com/pytorch/audio/pull/1137/files#diff-9623f973feeae59fa7215b73ad28d00f83c59bfa00f0743c096e4ab36fe5e25dR133) to C++.
  - Investigate using [`AT_DISPATCH_FLOATING_TYPES`](https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/aten/src/ATen/native/cpu/Activation.cpp#L61).
  - Patch the submodule to remove the pytorch deprecation warnings.
  - Add GPU implementation and compilation.
  - Refactor and update the API, see [warprnnt](https://github.com/1ytic/warp-rnnt) and [internal](https://fb.quip.com/tzrDA3DHqlY8).

cc @astaff, [internal](https://fb.quip.com/tzrDA3DHqlY8), #1099


build_tools/setup_helpers/extension.py
```diff
@@ -132,11 +132,60 @@ def get_ext_modules(debug=False):
...
+    include_path = _TP_TRANSDUCER_BASE_DIR / 'submodule' / 'include'
+
+    return CppExtension(
+        name=_TRANSDUCER_NAME,
+        sources=[os.path.realpath(path) for path in source_paths],
+        libraries=librairies,
+        include_dirs=[os.path.realpath(include_path)],
+        library_dirs=[os.path.realpath(build_path)],
+        extra_compile_args=extra_compile_args,
+        extra_objects=[str(build_path / f'lib{lib}.a') for lib in librairies],
```
cpuhrsch, line 189: nit: libraries

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
...
+    if (NOT APPLE)
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -O2")
+    ENDIF()
+
+    ADD_LIBRARY(warprnnt STATIC submodule/src/rnnt_entrypoint.cpp)
+
+ENDIF()
+
+
+add_executable(test_cpu submodule/tests/test_cpu.cpp submodule/tests/random.cpp )
```
cpuhrsch, line 129: I don't think this is necessary. We don't run these executables in CI or anything, so we don't have to build them.

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
...
+        set(CMAKE_SKIP_RPATH TRUE)
+    ENDIF ()
+
+ELSE()
+    #always skip for linux
+    set(CMAKE_SKIP_RPATH TRUE)
+ENDIF()
+
+
+IF (WITH_GPU)
```
cpuhrsch, line 100: I'd remove this for now until we add GPU support in a later PR

torchaudio/prototype/transducer.py
```diff
@@ -0,0 +1,147 @@
...
+        raise ValueError("{} must be contiguous".format(name))
+
+
+def check_dim(var, dim, name):
+    if len(var.shape) != dim:
+        raise ValueError("{} must be {}D".format(name, dim))
+
+
+def certify_inputs(log_probs, labels, lengths, label_lengths):
+    # check_type(log_probs, torch.float32, "log_probs")
```
cpuhrsch, line 123: Why is this commented out?

torchaudio/prototype/transducer.py
```diff
@@ -0,0 +1,147 @@
...
+    def backward(ctx, grad_output):
+        grad_output = grad_output.view(-1, 1, 1, 1).to(ctx.grads)
+        return ctx.grads.mul_(grad_output), None, None, None, None, None
+
+
+@_mod_utils.requires_module('_warp_transducer')
+def rnnt_loss(acts, labels, act_lens, label_lens, blank=0, reduction="mean"):
+    """RNN Transducer Loss
+
+    Args:
```
cpuhrsch, line 58: I think the documentation could be improved a bit. It could also be useful to reference the paper.

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
...
+option(WITH_OMP "compile warp-rnnt with openmp." ON)
+
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+set(CUDA_NVCC_FLAGS "${CUDA_NVCC_FLAGS} -Xcompiler -fPIC")
+
+if(NOT WITH_OMP)
+    add_definitions(-DRNNT_DISABLE_OMP)
+endif()
+
+if (WITH_OMP)
```
cpuhrsch, line 48: I'd remove this as well

build_tools/setup_helpers/extension.py
```diff
@@ -132,11 +132,60 @@ def get_ext_modules(debug=False):
...
+        _get_transducer_module(),
     ]
 
 
 class BuildExtension(TorchBuildExtension):
     def build_extension(self, ext):
         if ext.name == _EXT_NAME and _BUILD_SOX:
             _build_third_party()
+        if ext.name == _TRANSDUCER_NAME:
+            _build_transducer()
```
mthrok, line 144: This is adding `_warp_transducer.so`. This is not the way third-party libraries should be built in torchaudio.
All the third party library should be built as part of `_torchaudio.so` and statically linked into it, so that runtime library management is easy.

build_tools/setup_helpers/extension.py
```diff
@@ -132,11 +132,60 @@ def get_ext_modules(debug=False):
...
+    subprocess.run(
+        args=['cmake', str(_TP_TRANSDUCER_BASE_DIR), '-DWITH_OMP=OFF'],
+        cwd=build_dir,
+        check=True,
+    )
+    subprocess.run(
+        args=['cmake', '--build', '.'],
+        cwd=build_dir,
+        check=True,
+    )
```
mthrok, line 164: There should be only one pair of `cmake .` and `cmake --build` calls, which in turn will call into sox build process and transducer build process. (possibly parallel),

Please introduce top level `third_party` directory and put the existing third party libraries in a dedicated directory.

```
third_party/libsox
third_party/transducer
```

Then in `third_party/CMakeLists.txt`, do 

```cmake
add_subdirectory(libsox)
add_subdirectory(transducer)
```

Then when calling `cmake`, designate the top-level `third_party` directory as the source directory and let `cmake` figure out the rest.

build_tools/setup_helpers/extension.py
```diff
@@ -132,11 +132,60 @@ def get_ext_modules(debug=False):
...
 class BuildExtension(TorchBuildExtension):
     def build_extension(self, ext):
         if ext.name == _EXT_NAME and _BUILD_SOX:
             _build_third_party()
+        if ext.name == _TRANSDUCER_NAME:
+            _build_transducer()
         super().build_extension(ext)
+
+
+_TRANSDUCER_NAME = '_warp_transducer'
```
mthrok, line 148: This will get installed in global namespace, outside of `torchaudio` package directory.
Please put it in `torchaudio` package.

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
...
+ENDIF()
+
+project(rnnt_release)
+
+IF (NOT APPLE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2")
+ENDIF()
+
+IF (APPLE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -O2")
```
mthrok, line 16: This looks conflict with the `'-std=c++14'` flag given in `build_tools/setup_helpers/extension.py`

third_party/warp_transducer/binding.cpp
```diff
@@ -0,0 +1,47 @@
...
+            labels,
+            input_lengths,
+            label_lengths,
+            costs,
+            grads,
+            blank_label,
+            num_threads);
+}
+
+TORCH_LIBRARY(warprnnt_pytorch_warp_rnnt, m) {
```
mthrok, line 34: I think this should live in `torchaudio` namespace.

third_party/warp_transducer/binding.cpp
```diff
@@ -0,0 +1,47 @@
+#include <iostream>
+#include <numeric>
+
+#include <torch/extension.h>
```
mthrok, line 4: I think `<torch/script.h>` is the appropriate header.

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
+# Modified from HawkAaron/warp-transducer/CMakeLists.txt to build statically
+
+IF (APPLE)
+    cmake_minimum_required(VERSION 3.4)
+ELSE()
+    cmake_minimum_required(VERSION 2.8)
+ENDIF()
```
mthrok, line 7: This should require at least 3.5, which is used by `PyTorch`.

third_party/warp_transducer/CMakeLists.txt
```diff
@@ -0,0 +1,142 @@
...
+    set(CMAKE_SKIP_RPATH TRUE)
+ENDIF()
+
+
+IF (WITH_GPU)
+
+    MESSAGE(STATUS "Building static library with GPU support")
+
+    CUDA_ADD_LIBRARY(warprnnt STATIC submodule/src/rnnt_entrypoint.cu)
+    IF (!Torch_FOUND)
```
mthrok, line 105: If torch is not found, shouldn't it be failing?

torchaudio/__init__.py
```diff
@@ -1,4 +1,6 @@
 from . import extension
+from . import prototype
```
mthrok, line 2: I think it's better not to import `prototype` automatically, so that users have to do explicit import by themselves.

torchaudio/extension/__init__.py
```diff
@@ -1,7 +1,9 @@
 from .extension import (
     _init_extension,
+    _init_transducer_extension,
 )
 
 _init_extension()
+_init_transducer_extension()
```
mthrok, line 7: I think this initialization should happen in `torchaudio/prototype/__init__.py` so that the non-prototype part of the code is fully-unaware of prototype.

torchaudio/prototype/transducer.py
```diff
@@ -0,0 +1,147 @@
...
+    if not var.is_contiguous():
+        raise ValueError("{} must be contiguous".format(name))
+
+
+def check_dim(var, dim, name):
+    if len(var.shape) != dim:
+        raise ValueError("{} must be {}D".format(name, dim))
+
+
+def certify_inputs(log_probs, labels, lengths, label_lengths):
```
mthrok, line 133: `certify` is a uncommon name. what does it do?

torchaudio/prototype/transducer.py
```diff
@@ -0,0 +1,158 @@
...
+            (Default: ``'mean'``)
+    """
+
+    def __init__(self, blank=0, reduction="mean"):
+        super(RNNTLoss, self).__init__()
+        self.blank = blank
+        self.reduction = reduction
+        self.loss = _RNNT.apply
+
+    def forward(self, acts, labels, act_lens, label_lens):
```
cpuhrsch, line 102: If you don't want to copy-paste the docs from the functional you could reference it here within the documentation.

torchaudio/prototype/transducer.py
cpuhrsch: check should be a more common name that we could use here.

torchaudio/prototype/transducer.py
vincentqb: thanks for catching this: removed, no longer needed.

vincentqb:
Some TODOs:
- Use `TORCH_CHECK` to raise an error instead of writing to standard output, [here](https://github.com/pytorch/audio/pull/1137/files#diff-0a85052ed18261506c343352ee4ca33cc70b29c5b44e2819d6a22e37948ba0bcR67).
- Investigate the memory allocation for workspace, [here](https://github.com/pytorch/audio/pull/1137/files#diff-0a85052ed18261506c343352ee4ca33cc70b29c5b44e2819d6a22e37948ba0bcR40).

Some follow-ups:
- Move libsox to a third_party subfolder as suggested [above](https://github.com/pytorch/audio/pull/1137#discussion_r550321378).
- Migrate the [checks](https://github.com/pytorch/audio/pull/1137/files#diff-9623f973feeae59fa7215b73ad28d00f83c59bfa00f0743c096e4ab36fe5e25dR133) to C++.
- Add GPU implementation and compilation.
- Patch the warp RNN-T code to remove the pytorch deprecation warnings.

test/torchaudio_unittest/transducer_test.py
```diff
@@ -0,0 +1,303 @@
...
+
+from torchaudio_unittest import common_utils
+from torchaudio.prototype.transducer import RNNTLoss
+
+
+def get_basic_data(device):
+    # Test if example provided in README runs
+    # https://github.com/HawkAaron/warp-transducer
+
+    acts = torch.FloatTensor(
```
cpuhrsch, line 12: nit: use the factory function ```torch.tensor([xyz], dtype=torch.float)``` instead of the type constructor. Same applies to IntTensor.

.gitmodules
```diff
@@ -0,0 +1,4 @@
+[submodule "third_party/warp_transducer/submodule"]
+	path = third_party/transducer/submodule
+	url = https://github.com/HawkAaron/warp-transducer
+	branch = master
```
mthrok, line 4: Could you add `ignore = dirty` so that whatever the intermediate files in this directory would be ignored by git.

Also I am wondering the effect of `branch = master`. What happens if there is a new commit in the master branch?
Typically, it's not necessary to add branch.
https://github.com/pytorch/pytorch/blob/master/.gitmodules

torchaudio/csrc/register.cpp
```diff
@@ -77,5 +77,17 @@ TORCH_LIBRARY(torchaudio, m) {
   m.def(
       "torchaudio::sox_effects_apply_effects_file",
       &torchaudio::sox_effects::apply_effects_file);
+
+  //////////////////////////////////////////////////////////////////////////////
+  // transducer.cpp
+  //////////////////////////////////////////////////////////////////////////////
+  m.def("rnnt_loss(Tensor acts,"
```
mthrok, line 84: We need a guard for excluding the prototype from the binary for the release process.
We need to decide wether we want to build prototype by default, and define a way to toggle it, (which is starting from an environment variable, then passing/parsing that in CMake, then passing it as compiler definition, which will be tested with `#ifdef` macro) then modify the CI configuration so that nightly build contains prototype, but proper release won't.

build_tools/setup_helpers/extension.py
```diff
@@ -64,6 +64,7 @@ def _get_srcs():
 def _get_include_dirs():
     dirs = [
         str(_ROOT_DIR),
+        str(_TP_BASE_DIR / 'transducer' / 'submodule' / 'include'),
```
mthrok, line 67: I think installing the resulting library and header in the same directory as sox, i.e. `<_TP_BASE_DIR>/install/ [lib | include]` would make things simpler.

build_tools/setup_helpers/extension.py
```diff
@@ -94,6 +95,7 @@ def _get_extra_objects():
         ]
         for lib in libs:
             objs.append(str(_TP_INSTALL_DIR / 'lib' / lib))
+    objs.append(str(_TP_BASE_DIR / 'build' / 'transducer' / 'libwarprnnt.a'))
```
mthrok, line 98: This has to be conditional too.

third_party/transducer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    SET(CMAKE_SKIP_RPATH TRUE)
+ENDIF()
+
+ADD_LIBRARY(warprnnt STATIC submodule/src/rnnt_entrypoint.cpp)
+
+INSTALL(TARGETS warprnnt
+        LIBRARY DESTINATION "lib"
+        ARCHIVE DESTINATION "archives")
+
+INSTALL(FILES submodule/include/rnnt.h DESTINATION "submodule/include")
```
mthrok, line 38: "archives" directory looks uncommon. Typically, you can use `"lib"`. however, as I am suggesting in the previous comments, installing the resulting library and header in `third_party/install/[lib|include]` will make the dependency management simpler.

For that you can remove `PROJECT` command from this file, and set `CMAKE_INSTALL_PREFIX` to `${CMAKE_CURRENT_SOURCE_DIR}/install`, then just `install(TARGETS warprnnt)` as the default subdirectory for static libraries (`archive`) is `"lib"`.

torchaudio/prototype/transducer.py
```diff
@@ -0,0 +1,157 @@
...
+            (Default: ``'mean'``)
+    """
+
+    # NOTE manually done log_softmax for CPU version,
+    # log_softmax is computed within GPU version.
+    acts = torch.nn.functional.log_softmax(acts, -1)
+    return _RNNT.apply(acts, labels, act_lens, label_lens, blank, reduction)
+
+
+@_mod_utils.requires_module('torchaudio._torchaudio')
```
mthrok, line 85: I think it's ideal to have a finer checking than just checking C++ extension.

third_party/transducer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
+CMAKE_MINIMUM_REQUIRED(VERSION 3.5)
+
+PROJECT(rnnt_release)
+
+SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2")
```
mthrok, line 5: I wonder if this flag should be hardcoded here, and wonder why it's `-O2`, instead of `-O3`.

third_party/transducer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+IF(APPLE)
+    ADD_DEFINITIONS(-DAPPLE)
+ENDIF()
+
+INCLUDE_DIRECTORIES(submodule/include)
```
mthrok, line 11: Please do not use `include_directories`, this add the target directory to all the project, which is not necessary.

Use `target_include_directories` instead. Setting it as `PUBLIC` target, you might be able to get rid of the second call to `install`.

.gitmodules
cpuhrsch: This is about the build files etc., right? We should probably rather build outside of the submodule in third_party/transducer, if that's the only reason. Otherwise, we should be able to just standardize on PyTorch and add this.

third_party/transducer/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+
+ELSE()
+    # always skip for linux
+    SET(CMAKE_SKIP_RPATH TRUE)
+ENDIF()
+
+ADD_LIBRARY(warprnnt STATIC submodule/src/rnnt_entrypoint.cpp)
+
+INSTALL(TARGETS warprnnt
+        LIBRARY DESTINATION "lib"
```
cpuhrsch, line 35: In relation to the "mark dirty": So here, if you wanted to build outside of the submodule you might need to do ".../build/lib" etc.

I'd mark that as follow-up work.

test/torchaudio_unittest/transducer_test.py
```diff
@@ -0,0 +1,307 @@
...
+        costs, gradients = compute_with_pytorch_transducer(data=data)
+        np.testing.assert_allclose(costs, ref_costs, atol=atol, rtol=rtol)
+        self.assertEqual(logits_shape, gradients.shape)
+        if not np.allclose(gradients, ref_gradients, atol=atol, rtol=rtol):
+            for b in range(len(gradients)):
+                T = data["src_lengths"][b]
+                U = data["tgt_lengths"][b]
+                for t in range(gradients.shape[1]):
+                    for u in range(gradients.shape[2]):
+                        np.testing.assert_allclose(
```
cpuhrsch, line 288: self.assertEqual should be preferred

vincentqb:
Some more TODOs:
- [x] Remove numpy from tests
- [x] Guard prototype csrc and extension

Some more follow-ups:
- Pass along the `DEBUG` flag to cmake
- Guard prototype python files by omitting them from torchaudio
- Guard building third party transducer even if not added as an extension

Error below also happens on [master](https://app.circleci.com/pipelines/github/pytorch/audio/4428/workflows/f5bf08ff-03a5-4da6-8430-048ef9443abc/jobs/139818):
```
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform osx-64: {"python[version='>=2.7,<2.8.0a0|>=3.8,<3.9.0a0|>=3.5,<3.6.0a0|>=3.5']", "python[version='>=3.6,<3.7.0a0|>=3.7,<3.8.0a0|>=3.9,<3.10.0a0']", "python_abi[version='3.6.*|3.7.*',build='*_cp36m|*_cp37m']", "python[version='>=2.7,<2.8.0a0|>=3.6,<3.7.0a0|>=3.7,<3.8.0a0|>=3.9,<3.10.0a0|>=3.8,<3.9.0a0|>=3.5,<3.6.0a0']", 'python', 'python_abi=3.9[build=*_cp39]'}
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -0,0 +1,278 @@
...
+            RuntimeError, rnnt_loss, acts, labels, act_length, label_length
+        )
+
+    def test_basic_backward(self):
+        rnnt_loss = RNNTLoss()
+        acts, labels, act_length, label_length = get_data_basic(self.device)
+        loss = rnnt_loss(acts, labels, act_length, label_length)
+        loss.backward()
+
+    def _test_costs_and_gradients(
```
cpuhrsch, line 260: This could be inlined since it only has one call-site and is pretty small (but that's not the reason to remove an abstraction necessarily).

cpuhrsch (approved):

third_party/CMakeLists.txt
```diff
@@ -88,3 +88,5 @@ ExternalProject_Add(libsox
   # See https://github.com/pytorch/audio/pull/1026
   CONFIGURE_COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/build_codec_helper.sh ${CMAKE_CURRENT_SOURCE_DIR}/src/libsox/configure ${COMMON_ARGS} --with-lame --with-flac --with-mad --with-oggvorbis --without-alsa --without-coreaudio --without-png --without-oss --without-sndfile --with-opus --with-amrwb --with-amrnb --disable-openmp --without-sndio --without-pulseaudio
 )
+
+add_subdirectory(transducer)
```
mthrok, line 92: This one also needs to be guarded. I am trying not to touch RNN transducer with 

```
BUILD_SOX=1 python setup.py develop
```

```
CMake Error at transducer/CMakeLists.txt:32 (ADD_LIBRARY):
  Cannot find source file:

    submodule/src/rnnt_entrypoint.cpp
```

yet it is still trying to build transducer.

The fix is to define an option for CMake like `option(BUILD_TRANSDUCER "Build RNN Transducer" ON)`

Then from `build_tools/setup_helpers/extension.py`, do `cmake -DBUILD_TRANSDUCER=ON|OFF`.

torchaudio/csrc/transducer.cpp
```diff
@@ -0,0 +1,82 @@
...
+      default:
+        TORCH_CHECK(false,
+            std::string(__func__) + " not implemented for '" + toString(acts.scalar_type()) + "'"
+        );
+    }
+    return -1;
+}
+
+TORCH_LIBRARY_IMPL(torchaudio, CPU, m) {
+    m.impl("rnnt_loss", &cpu_rnnt_loss);
```
mthrok, line 79: @vincentqb Can you define a proper namespace? `torchaudio::<something>::rnnt_loss`

I am not sure how you want to move on, but if you have a plan to add different type of rnnt, then more descriptive name would work better later, like `warprnnt`

torchaudio/csrc/transducer.cpp
```diff
@@ -0,0 +1,82 @@
+#ifdef BUILD_TRANSDUCER
```
mthrok, line 1: Instead of guarding the whole file, we should exclude this from the list of source code when not building transducer binding.

torchaudio/csrc/register.cpp
mthrok: Also to decouple the different bindings. We should use `TORCH_LIBRARY_FRAGMENT` and register `rnnt` inside of `transducer.cpp`

build_tools/setup_helpers/extension.py
```diff
@@ -94,6 +99,8 @@ def _get_extra_objects():
         ]
         for lib in libs:
             objs.append(str(_TP_INSTALL_DIR / 'lib' / lib))
+    if _BUILD_TRANSDUCER:
+        objs.append(str(_TP_BASE_DIR / 'build' / 'transducer' / 'libwarprnnt.a'))
```
mthrok, line 103: This mean it is linking to the library in build directory, not the one installed with `install` command.

torchaudio/csrc/transducer.cpp
```diff
@@ -0,0 +1,82 @@
...
+        get_workspace_size(maxT, maxU, minibatch_size,
+                           false, &cpu_size_bytes);
+
+        std::vector<float> cpu_workspace(cpu_size_bytes / sizeof(float), 0);
+
+        compute_rnnt_loss(acts.data<float>(), grads.data<float>(),
+                          labels.data<int>(), label_lengths.data<int>(),
+                          input_lengths.data<int>(), alphabet_size,
+                          minibatch_size, costs.data<float>(),
+                          cpu_workspace.data(), options);
```
mthrok, line 50: Note `data()` is issuing a bunch of deprecation warnings;

```
/scratch/moto/torchaudio/torchaudio/csrc/transducer.cpp:44:44: warning: ‘T* at::Tensor::data() const [with T = float]’ is deprecated: Tensor.data<T>() is deprecated. Please use Tensor.data_ptr<T>() instead. [-Wdeprecated-declarations]
         compute_rnnt_loss(acts.data<float>(), grads.data<float>(),
                                            ^
In file included from /home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/include/ATen/Tensor.h:3:0,
                 from /home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/include/ATen/Context.h:4,
                 from /home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/include/ATen/ATen.h:9,
                 from /home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/include/torch/csrc/api/include/torch/types.h:3,
                 from /home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/include/torch/script.h:3,
                 from /scratch/moto/torchaudio/torchaudio/csrc/transducer.cpp:6:
```

third_party/CMakeLists.txt
vincentqb: Yes, this is the issue I referred to as "Guard building third party transducer even if not added as an extension" in the [follow-ups](https://github.com/pytorch/audio/pull/1137#issuecomment-754410503).

torchaudio/csrc/transducer.cpp
vincentqb: yes, that's one of the quick follow-up: i opened #1160 to remove this just now. feel free to stamp :)

build_tools/setup_helpers/extension.py
vincentqb: If they were not, I'd expect the nightlies to fail on my computer, right?
```
base ❯ conda create -n pytorch-nightly -c pytorch-nightly pytorch torchaudio            
base ❯ conda activate pytorch-nightly
pytorch-nightly ❯ python -c "from torchaudio.prototype.transducer import RNNTLoss; import torchaudio; print(torchaudio.__version__)"
0.8.0a0+361a146
pytorch-nightly ❯ 
```
Or did you find a different case that failed?

mthrok: It's not about nightly but about how CMake organize the intermediate files and directories inside of build directory. (nighty build has the library data statically linked. so, as long as build succeed, it is not an issue.) This is more about maintainability.

The file being pointed here is an intermediate product that is waiting to be installed. I do not think there is any guarantee that CMake will keep it this way, or CMake in different system could put it differently. Also this way, users do not have control over where the resulting binary is found.
(this way, the INSTALL command in CMakeLists.txt is not used)

So, more future proof way to do this is to perform `install` step and install the resulting binary in pre-determined directory. I cleaned this up in #1159

torchaudio/csrc/transducer.cpp
mthrok: Adding anonymous namespace in #1159 for the time being.

torchaudio/csrc/transducer.cpp
mthrok: addressed in #1159

third_party/CMakeLists.txt
mthrok: Addressed in #1159

mthrok:
@vincentqb I update the followup description for things addressed in #1159 and #1161. Please stamp these PRs when you have time.

For `Enable building transducer in nightlies only, disable for release.` I am thinking to add master environment variable that has higher precedence than `BUILD_TRANSDUCER`, like `DISABLE_PROTOTYPE` and propagating it from CCI configuration.

mthrok:
For C++ ABI issue ssee #880

third_party/warp_transducer/binding.cpp
vincentqb: (cc https://github.com/pytorch/pytorch/issues/18718#issuecomment-480913493)

PR #1138, opened 2020-12-30
Refactor LJSpeech unittest

AzizCode92:
- [x] refactor [ljspeech](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/ljspeech_test.py#L38-L59) unittest.

#1131

### DIFF
--- a/test/torchaudio_unittest/datasets/ljspeech_test.py
+++ b/test/torchaudio_unittest/datasets/ljspeech_test.py
@@ -2,8 +2,6 @@ import csv
 import os
 from pathlib import Path
 
-from torchaudio.datasets import ljspeech
-
 from torchaudio_unittest.common_utils import (
     TempDirMixin,
     TorchaudioTestCase,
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/datasets/ljspeech_test.py
```diff
@@ -12,59 +10,71 @@
...
     def _test_ljspeech(self, dataset):
         n_ite = 0
         for i, (waveform, sample_rate, transcript, normalized_transcript) in enumerate(
-            dataset
+                dataset
         ):
-            expected_transcript = self.transcripts[i]
-            expected_normalized_transcript = self.normalized_transcripts[i]
+            expected_transcript = _TRANSCRIPTS[i]
+            expected_normalized_transcript = _NORMALIZED_TRANSCRIPT[i]
```
mthrok, line 77: Could you make `get_mock_dataset ` return `_TRANSCRIPTS` and `_NORMALIZED_TRANSCRIPT`?

`get_mock_dataset` should be returning the expected data that correct Dataset implementation should be returning.

mthrok (approved):

PR #1139, opened 2020-12-30
Refactor LibriTTS unittest

AzizCode92:
- [x] refactor [libritts](https://github.com/pytorch/audio/blob/71214b48548b1dcb6ebd581dd36a9d0e60af6837/test/torchaudio_unittest/datasets/libritts_test.py#L30-L49) unittest.

#1131

### DIFF
--- a/test/torchaudio_unittest/datasets/libritts_test.py
+++ b/test/torchaudio_unittest/datasets/libritts_test.py
@@ -1,8 +1,6 @@
 import os
 from pathlib import Path
 
-from torchaudio.datasets.libritts import LIBRITTS
-
 from torchaudio_unittest.common_utils import (
     TempDirMixin,
     TorchaudioTestCase,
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/datasets/libritts_test.py
```diff
@@ -11,42 +9,54 @@
...
+
+_UTTERANCE_IDS = [
+    [19, 198, '000000', '000000'],
+    [26, 495, '000004', '000000'],
+]
+_ORIGINAL_TEXT = 'this is the original text.'
+_NORMALIZED_TEXT = 'this is the normalized text.'
+
+
+def get_mock_dataset(root_dir):
```
mthrok, line 22: Can you make `get_mock_dataset` return all the expected sample data, i.e. texts and utterance IDs, so that test suites do not need to refer to global variables?

Issue #1115, opened 2020-12-21
Support in-memory encoding/decoding


mthrok:
Add support for file-like object in I/O operations (info/load/save/sox_effects).

* [x] #1158 load
* [x] #1108 info
* [x] #1166 sox_effect
* [x] #1141 save func
* [x] #1165 smoke test
* (Optionally) Change argument name from `filepath` to `src`.

PR #1141, opened 2020-12-30
Support file-like object in save func

mthrok:
This adds file object support to `save` function.

Also this PR disables TP binary cache, as it seems to cause an occasional build error.

See #1115

### DIFF
--- a/torchaudio/backend/_soundfile_backend.py
+++ b/torchaudio/backend/_soundfile_backend.py
@@ -1,10 +1,11 @@
 """The new soundfile backend which will become default in 0.8.0 onward"""
+import os
 from typing import Tuple, Optional
 import warnings
 
 import torch
 from torchaudio._internal import module_utils as _mod_utils
-from .common import AudioMetaData
+from .common import AudioMetaData, get_ext
 
 
 if _mod_utils.is_module_available("soundfile"):
@@ -138,6 +139,7 @@ def save(
     sample_rate: int,
     channels_first: bool = True,
     compression: Optional[float] = None,
+    format: Optional[str] = None,
 ):
     """Save audio data to file.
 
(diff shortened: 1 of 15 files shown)


torchaudio/backend/_soundfile_backend.py
```diff
@@ -177,7 +182,11 @@ def save(
             "The argument is silently ignored."
         )
 
-    ext = str(filepath).split(".")[-1].lower()
+    try:
```
cpuhrsch, line 185: For now we could simply require the user to pass a format for a file-like object and then ask for feedback and see whether there's interest in this soundfile-like behavior that autodetect the destination format in some cases.

torchaudio/backend/common.py
```diff
@@ -19,6 +20,22 @@ def __init__(self, sample_rate: int, num_frames: int, num_channels: int):
         self.num_channels = num_channels
 
 
+def get_ext(
```
cpuhrsch, line 23: Based on above suggestion this could be removed

torchaudio/csrc/sox/io.cpp
```diff
@@ -147,6 +152,74 @@ std::tuple<torch::Tensor, int64_t> load_audio_fileobj(
...
+  }
+
+  torchaudio::sox_effects_chain::SoxEffectsChain chain(
+      /*input_encoding=*/get_encodinginfo("wav", tensor.dtype()),
+      /*output_encoding=*/sf->encoding);
+  chain.addInputTensor(&signal);
+  chain.addOutputFileObj(sf, &buffer.ptr, &buffer.size, &fileobj);
+  chain.run();
+
+  sf.close();
```
cpuhrsch, line 218: You could just do ```delete sf``` instead of creating a new method

cpuhrsch (approved):

Issue #1156, opened 2021-01-06
Replace pytest's paremeterization with parameterized


mthrok:
If you are interested working on this, please leave a comment here (so that other people will not do a duplicated work), then create a PR and tag @mthrok.

---

For historical and technical reasons, torchaudio's test suite cannot use `pytest` as a module. (we still use it as a runner in CircleCI, which is fine.)

There are couple of pytest utilities used in torchaudio's test code.

https://github.com/pytorch/audio/blob/6b07bcf80fafd77cb8bee32c316ce8b55323b868/test/torchaudio_unittest/functional/functional_cpu_test.py#L115-L163

We would like to replace `pytest.mark.parametrize` with `parameterized` module like the following

https://github.com/pytorch/audio/blob/3250d3df168c956389bd16956aa458ce111570d0/test/torchaudio_unittest/sox_compatibility_test.py#L305-L310

Steps
1. Convert the test function into test class using `TorchaudioTestCase`
2. In test method, use `parameterized.expand`
3. Additionally, when generating a tensor with random values, do it in the method and set the seed value first so that the test is more deterministic.

krishnakalyan3:
@mthrok I would be happy to work on this.

mthrok:
@krishnakalyan3 sure. thanks!

PR #1157, opened 2021-01-06
Replace pytest's paremeterization with parameterized 

krishnakalyan3:
Made Changes to

- [x] test/torchaudio_unittest/librosa_compatibility_test.py
- [x] test/torchaudio_unittest/functional/functional_cpu_test.py

fixes: #1156

### DIFF
--- a/test/torchaudio_unittest/functional/functional_cpu_test.py
+++ b/test/torchaudio_unittest/functional/functional_cpu_test.py
@@ -125,7 +125,7 @@ class TestDB_to_amplitude(common_utils.TorchaudioTestCase):
 
 class TestComplexNorm(common_utils.TorchaudioTestCase):
     @parameterized.expand(list(itertools.product(
-        [(1, 2, 1025, 400, 2),  (1025, 400, 2)],
+        [(1, 2, 1025, 400, 2), (1025, 400, 2)],
         [1, 2, 0.7]
     )))
     def test_complex_norm(self, shape, power):
@@ -138,7 +138,7 @@ class TestComplexNorm(common_utils.TorchaudioTestCase):
 
 class TestMaskAlongAxis(common_utils.TorchaudioTestCase):
     @parameterized.expand(list(itertools.product(
-        [(2, 1025, 400),  (1, 201, 100)],
+        [(2, 1025, 400), (1, 201, 100)],
         [100],
         [0., 30.],
         [1, 2]
@@ -156,6 +156,7 @@ class TestMaskAlongAxis(common_utils.TorchaudioTestCase):
 
         assert mask_specgram.size() == specgram.size()
         assert num_masked_columns < mask_param
+
 
 class TestMaskAlongAxisIID(common_utils.TorchaudioTestCase):
     @parameterized.expand(list(itertools.product(
--- a/test/torchaudio_unittest/functional/functional_cpu_test.py
+++ b/test/torchaudio_unittest/functional/functional_cpu_test.py
@@ -54,7 +54,7 @@ class TestComputeDeltas(common_utils.TorchaudioTestCase):
         specgram = torch.tensor([[[1.0, 2.0, 3.0, 4.0]]])
         expected = torch.tensor([[[0.5, 1.0, 1.0, 0.5]]])
         computed = F.compute_deltas(specgram, win_length=3)
-        torch.testing.assert_allclose(computed, expected)
+        self.assertEqual(computed, expected)
 
     def test_two_channels(self):
         specgram = torch.tensor([[[1.0, 2.0, 3.0, 4.0],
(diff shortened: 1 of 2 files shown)


test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -121,16 +121,18 @@ def test_DB_to_amplitude(self):
...
-])
-@pytest.mark.parametrize('power', [1, 2, 0.7])
-def test_complex_norm(complex_tensor, power):
-    expected_norm_tensor = complex_tensor.pow(2).sum(-1).pow(power / 2)
-    norm_tensor = F.complex_norm(complex_tensor, power)
-
-    torch.testing.assert_allclose(norm_tensor, expected_norm_tensor, atol=1e-5, rtol=1e-5)
+class TestComplexNorm(common_utils.TorchaudioTestCase):
+
+    @parameterized.expand([
```
krishnakalyan3, line 126: I suppose this is not quite right. `list(itertools.product(...)` might be the right way to go about this. However that would require importing `import itertools`. 

@mthrok which would be the best way to go about this?.

krishnakalyan3:
Following occurrences will be substituted (`pytest.mark.parametrize -> parameterized`)  eventually as part of this PR.

image

test/torchaudio_unittest/functional/functional_cpu_test.py
mthrok: Yeah, that's right. That's the pattern I use everywhere. `itertools` is a standard library so there is no problem.
Maybe it's good time to add a helper decorator for `@partameterized.expand(list(itertools.product` pattern, but let's not bother at this moment.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -121,16 +121,18 @@ def test_DB_to_amplitude(self):
...
+        (torch.randn(1, 2, 1025, 400, 2),  1),
+        (torch.randn(1025, 400, 2), 2)
+    ])
+    def test_complex_norm(self, complex_tensor, power):
```
mthrok, line 130: Performing `torch.randn` in decorator makes it difficult to control the randomness, so instead can you parameterize the shape?

```
def test_complex_norm(self, shape, power):
    torch.manual_seed(0)
    complex_tensor = torch.randn(*shape)
```

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -121,52 +123,57 @@ def test_DB_to_amplitude(self):
...
+    @parameterized.expand(list(itertools.product(
+        [(1, 2, 1025, 400, 2), (1025, 400, 2)],
+        [1, 2, 0.7]
+    )))
+    def test_complex_norm(self, shape, power):
+        torch.random.manual_seed(42)
+        complex_tensor = torch.randn(*shape)
+        expected_norm_tensor = complex_tensor.pow(2).sum(-1).pow(power / 2)
+        norm_tensor = F.complex_norm(complex_tensor, power)
+        torch.testing.assert_allclose(norm_tensor, expected_norm_tensor, atol=1e-5, rtol=1e-5)
```
mthrok, line 136: Can you replace `torch.testing.assert_allclose` with `self.assertEqual`?

the use of `torch.testing.assert_allclose` is discouraged now.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -81,6 +82,7 @@ def test_pitch(self, frequency):
 
 class TestDB_to_amplitude(common_utils.TorchaudioTestCase):
     def test_DB_to_amplitude(self):
+        torch.random.manual_seed(42)
```
mthrok, line 85: Let's leave this test untouched. #1113 is modifying this test suite.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -111,42 +113,49 @@ def test_amplitude_to_DB(self):
...
+            mono_complex_specgram[..., 1] * 1j
+        expected_complex_stretch = librosa.phase_vocoder(
+            mono_complex_specgram,
+            rate=rate,
+            hop_length=hop_length)
+
+        complex_stretch = complex_specgrams_stretch[index].numpy()
+        complex_stretch = complex_stretch[..., 0] + 1j * complex_stretch[..., 1]
+
+        assert np.allclose(complex_stretch, expected_complex_stretch, atol=1e-5)
```
mthrok, line 158: Can you use `self.assertEqual` here? You might need to wrap the expected using `torch.from_numpy`.

mthrok (commented):
Looks good. Made some suggestions.

mthrok (approved):

mthrok:
Thanks!

PR #1159, opened 2021-01-06
Clean up transducer build

mthrok:
This PR cleans up build process by

* Decoupling the registration of sox and transducer, using `TORCH_LIBRARY_FRAGMENT` macro.
* Instead of linking `libwarprnnt.a` from the build directory use the install directory, where other libsox related libraries are found.

When `BUILD_TRANSDUCER` is false-y value
* Do not build the external transducer
* Do not build the transducer binding (exclude it from the list of sources)

---

Remaining tweak for build (not in this PR)
* move the sox-related stuff to the dedicated directory under `third_directory` #1161 
* CXX11 ABI
* Hide symbols
* Add switch to exclude prototype from package

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -61,29 +61,31 @@ def _get_ela(debug):
 
 
 def _get_srcs():
-    return [str(p) for p in _CSRC_DIR.glob('**/*.cpp')]
+    srcs = [_CSRC_DIR / 'pybind.cpp']
+    srcs += list(_CSRC_DIR.glob('sox/**/*.cpp'))
+    if _BUILD_TRANSDUCER:
+        srcs += [_CSRC_DIR / 'transducer.cpp']
+    return [str(path) for path in srcs]
 
 
 def _get_include_dirs():
     dirs = [
         str(_ROOT_DIR),
     ]
-    if _BUILD_SOX:
+    if _BUILD_SOX or _BUILD_TRANSDUCER:
         dirs.append(str(_TP_INSTALL_DIR / 'include'))
-    if _BUILD_TRANSDUCER:
-        dirs.append(str(_TP_BASE_DIR / 'transducer' / 'submodule' / 'include'))
     return dirs
 
 
 def _get_extra_objects():
-    objs = []
+    libs = []
     if _BUILD_SOX:
         # NOTE: The order of the library listed bellow matters.
         #
         # (the most important thing is that dependencies come after a library
         # e.g., sox comes first, flac/vorbis comes before ogg, and
         # vorbisenc/vorbisfile comes before vorbis
-        libs = [
+        libs += [
             'libsox.a',
             'libmad.a',
             'libFLAC.a',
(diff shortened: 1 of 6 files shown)


third_party/transducer/CMakeLists.txt
```diff
@@ -1,20 +1,7 @@
-CMAKE_MINIMUM_REQUIRED(VERSION 3.5)
-
-PROJECT(rnnt_release)
-
-SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O2")
```
mthrok, line 5: Removing this as I do not see this reflected in build logs. `-O3` is used.

third_party/transducer/CMakeLists.txt
```diff
@@ -1,20 +1,7 @@
...
-
-IF(APPLE)
-    ADD_DEFINITIONS(-DAPPLE)
-ENDIF()
-
-INCLUDE_DIRECTORIES(submodule/include)
-
-SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
mthrok, line 13: PIC argument has to be consistent across all the libraries. The current configuration works without this explicit line, but I moved this to the top level `CMakeLitst.txt`

build_tools/setup_helpers/extension.py
```diff
@@ -97,27 +99,34 @@ def _get_extra_objects():
...
+            f'-DCMAKE_INSTALL_PREFIX={_TP_INSTALL_DIR}',
+            f'-DBUILD_SOX={"ON" if _BUILD_SOX else "OFF"}',
+            f'-DBUILD_TRANSDUCER={"ON" if _BUILD_TRANSDUCER else "OFF"}',
+            f'{_TP_BASE_DIR}'],
         cwd=build_dir,
         check=True,
     )
+    command = ['cmake', '--build', '.']
+    if _BUILD_TRANSDUCER:
+        command += ['--target', 'install']
```
mthrok, line 127: `instrall` target is only defined when building transducer.

third_party/CMakeLists.txt
```diff
@@ -1,9 +1,15 @@
-cmake_minimum_required(VERSION 3.1)
+cmake_minimum_required(VERSION 3.5)
 
 project(torchaudio_third_parties)
-include(ExternalProject)
 
-set(INSTALL_DIR ${CMAKE_CURRENT_SOURCE_DIR}/install)
+option(BUILD_SOX "Build libsox statically")
+option(BUILD_TRANSDUCER "Build transducer")
```
vincentqb, line 6: nit: "Build transducer statically" to echo libsox?
```suggestion
option(BUILD_TRANSDUCER "Build transducer statically")
```

third_party/CMakeLists.txt
```diff
@@ -1,9 +1,15 @@
...
+
+SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
vincentqb, line 8: nit: `SET` is capitalized but not the other cmake commands in this file :)

```suggestion
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```

third_party/transducer/CMakeLists.txt
```diff
@@ -30,9 +17,11 @@ ELSE()
 ENDIF()
 
 ADD_LIBRARY(warprnnt STATIC submodule/src/rnnt_entrypoint.cpp)
+target_include_directories(warprnnt PUBLIC submodule/include)
+set_target_properties(warprnnt PROPERTIES PUBLIC_HEADER submodule/include/rnnt.h)
```
vincentqb, line 21: nit: the commands are capitablized in this file :)
```suggestion
TARGET_INCLUDE_DIRECTORIES(warprnnt PUBLIC submodule/include)
SET_TARGET_PROPERTIES(warprnnt PROPERTIES PUBLIC_HEADER submodule/include/rnnt.h)
```

torchaudio/csrc/transducer.cpp
```diff
@@ -75,8 +75,19 @@ int64_t cpu_rnnt_loss(torch::Tensor acts,
...
 
+} // namespace
+
 TORCH_LIBRARY_IMPL(torchaudio, CPU, m) {
-    m.impl("rnnt_loss", &cpu_rnnt_loss);
+  m.impl("rnnt_loss", &cpu_rnnt_loss);
 }
 
-#endif
+TORCH_LIBRARY_FRAGMENT(torchaudio, m) {
```
vincentqb, line 84: `TORCH_LIBRARY_FRAGMENT ` is neat :)

vincentqb (approved):
LGTM

third_party/CMakeLists.txt
mthrok: `BUILD_SOX` has a slightly different connotation, as (ignoring Windows case) sox is bound anyway, but with `BUILD_SOX` it is binding the statically built one.

While `BUILD_TRANSDUCER` switch on/off the both build and bind, so there is no expected case where transducer library is dynamically bound.

But I guess it does not harm to be more precise.

third_party/CMakeLists.txt
mthrok: sure. I forgot to lowercase it.

third_party/transducer/CMakeLists.txt
mthrok: This is intended. only ancient CMake should use all capitals. I could have lowercased others but I left them untouched.

https://stackoverflow.com/a/32783814

PR #1161, opened 2021-01-06
Clean up libsox build and directory location

mthrok:
Now that `third_party` directory has something other than libsox, this PR adds directory dedicated for libsox and move stuff there.

Additionally, the temporary files created during build process goes to the build directory, not in the same directory as where libsox configuration is stored.


.gitignore
```diff
@@ -123,7 +123,5 @@ examples/interactive_asr/data/*.model
 examples/interactive_asr/data/*.pt
 
 # third parties
-third_party/archives/
 third_party/install/
-third_party/src/
-third_party/tmp/
```
vincentqb, line 129: nit: the folders `third_party/sox/src` and `third_party/sox/tmp` will no longer be ignored. is that intentional?

vincentqb (approved):
LGTM

.gitignore
mthrok: yes, they will go to build directory.

PR #1166, opened 2021-01-08
Support file-like object in apply_effects_file

mthrok:
The underlying implementation is same as `load` function. (both use `apply_effects_fileobj`) So this is just registering the C++ implementations, modifying frontend and adding test.

See also #1115


datumbox (approved):
LGTM!

mthrok:
thanks!

vincentqb:
@mthrok, can you confirm this is not breaking master [here](https://app.circleci.com/pipelines/github/pytorch/audio/4747/workflows/aec1a02d-89a9-46ab-b5a6-ec7a00229e2b/jobs/152468) ?

mthrok:
@vincentqb that failure is not originated from this. It is flaky in the occurrence but the same set of tests fail when it happens. I am hunting down the cause. and current hypothesis is that the underlying CCI environment is not consistent. Anyways it does not block the import you are working on.

PR #1167, opened 2021-01-08
Add SpectralCentroid transform

### LINKED ISSUE #1059: SpectralCentroid transform
## 🚀 Feature

torchaudio.transforms.SpectralCentroid

## Motivation

This is a common audio transform, included in Librosa.

## Pitch

A differentiable SpectralCentroid would be useful for audio neural networks.

## Alternatives

Here is the [librosa implementation](https://librosa.org/doc/0.8.0/_modules/librosa/feature/spectral.html#spectral_centroid).

## Additional context

Here is a simplified implementation I have made, based upon the librosa implementation:

```
def spectral_centroid(y, sr, n_fft, hop_length=512):
    S = torchaudio.transforms.Spectrogram(n_fft=n_fft, hop_length=hop_length, power=1.0)(y)

#    freq = fft_frequencies(sr=sr, n_fft=n_fft)
    freq = torch.linspace(0,
                       float(sr) / 2,
                       int(1 + n_fft//2))
                       #endpoint=True)

    if freq.ndim == 1:
        freq = freq.reshape((-1, 1))

    def tl1norm(S):
        return S / torch.sum(torch.abs(S), axis=0)
        
    # Column-normalize S
    return torch.sum(freq * tl1norm(S),
                  axis=0, keepdims=True)
```

NicolasHug:
Closes #1059

This PR adds a new `SpectralCentroid` layer, along with the `spectral_centroid` function.

I have the following design question:
Currently, `SpectralCentroid` and `spectral_centroid` take the raw waveform as input, and call `spectrogram(...)` internally. It might make more sense for them to take the spectrogram tensor as input instead? The reason I didn't go for that is that for the computation to be correct, the spectrogram must be computed with `power=1` (to get the amplitude values).

One thing I find slightly awkward with the current design is that `sample_rate` must be passed at the initialization of `SpectralCentroid` and `waveform` is only passed when calling the object. But as far as I understand, `sample_rate` and `waveform` should ideally be tied together?

#### Still TODO:

- [x] More tests
- [x] Fix Linting (probably)
- [ ] ~Allow for bounds as suggested in https://github.com/pytorch/audio/issues/1059#issuecomment-754622917~ (leaving that for another PR)
- [ ] Check if spectrogram is complex and error in this case (not sure if relevant)

CC @mthrok :)


mthrok:
Quick glance, the code looks good. Can you add torchscript compatibility test and batch consistency test?

- `torchscript_bc_test` is broken, so you can ignore it.
- I will disable the segfaulting in tests soon. It's added yesterday and causing issues.

test/torchaudio_unittest/torchscript_consistency_impl.py
```diff
@@ -624,3 +638,8 @@ def test_Vad(self):
         filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
         waveform, sample_rate = common_utils.load_wav(filepath)
         self._assert_consistency(T.Vad(sample_rate=sample_rate), waveform)
+
+    def test_SpectralCentroid(self):
+        filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
```
mthrok, line 643: Could you use either random Tensor or some synthetic wave here?
We are trying to minimize the use of external resource in test.

VAD is an exception here, as it requires audio with voice (although, since this is consistency test and not a an accuracy test, it does not need to use an actual voice, but that's a separate story).

test/torchaudio_unittest/batch_consistency_test.py
```diff
@@ -283,3 +283,14 @@ def test_batch_Vol(self):
         # Batch then transform
         computed = torchaudio.transforms.Vol(gain=1.1)(waveform.repeat(3, 1, 1))
         self.assertEqual(computed, expected)
+
+    def test_batch_spectral_centroid(self):
+        test_filepath = common_utils.get_asset_path('steam-train-whistle-daniel_simon.wav')
```
mthrok, line 288: Like in `torchscript_compatibility_test`, can you use random or pseudo waveform?
The use of I/O functions is a legacy and problematic practice.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -233,6 +233,17 @@ def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sampl
         self.assertEqual(
             torch_mfcc.type(librosa_mfcc_tensor.dtype), librosa_mfcc_tensor, atol=5e-3, rtol=1e-5)
 
+        # test SpectralCentroid
```
mthrok, line 236: Can you make this into a stand alone test method?
The code snippet above is legacy code and it's discouraged to perform multiple tests in one method.

torchaudio/transforms.py
```diff
@@ -1037,3 +1038,55 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+    __constants__ = ['sample_rate', 'n_fft', 'win_length', 'hop_length', 'pad']
+
+    def __init__(self,
+                 sample_rate: int,
+                 n_fft: int = 400,
+                 win_length: Optional[int] = None,
+                 hop_length: Optional[int] = None,
+                 pad: int = 0,
+                 window_fn: Callable[..., Tensor] = torch.hann_window,
+                 wkwargs: Optional[dict] = None) -> None:
```
mthrok, line 1072: Regarding `window_fn` and `wkwargs`, why not accept `window` like in `F.spectral_centroid`? That will simplify the interface. `window_fn` would also requires the returned value is `torch.Tensor`, otherwise it will not be able to register the buffer.

torchaudio/transforms.py
```diff
@@ -1037,3 +1038,54 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+                 pad: int = 0,
+                 window: Optional[Tensor] = None) -> None:
+        super(SpectralCentroid, self).__init__()
+        self.sample_rate = sample_rate
+        self.n_fft = n_fft
+        self.win_length = win_length if win_length is not None else n_fft
+        self.hop_length = hop_length if hop_length is not None else self.win_length // 2
+        if window is None:
+            window = torch.hann_window(self.win_length)
+        self.register_buffer('window', window)
```
NicolasHug, line 1078: I updated the signature to take `window` as input instead of `window_fn`.

Do I understand correctly that we should still be calling `register_buffer` as done here, even if the `window` tensor is allocated outside of `__init__`?

mthrok: > Do I understand correctly that we should still be calling `register_buffer` as done here, even if the `window` tensor is allocated outside of `__init__`?

I think so. Tensor object is a shared resource managed with reference counting, so in which stack frame the Tensor object should not matter, and for the purpose of serialization/deserialization of the Transform, it should be registered.

mthrok (approved):
Looks good. Thanks!

PR #1171, opened 2021-01-08
Add documentation for RNN transducer

vincentqb:
Add prototype RNN transducer loss to the documentation. This will need to be guarded off from releases.

Screen Shot 2021-01-11 at 11 48 48 AM

cc #1137

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -39,6 +39,7 @@ The :mod:`torchaudio` package consists of I/O, popular datasets and common audio
    compliance.kaldi
    kaldi_io
    utils
+   transducer
    
    
 .. toctree::
--- a/docs/source/transducer.rst
+++ b/docs/source/transducer.rst
@@ -1,0 +1,23 @@
+.. role:: hidden
+    :class: hidden-section
+
+torchaudio.prototype.transducer
+===============================
+
+.. currentmodule:: torchaudio.prototype.transducer
+
+.. note::
+
+    The RNN transducer loss is a prototype feature, see `here <https://pytorch.org/audio>`_ to learn more about the nomenclature. It is only available within the nightlies, and also needs to be imported explicitly using: :code:`from torchaudio.prototype.transducer import rnnt_loss, RNNTLoss`.
+
+rnnt_loss
+---------
+
+.. autofunction:: rnnt_loss
+
+:hidden:`RNNTLoss`
+~~~~~~~~~~~~~~~~~~
+
+.. autoclass:: RNNTLoss
+
+  .. automethod:: forward


docs/source/index.rst
```diff
@@ -39,6 +39,7 @@ The :mod:`torchaudio` package consists of I/O, popular datasets and common audio
    compliance.kaldi
    kaldi_io
    utils
+   transducer
```
mthrok, line 42: I think this `prototype` is the correct module name.

docs/source/transducer.rst
```diff
@@ -0,0 +1,21 @@
+.. role:: hidden
+    :class: hidden-section
+
```
mthrok, line 3: I think it's a good idea to note that prototype is nightly-only feature at the top of all the pages for prototype. So that there is a less chance of receiving the report like [this](https://github.com/pytorch/text/issues/1079)

docs/source/transducer.rst
```diff
@@ -0,0 +1,21 @@
...
+torchaudio.prototype.transducer
+===============================
+
+.. currentmodule:: torchaudio.prototype.transducer
```
mthrok, line 7: Did you verify the doc builds? Since prototype module has to be imported separately from `import torchaudio`, I think you need to add `import torchaudio.prototype` to doc config.

docs/source/index.rst
vincentqb: If you mean in terms of compilation, this does compile locally, see preview in description.

If you mean in terms of file organization, [torchtext](https://github.com/pytorch/text/tree/master/docs/source) also groups each prototype/experimental feature in its own doc file.

Is that what you meant?

docs/source/transducer.rst
vincentqb: I've labelled the issue as prototype and linked to the "source of truth" definition of prototype, which says "not part of release". Do you mean to say that you feel this may not be clear enough?

docs/source/transducer.rst
vincentqb: Yes, the compilation looks good to me. Do you see something missing from the preview in the description?

docs/source/transducer.rst
mthrok: Yes, I think we cannot be too careful here.

I suggest add `.. note::` or `.. warning::` and tell users to use nightly build in a more straightforward manner. 

Reading the classification, I feel like it is not even included in nightly build, which is another binary distribution.

> Prototype: These features are typically not available as part of binary distributions like PyPI or Conda, except sometimes behind run-time flags, and are at an early stage for feedback and testing.

docs/source/transducer.rst
mthrok: If it compiles, then that's okay. Let's see if the automated push also works.

docs/source/index.rst
mthrok: Looking at the `compliance.kaldi` I thought it had to be `prototype.transducer`. If the doc is compiled correctly then it's not an issue.

mthrok (approved):
Approving so that you are not blocked, can merge it and see how it is reflected at pytorch.org/audio

After the merge, you can run the nightly job manually from #306 and check how it looks on pytorch.org/audio. If you see an issue, feel free to open a new PR.

Issue #1053, opened 2020-11-20
Merge flake8 configurations


mthrok:
Currently we have `tox.ini` with only flake8-related configuration and `.flake8`.
In the workflow of torchaudio development we do not use `tox`.
I do not know the origin of them but we should merge these into `.flake8`.
PyTorch repo does not have `tox.ini` either so it's simpler for us to do the same thing.

bhargavkathivarapu:
@mthrok  Along with flake8 configurations can we include precommit hooks , so that style issues , imports ordering ,unused imports etc  won't come into commits. It seems Pytorch already has  a pre commit setup( https://github.com/pytorch/pytorch/blob/master/tools/git-pre-commit ) . May be we can extended that for this repo or use configuarations like [isort](https://github.com/PyCQA/isort) , [pre-commit](https://pre-commit.com/) along with flake8

mthrok:
@bhargavkathivarapu 

Thanks for the suggestion. However, in my experience, this kind of pre-commit hook is more counter-productive than helpful. I often push my WIP branch for the purpose of backup, because my development workflow sometimes involve multiple machines on network. I polish my changes only before publishing a PR and pre-commit hook only helps in those occasions.

Also one has to install pre-commit hook manually, so there is no automatic enforcement.

krishnakalyan3:
@mthrok, I would like to take a stab at this. 

From what I understand 
https://github.com/pytorch/audio/blob/master/.flake8
https://github.com/pytorch/audio/blob/master/tox.ini

should be merged to `.flake8` and `tox.ini` removed. Is this understanding correct?.
Is there a way to test this locally in my system?.

PR #1172, opened 2021-01-09
Merge flake8 configurations

krishnakalyan3:
#1053

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-ignore = E305,E402,E721,E741,F401,F403,F405,W503,W504,F821,F841,F999
-exclude = build,docs/source,_ext,third_party,docs/src,venv,torch/lib/gloo,torch/lib/pybind11,torch/lib/nanopb
+ignore = E305,E402,E721,E741,F403,F405,W503,W504,F999
+exclude = build,docs/source,_ext,third_party
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-ignore = E305,E402,E721,E741,F403,F405,W503,W504,F999
+ignore = E305,E402,E721,E741,F401,F403,F405,W503,W504,F999
 exclude = build,docs/source,_ext,third_party
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-ignore = E305,E402,E721,E741,F401,F403,F405,F999,W503,W504
-exclude = build,docs/source,_ext,third_party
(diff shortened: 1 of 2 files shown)


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-ignore = E305,E402,E721,E741,F401,F403,F405,F999,W503,W504
-exclude = build,docs/source,_ext,third_party
+ignore = E305,E402,E721,E741,F401,F403,F405,W503,W504,F821,F841,F999
+exclude = build,docs/source,_ext,third_party,docs/src,venv,torch/lib/gloo,torch/lib/pybind11,torch/lib/nanopb
```
mthrok, line 4: Let's remove unused ones. Only `build,docs/source,third_party` are relevant and the others are originated from PyTorch core.

.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
 max-line-length = 120
-ignore = E305,E402,E721,E741,F401,F403,F405,F999,W503,W504
-exclude = build,docs/source,_ext,third_party
+ignore = E305,E402,E721,E741,F401,F403,F405,W503,W504,F821,F841,F999
```
mthrok, line 3: What happens if `F821` and `F841` are not added here? They should not be added here I think.

mthrok: Can you also see what happens if removing 401, 402, 403, 405, 503, 504? I do not see a good reason in torchaudio that they should be ignored.

krishnakalyan3: ref https://github.com/pytorch/pytorch/blob/master/.flake8#L7

krishnakalyan3: I agree that we should remove `F821,  F841`

krishnakalyan3: `F503, F504` looks like its being ignore in pytorch too.

https://www.flake8rules.com/rules/W504.html
https://www.flake8rules.com/rules/W503.html

From what I understand they seem to be contradicting each other

krishnakalyan3: `403,405` being ignore by pytorch too. (Similar rules discouraging the use for * )


https://www.flake8rules.com/rules/F403.html
https://www.flake8rules.com/rules/F405.html

krishnakalyan3: image

 looks like we have `401` violations.

mthrok: Would it possible for you to fix it?

For `__init__.py`s, listing the imported stuff in `__all__` would do. https://stackoverflow.com/a/31079085
For others, unused import should be removed.

If that uncovers other issues, (like unused imported function is used from external module). We can just give up and do it in a follow up.

mthrok:
@krishnakalyan3 I am merging this one now. If you would like to follow up on the change to the rule. Feel free to open a new PR.

mthrok:
Thanks!

PR #1175, opened 2021-01-11
DOC: dynamically create a version pull-down using javascript

mattip:
Enhancement to create a version pulldown to switch between the documentation version URLs
```
https://pytorch.org/audio/master/
https://pytorch.org/audio/stable/ (which is a symlink to 0.7.0) and
https://pytorch.org/audio/0.7.0/
```

Adapted from pytorch/pytorch#46652 which is not yet merged. The nuances between the 2 documentation versions is not yet so pronounced in this repo, but it would be good to have this in place going forward.

- The version is now passed to sphinx from the environment, which is set via the build_docs.sh script
- The layout template has been changed
  - The javascript scripts are loaded in the header, this should be backported to the upstream [pytorch_sphinx_theme repo](https://github.com/pytorch/pytorch_sphinx_theme) since now the loading stanza appears twice on the page :(. I needed this to activate the new javascript function as the page loads.
  - A new snippet of javascript builds a `ul` list of versions on document loading (provides consistency across documentation versions)
    - The snippet will try to match the current document to the one from the other versions. If there is a match, switching versions will keep you at the same page. 
  - The version list is now below the search bar.

I could not get the page existence check to work asynchronously, so page loading time takes a hit. Is the UX improvement worth it? I was basing this off the [warning in the NumPy docs](https://numpy.org/doc/1.13/) when you view older documentation, [this](https://numpy.org/doc/1.13/reference/routines.random.html) is a page with no parallel in the latest version, so the message will pre-populate a search qeury (source code [here](https://github.com/numpy/doc/blob/master/_static/versionwarning.js_t)). maybe I missed something? In any case, the NumPy script only needs to check for the latest page, my adaptaion scans all the versions.

Help here would be welcome from people fluent in web front-end programming to style the output.


mattip:
ping @brianjo

brianjo:
> Help here would be welcome from people fluent in web front-end programming to style the output.

@ranman Could you take a look at this? Thanks!

.circleci/build_docs/build_docs.sh
```diff
@@ -2,12 +2,15 @@
...
 source ./packaging/pkg_helpers.bash
 export NO_CUDA_PACKAGE=1
 setup_env 0.8.0
 setup_wheel_python
 
 pushd docs
+VERSION=$1
 pip install -r requirements.txt
-make html
+make html
```
ranman, line 15: nit: extra space on html here

docs/source/_static/docversions.js
```diff
@@ -0,0 +1,56 @@
...
+         */ 
+        const pathPattern = /audio(\/build\/html\/|\/stable|\/master|\/[0-9.rc]+|)(.*)/;
+        const m = location.pathname.match(pathPattern);
+        base_url = 'https://pytorch.org/audio/' + versions[i];
+        if (m == null) {
+            url = base_url
+        } else {
+            url = 'https://pytorch.org/audio/' + versions[i] + m[2];
+            // Checks synchronously if the page exists, fail -> replace the link
+            $.ajax({url: url, cache: true, async: false}).fail(function(jqXHR, textStatus) {
```
ranman, line 46: Could we do an HTTP HEAD request instead of a full GET, then check if 404? Faster than trying to get the whole page.

docs/source/_static/docversions.js
```diff
@@ -0,0 +1,56 @@
...
+                url = base_url
+                });
+        }
+        document.write(open_list);
+        document.write('<a id=ID href="URL">VERSION</a> </li>\n'
+                        .replace('VERSION', label)
+                        .replace('URL', url));
+    }
```
ranman, line 54: nit: ideally we wouldn't do document.write inside the for loop but just once at the end. Probably not a big deal though.

docs/source/conf.py
```diff
@@ -80,11 +80,12 @@
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-# The short X.Y version.
-# TODO: change to [:2] at v1.0
-version = 'master '
-# The full version, including alpha/beta/rc tags.
-# TODO: verify this works as expected
+VERSION = os.environ.get('VERSION', False)
```
ranman, line 83: instead of False here could we just do the following:
`version = os.getenv('VERSION', 'master)`

ranman:
I left some comments on potential improvements. I think there's a better way to go about this but I'm not smart enough to figure it out and I don't want perfect to be the enemy of done.

mthrok:
Thanks @ranman 

@mattip Ping me once you address the comments. I can merge this and we can iterate on it with new PRs.

docs/source/_static/docversions.js
mattip: refactored to fix this and make the call asynchronous

docs/source/_static/docversions.js
mattip: refactored

docs/source/conf.py
mattip: adopted

mattip:
Fixed and rebased off master so the build should pass. Then the artifacts should be visible.

mattip:
@ranman @brianjo I added an upload for the artifacts, you can see the rendered docs with the pull down [here](https://147340-90321822-gh.circle-artifacts.com/0/docs/datasets.html). I am trying a little more to get the link in the pulldown to go to the parallel document in the other versions. What do you think of the styling? It seems it could be improved ...

ranman:
I like the changes. I’ll take a stab at the styling today and see if I can come up with anything.

mthrok:
Ping @mattip @ranman

PR #1176, opened 2021-01-12
Fix third party download cache config

mthrok:
Follow up of #1161. CircleCI cache directory was not updated and this is causing nightly to fail.


vincentqb (approved):
LGTM, please make sure to re-run the nightlies afterward to regenerate the docs.

Issue #1173, opened 2021-01-09
Add `bits_per_sample` to `info`


mthrok:
The legacy sox backend exposed libsox's structures (`sox_signalinfo_t` and `sox_encodinginfo_t`) and that allowed to access most of audio metadata. In the transition to the new backends, these structures were abstracted away and introduced `AudioMetaData` which only contain very basic information. 

Now we would like to add some metadata, in this time, with more consistent behaviors across the backends.

First, we can add `bits_per_sample` attribute. (See also https://github.com/pytorch/audio/issues/1094)
 
## Steps
Here is the gist of how we can add that.
### sox_io backend
1. the core of `info` function is implemented [here](https://github.com/pytorch/audio/blob/master/torchaudio/csrc/sox/io.cpp#L33-L50). In this function `sf->encoding.bits_per_sample` contains the `bits_per_sample` information. See also [sox_format_t](https://fossies.org/dox/sox-14.4.2/structsox__format__t.html).
2. Add `bits_per_sample` to `SignalInfo` class and getter. ([interface](https://github.com/pytorch/audio/blob/master/torchaudio/csrc/sox/io.h#L14-L26), [implementation](https://github.com/pytorch/audio/blob/7a36c5573d4fac310d694ab758052f8d5aa55ed3/torchaudio/csrc/sox/io.cpp#L13-L31))
`SignalInfo` class is the structure passed from C++ code to Python.
3. Add `bits_per_sample` to [`AudioMetaData` class](https://github.com/pytorch/audio/blob/master/torchaudio/backend/common.py#L6-L19)
This will allow client code to fetch the resulting `bits_per_sample` attributed.
4. Update [`info` function](https://github.com/pytorch/audio/blob/master/torchaudio/backend/sox_io_backend.py#L13-L35
) to pass the `bit_per_sample` from `SignalInfo` to `AudioMetaData`.

### soundfile backend
1. Update the `info` func to add the `bit_per_sample` data to `AudioMetaData`
[...]

PR #1177, opened 2021-01-12
Add bits_per_sample to info

NicolasHug:
Closes https://github.com/pytorch/audio/issues/1173

This PR adds a `bits_per_sample` attribute to `AudioMetaData` as returned by `info()`, with support for both `sox_io` and `SoundFile`. As SoundFile does not expose the bit depth directly, it is inferred from the `subtype` field as suggested in https://github.com/pytorch/audio/issues/1173.

`bits_per_sample` is left to 0 for lossy formats or when it cannot be inferred.
CC @mthrok

### DIFF
--- a/test/torchaudio_unittest/soundfile_backend/info_test.py
+++ b/test/torchaudio_unittest/soundfile_backend/info_test.py
@@ -1,3 +1,6 @@
+from unittest.mock import patch
+import warnings
+
 import torch
 from torchaudio.backend import _soundfile_backend as soundfile_backend
 from torchaudio._internal import module_utils as _mod_utils
@@ -10,8 +13,6 @@ from torchaudio_unittest.common_utils import (
     save_wav,
 )
 from .common import skipIfFormatNotSupported, parameterize
-
-import pytest
 
 if _mod_utils.is_module_available("soundfile"):
     import soundfile
@@ -104,29 +105,25 @@ class TestInfo(TempDirMixin, PytorchTestCase):
         assert info.num_channels == num_channels
         assert info.bits_per_sample == bits_per_sample
 
+    def test_unknown_subtype_warning(self):
+        """soundfile_backend.info issues a warning when the subtype is unknown
 
-def test_unknown_subtype_warning(tmp_path, monkeypatch):
-    """soundfile_backend.info issues a warning when the subtype is unknown
+        This will happen if a new subtype is supported in SoundFile: the _SUBTYPE_TO_BITS_PER_SAMPLE
+        dict should be updated.
+        """
+        dtype, sample_rate, duration, num_channels = 'float32', 16000, 1, 1
+        duration = 1
+        path = self.get_temp_path("data.wav")
+        data = get_wav_data(
+            dtype, num_channels, normalize=False, num_frames=duration * sample_rate
+        )
+        save_wav(path, data, sample_rate=16000)
 
-    This will happen if a new subtype is supported in SoundFile: the _SUBTYPE_TO_BITS_PER_SAMPLE
-    dict should be updated.
-    """
-
-    soundfile_info_original = soundfile.info
-
-    def info_wrapper(filepath):
-        # Wraps soundfile.info and sets the subtype to some unknown value
-        sinfo = soundfile_info_original(filepath)
-        sinfo.subtype = 'SOME_UNKNOWN_SUBTYPE'
-        return sinfo
-
-    monkeypatch.setattr(soundfile, "info", info_wrapper)
-
-    data = get_wav_data(
-        dtype='float32', num_channels=1, normalize=False, num_frames=16000
-    )
-    path = tmp_path / 'data.wav'
-    save_wav(path, data, sample_rate=16000)
-    with pytest.warns(UserWarning, match="subtype is unknown to TorchAudio"):
-        info = soundfile_backend.info(path)
-    assert info.bits_per_sample == 0
+        # We pretend the _SUBTYPE_TO_BITS_PER_SAMPLE dict knows no subtype / is empty.
+        # It's easier to mock the internal dict rather than to mock the entire soundfile.info function
+        with patch.dict(soundfile_backend._SUBTYPE_TO_BITS_PER_SAMPLE, {}, clear=True):
+            with warnings.catch_warnings(record=True) as w:
+                info = soundfile_backend.info(path)
+        assert len(w) == 1
+        assert "subtype is unknown to TorchAudio" in str(w[-1].message)
+        assert info.bits_per_sample == 0
--- a/test/torchaudio_unittest/soundfile_backend/info_test.py
+++ b/test/torchaudio_unittest/soundfile_backend/info_test.py
@@ -111,19 +111,17 @@ class TestInfo(TempDirMixin, PytorchTestCase):
         This will happen if a new subtype is supported in SoundFile: the _SUBTYPE_TO_BITS_PER_SAMPLE
         dict should be updated.
         """
-        dtype, sample_rate, duration, num_channels = 'float32', 16000, 1, 1
-        duration = 1
-        path = self.get_temp_path("data.wav")
-        data = get_wav_data(
-            dtype, num_channels, normalize=False, num_frames=duration * sample_rate
-        )
-        save_wav(path, data, sample_rate=16000)
+        def _mock_info_func(_):
+            class MockSoundFileInfo:
+                samplerate = 8000
+                frames = 356
+                channels = 2
+                subtype = 'UNSEEN_SUBTYPE'
+            return MockSoundFileInfo()
 
-        # We pretend the _SUBTYPE_TO_BITS_PER_SAMPLE dict knows no subtype / is empty.
-        # It's easier to mock the internal dict rather than to mock the entire soundfile.info function
-        with patch.dict(soundfile_backend._SUBTYPE_TO_BITS_PER_SAMPLE, {}, clear=True):
+        with patch("soundfile.info", _mock_info_func):
             with warnings.catch_warnings(record=True) as w:
-                info = soundfile_backend.info(path)
-        assert len(w) == 1
-        assert "subtype is unknown to TorchAudio" in str(w[-1].message)
-        assert info.bits_per_sample == 0
+                info = soundfile_backend.info("foo")
+                assert len(w) == 1
+                assert "UNSEEN_SUBTYPE subtype is unknown to TorchAudio" in str(w[-1].message)
+                assert info.bits_per_sample == 0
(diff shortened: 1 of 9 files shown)


torchaudio/backend/_soundfile_backend.py
```diff
@@ -27,7 +63,8 @@ def info(filepath: str, format: Optional[str] = None) -> AudioMetaData:
         AudioMetaData: meta data of the given audio.
     """
     sinfo = soundfile.info(filepath)
-    return AudioMetaData(sinfo.samplerate, sinfo.frames, sinfo.channels)
+    bits_per_sample = SUBTYPE_TO_BITS_PER_SAMPLE[sinfo.subtype]
```
mthrok, line 66: If PySoundFile (and the underlying libsndfile) is updated and a new format, say is added, this will raise KeyError.
This makes me wonder if we should add something like `-1` for meaning unsupported (unexpected) format. like
`SUBTYPE_TO_BITS_PER_SAMPLE.get(sinfo.subtype, -1)`

There is a tradeoff between throwing error or silently ignoring such case.
Throwing error will prevent users who only needs to know other information from using the function, but the issue is more visible to users so that they can file an issue and we can add support.
Ignoring and substituting a placeholder makes it harder for users to notice such case, while the function will keep working if they do not need an accurate `bits_per_sample` for the new format.

What do you think?

(libsox would simply fail with `RunTimeError` if it does not recognize a format)

torchaudio/backend/_soundfile_backend.py
```diff
@@ -11,6 +11,42 @@
     import soundfile
 
 
+# Mapping from soundfile subtype to number of bits per sample.
+# This is mostly heuristical and the value is set to 0 when it is irrelevant
+# (lossy formats) or when it can't be inferred.
+# The dict is inspired from
+# https://github.com/bastibe/python-soundfile/blob/744efb4b01abc72498a96b09115b42a4cabd85e4/soundfile.py#L66-L94
```
mthrok, line 18: Can you add references you looked at, for formats that are not immediately clear like ADPCMs? So that later, it's easy for other maintainer to look at it.

torchaudio/backend/_soundfile_backend.py
```diff
@@ -11,6 +11,42 @@
...
+SUBTYPE_TO_BITS_PER_SAMPLE = {
```
mthrok, line 19: Can you prefix it with "_" just in case?

mthrok (commented):
Implementation looks good. Thank you. There are (rather minor) comments I would like to hear your opinion.

mthrok:
Regarding the `download_third_parties_nix` test, it should be fixed on master branch, so rebasing or merging should resolve it.

torchaudio/backend/_soundfile_backend.py
NicolasHug: Great point, thanks for noting!

I feel like an error might be a bit annoying for users as the bit depth is probably not such a crucial info (could be wrong on this). So I went for issuing a warning instead.

Regarding the placeholder value, I'm going with 0 instead of -1 as 0 is already used to mean "irrelevant or unknown" (this is documented in `AudioMetaData`).

I don't feel strongly about any of these points so I'm happy to amend according to your comments!

torchaudio/backend/_soundfile_backend.py
```diff
@@ -27,7 +66,17 @@ def info(filepath: str, format: Optional[str] = None) -> AudioMetaData:
...
+    try:
+        bits_per_sample = _SUBTYPE_TO_BITS_PER_SAMPLE[sinfo.subtype]
+    except KeyError:
+        bits_per_sample = 0
+        warnings.warn(
+            "The {subtype} subtype is unknown to TorchAudio. As a result, the bits_per_sample "
+            "attribute will be set to 0. If you are seeing this warning, please "
+            "report by opening an issue on github (after checking for existing/closed ones). "
+            "You may otherwise ignore this warning."
+        )
```
mthrok, line 78: I think a regular `if` statement expresses the intent of the code simpler;

```python
if sinfo.subtype not in _SUBTYPE_TO_BITS_PER_SAMPLE:
    warnings.warn(...)
bits_per_sample = _SUBTYPE_TO_BITS_PER_SAMPLE.get(sinfo.subtype, 0)
```

torchaudio/backend/_soundfile_backend.py
```diff
@@ -27,7 +66,17 @@ def info(filepath: str, format: Optional[str] = None) -> AudioMetaData:
         AudioMetaData: meta data of the given audio.
     """
     sinfo = soundfile.info(filepath)
-    return AudioMetaData(sinfo.samplerate, sinfo.frames, sinfo.channels)
+    try:
+        bits_per_sample = _SUBTYPE_TO_BITS_PER_SAMPLE[sinfo.subtype]
+    except KeyError:
+        bits_per_sample = 0
+        warnings.warn(
+            "The {subtype} subtype is unknown to TorchAudio. As a result, the bits_per_sample "
```
mthrok, line 74: Did you mean the f-string here?

torchaudio/backend/_soundfile_backend.py
mthrok: > I feel like an error might be a bit annoying for users as the bit depth is probably not such a crucial info (could be wrong on this). So I went for issuing a warning instead.
> 
> Regarding the placeholder value, I'm going with 0 instead of -1 as 0 is already used to mean "irrelevant or unknown" (this is documented in `AudioMetaData`).
> 

Both points sound good to me. Thanks!

mthrok (approved):
Looks good. Thanks. I made some follow-up comments.

test/torchaudio_unittest/soundfile_backend/info_test.py
```diff
@@ -78,18 +85,49 @@ def test_ogg(self, sample_rate, num_channels):
...
+        subtype, bits_per_sample = subtype_and_bit_depth
+        soundfile.write(path, data, sample_rate, subtype=subtype)
 
         info = soundfile_backend.info(path)
         assert info.sample_rate == sample_rate
         assert info.num_frames == sample_rate * duration
         assert info.num_channels == num_channels
+        assert info.bits_per_sample == bits_per_sample
+
+    def test_unknown_subtype_warning(self):
```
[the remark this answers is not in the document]
mthrok: Actually, I think a context manager would be simpler, without the need of separate test suite.

like;
```python 
with patch.object(ProductionClass, 'method', return_value=None) as mock_method:
    thing = ProductionClass()
    thing.method(1, 2, 3)

mock_method.assert_called_once_with(1, 2, 3)
```

from https://docs.python.org/3/library/unittest.mock.html#quick-guide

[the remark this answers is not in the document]
NicolasHug: I might be missing something but I'm not sure how I can use `patch` to change the internal. logic within `soundfile.info`?

The main issue here is that `pytest`'s `monkeypatch` fixture can't be used within methods https://stackoverflow.com/a/42334155

Maybe the easiest solution is to have a single test function, so `monkeypatch` can be safely and cleanly used? That's what I pushed in the last commit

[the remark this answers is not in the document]
mthrok: Oh sorry, I missed the part that it was `pytest`'s `mokeypatch`.
Due to the technical reason within fbcode in the past, we do not use `pytest` as module in our test suites.
(This restriction no longer applies, but before we start using `pytest`, we will have to rewrite all the test infrastructure from PyTorch core, which is not trivial.)
On GitHub/CircleCI We use `pytest` command as test runner, which works fine with tests written with `unittest` module.
We restrict ourself to use `unittest` only. So we need to use `mock` and `patch` from `unittest` utility.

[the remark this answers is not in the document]
NicolasHug: >  Due to the technical reason within fbcode in the past, we do not use pytest as module in our test suites.

Ah, OK. Then seems like these instances should be removed:

```
(pt) ➜  audio git:(bits_per_sample) ✗ git grep "import pytest" 
test/torchaudio_unittest/functional/functional_cpu_test.py:import pytest
test/torchaudio_unittest/librosa_compatibility_test.py:import pytest
```

I will open a PR for that.

>  We restrict ourself to use unittest only. So we need to use mock and patch from unittest utility.

I could not find an easy way to properly patch `soundfile.info`, so I went for the equivalent strategy of mocking our internal mapping dict instead.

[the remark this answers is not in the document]
mthrok: > I could not find an easy way to properly patch `soundfile.info`, so I went for the equivalent strategy of mocking our internal mapping dict instead.

That's an interesting approach. But patching `soundfile.info` with a mock will makes it even simpler as you do not need a Tensor object.

```python
def _mock_info_func(_):
    class MockSoundFileInfo:
	samplerate = 8000
	frames = 356
	channels = 2
	subtype = 'UNSEEN_AUDIO_FORMAT'
    return MockSoundFileInfo()


from unittest.mock import patch

    def test_unknown_subtype_warning(self):
        """docstring ..."""
        with patch("soundfile.info", _mock_info_func):
            with warnings.catch_warnings(record=True) as w:
                info = soundfile_backend.info("foo")
                assert len(w) == 1
                assert "subtype is unknown to TorchAudio" in str(w[-1].message)
                assert info.bits_per_sample == 0
```

Also using the variable `w` after the context is closed feels a bit strange to me, so I would move the check inside of  `catch_warnigs` clause.

test/torchaudio_unittest/sox_io_backend/info_test.py
```diff
@@ -146,11 +155,12 @@ def test_amr_nb(self):
         sample_rate = 8000
         path = self.get_temp_path('data.amr-nb')
         sox_utils.gen_audio_file(
-            path, sample_rate=sample_rate, num_channels=num_channels, bit_depth=16, duration=duration)
+            path, sample_rate=sample_rate, num_channels=num_channels, duration=duration)
         info = sox_io_backend.info(path)
         assert info.sample_rate == sample_rate
         assert info.num_frames == sample_rate * duration
         assert info.num_channels == num_channels
+        assert info.num_channels == num_channels
```
mthrok, line 163: What is happening here?

test/torchaudio_unittest/sox_io_backend/info_test.py
```diff
@@ -146,11 +155,12 @@ def test_amr_nb(self):
         sample_rate = 8000
         path = self.get_temp_path('data.amr-nb')
         sox_utils.gen_audio_file(
-            path, sample_rate=sample_rate, num_channels=num_channels, bit_depth=16, duration=duration)
```
mthrok, line 149: Did you try to parameterize `bit_depth` here?

mthrok:
The failing tests are not related to this PR. Merging this.

PR #1158, opened 2021-01-06
Support in-memory decoding in load function

mthrok:
Split of #1108

PR #1181, opened 2021-01-15
Fix load from file object for small files and shorter bytes

mthrok:
In #1158 file-like object support was added to `load` function.
The code works an object that implements `read` protocol `read(size: int) -> bytes`.
The implementation expected that the `read` method will always returns the `bytes` with the exact requested length, but that should not be required.

This PR relax it with couple of tweaks.
1. Introduce `read_fileobj` helper function that calls `read` and retry until the requested length is fetched / EOF is reached.
1. Mark explicitly the end of file object read, and will not call the method again once it reaches the EOF.
1. Tweak the behavior around file opening (reading header and initializing decoder) so that buffer size can be changed, and  it works fine on tiny audio data.

### DIFF
--- a/torchaudio/csrc/sox/effects.cpp
+++ b/torchaudio/csrc/sox/effects.cpp
@@ -137,51 +137,65 @@ c10::intrusive_ptr<TensorSignal> apply_effects_file(
 
 #ifdef TORCH_API_INCLUDE_EXTENSION_H
 
+// Streaming decoding over file-like object is tricky because libsox operates on
+// FILE pointer. The folloing is what `sox` and `play` commands do
+//  - file input -> FILE pointer
+//  - URL input -> call wget in suprocess and pipe the data -> FILE pointer
+//  - stdin -> FILE pointer
+//
+// We want to, instead, fetch byte strings chunk by chunk, consume them, and
+// discard.
+//
+// Here is the approach
+// 1. Initialize sox_format_t using sox_open_mem_read, providing the initial
+// chunk of byte string
+//    This will perform header-based format detection, if necessary, then fill
+//    the metadata of sox_format_t. Internally, sox_open_mem_read uses fmemopen,
+//    which returns FILE* which points the buffer of the provided byte string.
+// 2. Each time sox reads a chunk from the FILE*, we update the underlying
+// buffer in a way that it
+//    starts with unseen data, and append the new data read from the given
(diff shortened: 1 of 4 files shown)


torchaudio/csrc/sox/effects.cpp
```diff
@@ -137,51 +137,65 @@ c10::intrusive_ptr<TensorSignal> apply_effects_file(
...
-  auto header = static_cast<std::string>(
-      static_cast<py::bytes>(fileobj.attr("read")(4096)));
-  memcpy(
-      static_cast<void*>(in_buf),
-      static_cast<void*>(const_cast<char*>(header.data())),
-      header.length());
+  // * This can be changed with `torchaudio.utils.sox_utils.set_buffer_size`.
+  auto capacity =
+      (sox_get_globals()->bufsiz > 256) ? sox_get_globals()->bufsiz : 256;
+  std::string buffer(capacity, '\0');
```
datumbox, line 183: Perhaps there is a better type on std instead of string to define your buffer? Perhaps `py::bytes`?

torchaudio/csrc/sox/utils.cpp
```diff
@@ -317,5 +317,37 @@ sox_encodinginfo_t get_encodinginfo(
...
+      std::ostringstream message;
+      message
+          << "Requested up to " << request << " bytes but, "
+          << "received " << chunk_len << " bytes. "
+          << "The given object does not confirm to read protocol of file object.";
+      throw std::runtime_error(message.str());
+    }
+
+    std::cerr << "req: " << request << ", fetched: " << chunk_len << std::endl;
+    std::cerr << "buffer: " << (void*)buffer << std::endl;
```
datumbox, line 342: remove logging.

datumbox (approved):
LGTM! I left only 2 non-blocking comments.

torchaudio/csrc/sox/effects.cpp
mthrok: Looking at [the implementation of `py::bytes`](https://github.com/pybind/pybind11/blob/48534089f7f1b0229bc7ae2e4d0f21dc9ad169b5/include/pybind11/pytypes.h#L1002-L1029), they do not provide direct access to the underlying memory. The only way to get an access to the memory is converting it to `std::string`. So I think it is straightforward to use `std::string`

PR #1182, opened 2021-01-18
Update setup.py to remove 2.7 ref

### LINKED ISSUE #477: Python 2 Deprecated

NicolasHug:
This PR removes alleged support to Python 2.7 in `setup.py`.
Closes  #477

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,9 @@ setup(
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Programming Language :: C++",
-        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: Implementation :: CPython",
         "Topic :: Multimedia :: Sound/Audio",
         "Topic :: Scientific/Engineering :: Artificial Intelligence"
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,6 @@ setup(
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Programming Language :: C++",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -73,7 +73,6 @@ def run(self):
         "Operating System :: Microsoft :: Windows",
         "Operating System :: POSIX",
         "Programming Language :: C++",
-        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
```
NicolasHug, line 76: Considering that `f-strings` are being used, torchaudio requires `>=3.6`. Should we be more specific and list

```py
"Programming Language :: Python :: 3.6",
"Programming Language :: Python :: 3.7",
"Programming Language :: Python :: 3.8",

```

?

mthrok: Thanks for the suggestion. I think that makes sense.

(FYI: The reason we can use f-string is that PyTorch core does not support 3.5. The support for 3.5. was dropped in July 2020 release IIRC)

mthrok (approved):
Looks good. Thanks!

PR #1190, opened 2021-01-20
Clean up sox/CMakeLists and build log

mthrok:
This PR 

1. redirects build logs for sox and its dependencies to file so that build log becomes simpler.
2. Tweak `third_party/sox/CMakeLists.txt` for improved readability.


datumbox (approved):
Stamping

PR #1187, opened 2021-01-20
Switch to cmake for build

mthrok:
Supersedes #1068

PR #1191, opened 2021-01-20
Update GPU test image to development

mthrok:
Changing the test Docker image for GPU tests to use Nvidia Docker `devel` image. This is necessary for #1187.


datumbox (approved):
Stamping

PR #1192, opened 2021-01-20
Disable torchscript BC test

mthrok:
TorchScript BC test checks that if dumped object from the previous version is compatible with the latest master. Disabling this test for the following reasons;
1. PyTorch does not guarantee backward compatibility of serialized TorchScript object across versions.
2. This test fails with OOM
3. BC-braking change was introduced in #1141 but there is no practical way to update a test.


datumbox (approved):
Stamping. This one seems to have additional failing tests, so please investigate prior merge. :)

mthrok:
> Stamping. This one seems to have additional failing tests, so please investigate prior merge. :)

These failures are flaky and I am investigating it. It was introduced in #1158.

PR #1193, opened 2021-01-21
Update unit test base Docker image

mthrok:
* Do not install system-wide sox, so that it would not get mixed up with static one torchaudio builds
* Do not install system-wide cmake. It's installed at test. 
* Move libsndfile installation to Docker image


datumbox (approved):
Stamp!

Issue #1183, opened 2021-01-18
Codec-based augmentation


mthrok:
With the support for file-like object in I/O function, torchaudio can apply codecs as a form of augmentation.
There are two components required to add such feature to torchaudio.

1. Python frontend
    The user-facing function.
2. Extend format for `save` function.
    Currently the audio format that "sox_io" save function supports is limited to certain formats. (wave, mp3, flac, opus, amb, amr-nb, sphere, ogg/vorbis), while the underlying libsox can handle more formats. We need to extend the supported formats.

We welcome the open source community contribution. If you are interested in working on this, please read the following description and leave a comment on which part you would like to work, so other people would not do a duplicated work. If you are interested in working on extended format support on `save` function, open one PR for one format.

### Implementation

The gist of the Python frontend looks like this. Save the given waveform in memory with codec and compression applied, then load it.

```python
# in torchaudio/functional/functional.py
import io


def apply_codec(waveform, sample_rate, format, channels_first=True, compression=None):
    bytes = io.BytesIO()
    torchaudio.save(bytes, waveform, sample_rate, channels_first, compression=compression, format=format)
    bytes.seek(0)
    waveform, _ = torchaudio.load(bytes, channels_first=channels_first)
    return waveform
```

### Testing

For a starter, we need a smoke test that verifies that the function does not crush for a variety of formats. Since we want this function to be able to work in Windows system, where libsox is not available, we need to test for "sox_io" backend and "soundfile" backend with new interface. Therefor, adding a base class where the test logic is implemented, and extending the class for "sox_io" backend and "soundfile" backend would do.

```python
class ApplyCodecTestBase:
    def test_codec(self, format, compression):
        # run the function 
        # maybe check the channels (number of frames can change depending on format like mp3)

class ApplyCodecSoxIOTest(ApplyCodecTestBase):
    backend = "sox_io"
    # parameterize the compression
    def test_mp3(self, compression):
        self.test_codec("mp3", compression)

class ApplyCodecSoundfileTest(ApplyCodecTestBase):
    backend = "soundfile-new"
    # similar to the above
```

For the detail of what parameters to be parameterized and how, see  the existing test;
https://github.com/pytorch/audio/blob/f1d8d1e0da44f6503c01ddcae2e40772b400ea2d/test/torchaudio_unittest/sox_io_backend/info_test.py#L56-L153

### Implementation
The save function is implemented in C++ backend. The following is the call stack.

`torchaudio.save`
[...]

PR #1194, opened 2021-01-21
[WIP] apply codec-based data augmentation

AzizCode92:
This PR addresses the issue #1183

### DIFF
--- a/.circleci/unittest/linux/scripts/run-clang-format.py
+++ b/.circleci/unittest/linux/scripts/run-clang-format.py
@@ -1,0 +1,340 @@
+#!/usr/bin/env python
+"""A wrapper script around clang-format, suitable for linting multiple files
+and to use for continuous integration.
+
+This is an alternative API for the clang-format command line.
+It runs over multiple files and directories in parallel.
+A diff output is produced and a sensible exit code is returned.
+
+"""
+
+import argparse
+import codecs
+import difflib
+import fnmatch
+import io
+import multiprocessing
+import os
+import signal
+import subprocess
+import sys
+import traceback
+
+from functools import partial
+
+try:
+    from subprocess import DEVNULL  # py3k
+except ImportError:
+    DEVNULL = open(os.devnull, "wb")
+
+
+DEFAULT_EXTENSIONS = 'c,h,C,H,cpp,hpp,cc,hh,c++,h++,cxx,hxx,cu'
+
+
+class ExitStatus:
+    SUCCESS = 0
+    DIFF = 1
+    TROUBLE = 2
+
+
+def list_files(files, recursive=False, extensions=None, exclude=None):
+    if extensions is None:
+        extensions = []
+    if exclude is None:
+        exclude = []
+
+    out = []
+    for file in files:
+        if recursive and os.path.isdir(file):
+            for dirpath, dnames, fnames in os.walk(file):
+                fpaths = [os.path.join(dirpath, fname) for fname in fnames]
+                for pattern in exclude:
+                    # os.walk() supports trimming down the dnames list
+                    # by modifying it in-place,
+                    # to avoid unnecessary directory listings.
+                    dnames[:] = [
+                        x for x in dnames
+                        if
+                        not fnmatch.fnmatch(os.path.join(dirpath, x), pattern)
+                    ]
+                    fpaths = [
+                        x for x in fpaths if not fnmatch.fnmatch(x, pattern)
+                    ]
+                for f in fpaths:
+                    ext = os.path.splitext(f)[1][1:]
+                    if ext in extensions:
+                        out.append(f)
+        else:
+            out.append(file)
+    return out
+
+
+def make_diff(file, original, reformatted):
+    return list(
+        difflib.unified_diff(
+            original,
+            reformatted,
+            fromfile='{}\t(original)'.format(file),
+            tofile='{}\t(reformatted)'.format(file),
+            n=3))
+
+
+class DiffError(Exception):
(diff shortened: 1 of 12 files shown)


AzizCode92:
Hi @mthrok, I have couple of questions:
1. where should I include the testing of this code? inside `test/torchaudio_unittest/functional/functional_impl.py` ?

For the test, I have done something like this 

```python
class CodecTestBase:
    @parameterized.expand(list(itertools.product(
        ["mp3", "wav", "flac"],
        [96, 128, 160, 192, 224, 256, 320],
    )), name_func=name_func)
    def test_codec(self, format, compression):
        torch.random.manual_seed(42)
        waveform = torch.rand(2, 44100 * 1)
        augmented_waveform = F.apply_codec(waveform, format, channels_first=True, compression=compression)
        # TODO: maybe check the channels (number of frames can change depending on format like mp3)

 ```
2. Is there a way I can verify the channels given the output waveform??
3.  should we also include the `channels_first` to the parameterized decorator?

**Comment by mthrok** [comment]
> Hi @mthrok, I have couple of questions:
> 
> 1. where should I include the testing of this code? inside `test/torchaudio_unittest/functional/functional_impl.py` ?
> 
> For the test, I have done something like this
> 
> ```python
> class CodecTestBase:
>     @parameterized.expand(list(itertools.product(
>         ["mp3", "wav", "flac"],
>         [96, 128, 160, 192, 224, 256, 320],
>     )), name_func=name_func)
>     def test_codec(self, format, compression):
>         torch.random.manual_seed(42)
>         waveform = torch.rand(2, 44100 * 1)
>         augmented_waveform = F.apply_codec(waveform, format, channels_first=True, compression=compression)
>         # TODO: maybe check the channels (number of frames can change depending on format like mp3)
> ```
> 
> 1. Is there a way I can verify the channels given the output waveform??
> 2. should we also include the other parameters like `channels_first`  and `compression` to the parameterized decorator?

Hi @AzizCode92 

Thanks for working on this. The implementation looks good so far.
To answer your question,
1. the test should be placed in ` test/torchaudio_unittest/functional/functional_cpu_test.py` because it's CPU only feature.
2. For the parameters, it's nice to have them covered, but we do not need to cover all the combination of formats and parameters. I suggest to add anther method that does the same thing as what you are describing above for  `format="wav"` and `channels_first=False`.

By the way, looking at the docstring you added,  I am wondering if we should make this feature require `sox_io` backend (means not supported on Windows) or allow the use of `soundfile` backend. With `soundfile` backend, the number of formats supported are very limited and the `compression` parameter has no effect. What do you think?

**Comment by AzizCode92** [comment]
Hi @mthrok, Thank you for your feedback.

> By the way, looking at the docstring you added, I am wondering if we should make this feature require sox_io backend (means not supported on Windows) or allow the use of soundfile backend. With soundfile backend, the number of formats supported are very limited and the compression parameter has no effect. What do you think?

Hmm, do the two parameters `compression` and `format`  have an effect on the data-augmentation we are trying to make? if yes, I guess we should make it require the `sox_io` backend.

**Comment by AzizCode92** [comment]
Now I am working on the unit-test. So for the smoke test, my idea is to save the augmented data, load it again and assert the sample_rate, the expected and the found data.
But here, inside the smoke test, what format should I use? what save/load methods should I use?

thanks

**Comment by mthrok** [comment]
Hi @AzizCode92 

> Hmm, do the two parameters `compression` and `format` have an effect on the data-augmentation we are trying to make? if yes, I guess we should make it require the `sox_io` backend. Otherwise, we would limit the parameters we can finetune to do the data-augmentation. Or we can include both `sox_io` and `soundfile` backends?

`compression` and `format` are the key ingredients for this augmentation. For example, when you want to train an ASR model for telephone conversation (say, for a call center application) and you do not have data collected from the particular environment you want to deploy the application, you need to train your initial model with something. If you have a clean dataset (like collected conversation from television) you can apply codecs (also background noise and reverberation) to make them sound like the conversation is happening over the phone.

Now I think that we can restrict the function to "sox_io" backend only although the code itself is compatible with "soundfile" backend, the lack of support for `compression` parameter make it not much useful for "soundfile" backend. You can use [this decorator](https://github.com/pytorch/audio/blob/99ed7183df3df8278cc46de432c31b254e31573e/torchaudio/backend/sox_io_backend.py#L13) to achieve that.

> Now I am working on the unit-test. So for the smoke test, my idea is to save the augmented data, load it again and assert the sample_rate, the expected and the found data. Is it the proper way to do it??
> Also inside the smoke test, what format should I use? what save/load methods should I use?

My view is that since `load` and `save` functions that are used inside of the implementation are separately and rigorously tested, and the new function does not do much more than `save` and `load`, we can just check that the function does not crush. (and some very cheap check like `sample_rate` as you suggested). Although the implementation detail around the use of `save` and `load` functions can change anytime, so it's not the best approach but I think for now that's enough for us to getting started.

**Comment by mthrok** [comment]
BTW, regarding the sample rate, I realized that some codecs require specific sample rate. (like AMR NB requires 8kHz) So we will need to incorporate this into the design later. (which will be a separate work than this PR)

For example, if I convert a WAV file into AMR NB with `sox` command, I get the following;

```
sox -V3 foo.wav foo.amr-nb

...

sox WARN formats: amr-nb can't encode at 32000Hz; using 8000Hz

...

sox INFO sox: effects chain: input        32000Hz  1 channels
sox INFO sox: effects chain: rate          8000Hz  1 channels
sox INFO sox: effects chain: dither        8000Hz  1 channels
sox INFO sox: effects chain: output        8000Hz  1 channels
```

PR #1198, opened 2021-01-25
Fix clang-format

mthrok:
clang-format job was checking if there is a difference after `clang-format` command  was used. This overlooked the case where the `clang-format` command was not found. This PR fixes it and fix C++ lint issue.

See example of failure case.
https://app.circleci.com/pipelines/github/pytorch/audio/4686/workflows/487fc3d8-037a-4662-b251-2f6ae4292d13/jobs/150112


datumbox (approved):
LGTM, stamping.

PR #1200, opened 2021-01-26
Apply codec-based data augmentation

AzizCode92:
This PR addresses the issue #1183


AzizCode92:
Hi @mthrok, this the baseline code I have made.
I apologize beforehand for accidentally closing the previous [PR](https://github.com/pytorch/audio/pull/1194).

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -1,13 +1,24 @@
...
+    skipIfNoExec,
+)
+from torchaudio_unittest.sox_io_backend.common import name_func
+
+import torchaudio
+import torchaudio.functional as F
+from torchaudio._internal import (
+    module_utils as _mod_utils,
+)
+from torchaudio.backend import sox_io_backend
```
mthrok, line 21: Is it your IDE that is sorting the import statements?
My interpretation of PEP8 here is that `torchaudio_unittest` is the local library so it should be the third group and `torch`, `torchaudio` and other libraries used for testing are the second group.
So I prefer `torchaudio` comes before `torchaudio_unittest`. But if this is something your IDE does automatically, I won't argue.

https://www.python.org/dev/peps/pep-0008/#imports
> Imports should be grouped in the following order:
>
> 1. Standard library imports.
> 2. Related third party imports.
> 3. Local application/library specific imports.
You should put a blank line between each group of imports.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -179,3 +191,31 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
 
         assert mask_specgrams.size() == specgrams.size()
         assert (num_masked_columns < mask_param).sum() == num_masked_columns.numel()
+
+
+@skipIfNoExec('sox')
+class ApplyCodecTestBase(TempDirMixin, TorchaudioTestCase):
```
mthrok, line 197: Can you rename this to `TestApplyCodec`? one thing, this is not a base class, the second, `TestXXX` is the common pattern used in torchaudio's test suite.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -179,3 +191,31 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
+        """
+        path = self.get_temp_path(f'data.{format}')
+        torch.random.manual_seed(42)
+        waveform = torch.rand(2, 44100 * 1)
+        sample_rate = 8000
+        augmented_data = F.apply_codec(waveform, sample_rate, format=format, channels_first=True,
+                                       compression=compression)
+        save_wav(path, augmented_data, sample_rate)
+        info = sox_io_backend.info(path)
+        assert info.sample_rate == sample_rate
```
mthrok, line 213: Since the data is saved with the given `sample_rate`. This assertion always passes. 
Since the `apply_codec` works in-memory manner, it is unnecessary to save data on file system. (`TempDirMixin` is not necessary)

Since it is not simple to check if the resulting data are correct,  I suggest that you check the shape (number of channels and conditionally number of frames).

```python
def _smoke_test(self, format, compression, *, check_num_frames):
    torch.random.manual_seed(42)
    sample_rate = 8000
    num_frames = 3.0 * sample_rate
    num_channels = 2
    waveform = torch.rand(num_channels, num_frames)
    augmented = F.apply_codec(
        waveform, sample_rate, format, channels_first=True, compression=compression)
    assert augmented.dtype == waveform.dtype
    assert augmented.shape[0] == num_channels
    if check_num_frames:
        assert augmented.shape[1] == num_frames
```

then for each formats,

```python
def test_wave(self):
    self._smoke_test("wav", compression=None, check_num_frames=True)

@parameterized.expand([96, ...])
def test_mp3(self, compression):
    self._smoke_test("mp3", compression, check_num_frames=False)
```

Note that WAVE is uncompressed format, so providing `compression` argument makes no difference.

Try to add other formats that are tested in `save_test` file, like `"opus"`. They have different valid value range for compression argument.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -179,3 +191,31 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
+
+    @_mod_utils.requires_module('torchaudio._torchaudio')
```
mthrok, line 215: You can use `@skipIfNoExtension` instead of `@_mod_utils.requires_module`, and you can put it on the class definition so that you do not need to do it on every method.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -179,3 +191,31 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
 
         assert mask_specgrams.size() == specgrams.size()
         assert (num_masked_columns < mask_param).sum() == num_masked_columns.numel()
+
+
+@skipIfNoExec('sox')
```
mthrok, line 196: `sox` command should not be required. This is a smoke test that does not require any external tool.

torchaudio/functional/functional.py
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
                            device=specgram.device).reshape((-1, 1))
     freq_dim = -2
     return (freqs * specgram).sum(dim=freq_dim) / specgram.sum(dim=freq_dim)
+
+
+def apply_codec(waveform, sample_rate, format, channels_first=True, compression=None) -> Tensor:
```
mthrok, line 980: Can you reorder the arguments like `waveform, sample_rate, compression=None, format, channels_first=True`?

The expected usage will change the value for `compression`, while `channels_first` will not be changed (at all).

torchaudio/functional/functional.py
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
...
+                * | ``MP3``: Either bitrate (in ``kbps``) with quality factor, such as ``128.2``, or
+                  | VBR encoding with quality factor such as ``-4.2``. Default: ``-4.5``.
+                * | ``FLAC``: compression level. Whole number from ``0`` to ``8``.
+                  | ``8`` is default and highest compression.
+                * | ``OGG/VORBIS``: number from ``-1`` to ``10``; ``-1`` is the highest compression
+                  | and lowest quality. Default: ``3``.
+            See the detail at http://sox.sourceforge.net/soxformat.html.
+
+    Returns:
+        Tensor: Dimension (..., time)
```
mthrok, line 1002: Similar to the above `(..., time)` is not accurate. Can you also mention that the number of frames might change for certain codecs?

torchaudio/functional/functional.py
```diff
@@ -848,11 +852,11 @@ def detect_pitch_frequency(
...
-    waveform: Tensor,
-    cmn_window: int = 600,
-    min_cmn_window: int = 100,
-    center: bool = False,
-    norm_vars: bool = False,
+        waveform: Tensor,
+        cmn_window: int = 600,
+        min_cmn_window: int = 100,
+        center: bool = False,
+        norm_vars: bool = False,
```
mthrok, line 859: (In general, it is a common practice not to touch the part unrelated to the main goal of the PR. You might be asked to revert it.)

mthrok (commented):
Hi @AzizCode92 

Thanks for working on this. It is looking good.

torchaudio/functional/functional.py
AzizCode92: Sorry, I didn't get this comment correctly. 
Since number of frames was not used inside this method, I didn't understand how to add it in the return statement inside the docstrings. Is it part of the shape of the output tensor? is our output tensor something like (..., num_frames)?

torchaudio/functional/functional.py
AzizCode92: Yes, you're totally right.
I think my IDE did it when I make code reformat.

test/torchaudio_unittest/functional/functional_cpu_test.py
AzizCode92: I have added `vorbis` and `flac`. By adding `opus` i got `RuntimeError: Unsupported file type: opus`?

AzizCode92: I also got an error inside the unit-test for the cases of mp3 files.
Something like this ```RuntimeError: Error loading audio file: failed to open file.```

AzizCode92:
Hi @mthrok, thanks a lot for you detailed feed-back.

torchaudio/functional/functional.py
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
                            device=specgram.device).reshape((-1, 1))
     freq_dim = -2
     return (freqs * specgram).sum(dim=freq_dim) / specgram.sum(dim=freq_dim)
+
+
+def apply_codec(waveform, sample_rate, compression, format, channels_first=True) -> Tensor:
```
mthrok, line 980: 1. Can you annotate the signature so that it is TorchScript compatible?
2. Can you reorder `compression` and `format`? `compression` depends on `format`.
3. Can you make compression parameter optional so that it defaults to `None` (and update the docstring)?
4. Can you decorate the function with `requires_module` so that this will be only available to torchaudio installation with C++ extension, as we discussed?

```python
def apply_codec(
    waveform: Tensor,
    sample_rate: int,
    format: str,
    compression: Optional[float] = None,
    channels_first: bool = True,
) -> Tensor:
```

torchaudio/functional/functional.py
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
...
+                  | ``8`` is default and highest compression.
+                * | ``OGG/VORBIS``: number from ``-1`` to ``10``; ``-1`` is the highest compression
+                  | and lowest quality. Default: ``3``.
+            See the detail at http://sox.sourceforge.net/soxformat.html.
+
+    Returns:
+        Tensor
+    """
+    bytes = io.BytesIO()
+    torchaudio.save(bytes, waveform, sample_rate, channels_first, compression=compression, format=format)
```
mthrok, line 1005: I realized that if the backend is set to `soundfile` it will use `soundfile`.
Can you use `torchaudio.backend.sox_io_backend.save` instead?

torchaudio/functional/functional.py
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
...
+    bytes.seek(0)
+    waveform, _ = torchaudio.load(bytes, channels_first=channels_first)
```
mthrok, line 1007: I realized that there are certain formats that has limitation on sample rate.

Can you replace `load` with `sox_effects.apply_effects_file`, and provide `[["rate", f"{sample_rate}"]]` so that they are re-sampled to original sample rate if necessary?

 ```python
augmented, _ = torchaudio.sox_effect.apply_effects_file(
    bytes, effects=[["rate", f"{sample_rate}"]], channels_first=channels_first, format=format)
return augmented
```

File: test/torchaudio_unittest/functional/functional_cpu_test.py (hunk 3 above)
Comment (mthrok) [reply test/torchaudio_unittest/functional/functional_cpu_test.py]: Oh sorry, I have forgotten that `torchaudio` can only read OPUS but cannot write.

File: torchaudio/functional/functional.py [hunk 12]
```diff
@@ -971,3 +975,34 @@ def spectral_centroid(
...
+def apply_codec(waveform, sample_rate, compression, format, channels_first=True) -> Tensor:
+    r"""
+    Applies codecs as a form of augmentation
+
+    Args:
+        waveform (Tensor): Audio data. Must be 2 dimensional. See also ```channels_first```
+        sample_rate (int): Sample rate of the audio waveform
+        format (str): file format
+        channels_first (bool):
+            When True, the returned Tensor has dimension ``[channel, time]``.
```
mthrok, line 989: Actually, `channels_first` value is for both input and returned Tensor. It tells which format the input Tensor is, then return the resulting Tensor in the same manner.

torchaudio/functional/functional.py
mthrok: You can use the something similar to [the docstring of I/O function](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/torchaudio/backend/sox_io_backend.py#L155-L159). If ``channels_first=True``, it has
            ``[channel, time]`` else ``[time, channel]``.

mthrok:
If you have time, adding tests for TorchScript consistency (like [this one](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/torchscript_consistency_impl.py#L538-L550)) would be nice.

mthrok:
Can you incorporate the changes from the latest master? I think the test failures will be fixed by #1181.

AzizCode92:
Hi @mthrok. 
The unit-test for `apply_codec` are green now.
I have added also a test for TorchScript consistency.
The PR is red due to an RuntimeError of the TorchScript consistency tests.
The error message looks something like this 
```E           
RuntimeError:            
E           save(str filepath, Tensor src, int sample_rate, bool channels_first=True, float? compression=None, str? format=None, str? dtype=None) -> (None):
E           Expected a value of type 'str' for argument 'filepath' but instead found type 'Tensor'.
```

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -548,6 +548,22 @@ def func(tensor):
         tensor = common_utils.get_whitenoise(sample_rate=44100)
         self._assert_consistency(func, tensor)
 
+    def test_apply_codec(self):
+
+        def func(tensor):
+            sample_rate = 8000,
+            format = "wav",
+            compression = None,
```
mthrok, line 556: The trailing commas here make the TorchScript compiler think that they are `tuple` type, thus the test is failing. Removing the commas should make the test work.

torchaudio/functional/functional.py
```diff
@@ -994,6 +1000,47 @@ def spectral_centroid(
...
+            Otherwise, they have dimension ``[time, channel]``.
+        compression (float):
+            Used for formats other than WAV. This corresponds to ``-C`` option of ``sox`` command.
+                * | ``MP3``: Either bitrate (in ``kbps``) with quality factor, such as ``128.2``, or
+                  | VBR encoding with quality factor such as ``-4.2``. Default: ``-4.5``.
+                * | ``FLAC``: compression level. Whole number from ``0`` to ``8``.
+                  | ``8`` is default and highest compression.
+                * | ``OGG/VORBIS``: number from ``-1`` to ``10``; ``-1`` is the highest compression
+                  | and lowest quality. Default: ``3``.
+            See the detail at http://sox.sourceforge.net/soxformat.html.
```
mthrok, line 1029: I am making updates to the original docstring of compression parameter. Instead of copy-pasting, can you redirect? Something like ``See :py:func:`torchaudio.backend.sox_io_backend.save`.`` should do.

test/torchaudio_unittest/functional/functional_cpu_test.py
```diff
@@ -211,3 +218,47 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
+                                  channels_first=True)
+        assert augmented.dtype == waveform.dtype
+        assert augmented.shape[0] == num_channels
+        if check_num_frames:
+            assert augmented.shape[1] == num_frames
+
+    def test_wave(self):
+        self._smoke_test("wav", compression=None, check_num_frames=True)
+
+    @parameterized.expand(list(itertools.product(
```
mthrok, line 248: I think `list(itertools.product(` is redundant here.

mthrok (commented):
Hi @AzizCode92 

Thanks for the work. Please refer to the comments. Once the tests pass, I think we can merge this.

torchaudio/functional/functional.py
mthrok: Also, the order of arguments in docstring does not match with the actual order.

mthrok:
Hi @AzizCode92 

I have added `encoding` and `bits_per_sample` options to `save` function. Can you also add the same parameters and pass them to `apply_codec` function?

AzizCode92:
Hi @mthrok, 
Sure, I am just having issues with building torchaudio from source for the moment.
By the way, I opened an [issue](https://github.com/pytorch/audio/issues/1266) for that.
Once solved, I can continue working on my PR again

AzizCode92:
Hi @mthrok, I need some help here regarding the `torchscript_consistency_impl.py`.
There are some RuntimeError caused by the `torchaudio.backend.sox_io_backend.save`.
Here is a copy of the Error message:

```
E           RuntimeError: 
E           
E           save(str filepath, Tensor src, int sample_rate, bool channels_first=True, float? compression=None, str? format=None, str? encoding=None, int? bits_per_sample=None) -> (None):
E           Expected a value of type 'str' for argument 'filepath' but instead found type 'Tensor'.
E           :
E             File "/root/project/env/lib/python3.7/site-packages/torchaudio-0.8.0a0+ec8326e-py3.7-linux-x86_64.egg/torchaudio/functional/functional.py", line 1035
E               """
E               bytes = io.BytesIO()
E               torchaudio.backend.sox_io_backend.save(bytes,
E               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
E                                                      waveform,
E                                                      sample_rate,
E           'apply_codec' is being compiled since it was called from 'func'
E             File "/root/project/test/torchaudio_unittest/functional/torchscript_consistency_impl.py", line 559
E                       compression = None
E                       channels_first = True
E                       return F.apply_codec(tensor,
E                              ~~~~~~~~~~~~~~~~~~~~~
E                                            sample_rate,
E                                            ~~~~~~~~~~~~
E                                            format,
E                                            ~~~~~~~
E                                            compression,
E                                            ~~~~~~~~~~~~
E                                            channels_first)
E                                            ~~~~~~~~~~~~~~ <--- HERE

../env/lib/python3.7/site-packages/torch/jit/_script.py:995: RuntimeError
```

mthrok:
> Hi @mthrok, I need some help here please regarding the `torchscript_consistency_impl.py`.
> There is a RuntimeError caused by the `torchaudio.backend.sox_io_backend.save`.
> Here is a copy of the Error message:
> 
> ```
> E           RuntimeError: 
> E           
> E           save(str filepath, Tensor src, int sample_rate, bool channels_first=True, float? compression=None, str? format=None, str? encoding=None, int? bits_per_sample=None) -> (None):
> E           Expected a value of type 'str' for argument 'filepath' but instead found type 'Tensor'.
> E           :
> E             File "/root/project/env/lib/python3.7/site-packages/torchaudio-0.8.0a0+ec8326e-py3.7-linux-x86_64.egg/torchaudio/functional/functional.py", line 1035
> E               """
> E               bytes = io.BytesIO()
> E               torchaudio.backend.sox_io_backend.save(bytes,
> E               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
> E                                                      waveform,
> E                                                      sample_rate,
> E           'apply_codec' is being compiled since it was called from 'func'
> E             File "/root/project/test/torchaudio_unittest/functional/torchscript_consistency_impl.py", line 559
> E                       compression = None
> E                       channels_first = True
> E                       return F.apply_codec(tensor,
> E                              ~~~~~~~~~~~~~~~~~~~~~
> E                                            sample_rate,
> E                                            ~~~~~~~~~~~~
> E                                            format,
> E                                            ~~~~~~~
> E                                            compression,
> E                                            ~~~~~~~~~~~~
> E                                            channels_first)
> E                                            ~~~~~~~~~~~~~~ <--- HERE
> 
> ../env/lib/python3.7/site-packages/torch/jit/_script.py:995: RuntimeError
> ```
> 
> Any hints please?

Hi @AzizCode92 

I am sorry, what I asked you was totally wrong. The said function is not compatible with TorchScript. (TorchScript compiler does not support Python-specific thing like file-like object). So adding TorchScript compatibility test is an impossible task. So sorry about that.

torchaudio/functional/functional.py
```diff
@@ -994,6 +1000,53 @@ def spectral_centroid(
...
 
+@_mod_utils.requires_module('torchaudio._torchaudio')
+def apply_codec(
+    waveform: Tensor,
+    sample_rate: int,
+    format: str,
+    compression: Optional[float] = None,
+    channels_first: bool = True,
+    encoding: Optional[str] = None,
+    bits_per_sample: Optional[int] = None,
```
mthrok, line 1011: Can you reorder the parameters as `waveform, sample_rate, channels_first, format, encoding, bits_per_sample`?

Sorry I am reverting what I originally asked (moving `channels_first` after `format`) but grouping format related parameters together looks nicer.

mthrok (commented):
Please remove TorchScript compatibility test and re-order the parameters (and docstring). Sorry about the flip-flopping the direction.
Other than that, I think it's good.

torchaudio/functional/functional.py
AzizCode92: Moving the `channels_first` (default_parameter) before `format` (non default_parameter) throws a syntax-error ```SyntaxError: non-default argument follows default argument```.
Also what about the `compression` parameter?

mthrok: Sorry, I made an invalid suggestion again.
let's do `waveform, sample_rate, format, channels_first, compression, encoding, bits_per_sample`

AzizCode92:
@mthrok, no worries :) 
Thanks a lot for your feedback.
I guess my PR is ready to be merged now.

mthrok (approved):
Looks good. Thank you for your contribution.
FYI, there is a good chance that we have to exclude this feature from the upcoming release because of the underlying instability in file-like object which needs attention. (#1229) I will try my best to hunt down the cause.

mthrok:
Thanks!

mthrok:
Hi @AzizCode92 

Checkout the new tutorial which demonstrates how to use `apply_codec` function. Try it out with Colab.

https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html#data-augmentation

AzizCode92:
Hi @mthrok, Sure I will do it asap.
Thanks and I am very happy that this feature is out there.

Issue #1197, opened 2021-01-25
Adding target `dtype` argument to `save` function


mthrok:
In `"sox_io"` backend and `"soundfile"` backend with new interface, when saving a Tensor to file with`save` function, it is expected to that client code converts the Tensor to the dtype with desired precision. It was pointed out that this is somewhat inconvenient and `save` function should handle type conversion for improved UX. (See #903)

To address this, we would like to add a new parameter `dtype` to `save` function so that dtype conversion is handled by `save` function.

## Specification

- `dtype` is optional string type, valid values are (`"uint8"`, `"int16"`, `"int32"`, `"float32"` and `None`)
- `dtype=None` means no conversion is performed.
- `dtype` parameter is only effective for float Tensor.
  - If user provide a Tensor with other dtype, (like `uint8`), `dtype` parameter is ignored and no conversion is performed.

In release 0.8
- The default value of `dtype` is `None`, for the sake of backward compatibility.

In release 0.9
- The default value of `dtype` is changed to `int16`.

## Migration plan

<table class="tg">
<thead>
  <tr>
    <th class="tg-4e34" colspan="3"><span style="font-weight:600">Release</span></th>
  </tr>
</thead>
<tbody>
  <tr>
    <td class="tg-q7lb">0.7</td>
    <td class="tg-q7lb">0.8 (Default to `dtype=None`)</td>
    <td class="tg-q7lb">0.9 (Default to `dtype=int16`)</td>
  </tr>
  <tr>
    <td class="tg-q7lb">The precision of saved audio <br>depends on the input Tensor dtype.</td>
    <td class="tg-q7lb"><ul>
<li>`dtype` parameter is added to `save` function</li>
<li>If the input Tensor is float type and `dtype` is `None`, <br>&nbsp;&nbsp;&nbsp;then issue warning saying the default value will be changed to `int16`.</li>
</ul></td>
    <td class="tg-q7lb"> `dtype` parameter defaults to `int16`.</td>
  </tr>
</tbody>
</table>

### "sox_io" backend

Addressed in #1204 
See the [comment](https://github.com/pytorch/audio/issues/1197#issuecomment-768548251)

### "soundfile" backend

1. Add `dtype` parameter to `save` functions of [`"soundfile"` backend with new interface](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/torchaudio/backend/_soundfile_backend.py#L183-L191). Set the default value of `dtype` parameter to `None`.
[...]

PR #1204, opened 2021-01-27
Add target `dtype` argument to `save` function for sox backend

carolineechen:
issue: #1197

### DIFF
--- a/test/torchaudio_unittest/sox_io_backend/save_test.py
+++ b/test/torchaudio_unittest/sox_io_backend/save_test.py
@@ -1,6 +1,7 @@
 import io
 import itertools
 
+import torch
 from torchaudio.backend import sox_io_backend
 from parameterized import parameterized
 
@@ -24,7 +25,7 @@ class SaveTestBase(TempDirMixin, PytorchTestCase):
         """`sox_io_backend.save` can save wav format."""
         path = self.get_temp_path('data.wav')
         expected = get_wav_data(dtype, num_channels, num_frames=num_frames)
-        sox_io_backend.save(path, expected, sample_rate)
+        sox_io_backend.save(path, expected, sample_rate, dtype=None)
         found, sr = load_wav(path)
         assert sample_rate == sr
         self.assertEqual(found, expected)
@@ -68,7 +69,7 @@ class SaveTestBase(TempDirMixin, PytorchTestCase):
         save_wav(src_path, data, sample_rate)
         # 2.1. Convert the original wav to mp3 with torchaudio
         sox_io_backend.save(
-            mp3_path, load_wav(src_path)[0], sample_rate, compression=bit_rate)
+            mp3_path, load_wav(src_path)[0], sample_rate, compression=bit_rate, dtype=None)
         # 2.2. Convert the mp3 to wav with Sox
         sox_utils.convert_audio_file(mp3_path, wav_path)
         # 2.3. Load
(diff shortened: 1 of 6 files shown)


test/torchaudio_unittest/sox_io_backend/save_test.py
```diff
@@ -415,10 +416,25 @@ def test_tensor_preserve(self, dtype):
...
+        ('uint8', (0, 255)),
+    ])
+    def test_dtype_conversion(self, dtype, range):
+        """`save` performs dtype conversion on float32 src tensors only."""
+        path = self.get_temp_path("data.wav")
+        data = get_wav_data("float32", 2)
+        sox_io_backend.save(path, data, 8000, dtype=dtype)
+        found = load_wav(path, normalize=False)[0]
+        self.assertEqual(found.dtype, getattr(torch, dtype))
+        assert(torch.all(found >= range[0]) and torch.all(found <= range[1]))
```
mthrok, line 436: This range assertion does not look quite right. In case of `int32` conversion, it will still pass even if the conversion does not happen.

How about performing exact comparison with Tensors that are small enough to catch the element wise parity?

```python
    @parameterized.expand([
         ('float32', torch.Tensor([-1.0, -0.5, 0, 0.5, 1.0]).to(torch.float32)),
         ('int32', torch.Tensor([-2147483648, ..., 0, ..., 2147483647]).to(torch.int32)),
        ...
    ])
    def test_dtype_conversion(self, dtype, expected):
        data = torch.Tensor([-1.0, -0.5, 0, 0.5, 1.0]).to(torch.dtype32)
        ...
        self.assertEqual(found, expected)
```

mthrok (commented):
Made some suggestion on testing, otherwise it looks good.

torchaudio/backend/sox_io_backend.py
```diff
@@ -243,12 +246,19 @@ def save(
         format (str, optional):
             Output audio format. This is required when the output audio format cannot be infered from
             ``filepath``, (such as file extension or ``name`` attribute of the given file object).
+        dtype (str, optional)
+            Output tensor dtype.
+            Valid values: ``"uint8", "int16", "int32", "float32", "float64", None``
+            ``dtype=None`` means no conversion is performed.
+            ``dtype`` parameter is only effective for ``float32`` Tensor.
     """
+    if src.dtype == torch.float32 and dtype == None:
```
mthrok, line 255: The `save` function has to be compatible with TorchScript compiler. Check out "Type Refinement" section to resolve `None` case.

https://pytorch.org/docs/stable/jit_language_reference.html#optional-type-refinement

I am not sure how TorchScript compiler processes `warnings.warn`. If the torchscript_compatibility test fails, you can move this warnings to `_save` function.

torchaudio/backend/sox_io_backend.py
```diff
@@ -243,12 +246,22 @@ def save(
...
             ``filepath``, (such as file extension or ``name`` attribute of the given file object).
+        dtype (str, optional)
+            Output tensor dtype.
+            Valid values: ``"uint8", "int16", "int32", "float32", "float64", None``
+            ``dtype=None`` means no conversion is performed.
+            ``dtype`` parameter is only effective for ``float32`` Tensor.
     """
+    if src.dtype == torch.float32 and dtype is None:
+        warnings.warn(
+            '`dtype` default value will be changed to `int16` in 0.9 release'
```
mthrok, line 257:
```suggestion
            '`dtype` default value will be changed to `int16` in 0.9 release. '
```

torchaudio/backend/sox_io_backend.py
```diff
@@ -243,12 +246,22 @@ def save(
...
+            'provide `dtype  argument to suppress this warning'
```
mthrok, line 258:
```suggestion
            'Specify `dtype` to suppress this warning.'
```

mthrok (approved):
Looks good. Can you fix the warning message?

mthrok:
Thanks!

Issue #1207, opened 2021-01-28
Expose normalization method to Mel transforms


mthrok:
`create_fb_matrix` function supports Slaney normalization. [[doc](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/torchaudio/functional/functional.py#L300-L301)]

There are three transforms that uses this function, [`MelScale`](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/torchaudio/transforms.py#L238-L270), [`InverseMelScale`](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/torchaudio/transforms.py#L301-L345) and [`MelSpectrogram`](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/torchaudio/transforms.py#L393-L449) (via `MelScale`). It was pointed out in #1058 that it is useful to expose the normalization option to these transforms. It is highly appreciated if someone can contribute this.

## Steps
1. Add `normalization` argument to the transforms. `MelScale`, `InverseMelScale` and `MelSpectrogram`. (default to `None` for the backward compatibility.
2. [Optional] **If** the addition of Slaney normalization to `MelSpectrogram` makes it numerically compatible with librosa, add a test to check the numerical parity against librosa in [`librosa compatibility test`](https://github.com/pytorch/audio/blob/674a71d1a239fd6bf48d5797053590f593efff08/test/torchaudio_unittest/librosa_compatibility_test.py) (Add a new test class.)
This is not confirmed, so please run simple test locally and tell us if this is the case. See also #1058

pratik-99:
Can I help out here?, what do I need do accomplish?

mthrok:
Hi @pratik-99 

Thanks. You can make a PR with the changes described above and ping me.

pratik-99:
hey @mthrok I'm new here can you explain me what do I exactly need to do I don't no the specifications of this project, can you help me out?

jieruan02:
Hi @pratik-99~ do you mind if I try this issue as well? I happen to need to complete some bootcamp task and this is one of them. Thank you!

PR #1212, opened 2021-01-29
Expose normalization method to Mel transforms

jieruan02:
This PR attempts to address issue #1207


mthrok (approved):
Looks good. Thanks!

torchaudio/transforms.py
```diff
@@ -418,6 +426,8 @@ class MelSpectrogram(torch.nn.Module):
             :attr:`center` is ``True``. Default: ``"reflect"``
         onesided (bool, optional): controls whether to return half of results to
             avoid redundancy. Default: ``True``
+        norm (Optional[str]): If 'slaney', divide the triangular mel weights by the width of the mel band
+        (area normalization). (Default: ``None``)
```
rbracco, line 430: Hi, thank you for this PR. This line has a small typo that is causing a formatting issue in the docs. 
image

Line 430 should be indented so the docs generator doesn't think it's a new argument. Thank you!

PR #1213, opened 2021-01-29
Add dtype parameter to soundfile backend

mthrok:
Resolves #1204


mthrok:
To be addressed by #1258

PR #1214, opened 2021-01-31
Improve Flake Rules 

krishnakalyan3:
Followup of #1172


krishnakalyan3:
Import files that need to be replaced by __all__ 
```
test/torchaudio_unittest/common_utils/__init__.py
torchaudio/backend/__init__.py
torchaudio/functional/__init__.py
torchaudio/sox_effects/__init__.py
torchaudio/models/__init__.py
torchaudio/__init__.py
```

krishnakalyan3:
@mthrok we have two options according to 

```
from .my_class import MyClass

__all__ = ['MyClass',]
```

or 

```
from .my_class import MyClass  # noqa: F401

```

mthrok:
> @mthrok we have two options according to
> 
> ```
> from .my_class import MyClass
> 
> __all__ = ['MyClass',]
> ```
> 
> or
> 
> ```
> from .my_class import MyClass  # noqa: F401
> ```
> 
> Confirming that the first option was preferred right?

Hi @krishnakalyan3 

Thanks for the follow up. Yes, the `__all__` is the preferred approach for public APIs, but for internal APIs, `noqa` is preferred. So, for `backend` module, let's use `noqa`.

mthrok (approved):
@krishnakalyan3 Thank you for working on this. It looks good. Is this ready to merge? (the failing I/O tests are flaky ones that I am hunting down the condition, so we can ignore.)

krishnakalyan3:
@mthrok yes. Looks like the tests are failing. I am not sure why.

mthrok:
Thanks!

PR #1215, opened 2021-02-01
Added AudioFolder to load a dataset from a local path

cassieview:
A function added to torchaudio.datasets to create a dataset from local files from a local path with some basic transforms (sample rate and spectrogram) build in.

### DIFF
--- a/torchaudio/datasets/folder.py
+++ b/torchaudio/datasets/folder.py
@@ -19,13 +19,11 @@ class AudioFolder(Dataset):
     Args:
         root (str): Path to the directory where the dataset is found or downloaded.
         suffix (str) : Audio file type, defaulted to ".WAV".
-        transform (callable, optional): A function/transform that  takes in the audio waveform
-            and returns a transformed version. E.g, ``transforms.Spectrogram``
-        target_transform (callable, optional): A function/transform that takes in the
-            target and transforms it.
-        loader (callable, optional): A function to load an image given its path.
-        is_valid_file (callable, optional): A function that takes path of an Image file
-            and check if the file is a valid file (used to check of corrupt files)
+        new_sample_rate (int) : Resample audio to new sample rate specified.
+        spectrogram_transform (bool): If True transform the audio waveform and returns it  
+        transformed into a spectrogram tensor.
+        label_name (str): The label is pulled from the folders in the path, this allows you to statically
+        define the label string.
     """
 
 
@@ -50,7 +48,7 @@ class AudioFolder(Dataset):
         """Load the n-th sample from the dataset.
 
         Args:
-            n (int): The index of the sample to be loaded
+            n (int): The index of the file to be loaded
 
         Returns:
             tuple: ``(waveform, sample_rate, label, filename)``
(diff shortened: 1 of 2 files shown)


mthrok:
Hi @cassieview 

Thanks for contribution. Was this discussed / specified somewhere?
I am concerned that this implementation provides specific solution to under-specified problem, and do not think this is suitable for library implementation without detailed discussion / conclusion that this implementation can cover the wide variety of use cases.
Can you provide justification of adding this?

torchaudio/datasets/folder.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+    def __init__(
+            self,
+            path: str,
+            suffix: str = ".wav",
+            pattern: str = "*/*",
+            new_sample_rate: int = None,
+            spectrogram_transform: bool = False,
+            label: str = None
```
mthrok, line 38: If this label is provided by user, the Dataset always returns the same label. In that case, users can always use their own label, so there is no benefit making this into a parameter.

torchaudio/datasets/folder.py
```diff
@@ -0,0 +1,73 @@
+import os
+from typing import Tuple
+import torchaudio
+from torch.utils.data import Dataset
+from torch import Tensor
+from pathlib import Path
+
+def load_audio_item(filepath: str, path: str, label_name: str) -> Tuple[Tensor, int, str, str]:
```
mthrok, line 8: Making this into a plain function makes it hard to customize the loading behavior.

Imagine use-cases like fixed-length portion is required (say 5 seconds chunk), and the audio data in file system are long (like one hour). Without customization, one has to load (decode) the entire audio and they are immediately discarded. This is inefficient.

Not being able to customize the load behavior reduces the value of having a generic dataset like this.

torchaudio/datasets/folder.py
cassieview: If left null then it doesn't set the label or pulls it from the subfolders. If I am pulling all one class from a folder then label comes back as an empty string. It is useful to be able to define it when creating a dataset of one class from one folder.

torchaudio/datasets/folder.py
cassieview: Let me look into this. Thanks!

mthrok: Before continue working on this, please open [an issue with Feature Request template](https://github.com/pytorch/audio/issues/new?assignees=&labels=&template=feature-request.md) and explain the benefit to have this dataset implementation in torchaudio for the broader audience, and spec it out. You can reference this PR as well, but we do need the goal defined first and clear design principle, otherwise we cannot accept this PR.

cassieview: Will do, thanks :)

torchaudio/datasets/folder.py
```diff
@@ -0,0 +1,73 @@
...
+            tuple: ``(waveform, sample_rate, label, filename)``
+        """
+        fileid = self._walker[n]
+
+        waveform, sample_rate, label, filename =  load_audio_item(fileid, self._path, self._label)
+
+        if self._new_sample_rate is not None:
+            waveform = torchaudio.transforms.Resample(sample_rate, self._new_sample_rate)(waveform)
+            sample_rate = self._new_sample_rate
+        if self._spectrogram_transform is not None:
```
iociveteres, line 65: By default False is provided to self._spectrogram_transform. This if checks for None, so it always goes in, and method always and by default returns a spectrogram tensor of unexpected shape, but not a wavefrom.

PR #1216, opened 2021-02-01
Update SpectralCentroid to accept window_fn and wkwargs

NicolasHug:
This PR updated `SpectralCentroid` transform to accept `window_fn` and `wkargs` instead of `window`, for consistency with the other transforms.

CC @mthrok


mthrok (approved):
Looks good. Thanks!

PR #1217, opened 2021-02-01
Fix segfault by calling ftell in place of tell_off

mthrok:
[`sox_format_t.tell_off`](https://fossies.org/dox/sox-14.4.2/structsox__format__t.html#a2016a9fa839f3139e3c2f64381b0c445) should be representing current offset in file, but there are cases it does not. This was causing segmentation fault in some cases. This PR fixes it by replacing it with `ftell` call and add extra check so that if the same thing should happen, it will throw runtime error instead of segmentation fault.

See also: [internal test log](https://www.internalfb.com/intern/testinfra/diagnostics/1125900086940494.281474997971466.1611950392/)


cpuhrsch (approved):

vincentqb (approved):
LGTM. Looks like internal tests will be enough for now.

Issue #1201, opened 2021-01-27
Par file registration


mthrok:
Currently torchaudio's functions related to `libsox` are bound via `TORCH_LIBRARY_FRAGMENT` macro in [`torchaudio/csrc/sox/register.cpp`](https://github.com/pytorch/audio/blob/22e7e877f1b50355e3e6f55e14d06f6574c4b951/torchaudio/csrc/sox/register.cpp#L5).

Towards the support for mobile deployment, we would like to change the library structure better suited for selective build. To prepare for this, we would like to change the source code structure, so that registration of custom ops are carried out **per source file**. This means that we want to remove the `torchaudio/csrc/sox/register.cpp` and use the `TORCH_LIBRARY_FRAGMENT` macro in CPP files where functions are implemented. (See [this example](https://github.com/pytorch/vision/blob/d732562f30b8a83af9ab074592af3dd01fe708ee/torchvision/csrc/ops/nms.cpp#L8-L21) from torchvision.)

## Steps

1. Hide `TensorSignal` class usage
    1. Replace `TensorSignal` type in registered functions
[`TensorSignal` class](https://github.com/pytorch/audio/blob/22e7e877f1b50355e3e6f55e14d06f6574c4b951/torchaudio/csrc/sox/utils.h#L29-L44) is used to pass around an audio `Tensor` and its metadata together. It is used by the following registered functions and this usage prevents from par-file registration. `TensorSignal` class should be replaced with `std::tuple<torch::Tensor, int64_t>` for Tensor object and sample rate.
        - [`load_audio_file`](https://github.com/pytorch/audio/blob/22e7e877f1b50355e3e6f55e14d06f6574c4b951/torchaudio/csrc/sox/io.h#L35-L41)
        - [`apply_effects_tensor` and `apply_effects_file`](https://github.com/pytorch/audio/blob/22e7e877f1b50355e3e6f55e14d06f6574c4b951/torchaudio/csrc/sox/effects.h#L18-L27).
    2. Remove the registration of `TensorSignal`
Once the usages of `TensorSignal` type are replaced, [the registration](https://github.com/pytorch/audio/blob/22e7e877f1b50355e3e6f55e14d06f6574c4b951/torchaudio/csrc/sox/register.cpp#L9-L17) can be removed.
    3. Change the Python code where `TensorSignal` was used.
[...]

PR #1221, opened 2021-02-01
Restructure code to allow per file registration of custom ops

prabhat00155:
Addresses https://github.com/pytorch/audio/issues/1201.


torchaudio/csrc/sox/effects.cpp
```diff
@@ -50,10 +50,11 @@ void shutdown_sox_effects() {
...
 }
 
-c10::intrusive_ptr<TensorSignal> apply_effects_tensor(
-    const c10::intrusive_ptr<TensorSignal>& input_signal,
-    std::vector<std::vector<std::string>> effects) {
-  auto in_tensor = input_signal->getTensor();
+std::tuple<torch::Tensor, int64_t> apply_effects_tensor(
+    std::tuple<torch::Tensor, int64_t> input_signal,
+    std::vector<std::vector<std::string>> effects,
+    c10::optional<bool>& channels_first) {
```
mthrok, line 56: I think `bool` type instead of `c10::optional<bool>` type better fits the context. All the Python code that calls this function has signature of `channels_first: bool = true`. So it should always have a boolean value.

mthrok: Also could you untangle `std::tuple<torch::Tensor, int64_t> input_signal` into `torch::Tensor waveform` and `int64_t sample_rate`? That way the interface between Python and C++ becomes flat (not need to construct a tuple `(waveform, sample_rate)`) so it looks better.

Then we have option to whether construct a `TensorSignal` object here or apply the same change to `addInputTensor` (and to utility functions used in there like `get_signalinfo`). I will let you decide as the later approach feels digressing from the main point of the PR.

mthrok (commented):
Thanks @prabhat00155 

The PR looks very good. Please refer to the comments for some additional points.

mthrok:
@prabhat00155 Thank you!

PR #1222, opened 2021-02-01
Use underscore for python files

vincentqb:
Suppress hyphenated python files internal warning, [here](https://www.internalfb.com/diff/D26180248). Doing so is also consistent with the other python filenames (which are using underscore instead of hyphen).

cc #1198


datumbox (approved):
LGTM!

Issue #1219, opened 2021-02-01
[Refactor] Simplify C++ interface for sox_io's `info` func


mthrok:
Currently, `torchaudio.info` function for `"sox_io"` backend gets the information via `SignalInfo` structure in C++.

https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/io.h#L14-L32

The signature of the actual function that performs querying looks like following;

https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/io.h#L34-L36

When binding this function (making it available on Python), the `SignalInfo` structure has to be bound before the function. When updating the `info` function, this forces updating multiple components.

Also the recently added function that performs the same operation on file-like object, which is bound via PyBind11, simplifies this via directly returning tuple object.

https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/io.h#L57-L59

`get_info_file` should also do the same and return tuple object directly.

## Steps

1. Change the interface / implementation of `torchaudio.info`
    1. `get_info_file` to return the tuple. The implementation should be swapped from [this](https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/io.cpp#L93-L98) to [this](https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/io.cpp#L245-L250)
    2. Update the Python frontend. [The construction of AudioMetaData](https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/backend/sox_io_backend.py#L23-L29) should be replaced with [the one with tuple destruction](https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/backend/sox_io_backend.py#L21).
2. Remove [the registration](https://github.com/pytorch/audio/blob/828df46a95310453dda13599a29fa22a1cc049da/torchaudio/csrc/sox/register.cpp#L42-L49).
[...]

PR #1232, opened 2021-02-03
Simplified C++ interface for sox_io's get_info_file()

prabhat00155:


prabhat00155:
Fixes https://github.com/pytorch/audio/issues/1219.

torchaudio/backend/sox_io_backend.py
```diff
@@ -16,17 +16,10 @@ def _info(
...
-        sinfo.get_sample_rate(),
-        sinfo.get_num_frames(),
-        sinfo.get_num_channels(),
-        sinfo.get_bits_per_sample(),
-        sinfo.get_encoding(),
-    )
+    sinfo = (torchaudio._torchaudio.get_info_fileobj(filepath, format)
+             if hasattr(filepath, 'read') else
+             torch.ops.torchaudio.sox_io_get_info(os.fspath(filepath), format))
+    return AudioMetaData(*sinfo)
```
mthrok, line 22: Can you try merge `_info` func and `info` in the same way as `load` function?

I think now TorchScript compiler can handle it without extracting the Python-only part to a helper function.

You can run the test locally with `pytest test/torchaudio_unittest/sox_io_backend/torchscript_test.py`

mthrok (approved):
Looks good, thanks. Could you try merging `_info` and `info` functions and see if it works? If it does not, let me know and I will merge this as-is.

mthrok:
Thanks!

PR #1226, opened 2021-02-02
Add encoding and bits_per_sample option to save function

mthrok:
The current release version's "sox_io" backend's `save` function changes the encoding of the audio file based on the dtype of the Tensor. For example, if the dtype is "float32", then it will be saved as 32bit floating point PCM. This behavior was taken from [SciPy's `scipy.io.wavefile.write` function](https://docs.scipy.org/doc/scipy/reference/generated/scipy.io.wavfile.write.html#scipy-io-wavfile-write). However it was pointed out that this is inconvenient for torchaudio users. Because most torchaudio's functionality works on float32 Tensor yet, the common audio formats typically retains only 16 bit, such as 16 bit signed integer PCM.

To resolve the inconvenience while keeping the functionality to support different encodings, this PR 
1. Add `encoding` and `bits_per_sample` parameters to `save` function.
2. ~For non-compressed format (such as "wav"), it defaults to 16-bit signed integer PCM. (This is BC-breaking behavior if users were dumping Tensor object without converting to the matching dtype)~ The default encoding/bit depth are now determined by dtype, which is not BC-breaking.

**Note** in https://github.com/pytorch/audio/pull/1204, we attempted to solve the issue by adding `dtype` argument, but after giving more thoughts I concluded that this approach is limited and the design is not intuitive, so we will replace the `dtype` argument with `encoding` and `bits_per_sample`. This will allow other encodings such as mu-law and a-law.
For the detailed behavior of the `encoding` and `bits_per_sample`, see the table at the bottom of this description. 

Along the way, the implementation of type conversion is updated to use the implementations that `libsox` provides. So the new implementation of `save` function yields the same result as `sox` command.

**Table: the provided values and the actual values to be used**
* For WAV format without `encoding` or `bits_per_sample` option, the encoding configuration is determined by `dtype`.
* If an invalid value is provided, it raises `RuntimeError`
* Other formats, like `mp3` and `vorbis`, do no support changing `encoding` and `bits_per_sample`, so if users provide some values, they will be ignored with warning.

<table width="672.0" cellspacing="0" cellpadding="0" class="t1">
  <tbody>
    <tr>
      <td colspan="3" valign="middle" class="td1">
        <p class="p1"><span class="s1"><b>Provided arguments</b></span></p>
      </td>
      <td colspan="2" valign="middle" class="td2">
        <p class="p1"><span class="s1"><b>Resulting value (if differ)</b></span></p>
      </td>
    </tr>
    <tr>
      <td valign="middle" class="td3">
[...]

PR #1233, opened 2021-02-03
Refactor `get_encodinginfo` logic

mthrok:
`sox_encodinginfo_t` is the structure used to communicate the encoding with `libsox`.
This structure is constructed/used in the following scenarios;
1. Pass Tensor object to `libsox` buffer.
2. Construct Tensor object from `libsox` buffer.
3. Save Tensor object to file or file-like object.

Originally, all the three cases used the same logic `get_encodinginfo(format string, Tensor dtype) -> sox_encodinginfo_t` but as I work on #1226 , I figured that it is better to use different logics for 1&2 and 3.

This PR makes the distinguish clearer.
For 1&2, the helper function is renamed to `get_tensor_encodinginfo` and it only takes `Tensor dtype` as argument (because Tensor is always raw, uncompressed format like WAV).
For 3, I renamed the function to `get_encodinginfo_for_save`. This function will go through major overhaul in #1226, so for this PR, its implementation is kept as-is.

### DIFF
--- a/torchaudio/csrc/sox/effects.cpp
+++ b/torchaudio/csrc/sox/effects.cpp
@@ -60,8 +60,8 @@ std::tuple<torch::Tensor, int64_t> apply_effects_tensor(
   // Create SoxEffectsChain
   const auto dtype = waveform.dtype();
   torchaudio::sox_effects_chain::SoxEffectsChain chain(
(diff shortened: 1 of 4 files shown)


datumbox (approved):
Stamping!

mthrok:
thanks!

Issue #1231, opened 2021-02-03
[Refactor] Move TorchScript tests for functional to dedicated directory.


mthrok:
The number of unit tests we have in torchaudio has grown a lot. It is critical to keep tests organized so as to make it easy for contributors to write tests for the feature they are working on.

TorchScript consistency tests are for validating that the execution results are same between Python eager mode and TorchScript runtime. We put all the tests in [test/torchaudio_unittest/torchscript_consistency_impl.py](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/torchscript_consistency_impl.py)(definitions of tests that are common for CPU and CUDA), [test/torchaudio_unittest/torchscript_consistency_cpu_test.py](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/torchscript_consistency_cpu_test.py) and [test/torchaudio_unittest/torchscript_consistency_cuda_test.py](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/torchscript_consistency_cuda_test.py). As the list of ops in `torchaudio.functional` grows, we have a dedicated directory for functional tests, [test/torchaudio_unittest/functional](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/functional). Moving the TorchScript consistency tests for functional to the directory better decouples the test module.

## Steps
1. Create the equivalent of `torchscript_consistency_impl.py`, `torchscript_consistency_cpu_test.py` and `torchscript_consistency_cuda_test.py` in [test/torchaudio_unittest/functional](https://github.com/pytorch/audio/blob/3651412bbcb0c4faf56bc7be7fe5e5ae2a7a63fd/test/torchaudio_unittest/functional) directory.
2. Move the tests for `functional` from the original location to the newly created scripts.

## Testing

1. Install nightly build of PyTorch
2. Clone repo and build torchaudio
3. Run test
    Run test with `pytest test/torchaudio_unittest/functional -v`

jieruan02:
I will give it a try

mthrok:
@jieruan02 

Thanks. Let me know if you need help.

jieruan02:
[...]

PR #1237, opened 2021-02-03
Refactor tests for functional

jieruan02:
This PR tries to address issue #1231.

Here is the pytest summary I got, let me know if requires further look. Also I didn't prefix the file name with functional since it's already under the functional directory, not sure if it conforms to the repo's rule or not, let me know~

====================================== warnings summary ======================================
test/torchaudio_unittest/functional/consistency_cpu_test.py::TestFunctionalFloat32::test_create_fb_matrix
test/torchaudio_unittest/functional/consistency_cpu_test.py::TestFunctionalFloat32::test_create_fb_matrix
test/torchaudio_unittest/functional/consistency_cpu_test.py::TestFunctionalFloat64::test_create_fb_matrix
test/torchaudio_unittest/functional/consistency_cpu_test.py::TestFunctionalFloat64::test_create_fb_matrix
/Users/jieruan/Library/Python/3.8/lib/python/site-packages/torchaudio/functional/functional.py:340: UserWarning: At least one mel filterbank has all zero values. The value for n_mels (10) may be set too high. Or, the value for n_freqs (100) may be set too low.
warnings.warn(

test/torchaudio_unittest/functional/consistency_cpu_test.py::TestFunctionalFloat32::test_griffinlim
/Users/jieruan/Library/Python/3.8/lib/python/site-packages/torch/functional.py:654: UserWarning: istft will require a complex-valued input tensor in a future PyTorch release. Matching the output from stft with return_complex=True. (Triggered internally at ../aten/src/ATen/native/SpectralOps.cpp:787.)
return _VF.istft(input, n_fft, hop_length, win_length, window, center, # type: ignore

======================= 103 passed, 107 skipped, 5 warnings in 28.85s ========================


test/torchaudio_unittest/functional/consistency_cpu_test.py
```diff
@@ -0,0 +1,14 @@
+import torch
+
+from torchaudio_unittest import common_utils
```
mthrok, line 3: Can you change the import to

```python
from torchaudio_unittest.common_utils import PytorchTestCase
```
Same goes to `cuda_test.py`.

mthrok (changes requested):
Please retain the name `torchscript_consistency_***.py`. If it's just `consistency`, then it is not immediately clear what kind of consistency it is about. (We have multiple kind of tests for checking parity.)

test/torchaudio_unittest/functional/consistency_cpu_test.py
jieruan02: no problem, I will do the both~

jieruan02:
I kind of wonder if there is some good local tests to run before pushing. my way of testing via ci seems pretty inefficient to me...

mthrok (changes requested):
Please keep the original name `torchscript_consistency`. `functional_consistency` is not clear what it means.

mthrok (changes requested):
Please rename 

`test/torchaudio_unittest/functional/functional_consistency_cpu_test.py`  into `test/torchaudio_unittest/functional/torchscript_consistency_cpu_test.py` and apply the same changes to currently called `functional_consistency_*.py` scripts.

jieruan02:
> Please rename
> 
> `test/torchaudio_unittest/functional/functional_consistency_cpu_test.py` into `test/torchaudio_unittest/functional/torchscript_consistency_cpu_test.py` and apply the same changes to currently called `functional_consistency_*.py` scripts.



> Please rename
> 
> `test/torchaudio_unittest/functional/functional_consistency_cpu_test.py` into `test/torchaudio_unittest/functional/torchscript_consistency_cpu_test.py` and apply the same changes to currently called `functional_consistency_*.py` scripts.

Oh I see, I misunderstood the part, that's why I omit the prefix at first place... Sure I will fix it now

mthrok:
Thanks!

jieruan02:
happy to help~
________________________________

PR #1243, opened 2021-02-05
Add Kaldi Pitch feature

mthrok:
This PR adds Kaldi Pitch Feature detailed in "A pitch extraction algorithm tuned for automatic speech recognition".

The interface is mostly same as `compute-kaldi-pitch-feats` CLI. Batch support is added via `at::parallel_for` function.

As the function binds the custom built `libkaldi`, it only supports CPU (and `float32`) at the moment.

### About the custom built Kaldi

Since kaldi is a large library, I choose to use only a subset of it by adding custom build process. In addition to that, to reduce the dependencies and make the build process, I reused the BLAS package PyTorch uses. For this, I added custom interface of Kaldi's matrix libraries. Therefor the algorithm runs on `torch::Tensor` class. However, some parts of the algorithms requires direct access to memory, so the resulting function is not differentiable. Also the resulting code is very slow (60x) at the moment due to the overhead of slicing operation. (Kaldi's feature implementations work element-wise, while PyTorch operates faster when operations are vectorized.)

Supersedes #1063


third_party/kaldi/src/matrix/kaldi-vector.h
```diff
@@ -0,0 +1,313 @@
...
+  // https://github.com/kaldi-asr/kaldi/blob/7fb716aa0f56480af31514c7e362db5c9f787fd4/src/matrix/kaldi-vector.h#L326-L330
+  void AddDiagMat2(
+      Real alpha,
+      const MatrixBase<Real>& M,
+      MatrixTransposeType trans = kNoTrans,
+      Real beta = 1.0) {
+    auto mat = M.tensor_;
+    if (trans == kNoTrans) {
+      tensor_ =
+          beta * tensor_ + torch::diag(torch::mm(mat, mat.transpose(1, 0)));
```
cpuhrsch, line 195: If this is called in a tight-loop with small Tensors, you might fare better inlining the operation and using the underlying tensor data pointer instead of calling into torch::mm repeatedly.

Also note that torch::mm does parallelization, but at::parallel_for disables nested parallelism for OpenMP (to avoid oversubscription), which is the default threadpool for PyTorch.

cpuhrsch (approved):

PR #1246, opened 2021-02-07
Add -Wall

mthrok:
Adding back `-Wall` flag, which was unintentionally removed during the CMake transition.
(There will be a better way to achieve this, but for now adding it to `CMAKE_CXX_FLAGS` so that all the compilation unit has the flag.)


NicolasHug (approved):
LGTM!

mthrok:
Thanks!

PR #1247, opened 2021-02-08
DOC Fix sphinx warnings and turn warnings into errors

NicolasHug:
Similar to what was done in `torchvision` https://github.com/pytorch/vision/pull/3290

This PR fixes sphinx warnings and sets the 

```py
napoleon_numpy_docstring = False
napoleon_google_docstring = True
```

for sphinx to issue warnings when the wrong parameter syntax is used.

The warnings are now turned into errors. Note: sphinx will now stop compilation at the first warning/error it sees; if we relied on sphinx>=8, we could use the `--keep-going` option to let it finish.

### DIFF
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -45,6 +45,7 @@ def deprecated(direction: str, version: Optional[str] = None):
     """
     def decorator(obj):
         # get __init__ if obj is a class, else get the function itself
+        # FIXME: remove when SignalInfo and EncodingInfo are properly removed
         func = getattr(obj, '__init__', obj)
 
         @wraps(func)
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -2,6 +2,7 @@ import warnings
 import importlib.util
 from typing import Optional
 from functools import wraps
+import inspect
 
 
 def is_module_available(*modules: str) -> bool:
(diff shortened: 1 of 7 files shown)


torchaudio/_internal/module_utils.py
```diff
@@ -43,7 +43,11 @@ def deprecated(direction: str, version: Optional[str] = None):
     Args:
         direction: Migration steps to be given to users.
     """
-    def decorator(func):
+    def decorator(obj):
+        # get __init__ if obj is a class, else get the function itself
+        # FIXME: remove when SignalInfo and EncodingInfo are properly removed
+        func = getattr(obj, '__init__', obj)
```
NicolasHug, line 49: For ref, this is needed to avoid a similar error as in https://github.com/astropy/astropy/issues/2051.

The main issue is that `decorator` doesn't  work on classes; it only works on functions but it's currently used for the `SignalInfo` and `EncodingInfo` classes which led to the warning / error.

I'm not sure the current changes properly makes the decorator work on classes - but it does successfully suppress the warning, and as the offending classes are deprecated and will be removed, I assumed it was good enough.

If needed, I'm happy to propose a proper fix and let the decorator work on classes, but that will be a bit more work

mthrok: Thanks for the clarification. `SignalInfo` and `EncodingInfo` classes are expected to be removed after the next release (if things go smoothly) so I think this workaround is good enough.

mthrok:
Looks like `deprecated` decorator is affected. What do you think is the best (easiest) way to resolve this? Is the way I added `deprecated` decorator invliad?

torchaudio/_internal/module_utils.py
NicolasHug: I couldn't make it work properly after all, so I removed the changes to the decorator and removed its us in the corresponding classes, and instead warned inside `__init__`

mthrok: Sounds good.

mthrok (approved):

PR #1250, opened 2021-02-08
List only one item par line in __all__

mthrok:
Follow-up of #1214 
Place one imported item per line so that **diff looks simpler when adding a new item**. (like in functional, which is what I am working in #1243) 

Note: I do not intend to start a discussion on how/if we should run formatter for the codebase. Please have that kind of discussion elsewhere.


datumbox (approved):
LGTM.

mthrok:
Thanks!

Issue #1251, opened 2021-02-08
Fix transducer test


mthrok:
RNN-Transducer is a recently added prototype module and users can opt-in/out when building torchaudio.

The unit test for RNNT is supposed to be skipped when RNNT is not available, but instead it skips only when C++ extension is not available. It still runs test when C++ extension is available but not compiled with RNNT.

https://github.com/pytorch/audio/blob/77de2b9616c47c114124a8e002447ad927799dbd/test/torchaudio_unittest/transducer_test.py#L274-L275

## Steps for fix
1. Replace the `skipIfNoExtension` decorator with one with logic specific to RNNT-loss
The implementation should check if `torch.ops.torchaudio.rnnt_loss` is accessible. (You can use `try ~ except ~` clause.)
The new decorator should live in the same file as `transducer_test.py`. (Let's call this `skipIfNoRNNT`)
2. Replace `PytorchTestCase` with `TorchaudioTestCase`
This is not related to the test skip logic, but `PytorchTestCase` is not the right class to be used here, so let's fix it.

## Build and test

1. Install nightly build of PyTorch
2. Clone repo
4. Run test with `pytest test/torchaudio_unittest/transducer_test.py -v` and confirm that test is not skipped and failed. (The test should not pass)
3. Fix the issue
4. Run the test again and see that it is skipped properly

Before the fix, it should fail with a message like `RuntimeError: No such operator torchaudio::rnnt_loss` as follow;

```
self = <[RuntimeError('No such operator torchaudio::__file__') raised in repr()] _OpNamespace object at 0x7f31d9e003b0>, op_name = 'rnnt_loss'

    def __getattr__(self, op_name):
        # Get the op `my_namespace::my_op` if available. This will also check
        # for overloads and raise an exception if there are more than one.
        qualified_op_name = '{}::{}'.format(self.name, op_name)
>       op = torch._C._jit_get_operation(qualified_op_name)
E       RuntimeError: No such operator torchaudio::rnnt_loss

/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/_ops.py:61: RuntimeError
```

After the fix, the test should be properly skipped.

jieruan02:
can I try this?

PR #1253, opened 2021-02-08
Fix transducer test

jieruan02:
This PR attempts to address issue #1251.

### DIFF
--- a/test/torchaudio_unittest/transducer_test.py
+++ b/test/torchaudio_unittest/transducer_test.py
@@ -1,7 +1,8 @@
+import unittest
 import torch
 from torchaudio.prototype.transducer import RNNTLoss
 
-from torchaudio_unittest import common_utils
+from torchaudio_unittest.common_utils import TorchaudioTestCase
 
 
 def get_data_basic(device):
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
     return costs, gradients
 
 
+def skipIfNoTransducer(test_item):
+    try:
+        torch.ops.torchaudio.rnnt_loss
+        return test_item
+    except RuntimeError:
+        return unittest.skip("torchaudio C++ extension is not compiled with RNN transducer loss")(test_item)
+
+
 class TransducerTester:
     def test_basic_fp16_error(self):
         rnnt_loss = RNNTLoss()
@@ -271,6 +280,6 @@ class TransducerTester:
         self.assertEqual(gradients, ref_gradients, atol=atol, rtol=rtol)
 
 
-@common_utils.skipIfNoExtension
-class CPUTransducerTester(TransducerTester, common_utils.PytorchTestCase):
+@skipIfNoTransducer
+class CPUTransducerTester(TransducerTester, TorchaudioTestCase):
     device = "cpu"


test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
     return costs, gradients
 
 
+def skipIfNoRNNT(test_item):
+    try:
+        import torch.ops.torchaudio.rnnt_loss as _
```
mthrok, line 247: Import looks a bit strange, can you just access the attribute?

```suggestion
        _ = torch.ops.torchaudio.rnn_loss
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
...
+        return test_item
+    except ImportError:
```
mthrok, line 249: With the above change, this should be `RuntimeError`

```
>>> _ = torch.ops.torchaudio.rnn_loss
  File "/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/_ops.py", line 61, in __getattr__
    op = torch._C._jit_get_operation(qualified_op_name)
RuntimeError: No such operator torchaudio::rnn_loss
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -271,6 +280,7 @@ def test_costs_and_gradients_B2_T4_U3_D3_fp32(self):
         self.assertEqual(gradients, ref_gradients, atol=atol, rtol=rtol)
 
 
+@skipIfNoRNNT
 @common_utils.skipIfNoExtension
```
mthrok, line 284: Please remove `@common_utils.skipIfNoExtension` as it is redundant.

mthrok (commented):
Thanks for working on this.

test/torchaudio_unittest/transducer_test.py
jieruan02: Thanks! I was trying to suppress the style check error for simple `import torch.ops.torchaudio.rnnt_loss`

test/torchaudio_unittest/transducer_test.py
jieruan02: l keep import because in my machine, if I only do `_ = torch.ops.torchaudio.rnn_loss` without import, it won't skip test. I think it's because when running on script, the code is not actually executed vs in interactive mode, so it wont trigger exception. but to pass the style check on github, I tried to do `_ = rnnt_loss`. let me know if you think of better alternative

mthrok: >  if I only do _ = torch.ops.torchaudio.rnn_loss without import, it won't skip test.

That might mean that you have a nightly-built torchaudio installed.
Can you verify that you do not have a torchaudio installed somewhere else?

jieruan02: yes I do have the nightly-build installed, but I thought that's the instruction in the issue #1251. So should I remove the import part and only do `_ = torch.ops.torchaudio.rnn_loss`? Thank you!

mthrok: The direction is to install nightly-build of PyTorch and not nightly-build of torchaudio.
Anyways, please just do the attribute access only.

jieruan02: got it, sorry about that, I misunderstand that point. will do so immediately

test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
     return costs, gradients
 
 
+def skipIfNoRNNT(test_item):
```
vincentqb, line 245: referring to "transducer" is more consistent with how we refer to it elsewhere
```suggestion
def skipIfNoTransducer(test_item):
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
...
+    try:
+        _ = torch.ops.torchaudio.rnnt_loss
```
vincentqb, line 247: nit: i see a [discussion above](https://github.com/pytorch/audio/pull/1253#discussion_r572971066) adding `_ = ...`, but `_ = ` isn't needed, right?
```suggestion
        torch.ops.torchaudio.rnnt_loss
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
...
+        return test_item
+    except RuntimeError:
+        return unittest.skip("torchaudio C++ extension is not compiled with RNNT")(test_item)
```
vincentqb, line 250: we could say "with the RNN transducer loss", or more explicitly:
```suggestion
        return unittest.skip("torchaudio C++ extension is not compiled with BUILD_TRANSDUCER=1")(test_item)
```

test/torchaudio_unittest/transducer_test.py
```diff
@@ -271,6 +280,6 @@ def test_costs_and_gradients_B2_T4_U3_D3_fp32(self):
         self.assertEqual(gradients, ref_gradients, atol=atol, rtol=rtol)
 
 
-@common_utils.skipIfNoExtension
-class CPUTransducerTester(TransducerTester, common_utils.PytorchTestCase):
+@skipIfNoRNNT
```
vincentqb, line 283: as above:
```suggestion
@skipIfNoTransducer
```

jieruan02: Thanks a lot for the suggestion! its very rigorous I got to keep up with your standard. will change it first

test/torchaudio_unittest/transducer_test.py
mthrok: @jieruan02 Please do not include `BUILD_TRANSDUCER=1` here.

* This could be changed anytime. When we update the build process, we update the documentations that we expect to find build process, which is README, CONTRIBUTING etc. A message in a test decorator is the place one would never expect to find a build direction so there is a high chance that this becomes obsolete and that case, it gives a bad developer experience.
*  It is not quite correct to say "skipping because it is not built with BUILD_TRANSDUCER=1". This message assumes that BUILD_TRANSDUCER=1 will build the required component, which is not wrong, but that is not guaranteed to be the only way. Why would give such an indirect, implied message, where the core of the logic is either it has certain feature compiled/available.

mthrok: Just say `torchaudio C++ extension is not compiled with the RNN transducer loss`

mthrok (commented):
Could you also change

```python
from torchaudio_unittest import common_utils
```

to 

```python
from torchaudio_unittest.common_utils import TorchaudioTestCase 
```
?

jieruan02:
> Could you also change
> 
> ```python
> from torchaudio_unittest import common_utils
> ```
> 
> to
> 
> ```python
> from torchaudio_unittest.common_utils import TorchaudioTestCase 
> ```
> 
> ?

sure, no problem!

test/torchaudio_unittest/transducer_test.py
```diff
@@ -241,6 +242,14 @@ def compute_with_pytorch_transducer(data):
     return costs, gradients
 
 
+def skipIfNoTransducer(test_item):
+    try:
+        torch.ops.torchaudio.rnnt_loss
+        return test_item
+    except RuntimeError:
+        return unittest.skip("torchaudio C++ extension is not compiled with BUILD_TRANSDUCER=1")(test_item)
```
mthrok, line 250: Please remove `BUILD_TRANSDUCER=1` here for the said reasons.

mthrok (changes requested):
Please remove `BUILD_TRANSDUCER=1` here for the said reasons.

mthrok:
Thanks!

jieruan02:
glad to help!

PR #1263, opened 2021-02-12
Fix dev setup direction

mthrok:


CONTRIBUTING.md
```diff
@@ -53,7 +53,7 @@ BUILD_SOX=1 python setup.py develop
 # or, for OSX
 # BUILD_SOX=1 MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py develop
 # for C++ debugging, please use DEBUG=1
-# DEBUG=1 python setup.py install
+# DEBUG=1 python setup.py develop
```
NicolasHug, line 56: Does `install` allow to edit code without having to re-install the package each time? I'm referring to https://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install

mthrok: `install` does not allow editing the code.

mthrok: also install places the resulting code in somewhere other than the cloned repo.
This can cause the case where documentation is built from installed package.
There were multiple reports from the contributors in the past.

NicolasHug: 🤦 brain fart, I was reading the diff the other way around and thought you wanted to go from `develop` to `install`, my bad!

NicolasHug (approved):

PR #1267, opened 2021-02-13
Improve MelSpectrogram librosa compatibility test

### LINKED ISSUE #1262: Improve MelSpectrogram librosa compatibility test
Currently the unit test for MelSpectrogram is varied amongst other tests.

https://github.com/pytorch/audio/blob/4f9b55201dea6d0dd5165a4f684d1160ee9ca94a/test/torchaudio_unittest/librosa_compatibility_test.py#L168-L234

This is very inconvenient from the viewpoint of maintainability because adding a new test for different argument is hard. Also it makes it hard to look into the property of test. (like is MelSpectrogram compatible against librosa? #1058)

We would like to extract the test for MelSpectrogram so that it is independently tested from the other functions.

## Steps
1. Extract the part that tests `MelSpectrogram`
https://github.com/pytorch/audio/blob/4f9b55201dea6d0dd5165a4f684d1160ee9ca94a/test/torchaudio_unittest/librosa_compatibility_test.py#L183-L193 and put it in a new method. (let's call `test_MelSpectrogram`).
2. Grab [the currently tested parameters](https://github.com/pytorch/audio/blob/4f9b55201dea6d0dd5165a4f684d1160ee9ca94a/test/torchaudio_unittest/librosa_compatibility_test.py#L247-L291), and put them as parameterized argument. See [example usage of `parameterized.expand`](https://github.com/pytorch/audio/blob/c3cb2015b13c92c1bf777ea251f42bf824f72209/test/torchaudio_unittest/backend/sox_io/save_test.py#L313-L322).
3. Add `norm=None|"slaney"` as a new parameter to be tested.

## Testing
1. Follow [the CONTRIBUTING guide](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md#development-installation) to set up development environment.
2. Run test with;
```
(cd test && pytest torchaudio_unittest/librosa_compatibility_test.py -vk 'MelSpectrogram')
```

worc3131:
Resolves #1262 

I've gone off piste from the steps laid out in the issue. Rather than splitting things out as a test and using parameterize, I decided to maintain more control by using unittest.subTest. I wanted to be able to generate lots of different variants, restrict them down to the relevant parameters for each component, and then remove any resultant duplicates (so as to cut down the test time), as well as maintaining the test skip on TravisCI. Trying to do all this in a decorator for parameterize was messy, so hopefully this approach is better. I'm concerned that it might just be over engineered for what you need though, and more difficult to understand than the dumb approach. Let me know what you think, I can drop back to doing the recommended way if you would like.

The change was tested by running:

(cd test && pytest torchaudio_unittest/librosa_compatibility_test.py)

### DIFF
--- a/test/torchaudio_unittest/librosa_compatibility_test.py
+++ b/test/torchaudio_unittest/librosa_compatibility_test.py
@@ -166,13 +166,6 @@ def _load_audio_asset(*asset_paths, **kwargs):
 class TestTransforms(common_utils.TorchaudioTestCase):
     """Test suite for functions in `transforms` module."""
 
-    @staticmethod
-    def _set_up_sound():
-        common_utils.set_audio_backend('default')
-        path = common_utils.get_asset_path('sinewave.wav')
-        sound, sample_rate = common_utils.load_wav(path)
-        sound_librosa = sound.cpu().numpy().squeeze()  # (64000)
-        return sound, sample_rate, sound_librosa
 
     @parameterized.expand([
         param(n_fft=400, hop_length=200, power=2.0),
@@ -181,7 +174,9 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         param(n_fft=200, hop_length=50, power=2.0),
     ])
     def test_spectrogram(self, n_fft, hop_length, power):
-        sound, sample_rate, sound_librosa = self._set_up_sound()
+        sample_rate = 16000
+        sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
+        sound_librosa = sound.cpu().numpy().squeeze()
         spect_transform = torchaudio.transforms.Spectrogram(
             n_fft=n_fft, hop_length=hop_length, power=power)
         out_librosa, _ = librosa.core.spectrum._spectrogram(
@@ -191,7 +186,7 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         self.assertEqual(out_torch, torch.from_numpy(out_librosa), atol=1e-5, rtol=1e-5)
 
     @parameterized.expand([
-        p._replace(kwargs={'norm': norm, **p.kwargs})
+        param(norm=norm, **p.kwargs)
         for p in [
             param(n_fft=400, hop_length=200, n_mels=128),
             param(n_fft=600, hop_length=100, n_mels=128),
@@ -200,7 +195,9 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         for norm in [None, 'slaney']
     ])
     def test_mel_spectrogram(self, n_fft, hop_length, n_mels, norm):
-        sound, sample_rate, sound_librosa = self._set_up_sound()
+        sample_rate = 16000
+        sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
+        sound_librosa = sound.cpu().numpy().squeeze()
         melspect_transform = torchaudio.transforms.MelSpectrogram(
             sample_rate=sample_rate, window_fn=torch.hann_window,
             hop_length=hop_length, n_mels=n_mels, n_fft=n_fft, norm=norm)
@@ -213,7 +210,7 @@ class TestTransforms(common_utils.TorchaudioTestCase):
             torch_mel.type(librosa_mel_tensor.dtype), librosa_mel_tensor, atol=5e-3, rtol=1e-5)
 
     @parameterized.expand([
-        p._replace(kwargs={'norm': norm, **p.kwargs})
+        param(norm=norm, **p.kwargs)
         for p in [
             param(n_fft=400, hop_length=200, power=2.0, n_mels=128),
             param(n_fft=600, hop_length=100, power=2.0, n_mels=128),
@@ -226,7 +223,9 @@ class TestTransforms(common_utils.TorchaudioTestCase):
     def test_s2db(self, n_fft, hop_length, power, n_mels, norm, skip_ci=False):
         if skip_ci and 'CI' in os.environ:
             self.skipTest('Test is known to fail on CI')
-        sound, sample_rate, sound_librosa = self._set_up_sound()
+        sample_rate = 16000
+        sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
+        sound_librosa = sound.cpu().numpy().squeeze()
         spect_transform = torchaudio.transforms.Spectrogram(
             n_fft=n_fft, hop_length=hop_length, power=power)
         out_librosa, _ = librosa.core.spectrum._spectrogram(
@@ -257,13 +256,14 @@ class TestTransforms(common_utils.TorchaudioTestCase):
     @parameterized.expand([
         param(n_fft=400, hop_length=200, n_mels=128, n_mfcc=40),
         param(n_fft=600, hop_length=100, n_mels=128, n_mfcc=20),
-        # NOTE: Test passes offline, but fails on TravisCI (and CircleCI), see #372.
         param(n_fft=200, hop_length=50, n_mels=128, n_mfcc=50, skip_ci=True),
     ])
     def test_mfcc(self, n_fft, hop_length, n_mels, n_mfcc, skip_ci=False):
         if skip_ci and 'CI' in os.environ:
             self.skipTest('Test is known to fail on CI')
-        sound, sample_rate, sound_librosa = self._set_up_sound()
+        sample_rate = 16000
+        sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
+        sound_librosa = sound.cpu().numpy().squeeze()
         librosa_mel = librosa.feature.melspectrogram(
             y=sound_librosa, sr=sample_rate, n_fft=n_fft,
             hop_length=hop_length, n_mels=n_mels, htk=True, norm=None)
@@ -296,7 +296,9 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         param(n_fft=200, hop_length=50),
     ])
     def test_spectral_centroid(self, n_fft, hop_length):
-        sound, sample_rate, sound_librosa = self._set_up_sound()
+        sample_rate = 16000
+        sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
+        sound_librosa = sound.cpu().numpy().squeeze()
         spect_centroid = torchaudio.transforms.SpectralCentroid(
             sample_rate=sample_rate, n_fft=n_fft, hop_length=hop_length)
         out_torch = spect_centroid(sound).squeeze().cpu()
--- a/test/torchaudio_unittest/librosa_compatibility_test.py
+++ b/test/torchaudio_unittest/librosa_compatibility_test.py
@@ -215,15 +215,11 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         self.assertEqual(
             power_to_db_torch.type(db_librosa_tensor.dtype), db_librosa_tensor, atol=5e-3, rtol=1e-5)
 
-    def assert_compatibilities_mfcc(self, n_fft, hop_length, n_mfcc, norm, sound, sample_rate, sound_librosa):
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -165,13 +165,7 @@ def _load_audio_asset(*asset_paths, **kwargs):
 @unittest.skipIf(not LIBROSA_AVAILABLE, "Librosa not available")
 class TestTransforms(common_utils.TorchaudioTestCase):
     """Test suite for functions in `transforms` module."""
-    def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sample_rate):
-        common_utils.set_audio_backend('default')
-        path = common_utils.get_asset_path('sinewave.wav')
-        sound, sample_rate = common_utils.load_wav(path)
```
worc3131, line 171: The sample_rate variable on assert_compatabilities was not being used as it was overwritten here. I've removed it, and I removed it from the parameter dictionaries below.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -244,51 +252,76 @@ def assert_compatibilities_spectral_centroid(self, sample_rate, n_fft, hop_lengt
...
+        for test_name, test_fn in compat_tests.items():
+            # what variables does our test function take
+            varnames = test_fn.__code__.co_varnames
+            params_list_restricted = [
+                {k: v for k, v in params.items() if k in varnames}
+                for params in params_list
+            ]
+            data_dict_restricted = {k: v for k, v in data_dict.items() if k in varnames}
+            # restrict to unique parameter combinations
+            params_list_unique = [dict(uniq) for uniq in {tuple(sorted(d.items())) for d in params_list_restricted}]
```
worc3131, line 319: I would ideally like to implement this in a util module and just have util.unique_dicts(params_list_restricted) here, so that it's clearer what this is doing and a comment isn't needed. Do you have a util module where this could go?

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -180,19 +174,30 @@ def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sampl
...
-            hop_length=hop_length, n_mels=n_mels, htk=True, norm=None)
+            hop_length=hop_length, n_mels=n_mels, htk=True, norm=norm)
         librosa_mel_tensor = torch.from_numpy(librosa_mel)
         torch_mel = melspect_transform(sound).squeeze().cpu()
         self.assertEqual(
             torch_mel.type(librosa_mel_tensor.dtype), librosa_mel_tensor, atol=5e-3, rtol=1e-5)
 
-        # test s2db
+    def assert_compatibilities_s2db(self, n_fft, hop_length, power, n_mels, norm, sound, sample_rate, sound_librosa):
+        spect_transform = torchaudio.transforms.Spectrogram(
```
worc3131, line 190: The issue talked about splitting out MelSpectrogram was assert_compatibilities, but it was being reused for the s2db and MFCC checks, so I decided to split it all out into individual a bunch of individual functions with some repeated set up work. Let me know what you think.

mthrok (commented):
So as to improve the test, we would like to run;
  1. different kinds of unit tests independently and 
  2. different parameters for the same uni test independently.

The original implementation met 2 by splitting the parameters into different test methods, but did not 1 because all the tests were implemented in the same method. (note that one method executed equals one unit test)

Using the `for` loop inside of tests to test different parameters is what we do not want, because, when one configuration fails, then the remaining tests are not executed. `subTest` might make it work but it's not good enough because when querying types of tests, the set of configurations defined under one method are counted together. They are not found until executed. Parameterizing with decorator nicely solves this. In addition to that, using decorator makes the test logic and parameters clearly separated. So please do not use `for` loop and `subTest`, instead use `parameterized.expand` like the all other tests do.

I like your suggestion of splitting the individual tests. This can achieve the point 1. Then all you need to do is to decorate the split test with `parameterized.expand`. Which will look like,

```python
@parameterized.expand([
    ...,
])
def test_spectrogram(self, ...):
   ...

@parameterized.expand([
    ...,
])
def test_melspectrogram(self, ...):
    ...

@parameterized.expand([
    ...,
])
def test_s2db(self, ...):
    ...

@parameterized.expand([
    ...,
])
def test_mfcc(self, ...):
    ...
```

With this approach, the test code becomes way much readable than using `for` loop. Also all the configuration becomes as if they are defined as different method. (they actually are)
The code for unit tests are typically maintained by those who do not have context of the library, so they have to be really simple and readable.

test/torchaudio_unittest/librosa_compatibility_test.py
mthrok: I honestly do no get what this par of the code is trying to achieve, but I think this is unnecessary simply if you use `parameterized.expand`.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -244,51 +252,78 @@ def assert_compatibilities_spectral_centroid(self, sample_rate, n_fft, hop_lengt
...
+                'n_mfcc': 40,
+            },
+            {
+                'n_fft': 200,
+                'hop_length': 50,
+                'power': 2.0,
+                'n_mels': 128,
+                'n_mfcc': 50,
+                'skip_CI': True,  # NOTE: Test passes offline, but fails on TravisCI (and CircleCI), see #372.
+            },
```
mthrok, line 292: Please move the parameters outside of test definition by using `parameterized.expand`

test/torchaudio_unittest/librosa_compatibility_test.py
mthrok: Sounds good.

test/torchaudio_unittest/librosa_compatibility_test.py
mthrok: Yeah, for the simplicity of task, I was going to suggest recreating the variable used by tests following MelSpectrogram, but splitting all the tests is much better for maintainability.

worc3131:
Hi @mthrok thanks for looking at this. I'm happy to go down the other route, explicit is better than implicit after all! I've written out the parameters for each test and deduplicated manually, which is a bit bulky but such is life. Having looked at unittest.subTest again this morning, it would expand out to give you the individual test results that you would like, except that it is not fully supported by pytest without a plug in, so it might be a nice tool for the future if not now.  https://github.com/pytest-dev/pytest/issues/1367

test/torchaudio_unittest/librosa_compatibility_test.py
worc3131: Ye it's fighting python, it would just be list(set(params_list_restricted)) to get the unique dicts except that dicts a mutable and so unhashable. It's not needed now.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -165,13 +165,23 @@ def _load_audio_asset(*asset_paths, **kwargs):
 @unittest.skipIf(not LIBROSA_AVAILABLE, "Librosa not available")
 class TestTransforms(common_utils.TorchaudioTestCase):
     """Test suite for functions in `transforms` module."""
-    def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sample_rate):
+
+    @staticmethod
+    def _set_up_sound():
         common_utils.set_audio_backend('default')
         path = common_utils.get_asset_path('sinewave.wav')
         sound, sample_rate = common_utils.load_wav(path)
```
mthrok, line 173: Could you 
1. Instead of loading data from file, generate one?
    ```python
    sample_rate = 16000
    sound = common_utils.get_sinusoid(n_channels=1, sample_rate=sample_rate)
    ```
2. Instead of making a common method, put the above code in each test method?

These points were not included in the original issue description, but also important and I would like to take advantage of this opportunity to address them.

The first point is to make the unit test not depend of other components. When our load function has some issue, this test could also fail, which makes it harder to see what's going on when looking at log. Unit test should be simple and independent. The second point is the extension of independency. Having common method often makes new contributors think that they have to use the same method for some reasons they do not know. If it's related to the core of the test logic, it is more beneficial to make fixture common, but in this case it is not, so making each unit test independent each other gives better maintenance experience.

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -180,19 +190,54 @@ def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sampl
...
-        # test mel spectrogram
+    @parameterized.expand([
+        p._replace(kwargs={'norm': norm, **p.kwargs})
+        for p in [
+            param(n_fft=400, hop_length=200, n_mels=128),
+            param(n_fft=600, hop_length=100, n_mels=128),
+            param(n_fft=200, hop_length=50, n_mels=128),
+        ]
+        for norm in [None, 'slaney']
+    ])
```
mthrok, line 201: Use of internal method `_replace` should be avoided.
We use `@parameterized.expand(list(itertools.product([....], [...]))` pattern for this kind of nesting.
Could you try the same?

test/torchaudio_unittest/librosa_compatibility_test.py
```diff
@@ -180,19 +190,54 @@ def assert_compatibilities(self, n_fft, hop_length, power, n_mels, n_mfcc, sampl
...
 
-        # test s2db
+    @parameterized.expand([
+        p._replace(kwargs={'norm': norm, **p.kwargs})
+        for p in [
+            param(n_fft=400, hop_length=200, power=2.0, n_mels=128),
+            param(n_fft=600, hop_length=100, power=2.0, n_mels=128),
+            param(n_fft=400, hop_length=200, power=3.0, n_mels=128),
+            # NOTE: Test passes offline, but fails on TravisCI (and CircleCI), see #372.
+            param(n_fft=200, hop_length=50, power=2.0, n_mels=128, skip_ci=True),
```
mthrok, line 222: Could you verify that both `test_s2db` and `test_mfcc` fail?
If one of them does not fail on CI, we would like to run the test.
From the previous implementation, I could not tell which part of the tests were failing.

mthrok (commented):
Thank you for the update. Now the resulting code look much easier to understand.

test/torchaudio_unittest/librosa_compatibility_test.py
worc3131: Despite having a leading underscore, I don't believe that `_replace` is an internal method. `parameterized.param` is a `collection.namedtuple`. As per the python docs the leading underscore here is to prevent naming conflicts, not because it is an internal method: "In addition to the methods inherited from tuples, named tuples support three additional methods and two attributes. To prevent conflicts with field names, the method and attribute names start with an underscore." That said, my fingers did shake when I typed it out, so I've replaced it with something else. I don't think that itertools.product would work well here, as it's more suited for trying every combination (the cartesian product) of n_fft, hop_length and n_mels.

test/torchaudio_unittest/librosa_compatibility_test.py
worc3131: I removed the `test_mfcc` skip and it passed. The `test_s2db` skip fails as per here:

https://app.circleci.com/pipelines/github/pytorch/audio/5014/workflows/5c7c3a8d-1e26-4b07-83dd-ed4ce2d5c22f/jobs/162780

test/torchaudio_unittest/librosa_compatibility_test.py
mthrok: Your update looks good to me. I was worried that the use of `_replace`, even if it's not internal method, forces contributors to learn the internal of `parameterized.param`, which would raise the bar for writing tests, which could lead to the situation where contributors would not like to contribute anything at all. But now the final form looks good and even better than `itertool` approach I ended up with. Thanks.

test/torchaudio_unittest/librosa_compatibility_test.py
mthrok: Thanks.

mthrok (approved):
Looks good. Thank you!

worc3131:
Perfect, thanks for all the feedback

Issue #1265, opened 2021-02-13
Use switch statements


mthrok:
In torchaudio's `libsox` bindings (I/O and sox effects), there are couple of locations where different operation is applied based on Tensor's dtype.

Currently the code uses `caffe2::TypeMeta` type to check the dtype, such as the following example;

https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/effects_chain.cpp#L83-L107

For this kind of execution flow, using `switch` statement makes the code cleaner. `caffe2::TypeMeta` is not an integral type so `switch` statement is not applicable. Converting it to `c10::ScalarType` make s it possible to use `switch` statement. For the above case, the resulting code looks like the following.

```c++
  switch (tensor_.dtype().toScalarType()) {
  case c10::ScalarType::Float: {
    auto ptr = tensor_.data_ptr<float_t>();
    for (size_t i = 0; i < *osamp; ++i) {
      obuf[i] = SOX_FLOAT_32BIT_TO_SAMPLE(ptr[i], effp->clips);
    }
    break;
  }
  case c10::ScalarType::Int: {
    auto ptr = tensor_.data_ptr<int32_t>();
    for (size_t i = 0; i < *osamp; ++i) {
      obuf[i] = SOX_SIGNED_32BIT_TO_SAMPLE(ptr[i], effp->clips);
    }
    break;
  }
  case c10::ScalarType::Short: {
    auto ptr = tensor_.data_ptr<int16_t>();
    for (size_t i = 0; i < *osamp; ++i) {
      obuf[i] = SOX_SIGNED_16BIT_TO_SAMPLE(ptr[i], effp->clips);
    }
    break;
  }
  case c10::ScalarType::Byte: {
    auto ptr = tensor_.data_ptr<uint8_t>();
    for (size_t i = 0; i < *osamp; ++i) {
      obuf[i] = SOX_UNSIGNED_8BIT_TO_SAMPLE(ptr[i], effp->clips);
    }
    break;
  }
  default:
    throw std::runtime_error("Unexpected dtype.");
  }
```

We would like to apply the same pattern to all the usage of `caffe2::TypeMeta`. The following is the list of such usage.

https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L105-L110

[...]

PR #1270, opened 2021-02-15
Replaced if-elseif-else with switch

prabhat00155:
Addresses https://github.com/pytorch/audio/issues/1265.

### DIFF
--- a/torchaudio/csrc/sox/utils.cpp
+++ b/torchaudio/csrc/sox/utils.cpp
@@ -102,11 +102,15 @@ void validate_input_tensor(const torch::Tensor tensor) {
     throw std::runtime_error("Input tensor has to be 2D.");
   }
 
-  const auto dtype = tensor.dtype();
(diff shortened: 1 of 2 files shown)


mthrok (approved):
Looks good, thanks!

torchaudio/csrc/sox/utils.cpp
```diff
@@ -209,22 +213,25 @@ namespace {
 
 std::tuple<sox_encoding_t, unsigned> get_save_encoding_for_wav(
     const std::string format,
-    const caffe2::TypeMeta dtype,
+    caffe2::TypeMeta dtype,
```
mthrok, line 216: I'm surprised that `toScalarType` is not marked as `const`. Do you think it makes sense to change the function signature to pass around `c10::ScalarType`, instead of `caffe2::TypeMeta`? (`caffe2` itself is absorbed to PyTorch, so seeing it as function signature feels strange too.)

Issue #1264, opened 2021-02-12
Add GSM format support to save function


mthrok:
We would like to add more format support to "sox_io"'s `save` function. [`Full Rate`](https://en.wikipedia.org/wiki/Full_Rate) is the codec used in mobile phone system and the `libsox` torchaudio binds have capability to handle this format.

### Implementation
1. Add `GSM` to [`Format` enum](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/types.h#L9-L18) and [`get_format_from_string`](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/types.cpp#L6-L26) function. The expected string form is `"gsm"`.
2. Add GSM branch in [`get_save_encoding`](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L290-L377) function. Similar to [the case MP3](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L303-L309), GSM does not take any encoding or bit depth option. Use `SOX_ENCODING_GSM, 16` for return value.
3. Add GSM branch in [`get_precision`](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L379) function. the return value should be 16.
### Update documentation
1. Update the docstring of `save` function. (add `"gsm"` in [`format`](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/backend/sox_io_backend.py#L190-L198) section  then [at the bottom before `Note:`](https://github.com/pytorch/audio/blob/master/torchaudio/backend/sox_io_backend.py#L255-L298) add some explanation from http://sox.sourceforge.net/soxformat.html (do text search `"gsm"`

### Add test
[...]

PR #1271, opened 2021-02-16
Added GSM format support to save function

imaginary-person:
### **SUMMARY**
Added the changes listed in #1264 towards supporting GSM format in the save function.

### **CAVEATS**
1. http://sox.sourceforge.net/soxformat.html doesn't list various parameters for `gsm`, so only a generic statement regarding GSM not supporting custom encodings/bit depth/compression has been added in the [documentation](https://github.com/imaginary-person/audio/blob/9b41ada5afb53b1dea8036ed0a5ae66e3addd7a1/torchaudio/backend/sox_io_backend.py#L294-L295), but I can revise it further, if required.

2. I also made some changes in the code-structure for readability, with most of them aimed at conformance with the Google style guide for C++.
I even added curly braces to case blocks within switch statements, but I can roll back such changes, if required, as the Google style guide doesn't mandate one to do so.

3. I'm submitted this PR since the issue has a good first issue label, but if it's still being worked on by someone, please disregard this PR.

Thanks!

### DIFF
--- a/test/torchaudio_unittest/backend/sox_io/save_test.py
+++ b/test/torchaudio_unittest/backend/sox_io/save_test.py
@@ -239,6 +239,12 @@ class SaveTest(SaveTestBase):
 
     @nested_params(
         ["path", "fileobj", "bytesio"],
+    )
+    def test_save_gsm(self, test_mode):
+        self.assert_save_consistency("gsm", test_mode=test_mode)
+
+    @nested_params(
+        ["path", "fileobj", "bytesio"],
         [
             None,
             -1,
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -195,7 +195,7 @@ def save(
             When ``filepath`` argument is file-like object, this argument is required.
 
             Valid values are ``"wav"``, ``"mp3"``, ``"ogg"``, ``"vorbis"``, ``"amr-nb"``,
-            ``"amb"``, ``"flac"`` and ``"sph"``.
+            ``"amb"``, ``"flac"``, ``"sph"``, and ``"gsm"``.
         encoding (str, optional): Changes the encoding for the supported formats.
             This argument is effective only for supported formats, cush as ``"wav"``, ``""amb"``
             and ``"sph"``. Valid values are;
@@ -290,6 +290,9 @@ def save(
 
     ``"amr-nb"``
         Bitrate ranging from 4.75 kbit/s to 12.2 kbit/s. Default: 4.75 kbit/s
+
+    ``"gsm"``
+        The GSM 06.10 Lossy Speech Compression only supports its default configuration.
 
     Note:
         To save into formats that ``libsox`` does not handle natively, (such as ``"mp3"``,
--- a/torchaudio/csrc/sox/types.cpp
+++ b/torchaudio/csrc/sox/types.cpp
@@ -4,22 +4,25 @@ namespace torchaudio {
 namespace sox_utils {
 
 Format get_format_from_string(const std::string& format) {
-  if (format == "wav")
+  if (format == "wav") {
     return Format::WAV;
-  if (format == "mp3")
+  } else if (format == "mp3") {
     return Format::MP3;
-  if (format == "flac")
+  } else if (format == "flac") {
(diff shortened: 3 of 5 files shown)


mthrok:
Hi @imaginary-person 

Thanks for the contribution. Let me talk with the original contributor first. If you would like, you can try HTK format too. It should be practically the same process.

imaginary-person:
Thank you, @mthrok. It seems that like `gsm`, even `htk` only uses default (non-customizable) `compression`/`encoding`/`bits_per_sample`. Please correct me if I'm wrong. Thanks again!

imaginary-person:
Hello @mthrok, `SOX_ENCODING_HTK` doesn't exist in the `sox` source-code. HTK encoding in `sox` actually uses `SOX_ENCODING_SIGN2` under the hood, but since `SOX_ENCODING_HTK` can't simply be substituted with `SOX_ENCODING_SIGN2`, please advise if a workaround can be used. Thanks!

mthrok:
> `SOX_ENCODING_HTK` doesn't exist in the `sox` source-code. HTK encoding in `sox` actually uses `SOX_ENCODING_SIGN2` under the hood

HTK format uses signed-integer encoding. [[code](https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/htk.c#L70)]

> but since `SOX_ENCODING_HTK` can't simply be substituted with `SOX_ENCODING_SIGN2`

Why not? There are other formats that saves samples in raw format like `wav`, such as [amb](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L300-L302), [sph](https://github.com/pytorch/audio/blob/05bff83fdec3e8f70f80bf7a1b89951bf7050114/torchaudio/csrc/sox/utils.cpp#L336-L373). I expect that using `SOX_ENCODING_SIGN2` should allow us to replicate the result `sox` produces.

> If I make `Format::HTK` correspond to `SOX_ENCODING_SIGN2` in `torchaudio/csrc/sox/utils.cpp`, then the save tests for `HTK` fail:
> 
> ```
> save_test.py:250: in test_save_htk
>     self.assert_save_consistency("htk", test_mode=test_mode)
> save_test.py:140: in assert_save_consistency
>     self.assertEqual(found, expected)
> /home/pytorch/torch/testing/_internal/common_utils.py:1183: in assertEqual
>     super().assertTrue(result, msg=self._get_assert_msg(msg, debug_msg=debug_msg))
> E   AssertionError: False is not true : Tensors failed to compare as equal!Attempted to compare equality of tensors with different sizes. Got sizes torch.Size([1, 48000]) and torch.Size([1, 24000]).
> ```

This is most likely the underlying encoder discards the channel information. http://sox.sourceforge.net/soxformat.html says, 

> .htk **Single channel** 16-bit PCM format used by HTK, a toolkit for building Hidden Markov Model speech processing tools.

Looking at the [code](https://github.com/dmkrepo/libsox/blob/b9dd1a86e71bbd62221904e3e59dfaa9e5e72046/src/htk.c#L70), this format seems to have a specific header but the sample storage is same as WAVE format. Also the source code has the link for the references. (if you are interested in)

https://labrosa.ee.columbia.edu/doc/HTKBook21/node57.html
https://labrosa.ee.columbia.edu/doc/HTKBook21/node61.html
https://labrosa.ee.columbia.edu/doc/HTKBook21/node58.html

Anyways, you can provide `num_channels` parameter to change the number of channels used in test. Can you try that?

imaginary-person:
Thanks for your help, @mthrok. I hadn't noticed that the number of channels used in that test were 2 by default.
By providing a `num_channels` parameter, tests for `htk` also passed, as you had expected.
I'll submit a separate PR for `htk`, so as to retain the existing code-style (I changed style in some files of this PR).

> Why not? There are other formats that saves samples in raw format like wav, such as amb, sph. I expect that using SOX_ENCODING_SIGN2 should allow us to replicate the result sox produces.

Yes, I just meant that I'd have to make some other change as well, such as the `num_channels` one you suggested.

imaginary-person:
Closed this PR as the original contributor submitted another one.

PR #1275, opened 2021-02-17
Added GSM format support to soxio's save function

prabhat00155:
Addresses https://github.com/pytorch/audio/issues/1264.

PR #1276, opened 2021-02-17
Add HTK format support to sox_io's save & info

imaginary-person:
With [advice & help](https://github.com/pytorch/audio/pull/1271#issuecomment-780040899) from @mthrok, added `htk` format to sox_io's save function & info test.

htk is a single channel 16-bit PCM format used by HTK, a toolkit for building Hidden Markov Model speech processing tools.

### DIFF
--- a/torchaudio/csrc/sox/utils.cpp
+++ b/torchaudio/csrc/sox/utils.cpp
@@ -424,10 +424,10 @@ unsigned get_precision(const std::string filetype, caffe2::TypeMeta dtype) {
   if (filetype == "amr-nb") {
     return 16;
   }
-  if (filetype == "htk") {
+  if (filetype == "gsm") {
     return 16;
   }
-  if (filetype == "gsm") {
+  if (filetype == "htk") {
     return 16;
   }
   throw std::runtime_error("Unsupported file type: " + filetype);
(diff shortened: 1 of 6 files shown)


mthrok (approved):
Hi @imaginary-person 

Thanks for the contribution. The PR looks good. Can you resolve the conflict due to #1275 ?

imaginary-person:
>  Can you resolve the conflict due to #1275 ?

Done. Thank you!

torchaudio/csrc/sox/utils.cpp
```diff
@@ -417,8 +424,12 @@ unsigned get_precision(const std::string filetype, caffe2::TypeMeta dtype) {
...
     return 16;
   }
-  if (filetype == "gsm")
+  if (filetype == "htk") {
     return 16;
+  }
+  if (filetype == "gsm") {
+    return 16;
+  }
   throw std::runtime_error("Unsupported file type: " + filetype);
```
mthrok, line 433: While this is technically correct, can you add HTK after GSM so that diff is simpler?

imaginary-person: Done. Thanks!
All builds are failing for unrelated reasons, BTW.

mthrok:
Thanks!

mthrok:
@imaginary-person 

I forgot to mention adding one validation. Could you do a follow-up?
Since HTK only supports one-channel data, we would like to raise an error if the input tensor is multi-channel. In the following two locations can you add a check for the number of channels, similar to `amr-nb` case?

https://github.com/pytorch/audio/blob/3488f3142c734326b54badfba4b166173647d1b2/torchaudio/csrc/sox/io.cpp#L265-L271

https://github.com/pytorch/audio/blob/3488f3142c734326b54badfba4b166173647d1b2/torchaudio/csrc/sox/io.cpp#L131-L135

imaginary-person:
@mthrok, thanks a lot for all your help so far!
I submitted #1291, but don't have the requisite permission to add you as a reviewer.

PR #1281, opened 2021-02-18
Remove instructions to build without building sox from README

vincentqb:


mthrok:
Considering the fact that we are now patching `libsox` and the process of building `libsox` has been stabilized, we should stop relying on external `libsox`.

vincentqb:
Indeed. I removed the instructions to build against external sox then.

mthrok (approved):
Looks good. Thanks!

cpuhrsch (approved):

PR #1285, opened 2021-02-19
torchaudio: Fix c extension compilation

seemethere:
After https://github.com/pytorch/pytorch/commit/a9f5e7229e92259bc44eb9898c04eaf200c7f81f was merged into upstream pytorch the passing of
references within torch c extension functions is no longer allowed. This
just removes the reference and passes by value instead.

Signed-off-by: Eli Uriegas <eliuriegas@fb.com>

PR #1300, opened 2021-02-23
Remove reference_cast in make_boxed_from_unboxed_functor (pytorch/pytorch#51319)

fmassa:
This PR cherry-picks a0446251ef7c2806fe36c601631fcae54667318d on top of master, by taking into account the changes from #1285

Summary:
Pull Request resolved: https://github.com/pytorch/pytorch/pull/51319

We were going out of our way to accommodate `IValue::to<Tensor>` returning a copy of the inner Tensor. `IValue::toTensor` is capable of returning a reference without copying, so if we use it directly, we can allow kernels that want to take `Tensor &` to do so!
As a bonus, we get reduced build times.
ghstack-source-id: 121378961

Reviewed By: bhosmer

Differential Revision: D26138549

fbshipit-source-id: b0f830527da360c542c815bef2f7e1692615b32a


mthrok (approved):
Thank you!

PR #1306, opened 2021-02-23
Fix uninitialised variable in sox/utils.cpp

fmassa:
When compiling torchaudio with `-Werror,-Wuninitialized`, we would get compilation errors due to `dummy` not being initialized as follows.

```
 pytorch/audio/torchaudio/csrc/sox/utils.cpp:179:53: error: variable 'dummy' is uninitialized when used here [-Werror,-Wuninitialized]
      ptr[i] = SOX_SAMPLE_TO_FLOAT_32BIT(buffer[i], dummy);
```

This PR fixes it by initializing it to zero.


mthrok (approved):
Looks good. Thanks!

PR #1291, opened 2021-02-19
Check for the number of channels in an htk file being 1

imaginary-person:
`htk` only supports single-channel, so check that the number of channels in an `htk` file or file-object being saved is 1.

PR #1308, opened 2021-02-23
Added HTK format support to sox_io's save & info

imaginary-person:
Cherrypicking #1276 & #1291 from the master branch


mthrok:
Note that the failing test `ci/circleci: unittest_linux_cpu_py3.9` is due to the fact that CPU PyTorch is getting installed even though the code is requesting CPU-only. So you can ignore it.

mthrok:
@imaginary-person Thanks!

Issue #704, opened 2020-06-07
Backprop support for lfilter


FBMachine:
## 🚀 Feature
It is currently not possible to backpropagate gradients through an lfilter because of this inplace operation:
https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L661

## Motivation
It's not worth the pytorch overhead to even use lfilter without backprop support (it's much faster when implemented using e.g. numba). When I saw that this was implemented here, I was hoping to use it instead of my own implementation (which is implemented as a custom RNN) as it is honestly too slow.

## Pitch

I would love to see that inplace operation replaced with something that would allow supporting backprop. I'm not sure what the most efficient way to do this is.

## Alternatives

I implemented transposed direct form II digital filters as custom RNNs, but the performance is pretty poor (which seems to be a problem with the fuser). This is the simplest version I tried, which works, but as I said it's quite slow.

```python
class DigitalFilterModel(jit.ScriptModule):
  def __init__(self):
    super(DigitalFilterModel, self).__init__()

  @jit.script_method
  def forward(self, x, coeffs, v1, v2, v3):
    # type: (Tensor, Tensor, Tensor, Tensor, Tensor) -> Tuple[Tensor, Tensor, Tensor, Tensor]
    seq_len = x.shape[1]
    output = torch.jit.annotate(List[Tensor], [])
    x = x.unbind(1)
    coeffs = coeffs.unbind(1)
    for i in range(seq_len):
      sample = x[i]
      out = coeffs[0] * sample + v1
      output.append(out)

      v1 = coeffs[1] * sample - coeffs[4] * out + v2
      v2 = coeffs[2] * sample - coeffs[5] * out + v3
      v3 = coeffs[3] * sample - coeffs[6] * out

    return torch.stack(output, 1), v1, v2, v3
```

Another alternative I've used when I only need to backprop through the filter, but not optimize the actual coefficients, is to take advantage of the fact that tanh is close to linear for very small inputs and design a standard RNN to be equivalent to the digital filter. Crushing the input, then rescaling the output to keep it linear gives a result very close to the original filter, but this is obviously quite a hack:

```python
class RNNTDFWrapper(nn.Module):
  def __init__(self, eps=0.000000001):
    super(RNNTDFWrapper, self).__init__()
    self.eps = eps
    self.rnn = nn.RNN(1, 4, 1, False, True)

  def set_coefficients(self, coeffs):
[...]

Issue #676, opened 2020-06-02
lfilter function numerical stability


jimchen90:
## 🐛 Bug

When using `lfilter` function in `torchaudio/functional.py`, I input two groups of biquad coefficient parameters (one group is normalized by a0, and the other is unnormalized). The maximum error between two output waveforms is 6e-4.
## To Reproduce

I have a group of coefficients calculated from the bass biquad filter:

```
[b0, b1, b2, a0, a1, a2] = 
[27.26210639724472 -37.770844011587144 14.561748502365816 20.54382514976342 -39.77708440115872 19.27378936027553]
```

After normalization by a0:
```
[b0/a0, b1/a0, b2/a0, a0/a0, a1/a0, a2/a0] = 
[1.3270219250069244 -1.8385497216920241 0.7088138842796521 1.0 -1.9362063350513272 0.9381791959272727]
```

The details of coefficient calculation are shown in Additional Test (list below). Then I compare the output waveforms and the number of elements that has error (absolute difference of output values) larger than 1e-4 is 12603. The maximum error between two waveforms is 6e-4.

```
filename = 'whitenoise.wav'
noise_filepath = common_utils.get_asset_path(filename)
waveform, sample_rate = torchaudio.load(noise_filepath, normalization=True)
device = waveform.device
dtype = waveform.dtype

output_waveform1 = F.lfilter(
waveform,
torch.tensor([a0, a1, a2], dtype=dtype, device=device),
torch.tensor([b0, b1, b2], dtype=dtype, device=device),
)

output_waveform2 = F.lfilter(
waveform,
torch.tensor([a0/a0, a1/a0, a2/a0], dtype=dtype, device=device),
torch.tensor([b0/a0, b1/a0, b2/a0], dtype=dtype, device=device),
)

atol = 1e-4
diff = abs(output_waveform2 - output_waveform1)
print((diff > atol).sum())
v, i = diff.topk(1)
print(v)

```
output result: 
12603
6e-4

## Expected behavior

There is negligible error between two output waveforms.

## Environment

* PyTorch Version: 1.6.0a0+78acc9d
* OS: Linux
* How you installed PyTorch: Conda
* Build command you used : As described above
* Python version: 3.7
* CUDA/cuDNN version: CUDA 10.1

## Additional context

The maximum error deceases when gain decreases.
The method to calculate the coefficients of bass (biquad) filter:
```
def bass_biquad_coefficient(
        #waveform: Tensor,
        sample_rate: int,
        gain: float,
[...]

PR #1310, opened 2021-02-24
Support backprop through lfilter parameters

yoyololicon:
This merge solve issue #704.

It moves the original python implementation of `lfilter` into c++ backend, and register a custom autograd kernel to support torchscript as @vincentqb mentioned in #704 . 

A simple test case is added to test whether the gradient is valid or not.

### Notes
Some differences to the old `lfilter`:

- The old implementation use direct-form I; the new one use direct-form II.
- A mix of indexing and matmul operation at
https://github.com/pytorch/audio/blob/e83d557a0d48249b20eba42952a2ed61a3d9644b/torchaudio/functional/filtering.py#L881

is replaced by a single `conv1d` function call.
https://github.com/yoyololicon/audio/blob/4e2ff32b50d56ce168fcee872c95ffc6cde82eaa/torchaudio/csrc/lfilter.cpp#L123


mthrok:
Hi @yoyololicon 

Thanks for the contribution. This looks a great improvement.

Questions:
 - Does the new implementation yield the same result as the previous one?
 - What do you think of the coverage of the input value of the autograd check test? Is it enough or should we testing more in different locations?
 - Would it be possible to provide a Python version (ideally as a separate PR)? The thing is that currently we do not build C++ extension on Windows so this approach will break the Windows package. We are planning to resolve this in a coming weeks (say, two weeks), until then we cannot merge this PR. Meanwhile, if you can make the original Python version differentiable, then we can merge it until the Windows support situation is resolved. After that we can discuss turning the rest of `lfilter` implementation into C++.

yoyololicon:
> Hi @yoyololicon
> 
> Thanks for the contribution. This looks a great improvement.
> 
> Questions:
> 
> * Does the new implementation yield the same result as the previous one?

It passed the tests listed in `test/torchaudio_unittest/sox_compatibility_test.py` so I think the output should be the same.

> * What do you think of the coverage of the input value of the autograd check test? Is it enough or should we testing more in different locations?

I think currently a gradient test on second order filter might be enough, cuz it's very common and is also the basis of biquad filters.

> * Would it be possible to provide a Python version (ideally as a separate PR)? The thing is that currently we do not build C++ extension on Windows so this approach will break the Windows package. We are planning to resolve this in a coming weeks (say, two weeks), until then we cannot merge this PR. Meanwhile, if you can make the original Python version differentiable, then we can merge it until the Windows support situation is resolved. After that we can discuss turning the rest of `lfilter` implementation into C++. (if you are fine being blocked on the Windows support, that's okay too)

A custom autograd function in Python frontend would break the torchscript support ability.
I suggest we can fallback to use the original version (which not fully support autograd) when c++ extension is not available on Windows. If this is not the case, I'm fine with waiting for a few weeks.

test/torchaudio_unittest/lfilter_autograd_test.py
```diff
@@ -0,0 +1,52 @@
+import torch
+import torchaudio.functional as F
+from parameterized import parameterized
+from torch.autograd import gradcheck
+from torchaudio_unittest import common_utils
+
+
+class AutogradLfilter(common_utils.TestBaseMixin):
```
mthrok, line 7: Thanks for adding auto-grad test. We were also considering adding the same kind of test (for native complex number support and verifying autograd support of existing functionals/transforms).

This class will be the model implementation of all the following autograd tests so I have couple of suggestions.

1. Please remove the helper methods `_get_a`, `_get_b` and `_get_waveform`, and add the equivalent in each function. (this appears to go against of principle of software engineering, but we want to keep unit tests independent each other as much as possible, and having this kind of helper function typically promotes other contributors to use the same and at the end it becomes difficult reason about the decision of parameters)
2. Please rename this class into `Autograd`, and have function name in each method. (like `test_lfilter_x_grad`)
3. Please split this into three different files like the other tests. (`impl`, `cpu_test`, `cuda_test`) This is for facebook's internal testing framework in which we enable/disable tests per file and we would like to split CPU and CUDA.

yoyololicon:
> What is the performance imprecation of conv1d replacement?

My initial idea is to make the source code much more readable, but seems like it can also bring some performance improvement.

Below is the profiling results on this specific part of lfilter:

```
indexing + matmul
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                  Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg       CPU Mem  Self CPU Mem      CUDA Mem  Self CUDA Mem    # of Calls  
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
           aten::copy_        39.23%     236.839ms        39.23%     236.839ms       1.184ms           0 b           0 b           0 b           0 b           200  
            aten::take        29.64%     178.937ms        29.64%     178.937ms       1.789ms     807.46 Mb     807.46 Mb           0 b           0 b           100  
          aten::matmul         0.39%       2.377ms        28.79%     173.818ms       1.738ms     269.15 Mb    -807.46 Mb           0 b           0 b           100  
      aten::contiguous         0.08%     481.909us        20.59%     124.307ms     621.537us     807.46 Mb           0 b           0 b           0 b           200  
          aten::repeat         0.20%       1.214ms        19.47%     117.533ms       1.175ms       1.58 Gb           0 b           0 b           0 b           100  
            aten::add_        19.26%     116.257ms        19.26%     116.257ms       1.163ms           0 b           0 b           0 b           0 b           100  
              aten::mm         7.26%      43.848ms         7.30%      44.086ms     440.863us     269.15 Mb           0 b           0 b           0 b           100  
          aten::arange         1.27%       7.642ms         2.62%      15.809ms      26.349us      67.32 Mb           0 b           0 b           0 b           600  
             aten::add         0.97%       5.873ms         1.00%       6.010ms      60.097us     100.93 Mb           0 b           0 b           0 b           100  
           aten::empty         0.32%       1.934ms         0.32%       1.934ms       2.417us       2.73 Gb       2.73 Gb           0 b           0 b           800  
       aten::unsqueeze         0.24%       1.450ms         0.28%       1.698ms       4.246us           0 b           0 b           0 b           0 b           400  
          aten::unfold         0.16%     948.707us         0.25%       1.481ms       4.937us           0 b           0 b           0 b           0 b           300  
            aten::view         0.18%       1.087ms         0.18%       1.087ms       2.718us           0 b           0 b           0 b           0 b           400  
       aten::transpose         0.12%     738.172us         0.17%       1.007ms       5.036us           0 b           0 b           0 b           0 b           200  
         aten::reshape         0.06%     373.146us         0.16%     979.680us       4.898us           0 b           0 b           0 b           0 b           200  
      aten::empty_like         0.06%     343.304us         0.16%     961.195us       9.612us     807.46 Mb           0 b           0 b           0 b           100  
      aten::as_strided         0.15%     894.105us         0.15%     894.105us       0.813us           0 b           0 b           0 b           0 b          1100  
             aten::mul         0.10%     615.297us         0.12%     701.129us       7.011us      12.50 Kb           0 b           0 b           0 b           100  
         aten::resize_         0.08%     458.102us         0.08%     458.102us       1.527us      33.66 Mb      33.66 Mb           0 b           0 b           300  
    aten::_unsafe_view         0.05%     298.902us         0.07%     422.488us       4.225us           0 b           0 b           0 b           0 b           100  
               aten::t         0.05%     281.180us         0.06%     384.573us       3.846us           0 b           0 b           0 b           0 b           100  
          aten::expand         0.04%     237.944us         0.06%     354.969us       1.775us           0 b           0 b           0 b           0 b           200  
          aten::stride         0.05%     272.478us         0.05%     272.478us       0.182us           0 b           0 b           0 b           0 b          1500  
       aten::expand_as         0.01%      84.950us         0.04%     230.088us       2.301us           0 b           0 b           0 b           0 b           100  
              aten::to         0.02%     132.702us         0.02%     132.702us       1.327us           0 b           0 b           0 b           0 b           100  
           aten::alias         0.02%      92.570us         0.02%      92.570us       0.926us           0 b           0 b           0 b           0 b           100  
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
Self CPU time total: 603.711ms

conv1d
----------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                        Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg       CPU Mem  Self CPU Mem      CUDA Mem  Self CUDA Mem    # of Calls  
----------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                aten::conv1d         0.06%     257.202us        99.72%     441.788ms       4.418ms     269.15 Mb           0 b           0 b           0 b           100  
           aten::convolution         0.07%     312.007us        99.66%     441.530ms       4.415ms     269.15 Mb           0 b           0 b           0 b           100  
          aten::_convolution         0.33%       1.447ms        99.59%     441.218ms       4.412ms     269.15 Mb           0 b           0 b           0 b           100  
    aten::mkldnn_convolution        98.70%     437.288ms        98.87%     438.045ms       4.380ms     269.15 Mb           0 b           0 b           0 b           100  
               aten::squeeze         0.24%       1.063ms         0.28%       1.237ms       6.186us           0 b           0 b           0 b           0 b           200  
             aten::unsqueeze         0.23%       1.006ms         0.26%       1.147ms       3.824us           0 b           0 b           0 b           0 b           300  
                 aten::empty         0.12%     551.754us         0.12%     551.754us       5.518us     269.15 Mb     269.15 Mb           0 b           0 b           100  
                  aten::view         0.11%     475.994us         0.11%     475.994us       4.760us           0 b           0 b           0 b           0 b           100  
            aten::as_strided         0.07%     315.852us         0.07%     315.852us       0.632us           0 b           0 b           0 b           0 b           500  
           aten::as_strided_         0.05%     204.563us         0.05%     204.563us       2.046us           0 b           0 b           0 b           0 b           100  
            aten::contiguous         0.03%     117.147us         0.03%     117.147us       0.390us           0 b           0 b           0 b           0 b           300  
----------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
Self CPU time total: 443.038ms
```

The new implementation reduce number of function calls, make the majority computation rely on a single `mkldnn_convolution` call; it also greatly reduce memory usage.

The profile script:

```python
import torch
import torch.nn.functional as F
import torch.autograd.profiler as profiler
import torch.utils.benchmark as benchmark


def index_matmul(x, b):
    n_channel, n_sample = x.shape
    dtype = x.dtype
    device = x.device
    n_order = b.shape[0]

    window_idxs = torch.arange(n_sample - n_order + 1, device=device).unsqueeze(
        0) + torch.arange(n_order, device=device).unsqueeze(1)
    window_idxs = window_idxs.repeat(n_channel, 1, 1)
    window_idxs += (
        torch.arange(
            n_channel, device=device).unsqueeze(-1).unsqueeze(-1) * n_sample
    )
    window_idxs = window_idxs.long()
    input_signal_windows = torch.matmul(
        b, torch.take(x, window_idxs)
    )
    return input_signal_windows

def conv1d(x, b):
    n_channel, n_sample = x.shape
    dtype = x.dtype
    device = x.device
    n_order = b.shape[0]
    return F.conv1d(x.unsqueeze(1), b.view(1, 1, n_order)).squeeze(1)


if __name__ == '__main__':
    torch.random.manual_seed(2434)
    b = torch.rand(3)
    x = torch.randn(16, 44100)
    x /= x.abs().max()
    
    with profiler.profile(profile_memory=True) as prof:
        for _ in range(100):
            index_matmul(x, b)
    
    print("indexing + matmul")
    print(prof.key_averages().table(sort_by="cpu_time_total"))


    with profiler.profile(profile_memory=True) as prof:
        for _ in range(100):
            conv1d(x, b)
    
    print("conv1d")
    print(prof.key_averages().table(sort_by="cpu_time_total"))

    y1 = index_matmul(x, b)
    y2 = conv1d(x, b)
    assert torch.allclose(y1, y2, atol=1e-7)
```

test/torchaudio_unittest/lfilter_autograd_test.py
yoyololicon: Thanks for suggestions!

cpuhrsch:
Thank you for this great contribution @yoyololicon! I'm taking this on to take some workload off of Moto :)

There are so many good things in here, I'd suggest we split this into three pieces so we can land more quickly
a) Landing the conv1d optimization within Python only 
b) Landing a full C++ forward implementation
c) Landing the C++ backward pass which also adds autograd support

a) and b) are fairly straightforward and should be quick to land, c) will need a bit more discussion due to numerical stability details etc.

While you're doing this we might already have solved Windows support for C++ extension, but your work is not blocked on it.

Thanks!
- Christian

cpuhrsch (changes requested):
See comment, please add me as the reviewer :)

yoyololicon (commented):
agree

yoyololicon:
> a) and b) are fairly straightforward and should be quick to land, c) will need a bit more discussion due to numerical stability details etc.

Should we create seperate request for each one?

mthrok:
> > a) and b) are fairly straightforward and should be quick to land, c) will need a bit more discussion due to numerical stability details etc.
> 
> Should we create seperate request for each one?

I recommend opening separate PRs for a) and b). Then after a) and b) are done, you can either close this one or update it to c). For now this one has a good discussion and benchmark numbers, so let's leave it open.

yoyololicon:
Profiling results running on a P620 gpu with same parameters:
```
indexing + matmul
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                  Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg       CPU Mem  Self CPU Mem      CUDA Mem  Self CUDA Mem    # of Calls  
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
          aten::matmul        10.66%       2.024ms        39.36%       7.470ms      74.696us           0 b           0 b     149.41 Mb    -403.76 Mb           100  
          aten::arange        10.72%       2.034ms        24.99%       4.743ms       7.906us           0 b           0 b      67.58 Mb           0 b           600  
          aten::repeat         3.82%     725.022us        18.14%       3.443ms      34.431us           0 b           0 b     807.47 Mb           0 b           100  
           aten::copy_         8.69%       1.650ms         8.69%       1.650ms       8.248us           0 b           0 b           0 b           0 b           200  
      aten::contiguous         1.34%     254.661us         8.33%       1.581ms       7.903us           0 b           0 b     403.76 Mb           0 b           200  
              aten::mm         6.44%       1.223ms         7.65%       1.451ms      14.512us           0 b           0 b     149.41 Mb           0 b           100  
          aten::unfold         4.67%     886.657us         7.16%       1.359ms       4.528us           0 b           0 b           0 b           0 b           300  
            aten::view         7.09%       1.346ms         7.09%       1.346ms       3.365us           0 b           0 b           0 b           0 b           400  
       aten::unsqueeze         5.72%       1.086ms         6.77%       1.285ms       3.212us           0 b           0 b           0 b           0 b           400  
           aten::empty         6.39%       1.213ms         6.39%       1.213ms       1.516us           0 b           0 b       1.43 Gb       1.43 Gb           800  
             aten::add         4.93%     935.917us         5.88%       1.116ms      11.162us           0 b           0 b     100.98 Mb           0 b           100  
            aten::take         5.57%       1.058ms         5.57%       1.058ms      10.578us           0 b           0 b     403.76 Mb     403.76 Mb           100  
             aten::mul         4.59%     870.274us         5.55%       1.054ms      10.539us           0 b           0 b      50.00 Kb           0 b           100  
    aten::_unsafe_view         0.98%     185.492us         4.24%     805.323us       8.053us           0 b           0 b           0 b           0 b           100  
         aten::reshape         1.01%     191.859us         3.65%     692.589us       3.463us           0 b           0 b           0 b           0 b           200  
            aten::add_         3.61%     685.445us         3.61%     685.445us       6.854us           0 b           0 b           0 b           0 b           100  
      aten::as_strided         3.05%     578.604us         3.05%     578.604us       0.526us           0 b           0 b           0 b           0 b          1100  
         aten::resize_         2.73%     518.079us         2.73%     518.079us       1.727us           0 b           0 b      33.79 Mb      33.79 Mb           300  
       aten::transpose         1.85%     350.566us         2.47%     467.854us       2.339us           0 b           0 b           0 b           0 b           200  
      aten::empty_like         0.88%     167.664us         2.38%     452.493us       4.525us           0 b           0 b     403.76 Mb           0 b           100  
          aten::stride         1.96%     371.699us         1.96%     371.699us       0.206us           0 b           0 b           0 b           0 b          1800  
               aten::t         1.17%     222.400us         1.72%     326.600us       3.266us           0 b           0 b           0 b           0 b           100  
          aten::expand         1.01%     191.119us         1.53%     289.546us       1.448us           0 b           0 b           0 b           0 b           200  
       aten::expand_as         0.35%      67.201us         1.02%     193.950us       1.939us           0 b           0 b           0 b           0 b           100  
           aten::alias         0.45%      84.716us         0.45%      84.716us       0.847us           0 b           0 b           0 b           0 b           100  
              aten::to         0.30%      57.575us         0.30%      57.575us       0.576us           0 b           0 b           0 b           0 b           100  
----------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
Self CPU time total: 18.980ms

conv1d
---------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
                       Name    Self CPU %      Self CPU   CPU total %     CPU total  CPU time avg       CPU Mem  Self CPU Mem      CUDA Mem  Self CUDA Mem    # of Calls  
---------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
               aten::conv1d         2.09%     115.218us        87.63%       4.837ms      48.367us           0 b           0 b     134.62 Mb           0 b           100  
          aten::convolution         2.07%     114.072us        85.55%       4.721ms      47.214us           0 b           0 b     134.62 Mb           0 b           100  
         aten::_convolution        11.17%     616.444us        83.48%       4.607ms      46.074us           0 b           0 b     134.62 Mb           0 b           100  
    aten::cudnn_convolution        47.62%       2.628ms        56.20%       3.102ms      31.017us           0 b           0 b     134.62 Mb           0 b           100  
            aten::unsqueeze        10.80%     596.204us        13.80%     761.660us       2.539us           0 b           0 b           0 b           0 b           300  
              aten::squeeze         8.38%     462.517us         9.91%     546.892us       2.734us           0 b           0 b           0 b           0 b           200  
                aten::empty         5.03%     277.499us         5.03%     277.499us       1.387us           0 b           0 b     134.62 Mb     134.62 Mb           200  
           aten::as_strided         4.53%     249.831us         4.53%     249.831us       0.500us           0 b           0 b           0 b           0 b           500  
                 aten::view         3.45%     190.142us         3.45%     190.142us       1.901us           0 b           0 b           0 b           0 b           100  
           aten::contiguous         2.38%     131.520us         2.38%     131.520us       0.329us           0 b           0 b           0 b           0 b           400  
               aten::stride         1.32%      72.752us         1.32%      72.752us       0.182us           0 b           0 b           0 b           0 b           400  
              aten::resize_         1.18%      64.994us         1.18%      64.994us       0.325us           0 b           0 b           0 b           0 b           200  
---------------------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  ------------  
Self CPU time total: 5.519ms

```
The speed improvements is a lot more obvious.

P

cpuhrsch:
@yoyololicon sorry for the confusion.

Now that #1318 merged we landed part a) of this work, which is using conv1d as a performance optimization. Thank you for that!

Next we can work on and merge the C++ version of the forward (b) and then update this PR to only contain the backward function and corresponding harness (c).

Thank you for your contribution and patience.

If you can't add me as a reviewer via the github UI (under section "Reviewers") just cc me on the PR in the description for a faster turn-around.

Thanks,
Christian

yoyololicon:
@cpuhrsch No problem, will start working on (b).

BTW, I saw you are already listed as reviewers, in my side of view.
image

cpuhrsch:
Now that the other two PRs were merged, we can rebase and re-review. I expect this PR will become much shorter :) Thank you for working this @yoyololicon!

yoyololicon (commented):
The custom backward function is added, based on original direct form I.
Seems like `skipIfNoExtension` was removed, is there any way to skip autograd check on windows?

mthrok:
> The custom backward function is added, based on original direct form I.
> Seems like `skipIfNoExtension` was removed, is there any way to skip autograd check on windows?

@yoyololicon 
For the time being please please use 

```
import platform

@unittest.skipIf(platform.system() == 'Windows'm "not supported on windows):
```

Windows support is coming very soon

mthrok (commented):
Could you move the tests into `torchaudio_unittest/functional`, then remove `lfilter` from the file name? I plan to add other autograd test and would like to consolidate all in one file.

yoyololicon:
> a) and b) are fairly straightforward and should be quick to land, c) will need a bit more discussion due to numerical stability details etc.

Currently, the custom backward function is actually just filter the gradients in reversed direction with the same parameters. If there is numerical problems in forward pass, it's very likely to happen in backward.

The instability of lfilter in torchaudio can be a result of:

1. Using float32 single precision (#676). This is unavoidable in PyTorch.
2. Using Direct Form instead of Transposed Form II. Direct Form is simple to implement (just a convolution and a IIR filter), but is well known to have stability issues; Transposed Form II is good for real-time application, but implement it efficiently in offline application is non-trivial.
3. Very high order filter (with order > 2). A solution is factoring the filter into cascaded second-order filters, and is also why SciPy recommend using `scipy.signal.sosfilt` instead of `scipy.signal.lfilter` in their documentation.

In future I would recommend implementing a custom second-order filter using Transposed Form II, and factorize the filter coefficients using root finding algorithm when higher order filter is needed.

cpuhrsch:
Ok that makes sense. As long as these particularities are documented and the user is aware of it based on the docstring, I think we'll be ok. Could you update the docstring for lfilter to explain potential sources for numerical instability and we can mark your suggestions as a follow-up issue?

torchaudio/csrc/lfilter.cpp
```diff
@@ -80,7 +80,7 @@ void lfilter_core_generic_loop(
   }
 }
 
-torch::Tensor lfilter_core(
+std::vector<torch::Tensor> lfilter_core(
```
cpuhrsch, line 83: semi-nit: using a std::tuple<at::Tensor, at::Tensor> should be faster and safer (type checking).

test/torchaudio_unittest/functional/autograd_cpu_test.py
```diff
@@ -0,0 +1,11 @@
+import torch
+from .autograd_impl import Autograd
+from torchaudio_unittest import common_utils
+import platform
+import unittest
+
+
+@unittest.skipIf(platform.system() == 'Windows', "not supported on windows")
```
mthrok, line 8: @yoyololicon We have landed https://github.com/pytorch/audio/pull/1345, and this should work on Windows too. Can you try removing this skip decorator?

yoyololicon:
@mthrok  Seems custom ops didn't get build on windows, could you help checking it?

mthrok:
> @mthrok Seems custom ops didn't get build on windows, could you help checking it?

@yoyololicon Looks like the commits 9395ad6 is missing from your branch. Can you try merge-commit or rebase on the master branch?

yoyololicon:
> > @mthrok Seems custom ops didn't get build on windows, could you help checking it?
> 
> @yoyololicon Looks like the commits [9395ad6](https://github.com/pytorch/audio/commit/9395ad6433a56ca9a8d5642d7f5c5386d3cd086e) is missing from your branch. Can you try merge-commit or rebase on the master branch?

commit merged, doesn't work.

mthrok:
Looking at the installation log from https://app.circleci.com/pipelines/github/pytorch/audio/5389/workflows/cb9960d0-3599-4f7a-9a9f-a597b49cba73/jobs/179910/steps

`lfilter.cpp` is compiled and I see that `lfilter.cpp.obj` is linked against `_torchaudio.dll`.

```
[3/6] C:\PROGRA~2\MICROS~2\2019\COMMUN~1\VC\Tools\MSVC\1428~1.293\bin\Hostx64\x64\cl.exe   /TP -DTORCH_API_INCLUDE_EXTENSION_H -D_torchaudio_EXPORTS -I..\..\..\ -I..\..\..\env\include -I..\..\..\env\Lib\site-packages\torch\include -I..\..\..\env\Lib\site-packages\torch\include\torch\csrc\api\include /DWIN32 /D_WINDOWS /GR /EHsc -Wall  /MD /O2 /Ob2 /DNDEBUG /EHsc /DNOMINMAX /wd4267 /wd4251 /wd4522 /wd4838 /wd4305 /wd4244 /wd4190 /wd4101 /wd4996 /wd4275 /bigobj -std:c++14 /showIncludes /Fotorchaudio\csrc\CMakeFiles\_torchaudio.dir\lfilter.cpp.obj /Fdtorchaudio\csrc\CMakeFiles\_torchaudio.dir\ /FS -c ..\..\..\torchaudio\csrc\lfilter.cpp
```

```
[5/6] cmd.exe /C "cd . && C:\Users\circleci\project\env\Lib\site-packages\cmake\data\bin\cmake.exe -E vs_link_dll --intdir=torchaudio\csrc\CMakeFiles\_torchaudio.dir --rc=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\rc.exe --mt=C:\PROGRA~2\WI3CF2~1\10\bin\100190~1.0\x64\mt.exe --manifests  -- C:\PROGRA~2\MICROS~2\2019\COMMUN~1\VC\Tools\MSVC\1428~1.293\bin\Hostx64\x64\link.exe  torchaudio\csrc\CMakeFiles\_torchaudio.dir\pybind.cpp.obj torchaudio\csrc\CMakeFiles\_torchaudio.dir\lfilter.cpp.obj torchaudio\csrc\CMakeFiles\_torchaudio.dir\overdrive.cpp.obj torchaudio\csrc\CMakeFiles\_torchaudio.dir\utils.cpp.obj  /out:torchaudio\csrc\_torchaudio.dll /implib:torchaudio\csrc\_torchaudio.lib /pdb:torchaudio\csrc\_torchaudio.pdb /dll /version:0.0 /machine:x64 /INCREMENTAL:NO  ..\..\..\env\Lib\site-packages\torch\lib\c10.lib  ..\..\..\env\Lib\site-packages\torch\lib\torch_python.lib  ..\..\..\env\libs\python37.lib  ..\..\..\env\Lib\site-packages\torch\lib\torch.lib  ..\..\..\env\Lib\site-packages\torch\lib\torch_cpu.lib  ..\..\..\env\Lib\site-packages\torch\lib\c10.lib  kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib  && cd ."
```

mthrok:
oh the C++ extension is not initialized properly.

```
..\env\lib\site-packages\torchaudio-0.9.0a0+cd24e10-py3.7-win-amd64.egg\torchaudio\extension\extension.py:14
  c:\users\circleci\project\env\lib\site-packages\torchaudio-0.9.0a0+cd24e10-py3.7-win-amd64.egg\torchaudio\extension\extension.py:14: UserWarning: torchaudio C++ extension is not available.
    warnings.warn('torchaudio C++ extension is not available.')
```

mthrok:
Looks like we need to fix the C++ extension initialization mechanism. Meanwhile you can skip the Windows test as previously done.

yoyololicon:
I see. Will wait until the issue been fixed.

mthrok:
> I see. Will wait until the issue been fixed.

Hi @yoyololicon

I have fixed the Windows issue in https://github.com/pytorch/audio/pull/1371. Can you merge it to your branch and see if it works?

mthrok:
🎉 it's working on Windows as well.

@cpuhrsch Please have another look. Testing-wise I think it's good to go.

torchaudio/functional/filtering.py
```diff
@@ -884,6 +884,10 @@ def lfilter(
 ) -> Tensor:
     r"""Perform an IIR filter by evaluating difference equation.
 
+    Note:
+        To avoid numerical problems, small filter order or short input sequence is prefered.
+        Using double precision could also minimize numerical precision errors.
```
cpuhrsch, line 889: Could I also ask to add a test that demonstrates the numerical precision errors (i.e. higher atol and rtol)? We can then use that test case to track progress. Also, this only applies to the backwards pass as far as I understand?

yoyololicon: Sure, I will add a test that demonstrate the effect of high order filter, probably filtering an unit impulse and check whether the output value exceed `1` or become `nan`. I also found that the length of input sequence doesn't effect much to the instability.

The numerical problem can happen both to the forward and the backward pass, cuz they share the same filter parameters.

yoyololicon:
@cpuhrsch A demonstration test is added.
I use a 9th order butterworth hi-pass filter to filter an unit impulse, the output should be the impulse response of filter.
With enough precision, the impulse response should not have absolute value greater than 1.
I adjust the pass band frequency so when using double precision the response is accurate enough, but when using single precision, the precision errors would accumulate enough to make the response have value greater than 1.

test/torchaudio_unittest/functional/numerical_test_impl.py
```diff
@@ -0,0 +1,23 @@
+import torch
+import torchaudio.functional as F
+from torchaudio_unittest import common_utils
+
+
+class NumericalStability(common_utils.TestBaseMixin):
```
mthrok, line 6: Couple of comments on this test;
1. I think this fits better in [the existing Lfilter test](https://github.com/pytorch/audio/blob/ea857940de9e3738166989ad3bf1726741a13f04/test/torchaudio_unittest/functional/functional_impl.py#L9). Can you move it there?
2. The intention and the assertion is not straightforward for those without context.
    - Can you add a docstring that explains what this test is supposed to be testing, specifically what `y` should look like if there is no numerical issue?
    - `assert torch.any(y.abs() > 1)` sounds strange. This looks like the ideal implementation of `lfilter` for `float32` has to be larger than `1` at somewhere. While I guess what you want to express is that "when `float32`, it is not guaranteed that `y` meets some criteria, like `y.abs() < 0.8`  (the case of `float64`)

yoyololicon: > Couple of comments on this test;
> 
> 1. I think this fits better in [the existing Lfilter test](https://github.com/pytorch/audio/blob/ea857940de9e3738166989ad3bf1726741a13f04/test/torchaudio_unittest/functional/functional_impl.py#L9). Can you move it there?

Sure.

> 2. The intention and the assertion is not straightforward for those without context.
>    
>    * Can you add a docstring that explains what this test is supposed to be testing, specifically what `y` should look like if there is no numerical issue?
>    * `assert torch.any(y.abs() > 1)` sounds strange. This looks like the ideal implementation of `lfilter` for `float32` has to be larger than `1` at somewhere. While I guess what you want to express is that "when `float32`, it is not guaranteed that `y` meets some criteria, like `y.abs() < 0.8`  (the case of `float64`)

Yup... I also found it looks kind of wierd and not straightforward. I was thinking about how to get ground truth data without using third party packages.

@mthrok How about pre-compute the ground truth target value offline using SciPy with high order filter, then load it when doing testing? For torch.float32 and torch.float64 we can give tolerance differently, maybe atol=1e-1 or something, then if we solve the numerical issue, we can lower it down. This sounds more like what @cpuhrsch wanted to do.

I'm not sure if it's ok to upload other file formats like `.npy` or `.pth` in test folder, but I saw some `.wav` and `.json` files in the `torchaudio_unittest/assets` folder.

cpuhrsch: This sounds good! Depending on the size of the test assets in my opinion a ```.pth``` file is fine. Of course another alternative is to have a test dependency on scipy, but I'm leaving that up to @mthrok. The advantage there would be further parameterization and ease of use.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -45,6 +45,24 @@ def test_shape(self, shape):
...
+        x[0] = 1
+        a = torch.tensor([1., -7.60545606, 25.80187885, -51.23717435,
+                          65.62093428, -56.20096964, 32.18274279, -11.88025302,
+                          2.56506938, -0.24677075], dtype=self.dtype, device=self.device)
+        b = torch.tensor([0.49676025, -4.47084227, 17.88336908, -41.72786118,
+                          62.59179178, -62.59179178, 41.72786118, -17.88336908,
+                          4.47084227, -0.49676025], dtype=self.dtype, device=self.device)
+        y = F.lfilter(x, a, b, False)
+        if self.dtype == torch.float64:
+            assert torch.all(y.abs() < 0.8)
```
cpuhrsch, line 62: You should be able to use ```self.assertLess(y.abs, 0.8)``` here, which will give a better error if this test where to fail.

cpuhrsch (approved):
Aside from scipy dependency question to generate test asset (@mthrok) this looks good to go.

yoyololicon:
Hi @mthrok , the stability test case was updated, could you help me review this?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -45,6 +45,34 @@ def test_shape(self, shape):
...
+                          65.62093428, -56.20096964, 32.18274279, -11.88025302,
+                          2.56506938, -0.24677075], dtype=self.dtype, device=self.device)
+        b = torch.tensor([0.49676025, -4.47084227, 17.88336908, -41.72786118,
+                          62.59179178, -62.59179178, 41.72786118, -17.88336908,
+                          4.47084227, -0.49676025], dtype=self.dtype, device=self.device)
+        yhat = F.lfilter(x, a, b, False)
+        if self.dtype == torch.float64:
+            self.assertEqual(yhat, y, atol=1e-4, rtol=1e-5)
+        else:
+            self.assertEqual(yhat, y, atol=10, rtol=1e-5)
```
mthrok, line 74: So my concerns about this approach were
1. it is not intuitively clear what the second assertion is intended.
2. if the implementation is updated and once `float32` passes the test, the code does not give the signal and we might forget to update the test by removing the branching here.

So, how about the following approach?
1. We keep the only the correct assertion.
```python
yhat = F.lfilter(x, a, b, False)
self.assertEqual(yhat, y, atol=1e-4, rtol=1e-5)
```
2. For `float32`, we overload the method in `functional_cpu_test.py` and decorate it with [`unittest.expectedFailure`](https://docs.python.org/3/library/unittest.html#unittest.expectedFailure) where `dtype=float32`.
```python
# in functional_cpu_test.py and functional_cuda_test.py
class TestLFilterFloat32(Lfilter, common_utils.PytorchTestCase):
    ...

    @unittest.expectedFailure
    def test_9th_order_filter_stability(self):
        super().test_9th_order_filter_stability()
```

This way, the test implementation looks like what is intended, and we can correctly acknowledge the failure for `float32`. If we update the implementation and the test starts to pass for `float32`, then the test will be reported as `unexpected success` so that we will not forget to remove the special treatment for `float32`.

mthrok (commented):
Hi @yoyololicon 

Sorry for my late response. I was away from the laptop. Please refer to my comments regarding test readability and maintainability.

test/torchaudio_unittest/functional/functional_impl.py
yoyololicon: Awesome, let's do this.

mthrok (approved):

mthrok:
@yoyololicon Thanks for this great contribution!

yoyololicon:
@mthrok @cpuhrsch Thank you guys for helping me submit this merge!
I'm really appreciated for your help and take time reviewing it.  It's a wonderful experience.

PR #1312, opened 2021-02-24
Removed check_length from validate_input_file()

prabhat00155:
Some audio formats like `gsm` does not have valid frame numbers when opened. But `libsox` can properly handle these audios, so checking if `length > 0` is not necessary and too strict.


mthrok (approved):
LGTM. 

follow-up - add load test for GSM.

prabhat00155:
> LGTM.
> 
> follow-up - add load test for GSM.

Shall I create a different PR for this?

mthrok:
> > LGTM.
> > follow-up - add load test for GSM.
> 
> Shall I create a different PR for this?

If you have time and would like to, go for it.

PR #1315, opened 2021-02-25
Apply functional batch consistency tests to batches of different items

### LINKED ISSUE #1294: Should batch consistency be tested with a batch of non-identical items?
The `batch_consistency_test.py` files check that behaviour is consistent on batches vs. items, but the current infrastructure does this with batches of identical items. This approach may miss bugs related to data leakage between items within a batch, for example #994. I suggest items should be different to ensure those kinds of issues are flagged.

I have an alternate implementation [here](https://github.com/jcaw/audio/tree/batch_consistency_different_items) which uses batches with different items (where possible). Would you like me to open a PR?

(I've just implemented it for the `functional` tests for now, but I can also expand it to the root `batch_consistency_check.py` too.)

jcaw:
The `batch_consistency_test.py` files check that behavior is consistent on batches vs. items, but the current infrastructure does this with batches of identical items. This approach may miss bugs related to data leakage between items within a batch, for example #994.

This PR makes changes `functional/batch_consistency_test.py` so (most) batches are generated with items that are not identical, then the entire batch is passed to the consistency check. Batch sizes are parameterized so tests are generated for single-item and three-item batches. I've also changed the white noise generator to generate different data in each channel (rather than duplicating one wave), and the consistency check now checks both inputs and outputs, to detect inconsistencies with in-place operations.

I am still repeating a single item in a couple of places - with a one-channel signal loaded from disk, and for pitch frequency. Using different frequencies within the same batch may be a more robust way to test pitch frequency.

I haven't touched the root `batch_consistency_test.py` yet, but it seems that this approach could be extended to that file too. Let me know if that sounds sensible.

Closes #1294

### DIFF
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -10,6 +10,8 @@ from torchaudio_unittest import common_utils
 
 
 @parameterized_class([
+    # Single-item batch isolates problems that come purely from adding a
+    # dimension (rather than processing multiple items)
     {"batch_size": 1},
     {"batch_size": 3},
 ])
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -187,11 +187,12 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     def test_vad_from_file(self):
         common_utils.set_audio_backend('default')
-        filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
+        filepath = common_utils.get_asset_path("vad-go-stereo-44100.wav")
         waveform, sample_rate = torchaudio.load(filepath)
-        self.assert_batch_consistency(
-            F.vad, waveform.repeat(self.batch_size, 1, 1),
-            sample_rate=sample_rate)
+        # Each channel is slightly offset - we can use this to create a batch
+        # with different items.
+        batch = waveform.view(2, 1, -1)
+        self.assert_batch_consistency(F.vad, batch, sample_rate=sample_rate)
 
     def test_vad_different_items(self):
         """Separate test to ensure VAD consistency with differing items."""
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -195,6 +195,13 @@ class TestFunctional(common_utils.TorchaudioTestCase):
             F.vad, waveform.unsqueeze(0).repeat(3, 1, 1),
             sample_rate=sample_rate)
 
+    def test_vad_different_items(self):
+        """Separate test to ensure VAD consistency with differing items."""
+        sample_rate = 44100
+        waveforms = torch.rand(3, 2, 100) - 0.5
+        self.assert_batch_consistencies(
+            F.vad, waveforms, sample_rate=sample_rate)
+
     @common_utils.skipIfNoExtension
     def test_compute_kaldi_pitch(self):
         sample_rate = 44100
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -35,7 +35,7 @@ class TestFunctional(common_utils.TorchaudioTestCase):
         batch_input = batch.clone()
         batch_result = functional(batch_input, *args, **kwargs)
 
-        self.assertEqual(items_input,  batch_input,  rtol=rtol, atol=atol)
+        self.assertEqual(items_input, batch_input, rtol=rtol, atol=atol)
         self.assertEqual(items_result, batch_result, rtol=rtol, atol=atol)
 
     def test_griffinlim(self):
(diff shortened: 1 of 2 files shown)


mthrok:
> Batch sizes are parameterized so tests are generated for single-item and three-item batches.

What do you think about the single-item test? I know this is what was there since the early time and I kept it in the first major test refactoring, but I cannot convince myself that it is serving well as a base case, and I feel something like 2 as a base case and 32 as a common use case might suit better for the purpose of the test. of course 32 increases computational cost for running the test.

> I've also changed the white noise generator to generate different data in each channel (rather than duplicating one wave), 

Sounds good. nice to know that it does not disturb other tests.

> and the consistency check now checks both inputs and outputs, to detect inconsistencies with in-place operations.

Nice!

> I am still repeating a single item in a couple of places - with a one-channel signal loaded from disk, and for pitch frequency. Using different frequencies within the same batch may be a more robust way to test pitch frequency.

Testing multiple frequency sounds a good suggestion and improvement.

> I haven't touched the root `batch_consistency_test.py` yet, but it seems that this approach could be extended to that file too. Let me know if that sounds sensible.

I am open for updating the other batch_consistency_test as well, but let's keep it a separate PR. Smaller PRs are easier to review, thus faster approval.

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -140,53 +142,70 @@ def test_amplitude_to_DB_not_channelwise_clamps(self):
...
+            F.sliding_window_cmn, waveforms, center=True, norm_vars=False)
+        self.assert_batch_consistency(
+            F.sliding_window_cmn, waveforms, center=False, norm_vars=True)
+        self.assert_batch_consistency(
+            F.sliding_window_cmn, waveforms, center=False, norm_vars=False)
+
+    def test_vad_from_file(self):
         common_utils.set_audio_backend('default')
         filepath = common_utils.get_asset_path("vad-go-mono-32000.wav")
         waveform, sample_rate = torchaudio.load(filepath)
```
mthrok, line 191: How about using `vad-go-stereo-44100.wav`? Its second channel is slightly delayed, so we can consider it different waveform for the first channel.

Screen Shot 2021-02-25 at 16 38 27

Also could you replace `torchaudio.load` with `common_utils.load_wav`, then remove `backend='default'` from the class? Using I/O function makes this test depend on both `F.vad` and `torchaudio.load`, which is not ideal for unit test as it makes harder to grasp what caused test failure.

jcaw: That's a good idea. I'll use that instead (and switch the loading method).

jcaw: I've updated this to use the dual-channel wav. For now I'm ignoring the parameterized batch dimension and passing a two-item batch, but that means the test is generated twice, so it might be worth extracting it.

jcaw:
>What do you think about the single-item test? 

I assumed that was there to isolate when issues came specifically from adding a dimension, rather than from processing multiple items. I could see the distinction coming up (e.g. reshaping issues) but I can't imagine dropping it would really impede diagnosis.

>2 as a base case and 32 as a common use case might suit better for the purpose of the test. of course 32 increases computational cost for running the test.

That's up to you guys. Just ran a quick test and it pushes the run time for the `batch_consistency_tests.py` suite to about 4 minutes on this laptop (i7-3630QM @ 2.40GHz), which seems a long time, but I haven't benchmarked the other suites. Do you think the overhead is worth it for catching issues that crop up between a batch size of 3 and 32?

>I am open for updating the other batch_consistency_test as well, but let's keep it a separate PR. Smaller PRs are easier to review, thus faster approval.

Awesome. In that case I'll get this done first.

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -140,53 +146,71 @@ def test_amplitude_to_DB_not_channelwise_clamps(self):
         assert (difference >= 1e-5).any()
 
     def test_contrast(self):
-        waveform = torch.rand(2, 100) - 0.5
-        self.assert_batch_consistencies(
-            F.contrast, waveform, enhancement_amount=80.)
+        torch.random.manual_seed(0)
+        waveforms = torch.rand(self.batch_size, 2, 100) - 0.5
```
mthrok, line 150: Could you also take this opportunity to replace these waveform generation with `get_whitenoise`?

mthrok (approved):

test/torchaudio_unittest/functional/batch_consistency_test.py
jcaw: Sure thing

jcaw:
I transferred the tensor creation to a batch variant of `get_whitenoise`, but in doing so I found a couple of problems with the suite. `F.vad` and `F.sliding_window_cmn` appear to take spectrograms as input (the shape is `(..., freq, time)`), but this suite is currently testing them with waveforms. Only certain combinations of `n_items` and `n_channels` will fail because of this, so I think that flew under the radar.

Would you like me to open a separate PR converting everything to white noise, and switching these tests to use spectrograms instead? I also parameterized the number of channels in order to isolate the issues, so I can include that too. I could also just add the changes to this PR.

(As a side note, `batch_size=1` turned out to be useful - only `batch_size=3` triggered the issue for the VAD test which did make it easier to diagnose, so I'd say keep it.)

jcaw:
Sorry about that - I got a couple of things wrong in that last comment. I've deleted it while I double check everything.

mthrok:
@jcaw

Thanks for the update. I will go ahead and merge this PR so that you can start working on the other batch consistency test. If you come up with an update on top of this PR, you can add that in the new PR.

Regarding my comment about replacing `randn` with white noise, I think you are right. The inputs are supposed to be spectrogram, so using `get_whitenoise` looks strange and confusing for the other maintainers. I think we can add a new helper function `get_spectrogram`, which generates spectrogram on the given waveform or whitenoise if not provided, using librosa.

PR #1319, opened 2021-02-26
Use c++ backend for lfilter when c++ extension is available

yoyololicon:
related to #1310


torchaudio/functional/filtering.py
```diff
@@ -887,11 +870,43 @@ def lfilter(
 
     output = padded_output_waveform[:, n_order - 1:]
 
-    if clamp:
-        output = torch.clamp(output, min=-1.0, max=1.0)
-
     # unpack batch
     output = output.reshape(shape[:-1] + output.shape[-1:])
```
cpuhrsch, line 874: It seems that this line is missing from the new ```lfilter``` and I don't find it in the C++ code. Did you mean to port it?

yoyololicon: This line can be found at L908, I don't think it's necessary to port it into c++ for autograd support.

cpuhrsch: On L908 I see torch.clamp, but what about the unpacking of the batch, i.e.
```
output = output.reshape(shape[:-1] + output.shape[-1:])
```
. 

```_lfilter_core``` does the reshape before returning (unpacking) but lfilter_core in lfilter.cpp does not.

File: torchaudio/functional/filtering.py [hunk 2]
```diff
@@ -887,11 +870,43 @@ def lfilter(
...
+        clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
+
+    Returns:
+        Tensor: Waveform with dimension of ``(..., time)``.
+    """
+    # pack batch
+    output = _lfilter(waveform, a_coeffs, b_coeffs)
+
+    if clamp:
+        output = torch.clamp(output, min=-1.0, max=1.0)
```
cpuhrsch, line 909: May I suggest we move ```output = output.reshape(shape[:-1] + output.shape[-1:])``` here as well and remove it from _lfilter_core above? That way it'll be applied to both the Python and C++ version. The reason the CI system didn't catch this is because there are no multidimensional tests for lfilter yet.

torchaudio/functional/filtering.py
yoyololicon: `lfilter_core` in cpp did reshaping at the return statement
```c++
  return output.view(shape);
```
I should use `reshape` instead of `view`, sorry for confusion, will fix it soon.

torchaudio/functional/filtering.py
yoyololicon: In that case, I would also suggest we move the packing part (L833 ~ L834 in `filtering.py` and L93 ~ L94 in `lfilter.cpp`) here just before calling `_filter`. In this way the required input and output of `_filter` will always be a matrix with shape `(n_channel, times)`.

cpuhrsch: Let's go with this, if you don't mind.

yoyololicon (commented):
@cpuhrsch changes resolved

cpuhrsch (approved):

cpuhrsch:
Excellent! Thank you very much! Now back to support for backwards :)

PR #1322, opened 2021-02-26
Update AudioMetaData docstring with encoding values

astaff:
Question: I think the docstring is leaking Sphinx tags. Should we remove them?

### DIFF
--- a/torchaudio/backend/common.py
+++ b/torchaudio/backend/common.py
@@ -16,8 +16,8 @@ class AudioMetaData:
             * ``PCM_U``: Unsigned integer linear PCM
             * ``PCM_F``: Floating point linear PCM
             * ``FLAC``: Flac, Free Lossless Audio Codec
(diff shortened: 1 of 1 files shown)


mthrok:
> Question: I think the docstring is leaking Sphinx tags. Should we remove them?

Where do you mean? The artifact looks good.
https://172095-90321822-gh.circle-artifacts.com/0/docs/backend.html

If you mean `[wikipedia]`,  `[xiph.org]` and other, they were residue from my copy-paste so we should remove it. Sorry about that.

astaff:
spoke IRL: (1) revisit sphinx tags concern later. (2) removed links from docstring.

mthrok:
> spoke IRL: (1) revisit sphinx tags concern later. (2) removed links from docstring.

`[wikipedia]` is still there. sorry...

astaff:
Fixed.

mthrok (approved):
looks good. Thanks!

PR #1325, opened 2021-03-01
Add subcategories to functional documentation

mthrok:
Adds subcategory to functional documentation, file `filtering` so that it's easier to see the separation in utilities and feature extractions.

### DIFF
--- a/docs/source/functional.rst
+++ b/docs/source/functional.rst
@@ -186,44 +186,47 @@ treble_biquad
 vad
 ---
 
+:hidden:`Feature Extractions`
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Looks nice from the rendered docs!

Is there a way to expand / collapse all categories at once?

mthrok:
> Is there a way to expand / collapse all categories at once?

I tried removing `:hidden:` role, but it did not help. I am not sure how we can achieve that.
I will merge this as-is for now.

mthrok:
Thanks!

PR #1326, opened 2021-03-01
[Experiment] integrate Kaldifeat

mthrok:


torchaudio/csrc/kaldifeat.cpp
```diff
@@ -0,0 +1,27 @@
...
+namespace kaldifeat {
+
+
+torch::Tensor compute_fbank(
+    const torch::Tensor &wave,
+    int64_t sample_rate) {
+  ::kaldifeat::FbankOptions option;
+  option.frame_opts.samp_freq = static_cast<float>(sample_rate);
+  option.device = wave.device();
+  ::kaldifeat::Fbank fbank(option);
```
csukuangfj, line 14: Is the input wave in the range [-32768, 32768]? 

If not, it is recommended to set `option.frame_opts.dither` to zero or to a very small
value (e.g., 1e-4), whose default is `1`.

PR #1338, opened 2021-03-02
Make sox selective

carolineechen:
support functionality to exclude libsox from C++ extension, so that we can eventually support C++ extension on Windows
- change behavior of `BUILD_SOX=0` to exclude libsox from being built + bound, different from the current implementation (dynamic build if =0)
- construct functions to detect whether or not libsox is built, to appropriately handle functions and unittests requiring sox. previously, this was tested using detection of C++ extension since libsox and extension were always built together, but this will no longer be the case

testing: `pytest torchaudio_unittest` for both `BUILDSOX=0` and `BUILDSOX=1` cases

### DIFF
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -5,13 +5,12 @@ import torch
 from torchaudio._internal import (
     module_utils as _mod_utils,
 )
-from torchaudio.utils.sox_utils import requires_sox
 
 import torchaudio
 from .common import AudioMetaData
 
 
-@requires_sox()
+@_mod_utils.requires_sox()
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def info(
         filepath: str,
@@ -56,7 +55,7 @@ def info(
     return AudioMetaData(*sinfo)
 
 
-@requires_sox()
+@_mod_utils.requires_sox()
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def load(
         filepath: str,
@@ -154,7 +153,7 @@ def load(
         filepath, frame_offset, num_frames, normalize, channels_first, format)
 
 
-@requires_sox()
+@_mod_utils.requires_sox()
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def save(
         filepath: str,
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -10,6 +10,7 @@ from torchaudio.utils.sox_utils import requires_sox
 import torchaudio
 from .common import AudioMetaData
 
+
 @requires_sox()
 @_mod_utils.requires_module('torchaudio._torchaudio')
 def info(
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -5,11 +5,12 @@ import torch
 from torchaudio._internal import (
     module_utils as _mod_utils,
 )
+from torchaudio.utils.sox_utils import requires_sox
 
 import torchaudio
 from .common import AudioMetaData
 
-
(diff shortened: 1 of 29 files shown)


torchaudio/utils/sox_utils.py
```diff
@@ -1,13 +1,27 @@
...
+    if is_sox_available():
+        def decorator(func):
+            return func
+    else:
+        def decorator(func):
+            @wraps(func)
+            def wrapped(*args, **kwargs):
+                raise RuntimeError(f'{func.__module__}.{func.__name__} requires sox')
+            return wrapped
+    return decorator
```
mthrok, line 21: The implementation looks good, but can you put this in `torchaudio.extension` module?

`torchaudio.utils.sox_utils` are intended for configuring the libsox, such as buffer size, while `is_sox_available` is for extension management. so when we make other components selective and add other availability functions (such as for Kaldi and RNN transducer), we can put all the `is_XXX_available` in one place.

mthrok (commented):
Looks good!

torchaudio/utils/sox_utils.py
```diff
@@ -1,13 +1,27 @@
...
+from functools import wraps
 
 import torch
 
-from torchaudio._internal import (
-    module_utils as _mod_utils,
-)
 
+def is_sox_available():
+    return torch.ops.torchaudio.is_sox_available()
```
mthrok, line 8: For windows support at the moment, the logic has to be `is C++ extension available` AND `is built with sox`

mthrok: so it's `is_module_available('torchaudio._torchaudio') and torch.ops.torchaudio.is_sox_available()`

torchaudio/utils/sox_utils.py
mthrok: Probably in `module_utils.py` is good.

torchaudio/backend/sox_io_backend.py
```diff
@@ -10,6 +10,7 @@
 from .common import AudioMetaData
 
 
+@_mod_utils.requires_sox()
 @_mod_utils.requires_module('torchaudio._torchaudio')
```
mthrok, line 14: I think the logic of `requires_sox` contains that of `requires_module('torchaudio._torchaudio')`, so we should be able to replace `'torchaudio._torchaudio'`.

torchaudio/csrc/utils.cpp
```diff
@@ -0,0 +1,17 @@
+#include <torch/script.h>
+
+namespace {
```
mthrok, line 3: can you add `torchaudio` namespace too?

torchaudio/sox_effects/__init__.py
```diff
@@ -1,4 +1,4 @@
-from torchaudio._internal import module_utils as _mod_utils
+from torchaudio._internal.module_utils import is_sox_available
```
mthrok, line 1: The reason why we adopted the import pattern of `from torchaudio._internal import module_utils as _mod_utils` is that the whole module (and functions inside of it) is intended for internal use, so we do not want to expose it in other module.

`from torchaudio._internal.module_utils import is_sox_available` makes the helper function accessible like `torchaudio.XXX.is_sox_available`. Of course this is not a documented function and it should not be used by users, but we want to be careful about it. This is because at the beginning of torchaudio project, many helper functions were defined in `__init__.py` without under score prefix and it was messy/confusing. If we import individual functions from internal module, users cannot tell from the outside of the package that the function resides in the internal module, thus not for users to use.

mthrok: Of course, you can do `from torchaudio._internal.module_utils import is_sox_available as _is_sox_available` too.

mthrok (approved):

PR #1339, opened 2021-03-02
Add AudioMetaData.__str__ override

### LINKED ISSUE #1323: Making `AudioMetaData` print friendly
`AudioMetaData` class reports meta-data of audio source. It is however not print friendly.

```python
print(torchaudio.info(src))
>>> <torchaudio.backend.common.AudioMetaData object at 0x7f1bc5cd2890>
```

It is nice if we can simply print the attributes like `dataclass` objects do.

```python
print(torchaudio.info(src))
>>> AudioMetaData(sample_rate=900, encoding="PCM", ...)
```

## Steps

There are two approaches I can think of
1. Add `__str__` method.
2. Use `dataclasses.dataclass`

For 2, the `info` function has to be TorchScript-compatible. This means that its return type `AudioMetaData` has to be TorchScript-able. For this reason, `dataclass` might not be applicable. This can be checked with the following test;

```bash
(cd test && pytest torchaudio_unittest/backend/sox_io/torchscript_test.py)
```

## Build and test

Please refer to the [contribution guide](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md) for how to setup development environment.

To test, 

```bash
(cd test && pytest torchaudio_unittest/backend/sox_io/torchscript_test.py torchaudio_unittest/backend/sox_io/info_test.py torchaudio_unittest/backend/soundfile_io/info_test.py)
```

majabojarska:
Makes `AudioMetaData` print-friendly.
Closes #1323

### DIFF
--- a/torchaudio/backend/common.py
+++ b/torchaudio/backend/common.py
@@ -38,3 +38,14 @@ class AudioMetaData:
         self.num_channels = num_channels
         self.bits_per_sample = bits_per_sample
         self.encoding = encoding
+
+    def __str__(self):
+        return (
+            f"{self.__class__.__name__}("
+            f"sample_rate={self.sample_rate}, "
+            f"num_frames={self.num_frames}, "
+            f"num_channels={self.num_channels}, "
+            f"bits_per_sample={self.bits_per_sample}, "
+            f"encoding={self.encoding}"
+            f")"
+        )
--- a/torchaudio/backend/common.py
+++ b/torchaudio/backend/common.py
@@ -41,7 +41,7 @@ class AudioMetaData:
 
     def __str__(self):
         return (
-            f"{self.__class__.__name__}("
+            f"AudioMetaData("
             f"sample_rate={self.sample_rate}, "
             f"num_frames={self.num_frames}, "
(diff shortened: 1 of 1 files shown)


torchaudio/backend/common.py
```diff
@@ -38,3 +38,14 @@ def __init__(
         self.num_channels = num_channels
         self.bits_per_sample = bits_per_sample
         self.encoding = encoding
+
+    def __str__(self):
+        return (
+            f"{self.__class__.__name__}("
```
mthrok, line 44: Although this is correct and more generic, TorchScript compiler has restriction in the kind of attributes it can access, and `self.__class__` is not accessible. Can you hard-code it?

mthrok (commented):
@majabojarska 


Thanks opening the PR. Could you try removing the access to `self.__class__` attribute?

torchaudio/backend/common.py
majabojarska: Thanks for the tip. I commited the necessary change.

mthrok:
Thanks!

Issue #1337, opened 2021-03-02
[Migration] Torchaudio Complex Tensor Support and Migration


mthrok:
# Torchaudio Complex Tensor Support and Migration

## Overview

`torchaudio` has been expressing complex numbers by having an extra dimension for real-part and imaginary-part. (We will refer this format as `"pseudo complex type"`)

```python
waveform = torch.randn(...)
spectrogram = torchaudio.functional.spectrogram(waveform, ..., power=None)

# the last dimension represents real and imaginary parts of complex tensors.
print(spectrogram.dtype, spectrogram.shape)
>>> torch.float32, torch.Size([... ,2])
```

PyTorch 1.6 introduced complex Tensor type, such as `torch.complex64` (`torch.cfloat`) and `torch.complex128` (`torch.cdouble`). (Will be refered as `"native complex type"`)

The natitve complex type comes with handy methods for complex operation such as `abs`, `angle` and `magphase`. (Please refer to the official documentation for the detail.)

Over the few coming releases, we plan to migrate torchaudio's functions and transforms to the native complex type. This issues describes the planned approaches/works/changes/timeline. If you have a question, a concern or a suggestion. Feel free to leave a comment.

## Migration Stages

We will perform the migration in multiple stages. At this moment, the completion of the later migration stages re not tied with specific releases yet.

### ✅ Stage 0 (`~ 0.8`)
Up to release 0.8, torchaudio exclusively used pseudo complex type. In PyTorch 1.7, PyTorch started the adaptation of native complex type and the migration of `torch.fft` namespace. Because of this, torchaudio already uses native complex type in some implementations (`F.vad`, `T.Vad`, `kaldi.spectrogram` and `kaldi.fbank`) but all the user facing APIs use use pseudo complex type.

### ✅ Stage 1 Add support for native complex type and deprecate pseudo complex type)

**Completed:** PyTorch 1.9 / torchaudio 0.9

#### Library code change
In this stage, torchaudio will support both pseudo complex type and native complex type. This means that

* Functions that accept complex input should be able to handle both pseudo and native complex types.
* Functions that return complex values can return both pseudo and native complex types.
    * For real-to-complex functions, a new argument `return_complex` will be added so that users can switch the behavior.

#### Test code update

In addition to the above library code changes, we are going to add a set of tests to make sure that native complex types work in common use cases. This includes;

* Gradient check
* JIT
* Performance 
* Distributed training
* `nn.Module` compatibility

[...]

PR #1340, opened 2021-03-02
Add autograd test for T.Spectrogram

mthrok:
Add test for checking auto grad. Part of #1337


test/torchaudio_unittest/autograd_test_impl.py
```diff
@@ -0,0 +1,38 @@
...
+
+class AutogradTestCase(TestBaseMixin):
+    def assert_grad(self, transform, *inputs, eps=1e-06, atol=1e-05, rtol=0.001):
+        transform = transform.to(self.device)
+
+        inputs_ = []
+        for i in inputs:
+            i.requires_grad = True
+            inputs_.append(i.to(dtype=self.dtype, device=self.device))
+        assert gradcheck(transform, inputs_, eps=eps, atol=atol, rtol=rtol)
```
anjali411, line 19: you probably also wanna run `gradgradcheck` for testing second order derivatives

test/torchaudio_unittest/autograd_cuda_test.py
```diff
@@ -0,0 +1,12 @@
+import torch
+from torchaudio_unittest.common_utils import (
+    PytorchTestCase,
+    skipIfNoCuda,
+)
+from torchaudio_unittest.autograd_test_impl import AutogradTestCase
+
+
+@skipIfNoCuda
+class AutogradCUDATest(AutogradTestCase, PytorchTestCase):
```
anjali411, line 10: I think it would be nice to have a device generic class similar to what we have in PyTorch for autograd tests https://github.com/pytorch/pytorch/blob/a3a2150409472fe6fa66f3abe9e795303786252c/test/test_autograd.py#L7931-L7937

mthrok: Sure, but let's differ on changing the test infrastructure. This is the common pattern used in torchaudio, so if we are going to change, I would like to change them in one go.

mthrok:
@anjali411 I have also added `T.MelSpectrogram`. let me know what you think.

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+class AutogradTestCase(TestBaseMixin):
+    def assert_grad(self, transform, *inputs, eps=1e-06, atol=1e-05, rtol=0.001):
+        transform = transform.to(self.device, self.dtype)
+
+        inputs_ = []
+        for i in inputs:
+            i.requires_grad = True
+            inputs_.append(i.to(dtype=self.dtype, device=self.device))
+        assert gradcheck(transform, inputs_, eps=eps, atol=atol, rtol=rtol)
+        assert gradgradcheck(transform, inputs_, eps=eps, atol=atol, rtol=rtol)
```
anjali411, line 20: you might also wanna test the non contiguous inputs here (`gen_non_contig_grad_outputs=True`)

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+
+class AutogradTestCase(TestBaseMixin):
+    def assert_grad(self, transform, *inputs, eps=1e-06, atol=1e-05, rtol=0.001):
+        transform = transform.to(self.device, self.dtype)
+
+        inputs_ = []
+        for i in inputs:
+            i.requires_grad = True
+            inputs_.append(i.to(dtype=self.dtype, device=self.device))
+        assert gradcheck(transform, inputs_, eps=eps, atol=atol, rtol=rtol)
```
anjali411, line 19: `self.assertTrue(gradcheck(transform, inputs_))`

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+import torchaudio.transforms as T
+
+from torchaudio_unittest.common_utils import (
+    TestBaseMixin,
+    get_whitenoise,
+)
+
+
+class AutogradTestCase(TestBaseMixin):
+    def assert_grad(self, transform, *inputs, eps=1e-06, atol=1e-05, rtol=0.001):
```
anjali411, line 12: you don't need to define these values `eps, atol, rtol` here

test/torchaudio_unittest/transforms/autograd_test_impl.py
anjali411, line 20: `self.assertTrue(gradgradcheck(transform, inputs_))

anjali411: synced with @albanD offline: this is not needed

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+        ({'pad': 0, 'normalized': True, 'power': 1.0}, ),
+        ({'pad': 3, 'normalized': True, 'power': 1.0}, ),
+        ({'pad': 0, 'normalized': False, 'power': 2.0}, ),
+        ({'pad': 3, 'normalized': False, 'power': 2.0}, ),
+        ({'pad': 0, 'normalized': True, 'power': 2.0}, ),
+        ({'pad': 3, 'normalized': True, 'power': 2.0}, ),
+    ])
+    def test_spectrogram(self, kwargs):
+        transform = T.Spectrogram(**kwargs)
+        waveform = get_whitenoise(sample_rate=8000, duration=0.05, n_channels=2)
```
albanD, line 38: I think you're missing the device/dtype args here no?
That might be simpler than moving stuff around in the assert_grad function.

test/torchaudio_unittest/transforms/autograd_test_impl.py
albanD, line 12: These default values don't seem to be used here, so maybe you can remove them?

test/torchaudio_unittest/transforms/autograd_test_impl.py
albanD, line 19: nit: use self.assertTrue()

test/torchaudio_unittest/transforms/autograd_test_impl.py
albanD: I don't think it is needed here.

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class AutogradTestCase(TestBaseMixin):
+    def assert_grad(self, transform, *inputs, eps=1e-06, atol=1e-05, rtol=0.001):
+        transform = transform.to(self.device, self.dtype)
+
+        inputs_ = []
+        for i in inputs:
+            i.requires_grad = True
+            inputs_.append(i.to(dtype=self.dtype, device=self.device))
```
albanD, line 18: nit you might want to assert that the dtype is float64 here because otherwise this test, with default eps/tolerance will fail.

test/torchaudio_unittest/transforms/autograd_test_impl.py
mthrok: Is there a reason that `assertTrue` is preferred?
I always think that the message returned by `assertTrue`, `AssertionError: False is not true` non sense, so I always use `assert`.

test/torchaudio_unittest/transforms/autograd_test_impl.py
mthrok: This is the pattern we use in `torchaudio`, where we define the tests in base class and set the `device/dtype` in separate class.
Also to minimize the chance to write inconsistent test, the conversion to the expected `device/dtype` is abstracted in helper functions.
Plus, since this tests only run on float64, I will stop parametrizing the `dtype`. will update the code soon

test/torchaudio_unittest/transforms/autograd_test_impl.py
albanD: I was mostly suggesting to use the `self.` variant of the assert instead of the default one.
The  reason is that it makes the test framework aware of it in a better way than catching the error from the assert. And it allows it to give better messages when running the whole test suite.

mthrok:
@albanD @anjali411 

What does this mean? How much should we be worried?

```
RuntimeError: Backward is not reentrant, i.e., running backward with same input and grad_output multiple times gives different values, although analytical gradient matches numerical gradient. The tolerance for nondeterminism was 0.0.
```
https://app.circleci.com/pipelines/github/pytorch/audio/5301/workflows/782e2d49-6f99-4e98-a240-b6597557d855/jobs/176159

albanD:
That means that your backward function is not deterministic.
If you expect it to be deterministic, then you might want to double check.
If this is expected, then you can increase the threshold by passing `nondet_tol: float` to gradcheck with the difference that you expect to see there (this defaults to 0.0). And you should document that the backward is not deterministic (at least that's what we do in core here https://pytorch.org/docs/stable/generated/torch.set_deterministic.html#torch.set_deterministic)

mthrok:
> That means that your backward function is not deterministic.
> If you expect it to be deterministic, then you might want to double check.
> If this is expected, then you can increase the threshold by passing `nondet_tol: float` to gradcheck with the difference that you expect to see there (this defaults to 0.0). And you should document that the backward is not deterministic (at least that's what we do in core here https://pytorch.org/docs/stable/generated/torch.set_deterministic.html#torch.set_deterministic)

Thanks for the info. This is an interesting finding. Let me dig into that.

mthrok:
@albanD Following the finding in https://github.com/pytorch/pytorch/issues/54093, I added `nondet_tol=1e-10`. Do you think this is tight enough?

albanD:
Interesting investigation!
In that case yes 1e-10 should be good enough!

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -0,0 +1,62 @@
...
+        # replication_pad1d_backward_cuda is not deteministic and
+        # gives very small (~2.7756e-17) difference.
+        #
+        # See https://github.com/pytorch/pytorch/issues/54093
+        transform = T.Spectrogram(**kwargs)
+        waveform = get_whitenoise(sample_rate=8000, duration=0.05, n_channels=2)
+        self.assert_grad(transform, [waveform], nondet_tol=1e-10)
+
+    def test_melspectrogram(self):
+        # replication_pad1d_backward_cuda is not deteministic and
```
anjali411, line 55: nit - `is nondeterministic`

anjali411 (approved):
LGTM!

mthrok:
Thanks!

PR #1342, opened 2021-03-02
make kaldi selective

carolineechen:
support functionality to compile without kaldi, similar to #1338

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -32,6 +32,7 @@ def _get_build(var):
 
 
 _BUILD_SOX = _get_build("BUILD_SOX")
+_BUILD_KALDI = _get_build("BUILD_KALDI")
 _BUILD_TRANSDUCER = _get_build("BUILD_TRANSDUCER")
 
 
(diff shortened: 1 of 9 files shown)


build_tools/setup_helpers/extension.py
```diff
@@ -32,6 +32,7 @@ def _get_build(var):
 
 
 _BUILD_SOX = _get_build("BUILD_SOX")
+_BUILD_KALDI = _get_build("BUILD_KALDI")
```
mthrok, line 35: Because we currently default to build Kaldi, in case the environment variable is missing, we need to default to `true` for this.

Please update the `_get_build` to allow override of default value.

mthrok (commented):
Looks good. Let's resolve CI issue and it's good merge.

build_tools/setup_helpers/extension.py
```diff
@@ -19,6 +19,9 @@
 
 
 def _get_build(var):
+    if (var == "BUILD_KALDI") and not os.environ.get(var):
+        return True
+
```
mthrok, line 24: Although this works for our purpose, let's keep this function generic. How about this?

```python
def _get_build(var, default_val=False):
    if val not in os.ennviron:
        return default_val
    val = os.environ.get(var, default_val)
    ...
```

carolineechen: good point 👍

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -99,11 +100,4 @@ def skipIfNoModule(module, display_name=None):
     'sox' not in torchaudio.list_audio_backends(), 'Sox backend not available')
 skipIfNoCuda = unittest.skipIf(not torch.cuda.is_available(), reason='CUDA not available')
 skipIfNoSox = unittest.skipIf(not is_sox_available(), reason='Sox not available')
-
-
-def skipIfNoExtension(test_item):
-    if is_module_available('torchaudio._torchaudio'):
-        return test_item
-    if 'TORCHAUDIO_TEST_FAIL_IF_NO_EXTENSION' in os.environ:
```
mthrok, line 107: follow-up (@mthrok): Revive this mechanism with new semantic. 

```
$ grep TORCHAUDIO_TEST_FAIL_IF_NO_EXTENSIO -r .circleci/
.circleci/unittest/linux/scripts/run_test.sh:export TORCHAUDIO_TEST_FAIL_IF_NO_EXTENSION=1
```

PR #1347, opened 2021-03-03
BC-Breaking: Rename `waveform` tensor in `F.sliding_window_cmn`

jcaw:
`F.sliding_window_cmn` appears to take a spectrogram as input, of shape `(..., freq, time)`. However, the name of the argument is `waveform`. I discussed this briefly with @mthrok [here](https://github.com/pytorch/audio/pull/1341#issuecomment-789406548) and [here](https://github.com/pytorch/audio/pull/1315#issuecomment-787478374). This PR changes the name of this argument to `specgram`. Since this is a BC-breaking change, I've opened a separate pull request. 

The output tensor also appears to be a spectrogram of the same shape as the input, so I've renamed that too, and updated the specification in the docstring.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -942,7 +942,7 @@ def detect_pitch_frequency(
 
 
 def sliding_window_cmn(
-    waveform: Tensor,
+    specgram: Tensor,
     cmn_window: int = 600,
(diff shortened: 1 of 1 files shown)


cpuhrsch (approved):

mthrok:
Thanks!

PR #1360, opened 2021-03-05
adding test for lfilter shape

aobo-y:
Add unit tests to verify the shape of output tensors for `lfilter` for variety of input Tensor shapes.


mthrok (approved):
Looks good. Thanks!

PR #1368, opened 2021-03-06
Adopt native complex dtype in griffnlim

mthrok:
Get rid of pseudo complex type in `F.griffinlim`.

Part of #1337 

`griffinlim` is tested
 - ✅ TorchScript on CPU/CUDA for `float32` and `float64`
 - ✅ `librosa` consistency on CPU
 - ✅ Batch consistency on CPU

**TODO**
 - [x] Autograd test (updated from #1421)
 - [x] Benchmark
 - [x] Remove the hardcoded `cfloat` in the intermediate variable.

## Benchmark

```bash
#!/usr/bin/env bash

OMP_NUM_THREADS=1 numactl --membind 0 --cpubind 0 python -m timeit -s """
import torch
import torchaudio;

batch_size = 32
num_channels = 1
num_freq = 201
num_frames = 400

torch.manual_seed(0);
spec = torch.randn(batch_size, num_channels, num_freq, num_frames, dtype=torch.float32, device='cpu');
griffinlim = torchaudio.transforms.GriffinLim()
""" """
griffinlim(spec)
"""
```

</details>

| device |           master (8d2eeb1)           |                This PR                |
|:------:|:------------------------------------:|:-------------------------------------:|
|    CPU | 1 loop, best of 5: 6.48 sec per loop | 1 loop, best of 5: 3.57 sec per loop  |
|    GPU | 1 loop, best of 5: 128 msec per loop | 1 loop, best of 5: 93.4 msec per loop |

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -8,8 +8,21 @@ import torchaudio.transforms as T
 from torchaudio_unittest.common_utils import (
     TestBaseMixin,
     get_whitenoise,
+    get_spectrogram,
     nested_params,
 )
+
+
+class _DeterministicWrapper(torch.nn.Module):
+    """Helper transform wrapper to make the given transform deterministic"""
+    def __init__(self, transform, seed=0):
+        super().__init__()
+        self.seed = seed
+        self.transform = transform
+
+    def forward(self, input: torch.Tensor):
+        torch.random.manual_seed(0)
+        return self.transform(input)
 
 
 class AutogradTestMixin(TestBaseMixin):
@@ -72,10 +85,13 @@ class AutogradTestMixin(TestBaseMixin):
     )
     def test_griffinlim(self, momentum, rand_init):
         n_fft = 400
-        n_frames = 5
+        power = 1
         n_iter = 3
-        spec = torch.rand(n_fft // 2 + 1, n_frames) * n_fft
-        transform = T.GriffinLim(n_fft=n_fft, n_iter=n_iter, momentum=momentum, rand_init=rand_init)
+        spec = get_spectrogram(
+            get_whitenoise(sample_rate=8000, duration=0.05, n_channels=2),
+            n_fft=n_fft, power=power)
+        transform = _DeterministicWrapper(
+            T.GriffinLim(n_fft=n_fft, n_iter=n_iter, momentum=momentum, rand_init=rand_init, power=power))
         self.assert_grad(transform, [spec])
 
     @parameterized.expand([(False, ), (True, )])
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -21,7 +21,7 @@ class _DeterministicWrapper(torch.nn.Module):
         self.transform = transform
 
     def forward(self, input: torch.Tensor):
-        torch.random.manual_seed(0)
+        torch.random.manual_seed(self.seed)
         return self.transform(input)
 
 
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -8,6 +8,7 @@ import torchaudio.transforms as T
 from torchaudio_unittest.common_utils import (
     TestBaseMixin,
     get_whitenoise,
+    nested_params,
 )
 
 
@@ -65,14 +66,17 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
 
-    @parameterized.expand([(0, ), (0.99, )])
-    def test_griffinlim(self, momentum):
+    @nested_params(
+        [0, 0.99],
+        [False, True],
+    )
+    def test_griffinlim(self, momentum, rand_init):
         n_fft = 400
         n_frames = 5
         n_iter = 3
         spec = torch.rand(n_fft // 2 + 1, n_frames) * n_fft
-        transform = T.GriffinLim(n_fft=n_fft, n_iter=n_iter, momentum=momentum, rand_init=False)
-        self.assert_grad(transform, [spec], nondet_tol=1e-10)
+        transform = T.GriffinLim(n_fft=n_fft, n_iter=n_iter, momentum=momentum, rand_init=rand_init)
+        self.assert_grad(transform, [spec])
 
     @parameterized.expand([(False, ), (True, )])
     def test_mfcc(self, log_mels):
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -125,6 +125,14 @@ def spectrogram(
     return spec_f
 
 
+def _get_complex_dtype(real_dtype: torch.dtype):
+    if real_dtype == torch.double:
+        return torch.cdouble
+    if real_dtype == torch.float:
+        return torch.cfloat
+    raise ValueError(f'Unexpected dtype {real_dtype}')
+
+
 def griffinlim(
         specgram: Tensor,
         window: Tensor,
@@ -180,23 +188,19 @@ def griffinlim(
 
     specgram = specgram.pow(1 / power)
 
-    # randomly initialize the phase
-    batch, freq, frames = specgram.size()
+    # initialize the phase
     if rand_init:
-        angles = 2 * math.pi * torch.rand(batch, freq, frames)
+        angles = torch.rand(
+            specgram.size(),
+            dtype=_get_complex_dtype(specgram.dtype), device=specgram.device)
     else:
-        angles = torch.zeros(batch, freq, frames)
-    angles = torch.stack([angles.cos(), angles.sin()], dim=-1) \
-        .to(dtype=specgram.dtype, device=specgram.device)
-    specgram = specgram.unsqueeze(-1).expand_as(angles)
+        angles = torch.full(
+            specgram.size(), 1,
+            dtype=_get_complex_dtype(specgram.dtype), device=specgram.device)
 
     # And initialize the previous iterate to 0
-    rebuilt = torch.tensor(0.)
-
+    tprev = torch.tensor(0., dtype=specgram.dtype, device=specgram.device)
     for _ in range(n_iter):
-        # Store the previous iterate
-        tprev = rebuilt
-
         # Invert with our current estimate of the phases
         inverse = torch.istft(specgram * angles,
                               n_fft=n_fft,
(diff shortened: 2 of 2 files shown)


torchaudio/functional/functional.py
```diff
@@ -177,12 +177,10 @@ def griffinlim(
...
     if rand_init:
-        angles = 2 * math.pi * torch.rand(batch, freq, frames)
+        angles = torch.rand(batch, freq, frames, dtype=torch.cfloat, device=specgram.device)
     else:
-        angles = torch.zeros(batch, freq, frames)
-    angles = torch.stack([angles.cos(), angles.sin()], dim=-1) \
-        .to(dtype=specgram.dtype, device=specgram.device)
-    specgram = specgram.unsqueeze(-1).expand_as(angles)
+        angles = torch.zeros(batch, freq, frames, dtype=torch.cfloat, device=specgram.device)
+        angles.real[:] = 1
```
mthrok, line 183: @anjali411 Do you know an elegant and TorchScript-compatible way to initialize Tensors with value `1+0i`?

anjali411: After #52881 is merged, `torch.full(batch, freq, frames, 1j, dtype=torch.cfloat, device=specgram.device)` should work

mthrok: it worked locally. pushing the change.

torchaudio/functional/functional.py
```diff
@@ -177,12 +177,9 @@ def griffinlim(
     # randomly initialize the phase
     batch, freq, frames = specgram.size()
     if rand_init:
-        angles = 2 * math.pi * torch.rand(batch, freq, frames)
+        angles = torch.rand(batch, freq, frames, dtype=torch.cfloat, device=specgram.device)
```
mthrok, line 180: TODO: check to see if `cfloat` is preferred.

anjali411: yeah or even take a step back and consider why we were creating float tensors by default and if it makes sense to  add a dtype `kwarg` that defaults to float

anjali411 (commented):
needs JIT and autograd tests to verify if everything works as expected

torchaudio/functional/functional.py
mthrok: @anjali411

I looked into this and the followings are my finding;

1. The use of `cfloat` here is fine when the input Tensor is `float32` and `float64`.
This is because the dtype of `angles` are updated to the correct one in the next `for` loop. Specifically, `specgram * angles` there produces the same dtype as `specgram` so after the first iteration, all the Tensor has the matching dtype as the input.
2. The above suggests that the `griffinlim` will not work for `float16` input type. But I did test and `pow` is not implemented for `float16`. (see the detail bellow) so I do not think it's and issue.

With this finding, I suggest to proceed with the current approach (using `cfloat` for intermediate variables) since it is working for `float32` and `float64`. Also note that `float16` is not in general tested in `torchaudio`.

---

The following is the result of running `griffinlim` for the different dtype on before/after this PR.

### Script
<Details><summary>script</summary>

```python
import torch
import torchaudio.transforms as T


def test(waveform, dtype):
    n_fft = 1024
    win_length = None
    hop_length = 512

    spectrogram = T.Spectrogram(
        n_fft=n_fft,
        win_length=win_length,
        hop_length=hop_length,
    )
    griffin_lim = T.GriffinLim(
        n_fft=n_fft,
        win_length=win_length,
        hop_length=hop_length,
    )

    spec = spectrogram(waveform).to(dtype)
    print('*******')
    print('input:', spec.dtype)
    recon = griffin_lim(spec)
    print('output:', recon.dtype)
    assert recon.dtype == spec.dtype
    return recon


torch.random.manual_seed(0)
waveform = torch.randn(2, 1024)


test(waveform, torch.float64)
test(waveform, torch.float32)
test(waveform, torch.float16)
```

<Details>

### Result
#### Before applying this PR (commit 0433b7aa)

```
*******
input: torch.float64
/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/functional.py:655: UserWarning: istft will require a complex-valued input tensor in a future PyTorch release. Matching the output from stft with return_complex=True.  (Triggered internally at  /opt/conda/conda-bld/pytorch_1617001512472/work/aten/src/ATen/native/SpectralOps.cpp:807.)
  return _VF.istft(input, n_fft, hop_length, win_length, window, center,  # type: ignore
output: torch.float64
*******
input: torch.float32
output: torch.float32
*******
input: torch.float16
  File "foo.py", line 36, in <module>
    test(waveform, torch.float16)
  File "foo.py", line 24, in test
    recon = griffin_lim(spec)
  File "/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1015, in _call_impl
    return forward_call(*input, **kwargs)
  File "/scratch/moto/torchaudio/torchaudio/transforms.py", line 189, in forward
    return F.griffinlim(specgram, self.window, self.n_fft, self.hop_length, self.win_length, self.power,
  File "/scratch/moto/torchaudio/torchaudio/functional/functional.py", line 168, in griffinlim
    specgram = specgram.pow(1 / power)
RuntimeError: "pow" not implemented for 'Half'
```

#### After applying this PR

```
*******
input: torch.float64
output: torch.float64
*******
input: torch.float32
output: torch.float32
*******
input: torch.float16
  File "foo.py", line 36, in <module>
    test(waveform, torch.float16)
  File "foo.py", line 24, in test
    recon = griffin_lim(spec)
  File "/home/moto/conda/envs/PY3.8-cuda101/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1015, in _call_impl
    return forward_call(*input, **kwargs)
  File "/scratch/moto/torchaudio/torchaudio/transforms.py", line 189, in forward
    return F.griffinlim(specgram, self.window, self.n_fft, self.hop_length, self.win_length, self.power,
  File "/scratch/moto/torchaudio/torchaudio/functional/functional.py", line 168, in griffinlim
    specgram = specgram.pow(1 / power)
RuntimeError: "pow" not implemented for 'Half'
```

anjali411: This PR https://github.com/pytorch/pytorch/pull/50999 recently added `pow` for `torch.float16`

anjali411: > The use of cfloat here is fine when the input Tensor is float32 and float64.
This is because the dtype of angles are updated to the correct one in the next for loop. Specifically, specgram * angles there produces the same dtype as specgram so after the first iteration, all the Tensor has the matching dtype as the input.

So in that case, I think it makes sense to instead create angles of the same dtype as `specgram` since multiplication in the for loop (`specgram` * `angles`) [promotes all input tensors to the same dtype](https://github.com/pytorch/pytorch/blob/a102adb55e00ea3bc567f6a700d19895f99e148a/aten/src/ATen/TensorIterator.h#L525-L531), so I think it makes sense to create `angles` with dtype `specgram.dtype`.

anjali411:
@mthrok this PR needs performance benchmarking!

torchaudio/functional/functional.py
mthrok: oh this could be the reason why autograd test in #1421 is failing.

mthrok: > This PR [pytorch/pytorch#50999](https://github.com/pytorch/pytorch/pull/50999) recently added `pow` for `torch.float16`

Looks like it's still landing? https://github.com/pytorch/pytorch/pull/55280



> So in that case, I think it makes sense to instead create angles of the same dtype as `specgram` since multiplication in the for loop (`specgram` * `angles`) [promotes all input tensors to the same dtype](https://github.com/pytorch/pytorch/blob/a102adb55e00ea3bc567f6a700d19895f99e148a/aten/src/ATen/TensorIterator.h#L525-L531), so I think it makes sense to create `angles` with dtype `specgram.dtype`.

Added the necessary change.

anjali411 (approved):
LGTM! Thanks Moto. let's merge this after #1421 is merged

torchaudio/functional/functional.py
```diff
@@ -180,52 +188,49 @@ def griffinlim(
...
     specgram = specgram.pow(1 / power)
 
-    # randomly initialize the phase
-    batch, freq, frames = specgram.size()
+    # initialize the phase
     if rand_init:
-        angles = 2 * math.pi * torch.rand(batch, freq, frames)
+        angles = torch.rand(
+            specgram.size(),
+            dtype=_get_complex_dtype(specgram.dtype), device=specgram.device)
```
anjali411, line 195: wait the angles here should be a floating point tensor, no?

mthrok: No, `angles` is complex value. `Griffin-Lim` algorithm iteratively optimizes the phase (or the direction in complex plain) of each element of the given spectrogram so that at the end `istft` give the original waveform.

mthrok: *The original implementation was constructing the complex value Tensor called `angles` from real valued tensor called `angle` and `magnitude`.

torchaudio/functional/functional.py
imaginary-person: Sorry, #50999 would hopefully reland soon via #55280, and it'd enable `pow` for `float16` on CPU.

`pow` for `float16` is already supported on CUDA.

mthrok: @imaginary-person Thanks for the info.

mthrok:
@anjali411 

The autograd tests are failing for the case `rand_init=True`. However, it does not fail on my local env with PyTorch nightly from March 29th.

```
[conda] pytorch                   1.9.0.dev20210329 py3.8_cuda10.1_cudnn7.6.3_0    pytorch-nightly
```

Failing test log: https://app.circleci.com/pipelines/github/pytorch/audio/5665/workflows/746a53f5-6f18-4f2e-b717-869bf342eaf4/jobs/194528

I am wondering if this is a regression. What do you think? The following is the corresponding code;

https://github.com/mthrok/audio/blob/17f61e73f8f4b0f314bb7356aa3a2222c4038347/torchaudio/functional/functional.py#L193-L201

yoyololicon:
> @anjali411
> 
> The autograd tests are failing for the case `rand_init=True`.

@mthrok 

Hmm, I think it's an expected behavior, cuz numerical method will (I'm not sure though) forward the function multiple times, and with `rand_init=True`, each forward pass will start with different initial phases, result in different output.

For functions like it which have random behavior internally, maybe a `.backward()` at the end of test is enough, to make sure that gradients can be propogated without problems.

mthrok:
> > @anjali411
> > The autograd tests are failing for the case `rand_init=True`.
> 
> @mthrok
> 
> Hmm, I think it's an expected behavior, cuz numerical method will (I'm not sure though) forward the function multiple times, and with `rand_init=True`, each forward pass will start with different initial phases, result in different output.
> 
> For functions like it which have random behavior internally, maybe a `.backward()` at the end of test is enough, to make sure that gradients can be propogated without problems.

Hi @yoyololicon 

Thanks for the insight. You are right. I added a wrapper that sets the random seed and the test passes now.
(I do not know why I thought it was working on my local env...)

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -8,9 +8,23 @@
...
 
+class _DeterministicWrapper(torch.nn.Module):
+    """Helper transform wrapper to make the given transform deterministic"""
+    def __init__(self, transform, seed=0):
+        super().__init__()
+        self.seed = seed
+        self.transform = transform
+
+    def forward(self, input: torch.Tensor):
+        torch.random.manual_seed(0)
```
anjali411, line 24:
```suggestion
        torch.random.manual_seed(self.seed)
```

mthrok:
@anjali411 This PR is ready for the final review.

PR #1370, opened 2021-03-07
Replace sox conversion macros with Tensor op

mthrok:
`SOX_SIGNED_16BIT_TO_SAMPLE` and `SOX_SIGNED_32BIT_TO_SAMPLE` uses left shift on signed integers, which is UB.
This PR replaces them (and other sox macros for `uint8` and `float` as well) with Tensor operations.

### DIFF
--- a/torchaudio/csrc/sox/effects_chain.cpp
+++ b/torchaudio/csrc/sox/effects_chain.cpp
@@ -108,7 +108,7 @@ int tensor_input_drain(sox_effect_t* effp, sox_sample_t* obuf, size_t* osamp) {
       throw std::runtime_error("Unexpected dtype.");
   }
   // Write to buffer
-  memcpy(obuf, chunk.data_ptr<int32_t>(), *osamp *4);
+  memcpy(obuf, chunk.data_ptr<int32_t>(), *osamp * 4);
   priv->index += *osamp;
   return (priv->index == num_samples) ? SOX_EOF : SOX_SUCCESS;
 }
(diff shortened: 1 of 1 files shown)


torchaudio/csrc/sox/effects_chain.cpp
```diff
@@ -69,49 +69,46 @@ int tensor_input_drain(sox_effect_t* effp, sox_sample_t* obuf, size_t* osamp) {
...
-      auto ptr = tensor_.data_ptr<float_t>();
-      for (size_t i = 0; i < *osamp; ++i) {
-        obuf[i] = SOX_FLOAT_32BIT_TO_SAMPLE(ptr[i], effp->clips);
-      }
+      // Need to convert to 64-bit precision so that
+      // values around INT32_MIN/MAX are handled correctly.
+      chunk = chunk.to(c10::ScalarType::Double);
+      chunk *= 2147483648.;
+      chunk.clamp_(INT32_MIN, INT32_MAX);
+      chunk = chunk.to(c10::ScalarType::Int);
```
cpuhrsch, line 89: I'm not sure that this will guarantee that chunk is contiguous afterwards. I'd call into "contiguous" anyway to be sure. Same applies for the other cases.

cpuhrsch (approved):

torchaudio/csrc/sox/effects_chain.cpp
mthrok: cool. will update

PR #1345, opened 2021-03-03
enable C++ extension on Windows

carolineechen:
after making sox and kaldi selective (#1338, #1342), we can enable C++ extension to be built on Windows without requiring bindings to either of them

PR #1371, opened 2021-03-08
Fix C++ extension initialization for Windows

mthrok:
The C++ extension for Windows we added in #1345 were not properly loaded. This PR fixes it by
1. Rename the extension to `_torchaudio.pyd` from `_torchaudio.dll`
2. Properly set the installation path to `_torchaudio.pyd` from `lib/_torchaudio.dll`

@carolineechen Can you verify that the test log does not contain the warning `UserWarning: torchaudio C++ extension is not available.` anymore?


carolineechen (approved):
Looks good! Windows logs no longer show the C++ extension UserWarning

PR #1379, opened 2021-03-10
Add TorchScript test for `F.phase_vocoder`

PengfeiWangWZ:
Add TorchScript test for `F.phase_vocoder`

Test plan
1. Go to test folder `cd test`
2. Run `pytest torchaudio_unittest/functional/torchscript_consistency_cpu_test.py -k phase_vocoder`

### DIFF
--- a/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
@@ -1,6 +1,5 @@
 """Test suites for jit-ability and its numerical compatibility"""
 import unittest
-import numpy as np
 
 import torch
 import torchaudio.functional as F
@@ -549,14 +548,15 @@ class Functional(common_utils.TestBaseMixin):
         self._assert_consistency(func, tensor)
 
     def test_phase_vocoder(self):
-        def func(tensor):
+        def func(tensor, device: torch.device = self.device):
             rate = 0.5
             hop_length = 256
             phase_advance = torch.linspace(
                 0,
-                np.pi * hop_length,
+                3.14 * hop_length,
                 tensor.shape[-3],
-                dtype=torch.float64)[..., None]
+                dtype=torch.float64,
+            ).to(device)[..., None]
             return F.phase_vocoder(tensor, rate, phase_advance)
 
         tensor = torch.randn(2, 1025, 400, 2)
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -547,6 +548,20 @@ def func(tensor):
         tensor = common_utils.get_whitenoise(sample_rate=44100)
         self._assert_consistency(func, tensor)
 
+    def test_phase_vocoder(self):
+        def func(tensor):
+            rate = 0.5
+            hop_length = 256
+            phase_advance = torch.linspace(
+                0,
+                np.pi * hop_length,
```
mthrok, line 557: Wow, I did not know that TorchScript support `np.pi` (maybe because it simply is a constant value).
However, we prefer not to depend on the third party package on the test unless that's absolutely necessary. 
Can you replace `np.pi` with something like `3.14` and remove `import numpy as np`?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -547,6 +548,20 @@ def func(tensor):
         tensor = common_utils.get_whitenoise(sample_rate=44100)
         self._assert_consistency(func, tensor)
 
+    def test_phase_vocoder(self):
+        def func(tensor):
+            rate = 0.5
+            hop_length = 256
+            phase_advance = torch.linspace(
```
mthrok, line 555: This test is failing on CUDA device, as `phase_advance` is not on the same device.
Can you change this helper function as the following so that `phase_advance` is moved to the test device?

```python
    def func(tensor, device: torch.device = self.device):
        ...
        def func(tensor, device: torch.device = self.device):
            rate = 0.5
            hop_length = 256
            phase_advance = torch.linspace(
		0,
                3.14 * hop_length,
		tensor.shape[-3],
		dtype=torch.float64,
            ).to(device)[..., None]
            return F.phase_vocoder(tensor, rate, phase_advance)
```

mthrok (commented):
Hi @PengfeiWangWZ 

Thanks for working on this!

mthrok (approved):
Looks good to me. Thanks!

PR #1382, opened 2021-03-10
[Experiment] Add YesNo dataset based on IterDataPipe

mthrok:
cc @VitalyFedyunin @ejguan @glaringlee

This is the initial PoC / study for adopting `IterDataPipe` in `torchaudio`.
`YesNo` dataset is very simple, probably too simple for benefitting from the advantages of `IterDataPipe`, but I wanted to use it as an opportunity to learn how `IterDataPipe` should be used.

### DIFF
--- a/torchaudio/datasets/yesno.py
+++ b/torchaudio/datasets/yesno.py
@@ -4,8 +4,7 @@ from dataclasses import dataclass
 from typing import List, Tuple, Union
 
 from torch import Tensor
-from torch.utils.data import Dataset
-from torch.data.utils import datapipes
+from torch.utils.data import Dataset, IterDataPipe
 
 import torchaudio
 from torchaudio.datasets.utils import (
@@ -94,7 +93,7 @@ class YesNoItem:
     sample_rate: int
 
 
-class ListYesNoItems(datapipes.iter.IterDataPipe):
+class ListYesNoItems(IterDataPipe):
     """Given a root directory, return the list of files"""
     def __init__(self, root):
         self.data_dir = os.path.join(root, 'waves_yesno')
@@ -107,7 +106,7 @@ class ListYesNoItems(datapipes.iter.IterDataPipe):
             yield path, label
 
 
-class LoadYesNoItem:
+class LoadYesNoItem(IterDataPipe):
     def __init__(self, data_pipe):
         self.data_pipe = data_pipe
 
--- a/torchaudio/datasets/yesno.py
+++ b/torchaudio/datasets/yesno.py
@@ -1,14 +1,17 @@
 import os
 from pathlib import Path
+from dataclasses import dataclass
 from typing import List, Tuple, Union
 
 from torch import Tensor
 from torch.utils.data import Dataset
+from torch.data.utils import datapipes
 
 import torchaudio
 from torchaudio.datasets.utils import (
     download_url,
     extract_archive,
+
 )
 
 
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/yesno.py
```diff
@@ -17,10 +19,108 @@
...
+
+
+class LoadYesNoItem(IterDataPipe):
+    def __init__(self, data_pipe):
+        self.data_pipe = data_pipe
+
+    def __iter__(self):
+        for path, label in self.data_pipe:
+            waveform, sample_rate = torchaudio.load(path)
+            yield YesNoItem(path, label, waveform, sample_rate)
```
ejguan, line 116: You don't need to have this datapipe. You can call `.map` over the source DataPipe with a function to decode the data.

ejguan, line 116: You can use `Map` to apply a function to each item in the pipeline.

torchaudio/datasets/yesno.py
```diff
@@ -17,10 +19,108 @@
...
+    """Given a root directory, return the list of files"""
+    def __init__(self, root):
+        self.data_dir = os.path.join(root, 'waves_yesno')
+        self.files = _RELEASE_CONFIGS['release1']['files']
+
+    def __iter__(self):
+        for filename in self.files:
+            path = os.path.join(self.data_dir, filename)
+            label = [int(c) for c in path.split("_")]
+            yield path, label
```
ejguan, line 106: Can you use [`ListDirFiles`](https://github.com/pytorch/pytorch/blob/f364e492dfb287e4043d37fffa1bcef55e2ac5dd/torch/utils/data/datapipes/iter/listdirfiles.py#L5) to generate a filename per iteration.

torchaudio/datasets/yesno.py
```diff
@@ -17,10 +19,108 @@
...
+
+    def __iter__(self):
+        for path, label in self.data_pipe:
+            waveform, sample_rate = torchaudio.load(path)
+            yield YesNoItem(path, label, waveform, sample_rate)
+
+
+def get_yesno_dataset(root_dir: str):  # , download=False, url=_RELEASE_CONFIGS["release1"]["url"]):
+    # TODO: download dataset if necessary
+    return LoadYesNoItem(ListYesNoItems(root_dir))
```
ejguan, line 121: You can use existing `DataPipe`s
```py
from torch.utils.data import datapipes

dp = datapipes.iter. ListDirFiles(root_dir)
dp = datapipes.iter.Map(dp, fn=loading_fn)
return dp
```
Or, you use the functional API (I prefer this way)
```py
return ListDirFiles(root_dir).map(fn=loading_fn)
```
where `loading_fn` converts file to `YesNoItem`.

ejguan (commented):
Hey @mthrok ,
Thank you for trying `IterDataPipe`. Several comments below:
- One goal for `IterDataPipe` is to reduce the memory in the pipeline by using `__iter__` to generate one item per iteration. As an example for users, it's better not to save filenames in the memory as a list and iterate over them.
- If the existing `DataPipe`s in the core repo have already covered the functionalities you want, there is no need to re-implement a new DataPipe.
Please let me know your opinion about it. Feel free to ping me when you have further questions. ☺️ 

:relaxed:
--



:relaxed:
--




:relaxed:
--



:relaxed:
--




:relaxed:
--




:relaxed:
--

PR #1389, opened 2021-03-13
Add support for 24-bit signed LPCM wav in sox_io backend

iseessel:


torchaudio/csrc/sox/utils.cpp
```diff
@@ -142,19 +146,24 @@ caffe2::TypeMeta get_dtype(
 
 caffe2::TypeMeta get_dtype_from_str(const std::string dtype) {
```
iseessel, line 147: I didn't see this function used anywhere, but I decided to change for completeness sake.

iseessel:
@mthrok Let me know what you think!

torchaudio/backend/sox_io_backend.py
```diff
@@ -73,9 +73,12 @@ def load(
 
             * 32-bit floating-point
             * 32-bit signed integer
+            * 24-bit signed integer
             * 16-bit signed integer
             * 8-bit unsigned integer (WAV only)
 
+            Note: 24-bit signed integers are converted to ``int32`` tensors.
```
mthrok, line 80: I think the fact that `int32` will be used as a substitute is better covered by the next block of docstring, so let's not mention it here. (Besides, note section inside of note section would not render well https://184250-90321822-gh.circle-artifacts.com/0/docs/backend.html#load)

torchaudio/backend/sox_io_backend.py
```diff
@@ -92,10 +95,12 @@ def load(
...
-    signed integer and 8-bit unsigned integer (24-bit signed integer is not supported),
-    by providing ``normalize=False``, this function can return integer Tensor, where the samples
-    are expressed within the whole range of the corresponding dtype, that is, ``int32`` tensor
-    for 32-bit signed PCM, ``int16`` for 16-bit signed PCM and ``uint8`` for 8-bit unsigned PCM.
+    signed integer, 24-bit signed integer, and 8-bit unsigned integer, by providing ``normalize=False``,
+    this function can return integer Tensor, where the samples are expressed within the whole range
+    of the corresponding dtype, that is, ``int32`` tensor for 32-bit signed PCM,
+    ``int16`` for 16-bit signed PCM and ``uint8`` for 8-bit unsigned PCM. Since torch does not
+    support ``int24`` dtype, 24-bit signed PCM are converted to ``int32`` tensors,
+    by implicitly performing 8 left-shifts per data chunk.
```
mthrok, line 103: Let's spare the detail of implementation (8 left-shifts), which can be changed anytime even if the chances are small. Just mention it will be converted to `int32` should suffice.

torchaudio/csrc/sox/utils.cpp
mthrok: Oh this is a helper function from removed feature, which we forgot to remove. Can you leave the function as-is, so that this PR contains only changes related to the 24-bit support. That way, if we need to revert the change, it is easier.

mthrok (commented):
Hi @iseessel

Thanks for working on this. It looks good overall.

torchaudio/csrc/sox/utils.cpp
```diff
@@ -118,15 +118,19 @@ caffe2::TypeMeta get_dtype(
...
         return torch::kUInt8;
-      case SOX_ENCODING_SIGN2: // 16-bit or 32-bit PCM WAV
+      case SOX_ENCODING_SIGN2: // 16-bit, 24-bit, or 32-bit PCM WAV
         switch (precision) {
           case 16:
             return torch::kInt16;
+          case 24: // Cast 24-bit to 32-bit. That is each data chunk will
+                   // implicitly perform 8 left-shifts.
+            std::cout << "Warning: Casting 24-bit signed PCM to torch::kInt32";
+            return torch::kInt32;
```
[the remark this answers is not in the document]
iseessel: Would you like me to add warnings to `apply_effects_file` and `apply_effects_fileobj`?

mthrok:
> 1. Would you like me to add warnings to apply_effects_file and apply_effects_fileobj?

If you have time and interest, please do so. The following is the direction I think works.
1. In [`apply_effects_file`](https://github.com/pytorch/audio/blob/2897f366a9a0a599409844d5a4bf8956c382512d/torchaudio/csrc/sox/effects.cpp#L91-L136) and [`apply_effects_fileobj`](https://github.com/pytorch/audio/blob/2897f366a9a0a599409844d5a4bf8956c382512d/torchaudio/csrc/sox/effects.cpp#L161-L237), change the logic of `get_dtype` to consider `normalize` option as well. (return `float32` if `normalize=True`). 
2. Add warning if `normalize=False` and the input WAV is 24-bit. We should be able to use [`TORCH_WARN_ONCE`](https://github.com/pytorch/pytorch/search?q=TORCH_WARN_ONCE)
3. Remove `normalize` option from [`convert_to_tensor`](https://github.com/pytorch/audio/blob/2897f366a9a0a599409844d5a4bf8956c382512d/torchaudio/csrc/sox/utils.cpp#L161-L201).

> 2. Should I put in a follow up PR to remove the dead code?

If you have time, please do so. I have filed a task T86709147 as well.

mthrok:
Thanks!

Issue #1240, opened 2021-02-04
RNN Transducer Loss


vincentqb:
This issue is to track the follow-up work to #1171, which introduced `rnnt_loss` and `RNNTLoss` as a [prototype](https://pytorch.org/audio/stable/index.html) in `torchaudio.prototype.transducer` using [HawkAaron's warp-transducer](https://github.com/HawkAaron/warp-transducer).

- Update documentation
  - [ ] Guard documentation (e.g. with [conditional doc build](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/conditions.html))
  - [x] Write e.g. [torchtext](https://pytorch.org/text/master/experimental_datasets.html) (#1171)
- Extend guards for prototype
  - [ ] Guard prototype python files by omitting them from torchaudio, see also https://github.com/pytorch/audio/pull/1137#discussion_r551496192
  - [x] Guard building third party transducer even if not added as an extension (#1159)
  - [ ] Enable building transducer in nightlies only, disable for release.
- Update building process
  - [ ] Pass along the `DEBUG` flag to cmake
  - [x] Remove hardcoded O2/O3 optimization, see https://github.com/pytorch/audio/pull/1137#discussion_r551498022 (#1159)
  - [x] Build within same folders as libsox, https://github.com/pytorch/audio/pull/1137#discussion_r551499829 and https://github.com/pytorch/audio/pull/1137#discussion_r551556305 (#1159)
  - [x] Move libsox to a third_party subfolder as suggested in https://github.com/pytorch/audio/pull/1137#discussion_r550321378 (#1161).
  - [ ] Add GPU implementation and compilation.
- Modernization
  - [ ] Migrate the [checks](https://github.com/pytorch/audio/pull/1137/files#diff-9623f973feeae59fa7215b73ad28d00f83c59bfa00f0743c096e4ab36fe5e25dR133) to C++.
  - [ ] Add Torchscriptability test ([internal](https://fb.workplace.com/groups/329222650990087/permalink/830958927483121/)).
  - [ ] Investigate using [`AT_DISPATCH_FLOATING_TYPES`](https://github.com/pytorch/pytorch/blob/22a34bcf4e5eaa348f0117c414c3dd760ec64b13/aten/src/ATen/native/cpu/Activation.cpp#L61).
  - [x] Update bindings to remove pytorch deprecation warnings. (#1160)
[...]

PR #1391, opened 2021-03-15
Migrate transducer input checks to C++

carolineechen:
Move input checks for RNNT from Python to C++

RNN Transducer Loss Issue: #1240

### DIFF
--- a/torchaudio/csrc/transducer.cpp
+++ b/torchaudio/csrc/transducer.cpp
@@ -23,20 +23,22 @@ int64_t cpu_rnnt_loss(
       "label_lengths must be int32 type");
   TORCH_CHECK(
       input_lengths.dtype() == torch::kInt32, "lengths must be int32 type");
-  TORCH_CHECK(acts.is_contiguous(), "log_probs must be contiguous");
+  TORCH_CHECK(acts.is_contiguous(), "acts must be contiguous");
   TORCH_CHECK(labels.is_contiguous(), "labels must be contiguous");
   TORCH_CHECK(
       label_lengths.is_contiguous(), "label_lengths must be contiguous");
   TORCH_CHECK(input_lengths.is_contiguous(), "lengths must be contiguous");
   TORCH_CHECK(
-      input_lengths.size(0) == acts.size(0), "must have a length per example.");
+      input_lengths.size(0) == acts.size(0),
+      "batch dimension mismatch between acts and input_lengths: each example must have a length");
   TORCH_CHECK(
       label_lengths.size(0) == acts.size(0),
-      "must have a label length per example.");
-  TORCH_CHECK(acts.dim() == 4, "log_probs must be 4D");
-  TORCH_CHECK(labels.dim() == 2, "labels must be 2D");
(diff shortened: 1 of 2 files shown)


torchaudio/csrc/transducer.cpp
```diff
@@ -17,11 +17,38 @@ int64_t cpu_rnnt_loss(
...
+  TORCH_CHECK(labels.is_contiguous(), "labels must be contiguous");
+  TORCH_CHECK(
+      label_lengths.is_contiguous(), "label_lengths must be contiguous");
+  TORCH_CHECK(input_lengths.is_contiguous(), "lengths must be contiguous");
+  TORCH_CHECK(
+      input_lengths.size(0) == acts.size(0), "must have a length per example.");
+  TORCH_CHECK(
+      label_lengths.size(0) == acts.size(0),
+      "must have a label length per example.");
+  TORCH_CHECK(acts.dim() == 4, "log_probs must be 4D");
```
mthrok, line 36: I am aware that these error messages are carried out from the original code, but I do not think `must be 4D` is user friendly. If the function is expecting the fixed shape, it should communicate that.  `4D (foo, bar, baz, boo)`.

torchaudio/csrc/transducer.cpp
```diff
@@ -17,11 +17,38 @@ int64_t cpu_rnnt_loss(
...
+      "label_lengths must be int32 type");
+  TORCH_CHECK(
+      input_lengths.dtype() == torch::kInt32, "lengths must be int32 type");
+  TORCH_CHECK(acts.is_contiguous(), "log_probs must be contiguous");
+  TORCH_CHECK(labels.is_contiguous(), "labels must be contiguous");
+  TORCH_CHECK(
+      label_lengths.is_contiguous(), "label_lengths must be contiguous");
+  TORCH_CHECK(input_lengths.is_contiguous(), "lengths must be contiguous");
+  TORCH_CHECK(
+      input_lengths.size(0) == acts.size(0), "must have a length per example.");
```
mthrok, line 32: Again, I am aware that these error messages are carried out from the original code, but `"must have a length per example."` is not helpful on what to do. @carolineechen Can you look into the code and rephrase it to a good instruction?

mthrok (approved):

torchaudio/csrc/transducer.cpp
```diff
@@ -17,11 +17,40 @@ int64_t cpu_rnnt_loss(
...
   int maxT = acts.size(1);
   int maxU = acts.size(2);
   int minibatch_size = acts.size(0);
   int alphabet_size = acts.size(3);
 
+  TORCH_CHECK(
+      at::max(input_lengths).item().toInt() == maxT, "input length mismatch");
+  TORCH_CHECK(
+      at::max(label_lengths).item().toInt() + 1 == maxU,
+      "output length mismatch");
```
vincentqb, line 52: (Follow-up beyond this PR: we'll want to improve this message too)

vincentqb (commented):
Thanks for working on this! I've noted a few small changes (and some things we'll think about in separate PRs)

PR #1393, opened 2021-03-15
Fix functional_cpu_test test name generation

carolineechen:
remove use of `name_func` from `functional_cpu_test.py` to unblock torchaudio import


mthrok (approved):

mthrok:
Thanks!

PR #1399, opened 2021-03-18
Add vanilla DeepSpeech model

discort:
- Add Documentation related to module models
- Add documentation to model `DeepSpeech`
- Add Unit-test to `DeepSpeech`
- Add model `DeepSpeech` according to its paper [Deep Speech: Scaling up end-to-end speech recognition](https://arxiv.org/abs/1412.5567)

Relates to https://github.com/pytorch/audio/issues/446

### DIFF
--- a/torchaudio/models/deepspeech.py
+++ b/torchaudio/models/deepspeech.py
@@ -7,17 +7,17 @@ __all__ = ["DeepSpeech"]
 class FullyConnected(nn.Module):
     """
     Args:
-        in_features: Number of input features
-        hidden_size: Internal hidden unit size.
+        n_feature: Number of input features
+        n_hidden: Internal hidden unit size.
     """
 
     def __init__(self,
-                 in_features: int,
-                 hidden_size: int,
+                 n_feature: int,
+                 n_hidden: int,
                  dropout: float,
                  relu_max_clip: int = 20) -> None:
         super(FullyConnected, self).__init__()
-        self.fc = nn.Linear(in_features, hidden_size, bias=True)
+        self.fc = nn.Linear(n_feature, n_hidden, bias=True)
         self.relu_max_clip = relu_max_clip
         self.dropout = dropout
 
@@ -37,32 +37,32 @@ class DeepSpeech(nn.Module):
     <https://arxiv.org/abs/1412.5567> paper.
 
     Args:
-        in_features: Number of input features
-        hidden_size: Internal hidden unit size.
-        num_classes: Number of output classes
+        n_feature: Number of input features
+        n_hidden: Internal hidden unit size.
+        n_class: Number of output classes
     """
 
     def __init__(self,
-                 in_features: int,
-                 hidden_size: int = 2048,
-                 num_classes: int = 40,
+                 n_feature: int,
+                 n_hidden: int = 2048,
+                 n_class: int = 40,
                  dropout: float = 0.0) -> None:
         super(DeepSpeech, self).__init__()
-        self.hidden_size = hidden_size
-        self.fc1 = FullyConnected(in_features, hidden_size, dropout)
-        self.fc2 = FullyConnected(hidden_size, hidden_size, dropout)
-        self.fc3 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.n_hidden = n_hidden
+        self.fc1 = FullyConnected(n_feature, n_hidden, dropout)
+        self.fc2 = FullyConnected(n_hidden, n_hidden, dropout)
+        self.fc3 = FullyConnected(n_hidden, n_hidden, dropout)
         self.bi_rnn = nn.RNN(
-            hidden_size, hidden_size, num_layers=1, nonlinearity='relu', bidirectional=True)
-        self.fc4 = FullyConnected(hidden_size, hidden_size, dropout)
-        self.out = nn.Linear(hidden_size, num_classes)
+            n_hidden, n_hidden, num_layers=1, nonlinearity='relu', bidirectional=True)
+        self.fc4 = FullyConnected(n_hidden, n_hidden, dropout)
+        self.out = nn.Linear(n_hidden, n_class)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """
         Args:
-            x (torch.Tensor): Tensor of dimension (batch_size, num_channels, input_length, num_features).
+            x (torch.Tensor): Tensor of dimension (batch, channel, time, feature).
         Returns:
-            Tensor: Predictor tensor of dimension (batch_size, input_length, number_of_classes).
+            Tensor: Predictor tensor of dimension (batch, time, class).
         """
         # N x C x T x F
         x = self.fc1(x)
@@ -77,14 +77,14 @@ class DeepSpeech(nn.Module):
         # T x N x H
         x, _ = self.bi_rnn(x)
         # The fifth (non-recurrent) layer takes both the forward and backward units as inputs
-        x = x[:, :, :self.hidden_size] + x[:, :, self.hidden_size:]
+        x = x[:, :, :self.n_hidden] + x[:, :, self.n_hidden:]
         # T x N x H
         x = self.fc4(x)
         # T x N x H
         x = self.out(x)
-        # T x N x num_classes
+        # T x N x n_class
         x = x.permute(1, 0, 2)
-        # N x T x num_classes
+        # N x T x n_class
         x = torch.nn.functional.log_softmax(x, dim=2)
-        # T x N x num_classes
+        # N x T x n_class
         return x
--- a/torchaudio/models/deepspeech.py
+++ b/torchaudio/models/deepspeech.py
@@ -18,19 +18,15 @@ class FullyConnected(nn.Module):
                  relu_max_clip: int = 20) -> None:
         super(FullyConnected, self).__init__()
         self.fc = nn.Linear(in_features, hidden_size, bias=True)
-        self.nonlinearity = nn.Sequential(*[
-            nn.ReLU(),
-            nn.Hardtanh(0, relu_max_clip)
-        ])
-        if dropout:
-            self.nonlinearity = nn.Sequential(*[
-                self.nonlinearity,
-                nn.Dropout(dropout)
-            ])
+        self.relu_max_clip = relu_max_clip
+        self.dropout = dropout
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         x = self.fc(x)
-        x = self.nonlinearity(x)
+        x = torch.nn.functional.relu(x)
+        x = torch.nn.functional.hardtanh(x, 0, self.relu_max_clip)
+        if self.dropout:
+            x = torch.nn.functional.dropout(x, self.dropout, self.training)
         return x
 
 
@@ -58,19 +54,15 @@ class DeepSpeech(nn.Module):
         self.fc3 = FullyConnected(hidden_size, hidden_size, dropout)
         self.bi_rnn = nn.RNN(
             hidden_size, hidden_size, num_layers=1, nonlinearity='relu', bidirectional=True)
-        self.nonlinearity = nn.ReLU()
         self.fc4 = FullyConnected(hidden_size, hidden_size, dropout)
-        self.out = nn.Sequential(*[
-            nn.Linear(hidden_size, num_classes),
-            nn.LogSoftmax(dim=2)
-        ])
+        self.out = nn.Linear(hidden_size, num_classes)
 
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """
         Args:
             x (torch.Tensor): Tensor of dimension (batch_size, num_channels, input_length, num_features).
         Returns:
-            Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
+            Tensor: Predictor tensor of dimension (batch_size, input_length, number_of_classes).
         """
         # N x C x T x F
         x = self.fc1(x)
@@ -90,5 +82,9 @@ class DeepSpeech(nn.Module):
         x = self.fc4(x)
         # T x N x H
         x = self.out(x)
+        # T x N x num_classes
+        x = x.permute(1, 0, 2)
+        # N x T x num_classes
+        x = torch.nn.functional.log_softmax(x, dim=2)
         # T x N x num_classes
         return x
--- a/torchaudio/models/deepspeech.py
+++ b/torchaudio/models/deepspeech.py
@@ -1,0 +1,88 @@
+import torch
+import torch.nn as nn
+
+__all__ = ["DeepSpeech"]
+
+
+class FullyConnected(nn.Module):
+    """
+    Args:
+        in_features: Number of input features
+        hidden_size: Internal hidden unit size.
+    """
+
+    def __init__(self,
+                 in_features: int,
+                 hidden_size: int,
+                 dropout: float,
+                 relu_max_clip: int = 20) -> None:
+        super(FullyConnected, self).__init__()
+        self.fc = nn.Linear(in_features, hidden_size, bias=True)
+        self.nonlinearity = nn.Sequential(*[
+            nn.ReLU(),
+            nn.Hardtanh(0, relu_max_clip)
+        ])
+        if dropout:
+            self.nonlinearity = nn.Sequential(*[
+                self.nonlinearity,
+                nn.Dropout(dropout)
+            ])
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.fc(x)
+        x = self.nonlinearity(x)
+        return x
+
+
+class DeepSpeech(nn.Module):
+    """
+    DeepSpeech model architecture from
+    `"Deep Speech: Scaling up end-to-end speech recognition"`
+    <https://arxiv.org/abs/1412.5567> paper.
+
+    Args:
+        in_features: Number of input features
+        hidden_size: Internal hidden unit size.
+        num_classes: Number of output classes
+    """
+
+    def __init__(self,
+                 in_features: int,
+                 hidden_size: int = 2048,
+                 num_classes: int = 40,
+                 dropout: float = 0.0) -> None:
+        super(DeepSpeech, self).__init__()
+        self.hidden_size = hidden_size
+        self.fc1 = FullyConnected(in_features, hidden_size, dropout)
+        self.fc2 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.fc3 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.bi_rnn = nn.RNN(
+            hidden_size, hidden_size, num_layers=1, nonlinearity='relu', bidirectional=True)
+        self.nonlinearity = nn.ReLU()
+        self.fc4 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.out = nn.Sequential(*[
+            nn.Linear(hidden_size, num_classes),
+            nn.LogSoftmax(dim=2)
+        ])
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # N x C x T x F
+        x = self.fc1(x)
+        # N x C x T x H
+        x = self.fc2(x)
+        # N x C x T x H
+        x = self.fc3(x)
+        # N x C x T x H
+        x = x.squeeze(1)
+        # N x T x H
+        x = x.transpose(0, 1)
+        # T x N x H
+        x, _ = self.bi_rnn(x)
+        # The fifth (non-recurrent) layer takes both the forward and backward units as inputs
+        x = x[:, :, :self.hidden_size] + x[:, :, self.hidden_size:]
+        # T x N x H
+        x = self.fc4(x)
+        # T x N x H
+        x = self.out(x)
+        # T x N x num_classes
+        return x
--- a/torchaudio/models/deepspeech.py
+++ b/torchaudio/models/deepspeech.py
@@ -1,10 +1,9 @@
 import torch
-import torch.nn as nn
 
 __all__ = ["DeepSpeech"]
 
 
-class FullyConnected(nn.Module):
+class FullyConnected(torch.nn.Module):
     """
     Args:
         n_feature: Number of input features
@@ -17,7 +16,7 @@ class FullyConnected(nn.Module):
                  dropout: float,
                  relu_max_clip: int = 20) -> None:
         super(FullyConnected, self).__init__()
-        self.fc = nn.Linear(n_feature, n_hidden, bias=True)
+        self.fc = torch.nn.Linear(n_feature, n_hidden, bias=True)
         self.relu_max_clip = relu_max_clip
         self.dropout = dropout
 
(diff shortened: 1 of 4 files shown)


vincentqb:
Thanks for the contribution! :) Before moving forward, we need to determine how to validate the model implementation, especially if the description implies that it should reproduce an existing paper. Here are strategies we have used in the past:
- For wav2letter and wavernn, we have a training pipeline showing convergence.
- In pytorch/vision#3182, we (also) provided pre-trained weights that worked well with the model.
- Did you use an existing pytorch implementation as reference that we could compare against and validate numerically?

Do you have suggestions on how to do so?

discort:
Thanks for response.
@vincentqb 

- I used wav2letter [implementation](https://github.com/pytorch/audio/pull/462) as an example of this PR.
- According to the [paper](https://github.com/pytorch/audio/pull/462), `DeepSpeech` was trained on 7380 hours on audio including private and paid datasets so unfortunately, I'm not able to reproduce exact the same result and provide the weights for this model as you did for `MobileNetV3`.
- I've made a quick search and didn't find existing precise implementations of `DeepSpeech`. I've seen only numerous implementations of `DeepSpeech2` () or `DeepSpeech` with custom modifications (are not mentioned in the paper), for instance [here](https://github.com/MyrtleSoftware/deepspeech/blob/995ce2d69c8da9b513a2b4e5e3f0b3bae5682373/src/deepspeech/networks/deepspeech.py).

Could you please share examples of pipeline as you did for wav2letter and wavernn? I could follow the same process for `DeepSpeech`. Or I could make a collab file and train the model on a simple dataset, let's say `dev-clean` [librispeech](https://www.openslr.org/12), train it on TPU and share the result here.

What do you think?

vincentqb:
> Thanks for response. @vincentqb

And thanks for contributing! :)

> * I've made a quick search and didn't find existing precise implementations of `DeepSpeech`. I've seen only numerous implementations of `DeepSpeech2` () or `DeepSpeech` with custom modifications (are not mentioned in the paper), for instance [here](https://github.com/MyrtleSoftware/deepspeech/blob/995ce2d69c8da9b513a2b4e5e3f0b3bae5682373/src/deepspeech/networks/deepspeech.py).

Ok, thanks for clarifying :) we'll just need to be careful about the phrasing in the description of the model to covey the idea that this might differ from the original implementation.

> Could you please share examples of pipeline as you did for wav2letter and wavernn? I could follow the same process for `DeepSpeech`.

Sure, they are in the torchaudio repo: [wav2letter](https://github.com/pytorch/audio/tree/master/examples/pipeline_wav2letter), [wavernn](https://github.com/pytorch/audio/tree/master/examples/pipeline_wavernn). (I've also updated the comment above with them.)

> Or I could make a collab file and train the model on a simple dataset, let's say `dev-clean` [librispeech](https://www.openslr.org/12), train it on TPU and share the result here.

If the data used for is not freely available to produce the pre-trained weights, then having a colab training pipeline using freely available data would be useful.  The [DeepSpeech2 paper](https://arxiv.org/pdf/1512.02595v1.pdf) quotes 8% WER on LibriSpeech for DS1 in Table 13, but that's using other data too for training. [This discussion](https://discourse.mozilla.org/t/deepspeech-accuracy-data-for-librispeeh/26067/2) gets 21% with only LibriSpeech, which is of course far from 8% mentioned (or [SOTA](https://paperswithcode.com/sota/speech-recognition-on-librispeech-test-clean)). What numbers would you expect to get by doing the experiment you suggest in colab?

Note that the [wav2letter](https://github.com/pytorch/audio/tree/master/examples/pipeline_wav2letter) pipeline reports only the CER and gets 15% CER, and does not have a language model.

discort:
> Ok, thanks for clarifying :) we'll just need to be careful about the phrasing in the description of the model to covey the idea that this might differ from the original implementation.

Proposed model in this PR matches the original DeepSpeech paper.

> Sure, they are in the torchaudio repo: wav2letter, wavernn. (I've also updated the comment above with them.)

Is it ok that I add train pipeline in this PR as you did for wav2letter and wavernn?

> If the data used for is not freely available to produce the pre-trained weights, then having a colab training pipeline using freely available data would be useful. The DeepSpeech2 paper quotes 8% WER on LibriSpeech for DS1 in Table 13, but that's using other data too for training. This discussion gets 21% with only LibriSpeech, which is of course far from 8% mentioned (or SOTA). What numbers would you expect to get by doing the experiment you suggest in colab?

I found that TPU is impractical so far to train models on CTCLoss(https://github.com/pytorch/xla/issues/2399) so I can't train DeepSpeech using `train-clean-100` `train-clean-360` `train-other-500` because of lack of resources.

The [mentioned](https://github.com/MyrtleSoftware/deepspeech/releases/tag/v0.1) DS1 implementation says that it achieves `15.80` WER on `dev-clean` (see results). But it has LSTM instead of RNN (like in original paper) so this MyrtleSoftware/deepspeech has more number of parameters.

In summary, here are thing what I can do:
- add train/val pipeline
- train model just on `train-clean-100` and provide pre-trained weights

@vincentqb 
What do you think about that?

vincentqb:
> > Sure, they are in the torchaudio repo: wav2letter, wavernn. (I've also updated the comment above with them.)
> 
> Is it ok that I add train pipeline in this PR as you did for wav2letter and wavernn?

This would be a good way of showing the convergence :) (by the way, once we are convinced of the convergence, I'm leaving the door open to only merging the model without the pipeline at the end for now to keep the PR small.)

> > If the data used for is not freely available to produce the pre-trained weights, then having a colab training pipeline using freely available data would be useful. The DeepSpeech2 paper quotes 8% WER on LibriSpeech for DS1 in Table 13, but that's using other data too for training. This discussion gets 21% with only LibriSpeech, which is of course far from 8% mentioned (or SOTA). What numbers would you expect to get by doing the experiment you suggest in colab?
> 
> I found that TPU is impractical so far to train models on [CTCLoss](https://github.com/pytorch/xla/issues/2399) so I can't train DeepSpeech using `train-clean-100` `train-clean-360` `train-other-500` because of lack of resources.

Having the pipeline here for train/validation also allows us to validate the pipeline, and potentially run it on our side with the other data you suggest.

> The [mentioned](https://github.com/MyrtleSoftware/deepspeech/releases/tag/v0.1) DS1 implementation says that it achieves `15.80` WER on `dev-clean` (see results). But it has LSTM instead of RNN (like in original paper) so this MyrtleSoftware/deepspeech has more number of parameters.
> 
> In summary, here are thing what I can do:
> 
> * add train/val pipeline
> * train model just on `train-clean-100` and provide pre-trained weights
> 
> @vincentqb
> What do you think about that?

Great! 

(I'm reading: `train-clean-100` for training and `dev-clean` for validation :) )

mthrok:
@discort 

> In summary, here are thing what I can do:
> 
> * add train/val pipeline
> * train model just on `train-clean-100` and provide pre-trained weights

If you can share the training script on early stage, feel free to share I can try it in our training cluster as well.

Also if you are writing a training script from scratch, please refer to [the distributed training utility](https://github.com/pytorch/audio/blob/master/examples/source_separation/utils/dist_utils.py) and [entry point](https://github.com/pytorch/audio/blob/35d68fddc756b0717bfcba6048fc8e2e95efb465/examples/source_separation/train.py) in source separation script. This one achieves a much better code organization and decoupling of training logic and setup.

discort:
Just a quick update.
I'm still training the model and it seems it's converged. The problem is in my weak GPU but I can share the temp log just to show you for a 'proof'. You can find a train script [here](https://github.com/discort/stt_models), it's based on examples from torchaudio.

[train.log](https://github.com/pytorch/audio/files/6287671/train.log)

@mthrok 
cc @vincentqb 

I'll update script for training DeepSpeech with dist.utility as you proposed. But I won't able to check it manually because I have a single GPU.

So next steps are:
- Update deepspeech train script by distributed parallel loader, etc.
- share the full train.log after 10 epochs (?)
- share the weights. (?)

Sounds good for you?

discort:
**UPD**.
trained on `train-clean-100` and `train-clean-360`. Achieved 0.33 WER on `dev-clean`. Did not apply regularization such as weight decay or dropout. 
pre-trained weights [here](https://github.com/discort/stt_models).

In the implementation from Mozilla's [DeepSpeech](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/training/deepspeech_training/util/flags.py#L37) they used 75 epochs, 0.05 dropout and beam decoder.

@vincentqb 
Let me know if you need more information.

vincentqb:
Thanks for sending the script and logs :) Just to clarify: in your log, the training reaches 130.5% WER and validation 33.3% WER, right?

I did a quick run of your script, and get a reduction from about 60 %WER to 33.6 %WER for validation, see [log](https://github.com/pytorch/audio/files/6311552/output.210414.2.txt). The training batch goes from around 100 %WER to 21.9 %WER in 14 epochs. I'll let it run a bit see if the validation continue to go down.

Do you have a version of DeepSpeech with dropout as done by [Mozilla](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/training/deepspeech_training/util/flags.py#L39)?

If you want to use a ctc beam search decoder, I'd suggest using [awni's](https://gist.github.com/awni/56369a90d03953e370f3964c826ed4b0). Given that there is no language model here, this should be the same as doing argmax and removing duplicates and blank. Does Mozilla do something different here?

discort:
> Thanks for sending the script and logs :) Just to clarify: in your log, the training reaches 130.5% WER and validation 33.3% WER, right?

Actually, I calculated WER in training incorrectly, so did not share it intentionally, then I fixed WER and measured WER for validation on `dev-clean` and got 33% WER in 8 epochs.

> Do you have a version of DeepSpeech with dropout as done by Mozilla?

Are you asking about probability coefficients for dropout?

> If you want to use a ctc beam search decoder, I'd suggest using awni's. Given that there is no language model here, this should be the same as doing argmax and removing duplicates and blank. Does Mozilla do something different here?

Thanks for sharing Awni's gist. I'll definitely take a look. To be honest I'm not into details how Mozilla's [implementation](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/native_client/ctcdecode/ctc_beam_search_decoder.cpp) works but at the first look it's based on https://github.com/parlance/ctcdecode so the principle must be the same.


How can I help you to move it forward?

discort:
Regarding the dropout:
In the original [paper](https://arxiv.org/pdf/1412.5567.pdf) they didn't mention what the exact dropout:

> During training we apply a dropout [19] rate between 5% - 10%. We apply dropout in the feedforward layers but not to the recurrent hidden activations.

In the implementation from [Mozilla](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/training/deepspeech_training/util/flags.py#L39-L44) they specify default value `0.05` for dropout just for the first FC layer. I'm not sure what coefficients did they use to get the best WER.

I can tweak the model implementation to accept dropout coefficients in the constructor (if it's needed):
It will be look something like:
```
class DeepSpeech(nn.Module):
    def __init__(self,
                 in_features: int,
                 hidden_size: int,
                 num_classes: int,
                 fc_dropouts: Tuple[float] = None) -> None:
        super(DeepSpeech, self).__init__()
        if fc_dropouts is None:
            fc_dropouts = (0, 0, 0, 0)
```

Anyway, I would like to see your thoughts about that.

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,88 @@
...
+        self.bi_rnn = nn.RNN(
+            hidden_size, hidden_size, num_layers=1, nonlinearity='relu', bidirectional=True)
+        self.nonlinearity = nn.ReLU()
+        self.fc4 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.out = nn.Sequential(*[
+            nn.Linear(hidden_size, num_classes),
+            nn.LogSoftmax(dim=2)
+        ])
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
```
vincentqb, line 68: Can you update the forward pass with the shape of the tensors? You can see the docstrings for the other models [here](https://github.com/pytorch/audio/tree/master/torchaudio/models). 

Also, note that we've been putting batch first (which is compatible with `DataParallel` also) in torchaudio. [If you like the flexibility, we could have a flag `batch_first` default to `True` also (though I know that [rnn](https://pytorch.org/docs/stable/generated/torch.nn.RNN.html) default to `False` unfortunately).]

vincentqb (commented):
> Actually, I calculated WER in training incorrectly, so did not share it intentionally, then I fixed WER and measured WER for validation on `dev-clean` and got 33% WER in 8 epochs.

Is the code and log in your [previous comment](https://github.com/pytorch/audio/pull/1399#issuecomment-816866303) corrected then?

> > Do you have a version of DeepSpeech with dropout as done by Mozilla?
> 
> Are you asking about probability coefficients for dropout?

I was asking for a code to copy paste and run with dropout so I can do a long run with it :) For now, I would not include it in the version for the PR.

When I look at [Mozilla](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/training/deepspeech_training/util/flags.py#L39-L44), I see layer 1 with 0.05 dropout as you mentioned, but also layer 2, 3, and 6 also defaulting to that value of 0.05.

> Thanks for sharing Awni's gist. I'll definitely take a look. To be honest I'm not into details how Mozilla's [implementation](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/native_client/ctcdecode/ctc_beam_search_decoder.cpp) works but at the first look it's based on https://github.com/parlance/ctcdecode so the principle must be the same.
> 
> How can I help you to move it forward?

Thanks for pointing out the decoder that Mozilla uses. We won't include this as part of the PR.

I've had to restart the long run. I'll post the result here.

torchaudio/models/deepspeech.py
discort: > Also, note that we've been putting batch first (which is compatible with DataParallel also) in torchaudio. [If you like the flexibility, we could have a flag batch_first default to True also (though I know that rnn default to False unfortunately).]

Apologies. I did not get this part of comment. Do you mean to put `batch_size` first in the model output?

vincentqb: no worries, let me clarify what I meant there :) 

if you look at [wav2letter](https://github.com/pytorch/audio/blob/master/torchaudio/models/wav2letter.py#L67), [convtasnet](https://github.com/pytorch/audio/blob/master/torchaudio/models/conv_tasnet.py#L297), [wavernn](https://github.com/pytorch/audio/blob/master/torchaudio/models/wavernn.py#L280), all take as input a tensor where the first dimension is a batch. we need to make the model consume a tensor with batch first like the others.

the reference to `batch_first` was made in connection to the interface for [rnn in pytorch](https://pytorch.org/docs/stable/generated/torch.nn.RNN.html#torch.nn.RNN). however, let me change my mind on this: let's not add a `batch_first` parameter to the constructor since no other model have that in torchaudio.

discort: > if you look at wav2letter, convtasnet, wavernn, all take as input a tensor where the first dimension is a batch. we need to make the model consume a tensor with batch first like the others.

ok, it's clear now. I'll add the similar docstrings.

discort:
> Here's a quick update for fun :) but I'll leave the code running longer, see current log.

Thank you so much for sharing it.  It's just a pleasure to see this graphs :) (especially when you can't reproduce them manually).

-----
The best validation WER so far (what I've found in the attached log) is `0.318`.
Just a couple of ideas of how to achieve lower train loss:
- to try a newest optimizer, for instance https://github.com/facebookresearch/madgrad
- orthogonal initialization for RNN

For lower validation WER:
You ran training on `train_data_urls=['train-clean-100', 'train-clean-360']`. 
For instance, in Mozilla's [implementation](https://github.com/mozilla/DeepSpeech/blob/5f566f442541d76f320ee081294cdef980c361d1/bin/import_librivox.py#L27-L37) and in [this](https://github.com/MyrtleSoftware/deepspeech/blob/995ce2d69c8da9b513a2b4e5e3f0b3bae5682373/src/deepspeech/run.py#L169-L171) they ran the training also on `train-other-500`, more training data behaves like a regularizer and should result in lower validation WER.

vincentqb:
I've ran the model also including train-other-500, see [log](https://github.com/pytorch/audio/files/6402103/deepspeech-960.txt). Best validation: 28.3 %WER.

train_loss_log_960
train_wer_960
valid_loss_960
valid_wer_960

We can see that the training continues to overfit, which is great in this context.

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        Args:
+            x (torch.Tensor): Tensor of dimension (batch_size, num_channels, input_length, num_features).
+        Returns:
+            Tensor: Predictor tensor of dimension (input_length, batch_size, number_of_classes).
```
vincentqb, line 73: Thanks for updating the docstring :) as mentioned in [comment above](https://github.com/pytorch/audio/pull/1399/files#r617686286), we want to keep batch first for the output too, e.g. [wav2letter](https://github.com/pytorch/audio/blob/master/torchaudio/models/wav2letter.py#L70), [conv_tasnet](https://github.com/pytorch/audio/blob/master/torchaudio/models/conv_tasnet.py#L300), [wavernn](https://github.com/pytorch/audio/blob/master/torchaudio/models/wavernn.py#L284).

```suggestion
            Tensor: Predictor tensor of dimension (batch_size, input_length, number_of_classes).
```

Making this change might not be ideal and means the pipeline (1) needs [another transpose](https://github.com/vincentqb/stt_models/blob/dataparallel/main.py#L161) to be compatible with CTC, (2) this follows the RNN convention though `batch_first` could be used, (3) might get a performance hit (though this should be measured). Thoughts?

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,94 @@
...
+        # T x N x H
+        x, _ = self.bi_rnn(x)
+        # The fifth (non-recurrent) layer takes both the forward and backward units as inputs
+        x = x[:, :, :self.hidden_size] + x[:, :, self.hidden_size:]
+        # T x N x H
+        x = self.fc4(x)
+        # T x N x H
+        x = self.out(x)
+        # T x N x num_classes
+        return x
```
vincentqb, line 94:
```suggestion
        # T x N x num_classes                                                                                                                                               
        x = x.permute(1, 0, 2)                                                                                                                                              
        # N x T x num_classes 
        return x
```

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,94 @@
...
+        if dropout:
+            self.nonlinearity = nn.Sequential(*[
+                self.nonlinearity,
+                nn.Dropout(dropout)
+            ])
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        x = self.fc(x)
+        x = self.nonlinearity(x)
+        return x
```
vincentqb, line 34:
```suggestion
        self.relu_max_clip = relu_max_clip
        self.dropout = dropout

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        x = self.fc(x)
        x = torch.nn.functional.relu(x)
        x = torch.nn.functional.hardtanh(x, 0, self.relu_max_clip)
        if self.dropout:
                x = torch.nn.functional.dropout(x, self.dropout, self.training)
        return x
```

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,94 @@
...
+                 num_classes: int = 40,
+                 dropout: float = 0.0) -> None:
+        super(DeepSpeech, self).__init__()
+        self.hidden_size = hidden_size
+        self.fc1 = FullyConnected(in_features, hidden_size, dropout)
+        self.fc2 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.fc3 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.bi_rnn = nn.RNN(
+            hidden_size, hidden_size, num_layers=1, nonlinearity='relu', bidirectional=True)
+        self.nonlinearity = nn.ReLU()
```
vincentqb, line 61: nit: is this used?

torchaudio/models/deepspeech.py
```diff
@@ -0,0 +1,94 @@
...
+        self.fc4 = FullyConnected(hidden_size, hidden_size, dropout)
+        self.out = nn.Sequential(*[
+            nn.Linear(hidden_size, num_classes),
+            nn.LogSoftmax(dim=2)
+        ])
```
vincentqb, line 66: As mentioned above, the sequential layer is not needed here. The `log_softmax` functional can be used in the forward function for readability.

vincentqb (commented):
Thanks a lot for working on this :) I gave a few comments around the organization of the model.

torchaudio/models/deepspeech.py
discort: nice catch.
removing

torchaudio/models/deepspeech.py
discort: thanks

torchaudio/models/deepspeech.py
discort: All suggestions look reasonable and I have already addressed your comments, thank you for sharing it with me.

However, I'm still don't follow your 3rd point. Could you please elaborate it a bit, why keeping `batch_first` can bring a performance hit?
In the mentioned example you are using simple `DataParallel` and here is my understanding of how it works (on 2 devices for simplicity):
- the model is copied into two devices
- dataloader splits a batch into 2 chunks
- then it feeds each model by each chunk during forward pass
- then, using the outputs of models, each device calculates ctc loss, the loss values are copied into single tensor (on RAM) and reduction is performed on cpu.
- on a backward pass, the gradients are computed on 2 devices and then they are 'all-reduce'd on cpu.

If I understood the process correctly, what's the intuition of keeping `batch_first` in the model's output?

vincentqb: In summary: let's have batch first for the output too for readability and consistency with the other models :)

> Could you please elaborate it a bit, why keeping `batch_first` can bring a performance hit?

The question about performance that I was raising was whether the forward method in `RNN` is "faster" when batch is first or not. For instance, this small user test [here](https://stackoverflow.com/questions/63822152/pytorch-rnn-is-more-efficient-with-batch-first-false) does not seem to find a significant difference in speed. Based on this small test, let's not worry about this here then.

> In the mentioned example you are using simple `DataParallel` and here is my understanding of how it works (on 2 devices for simplicity):

DataParallel applies only to the computation of the model:
* the model is copied to the two devices
* the batch is split in to along the first dimension and sent to each devices
* the forward is computed on each device
* the output tensors are concatenated along the first dimension
* the loss function is then computed on that new tensor

discort: Thank you for the detailed explanation.  Let me know if anything is wrong.

vincentqb (approved):
LGTM! I just did a re-run with the latest version, and got a slightly better best validation of 0.259 at epoch 22.

I'll push a commit to change the names of the parameter to be more consistent with WaveRNN and the convention described in the [README](https://github.com/pytorch/audio/blob/master/README.md#conventions), wait for the tests to run, and merge this PR. Thanks for working on this! :)

vincentqb:
(rebased)

Issue #1375, opened 2021-03-09
🚀 Enable gradient pass through all biquad filter function


yoyololicon:
## 🚀 Feature

I propose make all `*_biquad` filter function become differentiable respect to their parameters (`Q`, `gain`, `central_freq`... etc.)

## Motivation

Recently there is a trend on using differentiable IIR filters to learn frequency responses efficiently. 
In the last year DAFX2020 already about 3 papers described similar idea. Some are produced by well known companies (Native Instruments, iZotope). Some also released their [source code](https://github.com/boris-kuz/differentiable_iir_filters), but to enable automatic differentiation, they use an RNN-like, for-loop structure, which is relatively slow and memory inefficient. 
Because our implementation of IIR (the `lfilter`) use in-place operation, it opens up possibilities to offer fast and accurate differentiable IIR filters.

## Pitch

Change parameters (like `gain`, `Q`, `central_freq`) type hint from `float` to `Union[float, Tensor]`, and replace those `math.*` function to corressponding `torch.*` one.

## Additional context

For references:

- Differentiable IIR Filters for Machine Learning Applications
Boris Kuznetsov; Julian Parker; Fabian Esqueda
- Optimization of Cascaded Parametric Peak and Shelving Filters With Backpropagation Algorithm
Purbaditya Bhattacharya; Patrick Nowak; Udo Zölzer
- Neural Parametric Equalizer Matching Using Differentiable Biquads
Shahan Nercessian

mthrok:
Hi @yoyololicon 

Thanks for the proposal. We are also interested in differential DSP (like in (magenta)[https://magenta.tensorflow.org/ddsp]), but we did not get to work on it due to the lack of the engineering resource. I second your proposal, it sounds like a natural extension of the current `lfilter` update, it does not sound like involve BC-breaking change.

We will be needing tests for ensuring the autograd support for these filters, and we, torchaudio library, are new to the nature of autograd by ourselves so there might be some things that we need to sort out and make an official stance (like the standard for numerical precision mentioned in #1310, which is new for us and ideally we want to establish a good guideline for the tolerance.)

What do you think @cpuhrsch ?

cpuhrsch:
This sounds great! I'm all for this.

yoyololicon:
@cpuhrsch @mthrok 
I was prototyping the feature, and found that torch jit doesn't support union type like `Union[float, Tensor]`, leading to fail on torch script consistency tests.
[...]

PR #1400, opened 2021-03-19
add autograd to biquad filters

yoyololicon:
implement feature requested in #1375 

Affected functions:

- `biquad`
- `band_biquad`
- `bass_biquad`
- `allpass_biquad`
- `highpass_biquad`
- `lowpass_biquad`
- `bandpass_biquad`
- `bandreject_biquad`
- `equalizer_biquad`


torchaudio/functional/filtering.py
```diff
@@ -83,14 +89,20 @@ def allpass_biquad(
         http://sox.sourceforge.net/sox.html
         https://www.w3.org/2011/audio/audio-eq-cookbook.html#APF
     """
+    dtype = waveform.dtype
+    device = waveform.device
+    central_freq = _float2Tensor(central_freq, dtype, device).squeeze()
+    Q = _float2Tensor(Q, dtype, device).squeeze()
```
cpuhrsch, line 95: Why the call to squeeze here each time?

yoyololicon: To make sure that arguments pass to `biquad` in return statement are scalar tensors, so it can be jittable.
But I found that it's not necessary, cuz if user pass unit length tensor (like `torch.Tensor([1.])`) as input and use `torch.jit.script`, it will throw error in the first place before calling `biquad`.
I will remove these lines.

torchaudio/functional/filtering.py
```diff
@@ -11,6 +11,12 @@ def _dB2Linear(x: float) -> float:
     return math.exp(x * math.log(10) / 20.0)
 
 
+def _float2Tensor(x: float, dtype: torch.dtype, device: torch.device):
+    if not isinstance(x, Tensor):
```
cpuhrsch, line 15: nit: use [torch.is_tensor](https://pytorch.org/docs/master/generated/torch.is_tensor.html#torch.is_tensor).

torchaudio/functional/filtering.py
```diff
@@ -11,6 +11,12 @@ def _dB2Linear(x: float) -> float:
     return math.exp(x * math.log(10) / 20.0)
 
 
+def _float2Tensor(x: float, dtype: torch.dtype, device: torch.device):
```
cpuhrsch, line 14: This is very similar to [torch.as_tensor](https://pytorch.org/docs/master/generated/torch.as_tensor.html#torch.as_tensor). For example

```
>>> t = torch.as_tensor(4, dtype=torch.float)
>>> t
tensor(4.)
>>> t.dtype
torch.float32
```

torchaudio/functional/filtering.py
```diff
@@ -269,10 +302,17 @@ def biquad(
...
+    b2 = _float2Tensor(b2, dtype, device).view(1)
+    a0 = _float2Tensor(a0, dtype, device).view(1)
+    a1 = _float2Tensor(a1, dtype, device).view(1)
+    a2 = _float2Tensor(a2, dtype, device).view(1)
+
     output_waveform = lfilter(
         waveform,
-        torch.tensor([a0, a1, a2], dtype=dtype, device=device),
-        torch.tensor([b0, b1, b2], dtype=dtype, device=device),
+        torch.cat([a0, a1, a2]),
```
cpuhrsch, line 314: Here you could also use [torch.as_tensor](https://pytorch.org/docs/master/generated/torch.as_tensor.html#torch.as_tensor) and construct the array right away.
```
>>> torch.as_tensor([4, torch.tensor(2), 3], dtype=torch.float)
tensor([4., 2., 3.])
```

yoyololicon: Have tried this approach but it didn't pass `gradcheck` so roll back to use concatenate.

cpuhrsch (approved):
Looks fine to me! @mthrok can you take another final look?

mthrok (approved):
The change looks good and this is a very cool change. Thanks @yoyololicon 

Couple of thoughts (for follow-up)
1. Should we mention the support of autograd in documentation for the functions that are tested?
2. Can we introduce a helper method for `Autograd` test that handles `dtype`, `device` and `requires_grad` property, so that each test method looks simpler. Example -> https://github.com/pytorch/audio/blob/e4a0bd2ceb6c80ec90194ce519bbd9589926034b/test/torchaudio_unittest/transforms/autograd_test_impl.py
3. What about the second order gradient? i.e. `gradgradcheck`

yoyololicon:
> 2. Can we introduce a helper method for `Autograd` test that handles `dtype`, `device` and `requires_grad` property, so that each test method looks simpler. Example -> https://github.com/pytorch/audio/blob/e4a0bd2ceb6c80ec90194ce519bbd9589926034b/test/torchaudio_unittest/transforms/autograd_test_impl.py

Looks good, will work on it later if I have time.

> 3. What about the second order gradient? i.e. `gradgradcheck`

Current implementation can't support higher order derivative (still use inplace operation internally).
I'm not sure whether there is a need for it, personally a first order derivative is sufficient.

PR #1401, opened 2021-03-19
Fix broken list of checks

dmpolukhin:
Summary: Extra spaces broke list of checks.

Differential Revision: D27125520


.clang-tidy
```diff
@@ -1,32 +1,35 @@
 ---
-# NOTE there must be no spaces before the '-', so put the comma first.
+# NOTE there must be no spaces before the '-' and check name.
+# If you edit this listplease virify list of enabled check with
```
vincentqb, line 3:
```suggestion
# If you edit this list, please verify list of enabled check with
```

.clang-tidy
```diff
@@ -1,32 +1,35 @@
...
+#   clang-tidy --list-checks
```
vincentqb, line 4: To enforce this, we should add this check to CircleCI [here](https://github.com/pytorch/audio/tree/master/.circleci/unittest/linux/scripts)

mthrok:
D27354634 has been landed in fbcode. @parmeet verified that this PR contains the same diff. I am landing this for the sake of syncing GitHub and fbcode.

Issue #1269, opened 2021-02-15
RFC: The future of Kaldi compliance module


mthrok:
# Request For Comment: The future of Kaldi-compatible features

## Problems
`torchaudio.compliance.kaldi` implements functionalities that tries to reproduce Kaldi's feature extractions, and this module has many issues, and causing headache for maintainers.
1. Inconsistent design
While the rest of the `torchaudio` library is standardized to work with floating-point Tensor with value range `[-1.0, 1.0]`, the ported Kaldi implementations are not necessarily following this ([example](https://github.com/pytorch/audio/issues/371#issuecomment-625613872))
2. Does not support batch dimension well. #675 #1245
3. `compliance` is not the right name https://github.com/pytorch/audio/issues/281
4. While it's called `compliance`, it does not match with the result of Kaldi's CLI
I think the natural expectation that users get from the `compliance.kaldi` is that you can get a result that matches Kaldi, (and possibly in an easy manner)
#332, #328, #400, 
5. It's slow #908 #1057
The code translated verbatim from the original Kaldi's C++ code typically is slower than the original implementation. (100x is not uncommon) To overcome this the code has to be changed to adopt PyTorch-style operation. (Kaldi does par element access very efficiently, which incurs a lot of overhead for PyTorch framework) This incurs huge maintenance overhead, 1 for implementing our custom code and 2 for catching up with upstream Kaldi.

## Possible solutions

Before going into the detail of possible solutions, I note that removing Kaldi compatibility features from `torchaudio` was also mentioned as a possibility. It was not included in the original plan of `torchaudio`. I am not familiar on this matter and I am not advocating this but keeping it as a possibility.

The following describes some partial solutions and considerations I have put so far.

### Module location.

For problems like 1 (inconsistent design) and 3 (naming), we can add a new interface to `torchaudio.functional`.

We have some Kaldi-compatible functions in `torchaudio.functional` module, such as [sliding_window_cmn]( https://github.com/pytorch/audio/blob/d58ac213db1e361342b9b704b125b83214a7dcbb/torchaudio/functional/functional.py#L877-L883) and [compute_kaldi_pitch](https://github.com/pytorch/audio/blob/d58ac213db1e361342b9b704b125b83214a7dcbb/torchaudio/functional/functional.py#L1049-L1069). They are placed under `torchaudio.functional` because they are confirmed to work on floating-point Tensors and their behaviors are consistent with other feature implementations.

[...]

Issue #853, opened 2020-08-04
Move computation code from kaldi implementation into functional


vincentqb:
Functions in `torchaudio.compliance.kaldi` should only provide a kaldi-compatible interface to `torchaudio.functional`.

- [ ] [resample waveform](https://github.com/pytorch/audio/blob/master/torchaudio/compliance/kaldi.py#L892)

Suggestion on approaching other functions in `torchaudio.compliance.kaldi`?

PR #1402, opened 2021-03-19
Move resample to functional and add librosa comparison

carolineechen:
- add `resample` to `torchaudio.functional`, copied over from kaldi compliance `resample_waveform`
- adds batching to kaldi compliance interface `resample_waveform`, which now wraps `functional.resample`
- add test to compare torchaudio resample against `librosa.resample`

### DIFF
--- a/test/torchaudio_unittest/compliance_kaldi_test.py
+++ b/test/torchaudio_unittest/compliance_kaldi_test.py
@@ -3,7 +3,6 @@ import math
 
 import torch
 import torchaudio
-import torchaudio.functional as F
 import torchaudio.compliance.kaldi as kaldi
 
 from torchaudio_unittest import common_utils
@@ -179,21 +178,21 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
 
     def test_resample_waveform(self):
         def get_output_fn(sound, args):
-            output = F.resample(sound.to(torch.float32), args[1], args[2])
+            output = kaldi.resample_waveform(sound.to(torch.float32), args[1], args[2])
             return output
 
         self._compliance_test_helper(self.test2_filepath, 'resample', 32, 3, get_output_fn, atol=1e-2, rtol=1e-5)
 
     def test_resample_waveform_upsample_size(self):
-        upsample_sound = F.resample(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr * 2)
+        upsample_sound = kaldi.resample_waveform(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr * 2)
         self.assertTrue(upsample_sound.size(-1) == self.test1_signal.size(-1) * 2)
 
     def test_resample_waveform_downsample_size(self):
-        downsample_sound = F.resample(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr // 2)
+        downsample_sound = kaldi.resample_waveform(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr // 2)
         self.assertTrue(downsample_sound.size(-1) == self.test1_signal.size(-1) // 2)
 
     def test_resample_waveform_identity_size(self):
-        downsample_sound = F.resample(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr)
+        downsample_sound = kaldi.resample_waveform(self.test1_signal, self.test1_signal_sr, self.test1_signal_sr)
         self.assertTrue(downsample_sound.size(-1) == self.test1_signal.size(-1))
 
     def _test_resample_waveform_accuracy(self, up_scale_factor=None, down_scale_factor=None,
@@ -213,7 +212,7 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
         original_timestamps = torch.arange(0, duration, 1.0 / sample_rate)
 
         sound = 123 * torch.cos(2 * math.pi * 3 * original_timestamps).unsqueeze(0)
-        estimate = F.resample(sound, sample_rate, new_sample_rate).squeeze()
+        estimate = kaldi.resample_waveform(sound, sample_rate, new_sample_rate).squeeze()
 
         new_timestamps = torch.arange(0, duration, 1.0 / new_sample_rate)[:estimate.size(0)]
         ground_truth = 123 * torch.cos(2 * math.pi * 3 * new_timestamps)
@@ -240,11 +239,11 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
         for i in range(num_channels):
             multi_sound[i, :] *= (i + 1) * 1.5
 
-        multi_sound_sampled = F.resample(multi_sound, self.test1_signal_sr, self.test1_signal_sr // 2)
+        multi_sound_sampled = kaldi.resample_waveform(multi_sound, self.test1_signal_sr, self.test1_signal_sr // 2)
 
         # check that sampling is same whether using separately or in a tensor of size (c, n)
         for i in range(num_channels):
             single_channel = self.test1_signal * (i + 1) * 1.5
-            single_channel_sampled = F.resample(single_channel, self.test1_signal_sr,
-                                                self.test1_signal_sr // 2)
+            single_channel_sampled = kaldi.resample_waveform(single_channel, self.test1_signal_sr,
+                                                             self.test1_signal_sr // 2)
             self.assertEqual(multi_sound_sampled[i, :], single_channel_sampled[0], rtol=1e-4, atol=1e-7)
--- a/test/torchaudio_unittest/compliance_kaldi_test.py
+++ b/test/torchaudio_unittest/compliance_kaldi_test.py
@@ -3,6 +3,7 @@ import math
 
 import torch
 import torchaudio
+import torchaudio.functional as F
 import torchaudio.compliance.kaldi as kaldi
 
 from torchaudio_unittest import common_utils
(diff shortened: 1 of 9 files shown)


test/torchaudio_unittest/functional/librosa_compatibility_test.py
```diff
@@ -109,6 +109,25 @@ def test_amplitude_to_DB(self):
...
+        waveform, sample_rate = common_utils.load_wav(input_path)
+
+        upsample_rate = sample_rate * 2
+        downsample_rate = sample_rate // 2
+
+        ta_upsampled = F.resample(waveform, sample_rate, upsample_rate)
+        lr_upsampled = librosa.resample(waveform.squeeze(0).numpy(), sample_rate, upsample_rate)
+        lr_upsampled = torch.from_numpy(lr_upsampled).unsqueeze(0)
+
+        self.assertEqual(ta_upsampled, lr_upsampled, atol=1e-4, rtol=1e-5)
```
carolineechen, line 123: @vincentqb @cpuhrsch what's an acceptable margin of error for this comparison? using these values gives an assertion error ` Tensors failed to compare as equal!With rtol=1e-05 and atol=0.0001, found 28142 element(s) (out of 128000) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.006743520498275757 (0.23187145590782166 vs. 0.2386149764060974), which occurred at index (0, 127992).`

vincentqb: You can look at the [kaldi compliance](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/compliance_kaldi_test.py#L198) as guidance. As you can see, the tolerances are fairly high there too.

Since we are not changing the algorithm as part of this PR, these new tests are informational, and we take the threshold that makes them pass (after we make sure that we match the correct parameters of course). The threshold therefore becomes new signal/information that we learned from this exercise.

IMHO, based on the numbers you gave, assuming we use the correct corresponding parameters for each resampling functions, I'd say `atol=1e-2` which is also used [here](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/compliance_kaldi_test.py#L184).

torchaudio/compliance/kaldi.py
```diff
@@ -751,96 +749,3 @@ def mfcc(
...
-        window = torch.cos(t / lowpass_filter_width / 2)**2
-        kernel = torch.where(t == 0, torch.tensor(1.).to(t), torch.sin(t) / t)
-        kernel.mul_(window)
-        kernels.append(kernel)
-
-    scale = base_freq / orig_freq
-    return torch.stack(kernels).view(new_freq, 1, -1).mul_(scale), width
-
-
-def resample_waveform(waveform: Tensor,
```
vincentqb, line 806: This is breaking backward compatibility by removing the kaldi compliance implementation. If we do this, we should do this in two steps across two releases: the first adds a warning that this function will be removed (and what to use in its place), and the second finally removing this function.

That being said, the idea of the compliance interface was to provide a simple way of migrating to torchaudio. If this interface still fulfills this goal, then we should just keep this function around -- or otherwise thing of removing it #1269.

Another option is to keep the `resample_waveform` function here, and make it call the new functional, see #853. The latter doesn't affect the users, and the existing test will also be validating the new version. Let's start with that. (The only difference for this function is that it will now also support batching.)

test/torchaudio_unittest/compliance_kaldi_test.py
```diff
@@ -178,21 +179,21 @@ def test_mfcc_empty(self):
 
     def test_resample_waveform(self):
         def get_output_fn(sound, args):
-            output = kaldi.resample_waveform(sound.to(torch.float32), args[1], args[2])
```
vincentqb, line 181: as per comment below, let's not yet change the existing tests

vincentqb (commented):
Great, thanks! I'm glad to see this happening :)

torchaudio/compliance/kaldi.py
```diff
@@ -751,96 +749,3 @@ def mfcc(
...
-                      new_freq: float,
-                      lowpass_filter_width: int = 6) -> Tensor:
-    r"""Resamples the waveform at the new frequency. This matches Kaldi's OfflineFeatureTpl ResampleWaveform
-    which uses a LinearResample (resample a signal at linearly spaced intervals to upsample/downsample
-    a signal). LinearResample (LR) means that the output signal is at linearly spaced intervals (i.e
-    the output signal has a frequency of ``new_freq``). It uses sinc/bandlimited interpolation to
-    upsample/downsample the signal.
-
-    https://ccrma.stanford.edu/~jos/resample/Theory_Ideal_Bandlimited_Interpolation.html
-    https://github.com/kaldi-asr/kaldi/blob/master/src/feat/resample.h#L56
```
vincentqb, line 817: let's change the docstring to reference the functional: concretely, 
```
Resamples the waveform at the new frequency.

This is a wrapper around torchaudio.functional.resample.
```

torchaudio/compliance/kaldi.py
```diff
@@ -753,68 +752,13 @@ def mfcc(
...
-    r"""Resamples the waveform at the new frequency. This matches Kaldi's OfflineFeatureTpl ResampleWaveform
-    which uses a LinearResample (resample a signal at linearly spaced intervals to upsample/downsample
-    a signal). LinearResample (LR) means that the output signal is at linearly spaced intervals (i.e
-    the output signal has a frequency of ``new_freq``). It uses sinc/bandlimited interpolation to
-    upsample/downsample the signal.
+    r"""Resamples the waveform at the new frequency.
 
-    https://ccrma.stanford.edu/~jos/resample/Theory_Ideal_Bandlimited_Interpolation.html
-    https://github.com/kaldi-asr/kaldi/blob/master/src/feat/resample.h#L56
+    This is a wrapper around torchaudio.functional.resample
```
vincentqb, line 761: nit: let's add some formatting:
```suggestion
    This is a wrapper around ``torchaudio.functional.resample``.
```

test/torchaudio_unittest/functional/librosa_compatibility_test.py
```diff
@@ -109,6 +109,25 @@ def test_amplitude_to_DB(self):
...
+        downsample_rate = sample_rate // 2
+
+        ta_upsampled = F.resample(waveform, sample_rate, upsample_rate)
+        lr_upsampled = librosa.resample(waveform.squeeze(0).numpy(), sample_rate, upsample_rate)
+        lr_upsampled = torch.from_numpy(lr_upsampled).unsqueeze(0)
+
+        self.assertEqual(ta_upsampled, lr_upsampled, atol=1e-2, rtol=1e-5)
+
+        ta_downsampled = F.resample(waveform, sample_rate, downsample_rate)
+        lr_downsampled = librosa.resample(waveform.squeeze(0).numpy(), sample_rate, downsample_rate)
```
vincentqb, line 126: question for future work: why is `squeeze` needed before passing to librosa?

torchaudio/transforms.py
```diff
@@ -649,17 +648,7 @@ def forward(self, waveform: Tensor) -> Tensor:
             Tensor: Output signal of dimension (..., time).
         """
         if self.resampling_method == 'sinc_interpolation':
```
vincentqb, line 650: note: we'll leave the discussion around migrating the `resampling_method` parameter to a follow-up PR

torchaudio/compliance/kaldi.py
```diff
@@ -826,21 +770,4 @@ def resample_waveform(waveform: Tensor,
     Returns:
         Tensor: The waveform at the new frequency
     """
-    assert waveform.dim() == 2
```
vincentqb, line 829: This PR is also adding batching to kaldi compliance interface. Please add that note in the description of the PR so we know this feature has been added as part of this PR.

torchaudio/compliance/kaldi.py
```diff
@@ -753,68 +752,13 @@ def mfcc(
...
-    the output signal has a frequency of ``new_freq``). It uses sinc/bandlimited interpolation to
-    upsample/downsample the signal.
+    r"""Resamples the waveform at the new frequency.
 
-    https://ccrma.stanford.edu/~jos/resample/Theory_Ideal_Bandlimited_Interpolation.html
-    https://github.com/kaldi-asr/kaldi/blob/master/src/feat/resample.h#L56
+    This is a wrapper around ``torchaudio.functional.resample``.
 
     Args:
         waveform (Tensor): The input signal of size (c, n)
```
vincentqb, line 764: nit: (c, n) --> (..., time) as below
```suggestion
        waveform (Tensor): The input signal of size (..., time)
```

vincentqb (approved):
Alrighty, LGTM overall :)

torchaudio/compliance/kaldi.py
carolineechen: good catch 👍

test/torchaudio_unittest/functional/librosa_compatibility_test.py
vincentqb: have you looked into this @carolineechen ?

carolineechen: `waveform` has dimensions `(1, n)`, but [librosa.resample](https://librosa.org/doc/main/generated/librosa.resample.html) requires it to be of shape `(n,)` or `(2, n)`

test/torchaudio_unittest/functional/librosa_compatibility_test.py
carolineechen: I had only tried the librosa default values for this test, but yup, I can look into the other parameters

PR #1405, opened 2021-03-22
Fix checksum for the YESNO dataset

danieleds:
The length of the checksum for the YESNO dataset is inconsistent with the specified algorithm (MD5) and its value does not correspond to the other supported algorithm (sha256).

This PR updates the hash to the correct sha256 value.


mthrok:
Hi @danieleds 

Thanks for making the PR.

@carolineechen Could you take a look at this?

carolineechen (approved):
Looks good, thanks!

PR #1411, opened 2021-03-29
[ROCM] add rocm support

micmelesse:
This PR adds support for building pytorch audio in ROCM. We currently disable 5 tests. There is still more work to be done but with the changes in this PR, we can build and test pytorch audio on ROCM.

### DIFF
--- a/test/torchaudio_unittest/backend/soundfile/save_test.py
+++ b/test/torchaudio_unittest/backend/soundfile/save_test.py
@@ -11,7 +11,7 @@ from torchaudio_unittest.common_utils import (
     get_wav_data,
     load_wav,
     nested_params,
-    skipCUDAIfRocm,
+    skipIfRocm,
 )
 from .common import (
     fetch_wav_subtype,
@@ -281,7 +281,7 @@ class TestFileObject(TempDirMixin, PytorchTestCase):
         self._test_fileobj('wav')
 
     @skipIfFormatNotSupported("FLAC")
-    @skipCUDAIfRocm
+    @skipIfRocm
     def test_fileobj_flac(self):
         """Saving audio via file-like object works"""
         self._test_fileobj('flac')
--- a/test/torchaudio_unittest/backend/soundfile/save_test.py
+++ b/test/torchaudio_unittest/backend/soundfile/save_test.py
@@ -11,6 +11,7 @@ from torchaudio_unittest.common_utils import (
     get_wav_data,
     load_wav,
     nested_params,
+    skipCUDAIfRocm,
 )
 from .common import (
     fetch_wav_subtype,
@@ -280,6 +281,7 @@ class TestFileObject(TempDirMixin, PytorchTestCase):
         self._test_fileobj('wav')
 
     @skipIfFormatNotSupported("FLAC")
+    @skipCUDAIfRocm
     def test_fileobj_flac(self):
         """Saving audio via file-like object works"""
         self._test_fileobj('flac')
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -98,4 +98,5 @@ def skipIfNoModule(module, display_name=None):
 skipIfNoCuda = unittest.skipIf(not torch.cuda.is_available(), reason='CUDA not available')
 skipIfNoSox = unittest.skipIf(not is_sox_available(), reason='Sox not available')
 skipIfNoKaldi = unittest.skipIf(not is_kaldi_available(), reason='Kaldi not available')
-skipCUDAIfRocm =unittest.skipIf(os.getenv('PYTORCH_AUDIO_TEST_WITH_ROCM', '0') == '1', reason="test doesn't currently work on the ROCm stack")
+skipCUDAIfRocm = unittest.skipIf(os.getenv('PYTORCH_AUDIO_TEST_WITH_ROCM', '0') == '1',
+                                 reason="test doesn't currently work on the ROCm stack")
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -98,3 +98,4 @@ def skipIfNoModule(module, display_name=None):
 skipIfNoCuda = unittest.skipIf(not torch.cuda.is_available(), reason='CUDA not available')
 skipIfNoSox = unittest.skipIf(not is_sox_available(), reason='Sox not available')
 skipIfNoKaldi = unittest.skipIf(not is_kaldi_available(), reason='Kaldi not available')
+skipCUDAIfRocm =unittest.skipIf(os.getenv('PYTORCH_AUDIO_TEST_WITH_ROCM', '0') == '1', reason="test doesn't currently work on the ROCm stack")
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -98,5 +98,5 @@ def skipIfNoModule(module, display_name=None):
 skipIfNoCuda = unittest.skipIf(not torch.cuda.is_available(), reason='CUDA not available')
 skipIfNoSox = unittest.skipIf(not is_sox_available(), reason='Sox not available')
 skipIfNoKaldi = unittest.skipIf(not is_kaldi_available(), reason='Kaldi not available')
-skipCUDAIfRocm = unittest.skipIf(os.getenv('PYTORCH_AUDIO_TEST_WITH_ROCM', '0') == '1',
+skipIfRocm = unittest.skipIf(os.getenv('TORCHAUDIO_TEST_WITH_ROCM', '0') == '1',
                                  reason="test doesn't currently work on the ROCm stack")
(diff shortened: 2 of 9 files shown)


micmelesse:
retest

micmelesse:
@facebook-github-bot retest

test/torchaudio_unittest/backend/soundfile/save_test.py
```diff
@@ -280,6 +281,7 @@ def test_fileobj_wav(self):
         self._test_fileobj('wav')
 
     @skipIfFormatNotSupported("FLAC")
+    @skipCUDAIfRocm
```
mthrok, line 284: The functionality tested here should be all Python. What kind of error do you see when you build with ROCm?

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -98,3 +98,5 @@ def skipIfNoModule(module, display_name=None):
 skipIfNoCuda = unittest.skipIf(not torch.cuda.is_available(), reason='CUDA not available')
 skipIfNoSox = unittest.skipIf(not is_sox_available(), reason='Sox not available')
 skipIfNoKaldi = unittest.skipIf(not is_kaldi_available(), reason='Kaldi not available')
+skipCUDAIfRocm = unittest.skipIf(os.getenv('PYTORCH_AUDIO_TEST_WITH_ROCM', '0') == '1',
```
mthrok, line 101: 1. Instead of `PYTORCH_AUDIO`, `TORCHAUDIO` is the keywords we use everywhere.
2. This logic looks like it's inverted. When the `PYTORCH_AUDIO_TEST_WITH_ROCM` is `1`, it evaluates to `unittet.skipIf(True, ...)`. Which sounds counter intuitive. Am I reading it wrong?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -34,6 +37,7 @@ def func(tensor):
         tensor = common_utils.get_whitenoise()
         self._assert_consistency(func, tensor)
 
+    @skipCUDAIfRocm
```
mthrok, line 40: If you see what function is not supported by ROCm in the error message, could you add a comment?

mthrok (commented):
Thanks for the contribution.
As I have no experience/knowledge in ROCm, I will leave the review of the ROCm part to those with expertise, but one naive question I got is the following.

At this moment, torchaudio only contains Python code and custom CPU code. There is no custom CUDA code. (although we plan to do add some CUDA kernel)
I imagined that we could just build torchaudio with C++ complier and the corresponding PyTorch binary. Why does it need to ROCm at build time?

test/torchaudio_unittest/backend/soundfile/save_test.py
micmelesse: > Thanks for the contribution.
> As I have no experience/knowledge in ROCm, I will leave the review of the ROCm part to those with expertise, but one naive question I got is the following.
> 
> At this moment, torchaudio only contains Python code and custom CPU code. There is no custom CUDA code. (although we plan to do add some CUDA kernel)
> I imagined that we could just build torchaudio with C++ complier and the corresponding PyTorch binary. Why does it need to ROCm at build time?


@mthrok, the reason was because I encountered multiple cmake errors from various modules failing to find the path to various rocm libraries.  I added the LoadHip cmake file and use it find all the various rocm libs.

micmelesse:
> Thanks for the contribution.
> As I have no experience/knowledge in ROCm, I will leave the review of the ROCm part to those with expertise, but one naive question I got is the following.
> 
> At this moment, torchaudio only contains Python code and custom CPU code. There is no custom CUDA code. (although we plan to do add some CUDA kernel)
> I imagined that we could just build torchaudio with C++ complier and the corresponding PyTorch binary. Why does it need to ROCm at build time?


@mthrok, the reason was because I encountered multiple cmake errors from various modules failing to find the path to various rocm libraries.  I added the LoadHip cmake file and use it find all the various rocm libs.

test/torchaudio_unittest/backend/soundfile/save_test.py
micmelesse: It was a mismatch error.

`==================================================================================== FAILURES ====================================================================================
________________________________________________________________________ TestFileObject.test_fileobj_flac ________________________________________________________________________

self = <torchaudio_unittest.backend.soundfile.save_test.TestFileObject testMethod=test_fileobj_flac>

    @skipIfFormatNotSupported("FLAC")
    @skipCUDAIfRocm
    def test_fileobj_flac(self):
        """Saving audio via file-like object works"""
>       self._test_fileobj('flac')

test/torchaudio_unittest/backend/soundfile/save_test.py:287: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/torchaudio_unittest/backend/soundfile/save_test.py:277: in _test_fileobj
    self.assertEqual(expected, found, atol=1e-4, rtol=1e-8)
/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py:1285: in assertEqual
    exact_dtype=exact_dtype, exact_device=exact_device)
/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py:1285: in assertEqual
    exact_dtype=exact_dtype, exact_device=exact_device)
/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py:1296: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing -0.99993896484375 and 1.00006103515625 gives a difference of 2.0, but the allowed difference with rtol=1e-08 and atol=0.0001 is only 0.00010001000061035157!
============================================================================ short test summary info =============================================================================
FAILED test/torchaudio_unittest/backend/soundfile/save_test.py::TestFileObject::test_fileobj_flac - AssertionError: False is not true : Scalars failed to compare as equal! Com...
=============================================================================== 1 failed in 1.07s ================================================================================`

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
micmelesse: This also gives a a mismatch error.
```
==================================================================================== FAILURES ====================================================================================
_____________________________________________________________________ TestFunctionalFloat32.test_griffinlim ______________________________________________________________________

self = <torchaudio_unittest.functional.torchscript_consistency_cuda_test.TestFunctionalFloat32 testMethod=test_griffinlim>

    @skipCUDAIfRocm
    def test_griffinlim(self):
        def func(tensor):
            n_fft = 400
            ws = 400
            hop = 200
            window = torch.hann_window(ws, device=tensor.device, dtype=tensor.dtype)
            power = 2.
            momentum = 0.99
            n_iter = 32
            length = 1000
            rand_int = False
            return F.griffinlim(tensor, window, n_fft, hop, ws, power, n_iter, momentum, length, rand_int)
    
        tensor = torch.rand((1, 201, 6))
>       self._assert_consistency(func, tensor)

test/torchaudio_unittest/functional/torchscript_consistency_impl.py:55: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
test/torchaudio_unittest/functional/torchscript_consistency_impl.py:24: in _assert_consistency
/opt/conda/lib/python3.6/site-packages/torch/testing/_internal/common_utils.py:1251: in assertEqual
E   AssertionError: False is not true : Tensors failed to compare as equal!With rtol=1.3e-06 and atol=1e-05, found 996 element(s) (out of 1000) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 0.20305343717336655 (-0.11751212924718857 vs. 0.08554130792617798), which occurred at index (0, 53).
================================================================================ warnings summary ================================================================================
test/torchaudio_unittest/functional/torchscript_consistency_cuda_test.py::TestFunctionalFloat32::test_griffinlim
  /opt/conda/lib/python3.6/site-packages/torch/functional.py:656: UserWarning: istft will require a complex-valued input tensor in a future PyTorch release. Matching the output from stft with return_complex=True.  (Triggered internally at  /var/lib/jenkins/pytorch/aten/src/ATen/native/SpectralOps.cpp:793.)
    normalized, onesided, length, return_complex)

============================================================================ short test summary info =============================================================================
FAILED test/torchaudio_unittest/functional/torchscript_consistency_cuda_test.py::TestFunctionalFloat32::test_griffinlim - AssertionError: False is not true : Tensors failed to...
```

micmelesse:
@mthrok What should I do about the failing circleci. It seems to have nothing to do with my changes and I cannot rerun the test.

test/torchaudio_unittest/common_utils/case_utils.py
micmelesse: 1. I will make the change.
2. That is the functionality that we want. We do a similar thing in pytorch. We skip some tests on rocm even though they work on cuda. I can rename it to skipIfRocm to be more clear.

mthrok: Oh I see. I was indeed reading it wrong. Thanks for the clarification.

mthrok:
> @mthrok What should I do about the failing circleci test, binary_macos_conda_py3.7. It seems to have nothing to do with my changes and I cannot rerun the test.

Build issues are often caused by external reason or upstream issues so we can ignore it.

seemethere (approved):

mthrok:
Ideally, this should be tested and we should have ROCm build in our CI but that will be a lot of work and I do not want this PR to become stale for that reason, so I am going to merge this.

mthrok (approved):

Issue #1414, opened 2021-04-01
Autograd tests for Transforms


mthrok:
Until recent, we have been assumed that ops provided in torchaudio support autograd just because they are implemented with PyTorch. However, this assumption was not correct all the time. For example, in #704, it was pointed out that `lfitler` does not support autograd, and this was resolved in #1310 with proper unit tests by community contribution. Similarly, as a part of #1337, I have added autograd test to some transforms in #1340. We would like to extend the autograd testing to as many functionals and transforms as possible. We would like to ask your help.

## Steps
1. Pick a transform from the list bellow and see if someone is working on it. If not, leave a comment in this thread that you will be working on it.
2. Add a test in [AutogradTestMixin](https://github.com/pytorch/audio/blob/7d45851d97d35a703817e2b037df03d57b748018/test/torchaudio_unittest/transforms/autograd_test_impl.py#L29) similar to the existing tests.
**NOTE** Please first try adding test without providing `nondet_tol` to see if the transform's backward pass is deterministic or not. If you see `Backward is not reentrant` error message, that means it is not deterministic. Please report it back here, so that we can discuss how to handle it.
3. Run test with `(cd test && pytest torchaudio_unittest/transforms/autograd_*_test.py)`.
4. Make a PR and add `cc @mthrok` in the description (or comment)

 **Note**: We are not sure if all the transforms actually support autograd. If you find a Transform does not support autograd please report back in this thread.

For the instruction to set up development environment, please refer to [CONTRIBUTING.md](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md).

## Transforms
- [x] `AmplitudeToDB` #1447 
    is used by `MFCC` so it should support but having an independent test is be nice
- [x] `Spectrogram` #1340
- [x] `GriffinLim` #1421 
- [x] `MelScale` #1467 
    is used by `MelSpectrogram` so it should support but having an independent test is be nice
- ~`InverseMelScale`~
    this transform most likely does not support autograd
- [x] `MelSpectrogram` #1340
- [x] `MFCC` #1415
- ~`MuLawEncoding`~
- ~`MuLawDecoding`~
    `MuLaw`s are quantization operation, thus they do not support autograd. 
- [x] `Resample` #1416
- [x] `ComputeDeltas` #1422
- ~`TimeStretch`~ ~#1420~
    `TimeStretch` requires coordinate transform from cartesian to polar which uses `atan2`. `atan2` is not differentiable everywhere.
[...]

PR #1415, opened 2021-04-01
Autograd tests for Transforms MFCC

krishnakalyan3:
#1414

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -61,7 +61,7 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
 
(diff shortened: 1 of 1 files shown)


krishnakalyan3:
cc @mthrok

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -60,3 +60,9 @@ def test_melspectrogram(self):
         transform = T.MelSpectrogram(sample_rate=sample_rate)
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
+
+    def test_mfcc(self):
```
mthrok, line 64: Note from offline discussion:

We should check for different values of `log_mel` as it changes the code path and operation used to compute MFCC.

https://github.com/pytorch/audio/blob/52decd2a4346f783a6229603716a08c2640aebe9/torchaudio/transforms.py#L558-L562

mthrok (approved):
Looks good. Thanks!

PR #1416, opened 2021-04-01
Add Autograd test for T.Resample

krishnakalyan3:
ref #1414 

cc: @mthrok


krishnakalyan3:
We need parameters for `Resample`.  If yes what should be the range for `orig_freq`, `new_freq`?.

https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L651

mthrok:
> https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L651

@krishnakalyan3 

Thanks for the suggestion. I think we should test at least three cases. down sampling, up sampling and on sampling (just in case).  
we can try something like `orig_freq, new_freq = (8000, 8000), (8000, 4000), (4000, 8000)`.

mthrok:
Thanks!

PR #1420, opened 2021-04-02
Add autograd test to T.TimeStretch (and F.phase_vocoder)

mthrok:
**UPDATE**
Due to the conversion from cartesian coordinate to polar coordinate, `F.phase_vocoder` (and `T.TimeStretch`) is not differentiable everywhere.

---
Part of #1337 
~Looks like `rate=0.7` is not differentiable to second degree.~
~Let's see what are the value ranges of `rate`, that are differentiable to second degree.~
`T.TimeStretch` manipulates spectrogram in polar coordinate.
The angle conversion from Cartesian to Polar (`atan2(Img, Real)`) does not have a 
nice differentiability around `(x=0, y=0)` and `theta = +/- pi`, therefore `T.TimeStrech` is not differentiable there.

**TODO**
- [x] Check if `rate=0.7` issue happens without complex Tensor adaptation (#1430)
    -> Yes it does. In fact the adaptation of complex Tensor makes the TimeStretch work with rate in range 0.8 ~ 1.3.
    -> Filed the issue for `rate=0.7` pytorch/pytorch#55557

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -36,8 +36,8 @@ class AutogradTestMixin(TestBaseMixin):
     ):
         transform = transform.to(dtype=torch.float64, device=self.device)
 
-        # For the autograd test, float32 and cfloat64 are not good enough
-        # we make sure that all the tensors are of float64 or cfloat128
+        # gradcheck and gradgradcheck only pass if the input tensors are of dtype `torch.double` or 
+        # `torch.cdouble`, when the default eps and tolerance values are used.
         inputs_ = []
         for i in inputs:
             if torch.is_tensor(i):
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -36,6 +36,8 @@ class AutogradTestMixin(TestBaseMixin):
     ):
         transform = transform.to(dtype=torch.float64, device=self.device)
 
+        # For the autograd test, float32 and cfloat64 are not good enough
+        # we make sure that all the tensors are of float64 or cfloat128
         inputs_ = []
         for i in inputs:
             if torch.is_tensor(i):
@@ -172,10 +174,10 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=40, duration=1, n_channels=2)
         spectrogram = get_spectrogram(waveform, n_fft=n_fft, power=None)
 
-        # 1e-3 is still too close on CPU
+        # 1e-3 is too small (on CPU)
         epsilon = 1e-2
         too_close = spectrogram.abs() < epsilon
-        spectrogram[too_close] = spectrogram[too_close] / spectrogram[too_close].abs()
+        spectrogram[too_close] = epsilon * spectrogram[too_close] / spectrogram[too_close].abs()
         if test_pseudo_complex:
             spectrogram = torch.view_as_real(spectrogram)
         self.assert_grad(transform, [spectrogram])
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -154,7 +154,7 @@ class AutogradTestMixin(TestBaseMixin):
         [0.7, 0.8, 0.9, 1.0, 1.3],
         [False, True],
     )
-    def test_timestretch(self, rate, test_pseudo_complex):
+    def test_timestretch_non_zero(self, rate, test_pseudo_complex):
         """Verify that ``T.TimeStretch`` does not fail if it's not close to 0
 
         ``T.TimeStrech`` is not differentiable around 0, so this test checks the differentiability
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -153,33 +153,29 @@ class AutogradTestMixin(TestBaseMixin):
         [False, True],
     )
     def test_timestretch(self, rate, test_pseudo_complex):
-        """Verify that ``T.TimeStretch`` does not fail if it's not too close to 0
+        """Verify that ``T.TimeStretch`` does not fail if it's not close to 0
 
         ``T.TimeStrech`` is not differentiable around 0, so this test checks the differentiability
-        for cases where input is not zero, and different configurations of `TimeStretch`.
+        for cases where input is not zero.
 
-        Ideally, we should be testing on Spectrogram of random waveform but it is hard to control
-        the values around zeros.
+        As tested above, when spectrogram contains values close to zero, the gradients are unstable
+        and gradcheck fails.
+
+        In this test, we generate spectrogram from random signal, then we push the points around
+        zero away from the origin.
+
+        This process does not reflect the real use-case, and it is not practical for users, but
+        this helps us understand to what degree the function is differentiable and when not.
         """
         n_fft = 16
         transform = T.TimeStretch(n_freq=n_fft // 2 + 1, fixed_rate=rate)
-        waveform = torch.zeros(2, 40)
+        waveform = get_whitenoise(sample_rate=40, duration=1, n_channels=2)
         spectrogram = get_spectrogram(waveform, n_fft=n_fft, power=None)
 
-        # Epsilon values tried
-        #
-        # Note:
-        #   This is not experimental and comprehensive.
-        #   The result also depends on ``n_fft``.
-        #
-        #                 CPU / CUDA
-        # * 1e-1           ok / ok
-        # * 1e-2           ok / ok
-        # * 1e-3           ok / ok
-        # * 1e-3 + 1e-3j   ok / ok
-        # * 1e-4           ok / NG
-
-        spectrogram += 1e-3
+        # 1e-3 is still too close on CPU
+        epsilon = 1e-2
+        too_close = spectrogram.abs() < epsilon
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -71,3 +86,21 @@ def test_mfcc(self, log_mels):
...
+        [0.7, 0.8, 0.9, 1.0, 1.3],
+        [True, False],
+    )
+    def test_timestretch(self, rate, test_complex):
+        transform = T.TimeStretch(fixed_rate=rate)
+        spectrogram = get_spectrogram(return_complex=test_complex)
+        self.assert_grad(transform, [spectrogram])
+
+    @nested_params(
+        [0.7, 0.8, 0.9, 1.0, 1.3],
```
mthrok, line 100: @anjali411

So the finding is that shrinking the spectrogram in time axis with factor of `0.7` causes 2nd order gradient to be `NaN`, while it works for the other parameters.
What do you think is a good way to communicate this to users?

anjali411: Did it return `NaN`s before we merged the PR migrating `TimeStretch` to complex types? If not, one or more of `abs`, `angle` or `polar` is likely causing the issue.

mthrok: TODO: @mthrok to file an issue in PyTorch core for the sake of record.

mthrok: > Did it return `NaN`s before we merged the PR migrating `TimeStretch` to complex types? 

Testing this in #1430

mthrok: @anjali411 

I ran the test with the code with pseudo complex code and found that the same tests fail for not only `0.7` but also for `0.8`, `0.9` and `1.3`. So This suggests that complex Tensor made these function support autograd.

https://app.circleci.com/pipelines/github/pytorch/audio/5655/workflows/4c1f6dc6-deb6-4730-815f-31946b9af73d/jobs/194026

anjali411: That's interesting. Could you please look into which function is breaking the test (return `nan`s instead of the non `nan` expected value) for the input the test is being ran on? It would be nice to file an issue both for real and complex operators which are returning incorrect value on edge cases!

mthrok: Filed the issue https://github.com/pytorch/pytorch/issues/55557

mthrok:
@anjali411 I have removed the failing case `0.7`. CR please.

mthrok:
Hmm, now the tests are failing for all the cases....

anjali411:
this issue looks similar to https://github.com/pytorch/pytorch/issues/42952, only that the function to blame here is likely `abs`, `angle`, or `polar`. cc. @malfet

mthrok:
> @mthrok One thing to look into here is if the `phase_vocoder` function is expected to be differentiable overall, in which case, we might wanna define custom gradients.

@anjali411  That's a tough question. I am not sure if it is supposed to be a differential. The primal profit of autograd support for ops in torchaudio is that it might enable the E2E task of combined training. For that first-order gradient is enough.

mthrok:
@anjali411 

 My brief understanding of `F.phase_vocoder` algorithm is that it performs phase shift of complex spectrogram in polar coordinate, so it is inevitable to use `atan` when converting the input complex value (in cartesian form) to polar form. 

So, unless there is another formula, I say that `F.phase_vocoder` and `T.TimeStretch` do not support autograd, and it has nothing to do with native complex Tensor.

Lezcano:
For reference, there's no other formula for the gradient. See this plot for what's going on at zero:

mthrok:
Oh, what I meant by "another formula" is a formula for achieving the same result as the current `F.phase_vocoder` without converting the input spectrogram to polar coordinate.

Since the `angle` is applied to the input Tensor, I do not think it's practically possible to ensure non-zero input for the real use case, even if we can do so for testing.

mthrok:
> It would be enough to make sure that the tests do not test points near zero, but I do not know whether that's possible in this particular use case

@Lezcano 

Thanks for the advice. Following @anjali411's suggestion I tried to see if it's possible to come up with a guideline for workaround the zero values. No luck, I bet someone has already invented it if it's something I can come up with easily.

@anjali 

I gave some thoughts on if it is possible to work around the in-differentiability. I think, if we really want to tackle this, we need to re-invent arctangent function.



I think the partial derivatives can be expressed as
 [plot](https://www.wolframalpha.com/input/?i=plot+-y%2F%28x%5E2%2By%5E2%29)
 [plot](https://www.wolframalpha.com/input/?i=plot+x%2F%28x%5E2%2By%5E2%29)

And they go to as close as +/- infinity around zero depending on the phase.
For the sake of `T.TimeStretch`, it's okay to have zero gradient around the origin because in this context, `abs(z)==0` means that the input signal has only an negligible amount of energy in the corresponding frequency bin and for the sake of NN model training passing 0 gradient makes sense. Now the problem is how we can invent a formula that has a nice differentiability around 0 but gets closer to `atan2` as the input gets far from the origin.

Essentially this function needs to satisfy some properties like


and if z is not close to the origin,

and they connect soothly 

similarly their derivatives have to meet the properties like





but have the regular partial derivatives of `atan2` when z is far from 0, while they connect nicely.

Lezcano:
I really like the name `super_atan2`, not gonna lie :D

Now, the first thing that comes to mind is to interpolate with a function that decreases very quickly to zero. The typical example of such a function is `e^{-1/x}`. So, if we denote `r = (x²+y²)/α²`, for a small  `α > 0`, we could consider
`super_atan2(y,x) = (1-r)e^{-1/r} + r atan2(y,x)` on the ball of radius α `x²+y² < α²`, and then we would like to extend it to be `atan2(y,x)` outside of this ball. This extension is continuous but not differentiable, but that should be fine, I reckon.

Now, if we were to do this, it might be even simpler to just have a cut-off at some small α for the derivative, and declare it to be zero after that.

Now, all these things are clearly just hacks. The real question would then be: is it really necessary to make this function differentiable, or would it be better to leave it as it is, and then react depending on whether people file issues?

mthrok:
> Now, the first thing that comes to mind is to interpolate with a function that decreases very quickly to zero. The typical example of such a function is `e^{-1/x}`. So, if we denote `r = (x²+y²)/α²`, for a small `α > 0`, we could consider
> `super_atan2(y,x) = (1-r)e^{-1/r} + r atan2(y,x)` on the ball of radius α `x²+y² < α²`, and then we would like to extend it to be `atan2(y,x)` outside of this ball. This extension is continuous but not differentiable, but that should be fine, I reckon.
>
> Now, if we were to do this, it might be even simpler to just have a cut-off at some small α for the derivative, and declare it to be zero after that.

wow those are cool options! Thanks for the suggestions.

> Now, all these things are clearly just hacks. The real question would then be: is it really necessary to make this function differentiable, or would it be better to leave it as it is, and then react depending on whether people file issues?

Well, for the scope of native complex Tensor support, we do not need this. We can just say `TimeStretch` does not support autograd.
But this god me thinking, it's not just this op but any op that manipulates data in polar coordinate is not differentiable. In case of signal processing and spectrogram, having 0 is inevitable, so the regular `atan` makes it impossible to use polar coordinate in training pipeline, and having a workaround might open a new door for research. Of course I have no idea it will be useful or not.

At this moment it's more of a thought experiment, but this could be also an interesting research topic.

Lezcano:
What I would suggest is, rather than saying that it's not differentiable, I would write down the differentiability issue in a `.. warning::`. That's what we are doing in `torch.linalg`, where it's much less obvious when do the gradients explode. See for example the penultimate warning in https://pytorch.org/docs/master/linalg.html?highlight=linalg%20svd#torch.linalg.svd

mthrok:
Thanks for the tips. Will do so.

mthrok:
@anjali411 Following the tips from @Lezcano, I realized that we need a test to claim the differentiability at non-zero point. So I re-opened the PR. (sorry for flip-flopping)

In the new approach, I push the points close to zero away from zero. so that all the complex value elements have norm at least 1.

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -37,8 +38,12 @@ def assert_grad(
 
         inputs_ = []
         for i in inputs:
-            i.requires_grad = True
-            inputs_.append(i.to(dtype=torch.float64, device=self.device))
+            if torch.is_tensor(i):
+                i = i.to(
+                    dtype=torch.cdouble if i.is_complex() else torch.double,
```
anjali411, line 43: please add a comment to explain that we need this because we want to run gradcheck and gradcheck with double and cdouble

anjali411 (approved):

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -35,10 +36,16 @@ def assert_grad(
     ):
         transform = transform.to(dtype=torch.float64, device=self.device)
 
+        # For the autograd test, float32 and cfloat64 are not good enough
```
anjali411, line 39:
```suggestion
        # gradcheck and gradgradcheck require the input tensors to be of `torch.double` or `torch.cdouble`
        # if the default eps and tolerance are used
```

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -35,10 +36,16 @@ def assert_grad(
...
+        # we make sure that all the tensors are of float64 or cfloat128
```
anjali411, line 40:
```suggestion
```

PR #1421, opened 2021-04-03
Add autograd test for T.GriffinLim

yoyololicon:
ref #1414 
cc @mthrok 

Have incorrect jacobian when doing gradcheck.


test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -65,6 +65,14 @@ def test_melspectrogram(self):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
 
+    def test_griffinlim(self):
+        n_fft = 400
+        n_frames = 5
+        n_iter = 3
+        spec = torch.rand(n_fft // 2 + 1, n_frames) * n_fft
```
anjali411, line 72: Hi @yoyololicon `gradcheck` fails because the `spec` is of dtype `torch.float32` instead of `torch.float64`. @mthrok we should probably assert that the inputs are of `torch.float64` in `assert_grad` to better catch these errors.

mthrok: hmm, `assert_grad` does the conversion to `float64`.
Looking at [the log](https://app.circleci.com/pipelines/github/pytorch/audio/5624/workflows/68c709a1-9ffc-44c3-a0c2-c253ff475793/jobs/192386), the Tensors being compared have `float64` type and the first couple of elements match. So I think the error is coming from somewhere. `Griffin-Lim` is an iterative method so it could be errors accumulating. Maybe we can try with lower `n_iter` value, but if the `n_iter` value is different than practical use, then it defeats the purpose of having the test to ensure autograd support.

mthrok:
> Have incorrect jacobian when doing gradcheck.

Hi @yoyololicon 

Thanks for the PR. Maybe the reason why it is failing could be the use of `float()` in the middle of the `for` loop.

https://github.com/pytorch/audio/blob/90311ed11ff88042b8ea5e849a746aedb10a0f42/torchaudio/functional/functional.py#L196-L207

Could you modify the code so that all the intermediate variables have the same `dtype` as the input Tensor?

test/torchaudio_unittest/transforms/autograd_test_impl.py
mthrok: Could be caused by the `float()` in the middle of the `for` loop.

https://github.com/pytorch/audio/blob/90311ed11ff88042b8ea5e849a746aedb10a0f42/torchaudio/functional/functional.py#L196-L207

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -65,6 +65,14 @@ def test_melspectrogram(self):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
 
+    def test_griffinlim(self):
```
mthrok, line 68: Could you add a case where momentum is disabled, `momentum=0`?

The in-place multiplication is causing the test failure.

https://github.com/pytorch/audio/blob/90311ed11ff88042b8ea5e849a746aedb10a0f42/torchaudio/functional/functional.py#L227

test/torchaudio_unittest/transforms/autograd_test_impl.py
yoyololicon: Good catch! It's indeed the cause!

mthrok (approved):
Looks good. Thank you!

PR #1426, opened 2021-04-05
Remove an invalid resample comparison unit test

carolineechen:
unit test comparing librosa and torchaudio's numerical results for resampling was not correct -- resampling with a larger range of input data can result in numerical differences of greater than atol=1e-02 in the unit test, which is not enough to claim compatibility


mthrok (approved):
Looks good, thanks.

nit: I think the test itself is correctly written. It's just the threshold is not tight enough to claim the compatibility.

PR #1431, opened 2021-04-05
Add warning about complex tensor to spectrogram

mthrok:
Follow-up of #1366 .
cc. @anjali411 
part of #1337 

Replaced by #1445

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -86,11 +86,13 @@ def spectrogram(
     """
     if power is None and not return_complex:
         warnings.warn(
-            "spectrogram now supports returning native complex tensor "
+            "Spectrogram now supports returning native complex tensor "
             "when `power=None` by setting `return_complex=True`. "
-            "Currently, the function returns pseudo complex type (..., 2) by default, "
-            "but this will change in the future and `return_complex` would be set to True by default. "
-            "Please refer to https://github.com/pytorch/audio/issues/1337 for the detail.")
+            "Currently, this function returns pseudo complex type (..., 2) by default, "
+            "but this will change in the future and `return_complex` "
+            "will be set to True by default. Please refer to "
+            "https://github.com/pytorch/audio/issues/1337 "
+            "for more details about the migration plan.")
 
     if power is not None and return_complex:
         raise ValueError(
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -84,6 +84,14 @@ def spectrogram(
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is None and not return_complex:
+        warnings.warn(
+            "spectrogram now supports returning native complex tensor "
+            "when `power=None` by setting `return_complex=True`. "
+            "Currently, the function returns pseudo complex type (..., 2) by default, "
+            "but this will change in the future and `return_complex` would be set to True by default. "
+            "Please refer to https://github.com/pytorch/audio/issues/1337 for the detail.")
+
     if power is not None and return_complex:
         raise ValueError(
             'When `power` is provided, the return value is real-valued. '


torchaudio/functional/functional.py
```diff
@@ -84,6 +84,14 @@ def spectrogram(
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is None and not return_complex:
+        warnings.warn(
+            "spectrogram now supports returning native complex tensor "
```
anjali411, line 89: nit - `Spectrogram`

torchaudio/functional/functional.py
```diff
@@ -84,6 +84,14 @@ def spectrogram(
...
+            "when `power=None` by setting `return_complex=True`. "
+            "Currently, the function returns pseudo complex type (..., 2) by default, "
```
anjali411, line 91:
```suggestion
            "Currently, this function returns pseudo complex type (..., 2) by default, "
```

torchaudio/functional/functional.py
```diff
@@ -84,6 +84,14 @@ def spectrogram(
...
+            "but this will change in the future and `return_complex` would be set to True by default. "
```
anjali411, line 92:
```suggestion
            "but this will change in the future and `return_complex` would be set to `True` by default. "
```

torchaudio/functional/functional.py
```diff
@@ -84,6 +84,14 @@ def spectrogram(
...
+            "Please refer to https://github.com/pytorch/audio/issues/1337 for the detail.")
```
anjali411, line 93:
```suggestion
            "Please refer to https://github.com/pytorch/audio/issues/1337 for more details about the migration plan.")
```

torchaudio/functional/functional.py
```diff
@@ -84,6 +84,14 @@ def spectrogram(
         ``n_fft // 2 + 1`` and ``n_fft`` is the number of
         Fourier bins, and time is the number of window hops (n_frame).
     """
+    if power is None and not return_complex:
+        warnings.warn(
```
anjali411, line 88: I think we should explicitly mention it's a deprecation warning: 

"Deprecation warning: In a future TorchAudio release `spectrogram` ",
"will no longer return tensors of pseudo complex type (..., 2) by default. ",
"Instead, `return_complex` will be set to `True` by default and spectrogram",
"would return native complex tensors by default."

anjali411: actually even better, we could use https://docs.python.org/3/library/exceptions.html#DeprecationWarning

`warnings.warn(msg, category=DeprecationWarning)` where `msg` equals
```
"In a future TorchAudio release, `spectrogram` ",
"will no longer return tensors of pseudo complex type (..., 2) by default. ",
"Instead, `return_complex` will be set to `True` by default and spectrogram",
"would return native complex tensors by default."
```

mthrok: @anjali411 
I do not have a preference here but, according to [this comment](https://github.com/pytorch/audio/pull/523#discussion_r405673957) by @gchanan, PyTorch uses `UserWarning` for deprecation, so in `torchaudio` we used `UserWarning`.

However, I see mixed usage in pytorch.
https://github.com/pytorch/pytorch/search?q=DeprecationWarning 
https://github.com/pytorch/pytorch/blob/c371542efc31b1abfe6f388042aa3ab0cef935f2/torch/hub.py#L512

mthrok: @anjali411 Do we want to deprecate the pseudo complex in the upcoming release?

The planned migrations phases are as following and I was thinking to deprecate the native complex type in phase 2.
1. Add support for native complex type
2. Switch to native complex type by default
3. Remove pseudo complex type

mthrok: I gave some thoughts on this and now I think it makes more sense to give heads up on deprecation prior to defaulting native complex type.

anjali411 (approved):

mthrok:
See #1445

PR #1434, opened 2021-04-06
Do not print CPU info before running tests

mthrok:
This code was added as a part of investigation of inconsistent behavior
of file-like object support. Now the investigation is done and this code
does not provide much insight yet adds a bunch of lines to the log, so
this PR removes it.


carolineechen (approved):
lgtm!

PR #1435, opened 2021-04-06
refactors functional test

steveplazafb:
Merges lfilter and spectrogram classes together in the common implementation and modifies the cpu and gpu test definitions accordingly


mthrok:
Thanks!

PR #1437, opened 2021-04-06
Update InverseMelScale comparison test 

mthrok:
This PR replaces the existing test for InverseMelScale in `librosa_compatibliity_test` with a new test.

Similar to #1426, `test_InverseMelScale` in `librosa_compatibility_test` is not
ensuring the comaptibility to librosa. Having this test can give a wrong statement
about the librosa numerical compatibility of the function.

The new test compares the result of inverse mel scale against the reference spectrogram, so no need to use librosa.
This test serves as more like an insurance that the change to the implementation of `InverseMelScale` only improves the result, not the other way.


carolineechen (approved):

carolineechen (approved):
looks good!

PR #1438, opened 2021-04-07
Refactor `F.lfilter` and `F.*_biquad` autograd tests

yoyololicon:
Adopt the autograd test model from `transforms/autograd_test_impl.py`.
Also add more combinations of testing parameters.


mthrok (commented):
Thanks for the PR.

Although I like the thorough testing, it's hitting the CI job time out limit.
Can we reduce the time it requires for the test somehow?
Autograd tests are by nature expensive itself.

I think it's good enough to use only one value for `central_freq` and `Q`.
I could be wrong but it feels that `2`, `100`, `800`, `4000`, `8000` are in a safe range for gradient computation.
If it is around `0`, I would be more cautious. (but again, this is my intuition and not backed by anything. so if you disagree, please let me know)

Things like `const_skirt_gain` should be parameterized because that change the code execution path, and we want to cover all the path.

yoyololicon:
> I think it's good enough to use only one value for `central_freq` and `Q`.
> I could be wrong but it feels that `2`, `100`, `800`, `4000`, `8000` are in a safe range for gradient computation.
> If it is around `0`, I would be more cautious. (but again, this is my intuition and not backed by any hard fact. So if you disagree, please let me know)

Sounds reasonable, I agree.

mthrok:
Thanks!

PR #1436, opened 2021-04-06
Run functional librosa compatibility test on CUDA as well

mthrok:
This PR parameterizes the device placement of the librosa compatibility test for functional, so that the tests run on GPU as well.

PR #1439, opened 2021-04-07
Run transform librosa compatibility test on CUDA as well

mthrok:
Similar to #1436 this PR ports librosa compatibility test for Transforms to CPU/CUDA.


test/torchaudio_unittest/transforms/librosa_compatibility_cpu_test.py
```diff
@@ -0,0 +1,9 @@
+import torch
+
+from torchaudio_unittest.common_utils import PytorchTestCase
+from .librosa_compatibility_test_impl import TransformsTestBase
+
+
+class TestTransforms(TransformsTestBase, PytorchTestCase):
```
carolineechen, line 7: do we want to rename this `TestTransformsCPU` (and similarly for the CUDA test)? I see it being done both ways in existing unit tests ([with](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/functional/librosa_compatibility_cpu_test.py#L5), [without](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/functional/torchscript_consistency_cpu_test.py#L7)) but maybe this is something we want to keep consistent moving forward

carolineechen (approved):
left a minor comment/question, but otherwise lgtm

test/torchaudio_unittest/transforms/librosa_compatibility_cpu_test.py
mthrok: So I started out as very verbose test class name but in #1436, it was pointed out that I forgot to change it from CPU to CUDA when I copy pasted it, and that made me wonder if we need to be that verbose for this. CPU tests and CUDA tests are separated on file-basis, so there is not a case where you look at test log and do not know which device it is running the test.
So I am learning towards removing `CPU/CUDA` from the name of test class.
What do you think?

carolineechen: Yup, that makes sense to me. We can remove `CPU/CUDA` from test class names moving forward

PR #1444, opened 2021-04-09
Use torchaudio melscale 'slaney' instead of librosa in WaveRNN pipeline preprocessing

discort:


discort:
cc @mthrok @vincentqb

examples/pipeline_wavernn/main.py
```diff
@@ -270,11 +270,11 @@ def main(args):
 
     transforms = torch.nn.Sequential(
         torchaudio.transforms.Spectrogram(**melkwargs),
-        LinearToMel(
+        torchaudio.transforms.MelScale(
```
mthrok, line 273: If you have an extra bandwidth, can you add a test that verifies that `librosa.feature.melspectrogram(S=spectrogram, ...)` and `MelScale()(spectrogram)` produce the same result?

We do have a test for `MelSpectrogram` here

https://github.com/pytorch/audio/blob/35d68fddc756b0717bfcba6048fc8e2e95efb465/test/torchaudio_unittest/transforms/librosa_compatibility_test_impl.py#L62-L87

It will be nice to add one for `MelScale` as well.

mthrok: Talking about `MelSpectrogram`, how about using `MelSpectrogram` directly?

discort: Makes sense.
Fixed it and checked manually.

mthrok:
@vincentqb Review please.

vincentqb:
LGTM, but we need to make sure the convergence of the model has not been affected. @discort -- have you reran the training loop to see? I'm rerunning the model on my side to see if the convergence profile changed.

discort:
thanks for getting back to me @vincentqb 
Unfortunately I don't have resources to check the convergency. Let me know if there are any problems and next week I'll try to allocate some staff to check it manually.

vincentqb:
Alright, LGTM, thanks!

vincentqb (approved):
Thanks!

PR #1446, opened 2021-04-09
Save/load TorchScript object in test

mthrok:
This PR adds step to save/load TorchScript object while testing TorchScript compatibility.

Part of #1337

### DIFF
--- a/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
@@ -6,17 +6,21 @@ import torchaudio.functional as F
 from parameterized import parameterized
 
 from torchaudio_unittest import common_utils
+from torchaudio_unittest.common_utils import TempDirMixin, TestBaseMixin
 from torchaudio_unittest.common_utils import (
     skipIfRocm,
 )
 
 
-class Functional(common_utils.TestBaseMixin):
+class Functional(TempDirMixin, TestBaseMixin):
     """Implements test for `functinoal` modul that are performed for different devices"""
     def _assert_consistency(self, func, tensor, shape_only=False):
         tensor = tensor.to(device=self.device, dtype=self.dtype)
 
-        ts_func = torch.jit.script(func)
+        path = self.get_temp_path('func.zip')
+        torch.jit.script(func).save(path)
+        ts_func = torch.jit.load(path)
+
         output = func(tensor)
         ts_output = ts_func(tensor)
         if shape_only:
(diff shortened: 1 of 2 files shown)


anjali411:
@SplitInfinity can you take a look?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -6,17 +6,21 @@
...
 )
 
 
-class Functional(common_utils.TestBaseMixin):
+class Functional(TempDirMixin, TestBaseMixin):
     """Implements test for `functinoal` modul that are performed for different devices"""
     def _assert_consistency(self, func, tensor, shape_only=False):
         tensor = tensor.to(device=self.device, dtype=self.dtype)
 
-        ts_func = torch.jit.script(func)
```
anjali411, line 19: we shouldn't remove this and continue checking `scripted_fn(input) == fn(input)`

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -6,17 +6,21 @@
...
+        path = self.get_temp_path('func.zip')
+        torch.jit.script(func).save(path)
+        ts_func = torch.jit.load(path)
```
anjali411, line 22: This returns a scripted module whose forward is the function you saved.

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
SplitInfinity: I agree with the above. To clarify, `torch.jit.load` returns a scripted module with a `forward` method equivalent to the function that was serialized. The caller does not notice the difference because `__call__` for the module calls `forward`.

mthrok: Hmm, I have a mixed feeling about introducing another helper function.

First, this method **IS** the helper function used by the actual tests.
Secondly, my view is that if it's something that needs explanation with comment, keeping it along side of tests make it easier to follow the logic of tests.

anjali411: At the moment, you duplicate the same code logic at different places, which I think strongly suggests that we should just have one helper function that can be used in all the tests.

anjali411 (approved):
This PR looks good! I still think we should add a helper function to avoid replicating the same code logic at different sites, but won't block this PR just for that.

mthrok:
@anjali411 Thanks for the suggestion. Let me move on with the changes for now. The repeated code part, I have a different idea to resolve it.

PR #1447, opened 2021-04-11
Add autograd test for T.AmplitudeToDB

krishnakalyan3:
Ref #1414
cc @mthrok

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -124,8 +124,9 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
 
-    def test_amplitude_to_db(self):
+    @parameterized.expand([(None, ), (80, )])
+    def test_amplitude_to_db(self, top_db):
         sample_rate = 8000
-        transform = T.AmplitudeToDB()
+        transform = T.AmplitudeToDB(top_db=top_db)
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -123,3 +123,9 @@ def test_spectral_centroid(self):
         transform = T.SpectralCentroid(sample_rate=sample_rate)
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform], nondet_tol=1e-10)
+
+    def test_amplitude_to_db(self):
+        sample_rate = 8000
+        transform = T.AmplitudeToDB()
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
+        self.assert_grad(transform, [waveform], nondet_tol=1e-10)
```
mthrok, line 131: Did you try it without `nonfat_tol`?
With `top_db=None`, it's just a combination of `log10`, `clamp` and multiplications.
If any of them are non deterministic, would be nice to add a comment about it here.

Also can you parameterize `top_db`? Say, `None` and `80` as recommended by the doc.

https://github.com/pytorch/audio/blob/78c3480e3b86b971cc834f1d3cec162408ba1ba0/torchaudio/functional/functional.py#L280

mthrok (commented):
Hi @krishnakalyan3 

Thanks for working on this. Please refer to my comment.

test/torchaudio_unittest/transforms/autograd_test_impl.py
mthrok: @krishnakalyan3 

My bad, I realized that when `top_db` is provided the function uses `torch.amax`, which is not differentiable. I think we should only test for the case `top_db=None`.

mthrok (approved):
Looks good. Thanks!

PR #1450, opened 2021-04-12
Replace existing RNN-T prototype

carolineechen:
Replace existing prototype RNN-T and ensure cmake builds

Follow ups
- determine if we want to expose `compute_rnnt_alphas` and `compute_rnnt_betas`
- determine more consistent naming folder and variable naming structure (`transducer` --> `rnnt`?, ensure consistent c++ and python variable names)
- improve python test readability (`data` structure, better test descriptions), partially after GPU migration
- c++ code refactoring (split up `.h` files), after GPU migration
- migrate input checks from python to c++
- ~~update documentation~~
- ~~resolve cmake build warnings~~

cc Differential Revision: D27623004


CMakeLists.txt
```diff
@@ -56,7 +56,6 @@ endif()
 # Options
 option(BUILD_SOX "Build libsox statically" OFF)
 option(BUILD_KALDI "Build kaldi statically" ON)
-option(BUILD_TRANSDUCER "Enable transducer" OFF)
```
mthrok, line 59: I think we can keep this but default to `ON`. Keeping this allows faster build time with `BUILD_TRANSDUCER=OFF`. Thoughts?

vincentqb: Indeed, given that this is still a prototype, let's keep the option in, and leave the switch of default to later.

vincentqb:
you can also remove the submodule in this PR

torchaudio/__init__.py
```diff
@@ -21,6 +23,8 @@
 except ImportError:
     pass
 
+logging.getLogger().setLevel(logging.INFO)
```
mthrok, line 26: Is any of the Python code checked in use `logging`?
I might be missing, but I do not see any usage.
If it is not used, we can remove this.

mthrok:
Looks like test passed.

https://app.circleci.com/pipelines/github/pytorch/audio/5730/workflows/c131311b-12d9-4bc7-9f0a-640311cf797a/jobs/198018
```
torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_basic_backward PASSED [ 57%]
torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_basic_fp16_no_error PASSED [ 57%]
torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_costs_and_gradients_B1_T2_U3_D5_fp32_cpu PASSED [ 57%]
torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_costs_and_gradients_B2_T4_U3_D3_fp32_cpu PASSED [ 57%]
torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_costs_and_gradients_random_data_with_numpy_fp32_cpu PASSED [ 57%]
```

I see that there is a warnings issued from the new code. I think this is something needs to be addressed (among others) before moving the RNNT loss out of prototype.

```
test/torchaudio_unittest/prototype/transducer_test.py::CPUTransducerTester::test_costs_and_gradients_random_data_with_numpy_fp32_cpu
  /root/project/test/torchaudio_unittest/prototype/numpy_transducer.py:29: UserWarning: Legacy tensor constructor is deprecated. Use: torch.tensor(...) for creating tensors from tensor-like objects; or torch.empty(...) for creating an uninitialized tensor with specific sizes. (Triggered internally at  /opt/conda/conda-bld/pytorch_1618211099535/work/torch/csrc/utils/tensor_new.cpp:477.)
    costs = torch.FloatTensor(costs).to(device=device)
```

The use of `FloatTensor` makes me wonder, does RNNT-loss handle `float64` (and  optionally `float16`)?

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,289 @@
...
-
-    The implementation uses `warp-transducer <https://github.com/HawkAaron/warp-transducer>`__.
-
-    Args:
-        blank (int): blank label. (Default: ``0``)
-        reduction (string): If ``'sum'``, the output losses will be summed.
-            If ``'mean'``, the output losses will be divided by the target lengths and
-            then the mean over the batch is taken. If ``'none'``, no reduction will be applied.
-            (Default: ``'mean'``)
+class TransducerLoss(torch.nn.Module):
```
mthrok, line 146: Does this class also handle loss other than RNNT-Loss?
If not, can we name this more descriptive, like RNNTLoss? 

There are different kind of transducers. We cannot predict the future perfectly, but just in case we add other transducers, please name the class / functions more descriptive.

docs/source/transducer.rst
```diff
@@ -8,16 +8,11 @@ torchaudio.prototype.transducer
 
 .. note::
 
-    The RNN transducer loss is a prototype feature, see `here <https://pytorch.org/audio>`_ to learn more about the nomenclature. It is only available within the nightlies, and also needs to be imported explicitly using: :code:`from torchaudio.prototype.transducer import rnnt_loss, RNNTLoss`.
+    The RNN transducer loss is a prototype feature, see `here <https://pytorch.org/audio>`_ to learn more about the nomenclature. It is only available within the nightlies, and also needs to be imported explicitly using: :code:`from torchaudio.prototype.transducer import TransducerLoss`.
```
vincentqb, line 11: let's keep the original names `rnnt_loss` and `RNNTLoss`, so i don't expect the doc to need to change

torchaudio/prototype/__init__.py
```diff
@@ -0,0 +1,11 @@
+from .transducer import (
+    compute_alphas,
+    compute_betas,
+    TransducerLoss,
+)
+
+__all__ = [
+    'compute_alphas',
+    'compute_betas',
```
vincentqb, line 9: we'll need feedback on whether we want to expose these two in python, please add this to the follows-up needed (in the description)

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,289 @@
...
+    src_lengths,
+    tgt_lengths,
+    blank=-1,
+    clamp=-1,
+):
+    """
+    Wrapper function to compute alphas for RNNT loss.
+    This can also be used as a backpointer table, to backtrack
+    {from (T-1, U-1) to (0,0)} to find the best cost path.
+    Also, enables easy unit-testing of alphas with numpy
```
vincentqb, line 16: for follow-up again: docstrings will need to be updated and standardize

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,289 @@
...
+        reuse_logits_for_grads=True,
+    ):
         """
-        See documentation for RNNTLoss.
+        logits: Tensor of (B, max_T, max_U, D) containing output from joiner
+        targets: Tensor of (B, max_U - 1) containing targets with zero padded
+        src_lengths: Tensor of (B) containing lengths of each sequence from encoder
+        tgt_lengths: Tensor of (B) containing lengths of targets for each sequence
+        fused_log_smax: [Optional] set to false if calling log_softmax outside loss
+        reuse_logits_for_grads: save memory by reusing logits memory for grads
```
vincentqb, line 88: please see how the docstring was organized

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,289 @@
...
-
-    Args:
-        blank (int): blank label. (Default: ``0``)
-        reduction (string): If ``'sum'``, the output losses will be summed.
-            If ``'mean'``, the output losses will be divided by the target lengths and
-            then the mean over the batch is taken. If ``'none'``, no reduction will be applied.
-            (Default: ``'mean'``)
+class TransducerLoss(torch.nn.Module):
+    """
+    Parameters:
```
vincentqb, line 148: nit: docstring standardized

docs/source/transducer.rst
mthrok: I am okay with those names, but please do not reason on the ease of doc update. Giving a name with proper level of descriptiveness is far more important than reducing the doc change.
If this turns out to be a bad name, we should change them and we should do so before moving them out of prototype.

carolineechen:
per offline/internal discussion w/ @vincentqb, additionally address the following points before merging
- D27625959: Add comment to `transducer_test.py` referring to source for reference
- D27689359: Rename `test/rnn_transducer` to `test/transducer` to keep c++ naming consistent (for beta: ensure c++ and python naming consistent)
- D27689359: Remove `__init__` in prototype -- we do not import function from prototype as we want users to import implicitly to use at this stage

torchaudio/prototype/transducer.py
```diff
@@ -1,19 +1,26 @@
...
     clamp=-1,
 ):
-    """
-    Wrapper function to compute alphas for RNNT loss.
-    This can also be used as a backpointer table, to backtrack
-    {from (T-1, U-1) to (0,0)} to find the best cost path.
+    """Wrapper function to compute alphas for RNNT loss.
+
+    This can also be used as a backpointer table, to backtrack {from (T-1, U-1) to (0,0)} to find the best cost path.
     Also, enables easy unit-testing of alphas with numpy
```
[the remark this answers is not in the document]
carolineechen: this was resolved in the following commit [here](https://github.com/pytorch/audio/pull/1450/commits/de2a6e2bc004d7fbf1c6bd5777130f6125123c4f#diff-9623f973feeae59fa7215b73ad28d00f83c59bfa00f0743c096e4ab36fe5e25d)

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,359 @@
...
+
+def check_device(var1, name1, var2, name2):
+    if var1.device != var2.device:
+        raise ValueError(
+            "`{}` ({}) must be on the same ".format(name1, var1.device.type)
+            + "device as `{}` ({})".format(name2, var2.device.type)
+        )
+
+
+def check_inputs(logits, targets, src_lengths, tgt_lengths, blank):
```
vincentqb, line 306: please add to follow-ups that we will move those to c++

torchaudio/prototype/transducer.py
```diff
@@ -1,119 +1,359 @@
...
-        act_lens (Tensor): Tensor of dimension (batch) containing the length of each output sequence
-        label_lens (Tensor): Tensor of dimension (batch) containing the length of each output sequence
-        blank (int): blank label. (Default: ``0``)
-        reduction (string): If ``'sum'``, the output losses will be summed.
-            If ``'mean'``, the output losses will be divided by the target lengths and
-            then the mean over the batch is taken. If ``'none'``, no reduction will be applied.
-            (Default: ``'mean'``)
+        logits (Tensor): Tensor of (B, max_T, max_U, D) containing output from joiner
+        targets (Tensor): Tensor of (B, max_U - 1) containing targets with zero padded
+        src_lengths (Tensor): Tensor of (B) containing lengths of each sequence from encoder
```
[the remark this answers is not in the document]
vincentqb: let's make sure the python code is consistent with this. the C++ code ideally would use the same naming, but i'm leaving changing C++ as follow-up.

carolineechen:
After offline discussion, we have decided to hold off on merging RNN-T loss to master until we have both the CPU and GPU implementations ready to be merged.

Issue #1348, opened 2021-03-03
`F.vad` batch consistency behaviour


jcaw:
I isolated a potential issue with batch consistency and F.vad in #1341:

> [Describing the change] Tweak the VAD consistency tests by using longer samples for the random test, and using lower thresholds for the file test. **This now causes many of the VAD tests to fail.**

> This has me thinking - if the purpose of `F.vad` is to outright crop the input tensor (not, for example, zero it), shouldn't we expect it to produce different results on a sample vs. a batch? I believe my tweaks just contain the right conditions for that to happen, but I might be misreading. If this is the case, should `F.vad` be removed from the batch consistency tests?

@mthrok replied:

> I see. I did not review the design of `F.vad` / `T.Vad` in detail, but your finding sounds like that the design of `T.Vad` is ill-formed. (I thought that `F.vad` would zero-out the non-vocal part). To me it does not make sense to batch the results that could be cropped in any way. We should check the behavior of the `sox` command, but @astaff do you have an idea of how it is supposed to behave?

Opening a dedicated issue so the discussion can continue here.

astaff:
Thanks for bringing this up, @jcaw. Apologies for a delayed response.

Current vad implementation is based on the behavior of sox and "Attempts to trim silence and quiet background sounds..." (http://sox.sourceforge.net/sox.html). Sox produces a truncated waveform, not the same-sized filled with zeros.

Since sox implementation of vad has multi-channel support, batched input is treated as multiple channels, and the entire batch is truncated to the earliest start of voice activity in any channel. 

I see how the current behavior can be confusing and might require a more thorough design. For now we can disable batch consistency testing for vad and update the documentation explaining current behavior.

Thoughts, @moto and @vincentqb ?

mthrok:
@astaff 

Thanks for the explanation.

I do not think it make sense to treat channels and samples same here. For channels, we can make some strong assumptions reasonably, such as, audio tracks in different channels share the same time axis, and they record the same event, which can justify the implementation of `F.vad`. For samples, such assumption does not apply. Especially when training a model and randomly shuffling audio files from various recording situations.

So wrapping `F.vad` in `T.VAD` just because other functionals for filtering do so does not make sense. If we want to have VAD for batch processing, we need to think of an expected usage before we start designing it. VAD itself is a big thing in audio application domain and active research/production. There are novel NN approaches published for VADs.

[...]

PR #1451, opened 2021-04-13
Remove VAD from batch consistency tests

### LINKED ISSUE #1449: Remove T.VAD from transform batch consistency test
See https://github.com/pytorch/audio/issues/1348 for the background

`T.VAD` currently treats all the tracks in the input batch as belong to the same audio data. This means that it returns different results when samples are batched and samples are fed independently.

There is a test to ensure the consistency of operation result regardless of batched/un-bached, but `T.VAD` in this test does not make sense. We would like to remove it.

https://github.com/pytorch/audio/blob/6a677ac8bfa2fd97f6d30a1ec0308ded98ba8020/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py#L92-L95

# Steps

1. Remove the test

# Testing and building

Please refer to [`CONTRIBUTING.md`](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md).

jcaw:
As mentioned in #1348 and #1449, the VAD function trims the input tensor to the first instance of voice activity on any channel or item. Trimming batches this way may be undesirable as the item with earliest activity will dominate, but either way, the batch behaviour does not match the itemwise behaviour.

The functional VAD batch consistency tests currently pass out of coincidence, but they specify incorrect behaviour. This PR removes these tests.

Closes #1449

### DIFF
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -197,22 +197,6 @@ class TestFunctional(common_utils.TorchaudioTestCase):
             F.sliding_window_cmn, spectrogram, center=center,
             norm_vars=norm_vars)
 
-    def test_vad_from_file(self):
-        filepath = common_utils.get_asset_path("vad-go-stereo-44100.wav")
-        waveform, sample_rate = common_utils.load_wav(filepath)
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Looks good. Thanks!

Issue #1456, opened 2021-04-15
Clarify `sox_effects.apply_effects_tensor` as  CPU-only


mthrok:
## 📚 Documentation

`torchaudio.sox_effects.apply_effects_tensor` only works on CPU tensor and it should be mentioned in the doc.

It can be clarified in here https://github.com/pytorch/audio/blob/ad534c1139d61aa9d9f05cb311fd64e6c4708f45/torchaudio/sox_effects/sox_effects.py#L71

PR #1459, opened 2021-04-18
Clarify `sox_effects.apply_effects_tensor` as CPU-only in Doc

krishnakalyan3:
#1456

### DIFF
--- a/torchaudio/sox_effects/sox_effects.py
+++ b/torchaudio/sox_effects/sox_effects.py
@@ -14,6 +14,7 @@ def init_sox_effects():
 
     Note:
         You do not need to call this function manually. It is called automatically.
+        This function only works on CPU Tensors.
 
     Once initialized, you do not need to call this function again across the multiple uses of
     sox effects though it is safe to do so as long as :func:`shutdown_sox_effects` is not called yet.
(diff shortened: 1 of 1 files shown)


mthrok (commented):
Hi @krishnakalyan3 

The change looks good. Can you also add `Note:` section to clarify it?

Say

```
    Note:
        This function only works on CPU Tensors.
```

Like in 

https://github.com/pytorch/audio/blob/14dd917ec60fa69ce3f7c6e3f2eaf520e67928b5/torchaudio/sox_effects/sox_effects.py#L14-L17

torchaudio/sox_effects/sox_effects.py
```diff
@@ -14,6 +14,7 @@ def init_sox_effects():
 
     Note:
         You do not need to call this function manually. It is called automatically.
+        This function only works on CPU Tensors.
```
mthrok, line 17: This is not the right function. The note has to be added to `apply_effects_tensor` function.

mthrok (changes requested):
Hi @krishnakalyan3 

Thanks for the update, but the note is being added to a different function.

torchaudio/sox_effects/sox_effects.py
krishnakalyan3: Ah got a bit confused. Should have read your comments carefully.

mthrok (approved):

PR #1463, opened 2021-04-19
Refactor functional test

dhthompson:
- Refactor functional test logic into one place, `functional_impl.py`
- Tidy imports

Local test plan:
`cd test && pytest torchaudio_unittest/functional`


mthrok (commented):
Looks good. Let's wait until the tests in CI pass.

For a follow up we can run these tests (except `create_fb_matrix`, which does not depend on device /dtypes) in CPU/CUDA by setting the device and dtype of Tensors passed to the tested functions.

mthrok:
Thanks!

PR #1473, opened 2021-04-23
Functional tests generalized to GPU and CPU

msaroufim:


msaroufim:
@mthrok this passes the flake8 issues. Had to create a new PR since my updated commits weren't being reflected in the old PR

May be helpful to open up a separate task for flake8 standardization since many files don't pass the linter right now with the default flake8 settings

mthrok:
Hi @msaroufim 

The PR looks good. We just fixed Windows issue in #1474, can you incorporate the change?

Regarding the linting, indeed the torchaudio's configuration is different from the default one, but that's the way repo was set up, so I think it's better kept that way. Also the change about the linting always include certain degree of subjectivity, which will allow anyone to object from any angle, and often times, which leads the conversation nowhere, so I think it's better not to touch it. Even applying `black` is not straightforward because the publicly available version and fb internal version are different, and later internal tool will try to format differently.
As long as flake8 passes, the style is somewhat in a good side.

PR #1474, opened 2021-04-23
Move windows GPU images to previous tag

malfet:
Should be moved back to stable once
https://github.com/pytorch/pytorch/issues/56654 is closed


mthrok (approved):
Thanks!

PR #1481, opened 2021-04-30
Ensure axis masking operations are not in-place

### LINKED ISSUE #1478: torchaudio.transforms.TimeMasking is inplace
## 🐛 Bug

The function `torchaudio.transforms.TimeMasking` transforms the Tensor inline without expectation.

## To Reproduce

```python
import librosa
import torchaudio
import torch
fmask = torchaudio.transforms.TimeMasking(time_mask_param=5)

BN = 2
max_frames = 200
data = torch.rand((BN, max_frames*160+240))
mt = torchaudio.transforms.MelSpectrogram(sample_rate=16000, n_fft=512, win_length=400, hop_length=160,
                                                 window_fn=torch.hamming_window, n_mels=64)
spec = mt(data)

def plot_spectrogram(spec, title=None, ylabel='freq_bin', aspect='auto', xmax=None):
    fig, axs = plt.subplots(1, 1)
    axs.set_title(title or 'Spectrogram (db)')
    axs.set_ylabel(ylabel)
    axs.set_xlabel('frame')
    im = axs.imshow(librosa.power_to_db(spec), origin='lower', aspect=aspect)
    if xmax:
        axs.set_xlim((0, xmax))
        fig.colorbar(im, ax=axs)
        plt.show(block=False)

spec = mt(data)

fmask(spec) # expect `mask_spec = fmask(spec)` or `fmask(spec, inline=False)`
print((spec==mt(data)).all()) # False (but  True is expected)
```

## Expected behavior

expect `mask_spec = fmask(spec)` or `fmask(spec, inline=False)`

## Environment
```
PyTorch version: 1.8.1+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2

OS: Ubuntu 16.04.7 LTS (x86_64)
GCC version: (Ubuntu 5.4.0-6ubuntu1~16.04.12) 5.4.0 20160609
Clang version: Could not collect
CMake version: version 3.5.1

Python version: 3.8 (64-bit runtime)
Is CUDA available: True

[pip3] numpy==1.20.2
[pip3] pytorch-metric-learning==0.9.97
[pip3] torch==1.8.1
[pip3] torch-scatter==2.0.6
[pip3] torchaudio==0.8.1
[pip3] torchlars==0.1.2
[pip3] torchvision==0.7.0
[conda] blas                      1.0                         mkl    defaults
[conda] cudatoolkit               10.1.243             h6bb024c_0    defaults
[conda] mkl                       2020.2                      256    defaults
[conda] mkl-service               2.3.0            py38he904b0f_0    defaults
[conda] mkl_fft                   1.2.0            py38h23d657b_0    defaults
[conda] mkl_random                1.1.1            py38h0573a6f_0    defaults
[conda] numpy                     1.19.2           py38h54aff64_0    defaults
[conda] numpy-base                1.19.2           py38hfa32c7d_0    defaults
[conda] pytorch                   1.6.0           py3.8_cuda10.1.243_cudnn7.6.3_0    pytorch
[conda] pytorch-metric-learning   0.9.97                   pypi_0    pypi
[conda] torch-scatter             2.0.6                    pypi_0    pypi
[conda] torchaudio                0.6.0                      py38    pytorch
[conda] torchlars                 0.1.2                    pypi_0    pypi
[conda] torchvision               0.7.0                py38_cu101    pytorch
```

carolineechen:
As reported in #1478, `TimeMasking` performs in-place operations that modifies the input tensor. This is also the case for `FrequencyMasking`. 

This PR modifies the algorithm to not perform in-place operations on the input tensor and adds checks in the unit test to ensure that the input tensor is not being changed.

(Resolves #1478)

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -200,6 +200,7 @@ class Functional(TestBaseMixin):
     def test_mask_along_axis(self, shape, mask_param, mask_value, axis):
         torch.random.manual_seed(42)
         specgram = torch.randn(*shape, dtype=self.dtype, device=self.device)
+        specgram_copy = specgram.clone()
         mask_specgram = F.mask_along_axis(specgram, mask_param, mask_value, axis)
 
         other_axis = 1 if axis == 2 else 2
@@ -211,11 +212,13 @@ class Functional(TestBaseMixin):
 
         assert mask_specgram.size() == specgram.size()
         assert num_masked_columns < mask_param
+        self.assertEqual(specgram, specgram_copy)
 
     @parameterized.expand(list(itertools.product([100], [0., 30.], [2, 3])))
     def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
         torch.random.manual_seed(42)
         specgrams = torch.randn(4, 2, 1025, 400, dtype=self.dtype, device=self.device)
+        specgrams_copy = specgrams.clone()
 
         mask_specgrams = F.mask_along_axis_iid(specgrams, mask_param, mask_value, axis)
 
@@ -226,6 +229,7 @@ class Functional(TestBaseMixin):
 
         assert mask_specgrams.size() == specgrams.size()
         assert (num_masked_columns < mask_param).sum() == num_masked_columns.numel()
+        self.assertEqual(specgrams, specgrams_copy)
 
 
 class FunctionalComplex(TestBaseMixin):
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -233,27 +233,31 @@ class Functional(TestBaseMixin):
     def test_mask_along_axis_preserve(self, shape, mask_param, mask_value, axis):
         """mask_along_axis should not alter original input Tensor
 
-        https://github.com/pytorch/audio/issues/1478
+        Test is run 5 times to bound the probability of no masking occurring to 1e-10
+        See https://github.com/pytorch/audio/issues/1478
         """
         torch.random.manual_seed(42)
-        specgram = torch.randn(*shape, dtype=self.dtype, device=self.device)
-        specgram_copy = specgram.clone()
-        F.mask_along_axis(specgram, mask_param, mask_value, axis)
+        for _ in range(5):
+            specgram = torch.randn(*shape, dtype=self.dtype, device=self.device)
+            specgram_copy = specgram.clone()
+            F.mask_along_axis(specgram, mask_param, mask_value, axis)
 
-        self.assertEqual(specgram, specgram_copy)
+            self.assertEqual(specgram, specgram_copy)
 
     @parameterized.expand(list(itertools.product([100], [0., 30.], [2, 3])))
     def test_mask_along_axis_iid_preserve(self, mask_param, mask_value, axis):
         """mask_along_axis_iid should not alter original input Tensor
 
-        https://github.com/pytorch/audio/issues/1478
+        Test is run 5 times to bound the probability of no masking occurring to 1e-10
+        See https://github.com/pytorch/audio/issues/1478
         """
         torch.random.manual_seed(42)
-        specgrams = torch.randn(4, 2, 1025, 400, dtype=self.dtype, device=self.device)
-        specgrams_copy = specgrams.clone()
-        F.mask_along_axis_iid(specgrams, mask_param, mask_value, axis)
+        for _ in range(5):
+            specgrams = torch.randn(4, 2, 1025, 400, dtype=self.dtype, device=self.device)
+            specgrams_copy = specgrams.clone()
+            F.mask_along_axis_iid(specgrams, mask_param, mask_value, axis)
 
-        self.assertEqual(specgrams, specgrams_copy)
+            self.assertEqual(specgrams, specgrams_copy)
 
 
 class FunctionalComplex(TestBaseMixin):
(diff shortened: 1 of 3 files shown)


vincentqb:
I agree the functional should not modify in place. However, it is possible someone in the wild was relying on that. This PR is changing the behavior of the function silently. I do not see a way of detecting users who would have relied on that feature though. Should we raise a warning for one release to warn of this behavior? Or would calling this a bug fix in the release notes be enough?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -211,11 +212,13 @@ def test_mask_along_axis(self, shape, mask_param, mask_value, axis):
 
         assert mask_specgram.size() == specgram.size()
         assert num_masked_columns < mask_param
+        self.assertEqual(specgram, specgram_copy)
```
vincentqb, line 215: the intention here is to specifically test that the original is not changed. at a minimum, i'd expect a comment justifying why this is here.  however, i would usually expect a separate test for this, so we know that this was the intention of the test. having a separate test would also make it easier to extend to other functionals if we want to.

thoughts?

mthrok: I agree with this point. Each unit test should be testing only one thing, so that when a unit test fails we know what aspect of the functionality is not met.

Can you make new test methods with short description and reference the issue number in docstring.
See examples of test descriptions like

https://github.com/pytorch/audio/blob/0c263a935f178490bda78c046e220e84b095447d/test/torchaudio_unittest/backend/sox_io/save_test.py#L381-L382

https://github.com/pytorch/audio/blob/0c263a935f178490bda78c046e220e84b095447d/test/torchaudio_unittest/backend/sox_io/load_test.py#L284-L293

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -227,6 +227,34 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
 
+    @parameterized.expand(
+        list(itertools.product([(2, 1025, 400), (1, 201, 100)], [100], [0., 30.], [1, 2]))
+    )
+    def test_mask_along_axis_preserve(self, shape, mask_param, mask_value, axis):
+        """mask_along_axis should not alter original input Tensor
+
+        https://github.com/pytorch/audio/issues/1478
+        """
+        torch.random.manual_seed(42)
```
mthrok, line 238: This is a tricky one. The goal of this test is to ensure that "when mask is randomly applied, the original Tensor is not altered", and in #1478, we learned that bug happens stochastically. So we need to control the randomness in a way that it always hits the condition that caused the issue in #1478.

Simply setting the random seed has a positive and negative effects here. The positive aspect is that the test is repeatable. Assuming that the environment (hardware / software (including everything from OS, PyTorch and CUDA) is the same, we can repeat the test and expect it to produce the same result. But the negative aspect is, we cannot be sure if this specific seed value applies to any configuration (HW/SW) in future? We do not know and it's not likely. If something about the random generator has been changed in the future, and if this seed value stopped hitting the condition, then the test becomes moot.

There are couple of approaches to overcome this.
1. Set the test configuration to always hit the correct condition of the reported bug.
For example, if we make the test to mask all the elements of the input tensor, we can be sure that the test meets the requirement. However, this diverges from the expected usages (should be masking a part of the input, not all), and looking at [the signature/docstring of function](https://github.com/pytorch/audio/blob/0c263a935f178490bda78c046e220e84b095447d/torchaudio/functional/functional.py#L713-L731) tested, it's not straightforward to do so. (It could be, but the docstring is hard to understand.)
2. Patch random generator for the sake of testing.
If there is no other solution, we can patch the random generator to change the behavior in our favor. This kind of technique is often used in function that relies on external resource (like HTTP access, for example [`moto`](https://github.com/spulec/moto) makes it possible to test your AWS app without internet connection). But this complicates the test logic, which increases the chance of writing a wrong test.
3. Bound the probability of this test not hitting the condition.
Another approach is to bound the probability that the test hits the correct condition. Say that this one attempt will hit the bug condition with probability `p`. If you repeat the procedure `n` times, assuming that the implementation of pseudo random generator meets [iid](https://en.wikipedia.org/wiki/Independent_and_identically_distributed_random_variables), (which I think is a reasonable assumption for this context even though it is not in strict sense) then the probability that the test hits the condition at least once becomes `1 - (1 - p)^n`. For `p=0.6` and `n=10`, we get something like 99.989 % of hitting the bug condition. The good thing about this approach is that we can still set the seed value once (and only once at the very beginning) and we expect the reproducibility.

carolineechen: I agree with everything here, and feel that option 3 makes the most sense to me. The success in #1478 (when the input tensor is not changed) takes place when `value = torch.rand(1) * mask_param` is equal to `0`, which results in `mask_start`=`mask_end` and no masking takes place. A `mask_param` value of 100 in all these tests indicates a 1% probability of `value=0` and not hitting the condition. I think therefore looping it over 5 times should be sufficient (> 99.9999%), what do you think? Also, should this reasoning be explained in the documentation or is linking the issue sufficient?

Additionally, I have run this test on the previous implementation and see that all these tests fail, indicating that we do hit the condition (although I do agree that the test as is does not ensure this is always true, if we end up changing parameters or if the state of the random seed changes)

mthrok: > Additionally, I have run this test on the previous implementation and see that all these tests fail, indicating that we do hit the condition 

Good to know. Thanks for taking the proper care.

> A `mask_param` value of 100 in all these tests indicates a 1% probability of `value=0` and not hitting the condition. I think therefore looping it over 5 times should be sufficient (> 99.9999%), what do you think?

Yes, 5 sounds good enough.

> Also, should this reasoning be explained in the documentation or is linking the issue sufficient?

Yes, explaining it in the docstring is more helpful for future maintainers. (which is how I learned this kind of technique in the past)

mthrok (approved):
Looks good. Thanks!

PR #1482, opened 2021-04-30
 Adding test for T.SlidingWindowCmn

pavithranrao:
Autograd tests for Transforms #1414

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -157,7 +157,6 @@ class AutogradTestMixin(TestBaseMixin):
     @parameterized.expand([
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
-        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
         ({'cmn_window': 500, 'min_cmn_window': 50, 'center': False, 'norm_vars': False}, ),
     ])
     def test_sliding_window_cmn(self, kwargs):
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -154,6 +154,19 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform])
 
+    @parameterized.expand([
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': True}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
+        ({'cmn_window': 500, 'min_cmn_window': 50, 'center': False, 'norm_vars': False}, ),
+    ])
+    def test_sliding_window_cmn(self, kwargs):
+        sample_rate = 8000
+        transform = T.SlidingWindowCmn(**kwargs)
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
+        self.assert_grad(transform, [waveform])
+
     @unittest.expectedFailure
     def test_timestretch_zeros_fail(self):
         """Test that ``T.TimeStretch`` fails gradcheck at 0
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -157,7 +157,6 @@ class AutogradTestMixin(TestBaseMixin):
     @parameterized.expand([
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
-        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': True}, ),
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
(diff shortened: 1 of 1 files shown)


pavithranrao:
@carolineechen Can you please review this?

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -154,6 +154,18 @@ def test_vol(self, gain, gain_type):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform])
 
+    @parameterized.expand([
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
```
carolineechen, line 160: this set of params is a duplicate of the first one -- could you change or remove it?

carolineechen (changes requested):

test/torchaudio_unittest/transforms/autograd_test_impl.py
pavithranrao: Thanks for catching that, will change it.

pavithranrao: Removed the duplicate test case. Thanks!

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -157,7 +157,6 @@ def test_vol(self, gain, gain_type):
     @parameterized.expand([
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
         ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
-        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': True}, ),
```
carolineechen: @mthrok this set of params causes the test to fail with `RuntimeError: Jacobian mismatch for output 0 with respect to input 0`. Any idea why?

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -154,6 +154,19 @@ def test_vol(self, gain, gain_type):
...
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': True}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': True}, ),
+        ({'cmn_window': 500, 'min_cmn_window': 50, 'center': False, 'norm_vars': False}, ),
+    ])
+    def test_sliding_window_cmn(self, kwargs):
+        sample_rate = 8000
+        transform = T.SlidingWindowCmn(**kwargs)
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
```
mthrok, line 167: The input to `SlidingWindowCmn` is supposed to be spectrogram.
This has been fixed in the master documentation https://pytorch.org/audio/master/functional.html#torchaudio.functional.sliding_window_cmn

Can you use `get_spectrogram`, then flip the last axis so that Tensor dimension is `[... time, freq]`?

test/torchaudio_unittest/transforms/autograd_test_impl.py
mthrok: This means that when `norm_vars=True`, some operation is not differential to the 2nd degree.

It's somewhere here but it is not immediately clear to me.

https://github.com/pytorch/audio/blob/0c263a935f178490bda78c046e220e84b095447d/torchaudio/functional/functional.py#L1073-L1100

What we want to do is
1. Identify which part is causing this.
2. Change the code if that is possible without performance degradation.

However, they are beyond the scope of this PR, so here, we can set `nondet_tol` and add docstring saying it's not 2nd-order differentiable when `norm_vars=True`, like in Spectrogram.

https://github.com/pytorch/audio/blob/0c263a935f178490bda78c046e220e84b095447d/test/torchaudio_unittest/transforms/autograd_test_impl.py#L71-L77

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -154,6 +154,19 @@ def test_vol(self, gain, gain_type):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform])
 
+    @parameterized.expand([
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': False}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': False, 'norm_vars': True}, ),
+        ({'cmn_window': 600, 'min_cmn_window': 100, 'center': True, 'norm_vars': True}, ),
+        ({'cmn_window': 500, 'min_cmn_window': 50, 'center': False, 'norm_vars': False}, ),
```
mthrok, line 162: `cmn_window =600` and `min_cmn_window=100` look too big for the input with `8000 * 0.05 == 400` (then FFT applied) can you make them somewhat smaller than the number of frames in time axis of the input tensor?

pavithranrao: thanks @mthrok will try to implement your suggestions

test/torchaudio_unittest/transforms/autograd_test_impl.py
pavithranrao: From the [doc](https://pytorch.org/docs/stable/generated/torch.stft.html)  of `torch.stft` I could find that it returns a tensor in the shape  (* × N × T) so do you suggest using `torch.transpose(-2, -1)` on the output?

pavithranrao:
unittest seems to pass with @mthrok 's suggestions

```
torchaudio_unittest\functional\batch_consistency_test.py::TestFunctional_0::test_sliding_window_cmn_False_False PASSED [ 48%]
torchaudio_unittest\functional\batch_consistency_test.py::TestFunctional_0::test_sliding_window_cmn_False_True PASSED [ 48%]
torchaudio_unittest\functional\batch_consistency_test.py::TestFunctional_0::test_sliding_window_cmn_True_False PASSED [ 48%]
torchaudio_unittest\functional\batch_consistency_test.py::TestFunctional_0::test_sliding_window_cmn_True_True PASSED [ 48%]
```

PR #1479, opened 2021-04-29
Replace existing prototype RNNT

carolineechen:
Summary: Replace the current torchaudio RNNT (using warp-transducer) to one without external library dependencies

Differential Revision: D28094572

Follow ups:
- check the status of the output in `cpu_transducer.h`
- improve `csrc` files (better `.h` `.cpp` organization, variable renaming)
- migrate input checks to C++

PR #1483, opened 2021-04-30
Add GPU RNNT Loss

carolineechen:
Summary: In #1479, we added support for CPU RNNT loss. This PR adds a GPU version of RNNT loss

Differential Revision: D28128853

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -38,6 +38,7 @@ _BUILD_SOX = False if platform.system() == 'Windows' else _get_build("BUILD_SOX"
 _BUILD_KALDI = False if platform.system() == 'Windows' else _get_build("BUILD_KALDI", True)
 _BUILD_TRANSDUCER = _get_build("BUILD_TRANSDUCER")
 _USE_ROCM = _get_build("USE_ROCM")
+_USE_CUDA = torch.cuda.is_available()
 
 
 def get_ext_modules():
@@ -76,6 +77,7 @@ class CMakeBuild(build_ext):
             "-DBUILD_TORCHAUDIO_PYTHON_EXTENSION:BOOL=ON",
             "-DBUILD_LIBTORCHAUDIO:BOOL=OFF",
             f"-DUSE_ROCM:BOOL={'ON' if _USE_ROCM else 'OFF'}",
+            f"-DUSE_CUDA:BOOL={'ON' if _USE_CUDA else 'OFF'}",
         ]
         build_args = [
             '--target', 'install'
--- a/test/torchaudio_unittest/rnnt/rnnt_loss_cuda_test.py
+++ b/test/torchaudio_unittest/rnnt/rnnt_loss_cuda_test.py
@@ -1,0 +1,10 @@
+import torch
+from .rnnt_loss_impl import RNNTLossTest
+from torchaudio_unittest import common_utils
+from .utils import skipIfNoTransducer
+
+
+@skipIfNoTransducer
+@common_utils.skipIfNoCuda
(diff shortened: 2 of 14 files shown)


build_tools/setup_helpers/extension.py
```diff
@@ -38,6 +38,7 @@ def _get_build(var, default=False):
 _BUILD_KALDI = False if platform.system() == 'Windows' else _get_build("BUILD_KALDI", True)
 _BUILD_TRANSDUCER = _get_build("BUILD_TRANSDUCER")
 _USE_ROCM = _get_build("USE_ROCM")
+_WITH_CUDA = True if torch.cuda.is_available() else False
```
mthrok, line 41: PyTorch core uses `USE_CUDA`. https://github.com/pytorch/pytorch/search?l=CMake&q=USE_CUDA 
So we should align with them.

carolineechen:
@malfet @seemethere Could I get your input on supporting custom CUDA kernel in torchaudio?

One of our features (RNN transducer loss) for our the upcoming release involves a custom CUDA kernel, but this isn't something that torchaudio currently supports. I have a minimal working CUDA build in this PR that compiles successfully on AWS and in the CircleCI builds (mac builds are failing on master for an unrelated reason), but I’m not very familiar with CUDA builds. Are there any additional flags that we should include (torchvision includes [these](https://github.com/pytorch/vision/blob/19ad0bbc5e26504a501b9be3f0345381d6ba1efc/CMakeLists.txt#L8)), or anything platform specific (ex/ for Windows) that we should add?

build_tools/setup_helpers/extension.py
```diff
@@ -38,6 +38,7 @@ def _get_build(var, default=False):
 _BUILD_KALDI = False if platform.system() == 'Windows' else _get_build("BUILD_KALDI", True)
 _BUILD_TRANSDUCER = _get_build("BUILD_TRANSDUCER")
 _USE_ROCM = _get_build("USE_ROCM")
+_USE_CUDA = True if torch.cuda.is_available() else False
```
malfet, line 41: Nit
```suggestion
_USE_CUDA = torch.cuda.is_available()
```

torchaudio/csrc/rnnt/gpu/compute_alphas.cu
```diff
@@ -0,0 +1,73 @@
+#include <THC/THC.h>
```
malfet, line 1: Do you really need to include THC?

malfet (approved):

torchaudio/csrc/rnnt/gpu/compute_alphas.cu
carolineechen: I can replace THC with something like `<c10/cuda/CUDAStream.h>` to get the same result -- is there a standard for supporting the `getCurrentCUDAStream` operator?

test/torchaudio_unittest/rnnt/rnnt_loss_cuda_test.py
```diff
@@ -0,0 +1,10 @@
+import torch
+from .rnnt_loss_impl import RNNTLossTest
+from torchaudio_unittest import common_utils
+from .utils import skipIfNoTransducer
+
+
+@skipIfNoTransducer
```
vincentqb, line 7: nit: just realized the name has not been standardized from Transducer to RNNT, but this can be changed after this PR

vincentqb (approved):
Awesome, LGTM!

torchaudio/csrc/rnnt/gpu/compute_alphas.cu
cpuhrsch: Including <c10/cuda/CUDAStream.h> is fine. THC is legacy and on its way out.

Issue #1487, opened 2021-05-05
Resampling


carolineechen:
We are revisiting the resampling feature, the way it is implemented and intended usage. Here are the improvements we plan to make:

#### Add `rolloff` as an adjustable parameter
- Currently a fixed value of 0.99 is used, but exposing the parameter can allow for control of the degree of anti-aliasing and highest frequency allowed when resampling.
#### Kaiser window support
- Currently, only sinc window is supported. Kaiser window will support an additional shape parameter (β) that controls the smoothness and impulse width of the filter, allowing for a wider range of resampling capabilities and signal quality.

## Improve `transforms.Resample` speed
- Improve speed of `transforms.Resample` by precomputing and caching the window kernel computation, which is the same for a given set of resampling parameters. After this change, we encourage users to use a single `transforms.Resample` class over `functional.resample` when calling resample multiple times with the same parameters, such as in cases of data loading.

#### warning for non-integer valued sampling rates
-  Sampling rates are currently being cast to ints during computation (see #891), as the resampling algorithm dependent on an integer ratio for the original / new frequency. We will issue a warning (error in the next release) and encourage users to manually convert their non-integer sampling rate parameters to integers that retain this same ratio rather than forcing integer inputs (ex/ using `orig_freq=8`, `new_freq=1` instead of `orig_freq=44100`, `new_freq=5512.5`).
- If there are situations requiring float sampling rates (that can not be manually converted to integers) that would benefit from approximate resampling and removing the int requirement, please leave a comment below so that we are aware of these use cases
#### Deprecate and remove `resample_waveform` from kaldi compliance
- `resample_waveform` in kaldi compliance simply calls `functional.resample`

## Progress Tracker
- [x] #1488 expose `rolloff` parameter
- [x] #1509 add Kaiser window support
- [x] #1499 #1514 cache `transforms.Resample` kernel for speedup
- [x] #1490 add warning for non-integer sampling rates
- [x] #1533 deprecating kaldi `resample_waveform`
- [x] (`0.10`) remove kaldi `resample_waveform`

cc @mthrok

mthrok:
cc @small-yellow-duck @faroit @adefossez  Please let us know if you have an opinion on resampling.

small-yellow-duck:
Earlier discussion here:
https://github.com/pytorch/audio/issues/908
- some discussion about how to test the module against librosa

Earlier PR here:
https://github.com/pytorch/audio/pull/911
- uses pytorch conv ops to do faster resampling
[...]

PR #1488, opened 2021-05-05
Add rolloff param to resampling

carolineechen:
Expose the `rolloff` parameter in resampling functions, which is can be used to control the degree of anti-aliasing and amount of high frequencies being filtered out in resampling.

cc #1487


mthrok:
Hi @adefossez 

We are revisiting resampling functions in `torchaudio`, and we appreciate if you can take a look at this PR. I remember you mention that `rolloff` affects aliasing for both downsampling and upsampling. I think this PR is harmless and uncontroversial, but extra eyes are nice.

mthrok (approved):

PR #1490, opened 2021-05-05
Add warning for non-integer resampling frequencies

carolineechen:
Add warning and suggestion when the frequency inputs to the resample function are not integer values

cc #1487

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -261,16 +261,17 @@ class Functional(TestBaseMixin):
 
     def test_resample_no_warning(self):
         sample_rate = 44100
-        waveform = get_whitenoise(sample_rate=sample_rate)
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.1)
 
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
-            F.resample(waveform, sample_rate, sample_rate / 2)
+            F.resample(waveform, float(sample_rate), sample_rate / 2.)
         assert len(w) == 0
 
     def test_resample_warning(self):
+        """resample should throw a warning if an input frequency is not of an integer value"""
         sample_rate = 44100
-        waveform = get_whitenoise(sample_rate=sample_rate)
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.1)
 
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -9,7 +9,7 @@ import torchaudio.functional as F
 from parameterized import parameterized
 from scipy import signal
 
-from torchaudio_unittest.common_utils import TestBaseMixin, get_sinusoid, nested_params
+from torchaudio_unittest.common_utils import TestBaseMixin, get_sinusoid, nested_params, get_whitenoise
 
 
 class Functional(TestBaseMixin):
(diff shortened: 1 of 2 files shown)


torchaudio/functional/functional.py
```diff
@@ -1372,6 +1372,16 @@ def resample(
...
     assert orig_freq > 0.0 and new_freq > 0.0
 
+    if not (int(orig_freq) == orig_freq and int(new_freq) == new_freq):
+        warnings.warn(
+            "Non-integer frequencies are cast to integers and may result in poor resampling quality."
+            "The algorithm is dependent on the integer ratio between `orig_freq` and `new_freq`,"
+            "so please manually convert frequencies to integer values that maintain the ratio"
+            "prior to passing them into the function"
+            "Example: To downsample a 44100 hz waveform by a factor of 8, you can use"
+            "`orig_freq=8` and `new_freq=1` instead of `orig_freq=44100` and `new_freq=5512.5`"
```
mthrok, line 1383: I think we can go farther and say it will raise an error in the future.
@cpuhrsch what do you think?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -227,6 +227,24 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
         assert mask_specgrams.size() == specgrams.size()
         assert (num_masked_columns < mask_param).sum() == num_masked_columns.numel()
 
+    def test_resample_no_warning(self):
+        sample_rate = 44100
+        waveform = get_whitenoise(sample_rate=sample_rate)
```
mthrok, line 264: nit: since we do not care about the resulting tensor, we can use something much smaller than a Tensor with 44100 elements, while keeping the value of `sample_rate = 44100`.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -227,6 +227,24 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            F.resample(waveform, sample_rate, sample_rate / 2)
```
mthrok, line 268: I think it's better to explicitly convert to float type here.

```suggestion
            F.resample(waveform, float(sample_rate), sample_rate / 2.)
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -227,6 +227,24 @@ def test_mask_along_axis_iid(self, mask_param, mask_value, axis):
...
+        assert len(w) == 0
+
+    def test_resample_warning(self):
```
mthrok, line 271: Can you add docstring stating what the expected behavior is?

mthrok (approved):

torchaudio/functional/functional.py
cpuhrsch: I think it's fine if we throw an exception in the next release, but I'd create a pre-emptive issue for this and link it here so that users can chime in whether they want this or not. The overall topic seems to be "non-integral sampling rates", so maybe we can link to an umbrella issue like that.

mthrok: @carolineechen 

Can you plan the future changes and make a note in #1487, then mention that in the message ?

mthrok (approved):
Looks good. Thanks!

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -259,6 +259,31 @@ def test_mask_along_axis_iid_preserve(self, mask_param, mask_value, axis):
...
+        sample_rate = 44100
+        waveform = get_whitenoise(sample_rate=sample_rate, duration=0.1)
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            F.resample(waveform, sample_rate, 5512.5)
+        assert len(w) == 1
+
+
+class FunctionalComplex(TestBaseMixin):
```
mthrok, line 282: @carolineechen I think this was introduced by mistake.

mthrok: Addressed in #1524

PR #1445, opened 2021-04-09
Add deprecation warnings to functions for complex

mthrok:
cc. @anjali411 
part of #1337

PR #1492, opened 2021-05-06
Add deprecation warnings to magphase and ComplexNorm

mthrok:
@anjali411 I noticed that there are missing deprecation warnings. The style of message is same as the other ones in #1445.


anjali411 (approved):

mthrok:
thanks!

PR #1494, opened 2021-05-06
Migrate RNNTL input checks to C++

carolineechen:
move runtime checks for RNNT loss from python to C++

cc #1240

### DIFF
--- a/torchaudio/csrc/rnnt/cpu/compute.cpp
+++ b/torchaudio/csrc/rnnt/cpu/compute.cpp
@@ -15,6 +15,62 @@ std::tuple<torch::Tensor, c10::optional<torch::Tensor>> compute(
     double clamp,
     bool fused_log_smax = true,
     bool reuse_logits_for_grads = true) {
+  TORCH_CHECK(
+      logits.device().type() == targets.device().type(),
+      "logits and targets must be on the same device");
+  TORCH_CHECK(
+      logits.device().type() == src_lengths.device().type(),
+      "logits and logit_lengths must be on the same device");
+  TORCH_CHECK(
+      logits.device().type() == tgt_lengths.device().type(),
+      "logits and target_lengths must be on the same device");
+
+  TORCH_CHECK(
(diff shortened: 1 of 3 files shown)


torchaudio/csrc/rnnt/cpu/compute.cpp
```diff
@@ -15,6 +15,62 @@ std::tuple<torch::Tensor, c10::optional<torch::Tensor>> compute(
...
+
+  TORCH_CHECK(logits.is_contiguous(), "logits must be contiguous");
+  TORCH_CHECK(targets.is_contiguous(), "targets must be contiguous");
+  TORCH_CHECK(src_lengths.is_contiguous(), "logit_lengths must be contiguous");
+  TORCH_CHECK(tgt_lengths.is_contiguous(), "target_lengths must be contiguous");
+
+  TORCH_CHECK(
+      logits.dim() == 4, "logits must be 4-D (batch, time, target, class)");
+  TORCH_CHECK(
+      targets.dim() == 2, "targets must be 2-D (batch, max target length)");
```
vincentqb, line 46: nit: since the "target" dimension must be the same:
```suggestion
  TORCH_CHECK(
      logits.dim() == 4, "logits must be 4-D (batch, time, max target length, class)");
  TORCH_CHECK(
      targets.dim() == 2, "targets must be 2-D (batch, max target length)");
```
same below

vincentqb (approved):
LGTM

torchaudio/csrc/rnnt/cpu/compute.cpp
carolineechen: the "target" dimensions are actually different -- we have targets.shape[1] = max target length, while logits.shape[2] = max target length + 1. I think we should leave this as is for now, since this matches the documentation in the python code [here](https://github.com/pytorch/audio/blob/master/torchaudio/prototype/rnnt_loss.py#L165), but we can create a follow up PR in both places to make it more clear if necessary.

vincentqb: oops! indeed, i agree with you :)

PR #1497, opened 2021-05-10
Add CUDA binary builds

malfet:


.circleci/config.yml
```diff
@@ -71,11 +71,14 @@ binary_common: &binary_common
     python_version:
       description: "Python version to build against (e.g., 3.8)"
       type: string
+    cuda_version:
```
seemethere, line 74: We should also probably add something to specify the correct docker image to use as well since I think that'll probably be an issue when attempting to build

seemethere: We should also try and move the conda builds off of `pytorch/conda-cuda` and onto `pytorch/conda-builder:${cu_version}`

PR #1498, opened 2021-05-11
Add autograd tests for TimeMasking/FrequencyMasking

kiri11:
Context: #1414

I combined these two transforms in one test because they're actually the same [`_AxisMasking` transform](https://github.com/pytorch/audio/blob/c2740644ee31d29861d3d9977a43af1e71779054/torchaudio/transforms.py#L856), but just on different axis. I even thought of testing the `_AxisMasking` directly, but decided against it because it's not part of the API.

Test fails without the `_DeterministicWrapper` likely due to random behaviour [here](https://github.com/pytorch/audio/blob/c2740644ee31d29861d3d9977a43af1e71779054/torchaudio/functional/functional.py#L792).

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -132,8 +132,15 @@ class AutogradTestMixin(TestBaseMixin):
         spectrogram = get_spectrogram(
             get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2),
             n_fft=n_fft, power=1)
+        spectrogram2 = get_spectrogram(
+            # Add second white noise with another seed, so they are different
+            get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2, seed=31),
+            n_fft=n_fft, power=1)
         deterministic_transform = _DeterministicWrapper(masking_transform(400))
+        deterministic_transform_iid = _DeterministicWrapper(masking_transform(400, True))
+
         self.assert_grad(deterministic_transform, [spectrogram])
+        self.assert_grad(deterministic_transform_iid, [spectrogram + spectrogram2])
 
     def test_spectral_centroid(self):
         sample_rate = 8000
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -125,6 +125,23 @@ def test_fade(self, fade_shape):
...
+            n_fft=n_fft, power=1)
+        spectrogram2 = get_spectrogram(
+            # Add second white noise with another seed, so they are different
+            get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2, seed=31),
+            n_fft=n_fft, power=1)
+        deterministic_transform = _DeterministicWrapper(masking_transform(400))
+        deterministic_transform_iid = _DeterministicWrapper(masking_transform(400, True))
+
+        self.assert_grad(deterministic_transform, [spectrogram])
+        self.assert_grad(deterministic_transform_iid, [spectrogram + spectrogram2])
```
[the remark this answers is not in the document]
kiri11: Comments addressed in the next commit

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -125,6 +125,30 @@ def test_fade(self, fade_shape):
...
+    def test_masking_iid(self, masking_transform):
+        sample_rate = 8000
+        n_fft = 400
+        specs = [get_spectrogram(
+            get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2, seed=i),
+            n_fft=n_fft, power=1)
+            for i in range(3)
+        ]
+
+        batch = torch.cat(specs)
```
mthrok, line 148: Can you add assertion in the `assert batch.ndim == 4`?
The code in the question automatically falls back to non-iid code path, so having assertion here makes it simpler to review the code.

https://github.com/pytorch/audio/blob/1f136671b84071a2fe1d5b762df64f3a76310c31/torchaudio/transforms.py#L884

kiri11: Done!

mthrok:
@kiri11  Thanks!

PR #1499, opened 2021-05-11
Precompute transforms.Resample kernel

carolineechen:
The kernel used for resampling computations is the same for a given set of resampling parameters, but is currently being recomputed every call to `resample`. To make computation more efficient for `transforms.Resample`, we can precompute the kernel and pass it in to the function instead.

Follow-ups:
- arbitrary source resampling transforms
- [BC-breaking] do not move kernel device and dtype during computation

cc #1487

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -1299,10 +1299,28 @@ def compute_kaldi_pitch(
 
 
 def _get_sinc_resample_kernel(
-        orig_freq: int,
-        new_freq: int,
+        orig_freq: float,
+        new_freq: float,
+        gcd: int,
         lowpass_filter_width: int,
         rolloff: float):
+
+    if not (int(orig_freq) == orig_freq and int(new_freq) == new_freq):
+        warnings.warn(
+            "Non-integer frequencies are being cast to ints and may result in poor resampling quality "
+            "because the underlying algorithm requires an integer ratio between `orig_freq` and `new_freq`. "
+            "Using non-integer valued frequencies will throw an error in the next release. "
+            "To work around this issue, manually convert both frequencies to integer values "
+            "that maintain their resampling rate ratio before passing them into the function "
+            "Example: To downsample a 44100 hz waveform by a factor of 8, use "
+            "`orig_freq=8` and `new_freq=1` instead of `orig_freq=44100` and `new_freq=5512.5` "
+            "For more information or to leave feedback about this change, please refer to "
+            "https://github.com/pytorch/audio/issues/1487."
+        )
+
+    orig_freq = int(orig_freq) // gcd
+    new_freq = int(new_freq) // gcd
+
     assert lowpass_filter_width > 0
     kernels = []
     base_freq = min(orig_freq, new_freq)
@@ -1353,11 +1371,15 @@ def _get_sinc_resample_kernel(
 
 def _apply_sinc_resample_kernel(
         waveform: Tensor,
-        orig_freq: int,
-        new_freq: int,
+        orig_freq: float,
+        new_freq: float,
+        gcd: int,
         kernel: Tensor,
         width: int,
 ):
+    orig_freq = int(orig_freq) // gcd
+    new_freq = int(new_freq) // gcd
+
     # pack batch
     shape = waveform.size()
     waveform = waveform.view(-1, shape[-1])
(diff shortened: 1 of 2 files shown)


torchaudio/functional/functional.py
```diff
@@ -1377,9 +1378,16 @@ def resample(
             but less efficient. We suggest around 4 to 10 for normal use. (Default: ``6``)
         rolloff (float, optional): The roll-off frequency of the filter, as a fraction of the Nyquist.
             Lower values reduce anti-aliasing, but also reduce some of the highest frequencies. (Default: ``0.99``)
+        kernel (Tensor, optional): Tensor of dimension (f, 1, w) representing the windowed sinc function that is
```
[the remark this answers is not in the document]
carolineechen: 1. I don't think passing in width will make computation flow (much) simpler -- my concern is that different libraries may compute kernels differently or use different convolution sizes (which is what width is being used for), and it is therefore hard to do sanity checks on their input to make sure it is reasonable and aligns with their parameters. Also because there is simplification to be done to create these kernels (like dividing by gcd) and it isn't super intuitive, I don't think kernel should be a parameter that should be exposed to users (unless we also expose the kernel function and provide concrete documentation for it)
2. `%timeit torchaudio.functional.resample(sig_torch, P, Q)` gives `10.6 ms ± 218 µs per loop` 
`resampler = torchaudio.transforms.Resample(); %timeit resampler(sig_torch)` gives `1.46 ms ± 95.6 µs per loop`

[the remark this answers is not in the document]
mthrok: We may want to introduce a new Transform for fixed sampling rate, for faster computation.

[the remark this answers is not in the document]
carolineechen: Yes, the kernel is dependent on a set of parameters (old_freq, new_freq, lowpass_filter_width, rolloff,..) and is the same for that given set of parameters. Currently in transforms.Resample, all of these parameters are being passed in init, and the only parameter for forward is the waveform itself, so I think it is fine to reuse the current transform.

torchaudio/transforms.py
```diff
@@ -653,12 +657,16 @@ def __init__(self,
                  lowpass_filter_width: int = 6,
                  rolloff: float = 0.99) -> None:
         super(Resample, self).__init__()
```
mthrok, line 659: You might want to reuse the logic to issue the warning about non-integer sampling rate values somehow . So that it will be triggered from `T.Resample` as well.

carolineechen: also, the `__init__` input types currently indicates an int for the frequencies, but the docstring mention float. which one should be changed?

mthrok: Since `F.resample`. Expects `float`, let's align it with `float`.

torchaudio/transforms.py
```diff
@@ -647,18 +651,36 @@ class Resample(torch.nn.Module):
...
+                "Non-integer frequencies are being cast to ints and may result in poor resampling quality "
+                "because the underlying algorithm requires an integer ratio between `orig_freq` and `new_freq`. "
+                "Using non-integer valued frequencies will throw an error in the next release. "
+                "To work around this issue, manually convert both frequencies to integer values "
+                "that maintain their resampling rate ratio before passing them into the function "
+                "Example: To downsample a 44100 hz waveform by a factor of 8, use "
+                "`orig_freq=8` and `new_freq=1` instead of `orig_freq=44100` and `new_freq=5512.5` "
+                "For more information or to leave feedback about this change, please refer to "
+                "https://github.com/pytorch/audio/issues/1487."
+            )
```
mthrok, line 672: Copy-pasting the chunk is cumbersome for the future maintenance. Instead, this logic of GCD and integer validation should go inside of `_get_sinc_resample_kernel`. That way, one code is used from different entry points.
In future, when we want to add more variations of kernels, we can extract the GCD and integer validation and make one helper function, which is called from `_get_sinc_resample_kernel` and the new kernel function.

torchaudio/functional/functional.py
```diff
@@ -1353,6 +1369,34 @@ def _get_sinc_resample_kernel(
...
+        kernel: Tensor,
+        width: int,
+):
+    orig_freq = int(orig_freq) // gcd
+    new_freq = int(new_freq) // gcd
+
+    # pack batch
+    shape = waveform.size()
+    waveform = waveform.view(-1, shape[-1])
+    kernel = kernel.to(device=waveform.device, dtype=waveform.dtype)
```
mthrok, line 1386: Considering the fact that this function will be called from Transform, we might want to avoid moving parameter to a certain device.

In a use case like 
1. Initialize `Resample`
2. Move the pipeline to a specific device
3. Run a inference.

and let's say that the pipeline is in GPU but the waveform is on CPU. We want the pipeline to fail so that users can fix the pipeline, instead of performing the operation on CPU, which incurs the cost of moving kernel from GPU to CPU. (well and then it will fail in the next step of pipeline, which expects the Tensor to be on GPU)

I think, moving the `kernel` to target device can happen in between `_get_kernel` function and `_apply_kernel` functions.

carolineechen: sure, where do you think target device/dtype should initially be set in transforms?

mthrok: > sure, where do you think target device/dtype should initially be set in transforms?

They are expected to be set by users with `.to`, so we do not need to set it explicitly.

```python
resamp = T.Resample(...)
resamp.to(device, dtype)
```

carolineechen: as discussed offline, this is BC-breaking and left as a follow-up

mthrok (approved):
Looks good. Thanks!

Issue #1454, opened 2021-04-14
Inconsistent default and TS-incompatible lazy behavior in MelScale


mthrok:
1. All the spectrogram related transforms have default frequency bin `n_fft: int = 400`, while `MelScale` and `InverseMelScale` is `n_stft: Optional[int] = None`.

2. In `MelScale`, when `n_stft=None`, it tries to resize the buffer in `forward`, but this causes TorchScripted (and loaded from file) version fail.

https://github.com/pytorch/audio/blob/931555c1b4ebddf74bb5d439ea197dc1d1691a05/torchaudio/transforms.py#L287-L308

https://app.circleci.com/pipelines/github/pytorch/audio/5716/workflows/fe399658-a33f-47f2-8227-7750b2f0af2f/jobs/197223/tests#failed-test-0

```
>       return callable(*args, **kwargs)
E       RuntimeError: The following operation failed in the TorchScript interpreter.
E       Traceback of TorchScript, serialized code (most recent call last):
E         File "code/__torch__/torchaudio/transforms.py", line 20, in forward
E           if torch.eq(torch.numel(self.fb), 0):
E             tmp_fb = _0(torch.size(specgram0, 1), 0., 8000., 128, 16000, self.norm, self.mel_scale, )
E             _1 = torch.resize_(self.fb, torch.size(tmp_fb), memory_format=None)
E                  ~~~~~~~~~~~~~ <--- HERE
E             _2 = torch.copy_(self.fb, tmp_fb, False)
E           else:
E       
E       Traceback of TorchScript, original code (most recent call last):
E         File "/root/project/env/lib/python3.9/site-packages/torchaudio-0.9.0a0+bb886e7-py3.9-linux-x86_64.egg/torchaudio/transforms.py", line 302, in forward
E                                               self.mel_scale)
E                   # Attributes cannot be reassigned outside __init__ so workaround
E                   self.fb.resize_(tmp_fb.size())
E                   ~~~~~~~~~~~~~~~ <--- HERE
E                   self.fb.copy_(tmp_fb)
E           
E       RuntimeError: Trying to resize storage that is not resizable at /opt/conda/conda-bld/pytorch_1617951974812/work/aten/src/TH/THStorageFunctions.cpp:87
```

To reproduce,
1. construct `MelScale` with `n_stft=None`.
2. Script the transform and save on file
3. Load the transform from file and feed a spectrogram Tensor.

Once the transform is scripted and dumped, there is no way to fix this issue.
[...]

PR #1505, opened 2021-05-13
Melscale torchscript bugfix

borundev:
Added a check in __pre_scriptable__ method for MelScale to make that self.n_stft is not None at the time of scripting.

See https://github.com/pytorch/audio/issues/1454 for details.

### DIFF
--- a/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
@@ -59,9 +59,12 @@ class Transforms(TempDirMixin, TestBaseMixin):
         spec = torch.rand((6, 201))
         self._assert_consistency(T.AmplitudeToDB(), spec)
 
+    def test_MelScale_invalid(self):
+        with self.assertRaises(ValueError):
+            torch.jit.script(T.MelScale())
+
     def test_MelScale(self):
         spec_f = torch.rand((1, 201, 6))
-        self._assert_consistency(T.MelScale(), spec_f)
         self._assert_consistency(T.MelScale(n_stft=201), spec_f)
 
     def test_MelSpectrogram(self):
--- a/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
@@ -61,6 +61,7 @@ class Transforms(TempDirMixin, TestBaseMixin):
 
     def test_MelScale(self):
         spec_f = torch.rand((1, 201, 6))
+        self._assert_consistency(T.MelScale(), spec_f)
         self._assert_consistency(T.MelScale(n_stft=201), spec_f)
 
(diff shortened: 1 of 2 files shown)


torchaudio/transforms.py
```diff
@@ -276,14 +276,24 @@ def __init__(self,
...
             self.mel_scale)
         self.register_buffer('fb', fb)
 
+    def __prepare_scriptable__(self):
+        r"""
+        Prepare self to be scripted
+        Returns:
+            MelScale: self
+        """
+        assert self.n_stft is not None, ValueError("n_stft must be provided at construction")
```
mthrok, line 294: I realized that checking `self.fb.numel() != 0`, in the similar way as in the `forward` method, is more appropriate here. Can you change it? My original direction was not precise.

torchaudio/transforms.py
```diff
@@ -276,14 +276,24 @@ def __init__(self,
...
         assert f_min <= self.f_max, 'Require f_min: {} < f_max: {}'.format(f_min, self.f_max)
 
-        fb = torch.empty(0) if n_stft is None else F.create_fb_matrix(
+        fb = torch.empty(0) if self.n_stft is None else F.create_fb_matrix(
             n_stft, self.f_min, self.f_max, self.n_mels, self.sample_rate, self.norm,
             self.mel_scale)
         self.register_buffer('fb', fb)
 
+    def __prepare_scriptable__(self):
+        r"""
```
mthrok, line 289: Since this is an internal API and library users do not use this directly, it's better to leave messages for the later maintainers, explaining why.

something like the following

```
If `self.fb` is empty, the `forward` function will try to resize the parameter, 
which does not work once the transform is scripted, but this error does not happen 
until the transform is executed. This is inconvenient especially if the resulting 
TorchScript object is executed in other environments. Therefore, we check the 
validity of `self.fb` here and fail if the resulting TS does not work.
```

test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
```diff
@@ -61,6 +61,7 @@ def test_AmplitudeToDB(self):
 
     def test_MelScale(self):
         spec_f = torch.rand((1, 201, 6))
+        self._assert_consistency(T.MelScale(), spec_f)
         self._assert_consistency(T.MelScale(n_stft=201), spec_f)
```
mthrok, line 65: The test to be added is for checking the different aspect of the `MelScale` class.
Therefore, we want to add it as a separate test method.

Let's call this new test `def test_MelScale_invalid(self)`, inside of this test, we want to make sure that scripting `MelScale` with empty `fb` would fail. i.e. `torch.jit.script(T.MelScale())`, should fail.

For this we can use [TestCase.assertRaises](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises).

mthrok (approved):
Looks good. Thanks.

PR #1507, opened 2021-05-15
Torchscriptable RNNT Loss

carolineechen:
- move RNNT loss autograd function to C++
- add torchscriptability tests

cc #1240

### DIFF
--- a/torchaudio/csrc/rnnt/autograd.cpp
+++ b/torchaudio/csrc/rnnt/autograd.cpp
@@ -1,0 +1,70 @@
+#include <torch/script.h>
+#include <torchaudio/csrc/rnnt/compute.h>
+
+namespace torchaudio {
+namespace rnnt {
+
+class RNNTLossFunction : public torch::autograd::Function<RNNTLossFunction> {
+  public:
+    static torch::autograd::tensor_list forward(
+        torch::autograd::AutogradContext* ctx,
+        torch::Tensor& logits,
+        const torch::Tensor& targets,
+        const torch::Tensor& src_lengths,
+        const torch::Tensor& tgt_lengths,
+        int64_t blank,
+        double clamp,
+        bool fused_log_smax = true,
+        bool reuse_logits_for_grads = true) {
+      at::AutoNonVariableTypeMode g;
+      torch::Tensor undef;
+      auto result = rnnt_loss(logits, targets, src_lengths, tgt_lengths, blank, clamp,
+                              fused_log_smax, reuse_logits_for_grads);
+      auto costs = std::get<0>(result);
+      auto grads = std::get<1>(result).value_or(undef);
+      std::cout << grads << std::endl;
+      ctx->save_for_backward({grads});
+      return {costs, grads};
+    }
+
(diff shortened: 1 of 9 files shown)


torchaudio/csrc/rnnt/autograd.cpp
```diff
@@ -0,0 +1,75 @@
...
+      tgt_lengths,
+      blank,
+      clamp,
+      fused_log_smax,
+      reuse_logits_for_grads);
+  return std::make_tuple(results[0], results[1]);
+}
+
+TORCH_LIBRARY_IMPL(torchaudio, Autograd, m) {
+  m.impl("rnnt_loss", rnnt_loss_autograd);
```
vincentqb, line 71: does binding RNNTLossFunction::apply work directly?

carolineechen: It doesn't immediately work because of some C++ errors and I think the return types would need to be changed also

torchaudio/prototype/rnnt_loss.py
```diff
@@ -166,17 +105,30 @@ def rnnt_loss(
...
-        blank,
-        clamp,
-        fused_log_softmax,
-        reuse_logits_for_grads,
-    )
-    return cost
+    # move everything to the same device.
+    targets = targets.to(device=logits.device)
+    logit_lengths = logit_lengths.to(device=logits.device)
+    target_lengths = target_lengths.to(device=logits.device)
```
vincentqb, line 111: i can see a situation where a user moved all but the logits accidentally, and would encounter a very slow computation of the loss function. do we do this elsewhere in the codebase? were the checks in the c++ code not catching this already and raising an error?

torchaudio/prototype/rnnt_loss.py
```diff
@@ -166,17 +105,30 @@ def rnnt_loss(
...
+
+    # make sure all int tensors are of type int32.
+    targets = targets.int()
+    logit_lengths = logit_lengths.int()
+    target_lengths = target_lengths.int()
```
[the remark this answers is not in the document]
carolineechen: you're right, the C++ checks should catch these failures

mthrok (approved):
Looks good.

torchaudio/prototype/rnnt_loss.py
```diff
@@ -166,17 +87,20 @@ def rnnt_loss(
...
-        logit_lengths,
-        target_lengths,
-        blank,
-        clamp,
-        fused_log_softmax,
-        reuse_logits_for_grads,
-    )
-    return cost
+    if blank < 0:  # reinterpret blank index if blank < 0.
+        blank = logits.shape[-1] + blank
```
vincentqb, line 91: is this also needed for `alphas` and `betas`?

carolineechen: It seems likely that this may be the case, but will leave this as a followup. Per offline discussion, it may be a better idea to just remove these functions which are not used internally and also not exposed to the user.

PR #1509, opened 2021-05-17
Add kaiser window support to resampling

carolineechen:
Add kaiser window as an option for resampling
Jupyter notebook of some results [kaiser_resampling__1_.pdf](https://github.com/pytorch/audio/files/6497164/kaiser_resampling__1_.pdf)

cc #1487

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -1303,7 +1303,9 @@ def _get_sinc_resample_kernel(
         new_freq: float,
         gcd: int,
         lowpass_filter_width: int,
-        rolloff: float):
+        rolloff: float,
+        resampling_method: str,
+        beta: float = 6.):
 
     if not (int(orig_freq) == orig_freq and int(new_freq) == new_freq):
         warnings.warn(
@@ -1352,15 +1354,20 @@ def _get_sinc_resample_kernel(
     # they will have a lot of almost zero values to the left or to the right...
     # There is probably a way to evaluate those filters more efficiently, but this is kept for
     # future work.
-    idx = torch.arange(-width, width + orig_freq)
+    idx = torch.arange(-width, width + orig_freq, dtype=torch.float64)
 
     for i in range(new_freq):
         t = (-i / new_freq + idx / orig_freq) * base_freq
         t = t.clamp_(-lowpass_filter_width, lowpass_filter_width)
-        t *= math.pi
-        # we do not use torch.hann_window here as we need to evaluate the window
+
+        # we do not use built in torch windows here as we need to evaluate the window
         # at specific positions, not over a regular grid.
-        window = torch.cos(t / lowpass_filter_width / 2)**2
+        if resampling_method == "sinc_interpolation":
+            window = torch.cos(t * math.pi / lowpass_filter_width / 2)**2
+        elif resampling_method == "kaiser_window":
+            beta = torch.tensor(beta, dtype=float)
+            window = torch.i0(beta * torch.sqrt(1 - (t / lowpass_filter_width) ** 2)) / torch.i0(beta)
+        t *= math.pi
         kernel = torch.where(t == 0, torch.tensor(1.).to(t), torch.sin(t) / t)
         kernel.mul_(window)
         kernels.append(kernel)
@@ -1403,6 +1410,8 @@ def resample(
         new_freq: float,
         lowpass_filter_width: int = 6,
         rolloff: float = 0.99,
+        resampling_method: str = "sinc_interpolation",
+        **kwargs,
 ) -> Tensor:
     r"""Resamples the waveform at the new frequency. This matches Kaldi's OfflineFeatureTpl ResampleWaveform
     which uses a LinearResample (resample a signal at linearly spaced intervals to upsample/downsample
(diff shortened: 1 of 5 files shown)


torchaudio/functional/functional.py
```diff
@@ -1303,7 +1303,9 @@ def _get_sinc_resample_kernel(
         new_freq: float,
         gcd: int,
         lowpass_filter_width: int,
-        rolloff: float):
+        rolloff: float,
+        resampling_method: str,
+        beta: float = 6.):
```
carolineechen, line 1308: According to https://numpy.org/doc/stable/reference/generated/numpy.kaiser.html, a value of 6 is similar to a Hanning window shape (used for sinc_interpolation), but I'm open to other default values as well. Resampy uses ~8.55 for their kaiser_fast and ~14.77 for kaiser_best.

torchaudio/functional/functional.py
```diff
@@ -1352,15 +1354,20 @@ def _get_sinc_resample_kernel(
     # they will have a lot of almost zero values to the left or to the right...
     # There is probably a way to evaluate those filters more efficiently, but this is kept for
     # future work.
-    idx = torch.arange(-width, width + orig_freq)
+    idx = torch.arange(-width, width + orig_freq, dtype=torch.float64)
```
carolineechen, line 1357: through running separate scripts, I realized that #1499 introduced rounding errors if waveform dtype was greater than float32. adding dtype=float64 here to retain implementation accuracy from the prior version, and this implementation will likely be improved after additional discussion on transforms kernel dtype/device computation

torchaudio/functional/functional.py
```diff
@@ -1352,15 +1354,20 @@ def _get_sinc_resample_kernel(
...
     for i in range(new_freq):
         t = (-i / new_freq + idx / orig_freq) * base_freq
         t = t.clamp_(-lowpass_filter_width, lowpass_filter_width)
-        t *= math.pi
-        # we do not use torch.hann_window here as we need to evaluate the window
+
+        # we do not use built in torch windows here as we need to evaluate the window
         # at specific positions, not over a regular grid.
-        window = torch.cos(t / lowpass_filter_width / 2)**2
+        if resampling_method == "sinc_interpolation":
```
mthrok, line 1365: Here is an idea. Since `"sinc_interpolation"` is not semantically correct, we deprecate it and add `"hann_window"` and change the default value to `"hann_window"`. (or was the default `square window`? I forgot)

Since there was no other choice before, we do not expect that a user would actually be passing `"sinc_interpolation"`, but if so we issue a warning to change it to `"hann_window"`, and in the future release, we can remove `"sinc_interpolation"` value. That way, the choices in the value of `resampling_method` makes more sense.

torchaudio/functional/functional.py
```diff
@@ -1403,6 +1410,8 @@ def resample(
         new_freq: float,
         lowpass_filter_width: int = 6,
         rolloff: float = 0.99,
+        resampling_method: str = "sinc_interpolation",
+        **kwargs,
```
mthrok, line 1414: Can you check if `**kwargs` is supported by TorchScript? IIRC, it's not supported. That case, we need to think of a structured way to select window type and pass window parameter.

torchaudio/functional/functional.py
mthrok: After the offline conversation. I think we can leave this as-is. Sinc interpolation is used when apply kaiser window as-well, but as a function option, we can interpret it as "just sinc interpolation".

torchaudio/functional/functional.py
```diff
@@ -1303,7 +1303,9 @@ def _get_sinc_resample_kernel(
         new_freq: float,
         gcd: int,
         lowpass_filter_width: int,
-        rolloff: float):
+        rolloff: float,
+        resampling_method: str,
+        beta: Optional[float]):
```
mthrok, line 1309: Can you add validation for `resample_method` at the beginning?

torchaudio/transforms.py
```diff
@@ -657,19 +657,22 @@ class Resample(torch.nn.Module):
     Args:
         orig_freq (float, optional): The original frequency of the signal. (Default: ``16000``)
         new_freq (float, optional): The desired frequency. (Default: ``16000``)
-        resampling_method (str, optional): The resampling method. (Default: ``'sinc_interpolation'``)
+        resampling_method (str, optional): The resampling method.
+            Options: [``sinc_interpolation``, ``kaiser_window``] (Default: ``'sinc_interpolation'``)
```
mthrok, line 661: followup: This docstring can be improved.

torchaudio/functional/functional.py
```diff
@@ -1318,9 +1320,15 @@ def _get_sinc_resample_kernel(
...
         )
 
+    if resampling_method not in ['sinc_interpolation', 'kaiser_window']:
+        raise ValueError('Invalid resampling method: {}'.format(resampling_method))
+
     orig_freq = int(orig_freq) // gcd
     new_freq = int(new_freq) // gcd
 
+    if resampling_method == "kaiser_window" and beta is None:
+        beta = 6.
```
mthrok, line 1330: my personal preference for default beta value is `kaiser_fast` because it feels balanced, but it's subjective and I am fine with 6 as well.

mthrok (approved):

Issue #1468, opened 2021-04-21
Voice Activity Detection


vincentqb:
I'd love to hear from our users what is the impact of their favorite VAD algorithm on their model's performance at both training and inference (offline, online, streaming).
- Is there one we should add in torchaudio?
- Does anyone uses [webrtc's vad](https://github.com/wiseman/py-webrtcvad) implementation mentioned in [comment](https://github.com/pytorch/audio/issues/1348#issuecomment-817904097)?

We currently have in torchaudio
- [the example interactive ASR pipeline](https://github.com/pytorch/audio/blob/master/examples/interactive_asr/vad.py)
- [sox re-implementation in native pytorch](https://github.com/pytorch/audio/blob/master/torchaudio/functional/filtering.py#L1347)
- [apply_effects_tensor with VAD effect](https://github.com/pytorch/audio/blob/master/torchaudio/sox_effects/sox_effects.py#L55)

cc @mthrok @astaff

PetrochukM:
Voice Activity Detection has been important for TTS and STT. It's useful for segmenting large datasets before training. I'd love a basic VAD implementation akin to:
https://maelfabien.github.io/project/Speech_proj/#pros-and-cons

`webrtc` didn't work for me because it didn't expose enough parameters for tunning. The default version of `webrtc` didn't work well for clean datasets... It tended to overcorrect for noise. It'd cut off unvoiced consonants because it thought they were noise... I tried all of the available settings in `py-webrtcvad` to correct for this. 

I don't think there is one perfect algorithm for VAD because you need to make all sorts of assumptions about noise and voice levels. So, I prefer a VAD which can be tuned by hand for different situations: no noise, minimal background noise, medium background noise, etc...  

The one currently in `torchaudio` didn't work for me because it is focused on audio trimming instead of detecting voice throughout the audio.

vincentqb:
great response @PetrochukM, thanks a lot for the input :)

vincentqb:
(quick note: some elements of [the algorithm you suggested](https://maelfabien.github.io/project/Speech_proj/#pros-and-cons) are also similar to one of the [pitch detection algorithm](https://github.com/pytorch/audio/blob/master/torchaudio/functional/functional.py#L975) we have)

PetrochukM:
Oh. I didn't. Thanks for the tip! It looks pretty close to what I was describing :) 

[...]

PR #1513, opened 2021-05-18
Update VAD docstring and check for input shape length

astaff:
This PR:
* adds warning when input tensor has more than two dimensions
* updates docstring for Vad transform module and function

Reviewer note:
Please check the wording of the warning message

This should bring the conversation in #1348 to a conclusion, while broader design discussions are still happening in #1468

### DIFF
--- a/test/torchaudio_unittest/transforms/sox_compatibility_test.py
+++ b/test/torchaudio_unittest/transforms/sox_compatibility_test.py
@@ -66,15 +66,14 @@ class TestFunctionalFiltering(TempDirMixin, TorchaudioTestCase):
         self.assert_sox_effect(result, path, ['vad'])
 
     def test_vad_warning(self):
+        """vad should throw a warning if input dimension is greater than 2"""
         sample_rate = 41100
+
         data = torch.rand(5, 5, sample_rate)
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
             T.Vad(sample_rate)(data)
         assert len(w) == 1
-
-    def test_vad_no_warning(self):
-        sample_rate = 41100
 
         data = torch.rand(5, sample_rate)
         with warnings.catch_warnings(record=True) as w:
--- a/test/torchaudio_unittest/transforms/sox_compatibility_test.py
+++ b/test/torchaudio_unittest/transforms/sox_compatibility_test.py
@@ -1,3 +1,6 @@
+import warnings
+
+import torch
 import torchaudio.transforms as T
 from parameterized import parameterized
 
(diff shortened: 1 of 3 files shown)


test/torchaudio_unittest/transforms/sox_compatibility_test.py
```diff
@@ -61,3 +64,26 @@ def test_vad(self, filename):
         data, sample_rate = load_wav(path)
         result = T.Vad(sample_rate)(data)
         self.assert_sox_effect(result, path, ['vad'])
+
+    def test_vad_warning(self):
```
mthrok, line 68: Can you provide the docstring which describes the condition when the warning should be issued and when not, so that any future maintainer can understand the intend easily?

I think we can merge `test_vad_warning` and `test_vad_no_warning`.

torchaudio/functional/filtering.py
```diff
@@ -1478,6 +1482,16 @@ def vad(
 
     # pack batch
     shape = waveform.size()
+
+    if len(shape) > 2:
+        warnings.warn(
+            "Expected input tensor shape length of 1 for single channel"
+            f" or 2 for multi-channel. Got {len(shape)} instead.\n"
+            "Batch semantics is not supported yet.\n"
```
mthrok, line 1490: * Here, I think `dimension` (as in `ndim` attribute) is more intuitive than `shape length`.
* > Batch semantics is not supported yet.

    Let's not say `"yet"`, as we do not know if we can support with the current interface.
* In general, we do not use `\n` in warnings. But if you find counter example, that's fine as well.

mthrok (approved):
Overall it looks good to me.

mthrok (approved):
Looks good. Thanks!

torchaudio/functional/filtering.py
```diff
@@ -1420,6 +1424,15 @@ def vad(
         http://sox.sourceforge.net/sox.html
     """
 
+    if waveform.ndim > 2:
+        warnings.warn(
+            "Expected input tensor dimension of 1 for single channel"
+            f" or 2 for multi-channel. Got {waveform.ndim} instead. "
+            "Batch semantics is not supported. "
+            "Please refer to https://github.com/pytorch/audio/issues/1348"
+            " and https://github.com/pytorch/audio/issues/1468."
```
vincentqb, line 1433: nit: two links might be confusing. the first is the most relevant one to this issue.

torchaudio/functional/filtering.py
```diff
@@ -1374,7 +1375,10 @@ def vad(
     so in order to trim from the back, the reverse effect must also be used.
 
     Args:
-        waveform (Tensor): Tensor of audio of dimension `(..., time)`
+        waveform (Tensor): Tensor of audio of dimension `(channels, time)` or `(time)`
```
vincentqb, line 1378: nit: "channel", see [convention](https://github.com/pytorch/audio/blob/master/README.md#conventions)

PR #1516, opened 2021-05-19
Add F.resample torchscript test

carolineechen:
add torchscriptiibility test for `F.resample`

### DIFF
--- a/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
@@ -592,12 +592,17 @@ class Functional(TempDirMixin, TestBaseMixin):
         self._assert_consistency(func, tensor)
 
     def test_resample(self):
-        def func(tensor):
+        def func_sinc(tensor):
             sr1, sr2 = 16000., 8000.
-            return F.resample(tensor, sr1, sr2)
+            return F.resample(tensor, sr1, sr2, resampling_method="sinc_interpolation")
+
+        def func_kaiser(tensor):
+            sr1, sr2 = 16000., 8000.
+            return F.resample(tensor, sr1, sr2, resampling_method="kaiser_window")
 
         tensor = common_utils.get_whitenoise(sample_rate=16000)
-        self._assert_consistency(func, tensor)
+        self._assert_consistency(func_sinc, tensor)
+        self._assert_consistency(func_kaiser, tensor)
 
     @parameterized.expand([(True, ), (False, )])
     def test_phase_vocoder(self, test_paseudo_complex):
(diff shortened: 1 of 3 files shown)


test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -591,6 +591,14 @@ def func(tensor):
         tensor = common_utils.get_whitenoise(sample_rate=44100)
         self._assert_consistency(func, tensor)
 
+    def test_resample(self):
```
mthrok, line 594: Looks good. Given tests pass, can we parameterize `resample_method` to cover the case for `"kaiser_window"`?

mthrok (approved):

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
carolineechen: tests are failing locally after rebasing due to new changes with kaiser window and constructing a tensor out of an Optional[float] type, but I will look into it

torchaudio/functional/functional.py
```diff
@@ -1422,7 +1420,7 @@ def resample(
         lowpass_filter_width: int = 6,
         rolloff: float = 0.99,
         resampling_method: str = "sinc_interpolation",
-        beta: Optional[float] = None,
+        beta: float = 14.769656459379492,
```
carolineechen, line 1423: @mthrok is it okay to use type float instead of Optional[float] even though we don't use beta in the case of resampling_method=sinc_interpolation? I get a jit error because of `beta_tensor = torch.tensor(beta)` when beta is of type `Optional[float]`, not sure if there is a better way to handle this

mthrok: I think you need to perform [optional type refinement](https://pytorch.org/docs/stable/jit_language_reference.html#optional-type-refinement). 

```
beta_ = 14.769656459379492
if beta is not None:
    beta_ = beta
beta_tensor = torch.tensor(beta_)
```

mthrok: or 

```python
if beta is not None:
    beta_tensor = torch.tensor(beta)
else:
    beta_tensor = torch.tensor(14.769656459379492)
```

torchaudio/functional/functional.py
```diff
@@ -1373,9 +1370,12 @@ def _get_sinc_resample_kernel(
...
         if resampling_method == "sinc_interpolation":
             window = torch.cos(t * math.pi / lowpass_filter_width / 2)**2
-        elif resampling_method == "kaiser_window":
-            beta = torch.tensor(beta, dtype=float)
-            window = torch.i0(beta * torch.sqrt(1 - (t / lowpass_filter_width) ** 2)) / torch.i0(beta)
+        else:
+            # kaiser_window
+            if beta is None:
+                beta = 14.769656459379492
+            beta_tensor = torch.tensor(float(beta))
```
mthrok, line 1377: 😮 Nicely done. I did not know optional type refinement can be done this elegantly.

mthrok (approved):

PR #1518, opened 2021-05-19
Remove unused RNNTL functions

carolineechen:
- remove _rnnt_loss_alphas and _rnnt_loss_betas, which are currently internal, unused, and not something we want to expose to users
- remove transducer.h file, which is not used


mthrok (approved):
token approval

vincentqb (approved):
LGTM!

PR #1519, opened 2021-05-20
Improve resampling documentation

carolineechen:
Improve resampling documentation

cc #1487

### DIFF
--- a/torchaudio/compliance/kaldi.py
+++ b/torchaudio/compliance/kaldi.py
@@ -770,6 +770,8 @@ def resample_waveform(waveform: Tensor,
             but less efficient. We suggest around 4 to 10 for normal use. (Default: ``6``)
         rolloff (float, optional): The roll-off frequency of the filter, as a fraction of the Nyquist.
             Lower values reduce anti-aliasing, but also reduce some of the highest frequencies. (Default: ``0.99``)
+        resampling_method (str, optional): The resampling method to use.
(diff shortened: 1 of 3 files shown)


torchaudio/compliance/kaldi.py
```diff
@@ -770,6 +770,8 @@ def resample_waveform(waveform: Tensor,
             but less efficient. We suggest around 4 to 10 for normal use. (Default: ``6``)
         rolloff (float, optional): The roll-off frequency of the filter, as a fraction of the Nyquist.
             Lower values reduce anti-aliasing, but also reduce some of the highest frequencies. (Default: ``0.99``)
+        resampling_method (str, optional): The resampling method to use.
+            Options: [``sinc_interpolation``, ``kaiser_window``] (Default: ``'sinc_interpolation'``)
```
carolineechen, line 774: @mthrok any thoughts on whether or not we should also add beta as a parameter to kaldi's resample waveform? this would make it more consistent with the trasnforms/functional form, but not "kaldi-compliant". `resampling_method` was added for easier testing which was done in `compliance_kaldi_test` -- if we remove beta, do we also want to remove `resampling_method` once we migrate the tests to functional?

mthrok (approved):

mthrok:
Let's add the possibility to remove `resample_waveform` from `compliance.kaldi` module in #1487 then discuss how we can proceed.

PR #1520, opened 2021-05-20
Migrate kaldi resample tests to functional

carolineechen:
Move resample tests from kaldi_compliance to functional, since we are deprecating kaldi's resample_waveform function and it is also standard to unit test in functional.

Follow up: 
- simplify sample rates in tests https://github.com/pytorch/audio/pull/1520#discussion_r636475527

### DIFF
--- a/test/torchaudio_unittest/compliance_kaldi_test.py
+++ b/test/torchaudio_unittest/compliance_kaldi_test.py
@@ -52,12 +52,6 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
     test_filepath = common_utils.get_asset_path('kaldi_file.wav')
     test_filepaths = {prefix: [] for prefix in compliance_utils.TEST_PREFIX}
 
-    def setUp(self):
-        super().setUp()
-
-        # test audio file corresponding to saved kaldi ark files
-        self.test_filepath = common_utils.get_asset_path('kaldi_file_8000.wav')
-
     # separating test files by their types (e.g 'spec', 'fbank', etc.)
     for f in os.listdir(kaldi_output_dir):
         dash_idx = f.find('-')
--- a/test/torchaudio_unittest/compliance_kaldi_test.py
+++ b/test/torchaudio_unittest/compliance_kaldi_test.py
@@ -7,7 +7,6 @@ import torchaudio.compliance.kaldi as kaldi
 
 from torchaudio_unittest import common_utils
 from .compliance import utils as compliance_utils
-from parameterized import parameterized
 
 
 def extract_window(window, wave, f, frame_length, frame_shift, snip_edges):
@@ -56,11 +55,8 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
     def setUp(self):
         super().setUp()
 
-        # test signal for testing resampling
-        self.test_signal_sr = 16000
-        self.test_signal = common_utils.get_whitenoise(
-            sample_rate=self.test_signal_sr, duration=0.5,
-        )
+        # test audio file corresponding to saved kaldi ark files
+        self.test_filepath = common_utils.get_asset_path('kaldi_file_8000.wav')
 
     # separating test files by their types (e.g 'spec', 'fbank', etc.)
     for f in os.listdir(kaldi_output_dir):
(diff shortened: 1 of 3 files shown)


test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -259,6 +288,58 @@ def test_mask_along_axis_iid_preserve(self, mask_param, mask_value, axis):
...
+        for i in range(1, 20):
+            self._test_resample_waveform_accuracy(down_scale_factor=i * 2, resampling_method=resampling_method)
+
+    @parameterized.expand([("sinc_interpolation"), ("kaiser_window")])
+    def test_resample_waveform_upsample_accuracy(self, resampling_method):
+        for i in range(1, 20):
+            self._test_resample_waveform_accuracy(up_scale_factor=1.0 + i / 20.0, resampling_method=resampling_method)
+
+    @parameterized.expand([("sinc_interpolation"), ("kaiser_window")])
+    def test_resample_waveform_multi_channel(self, resampling_method):
```
mthrok, line 323: I think this test is better suited in `batch_consistency_test.py`. The similar fixture is already there, so we should be able to simplify the test code.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -259,6 +288,58 @@ def test_mask_along_axis_iid_preserve(self, mask_param, mask_value, axis):
...
+
+    @parameterized.expand([("sinc_interpolation"), ("kaiser_window")])
+    def test_resample_waveform_downsample_accuracy(self, resampling_method):
+        for i in range(1, 20):
+            self._test_resample_waveform_accuracy(down_scale_factor=i * 2, resampling_method=resampling_method)
+
+    @parameterized.expand([("sinc_interpolation"), ("kaiser_window")])
+    def test_resample_waveform_upsample_accuracy(self, resampling_method):
+        for i in range(1, 20):
+            self._test_resample_waveform_accuracy(up_scale_factor=1.0 + i / 20.0, resampling_method=resampling_method)
```
mthrok, line 320: Can you move the for loop out of the method and put it in parameter, so that tests are independent each other?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -13,6 +13,35 @@
...
+        # resample the signal and compare it to the ground truth
+        n_to_trim = 20
+        sample_rate = 1000
+        new_sample_rate = sample_rate
+
+        if up_scale_factor is not None:
+            new_sample_rate *= up_scale_factor
+
+        if down_scale_factor is not None:
+            new_sample_rate //= down_scale_factor
```
mthrok, line 27: I think, simply passing the original sample rate and the new sample rate is more readable, rather than `up_scale_factor` / `down_scale_factor`.

But since this PR is about moving the tests, so this does not need to be addressed.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -13,6 +13,35 @@
 
 
 class Functional(TestBaseMixin):
+    def _test_resample_waveform_accuracy(self, up_scale_factor=None, down_scale_factor=None,
+                                         resampling_method="sinc_interpolation", atol=1e-1, rtol=1e-4):
```
mthrok, line 17: Could you try and see if `atol=1e-1` can be tightened?

mthrok (approved):
Looks good overall.

test/torchaudio_unittest/functional/functional_impl.py
carolineechen: It doesn't look like it

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -197,6 +197,22 @@ def test_sliding_window_cmn(self, center, norm_vars):
...
             norm_vars=norm_vars)
 
+    @parameterized.expand([("sinc_interpolation"), ("kaiser_window")])
+    def test_resample_waveform(self, resampling_method):
+        num_channels = 3
+        sr = 16000
+        new_sr = sr // 2
+        waveform = common_utils.get_whitenoise(sample_rate=sr, duration=0.5,)
+
+        multi_sound = waveform.repeat(num_channels, 1)  # (num_channels, 8000 smp)
```
mthrok, line 207: `get_whitenoise` has `n_channels` option that generates different waveforms for each channel.

https://github.com/pytorch/audio/blob/fad19fabd036879fc3e6a78d544b94c4440c1666/test/torchaudio_unittest/common_utils/data_utils.py#L42

test/torchaudio_unittest/compliance_kaldi_test.py
```diff
@@ -56,11 +55,8 @@ class Test_Kaldi(common_utils.TempDirMixin, common_utils.TorchaudioTestCase):
     def setUp(self):
         super().setUp()
 
-        # test signal for testing resampling
-        self.test_signal_sr = 16000
-        self.test_signal = common_utils.get_whitenoise(
-            sample_rate=self.test_signal_sr, duration=0.5,
-        )
+        # test audio file corresponding to saved kaldi ark files
+        self.test_filepath = common_utils.get_asset_path('kaldi_file_8000.wav')
```
mthrok, line 59: Is this for lint issue? I see that `test_filepath` attribute is already defined at class level. So this appears redundant and it seems that we can get rid of the whole `setUp` method.

test/torchaudio_unittest/functional/batch_consistency_test.py
carolineechen: Using `n_channels` produces a new input tensor which is failing for linux cpu batch consistency unittests (see [here](https://app.circleci.com/pipelines/github/pytorch/audio/6167/workflows/3de77b09-51b5-477a-b969-a83d05954a63/jobs/224472)). Is it okay to increase atol (from 1e-8 to something like 4e-8)?

mthrok: Yeah, sounds good.

carolineechen: actually just noticed the original test used [these values](https://github.com/pytorch/audio/pull/1520/files#diff-a779529b921800f7e0a12fdf7289f2915284fbbaae1ca22149b83dc2089bb78eL251), so going to reuse those

test/torchaudio_unittest/functional/functional_impl.py
mthrok: Please add this^ as a follow-up work item. The implementation can be much simplified.

test/torchaudio_unittest/functional/batch_consistency_test.py
mthrok: sounds good.

mthrok (approved):

PR #1521, opened 2021-05-20
Remove IO function from batch_consistency_test

jachymuv:
Replace IO function torchaudio.load() usage in test by on-the-fly generated white noise


mthrok (commented):
Hi @jachymuv 

Thanks for the contribution. It overall looks good.
Could you change `sample_rate=8000` and `duration=1`?
For this test, we do not need such a long audio data.

jachymuv:
OK, changed to duration=1 and sampling_rate=8000. The last test for spectral centroid I left intact for this PR, so there is still sr=44100 as it was before.

mthrok (approved):
The change looks good. Thanks!

But the Windows CI is having infrastructure issue, let me come back and rerun tests before merging this.

Issue #1503, opened 2021-05-12
Add file path in error messages of I/O operations


mthrok:
Ref #1502 

We want to add file path in the error message when I/O operations fail.

The corresponding parts are;

https://github.com/pytorch/audio/blob/1f136671b84071a2fe1d5b762df64f3a76310c31/torchaudio/csrc/sox/io.cpp#L16-L24

https://github.com/pytorch/audio/blob/1f136671b84071a2fe1d5b762df64f3a76310c31/torchaudio/csrc/sox/io.cpp#L117-L127

https://github.com/pytorch/audio/blob/14dd917ec60fa69ce3f7c6e3f2eaf520e67928b5/torchaudio/csrc/sox/effects.cpp#L98-L104

https://github.com/pytorch/audio/blob/14dd917ec60fa69ce3f7c6e3f2eaf520e67928b5/torchaudio/csrc/sox/utils.cpp#L84-L87

Note that there the first two checks the file validity directly where as the third case uses the helper function that performs the same check. This helper function is also used by in-memory I/O operations, which does not have paths. So there is a design choice to be made either we want to consolidate the check to this helper function or we want to keep doing the direct checks and add error message generation there. If we consolidate the error handling in the helper function, then we will need to optionally pass the file path that need to be included in the error message.

We need to construct error message dynamically, by attaching the path. For this one can use `std::ostream` as the example bellow.

https://github.com/pytorch/audio/blob/14dd917ec60fa69ce3f7c6e3f2eaf520e67928b5/torchaudio/csrc/sox/types.cpp#L27-L29

## Steps
1. Make the change described above.
2. Add test that tries to load a file that does not exist and inspect the error message to see if the provided path exist.
3. Make a PR and tag `cc @mthrok` in description or comment.

## Note

For setting up the development environment, please refer to [CONTRIBUTING.md](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md) and [torchaudio_unittest](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/README.md)

mthrok:
Temporarily closing for internal task assignment.

PR #1523, opened 2021-05-21
Add file path to io errors

deniskokarev:
Add file path in error messages of I/O operations #1503
cc @mthrok

### DIFF
--- a/test/torchaudio_unittest/backend/sox_io/load_test.py
+++ b/test/torchaudio_unittest/backend/sox_io/load_test.py
@@ -528,8 +528,8 @@ class TestFileObjectHttp(HttpServerMixin, PytorchTestCase):
 class TestLoadNoSuchFile(PytorchTestCase):
     def test_load_fail(self):
         """
-        Must fail to load a non-existing file with a very specific message
+        When attempted to load a non-existing file, error message must contain the file path.
         """
-        path = "__No_SuCh_FiLe__"
+        path = "non_existing_audio.wav"
         with self.assertRaisesRegex(RuntimeError, "^Error loading audio file: failed to open file {0}$".format(path)):
             sox_io_backend.load(path)
--- a/test/torchaudio_unittest/backend/sox_io/load_test.py
+++ b/test/torchaudio_unittest/backend/sox_io/load_test.py
@@ -527,7 +527,8 @@ class TestFileObjectHttp(HttpServerMixin, PytorchTestCase):
 class TestLoadNoSuchFile(PytorchTestCase):
     def test_load_fail(self):
         """
+        Must fail to load a non-existing file with a very specific message
         """
         path = "__No_SuCh_FiLe__"
-        with self.assertRaisesRegex(RuntimeError, "^Error loading audio file: failed to open file.$"):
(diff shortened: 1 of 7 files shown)


test/torchaudio_unittest/backend/sox_io/load_test.py
```diff
@@ -522,3 +522,14 @@ def test_frame(self, frame_offset, num_frames):
 
         assert sr == sample_rate
         self.assertEqual(expected, found)
+
+
+@skipIfNoSox
+class TestLoadNoSuchFile(PytorchTestCase):
+    def test_load_fail(self):
+        """
+        Must fail to load a non-existing file with a very specific message
```
mthrok, line 531:
```suggestion
        When attempted to load a non-existing file, error message must contain the file path.
```

test/torchaudio_unittest/backend/sox_io/load_test.py
```diff
@@ -522,3 +522,14 @@ def test_frame(self, frame_offset, num_frames):
...
+        """
+        path = "__No_SuCh_FiLe__"
```
mthrok, line 533: This looks very strange. Can we use an insignificant name, so that future maintainers focus naturally goes to test logic, instead of the file name.

```suggestion
        path = "non_existing_audio.wav"
```

test/torchaudio_unittest/backend/sox_io/save_test.py
```diff
@@ -20,6 +21,9 @@
     name_func,
     get_enc_params,
 )
+from torch import (
+    zeros,
+)
```
mthrok, line 26: 1. Please use `import torch` and `torch.zeros`.
2. This import should be placed before `from torchaudio.backend import sox_io_backend`, following PEP8.

test/torchaudio_unittest/backend/sox_io/save_test.py
```diff
@@ -387,3 +391,14 @@ def test_save_tensor_preserve(self, dtype):
         sox_io_backend.save(path, data, 8000)
 
         self.assertEqual(data, expected)
+
+
+@skipIfNoSox
+class TestSaveErr(PytorchTestCase):
```
mthrok, line 397: Please use more descriptive name

```suggestion
class TestSaveNonExistingDirectory(PytorchTestCase):
```

test/torchaudio_unittest/backend/sox_io/save_test.py
```diff
@@ -387,3 +391,14 @@ def test_save_tensor_preserve(self, dtype):
...
+    def test_save_fail(self):
+        """
+        Must fail to save into a non-existing directory/file with a very specific message
+        """
+        path = os.path.join("__No_SuCh_DiR__", "__No_SuCh_FiLe__.wav")
```
mthrok, line 402:
```suggestion
        path = os.path.join("non_existing_directory", "foo.wav")
```

mthrok (changes requested):
Thanks for the contribution. Could you also add test for `info` function?

test/torchaudio_unittest/backend/sox_io/save_test.py
deniskokarev: made this update

mthrok (approved):
Looks good. Thanks!

PR #1525, opened 2021-05-24
Add wav2vec2.0 model

mthrok:
This PR adds

- TorchScript-able `Wav2Vec2Model` class
- Factory functions for three configurations presented in the paper 
  - `wav2vec2_base`
  - `wav2vec2_large`
  - `wav2vec2_large_lv60k`

ref: #1506

Issue #1506, opened 2021-05-13
wav2vec2.0 model support in torchaudio


mthrok:
# wav2vec2.0 model with TorchScript support

Following the popularity of wav2vec2.0 models, we plan to add wav2vec2.0 model in torchaudio.

## What are we going to add?

1. Model definitions in `torchaudio`, that supports TorchScript, specifically `torch.jit.script`.
2. Conversion of pretrained models from Hugging Face's `transformers` (and `fairseq`).*

\* For importing models of these libraries, you will need to install them, but they will be optional dependencies of `torchaudio`.

### PRs
- [x] #1529 TorchScript-able Model definition
- [x] #1530 Import script from Hugging Face `transformers`.
- [x] #1531 Import script from `fairseq`.
- [x] #1538 C++ example.

## What is the value of this addition?

With the support for TorchScript, one can deploy wav2vec2.0 models into non-Python environments, such as C++ applications and mobiles.

There is [an example of running wav2vec2.0 in mobile](https://github.com/pytorch/android-demo-app/tree/master/SpeechRecognition), but due to the limitation on TorchScript support, it is constrained on the fixed number of time frames. We want to improve this by adding a model definition that supports scripting method.

## So is this inference only?

At the beginning, our focus and the intended use is inference. However, since the model is authored with `torch.nn.Module`, you can still plug-in it to your training as well. Note that the initial version will behave differently in training mode. (Things like random layer drop in Transformer or weight normalization will not be included.)

For training, we recommend using `fairseq`, and Hugging Face `transformers`, which are confirmed to work.

## Is there a prototype?

You can see the example of running wav2vec2.0 pre-trained models in [the prototype branch](https://github.com/pytorch/audio/tree/wav2vec2-prototype/examples/libtorchaudio/speech_recognition).

## What are the accuracies?

With the prototype above, we ran models in C++ and evaluated them on Librispeech.

For [models from `fairseq` repository](https://github.com/pytorch/fairseq/tree/master/examples/wav2vec), we also ran them in Python, and got the exact same results.

| Architecture                              | Fine Tune  | test-clean | test-other |
| ----------------------------------------- | ----------:|-----------:| ----------:|
| Base <br/><code>wav2vec_small_960h</code> | 960h | 3.1 | 7.7 |
| Large <br/><code>wav2vec_big_960h</code>  | 960h | 2.6 | 5.9 |
[...]

PR #1529, opened 2021-05-26
Add wav2vec2.0 model

mthrok:
This PR adds

- TorchScript-able `Wav2Vec2Model` class
- Factory functions for three configurations presented in the paper 
  - `wav2vec2_base`
  - `wav2vec2_large`
  - `wav2vec2_large_lv60k`

ref: #1506 
supersedes #1525


mthrok:
Quantization tests are failing for macOS CI. But I tried it locally with the latest PyTorch nightly, and it worked fine. 

```
2021-05-25 23:45:43 moto@moto-mbp:/Users/moto/Development/torchaudio  (base)
% pytest test/torchaudio_unittest/models/wav2vec2 -k quant -v
============================================================================================================ test session starts =============================================================================================================
platform darwin -- Python 3.8.5, pytest-6.2.2, py-1.10.0, pluggy-0.13.1 -- /Users/moto/miniconda3/bin/python
cachedir: .pytest_cache
rootdir: /Users/moto/Development/torchaudio
collected 18 items / 12 deselected / 6 selected

test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_0 PASSED                                                                                                                                      [ 16%]
test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_1 PASSED                                                                                                                                      [ 33%]
test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_2 PASSED                                                                                                                                      [ 50%]
test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_torchscript_0 PASSED                                                                                                                          [ 66%]
test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_torchscript_1 PASSED                                                                                                                          [ 83%]
test/torchaudio_unittest/models/wav2vec2/model_test.py::TestWav2Vec2Model::test_quantize_torchscript_2 PASSED                                                                                                                          [100%]

===================================================================================================== 6 passed, 12 deselected in 52.63s ======================================================================================================
```

@vkuzo Have you ever seen an error like this? I guess it's CI / PyTorch nightly package issue, but if you have an insight, that will be helpful as well.

From https://app.circleci.com/pipelines/github/pytorch/audio/6076/workflows/6958ea54-1bb4-482e-8d7b-c42de0bf0a4b/jobs/216325

```
self = <[AttributeError("'LinearPackedParams' object has no attribute '_packed_params'",) raised in repr()] LinearPackedParams object at 0x7faa652edef0>
weight = tensor([[0.]], size=(1, 1), dtype=torch.qint8,
       quantization_scheme=torch.per_tensor_affine, scale=1.0, zero_point=0)
bias = None

    @torch.jit.export
    def set_weight_bias(self, weight: torch.Tensor, bias: Optional[torch.Tensor]) -> None:
        if self.dtype == torch.qint8:
>           self._packed_params = torch.ops.quantized.linear_prepack(weight, bias)
E           RuntimeError: Didn't find engine for operation quantized::linear_prepack NoQEngine

```

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+                Valid length of each output sample. shape: ``[batch, ]``.
+        """
+        if x.ndim != 2:
+            raise ValueError(
+                "Expected the input Tensor to be 2D (batch, time), "
+                "but received {list(x.shape)}")
+
+        x = x.unsqueeze(1)  # (batch, channel==1, frame)
+        for layer in self.conv_layers:
+            x, length = layer(x, length)  # (batch, feathre, frame)
```
cpuhrsch, line 109: nit: feature

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+        self.embed_dim = embed_dim
+        self.conv = nn.Conv1d(
+            in_channels=embed_dim,
+            out_channels=embed_dim,
+            kernel_size=kernel_size,
+            padding=kernel_size // 2,
+            groups=groups,
+        )
+        self.conv = nn.utils.weight_norm(self.conv, name="weight", dim=2)
+        self.num_remove: int = 1 if kernel_size % 2 == 0 else 1
```
cpuhrsch, line 173: Isn't this always 1?

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+            shape_ = (batch_size, 1, length, length)
+            if attention_mask.size() != shape_:
+                raise ValueError(
+                    f"The expected attention mask shape is {shape_}. "
+                    f"Found {attention_mask.size()}."
+                )
+
+        shape = (batch_size, length, self.num_heads, self.head_dim)
+        q = self.q_proj(x).view(*shape).transpose(2, 1)  # B, nH, L, Hd
+        k = self.k_proj(x).view(*shape).permute(0, 2, 3, 1)  # B, nH, Hd, L
```
cpuhrsch, line 268: Why permute and not transpose?

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+        if attention_mask is not None:
+            shape_ = (batch_size, 1, length, length)
+            if attention_mask.size() != shape_:
+                raise ValueError(
+                    f"The expected attention mask shape is {shape_}. "
+                    f"Found {attention_mask.size()}."
+                )
+
+        shape = (batch_size, length, self.num_heads, self.head_dim)
+        q = self.q_proj(x).view(*shape).transpose(2, 1)  # B, nH, L, Hd
```
cpuhrsch, line 267: All these projections consume the same input, so you could do it [using one linear](https://github.com/pytorch/pytorch/blob/master/torch/nn/functional.py#L4683-L4736)

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+        x = x + self.pos_conv_embed(x)
+
+        if self.layer_norm_first:
+            x = self.layer_norm(x)
+
+        x = self.dropout(x)
+        for layer in self.layers:
+            # It is more readable if the following condition is flipped and we use `continue`,
+            # but TorchScript does not support `continue`.
+            if not self.training or torch.rand(1).item() > self.layer_drop:
```
cpuhrsch, line 394: If you find the inverse more readable you can put the whole block into a not
```not(self.training and torch.rand(1).item() <= self.layer_drop)```

File: torchaudio/models/wav2vec2/components.py [hunk 6]
```diff
@@ -0,0 +1,689 @@
...
+          - Large:
+            https://github.com/pytorch/fairseq/blob/425c36eafff535fe7337f8bdd5ace22ebacc78cb/examples/wav2vec/config/pretraining/wav2vec2_large_librivox.yaml#L52
+        * "conv_feature_layers"
+          - Def, base and large:
+            https://github.com/pytorch/fairseq/blob/425c36eafff535fe7337f8bdd5ace22ebacc78cb/fairseq/models/wav2vec/wav2vec2.py#L94-L100
+        * "conv_bias"
+          - Def and base:
+            https://github.com/pytorch/fairseq/blob/425c36eafff535fe7337f8bdd5ace22ebacc78cb/fairseq/models/wav2vec/wav2vec2.py#L101-L103
+          - Large:
+            https://github.com/pytorch/fairseq/blob/425c36eafff535fe7337f8bdd5ace22ebacc78cb/examples/wav2vec/config/pretraining/wav2vec2_large_librivox.yaml#L61
```
vincentqb, line 480: nit: would it be more readable in the documentation to create links instead of just having the url?

torchaudio/models/wav2vec2/components.py
```diff
@@ -0,0 +1,689 @@
...
+        x = self.feature_projection(features)
+
+        mask: Optional[Tensor] = None
+        if lengths is not None:
+            batch_size, max_len, _ = x.shape
+            # create mask for padded elements and zero-out them
+            mask = torch.arange(max_len).expand(batch_size, max_len) >= lengths[:, None]
+            x[mask] = 0.0
+            # extend the mask to attention shape and set weight
+            mask = -10000.0 * mask[:, None, None, :].to(dtype=features.dtype)
```
cpuhrsch, line 429: There's contention around what the right value here is. Parlai uses [neginf](https://github.com/facebookresearch/ParlAI/blob/2f48dcd9ab7b44b09001807b6a8a5e783972ad59/parlai/utils/torch.py#L37) that depends on the input dtype, nn.MHA uses [float(-inf)](https://github.com/pytorch/pytorch/blob/032d6b0643f10fdfd5bcc1249d045f86c4ecc454/torch/nn/functional.py#L5074) which likely has issues with lower precision dtypes and FasterTransformer uses [-10000](https://github.com/NVIDIA/FasterTransformer/blob/main/fastertransformer/cuda/attention_kernels.cu#L423) as well, fairseq uses [float("-inf")](https://github.com/pytorch/fairseq/blob/207254bf56374831d08c20064ca7a2740a871ce5/fairseq/modules/multihead_attention.py#L353) as well. I think the -10000.0 is fine, but I'm curious about your reasoning behind this.

mthrok: This value is from [HuggingFaces's implementation of Wav2Vec2.0](https://github.com/huggingface/transformers/blob/b3429ab678d0c782b0a1cc6df76f848b5c013f91/src/transformers/models/wav2vec2/modeling_wav2vec2.py#L533).

Let me try `float("-inf")` and if the test passes, then I will switch to `float("-inf")`.

torchaudio/models/wav2vec2/components.py
mthrok: Thanks for the suggestion but, this is an internal API so, it's more of detailed dev notes for future maintainers including myself, and not really intended for sphinx rendering.

torchaudio/models/wav2vec2/components.py
mthrok: I think I need to do `transpose` twice to achieve this, and I thought `permute` is more readable.

torchaudio/models/wav2vec2/model.py
```diff
@@ -0,0 +1,239 @@
...
+        Args:
+            waveforms (Tensor): Audio tensor of shape ``(batch, frames)``.
+            lengths (Tensor, optional):
+                Indicates the valid length of each audio sample in the batch.
+                Shape: ``(batch, )``.
+
+        Returns:
+            Tensor:
+                The sequences of probability distribution (in negative log-likelihood) over labels.
+                Shape: ``(batch, frames)``.
```
mthrok, line 79: TODO: fix this
- it's logit
- shape is 3D

torchaudio/models/wav2vec2/components.py
mthrok: I cannot make it work with `float("-inf")`, so I will stick with `-10000`.

torchaudio/models/wav2vec2/components.py
cpuhrsch: Ah I see, you're merging the transpose needed for the weights below

cpuhrsch (approved):

vkuzo:
> @vkuzo Have you ever seen an error like this? I guess it's CI / PyTorch nightly package issue, but if you have an insight, that will be helpful as well.

> E           RuntimeError: Didn't find engine for operation quantized::linear_prepack NoQEngine

This means that neither `fbgemm` or `qnnpack` are available in the environment.  Quantization is not supported if neither of those are available, unless the user is using a custom out of tree quantization backend.  We have seen this before, and usually we just gate the test by availability of `fbgemm`/`qnnpack`.  You could check out the 
`override_qengines` decorator in `torch/testing/_internal/common_quantized.py` for a util to do this.

PR #1530, opened 2021-05-26
Add huggingface importer

mthrok:
This PR adds

- Script to import Wav2Vec2Model from Hugging Face Transformer.

ref: #1506 

TODO
- [ ] Add `transformers` to test env so that tests actually run.

**Note**
This PR is built upon #1529 and is targeted against #1529.


cpuhrsch (approved):
LGTM!

PR #1533, opened 2021-05-31
Deprecate kaldi resample

mthrok:
Deprecating `torchaudio.compliance.kaldi.resample_waveform`, which is exactly same as `torchaudio.functional.resample`.


carolineechen (approved):
LGTM

PR #1536, opened 2021-05-31
remove reuse logits for grads from python interface.

vincentqb:
Follow up to https://github.com/pytorch/audio/pull/1532 (and needs to lands after #1532)

The default for `reuse_logits_for_grads=True` can also be confusing to users and lead to the inputs being modified in place.
- This is not the usual behavior of a loss function and needs to be changed. 
- When a function changes its inputs in place, it also needs to report this to autograd by marking them as dirty.

I’ll simply remove this option for now.

### DIFF
--- a/torchaudio/prototype/rnnt_loss.py
+++ b/torchaudio/prototype/rnnt_loss.py
@@ -15,7 +15,6 @@ def rnnt_loss(
     blank: int = -1,
     clamp: float = -1,
     fused_log_softmax: bool = True,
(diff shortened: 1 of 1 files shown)


carolineechen:
we should also make sure to remove this parameter from the tests and c++ source code as well

vincentqb:
Closed by [comment](https://github.com/pytorch/audio/pull/1532#discussion_r647660456).

PR #1537, opened 2021-06-01
Ensure resampling identity is unchanged

carolineechen:
- return the original waveform when original and new sampling rates are the same for resample
- add unit tests for identity case
- remove a unit test that tests compatibility against kaldi's resample, since it is being deprecated


mthrok (approved):
Looks good. Can you cherry pick this to `release/0.9` branch as well?

PR #1549, opened 2021-06-02
[BC-Breaking] Default to native complex type when returning raw spect…

mthrok:
…rogram

As planed in https://github.com/pytorch/audio/issues/1337

- This code changes the return type of spectrogram to be native complex dtype,
when (and only when) returning raw (complex-valued) spectrogram.
- Change `return_complex=False` to `return_complex=True` in spectrogram ops.
- `return_complex` is only effective when `power` is `None`. It is ignored for
cases where `power` is not `None`. Because the returned Tensor is power spectrogram,
which is real-valued Tensors.

### DIFF
--- a/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
@@ -51,7 +51,21 @@ class Functional(TempDirMixin, TestBaseMixin):
 
         self.assertEqual(ts_output, output)
 
-    def test_spectrogram(self):
+    def test_spectrogram_complex(self):
+        def func(tensor):
+            n_fft = 400
+            ws = 400
+            hop = 200
+            pad = 0
+            window = torch.hann_window(ws, device=tensor.device, dtype=tensor.dtype)
+            power = None
+            normalize = False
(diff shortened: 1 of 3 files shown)


anjali411 (approved):

anjali411:
maybe we should create a bc-breaking label for torchaudio? Worth preemptively adding a bc-breaking note that would go in the release notes.

> return_complex is only effective when power is None. It is ignored for
cases where power is not None. Because the returned Tensor is power spectrogram,
which is real-valued Tensors.

should this go in the doc?

mthrok:
> maybe we should create a bc-breaking label for torchaudio? Worth preemptively adding a bc-breaking note that would go in the release notes.

Yeah, we can do that. Currently we track it with `[BC-Breaking]` tag. We do have section about BC-breaking in release note. We correct them manually at release time.

> > return_complex is only effective when power is None. It is ignored for
> > cases where power is not None. Because the returned Tensor is power spectrogram,
> > which is real-valued Tensors.
> 
> should this go in docs as a note?

It kind of does, but let me update.
https://github.com/pytorch/audio/blob/f2a4aac0cc19ebfc296073bb309d4346b70862c9/torchaudio/functional/functional.py#L79

PR #1553, opened 2021-06-04
Set removal version of pseudo complex support

mthrok:
As planned in #1337

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -523,7 +523,8 @@ def mu_law_decoding(
     "Please convert the input Tensor to complex type with `torch.view_as_complex` then "
     "use `torch.abs`. "
     "Please refer to https://github.com/pytorch/audio/issues/1337 "
-    "for more details about torchaudio's plan to migrate to native complex type."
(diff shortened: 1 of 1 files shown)


torchaudio/functional/functional.py
```diff
@@ -523,7 +523,8 @@ def mu_law_decoding(
     "Please convert the input Tensor to complex type with `torch.view_as_complex` then "
     "use `torch.abs`. "
     "Please refer to https://github.com/pytorch/audio/issues/1337 "
-    "for more details about torchaudio's plan to migrate to native complex type."
+    "for more details about torchaudio's plan to migrate to native complex type.",
+    version="0.11",
```
anjali411, line 527: what is this for?

anjali411 (approved):

torchaudio/functional/functional.py
mthrok: It will set the message to `... and will be removed from 0.11 release.`.

https://github.com/pytorch/audio/blob/d4d09074f4cd0b92f168836638b8ecd81186a4c9/torchaudio/_internal/module_utils.py#L54-L56

mthrok:
thanks!

PR #1554, opened 2021-06-04
Remove override CSS

mthrok:
Currently, hyperlinks created inside of PyTorch doc section title are not rendered correctly. (see https://github.com/pytorch/pytorch_sphinx_theme/issues/115)

When I was adding `bibtex` extension, I was placing the bibliography inside of `pytorch.article` (docstring) which suffers from the same rendering issue above. I added `override.css` to workaround this, but later I decided to put all the reference at the bottom of each page, which is outside of `pytorch.article` (docstring), so it no longer suffers from this, and the workaround of override CSS is not necessary.

### DIFF
--- a/docs/source/_static/css/override.css
+++ b/docs/source/_static/css/override.css
@@ -1,8 +1,0 @@
-/* Fix for bibtex reference */
-dl.footnote.brackets > dt.label > span.brackets > a.fn-backref {
-    position: inherit
-}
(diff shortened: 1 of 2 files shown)


vincentqb (approved):
[Links](https://230761-90321822-gh.circle-artifacts.com/0/docs/models.html#luo-2019) LGTM

PR #1555, opened 2021-06-04
[BC-Breaking] Remove kaldi.resample_waveform

mthrok:
`torchaudio.compliance.kaldi.resample_waveform` has been replaced with `torchaudio.funcitonal.resample`.


carolineechen (approved):
make sure to remove `resample_waveform` from docs / ensure doc build is passing before merging, but otherwise looks good 😃 thanks!

PR #1556, opened 2021-06-06
PoC: Use UninitializedBuffer for resampling kernel cache

mthrok:
`T.Resample` precomputes and caches resampling kernel for performance improvement.
Currently, the kernel computed at the construction time and is cached with float32 first.
This causes degredation if one wants to perform resampling on float64.

In 0.9.0, we decided to use float32 for initial caching for the sake of minimum disruption
to user experience. (The original implementation computed the kernel on-the-fly on the same
device/dtype of the input Tensor, but most of use cases we are aware of are CPU/float32.)
We are now asking users to use `.to` to move the Module to the appropriate device/dtype.
But if users are using float64, this is numerically BC-breaking.
For DL application this might be fine, but for sound engineering, it might be not.

This PR adopts `LazyModuleMixin` and `UninitializedBuffer` so that resampling kernel is
computed and cached at the first forward call, assuming that `T.Resample` object has been
moved to the right `device/dtype`. This resolves the numerical BC-breaking for float64 case.

The downside of this adoptation is that
1. To TorchScript the resulting model, one has to do dry-run so that the kernel buffer is materialized. This is another BC-breaking.
2. It issues `UserWarning: Lazy modules are a new feature under heavy development so changes to the API or functionality can happen at any moment.`.

### DIFF
--- a/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
@@ -18,6 +18,9 @@ class Transforms(TempDirMixin, TestBaseMixin):
         tensor = tensor.to(device=self.device, dtype=self.dtype)
         transform = transform.to(device=self.device, dtype=self.dtype)
 
+        # Perform dry-run so that UninitializedBuffer/Parameter are materialized
+        transform(tensor)
+
         path = self.get_temp_path('transform.zip')
         torch.jit.script(transform).save(path)
         ts_transform = torch.jit.load(path)
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -7,6 +7,7 @@ from typing import Callable, Optional
 import torch
 from torch import Tensor
 from torchaudio import functional as F
+from torch.nn.parameter import UninitializedBuffer
 
 from .functional.functional import (
     _get_sinc_resample_kernel,
@@ -645,7 +646,7 @@ class MuLawDecoding(torch.nn.Module):
         return F.mu_law_decoding(x_mu, self.quantization_channels)
 
 
-class Resample(torch.nn.Module):
+class _Resample(torch.nn.Module):
     r"""Resample a signal from one frequency to another. A resampling method can be given.
 
     Note:
@@ -666,14 +667,15 @@ class Resample(torch.nn.Module):
         beta (float or None): The shape parameter used for kaiser window.
     """
 
-    def __init__(self,
-                 orig_freq: float = 16000,
-                 new_freq: float = 16000,
-                 resampling_method: str = 'sinc_interpolation',
-                 lowpass_filter_width: int = 6,
-                 rolloff: float = 0.99,
-                 beta: Optional[float] = None) -> None:
-        super(Resample, self).__init__()
+    def __init__(
+            self,
+            orig_freq: float = 16000,
+            new_freq: float = 16000,
+            resampling_method: str = 'sinc_interpolation',
+            lowpass_filter_width: int = 6,
+            rolloff: float = 0.99,
+            beta: Optional[float] = None) -> None:
+        super().__init__()
 
         self.orig_freq = orig_freq
         self.new_freq = new_freq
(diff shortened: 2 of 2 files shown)


mthrok:
@carolineechen We can add this `LazyResample` alongside with the `Resample` and direct users with need for `float64` to use it. Then later, when `LazyModuleMixin` becomes stable, we can merge two.

carolineechen:
I also want to note that #1514 was BC-breaking for other dtypes as well. Instead of the entire operation being performed with the final dtype, we first compute the kernel with `float64` to maintain maximal precision, then cast it to `float32`, and cast it once again when the user calls `resampler.to`. With dtypes smaller than `float64`, the computation in `float64` will result in a higher precision intermediate calculation and therefore slightly different numerical results when it is cast, and with `float64`, the intermediate cast to `float32` causes it to lose a small amount of precision.

cc #1487

ezyang:
didn't read in depth, but dumb question: Can you add a `dtype=` parameter to the Module in question?

mthrok:
@ezyang 

It totally works. I just did not think of it, because I was constrained on the `Module.to(device, dtype)` semantics. I will update the PR.

mthrok:
@carolineechen I updated the PR to add `dtype`. Review please.

test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
```diff
@@ -18,6 +18,9 @@ def _assert_consistency(self, transform, tensor):
         tensor = tensor.to(device=self.device, dtype=self.dtype)
         transform = transform.to(device=self.device, dtype=self.dtype)
 
+        # Perform dry-run so that UninitializedBuffer/Parameter are materialized
```
carolineechen, line 21: I don't think this is necessary if taking the `dtype` initialization approach

torchaudio/transforms.py
```diff
@@ -664,28 +664,41 @@ class Resample(torch.nn.Module):
...
-        super(Resample, self).__init__()
+    def __init__(
+            self,
+            orig_freq: float = 16000,
+            new_freq: float = 16000,
+            resampling_method: str = 'sinc_interpolation',
+            lowpass_filter_width: int = 6,
+            rolloff: float = 0.99,
+            beta: Optional[float] = None,
+            *,
```
carolineechen, line 684: what is * used for here?

torchaudio/transforms.py
```diff
@@ -664,28 +664,41 @@ class Resample(torch.nn.Module):
         rolloff (float, optional): The roll-off frequency of the filter, as a fraction of the Nyquist.
             Lower values reduce anti-aliasing, but also reduce some of the highest frequencies. (Default: ``0.99``)
         beta (float or None): The shape parameter used for kaiser window.
+        dtype (torch.device, optional):
+            Determnines the precision that resampling kernel is pre-computed and cached. If not provided,
+            kernel is computed with ``torch.float64`` then cached as ``torch.float32``.
+            If you need higher precision, provide ``torch.float64``, and the pre-computed kernel is computed and
```
carolineechen, line 670: A few questions here
- if the user provides `dtype=torch.float64`, should they still provide `.to(torch.float64)` afterwards or is that unnecessary?
- should we assume that it is preferable for the user to have the generation in `float64` even if they are using something like float16? using `.to(dtype)` in these cases will improve precision, but the resulting behavior will be different than the functional, which does the entire computation in the final dtype

test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
mthrok: oops I forgot to remove it.

torchaudio/transforms.py
mthrok: It's for keyword-only arguments. For Tensor-related operations, the common pattern in PyTorch is ..., `(*, out=None, dtype=None, layout=torch.strided, device=None, requires_grad=False)`.  See https://pytorch.org/docs/stable/generated/torch.zeros.html#torch.zeros

Now we can argue if we should follow the same pattern in transform or not, but unlike the other parameters that control the overall nature of how resampling should run, this `dtype` argument, in my eyes, has different properties. And when we think of the possibility to add more arguments for resampling algorithm itself, I would like to keep the possibility for us to add these new arguments before the `dtype` arguments without BC-breaking.

torchaudio/transforms.py
mthrok: > * if the user provides `dtype=torch.float64`, should they still provide `.to(torch.float64)` afterwards or is that unnecessary?

It depends. If the transform object user is manipulating is only composed of `Resample`, then it makes no difference, but if the processing pipeline is composed of multiple transforms (say, resample -> spectrogram -> mel-scale), then user will have to call `.to` for the entire transforms. The result-wise it should be same.

> * should we assume that it is preferable for the user to have the generation in `float64` even if they are using something like float16? 

Yes, that's my view. The only reason we are adding `dtype` is otherwise we cannot provide a mean to resample in higher precision. We are not trying to get rid of the need for users to call `.to`. Therefore, in the doc, I only instructed the case where users absolutely are necessary to use it.

> using `.to(dtype)` in these cases will improve precision, but the resulting behavior will be different than the functional, which does the entire computation in the final dtype

I think 1. users can still provide `dtype=float16`, if they want the exact same result as `F.resample`. 2. it's okay to have small discrepancy here, as it presumably produces better result (the quality of resampled audio).

carolineechen (approved):
Looks good, thanks!

PR #1558, opened 2021-06-07
Replace simple_ctc with Python greedy decoder

mthrok:
For ASR example with Wav2Vec2, I added some C++ based CTC decoder without language model. This should practically yield the same result as greedy decoding which can be implemented in Python very easily. For the purpose of example, this Python implementation is good enough, and it reduces compilation time and make the example simpler.

TODO:
 - [x] Test the resulting pipeline and ensure WERs are not affected.
     - [x] With `wav2vec_small_960h`, this gets `3.1` for `test-clean` and `7.7` for `test-other` dataset.
     - [x] With `facebook/wav2vec2-base-960h`, this gets `3.1` for `test-clean` and `7.7` for `test-other` dataset.

### DIFF
--- a/examples/libtorchaudio/speech_recognition/greedy_decoder.py
+++ b/examples/libtorchaudio/speech_recognition/greedy_decoder.py
@@ -16,17 +16,13 @@ class Decoder(torch.nn.Module):
             str: The resulting transcript
         """
         best_path = torch.argmax(logits, dim=-1)  # [num_seq,]
-        prev = ''
+        best_path = torch.unique_consecutive(best_path, dim=-1)
         hypothesis = ''
         for i in best_path:
             char = self.labels[i]
-            if char == prev:
-                continue
             if char == '<s>':
-                prev = ''
                 continue
-            prev = char
             if char == '|':
                 char = ' '
             hypothesis += char
-        return hypothesis.replace('|', ' ')
+        return hypothesis
--- a/examples/libtorchaudio/speech_recognition/greedy_decoder.py
+++ b/examples/libtorchaudio/speech_recognition/greedy_decoder.py
@@ -1,0 +1,32 @@
+import torch
+
+
+class Decoder(torch.nn.Module):
+    def __init__(self, labels):
+        super().__init__()
+        self.labels = labels
+
+    def forward(self, logits: torch.Tensor) -> str:
+        """Given a sequence logits over labels, get the best path string
+
+        Args:
+            logits (Tensor): Logit tensors. Shape `[num_seq, num_label]`.
+
+        Returns:
+            str: The resulting transcript
+        """
+        best_path = torch.argmax(logits, dim=-1)  # [num_seq,]
+        prev = ''
+        hypothesis = ''
+        for i in best_path:
+            char = self.labels[i]
+            if char == prev:
+                continue
+            if char == '<s>':
+                prev = ''
+                continue
+            prev = char
+            if char == '|':
+                char = ' '
+            hypothesis += char
+        return hypothesis.replace('|', ' ')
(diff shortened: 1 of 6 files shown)


examples/libtorchaudio/speech_recognition/greedy_decoder.py
```diff
@@ -0,0 +1,32 @@
...
+
+        Returns:
+            str: The resulting transcript
+        """
+        best_path = torch.argmax(logits, dim=-1)  # [num_seq,]
+        prev = ''
+        hypothesis = ''
+        for i in best_path:
+            char = self.labels[i]
+            if char == prev:
```
vincentqb, line 23: nit: have you tried if `torch.unique_consecutive` improves performance by getting rid of this case?

examples/libtorchaudio/speech_recognition/greedy_decoder.py
```diff
@@ -0,0 +1,32 @@
...
+                continue
+            if char == '<s>':
+                prev = ''
+                continue
+            prev = char
+            if char == '|':
+                char = ' '
+            hypothesis += char
+        return hypothesis.replace('|', ' ')
```
vincentqb, line 32: nit: does the case `if char == '|'` already takes care of this?

examples/libtorchaudio/speech_recognition/greedy_decoder.py
vincentqb: if using `torch.unique_consecutive`, it might turn out faster to use comprehension, e.g.
```python
best_path = torch.argmax(logits, dim=-1)
best_path = torch.unique_consecutive(best_path, dim=-1)
best_path = "".join(self.labels[i] for i in best_path)
best_path = best_path.replace("<s>", "").replace("|", " ")
return best_path
```
or
```python
best_path = torch.argmax(logits, dim=-1)
best_path = torch.unique_consecutive(best_path, dim=-1)
best_path = "".join(self.labels[i] for i in best_path if self.labels[i] != "<s>")
best_path = best_path.replace("|", " ")
return best_path
```

P.S. I find the first option also most readable :)

vincentqb (commented):
I like a simpler python implementation like this :)

examples/libtorchaudio/speech_recognition/greedy_decoder.py
```diff
@@ -0,0 +1,28 @@
+import torch
+
+
+class Decoder(torch.nn.Module):
```
carolineechen, line 4: I think it would be good to be more specific and name this `GreedyDecoder`, since we may be adding other decoders to torchaudio as well, and the wav2letter pipeline also calls it `GreedyDecoder` [here](https://github.com/pytorch/audio/blob/14dd917ec60fa69ce3f7c6e3f2eaf520e67928b5/examples/pipeline_wav2letter/ctc_decoders.py)

carolineechen (approved):
Left a minor comment, but LGTM

examples/libtorchaudio/speech_recognition/greedy_decoder.py
mthrok: So the reason why this is simply called `Decoder` is that in `build_pipeline_....py`, there are components `Loader`/`Decoder`/`Encoder` to illustrate the ASR components in a simplest manner.
If, in future, torchaudio natively supports a decoder, we will update the implementation of this class to serve as an adopter/guru code of the decoder from torchaudio. The interface will be specialized for the example code.
So I see this as a container class. (and currently the core decoding algorithm is missing so we default to hard-coded greedy algorithm)

agupta54:
how can we use language model for decoding with wav2vec2 torchscript models?

mthrok:
Thanks, the script looks like it's taken from fairseq. We will provide a similar solution.

PR #1568, opened 2021-06-10
0.9.0 generate new docs

carolineechen:


0.9.0/.buildinfo
```diff
@@ -1,4 +0,0 @@
-# Sphinx build info version 1
```
carolineechen, line 1: should this file have been deleted (as is the case from following README) or just updated?

0.9.0/_static/pygments.css
```diff
@@ -1,7 +1,7 @@
 pre { line-height: 125%; }
-td.linenos pre { color: #000000; background-color: #f0f0f0; padding-left: 5px; padding-right: 5px; }
```
carolineechen, line 2: same with this file -- is this okay or should it be changed back

mthrok: Can you enable your gh-pages build and skim through the doc to see if there is any obvious layout change? If not we can keep the change.

0.9.0/.buildinfo
mthrok: I think it's okay either way, as long as the resulting page is rendered correctly. We can come back to this if it affects GitHub's page deployment.

mthrok:
Did you install the dependencies as directed in doc build process? i.e. `pip install -r requirements.txt` in doc directory. If you did that, then things should be fine. The differences on CSS and such might be coming from the upstream repo.

mthrok (approved):
Token approval

PR #1572, opened 2021-06-13
Update MelSpectrogram power annotation

### LINKED ISSUE #1570: Using MelSpec when power=None
## To Reproduce

1. Use MelSpectrogram(..., power=None)

```
mel_spec = MelSpectrogram(
            sample_rate=self.feat_params["sample_rate"],
            n_fft=self.feat_params["n_window"],
            win_length=self.feat_params["n_window"],
            hop_length=self.feat_params["hop_length"],
            f_min=self.feat_params["f_min"],
            f_max=self.feat_params["f_max"],
            n_mels=self.feat_params["n_mels"],
            window_fn=torch.hamming_window,
            wkwargs={"periodic": False},
            power=None,
            )
mel_utt = mel_spec(utt_in_time)
```

## Environment

Please copy and paste the output from our
[environment collection script](https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py)
(or fill out the checklist below manually).

You can get the script and run it with:
```
wget https://raw.githubusercontent.com/pytorch/pytorch/master/torch/utils/collect_env.py
# For security purposes, please check the contents of collect_env.py before running it.
python collect_env.py
```

 - PyTorch Version (e.g., 1.0): 1.8.1+cu111
 - Torchaudio: 0.8.1
 - OS (e.g., Linux): Linux
 - How you installed PyTorch (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source): from pytorch.org
 - Python version: 3.85
 - CUDA/cuDNN version: 11.1
 - GPU models and configuration: A-100
 - Any other relevant information:

## Additional context
I actually wanted to implement time-stretching here. Think what I shall do is to use Spectrogram(..., power=0) to transform time-domain signal into frequency domain with complex numbers, and then use TimeStretch to do the time stretching. If I want to get the Mel_spec, I shall use MelScale to the obtained data from TimeStrecth right?
Anyways, just found there is a bug when am using Mel_Spec with power=0.

mthrok:
Resolves #1570 

Mel scale frequency is in general defined/used on power spectrogram and as far as we know, it is not defined for raw (complex value) spectrogram. When `power=None`, `MelSpectrogram` fails because of extra dimension (when using pseudo complex) or type mismatch between float and cfloat (when using native complex).

This PR changes type annotation so that power is only float.

@carolineechen Can you review? I think we can cherry-pick this to 0.9. Aldo Do you think this should be labeled BC-breaking? `power=None` never worked, so it does not break anything.

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -445,7 +445,7 @@ class MelSpectrogram(torch.nn.Module):
         n_mels (int, optional): Number of mel filterbanks. (Default: ``128``)
         window_fn (Callable[..., Tensor], optional): A function to create a window tensor
             that is applied/multiplied to each frame/window. (Default: ``torch.hann_window``)
-        power (float, optional): Exponent for the magnitude spectrogram,
+        power (float): Exponent for the magnitude spectrogram,
             (must be > 0) e.g., 1 for energy, 2 for power, etc. (Default: ``2``)
         normalized (bool, optional): Whether to normalize by magnitude after stft. (Default: ``False``)
         wkwargs (Dict[..., ...] or None, optional): Arguments for window function. (Default: ``None``)
@@ -476,7 +476,7 @@ class MelSpectrogram(torch.nn.Module):
                  pad: int = 0,
                  n_mels: int = 128,
                  window_fn: Callable[..., Tensor] = torch.hann_window,
-                 power: Optional[float] = 2.,
+                 power: float = 2.,
                  normalized: bool = False,
                  wkwargs: Optional[dict] = None,
                  center: bool = True,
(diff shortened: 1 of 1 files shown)


torchaudio/transforms.py
```diff
@@ -445,7 +445,7 @@ class MelSpectrogram(torch.nn.Module):
         n_mels (int, optional): Number of mel filterbanks. (Default: ``128``)
         window_fn (Callable[..., Tensor], optional): A function to create a window tensor
             that is applied/multiplied to each frame/window. (Default: ``torch.hann_window``)
-        power (float, optional): Exponent for the magnitude spectrogram,
+        power (float): Exponent for the magnitude spectrogram,
```
carolineechen, line 448: we should leave in `optional` here because there is a default value. (I did not include `or None` when I updated MelSpectrogram documentation because of uncertainty so no change should be needed here)

mthrok: Good catch. Updated.

carolineechen (approved):

PR #1577, opened 2021-06-15
fix tag build so it triggers prerequisites

mattip:
Port pytorch/vision#3998 and pytorch/vision#4053 so that adding a tag will trigger a documentation build-and-upload. The problem was that the upload depends on both the wheel and the build_docs builds, but those did not have the correct filter so they were not running.

Similar to pytorch/text#1332

It would be nice to see this merged before the next tag event, to make sure the docs are now building (the last few tags did not generate docs), although #1568 already generated the 0.9 docs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -686,6 +686,12 @@ workflows:
           wheel_docker_image: pytorch/manylinux-rocm:4.1
       - binary_linux_wheel:
           cuda_version: cpu
+          filters:
+            branches:
+              only:
+              - /.*/
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*-rc[0-9]+/
           name: binary_linux_wheel_py3.8_cpu
           python_version: '3.8'
           requires:
@@ -991,6 +997,12 @@ workflows:
           name: binary_windows_conda_py3.9_cu111
           python_version: '3.9'
       - build_docs:
+          filters:
+            branches:
+              only:
+              - /.*/
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*-rc[0-9]+/
           name: build_docs
           python_version: '3.8'
           requires:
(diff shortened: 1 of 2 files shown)


mattip:
I am pretty sure this needs to be cherry-picked to the 0.9 release branch

.circleci/config.yml
```diff
@@ -991,6 +997,12 @@ workflows:
           name: binary_windows_conda_py3.9_cu111
           python_version: '3.9'
       - build_docs:
+          filters:
+            branches:
+              only:
+              - /.*/
+            tags:
+              only: /v[0-9]+(\.[0-9]+)*-rc[0-9]+/
```
mthrok, line 1005: @mattip 

This regex looks like it will only match RC. Is this intended?

mthrok: https://regex101.com/r/KPoLgG/1

mthrok:
Merging this to see if it works no release/tag.

mthrok:
The new workflow

https://app.circleci.com/pipelines/github/pytorch/audio/6328/workflows/5b1b516e-97cf-461e-9db9-388da101b21f

However I just realized that `build_doc` was running before as well.

https://app.circleci.com/pipelines/github/pytorch/audio/6320/workflows/1aed74f6-1361-4b01-9b4e-4b9536e16510

@mattip Isn't it `upload_doc` job that has to be triggered on tag?

mattip:
Hmm. Not sure what is going on. There should be a specific tag build, this looks like a nightly build.

mthrok:
tag build triggers nightly build (and only nightly build) because only nightly build has upload process

mattip:
Right, the tag symbol shows at the top. It should have triggered the `binary_linux_wheel_py3.8_cpu` (no nightly prefix) due to the filter. Maybe merging the changed filter together with the tag got circleCI confused?

PR #1579, opened 2021-06-15
add v0.9.0 documentation, adjust versions.html, stable

mattip:
This is a backup in case #1577 doesn't work when the repo is tagged for the release. If #1577 does work and the 0.9.0 documentation is generated automatically, then only the small changes to adjust versions.html and make stable point to 0.9.0 should be needed.

This is the result of
- installing the pytorch 1.9 latest rc
- installing audio from the v0.9.0-rc6 tag
```
cd docs
make html
cd ..
git checkout gh-pages
mkdir 0.9.0
cp -r docs/build/html/* 0.9.0
```

I also edited the versions.html to add the version documentation, and changed the symlink for stable to point to 0.9.0.

Please check carefully that there are no post-rc1 documentation changes (that belong in master and not the release).

Hopefully this can be closed and not merged after the tag triggers a doc build/push. In any case, please ping me when the tag goes in, I am curious to see what happens.


mattip:
Whoops, the pages already exist. Closing.

PR #1539, opened 2021-06-01
Set release version

mthrok:
Set versions numbers
- release version
- base PyTorch version
- documentation version

Screen Shot 2021-06-01 at 14 15 35

PR #1581, opened 2021-06-15
DOC: Remove git tag from 0.9.0 pages titles

mattip:
The page titles (what you see when you hover over the page's tab in the browser) still had the git tag.

`for f in $(find . -name "*.html"); do sed -i -e"s/0\.9\.0a0+1437208/0.9.0/" $f; done`

### DIFF
--- a/0.9.0/_modules/index.html
+++ b/0.9.0/_modules/index.html
@@ -9,7 +9,7 @@
   
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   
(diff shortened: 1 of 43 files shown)


mthrok:
I think I introduced this here. I will update the release runback to not repeat this.

https://github.com/pytorch/audio/pull/1539/files#diff-008dcb3426febd767787b1521f1fe33086313b927ea37eaab86df5fa88a51698L88

PR #1584, opened 2021-06-18
Update the installation and testing instructions.

yangarbiter:
Here are the changes I made:
- Add optional installation instructions to CONTRIBUTING.md
- Remove duplicated installation information in test/torchaudio_unittest/README.md  (since we already direct people to CONTRIBUTING.md
- Fix the link to backend io tests in test/torchaudio_unittest/README.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,7 +58,6 @@ pip install cmake ninja
 ```bash
 git clone https://github.com/pytorch/audio.git
 cd audio
-git submodule sync
 git submodule update --init --recursive
 BUILD_SOX=1 python setup.py develop
 # or, for OSX
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,8 +55,6 @@ conda install pytorch -c pytorch-nightly
 pip install cmake ninja
 ```
 
-### Install other requirements
-
 ```bash
 git clone https://github.com/pytorch/audio.git
 cd audio
@@ -69,7 +67,7 @@ BUILD_SOX=1 python setup.py develop
 # DEBUG=1 python setup.py develop
 ```
 
-If you built sox, set the `PATH` variable so that the tests properly use the newly built `SoX` binary:
+If you built sox, set the `PATH` variable so that the tests properly use the newly built `sox` binary:
 
 ```bash
 export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
@@ -78,13 +76,13 @@ export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
 The following dependencies are also needed for testing:
 
 ```bash
-pip install typing pytest scipy numpy parameterized pyyaml
+pip install typing pytest scipy numpy parameterized
 ```
 
 Optional packages to install if you want to run related tests:
 
 ```bash
-pip install librosa requests soundfile kaldi_io transformers fariseq
+pip install librosa requests soundfile kaldi_io transformers
 ```
 
 ## Development Process
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,9 +80,13 @@ pip install typing pytest scipy numpy parameterized
 
 Optional packages to install if you want to run related tests:
 
-```bash
-pip install librosa requests soundfile kaldi_io transformers
-```
+- `librosa`
+- `requests`
+- `soundfile`
+- `kaldi_io`
+- `transformers`
+- `fairseq` (it has to be newer than `0.10.2`, so you will need to install from
+  source. Commit `e6eddd80` is known to work.)
 
 ## Development Process
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,9 +55,12 @@ conda install pytorch -c pytorch-nightly
 pip install cmake ninja
 ```
 
+### Install other requirements
+
 ```bash
 git clone https://github.com/pytorch/audio.git
 cd audio
+git submodule sync
 git submodule update --init --recursive
 BUILD_SOX=1 python setup.py develop
 # or, for OSX
@@ -66,7 +69,7 @@ BUILD_SOX=1 python setup.py develop
 # DEBUG=1 python setup.py develop
 ```
 
-If you built sox, set the `PATH` variable so that the tests properly use the newly built `sox` binary:
+If you built sox, set the `PATH` variable so that the tests properly use the newly built `SoX` binary:
 
 ```bash
 export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
@@ -75,7 +78,13 @@ export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
 The following dependencies are also needed for testing:
 
 ```bash
-pip install typing pytest scipy numpy parameterized
+pip install typing pytest scipy numpy parameterized pyyaml
+```
+
+Optional packages to install if you want to run related tests:
+
+```bash
+pip install librosa requests soundfile kaldi_io transformers fariseq
 ```
 
 ## Development Process
--- a/test/torchaudio_unittest/README.md
+++ b/test/torchaudio_unittest/README.md
@@ -9,6 +9,7 @@ For testing, the following packages are needed:
 
 ```bash
 pip install typing pytest scipy numpy parameterized
+pip install -r requirements.txt
 ```
 
 Make sure to follow the installation instruction in the [contributing
@@ -73,7 +74,9 @@ The following test modules are defined for corresponding `torchaudio` module/fun
 - [`torchaudio.compliance.kaldi`](./compliance_kaldi_test.py)
 - [`torchaudio.kaldi_io`](./kaldi_io_test.py)
 - [`torchaudio.sox_effects`](./sox_effect)
-- [`torchaudio.save`, `torchaudio.load`, `torchaudio.info`](./io_test.py)
+- [`torchaudio.save`, `torchaudio.load`, `torchaudio.info`]
+    - soundfile backend: [[save](./backend/soundfile/save_test.py), [load](./backend/soundfile/load_test.py), [info](./backend/soundfile/info_test.py)]
+    - SoX backend: [[save](./backend/sox_io/save_test.py), [load](./backend/sox_io/load_test.py), [info](./backend/sox_io/info_test.py)]
 
 ### Test modules that do not fall into the above categories
 - [test_dataloader.py](./dataloader_test.py)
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -55,9 +55,16 @@ conda install pytorch -c pytorch-nightly
 pip install cmake ninja
 ```
 
+### Install other requirements
+
+```bash
+pip install -r requirements.txt
```
mthrok, line 61: We do not actively maintain `requirements.txt`. (which is not good, but at the same time I always wonder the value of `requirements.txt`, especially there is similar section in `setup.py`) 

We have better description in line 85. Can we instead update it?
`librosa`, `requests`, `pyyaml`, are missing from there.
Other missing ones are `fairseq` and `transformers`, but they are very purpose-specific, so we can omit.

CONTRIBUTING.md
```diff
@@ -55,9 +55,16 @@ conda install pytorch -c pytorch-nightly
...
+```
+
 ```bash
 git clone https://github.com/pytorch/audio.git
 cd audio
+git submodule sync
```
mthrok, line 67: I do not think this is necessary.
Our CI just works without it https://github.com/pytorch/audio/blob/7deea25948383fdaad63893b7f5676e06fd8a5ad/.circleci/unittest/linux/scripts/install.sh#L44

test/torchaudio_unittest/README.md
```diff
@@ -9,6 +9,7 @@ For testing, the following packages are needed:
 
 ```bash
 pip install typing pytest scipy numpy parameterized
+pip install -r requirements.txt
```
mthrok, line 12: Instead adding duplicated information, can you replace the whole section (from `For testing, the following ...`) with a redirection to `CONTRIBUTING.md`? In all the `help-wanted` issues, we tell contributors to follow the instruction in `CONTRIBUTING.md`.

CONTRIBUTING.md
vincentqb: Both @yangarbiter and I just ran into compilation errors with kaldi despite running submodule update on fresh clone. @yangarbiter noticed we simply needed to sync. Since we are reminding users to both clone and update submodule, it seems like we might as well remind them to sync. @mthrok, have you ran seen [this error](https://github.com/pytorch/audio/files/6677554/buffer.txt)?

mthrok: `pip` cannot be used for source build. Please follow the documentation and use `python setup.py develop`.

mthrok: Note: the proper way to change the build process so that one can perform `pip install` is to perform submodule update in somewhere `setup.py`. That way, `pip install git+https://github.com/pytorch/audio` will work. 
ref: https://github.com/pytorch/text/pull/1154

CONTRIBUTING.md
yangarbiter: Sure, I'll update them.

CONTRIBUTING.md
yangarbiter: I thought `git submodule update` and `git submodule sync` are doing slightly different things ([(link)](https://stackoverflow.com/questions/45678862/git-submodule-update-vs-git-submodule-sync/45679261) ). It appears that if you just pull from the current master then it should be enough to do only `git submodule update`. But if you `git submodule update` a while ago and the url for the submodule have changed, then you'll need to do `git submodule sync` to update the url.

mthrok: Yeah, `sync` is necessary when there is a change on submodule. But here, the context is for first time, as the explanation starts from `git clone`.

How about adding some comments when you need it?

```bash
git submodule sync  # Only when submodule URLs are changed upstream.
```

vincentqb: > `pip` cannot be used for source build. Please follow the documentation and use `python setup.py develop`.

The issue occurs with `python setup.py develop` too, see [buffer.txt](https://github.com/pytorch/audio/files/6678092/buffer.txt).

> How about adding some comments when you need it?
> 
> ```shell
> git submodule sync  # Only when submodule URLs are changed upstream.
> ```

SGTM

CONTRIBUTING.md
yangarbiter: I think we also need `soundfile` otherwise many tests will fail. Should we also add it to line 85?

CONTRIBUTING.md
mthrok: I think this is due to something other than submodule. Looking at the log Kaldi is pulled correctly. The submodule has not been updated.

CONTRIBUTING.md
mthrok: Which ones are failing? When a dependency is missing our tests are supposed to be skipping them.

yangarbiter: Something like this one
```FAILED torchaudio_unittest/backend/soundfile/save_test.py::MockedSaveTest::test_sph_sph_int16_16000_1_True_ALAW_32 - ModuleNotFoundError: No module named 'soundfile'```
also some in this file ```https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/kaldi_io_test.py```

Comment (mthrok) [reply CONTRIBUTING.md:61]: Ok, these are bugs, I will fix them. So, can you update the description so that all of them are optional?
`librosa`, `requests`, `soundfile`, `kaldi_io`, `transformers` and `fariseq`.

File: CONTRIBUTING.md [hunk 4]
```diff
@@ -55,9 +55,12 @@ conda install pytorch -c pytorch-nightly
 pip install cmake ninja
 ```
 
+### Install other requirements
+
```
vincentqb, line 59: the content under this title belongs to the previous section.
```suggestion
```

CONTRIBUTING.md
yangarbiter: Thanks for pointing it out. I've opened an issue (#1589) regarding this.

CONTRIBUTING.md
```diff
@@ -66,7 +67,7 @@ BUILD_SOX=1 python setup.py develop
 # DEBUG=1 python setup.py develop
 ```
 
-If you built sox, set the `PATH` variable so that the tests properly use the newly built `sox` binary:
+If you built sox, set the `PATH` variable so that the tests properly use the newly built `SoX` binary:
```
mthrok, line 70: This is wrong. `sox` is a command line.

CONTRIBUTING.md
```diff
@@ -75,7 +76,13 @@ export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
 The following dependencies are also needed for testing:
 
 ```bash
-pip install typing pytest scipy numpy parameterized
+pip install typing pytest scipy numpy parameterized pyyaml
```
mthrok, line 79: `pyyaml` is not required. It is required when you use `fairseq`.

CONTRIBUTING.md
```diff
@@ -75,7 +76,13 @@ export PATH="<path_to_torchaudio>/third_party/install/bin:${PATH}"
...
+```
+
+Optional packages to install if you want to run related tests:
+
+```bash
+pip install librosa requests soundfile kaldi_io transformers fariseq
```
mthrok, line 85: The latest version of `fairseq` from `pip install` does not work at the moment. So please do not put it in the command line.

test/torchaudio_unittest/README.md
```diff
@@ -73,7 +69,9 @@ The following test modules are defined for corresponding `torchaudio` module/fun
 - [`torchaudio.compliance.kaldi`](./compliance_kaldi_test.py)
 - [`torchaudio.kaldi_io`](./kaldi_io_test.py)
 - [`torchaudio.sox_effects`](./sox_effect)
-- [`torchaudio.save`, `torchaudio.load`, `torchaudio.info`](./io_test.py)
+- [`torchaudio.save`, `torchaudio.load`, `torchaudio.info`]
```
mthrok, line 72: Can you just point `torchaudio.backend`? As you noticed we refactor the test often and this detailed links get obsolete.

yangarbiter: Sure, I've updated it in the latest commit.

CONTRIBUTING.md
yangarbiter: Sure, I've updated it in the latest commit.

CONTRIBUTING.md
yangarbiter: I've removed pyyaml completely since we are not including fariseq for now.

CONTRIBUTING.md
yangarbiter: Thanks for pointing it out. I've fixed it in the latest command,.

CONTRIBUTING.md
```diff
@@ -78,6 +78,12 @@ The following dependencies are also needed for testing:
 pip install typing pytest scipy numpy parameterized
 ```
 
+Optional packages to install if you want to run related tests:
+
+```bash
+pip install librosa requests soundfile kaldi_io transformers
```
mthrok, line 84: Installation of `librosa` is sometimes tricky. 

https://github.com/pytorch/audio/blob/380800ce09ef9b31475d298eb3e60f096056caca/.circleci/unittest/linux/scripts/install.sh#L55

Instead of giving the instruction with `pip install`, can you just simply list these libraries?
Something like

- `librosa`
- `requests`
- `soundfile`
- `kaldi_io`
- `transformers`
- `fairseq` (it has to be newer than `0.10.2`, so you will need to install from source. Commit `e6eddd80` is known to work.)

CONTRIBUTING.md
yangarbiter: We don't have concrete evidence showing `git submodule sync` fixes this bug. I'll remove this line from README.

test/torchaudio_unittest/README.md
yangarbiter: I've addressed this here https://github.com/pytorch/audio/pull/1584/files#diff-937f67c5380ec39c13ab0d069d3dff30ec219a3fe7c8725605caa95120b65e0cR8

CONTRIBUTING.md
yangarbiter: I've removed it. Thanks :)

CONTRIBUTING.md
yangarbiter: I've added these [here](https://github.com/pytorch/audio/pull/1584/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R81)

mthrok (approved):

vincentqb (approved):
LGTM. Given that this relates only to documentation, @yangarbiter, you can go ahead and merge without rerunning the tests.

PR #1585, opened 2021-06-18
Fix HF wav2vec2 integration test

mthrok:
It seems that the return value of intermediate layer has been changed. 
The fixed assertion is redundant so we can remove it if there are too many changes on transformer side, but let's see how it goes.


carolineechen (approved):

vincentqb (approved):
LGTM, all green :)

Issue #226, opened 2019-08-10
Request preemphasis and deemphasis modules


AppleHolic:
## 🚀 Feature
I request Preemphasis / Deemphasis modules. [wiki](https://en.wikipedia.org/wiki/Preemphasis_improvement) In several sound related deep-learning tasks, These module are often used to help sound quality.

## Motivation

In my speech enhancement case, that model usually generate high frequency noises without preemphasis. With this method, it brings many helps to enhance enhancement quality. But, to use of it on training models, it should be implemented with pytorch.

## Pitch and Alternatives

I already did code first version of preemphasis and deemphasis written in pytorch. 
So, after checking out below code has some bugs, I suggest to add this modules to transforms.py

- [Permalink of my github repo](https://github.com/AppleHolic/pytorch_sound/blob/b741b994795830c35c9627117d76727c6c4bf550/pytorch_sound/models/sound.py#L64)

- code snippet

```python
class PreEmphasis(torch.nn.Module):

    def __init__(self, coef: float = 0.97):
        super().__init__()
        self.coef = coef
        # make kernel
        # In pytorch, the convolution operation uses cross-correlation. So, filter is flipped.
        self.register_buffer(
            'flipped_filter', torch.FloatTensor([-self.coef, 1.]).unsqueeze(0).unsqueeze(0)
        )

    def forward(self, input: torch.tensor) -> torch.tensor:
        assert len(input.size()) == 3, 'The number of dimensions of input tensor must be 3!'
        # reflect padding to match lengths of in/out
        input = F.pad(input, (1, 0), 'reflect')
        return F.conv1d(input, self.flipped_filter)


class InversePreEmphasis(torch.nn.Module):
    """
    Implement Inverse Pre-emphasis by using RNN to boost up inference speed.
    """

    def __init__(self, coef: float = 0.97):
        super().__init__()
        self.coef = coef
        self.rnn = torch.nn.RNN(1, 1, 1, bias=False, batch_first=True)
        # use originally on that time
        self.rnn.weight_ih_l0.data.fill_(1)
        # multiply coefficient on previous output
        self.rnn.weight_hh_l0.data.fill_(self.coef)

    def forward(self, input: torch.tensor) -> torch.tensor:
        x, _ = self.rnn(input.transpose(1, 2))
        return x.transpose(1, 2)
```

### Sample Plots

- scipy codes

```python
from scipy import signal

def preemphasis(x, coeff=0.97):
[...]

PR #1587, opened 2021-06-18
feat: lfilter with filterbanks support

### LINKED ISSUE #1528: Support multiple filters in lfilter

yoyololicon:
resolves #1528 , relates to #1561 

This merge add the ability to apply multiple filters in `F.lfilter`. 
When the input shape is `(..., time)`, if the shape of filter coefficients is `(number_of_filters, filter_order)`, the shape of output will be `(...., number_of_filters, time)`, which is the input shape *plus* one additional dimension from the filters. 

Before feeding to the cpp backend, the python frontend first duplicate the input `number_of_filters`  times and stacking them to form a shape of `(...., number_of_filters, time)`.

The cpp backend assume the shape of `waveform` and  `*_coeffs` are `(batch, number_of_filters, time)` and `(number_of_filters, filter_order)` (before are just `(batch, time)` and `(filter_order,)`). Different filters is applied parallelly at the second dimension of `waveform`.

cc @mthrok

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -80,6 +80,19 @@ class Functional(TestBaseMixin):
         output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
         assert shape == waveform.size() == output_waveform.size()
 
+    @parameterized.expand([
+        ((44100,), (2, 3), (2, 44100)),
+        ((3, 44100), (1, 3), (3, 1, 44100)),
+        ((1, 2, 44100), (3, 3), (1, 2, 3, 44100))
+    ])
+    def test_lfilter_filterbanks_shape(self, input_shape, coeff_shape, target_shape):
+        torch.random.manual_seed(42)
+        waveform = torch.rand(*input_shape, dtype=self.dtype, device=self.device)
+        b_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
+        a_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
+        output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
+        assert target_shape == output_waveform.size()
+
     def test_lfilter_9th_order_filter_stability(self):
         """
         Validate the precision of lfilter against reference scipy implementation when using high order filter.
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -67,30 +67,21 @@ class Functional(TestBaseMixin):
         assert output_signal.max() > 1
 
     @parameterized.expand([
-        ((44100,),),
-        ((3, 44100),),
-        ((2, 3, 44100),),
-        ((1, 2, 3, 44100),)
+        ((44100,), (4,), (44100,)),
+        ((3, 44100), (4,), (3, 44100,)),
+        ((2, 3, 44100), (4,), (2, 3, 44100,)),
+        ((1, 2, 3, 44100), (4,), (1, 2, 3, 44100,)),
+        ((44100,), (2, 4), (2, 44100)),
+        ((3, 44100), (1, 4), (3, 1, 44100)),
+        ((1, 2, 44100), (3, 4), (1, 2, 3, 44100))
     ])
-    def test_lfilter_shape(self, shape):
-        torch.random.manual_seed(42)
-        waveform = torch.rand(*shape, dtype=self.dtype, device=self.device)
-        b_coeffs = torch.tensor([0, 0, 0, 1], dtype=self.dtype, device=self.device)
-        a_coeffs = torch.tensor([1, 0, 0, 0], dtype=self.dtype, device=self.device)
-        output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
-        assert shape == waveform.size() == output_waveform.size()
-
-    @parameterized.expand([
-        ((44100,), (2, 3), (2, 44100)),
-        ((3, 44100), (1, 3), (3, 1, 44100)),
-        ((1, 2, 44100), (3, 3), (1, 2, 3, 44100))
-    ])
-    def test_lfilter_filterbanks_shape(self, input_shape, coeff_shape, target_shape):
+    def test_lfilter_shape(self, input_shape, coeff_shape, target_shape):
         torch.random.manual_seed(42)
         waveform = torch.rand(*input_shape, dtype=self.dtype, device=self.device)
         b_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
         a_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
         output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
+        assert input_shape == waveform.size()
         assert target_shape == output_waveform.size()
 
     def test_lfilter_9th_order_filter_stability(self):
--- a/torchaudio/functional/filtering.py
+++ b/torchaudio/functional/filtering.py
@@ -941,18 +941,19 @@ def lfilter(
 
     Args:
         waveform (Tensor): audio waveform of dimension of ``(..., time)``.  Must be normalized to -1 to 1.
-        a_coeffs (Tensor): denominator coefficients of difference equation of dimension of ``(*, n_order + 1)``.
-                                Where * is the optional number of filter banks.
+        a_coeffs (Tensor): denominator coefficients of difference equation of dimension of either 
+                                1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
                                 Lower delays coefficients are first, e.g. ``[a0, a1, a2, ...]``.
                                 Must be same size as b_coeffs (pad with 0's as necessary).
-        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of ``(*, n_order + 1)``.
-                                Where * is the optional number of filter banks.
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of either 
+                                1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
                                 Lower delays coefficients are first, e.g. ``[b0, b1, b2, ...]``.
                                 Must be same size as a_coeffs (pad with 0's as necessary).
         clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
 
     Returns:
-        Tensor: Waveform with dimension of ``(..., *, time)``.
+        Tensor: Waveform with dimension of either ``(..., num_filters, time)`` if ``a_coeffs`` and ``b_coeffs`` are 2D Tensors, 
+                or ``(..., time)`` otherwise.
     """
     # pack batch
     shape = waveform.size()
--- a/torchaudio/functional/filtering.py
+++ b/torchaudio/functional/filtering.py
@@ -941,19 +941,19 @@ def lfilter(
 
     Args:
         waveform (Tensor): audio waveform of dimension of ``(..., time)``.  Must be normalized to -1 to 1.
-        a_coeffs (Tensor): denominator coefficients of difference equation of dimension of either 
+        a_coeffs (Tensor): denominator coefficients of difference equation of dimension of either
                                 1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
                                 Lower delays coefficients are first, e.g. ``[a0, a1, a2, ...]``.
                                 Must be same size as b_coeffs (pad with 0's as necessary).
-        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of either 
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of either
                                 1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
                                 Lower delays coefficients are first, e.g. ``[b0, b1, b2, ...]``.
                                 Must be same size as a_coeffs (pad with 0's as necessary).
         clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
 
     Returns:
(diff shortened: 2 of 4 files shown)


test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -80,6 +80,19 @@ def test_lfilter_shape(self, shape):
         output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
         assert shape == waveform.size() == output_waveform.size()
 
+    @parameterized.expand([
+        ((44100,), (2, 3), (2, 44100)),
+        ((3, 44100), (1, 3), (3, 1, 44100)),
+        ((1, 2, 44100), (3, 3), (1, 2, 3, 44100))
```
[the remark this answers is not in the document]
mthrok: Also, can you merge `test_lfilter_shape` into this test?
That way, all the specifications about the shape is in one place and easy to see the whole design.

torchaudio/functional/filtering.py
```diff
@@ -938,20 +941,31 @@ def lfilter(
...
-                                 Must be same size as a_coeffs (pad with 0's as necessary).
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of ``(*, n_order + 1)``.
+                                Where * is the optional number of filter banks.
+                                Lower delays coefficients are first, e.g. ``[b0, b1, b2, ...]``.
+                                Must be same size as a_coeffs (pad with 0's as necessary).
         clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
 
     Returns:
-        Tensor: Waveform with dimension of ``(..., time)``.
+        Tensor: Waveform with dimension of ``(..., *, time)``.
```
mthrok, line 955: I think that, things about shape and filter bank (noted as `*` in the current form) needs more clarification.
Specifically, the distinction between the cases of 1D coeffs and 2D coeffs should be explained.
(even though I think the use of `*` here is very sophisticated.)

mthrok (commented):
Overall, it looks good. The semantic around filter bank looks straightforward though documentation can be improved.

If I understand correctly, the implementation cannot handle the case where one wants to apply different filters per channel. (though I am not sure if that's valid use-case.) Is that correct?

torchaudio/functional/filtering.py
yoyololicon: How about
```suggestion
        Tensor: Waveform with dimension of either ``(..., *, time)`` if ``a_coeffs`` and ``b_coeffs`` are 2D Tensors, 
        or ``(..., time)`` otherwise.
```

yoyololicon:
> Overall, it looks good. The semantic around filter bank looks straightforward though documentation can be improved.
> 
> If I understand correctly, the implementation cannot handle the case where one wants to apply different filters per channel. (though I am not sure if that's valid use-case.) Is that correct?

No, it cannot apply different filters per channel. All signals are applied with same set of filters, regardless.

torchaudio/functional/filtering.py
```diff
@@ -938,20 +941,31 @@ def lfilter(
...
     # pack batch
     shape = waveform.size()
-    waveform = waveform.reshape(-1, shape[-1])
+    assert a_coeffs.size() == b_coeffs.size()
+    assert a_coeffs.ndim <= 2
+
+    waveform = waveform.reshape(-1, 1, shape[-1])
+    if a_coeffs.ndim > 1:
+        shape = shape[:-1] + (a_coeffs.shape[0], shape[-1])
+        waveform = waveform.repeat(1, a_coeffs.shape[0], 1)
```
mthrok, line 965: Instead of repeating the waveform manually _and_ computing the shape manually, how about overwriting `shape` after repeating? I think that will be slightly easier to understand.

```python
waveform = waveform.repeat(...)
shape = waveform.size()
```

torchaudio/functional/filtering.py
mthrok: This looks good, but can we simply replace the `*` with something like `num_filters`?

(And in the docstring of `a_coeffs`, `b_coeffs`, can we explicitly say `either 1D with shape (num_order + 1) or 2D with shape (num_filters, num_order + 1)`?)

The thing is, ellipsis is used in many docstrings, but asterisk is very unique to this function, so I think that explicitly mentioning the difference of 1D and 2D are easier to understand.

mthrok (approved):
Hi @yoyololicon 

Sorry for not getting back sooner.
I think this PR is good, but if you can consider the comments, that would be great. 
Thanks!

yoyololicon:
# Benchmarks
I have managed to build the sources with cuda support using docker container. 
The benchmarks was conducted on a linux machine with i5 4790K, 32G of RAM, and a RTX 3070.
The result indicate that the runtime on CUDA is roughly doubled. 

## Before

### CPU
```
                   |  forward  |  backward
      [32, 256]    |    307.2  |    808.0 
      [32, 1024]   |    614.9  |   1512.9 
      [32, 4096]   |   2036.1  |   5199.6 
      [64, 256]    |    399.5  |    992.8 
      [64, 1024]   |   1027.6  |   2417.2 
      [64, 4096]   |   7028.7  |  13454.1 
      [128, 256]   |    624.4  |   1493.7 
      [128, 1024]  |   1986.3  |   4805.7 
      [128, 4096]  |  17420.1  |  31806.2 
      [32, 256]    |    291.6  |    813.7 
      [32, 1024]   |    531.0  |   1417.3 
      [32, 4096]   |   1754.9  |   4138.1 
      [64, 256]    |    357.4  |    968.6 
      [64, 1024]   |    797.2  |   1973.9 
      [64, 4096]   |   6914.0  |  13116.0 
      [128, 256]   |    535.3  |   1398.3 
      [128, 1024]  |   1522.3  |   4060.9 
      [128, 4096]  |  17863.6  |  30650.4 
      [32, 256]    |    268.3  |    822.9 
      [32, 1024]   |    491.3  |   1395.4 
      [32, 4096]   |   1539.7  |   4215.9 
      [64, 256]    |    347.6  |    996.4 
      [64, 1024]   |    708.5  |   1801.9 
      [64, 4096]   |   7666.7  |  13461.7 
      [128, 256]   |    479.9  |   1306.3 
      [128, 1024]  |   1428.4  |   4137.1 
      [128, 4096]  |  18735.3  |  31495.1 

```

### CUDA

```
                   |  forward  |  backward
      [32, 256]    |    10.7   |    19.6  
      [32, 1024]   |    41.7   |    76.5  
      [32, 4096]   |   159.8   |   297.6  
      [64, 256]    |    10.3   |    19.7  
      [64, 1024]   |    41.8   |    75.9  
      [64, 4096]   |   157.9   |   299.2  
      [128, 256]   |    10.5   |    19.8  
      [128, 1024]  |    41.3   |    75.9  
      [128, 4096]  |   164.3   |   301.0  
      [32, 256]    |    10.6   |    19.6  
      [32, 1024]   |    40.2   |    77.6  
      [32, 4096]   |   161.9   |   302.6  
      [64, 256]    |    10.7   |    19.9  
      [64, 1024]   |    40.0   |    75.6  
      [64, 4096]   |   159.8   |   299.9  
      [128, 256]   |    10.9   |    19.8  
      [128, 1024]  |    41.4   |    75.2  
      [128, 4096]  |   170.6   |   301.2  
      [32, 256]    |    10.8   |    19.6  
      [32, 1024]   |    40.3   |    76.8  
      [32, 4096]   |   160.5   |   301.7  
      [64, 256]    |    10.6   |    19.7  
      [64, 1024]   |    40.3   |    75.5  
      [64, 4096]   |   163.7   |   302.8  
      [128, 256]   |    10.7   |    19.8  
      [128, 1024]  |    42.8   |    75.9  
      [128, 4096]  |   165.5   |   299.8  

Times are in milliseconds (ms).
```

## After

### CPU
```
                   |  forward  |  backward
      [32, 256]    |    322.1  |    835.5 
      [32, 1024]   |    635.2  |   1538.1 
      [32, 4096]   |   2164.4  |   5424.1 
      [64, 256]    |    410.9  |   1026.8 
      [64, 1024]   |   1010.5  |   2299.0 
      [64, 4096]   |   6981.6  |  12721.7 
      [128, 256]   |    616.4  |   1451.6 
      [128, 1024]  |   1886.5  |   4810.8 
      [128, 4096]  |  17436.0  |  31758.8 
      [32, 256]    |    305.4  |    854.4 
      [32, 1024]   |    550.3  |   1464.6 
      [32, 4096]   |   1733.3  |   4914.4 
      [64, 256]    |    370.5  |   1003.6 
      [64, 1024]   |    791.4  |   1927.2 
      [64, 4096]   |   6829.2  |  12608.5 
      [128, 256]   |    536.6  |   1383.8 
      [128, 1024]  |   1429.7  |   4541.2 
      [128, 4096]  |  17876.0  |  31664.5 
      [32, 256]    |    283.0  |    850.6 
      [32, 1024]   |    492.4  |   1446.2 
      [32, 4096]   |   1586.0  |   4792.5 
      [64, 256]    |    362.8  |    978.7 
      [64, 1024]   |    735.8  |   1842.1 
      [64, 4096]   |   7331.0  |  12645.4 
      [128, 256]   |    492.1  |   1351.4 
      [128, 1024]  |   1265.6  |   3954.7 
      [128, 4096]  |  18502.5  |  33391.6 

```

### CUDA

```
                   |  forward  |  backward
      [32, 256]    |    19.6   |    37.9  
      [32, 1024]   |    77.4   |   148.7  
      [32, 4096]   |   314.3   |   596.4  
      [64, 256]    |    19.7   |    38.5  
      [64, 1024]   |    82.8   |   147.6  
      [64, 4096]   |   328.1   |   589.7  
      [128, 256]   |    20.5   |    38.0  
      [128, 1024]  |    82.3   |   148.7  
      [128, 4096]  |   323.4   |   595.7  
      [32, 256]    |    19.9   |    37.9  
      [32, 1024]   |    77.4   |   150.9  
      [32, 4096]   |   308.5   |   597.6  
      [64, 256]    |    19.7   |    38.6  
      [64, 1024]   |    82.7   |   147.6  
      [64, 4096]   |   330.5   |   590.1  
      [128, 256]   |    20.8   |    38.2  
      [128, 1024]  |    80.9   |   149.3  
      [128, 4096]  |   330.8   |   595.9  
      [32, 256]    |    19.8   |    37.9  
      [32, 1024]   |    78.2   |   150.8  
      [32, 4096]   |   309.5   |   598.0  
      [64, 256]    |    20.1   |    38.1  
      [64, 1024]   |    82.6   |   148.9  
      [64, 4096]   |   323.4   |   589.2  
      [128, 256]   |    20.9   |    38.2  
      [128, 1024]  |    81.4   |   148.7  
      [128, 4096]  |   322.0   |   596.0  

Times are in milliseconds (ms).
```

mthrok:
Hi @yoyololicon 

Thanks for the benchmark.
The shapes in the table, are they the size of input batch?
I am also curious to learn the performance difference of multiple filters.

What do you think is the feasibility of splitting CPU/CUDA code?
I assume that the extra overhead for CUDA case is coming from the newly added for-loop.

BTW, I found an attempt to use RNN in https://github.com/pytorch/audio/issues/226. Although it is also pointed out that RNN without activation would not improve the performance as much. https://discuss.pytorch.org/t/vanillarnn-without-activation-function/26507/2. I wonder if this could help us in custom CUDA `lfilter` code.

yoyololicon:
Hi @mthrok ,

> Thanks for the benchmark.
> The shapes in the table, are they the size of input batch?
> I am also curious to learn the performance difference of multiple filters.

Yes, they are the size of input. I can do the multiple filters case later.

> 
> What do you think is the feasibility of splitting CPU/CUDA code?
> I assume that the extra overhead for CUDA case is coming from the newly added for-loop.

The overhead concern is technically true, although the actual implementation is done by replacing `addmv` and a for-loop with `bmm` and some extra transpose operations.
To completely avoid these overheads, I think writing custom cuda kernels is needed.

> 
> BTW, I found an attempt to use RNN in #226. Although it is also pointed out that RNN without activation would not improve the performance as much. https://discuss.pytorch.org/t/vanillarnn-without-activation-function/26507/2. I wonder if this could help us in custom CUDA `lfilter` code.

Hmm, I can't see much potentials on this direction of using RNN,  in our case **for now**, cuz RNN is strictly a case of second-order IIR, and cannot generalize to arbitrary order. 
I do see some possibilities of using RNN if we can implement lfilter in [state space formulation](https://ccrma.stanford.edu/~jos/filters/State_Space_Realization.html), but I haven't look into it carefully yet.

mthrok:
@yoyololicon Sorry for not getting back to you. I am very busy these days and I should be able to get back to this at the end of this week.

mthrok (approved):

mthrok:
@yoyololicon Again, really sorry for taking a long time. Let's continue to the next step.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -67,18 +67,22 @@ def test_lfilter_clamp(self):
...
-        waveform = torch.rand(*shape, dtype=self.dtype, device=self.device)
-        b_coeffs = torch.tensor([0, 0, 0, 1], dtype=self.dtype, device=self.device)
-        a_coeffs = torch.tensor([1, 0, 0, 0], dtype=self.dtype, device=self.device)
+        waveform = torch.rand(*input_shape, dtype=self.dtype, device=self.device)
+        b_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
+        a_coeffs = torch.rand(*coeff_shape, dtype=self.dtype, device=self.device)
         output_waveform = F.lfilter(waveform, a_coeffs, b_coeffs)
-        assert shape == waveform.size() == output_waveform.size()
+        assert input_shape == waveform.size()
+        assert target_shape == output_waveform.size()
```
mthrok, line 86: Follow-up: We should add tests to verify that the resulting signals are same whether multiple filters are applied individually or as a bank.

yoyololicon: Got it, will add batch consistency tests and also benchmarks of multiple filters in later commits.

PR #1591, opened 2021-06-18
Add return type for RNNT loss

carolineechen:
add return type in documentation for rnnt loss


mthrok (approved):

torchaudio/prototype/rnnt_loss.py
```diff
@@ -98,6 +101,9 @@ def forward(
             targets (Tensor): Tensor of dimension (batch, max target length) containing targets with zero padded
             logit_lengths (Tensor): Tensor of dimension (batch) containing lengths of each sequence from encoder
             target_lengths (Tensor): Tensor of dimension (batch) containing lengths of targets for each sequence
+
+        Returns:
+            Tensor: loss with the reduction option applied.
```
mthrok, line 106: Can you add description about the shape?

Example in CE loss https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html
>> scalar. If reduction is 'none', then the same size as the target:  ...

vincentqb (approved):
LGTM

PR #1601, opened 2021-06-21
Porting edit distance from torchaudio.metrics to torchaudio.functional

yangarbiter:
Closes #927.

I am porting edit distance from torchaudio.metrics to torchaudio.functional.

There are two things I would need comments on:
1. Currently, metrics.py is removed. Do we need to maintain backwards competibility for metrics.levenshtein_distance?
2. Currently, the implementations for character_edit_distance and word_edit_distance are duplicated because that torchscript does not supported Union[str, List[str]] type. I am wondering if there is a way to get around it (see below for more details).

See [link](https://github.com/pytorch/audio/pull/1601/files#diff-f69839488ce70c39d8c3db8c87790ac963364faf8aefcb38ff5b026ee9dca8c7R1451) for the current implementation. With the following implementation for character_edit_distance and word_edit_distance, I would get `General Union types are not currently supported. Only Union[T, NoneType] (i.e. Optional[T]) is supported.` (see Error message 1 below for the full error)

```python
def _edit_distance(r: Union[str, List[str]], h: Union[str, List[str]]) -> int:
    # Initialisation
    dold = [i for i in range(len(h)+1)]
    dnew = [0 for _ in range(len(h)+1)]

    # Computation
    for i in range(1, len(r) + 1):
        dnew[0] = i
        for j in range(1, len(h) + 1):
            if r[i - 1] == h[j - 1]:
                dnew[j] = dold[j - 1]
            else:
                substitution = dold[j - 1] + 1
                insertion = dnew[j - 1] + 1
                deletion = dold[j] + 1
                dnew[j] = min(substitution, insertion, deletion)

        dnew, dold = dold, dnew

    return int(dold[-1])


def character_edit_distance(r: str, h: str) -> int:
    return _edit_distance(r, h)


def word_edit_distance(r: List[str], h: List[str]) -> int:
    return _edit_distance(r, h)
```

If I removed the types for `_edit_distance` like below:
```python
def _edit_distance(r, h) -> int:
```
The types for r and h are inferred as Tensor and I would also get an error (`Expected a value of type 'Tensor (inferred)' for argument 'r' but instead found type 'str'. Inferred 'r' to be of type 'Tensor' because it was not annotated with an explicit type.` See error message 2 for more details).

I am wondering if there is a way to get around this without duplicating the implementation.

Error message 1:
```
E           RuntimeError:
E           General Union types are not currently supported. Only Union[T, NoneType] (i.e. Optional[T]) is supported.:
E             File "/Users/yyyang/miniconda3/envs/torchaudio/lib/python3.9/site-packages/torchaudio-0.10.0a0+0eae456-py3.9-macosx-10.9-x86_64.egg/torchaudio/functional/functional.py", line 1451
E           def _edit_distance(r: Union[str, List[str]], h: Union[str, List[str]]) -> int:
E                                 ~~~~~~~~~~~~~~~~~~~~~ <--- HERE
E               """
E               """
E           '_edit_distance' is being compiled since it was called from 'character_edit_distance'
E             File "/Users/yyyang/miniconda3/envs/torchaudio/lib/python3.9/site-packages/torchaudio-0.10.0a0+0eae456-py3.9-macosx-10.9-x86_64.egg/torchaudio/functional/functional.py", line 1492
E                   int: The distance between the reference and the hypothesis.
E               """
E               return _edit_distance(r, h)
E                      ~~~~~~~~~~~~~~~~~~~ <--- HERE
E           'character_edit_distance' is being compiled since it was called from 'func'
E             File "/Users/yyyang/levenshtein/audio/test/torchaudio_unittest/functional/torchscript_consistency_impl.py", line 601
E                       ref = "abc"
E                       hyp = "bcd"
E                       return F.character_edit_distance(ref, hyp)
E                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
``` 

Error message 2:
```
E           RuntimeError:
E
E           _edit_distance(Tensor r, Tensor h) -> (int):
E           Expected a value of type 'Tensor (inferred)' for argument 'r' but instead found type 'str'.
E           Inferred 'r' to be of type 'Tensor' because it was not annotated with an explicit type.
E           :
E             File "/Users/yyyang/miniconda3/envs/torchaudio/lib/python3.9/site-packages/torchaudio-0.10.0a0+0eae456-py3.9-macosx-10.9-x86_64.egg/torchaudio/functional/functional.py", line 1492
E                   int: The distance between the reference and the hypothesis.
E               """
E               return _edit_distance(r, h)
E                      ~~~~~~~~~~~~~~ <--- HERE
E           'character_edit_distance' is being compiled since it was called from 'func'
E             File "/Users/yyyang/levenshtein/audio/test/torchaudio_unittest/functional/torchscript_consistency_impl.py", line 601
E                       ref = "abc"
E                       hyp = "bcd"
E                       return F.character_edit_distance(ref, hyp)
E                              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
```


torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+
+
+def character_edit_distance(r: str, h: str) -> int:
+    """
+    Calculate the character level edit (Levenshtein) distance between two strings.
+
+    The function computes an edit distance allowing deletion, insertion and substitution.
+    The result is an integer. Users may want to normalize by the length of the reference.
+
+    This can be used to compute the edit distance for instance between two strings,
```
carolineechen, line 1458: Is this statement still true now that it's split into two functions? This flexibility also makes it inconsistent with the function typing and documentation

examples/pipeline_wav2letter/metrics.py
```diff
@@ -1,38 +0,0 @@
-from typing import List, Union
-
-
-def levenshtein_distance(r: Union[str, List[str]], h: Union[str, List[str]]):
```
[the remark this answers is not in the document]
yangarbiter: I just realized that this function is implemented in the examples folder. Do we still want to maintain it's backward compatibility?

test/torchaudio_unittest/metrics.py
```diff
@@ -0,0 +1,30 @@
+from parameterized import parameterized
+from torchaudio.functional import character_edit_distance, word_edit_distance
+from torchaudio_unittest import common_utils
+
+
+class TestLevenshteinDistance(common_utils.TorchaudioTestCase):
```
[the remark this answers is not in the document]
yangarbiter: Thanks for the suggestion!

torchaudio/functional/functional.py
yangarbiter: I think this is still true as now we are just splitting up charter and word level edit distance. You may delete, insert or substitute  a charter or a word at a time.

examples/pipeline_wav2letter/metrics.py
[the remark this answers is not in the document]
mthrok: It is okay to make changes in examples since they are not part of released packages.

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -593,6 +593,26 @@ def func(tensor):
...
 
+    def test_character_edit_distance(self):
+
+        def func(tensor):
+            ref = "abc"
+            hyp = "bcd"
+            return F.character_edit_distance(ref, hyp)
+
+        dummy = torch.zeros(1, 1)
+        self._assert_consistency(func, dummy)
```
mthrok, line 604: Please do not use `_assert_consistency` here.
It is not designed for functions that do not accept tensor.
Please implement the logic directly in the test method.
Also instead of hard coding `ref` and `hyp`, make them function argument in `func`.

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -593,6 +593,26 @@ def func(tensor):
...
+
+    def test_word_edit_distance(self):
+
+        def func(tensor):
+            ref = ["hello", "world"]
+            hyp = ["hello", "planet"]
+            return F.word_edit_distance(ref, hyp)
+
+        dummy = torch.zeros(1, 1)
+        self._assert_consistency(func, dummy)
```
mthrok, line 614: Similar to the above, instead of `_assert_consistency`, implement the correct logic.

torchaudio/functional/functional.py
mthrok: I think what @carolineechen meant is that the following part is inconsistent with what type annotation / docstring says.

```
This can be used ... or two list of words. Note that, if a string and a list of words is provided, the distance
     will be computed between the "string" sequence, and the "list of words" sequence.
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
                                               resampling_method, beta, waveform.device, waveform.dtype)
     resampled = _apply_sinc_resample_kernel(waveform, orig_freq, new_freq, gcd, kernel, width)
     return resampled
+
+
+def character_edit_distance(r: str, h: str) -> int:
```
mthrok, line 1451: Please give arguments proper, readable, easy-to-understand names.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+    will be computed between the "string" sequence, and the "list of words" sequence.
+
+    Args:
+        r (str): the reference string to compare.
+        h (str): the hypothesis string (e.g. the predicted string) to compare.
+    Returns:
+        int: The distance between the reference and the hypothesis.
+    """
+
+    # Initialisation
```
mthrok, line 1469: Please remove these comments, that are too obvious and do not provide any additional information.
The code should be self-explanatory.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+                insertion = dnew[j - 1] + 1
+                deletion = dold[j] + 1
+                dnew[j] = min(substitution, insertion, deletion)
+
+        dnew, dold = dold, dnew
+
+    return int(dold[-1])
+
+
+def word_edit_distance(r: List[str], h: List[str]) -> int:
```
mthrok, line 1490: Please give arguments proper, readable, easy-to-understand names.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+    """
+    Calculate the word level edit (Levenshtein) distance between two sentences.
+    The sentences should be represented in a list of strings.
+
+    The function computes an edit distance allowing deletion, insertion and substitution.
+    The result is an integer. Users may want to normalize by the length of the reference.
+
+    This can be used to compute the edit distance for instance between two strings,
+    or two list of words. Note that, if a string and a list of words is provided, the distance
+    will be computed between the "string" sequence, and the "list of words" sequence.
```
mthrok, line 1500: Same as the above discussion, this paragraph is inconsistent with docstring/type annotation.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+
+    Args:
+        r (list of str): the reference sentence to compare.
+        h (list of str): the hypothesis sentence (e.g. the predicted string) to compare.
+    Returns:
+        int: The distance between the reference and the hypothesis.
+    """
+
+    # Initialisation
```
mthrok, line 1509: Same as the above. Please remove these comments, that are too obvious and do not provide any additional information.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+
+    Args:
+        r (str): the reference string to compare.
+        h (str): the hypothesis string (e.g. the predicted string) to compare.
+    Returns:
+        int: The distance between the reference and the hypothesis.
+    """
+
+    # Initialisation
+    dold = [i for i in range(len(h)+1)]
```
mthrok, line 1470: Using intermediate variable `i` without extra computation is redundant. 
```suggestion
    dold = list(range(len(h)+1))
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+    dnew = [0 for _ in range(len(h)+1)]
```
mthrok, line 1471: Since it's a list of primitive, the list can be initialized without creating intermediate variable, which is never used.

```suggestion
    dnew = [0] * (len(h) + 1)
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
+
+    # Computation
+    for i in range(1, len(r) + 1):
+        dnew[0] = i
+        for j in range(1, len(h) + 1):
```
mthrok, line 1476: `len(h) + 1` is called three times. Please put it in a variable.

mthrok (commented):
Could you add a reference to the definition of these distances?
Without reference, I cannot tell if the current test coverage is enough.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,82 @@ def resample(
...
     resampled = _apply_sinc_resample_kernel(waveform, orig_freq, new_freq, gcd, kernel, width)
     return resampled
+
+
+def character_edit_distance(r: str, h: str) -> int:
+    """
+    Calculate the character level edit (Levenshtein) distance between two strings.
+
+    The function computes an edit distance allowing deletion, insertion and substitution.
+    The result is an integer. Users may want to normalize by the length of the reference.
```
mthrok, line 1456: Can you also mention if this distance is symmetric or asymmetric?

mthrok: If the distance is symmetric, I think function arguments should not be called hypothesis and reference. That is the context of caller side and it does not concern in this function.

mthrok:
I am not entirely sure if TorchScript support is necessary, especially if it is forcing the duplicated implementation.
What is the reason / motivation behind TS support other than "because other functions support"?

torchaudio/functional/functional.py
mthrok: oh maybe it's implemented this way because of TS-support.

torchaudio/functional/functional.py
mthrok: oh maybe it's implemented this way because of TS-support.

mthrok:
In `character_edit_distance`, how about converting the given strings to list of characters and pass it to `word_edit_distance`?

vincentqb:
> I am not sure if TorchScript support is necessary, especially if it is forcing the duplicated implementation.
> What is the reason / motivation behind TS support other than "because other functions support"?

The cost of two exact copies is actually very low, especially if we add a comment in code justifying why that is. The cost is not even visible to the end user, and yet the benefit is a consistent experience and compatibility across the library. Do you see other costs @mthrok?

@yangarbiter -- Please confirm that there is no significant difference in runtime between python and jitted version (and share the numbers).

mthrok:
> > I am not sure if TorchScript support is necessary, especially if it is forcing the duplicated implementation.
> > What is the reason / motivation behind TS support other than "because other functions support"?
> 
> The cost of two exact copies is actually very low, especially

Actually, no. The maintenance cost of duplicated code is never low. In software engineering, having the duplicated code is something that should be avoided. It is rare that duplicated code is justified. Which is why there is saying like [Don't Repeat Yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) or [Rule of three/five/zero](https://en.cppreference.com/w/cpp/language/rule_of_three).

This is why we need a very good reason for supporting TS in this function. Firstly, when considering the research to production story, I do not know a production system that has to have edit distance. In when evaluating ASR system, typically, transcription is generated and saved somewhere for the later reference, and then some metrics are calculated. Secondly, these functions do not handle a Tensor and they only manipulate [list of] strings. So it is not as hard to write in other languages. PyTorch's research to production story is important because otherwise it is not straightforward to deploy models trained on PyTorch, and when deploying model, the preprocessing should be accompanied so that the same preprocessing can be applied. Now edit distance does not fit in this picture.

Therefore again I ask, what is the definitive reason that edit-distance has to support TorchScript?

vincentqb:
On torchscriptability for the edit distance, the advantage is general consistency of the library and the ecosystem -- which can lead to new application (I could imagine a scenario where a user would like to monitor WER in production for instance along with the rest of a code that was jitted.). That being said, I'm open to not supporting torchscript at this point for the edit distance, if no satisfactory solution is found. Further user feedback on this issue would help.

Let's take a step back though and list what options we have.

1. Two functions with the same code but different signature. This is what is currently done in the pull request. This supports torchscript, but duplicates code.
2. One main function and two wrapper exposing different signatures. As mentioned in the description, this is not torchscriptable.
3. One edit distance function, annotated with List[str] only. When List[str] is used, this is WER. When str is used in python, this is CER. This is torchscriptable and support List[str] when torchscripted, and only has one code. In the documentation, we can add that it also supports str cases in python, but for torchscriptability reason is only annotated with List[str]. We do that somewhere else in the code, see below. (as noted by @yangarbiter we would need an assert to make sure both are the same type when ran in python)
4. @suo -- do you have another solution for exposing Union[str, List[str]] implementations to torchscript? thanks!

As noted by @yangarbiter, in the future, we might want to migrate this or the for-loop to C++ as done for lfilter. This may also affect which option we go too.

On 3:
```
./backend/sox_io_backend.py:                  * This argument is intentionally annotated as ``str`` only due to
./backend/sox_io_backend.py:            Note: This argument is intentionally annotated as ``str`` only due to
./backend/sox_io_backend.py:            This function also handles ``pathlib.Path`` objects, but is annotated
./backend/soundfile_backend.py:        ``filepath`` argument is intentionally annotated as ``str`` only, even though it accepts
./backend/soundfile_backend.py:        which has a restriction on type annotation due to TorchScript compiler compatiblity.
./backend/soundfile_backend.py:        ``filepath`` argument is intentionally annotated as ``str`` only, even though it accepts
./backend/soundfile_backend.py:        which has a restriction on type annotation due to TorchScript compiler compatiblity.
./backend/soundfile_backend.py:        ``filepath`` argument is intentionally annotated as ``str`` only, even though it accepts
./backend/soundfile_backend.py:        which has a restriction on type annotation due to TorchScript compiler compatiblity.
./sox_effects/sox_effects.py:            Note: This argument is intentionally annotated as ``str`` only for
```

torchaudio/functional/functional.py
yangarbiter: Yes, this is for TS-support. See [this internal link](https://colab.research.google.com/drive/12QfAEq46Ra0wkQLyoHid7ZtK9kr6Ixdp?usp=sharing)

torchaudio/functional/functional.py
yangarbiter: Yes, this is for TS-support. See [this internal link](https://colab.research.google.com/drive/12QfAEq46Ra0wkQLyoHid7ZtK9kr6Ixdp?usp=sharing). Torchscript doesn't support generators.

torchaudio/functional/functional.py
yangarbiter: I'll remove them. Thanks :).

torchaudio/functional/functional.py
yangarbiter: I'll remove them. Thanks :)

torchaudio/functional/functional.py
yangarbiter: I'll remove them. Thanks :)

yangarbiter (commented):
Here is the [internal link](https://colab.research.google.com/drive/12QfAEq46Ra0wkQLyoHid7ZtK9kr6Ixdp?usp=sharing) to the performance between python and jitted version.

torchaudio/functional/functional.py
yangarbiter: Yes, it is symmetric, I'll change the name too string1 and string2 for character_edit_distance and sentence1 and sentence2 for word_edit_distance.

torchaudio/functional/functional.py
yangarbiter: I've changed it to string1 and string2.

torchaudio/functional/functional.py
yangarbiter: I've changed to sentence1 and sentence 2.

torchaudio/functional/functional.py
yangarbiter: It is fixed [here](https://github.com/pytorch/audio/pull/1601/files#diff-f69839488ce70c39d8c3db8c87790ac963364faf8aefcb38ff5b026ee9dca8c7R1498)

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -382,6 +382,31 @@ def test_phase_vocoder_shape(self, rate, test_pseudo_complex):
...
+            ["abc", "", 3],
+            ["", "", 0],
+            ["abc", "abc", 0],
+            ["aaa", "aba", 1],
+            ["aba", "aaa", 1],
+            ["aa", "aaa", 1],
+            ["aaa", "aa", 1],
+            ["abc", "bcd", 2],
+        ]
+    )
```
mthrok, line 396: Can you organize the test cases;
1. Split test methods for `equal`, `deletion`, `substitution`.
1. For the case of `deletion` and `substitution` methods, check the swapped case inside of the method.
1. For the `deletion` and `substitution` methods, add cases where there are 1, 2 and 3 occurrences and make sure that they work even when the event occurs at the beginning and/or end of string.
1. For all the methods, please include the cases with empty string and string with multi-byte (unicode) characters, with some strings that are under [unicode equivalence](https://en.wikipedia.org/wiki/Unicode_equivalence).

yangarbiter:
Thanks to @mthrok's suggestion, we have a way to refactor the duplicated implementation (see this [link](https://github.com/pytorch/audio/pull/1601/files#diff-f69839488ce70c39d8c3db8c87790ac963364faf8aefcb38ff5b026ee9dca8c7R1485))

Also, I did a quick test on leetcode and it seems our current implementation is correct.
Screen Shot 2021-06-22 at 4 07 28 PM

I still add a few more tests to make sure we can still test things well in the future. However, the torchscript is failing for the unicode case. The test cases that failed are the torchscript consistency tests. More specifically test_character_edit_distance with input `["ᑌᑎI", "ᑌᑎIᑕO"]` and `["0ᑌᑎI", "ᑌᑎIᑕO"]`. I am not sure how unicode is represented in the torchscript. Does anyone have suggestions for debugging? Please see below for the error messages.

```
(torchaudio) 08:32:29 [130] @yyyang-mbp ➜ audio/test/torchaudio_unittest git:(levenshtein) ✗ 0s & pytest functional/torchscript_consistency_cpu_test.py
================================================================================================= test session starts =================================================================================================
platform darwin -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /Users/yyyang/levenshtein/audio
plugins: anyio-2.2.0, xdist-2.3.0, forked-1.3.0
collected 134 items

functional/torchscript_consistency_cpu_test.py ...............F....F...............................................s.ssssss......F....F..s....s.....s...sss.....s.....s....s.........                           [100%]

====================================================================================================== FAILURES =======================================================================================================
______________________________________________________________________________ TestFunctionalFloat32.test_character_edit_distance_06__I _______________________________________________________________________________

a = (<torchaudio_unittest.functional.torchscript_consistency_cpu_test.TestFunctionalFloat32 testMethod=test_character_edit_distance_06__I>,)

    @wraps(func)
    def standalone_func(*a):
>       return func(*(a + p.args), **p.kwargs)

../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/parameterized/parameterized.py:533:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
functional/torchscript_consistency_impl.py:622: in test_character_edit_distance
../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1419: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing 4 and 2 gives a difference of 2, but the allowed difference with rtol=0 and atol=0 is only 0!
______________________________________________________________________________ TestFunctionalFloat32.test_character_edit_distance_11_0_I ______________________________________________________________________________

a = (<torchaudio_unittest.functional.torchscript_consistency_cpu_test.TestFunctionalFloat32 testMethod=test_character_edit_distance_11_0_I>,)

    @wraps(func)
    def standalone_func(*a):
>       return func(*(a + p.args), **p.kwargs)

../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/parameterized/parameterized.py:533:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
functional/torchscript_consistency_impl.py:622: in test_character_edit_distance
../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1419: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing 5 and 3 gives a difference of 2, but the allowed difference with rtol=0 and atol=0 is only 0!
______________________________________________________________________________ TestFunctionalFloat64.test_character_edit_distance_06__I _______________________________________________________________________________

a = (<torchaudio_unittest.functional.torchscript_consistency_cpu_test.TestFunctionalFloat64 testMethod=test_character_edit_distance_06__I>,)

    @wraps(func)
    def standalone_func(*a):
>       return func(*(a + p.args), **p.kwargs)

../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/parameterized/parameterized.py:533:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
functional/torchscript_consistency_impl.py:622: in test_character_edit_distance
../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1419: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing 4 and 2 gives a difference of 2, but the allowed difference with rtol=0 and atol=0 is only 0!
______________________________________________________________________________ TestFunctionalFloat64.test_character_edit_distance_11_0_I ______________________________________________________________________________

a = (<torchaudio_unittest.functional.torchscript_consistency_cpu_test.TestFunctionalFloat64 testMethod=test_character_edit_distance_11_0_I>,)

    @wraps(func)
    def standalone_func(*a):
>       return func(*(a + p.args), **p.kwargs)

../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/parameterized/parameterized.py:533:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
functional/torchscript_consistency_impl.py:622: in test_character_edit_distance
../../../../miniconda3/envs/torchaudio/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1419: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing 5 and 3 gives a difference of 2, but the allowed difference with rtol=0 and atol=0 is only 0!
================================================================================================== warnings summary ===================================================================================================

..... omitted warnings

=============================================================================================== short test summary info ===============================================================================================
FAILED functional/torchscript_consistency_cpu_test.py::TestFunctionalFloat32::test_character_edit_distance_06__I - AssertionError: False is not true : Scalars failed to compare as equal! Comparing 4 and 2 gives a...
FAILED functional/torchscript_consistency_cpu_test.py::TestFunctionalFloat32::test_character_edit_distance_11_0_I - AssertionError: False is not true : Scalars failed to compare as equal! Comparing 5 and 3 gives ...
FAILED functional/torchscript_consistency_cpu_test.py::TestFunctionalFloat64::test_character_edit_distance_06__I - AssertionError: False is not true : Scalars failed to compare as equal! Comparing 4 and 2 gives a...
FAILED functional/torchscript_consistency_cpu_test.py::TestFunctionalFloat64::test_character_edit_distance_11_0_I - AssertionError: False is not true : Scalars failed to compare as equal! Comparing 5 and 3 gives ...
=============================================================================== 4 failed, 114 passed, 16 skipped, 10 warnings in 11.98s ===============================================================================
```

test/torchaudio_unittest/functional/functional_impl.py
yangarbiter: I've added these tests and grouped them. However, we may have problem for unicode with torchscript. Please see below.

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
yangarbiter: I've fixed it in the new commit.

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
yangarbiter: Thanks for the suggestion, I've fixed them in the new commit.

torchaudio/functional/functional.py
yangarbiter: Thanks for pointing it out. I've removed these comments :)

mthrok:
> I still add a few more tests to make sure we can still test things well in the future. However, the torchscript is failing for the unicode case. The test cases that failed are the torchscript consistency tests. More specifically test_character_edit_distance with input `["ᑌᑎI", "ᑌᑎIᑕO"]` and `["0ᑌᑎI", "ᑌᑎIᑕO"]`. I am not sure how unicode is represented in the torchscript. Does anyone have suggestions for debugging? Please see below for the error messages.
> 

This is expected. The treatment of unicode is very complicated and often it's the implementation detail. Different language/framework/applications have different needs/standards to compare unicode strings.

Rust's tutorial gives some insight about how unicode is complicated. Rust and C++ are different but the structure is somewhat similar, so the same difficulty applies.
https://doc.rust-lang.org/book/ch08-02-strings.html#bytes-and-scalar-values-and-grapheme-clusters-oh-my
> Another point about UTF-8 is that there are actually three relevant ways to look at strings from Rust’s perspective: as bytes, scalar values, and grapheme clusters (the closest thing to what we would call letters).

Python has its own logic for comparing unicode characters and that cannot be translated to C++ via TorchScript, which means that string comparison cannot be TorchScripted.

vincentqb (commented):
Thanks @mthrok for the suggestions [here](https://github.com/pytorch/audio/pull/1601#issuecomment-865994136) and [here](https://github.com/pytorch/audio/pull/1601#issuecomment-866972482), I've added to the [list](https://github.com/pytorch/audio/pull/1601#issuecomment-866250514) for completeness.

> I still add a few more tests to make sure we can still test things well in the future. However, the torchscript is failing for the unicode case. The test cases that failed are the torchscript consistency tests. More specifically test_character_edit_distance with input `["ᑌᑎI", "ᑌᑎIᑕO"]` and `["0ᑌᑎI", "ᑌᑎIᑕO"]`. I am not sure how unicode is represented in the torchscript. Does anyone have suggestions for debugging?

Does torchtext handle unicode in strings, @parmeet? How about torchscripting in this case? cc @suo also

vincentqb:
> My guess is that when scripting a string comparison operation, it is translated to byte comparison of C++, which is a different operation. which means this function is not TorchScriptable, which means we do not need to care about the TorchScript-compatibility and we can use one function implementation and annotate the function as `Sequence`.

I agree that this pull request does not need to be blocked on resolving this issue. We can mark as follow-up to hear from @suo and @parmeet on this. In the meanwhile, using the original `Sequence`, as you also suggest here, would be appropriate.

I recommend adding a check that the two sequences are of the same type to avoid confusion like comparing ["Hello"] and "Hello".

vincentqb:
> > My guess is that when scripting a string comparison operation, it is translated to byte comparison of C++, which is a different operation. which means this function is not TorchScriptable, which means we do not need to care about the TorchScript-compatibility and we can use one function implementation and annotate the function as `Sequence`.

as discussed with @yangarbiter offline: i meant to add that I would also be ok with one `edit_distance` annotated with  `List[str]` if we provide two examples in the docstring on how to use each for word/char error rate using it. this has the advantage of 
- reminding/hinting the user to make sure both inputs are of the same type
- supporting torchscript when strings are not unicode (over Seq annotation).

@mthrok, what do you think of this proposal?

yangarbiter:
We will take the one function implementation and annotate the function as `Sequence` route. I've made the changes in the latest commit. Please check.

mthrok:
> as discussed with @yangarbiter offline: i meant to add that I would also be ok with one `edit_distance` annotated with `List[str]` if we provide two examples in the docstring on how to use each for word/char error rate using it. this has the advantage of
> 
> * reminding/hinting the user to make sure both inputs are of the same type
> * supporting torchscript when strings are not unicode (over Seq annotation).
> 
> @mthrok, what do you think of this proposal?

I do not see why it should be limited to `List[str]` if the use case of char-level distance is expected.

I think we should make sure that this function does not support TorchScript. Accidentally scripting the function will cause a situation where it returns wrong result and slower execution due to unnecessary string copy. That will leave users very confusing situation. So I propose to use `torch.jit.unused` decorator, in combination of `Sequence` type.

mthrok:
> I recommend adding a check that the two sequences are of the same type to avoid confusion like comparing ["Hello"] and "Hello".

I am not sure how this can be enforced for containers. Also since we are free from TorchScript, which is strongly-typed language, we should take advantage of duck-typing here. Anything that resemble sequence is accepted. We should explain in the docstring that valid usage would be two sequences of same types, but it's up to users to decide what to pass and it's users' responsibility to pass the valid arguments.

yangarbiter:
@mthrok If the valid usages are passing in str or List[str], shouldn't we just set the type to Union[str, List[str]] along with the `torch.jit.unused` decorator?

yangarbiter:
I just added the `torch.jit.unused` decorator and updated the descriptions in the docstring.

mthrok:
Regarding `List[str]` vs `Sequence`, I think `bytes` (byte string) and `tuple` are also valid use case, and there should be more that I am not aware at the moment. So using `Sequence` will be simple.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,55 @@ def resample(
...
+
+    Args:
+        string1 (str): the first string to compare.
+        string2 (str): the second string to compare.
+    Returns:
+        int: The distance between the first and second strings.
+    """
+    sentence1 = [s for s in string1]
+    sentence2 = [s for s in string2]
```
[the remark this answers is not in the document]
parmeet: @vincentqb  Yes, we do have unicode support in torchtext Vocab that is also torchscriptable. I am not sure what the context is, could you be more specific what you need to know? 

In general, C++ is encoding agnostic. When you pass a str object from python to C++, that str need to be encoded. Pybind11 do this implicitly for you (by default encode in UTF-8). With Pybind11 you can even pass the str object reference and do the encoding in C++ that save you a copy operation (refer to code [here](https://github.com/pytorch/text/blob/master/torchtext/csrc/register_bindings.cpp#L115)). for torchscipting, torchbind registration mechanism would do the job as it support string types (refer to code [here](https://github.com/pytorch/text/blob/master/torchtext/csrc/register_bindings.cpp#L243)).

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,45 @@ def resample(
...
     resampled = _apply_sinc_resample_kernel(waveform, orig_freq, new_freq, gcd, kernel, width)
     return resampled
+
+@torch.jit.unused
+def edit_distance(seq1: Sequence, seq2: Sequence) -> int:
+    """
+    Calculate the word level edit (Levenshtein) distance between two seqences.
+
+    The function computes an edit distance allowing deletion, insertion and
+    substitution.  The result is an integer.
```
vincentqb, line 1456: nit:

```suggestion
    substitution. The result is an integer.
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,45 @@ def resample(
...
+
+    For valid usage, the two input sequences should be the same type.If a pair
```
vincentqb, line 1458: nit:
```suggestion
    For valid usage, the two input sequences should be the same type. If a pair
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,45 @@ def resample(
...
+    strings are given, the output would be the edit distance between two words
+    (character edit distance); if two lists of strings are given, the output
+    would be the edit distance between sentences (word edit distance). Users
+    may want to normalize the output by the length of the reference sequence.
+
+    This function is not type annotated, so torchscipt is not supported for this
+    function.
```
vincentqb, line 1465: nit:

```suggestion
    torchscipt is not supported for this function.
```

yangarbiter:
Hi @mthrok, I think I've addressed every comments. Please let me know if I missed any of them and if the PR is ready to merge.
Thanks.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -382,6 +382,43 @@ def test_phase_vocoder_shape(self, rate, test_pseudo_complex):
...
+            [["Tᕮ᙭T", "XD"], ["world", "hello"], 2],
+            [["", "XD"], ["world", ""], 2],
+            ["aba", "   ", 3],
+
+            [["hello", "world"], ["world", "hello", "!"], 2], # mix deletion and substitution
+            [["Tᕮ᙭T", "world", "LOL", "XD"], ["world", "hello", "ʕ•́ᴥ•̀ʔっ"], 3],
+        ]
+    )
+    def test_simple_case_edit_distance(self, ref, hyp, distance):
+        assert F.edit_distance(ref, hyp) == distance
```
mthrok, line 420: Some cases are tested for flipped version but some are not. Can you check the flipped version as well?
Also can you rename `ref`/`hyp` to `seq1`/`seq2`?

```python
assert F.edit_distance(seq1, seq2) == distance
assert F.edit_distance(seq2, seq1) == distance
```

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,45 @@ def resample(
                                               resampling_method, beta, waveform.device, waveform.dtype)
     resampled = _apply_sinc_resample_kernel(waveform, orig_freq, new_freq, gcd, kernel, width)
     return resampled
+
```
mthrok, line 1449: There should be two lines between function definitions.

torchaudio/functional/__init__.py
```diff
@@ -88,4 +89,5 @@
     'vad',
     'apply_codec',
     'resample',
+    "edit_distance",
```
mthrok, line 92: Can you use single quotation for the consistency with the others?

mthrok: In `functional.py`, most of them use double quotations. so it's play to leave it as-is. (and please do not try to fix the others that's not the concern of this PR)

mthrok (commented):
Mostly looks good. Some minor style issues. Please fix them.

https://app.circleci.com/pipelines/github/pytorch/audio/6447/workflows/fbf5cd2a-dd9d-49c3-9cca-22efc3fbe1ee/jobs/248436

test/torchaudio_unittest/functional/functional_impl.py
yangarbiter: Thanks for the reminder. I've fixed these [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR421).

torchaudio/functional/functional.py
yangarbiter: I've addressed this [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR422). I've also run `flake8` and added one more space before inline comments [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR407) and change the list comprehension to list function [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR407). Please check.

torchaudio/functional/__init__.py
yangarbiter: This is addressed [here](https://github.com/pytorch/audio/pull/1601/files#diff-efef667849875874740a56ce4170b267e3ec6a70c91670864e6ad4e3a4202509R92).

mthrok (approved):
Approving to unblock you, but it would be nice if you can address the latest comments.

vincentqb (approved):
LGTM! Please do address the feedback given below before merging.

torchaudio/functional/functional.py
```diff
@@ -1444,3 +1446,46 @@ def resample(
...
+    The function computes an edit distance allowing deletion, insertion and
+    substitution. The result is an integer.
+
+    For most applications, the two input sequences should be the same type.
+    If a pair strings are given, the output would be the edit distance between
+    two words (character edit distance); if two lists of strings are given, the
+    output would be the edit distance between sentences (word edit distance).
+    Users may want to normalize the output by the length of the reference
+    sequence.
```
vincentqb, line 1464:
```suggestion
    Compute the (Levenshtein) edit distance between two sequences,
    allowing deletion, insertion and substitution. The result is an integer.

    For most applications, the two input sequences should be the same type.
    If two strings are given, the output is the edit distance between the two
    (character edit distance). If two lists of strings are given, the
    output is the edit distance between sentences (word edit distance).
    Users may want to normalize the output by the length of the reference
    sequence.
```

vincentqb: nit: misspelled sequence

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -382,6 +382,44 @@ def test_phase_vocoder_shape(self, rate, test_pseudo_complex):
...
+
+            ["aaa", "aba", 1],  # substitution
+            ["aba", "aaa", 1],
+            ["aba", "   ", 3],
+
+            ["abc", "bcd", 2],  # mix deletion and substitution
+            ["0ᑌᑎI", "ᑌᑎIᑕO", 3],
+
+            # sentences
+            [["hello", "", "Tᕮ᙭T"], ["hello", "", "Tᕮ᙭T"], 0],  # equal
```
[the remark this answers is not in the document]
yangarbiter: I've added them [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR410) and [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR406).

torchaudio/functional/functional.py
yangarbiter: Updated it [here](https://github.com/pytorch/audio/pull/1601/files#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR406)

vincentqb: There is more than just the typo to change. I've updated some of the phrasing:
```
    Calculate the (Levenshtein) edit distance between two sequences.

    The function computes an edit distance allowing deletion, insertion and
    substitution. The result is an integer.

    For most applications, the two input sequences should be the same type.
    If two strings are given, the output is the edit distance between
    the two strings (character edit distance). If two lists of strings are given, the
    output is the edit distance between sentences (word edit distance).
    Users may want to normalize the output by the length of the reference
    sequence.
```

vincentqb:
The errors that I see below don't currently occur on master as far as I can tell. In any case, they are unrelated to this PR, and so we can ignore those.
```
FAILED: torchaudio/csrc/CMakeFiles/_torchaudio.dir/rnnt/autograd.cpp.o 
/usr/bin/c++ -DINCLUDE_KALDI -DINCLUDE_SOX -DTORCH_API_INCLUDE_EXTENSION_H -DUSE_CUDA -D_torchaudio_EXPORTS -I../../ -I../../env/include/python3.7m -I../../third_party/sox/../install/include -I../../third_party/kaldi/src -I../../third_party/kaldi/submodule/src -isystem ../../env/lib/python3.7/site-packages/torch/include -isystem ../../env/lib/python3.7/site-packages/torch/include/torch/csrc/api/include -isystem /usr/local/cuda/include -Wall -D_GLIBCXX_USE_CXX11_ABI=0 -O3 -DNDEBUG -fPIC -D_GLIBCXX_USE_CXX11_ABI=0 -std=gnu++14 -MD -MT torchaudio/csrc/CMakeFiles/_torchaudio.dir/rnnt/autograd.cpp.o -MF torchaudio/csrc/CMakeFiles/_torchaudio.dir/rnnt/autograd.cpp.o.d -o torchaudio/csrc/CMakeFiles/_torchaudio.dir/rnnt/autograd.cpp.o -c ../../torchaudio/csrc/rnnt/autograd.cpp
../../torchaudio/csrc/rnnt/autograd.cpp: In function ‘std::tuple<at::Tensor, c10::optional<at::Tensor> > torchaudio::rnnt::rnnt_loss_autograd(at::Tensor&, const at::Tensor&, const at::Tensor&, const at::Tensor&, int64_t, double, bool, bool)’:
../../torchaudio/csrc/rnnt/autograd.cpp:56:7: error: ‘AutoDispatchBelowADInplaceOrView’ is not a member of ‘at’
   at::AutoDispatchBelowADInplaceOrView guard;
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../torchaudio/csrc/rnnt/autograd.cpp:56:7: note: suggested alternative: ‘AutoDispatchBelowInplaceOrView’
   at::AutoDispatchBelowADInplaceOrView guard;
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
       AutoDispatchBelowInplaceOrView
```
Let's keep an eye on master in case they occur there too.

The most [levenshtein tests](https://app.circleci.com/pipelines/github/pytorch/audio/6458/workflows/a2429f4d-3d53-4147-8866-67e4e316aa17/jobs/249468) are passing
```
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_00_ PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_01_abc PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_02__I_O PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_03_abc PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_04_aa PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_05_aaa PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_06__I PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_07_aaa PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_08_aba PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_09_aba PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_10_abc PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_11_0_I PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_12 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_13 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_14 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_15 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_16 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_17 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_18 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_19 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_20_aba PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_21 PASSED [ 45%]
torchaudio_unittest/functional/functional_cpu_test.py::TestFunctionalFloat32::test_simple_case_edit_distance_22 PASSED [ 45%]
```

PR #1602, opened 2021-06-22
Rename RNNT loss C++ parameters

carolineechen:
rename variables in RNNT loss C++ code to be consistent with Python interface


vincentqb (approved):
LGTM, but let's wait on the test to pass :)

vincentqb:
@carolineechen -- given that the gpu tests are passing, if the tests pass locally on cpu, you can go ahead and merge this.

PR #1603, opened 2021-06-22
Rename transducer to RNNT

carolineechen:
for consistency, use `RNNT` instead of `transducer` to refer to RNN transducer


vincentqb (approved):
LGTM, but let's wait for the tests to pass :)

PR #265, opened 2019-09-03
Revert "Change Tensor.data<T> to Tensor.data_ptr<T> (#261)"

vincentqb:
Compilation error noticed in #262.

First PR to crash is #261 according to [this](https://travis-ci.org/pytorch/audio/pull_requests). Unknown status from [this](https://travis-ci.org/pytorch/audio/builds).

This reverts commit 860edcdb13e07d017ae0d081d4e2e3034b5e455f.

PR #1604, opened 2021-06-22
[BC-Breaking] Default to PCM_16 for flac on soundfile backend

### LINKED ISSUE #1592: soundfile save test failing
## 🐛 Bug

I've tried to build torchaudio from the source but one test is consistently failing (backend/soundfile/save_test.py::TestFileObject::test_fileobj_flac). This happens on both my Linux and OSX machine.

## To Reproduce

1. `git clone https://github.com/pytorch/audio`
2. `cd audio`
3. `git submodule update --init --recursive`
4. `conda create -n temp2 -c pytorch-nightly pytorch`
5. `conda activate temp2`
6. `pip install typing pytest scipy numpy parameterized pyyaml`
7. `NO_CUDA=1 BUILD_SOX=1 MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py install` (for linux machine, it is build with `BUILD_SOX=1 python setup.py install`)
8. `export PATH="${PATH_TO_TORCHAUDIO}/audio/third_party/install/bin:${PATH}"`
9. `cd test/torchaudio_unittest`
10. `Pip install soundfile`
11. `pytest backend/soundfile/save_test.py` (more specifically `pytest backend/soundfile/save_test.py::TestFileObject::test_fileobj_flac`)

Here is the pytest message by running (`pytest backend/soundfile/save_test.py`).

```bash
============================================================================================================================ test session starts =============================================================================================================================
platform linux -- Python 3.9.5, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: /data/home/yyyang/temp/audio
plugins: xdist-2.3.0, forked-1.3.0
collected 958 items

backend/soundfile/save_test.py ....................................................................................................................................................................................................................................... [ 24%]
...................................................................................................................................................................................................................................................................... [ 51%]
...................................................................................................................................................................................................................................................................... [ 78%]
.......................................................................................................................................................................................................F...                                                            [100%]
================================================================================================================================== FAILURES ==================================================================================================================================
______________________________________________________________________________________________________________________ TestFileObject.test_fileobj_flac ______________________________________________________________________________________________________________________

self = <torchaudio_unittest.backend.soundfile.save_test.TestFileObject testMethod=test_fileobj_flac>

    @skipIfFormatNotSupported("FLAC")
    @skipIfRocm
    def test_fileobj_flac(self):
        """Saving audio via file-like object works"""
>       self._test_fileobj('flac')

backend/soundfile/save_test.py:287:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
backend/soundfile/save_test.py:277: in _test_fileobj
    self.assertEqual(expected, found, atol=1e-4, rtol=1e-8)
/fsx/users/yyyang/conda/envs/torchaudio-upstream/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1407: in assertEqual
    self.assertEqual(x_, y_, atol=atol, rtol=rtol, msg=msg,
/fsx/users/yyyang/conda/envs/torchaudio-upstream/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1407: in assertEqual
    self.assertEqual(x_, y_, atol=atol, rtol=rtol, msg=msg,
/fsx/users/yyyang/conda/envs/torchaudio-upstream/lib/python3.9/site-packages/torch/testing/_internal/common_utils.py:1419: in assertEqual
E   AssertionError: False is not true : Scalars failed to compare as equal! Comparing -0.99993896484375 and 1.00006103515625 gives a difference of 2.0, but the allowed difference with rtol=1e-08 and atol=0.0001 is only 0.00010001000061035157!
============================================================================================================================== warnings summary ==============================================================================================================================
test/torchaudio_unittest/backend/soundfile/save_test.py: 10 warnings
  /data/home/yyyang/temp/audio/test/torchaudio_unittest/common_utils/wav_utils.py:77: WavFileWarning: Chunk (non-data) not understood, skipping it.
    sample_rate, data = scipy.io.wavfile.read(path)

========================================================================================================================== short test summary info ===========================================================================================================================
FAILED backend/soundfile/save_test.py::TestFileObject::test_fileobj_flac - AssertionError: False is not true : Scalars failed to compare as equal! Comparing -0.99993896484375 and 1.00006103515625 gives a difference of 2.0, but the allowed difference with rtol=1e-08 a...
================================================================================================================= 1 failed, 957 passed, 10 warnings in 6.80s =================================================================================================================
```

## Expected behavior

The test should pass.

## Environment

 see above
 533c2d47f7c03370a26b9744f9ec2093857fc43a
```
>>> soundfile.__version__
'0.10.3'
>>> torchaudio.__version__
'0.10.0a0+533c2d4'
>>> scipy.__version__
'1.6.3'
```

OSX machine:
```
PyTorch version: 1.10.0.dev20210618
Is debug build: False

OS: macOS 11.4 (x86_64)
Clang version: 12.0.5 (clang-1205.0.22.9)
CMake version: version 3.20.4
Libc version: N/A

Python version: 3.9 (64-bit runtime)
Python platform: macOS-10.16-x86_64-i386-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

[pip3] numpy==1.20.3
[pip3] torch==1.10.0.dev20210618
[pip3] torchaudio==0.10.0a0+533c2d4
[conda] blas                      1.0                         mkl
[conda] mkl                       2021.2.0           hecd8cb5_269
[conda] numpy                     1.20.3                   pypi_0    pypi
[conda] pytorch                   1.10.0.dev20210618         py3.9_0    pytorch-nightly
[conda] torchaudio                0.10.0a0+533c2d4          pypi_0    pypi
```

Linux machine:
```
CUDA used to build PyTorch: 10.2

OS: Ubuntu 18.04.5 LTS (x86_64)
GCC version: (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Clang version: Could not collect
CMake version: version 3.19.6
Libc version: glibc-2.27

Python version: 3.9 (64-bit runtime)
Python platform: Linux-5.4.0-1041-aws-x86_64-with-glibc2.27
Is CUDA available: False
CUDA runtime version: 11.1.105
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Probably one of the following:
/usr/local/cuda-10.1/targets/x86_64-linux/lib/libcudnn.so.7.6.5
/usr/local/cuda-10.2/targets/x86_64-linux/lib/libcudnn.so.7.6.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_adv_infer.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_adv_train.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_cnn_infer.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_cnn_train.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_ops_infer.so.8.0.5
/usr/local/cuda-11.0/targets/x86_64-linux/lib/libcudnn_ops_train.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_adv_infer.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_adv_train.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_cnn_infer.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_cnn_train.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_ops_infer.so.8.0.5
/usr/local/cuda-11.1/targets/x86_64-linux/lib/libcudnn_ops_train.so.8.0.5
HIP runtime version: N/A
MIOpen runtime version: N/A

[pip3] numpy==1.20.3
[pip3] torch==1.10.0.dev20210618
[pip3] torchaudio==0.10.0a0+533c2d4
[conda] blas                      1.0                         mkl
[conda] cudatoolkit               10.2.89              hfd86e86_1
[conda] mkl                       2021.2.0           h06a4308_296
[conda] numpy                     1.20.3                   pypi_0    pypi
[conda] pytorch                   1.10.0.dev20210618 py3.9_cuda10.2_cudnn7.6.5_0    pytorch-nightly
[conda] torchaudio                0.10.0a0+533c2d4           dev_0    <develop>
```

yangarbiter:
The test for backend/soundfile/save_test.py::TestFileObject::test_fileobj_flac is not working correctly. Please refer to #1592 for more details.

closes #1592

### DIFF
--- a/torchaudio/backend/soundfile_backend.py
+++ b/torchaudio/backend/soundfile_backend.py
@@ -283,7 +283,7 @@ def _get_subtype(
         if encoding:
             raise ValueError("flac does not support encoding.")
         if not bits_per_sample:
-            return "PCM_24"
+            return "PCM_16"
         if bits_per_sample > 24:
             raise ValueError("flac does not support bits_per_sample > 24.")
         return "PCM_S8" if bits_per_sample == 8 else f"PCM_{bits_per_sample}"
@@ -382,8 +382,8 @@ def save(
 
     ``"flac"``
         - 8-bit
-        - 16-bit
-        - 24-bit (default)
+        - 16-bit (default)
+        - 24-bit
 
     ``"ogg"``, ``"vorbis"``
         - Doesn't accept changing configuration.
--- a/torchaudio/backend/soundfile_backend.py
+++ b/torchaudio/backend/soundfile_backend.py
@@ -416,6 +416,9 @@ def save(
 
     if bits_per_sample not in (None, 8, 16, 24, 32, 64):
         raise ValueError("Invalid bits_per_sample.")
+    if bits_per_sample == 24:
+        warnings.warn("Saving audio with 24 bits per sample might warp samples near -1. "
+                      "Using 16 bits per sample might be able to avoid this.")
     subtype = _get_subtype(src.dtype, ext, encoding, bits_per_sample)
 
     # sph is a extension used in TED-LIUM but soundfile does not recognize it as NIST format,


mthrok (commented):
Looks good. 

Can you also remove the following line? ROCm should have nothing to do with soundfile and I was puzzled why it was happening, but it seems the test was broken.

https://github.com/pytorch/audio/blob/9e36281637ac99b7d3ab2b8d555ad98b7ceee9b2/test/torchaudio_unittest/backend/soundfile/save_test.py#L284

torchaudio/backend/soundfile_backend.py
```diff
@@ -283,7 +283,7 @@ def _get_subtype(
         if encoding:
             raise ValueError("flac does not support encoding.")
         if not bits_per_sample:
-            return "PCM_24"
+            return "PCM_16"
         if bits_per_sample > 24:
```
mthrok, line 287: I wonder if we should also prohibit 24 as it could destroy the signal.
@carolineechen What do you think?

carolineechen: I'm not sure I completely understand -- why could using 24-bit destroy the signal?

mthrok: Some process inside of `soundfile` (or `libsndfile` underneath) causes warping from `-1` to `1`. 
https://github.com/pytorch/audio/issues/1592#issuecomment-865141782
I am not sure which and where.

The format is supposed to support up to 32 bits.
https://xiph.org/flac/format.html
> FLAC supports from 4 to 32 bits per sample. 

Using `PCM_16` avoid this problem and `soundfile` defaults to it. IIRC, `soundfile` does not support 32 bit so we raise an error, but for `PCM_24` it does not fail, but gives incorrect result. At least I think we should warn.

carolineechen: I see, it looks like it's being caused by something in `libsndfile` where converting from wav to flac creates warping/artifacts ([#274](https://github.com/bastibe/python-soundfile/issues/274), [#265](https://github.com/bastibe/python-soundfile/issues/265)). I think adding a warning makes sense, but I'm not sure about prohibiting 24-bit since it seems this may also be an issue with other bit sizes as well

mthrok: @carolineechen  Thanks for the issue links, then let's add warning and move on here. How does the following message sound?

@yangarbiter Can you add warning something like;

```python
if bits_per_sample == 24:
    warnings.warn("Saving audio with 24 bits per sample might warp samples near -1. Using 16 bits per sample might be able to avoid this.")
```

yangarbiter: I think the description here is not very precise. Because this distortion not only happens in PCM_24, it also happens in PCM_16. See the below example, the assertion in line 271 will not pass.
Screen Shot 2021-06-23 at 1 02 14 PM

The reason that we can pass the current test is that the assertion here
https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/backend/soundfile/save_test.py#L277

The variables `expected` and `found` are distorted in the same way.

Because of this, the warning I would suggest is
```
if bits_per_sample == None and ext == "flac":
    warnings.warn("We have changed the default bits_per_sample for flac files from 24 to 16. Please set bits_per_sample to 24 to maintain the same behavior.")
```

mthrok: > Because this distortion not only happens in PCM_24, it also happens in PCM_16

Can you show how it does not pass the assertion? If the reason is absolute difference is slightly larger than the given `atol=1e-4`, then it's fine, because you are saving data in 16 bits. But if the reason is another case of -1 -> 1 warping then it's the real concern.

>     warnings.warn("The default bits_per_sample for flac files has been changed from 24 to 16. Please set bits_per_sample to 24 to maintain the same behavior.")

The core of the concern here is the fact that 24bit could cause warping, and that's what we want users to be aware of it. It's so subtle and difficult to notice.

mthrok: Actually, now that I realized that you are not talking about torch audio's soundfile backend but you are talking about soundfile itself. 

The assertion failure you are mentioning is expected to happen, and that's not how save function is tested. And your modification is invalid for our context. Please refer to http://www.mega-nerd.com/libsndfile/FAQ.html#Q010 for why your assertion is failing.

yangarbiter: I think this post is talking about small error when you write and read the same file due to float conversion (correct me if I am wrong). Sorry that I should have provide more detail about this. I added a few prints to the code. These error does not seem to be errors due to float conversion.

Screen Shot 2021-06-24 at 8 29 42 AM

Here are the print results.
Screen Shot 2021-06-24 at 8 27 46 AM

You can see that the fourth and fifth row for the `expected` variable goes completely off. Please let me know if you are also able to reproduce this error.

yangarbiter:
Now it appears that not only PCM_24 will break things, PCM_16 will also distort the waveform. Here are two questions I would like comments on:
1. Do we want to add a general warning that says soundfile has a known issue of distorting waveform? We can link to soundfile's issue ([#274](https://github.com/bastibe/python-soundfile/issues/274), [#265](https://github.com/bastibe/python-soundfile/issues/265)) in the warning.
2. To fix the test, do we want to change the default value of `bits_per_sample` to match soundfile’s default (BC breaking) or just to change the value in the test (not BC breaking).

yangarbiter:
Here is the code for generating the comparison between 16 bits and 24 bits.
Screen Shot 2021-07-06 at 1 28 19 PM

Here are test_16.flac and test_24.flac that is generated shown in audacity
Screen Shot 2021-07-06 at 1 26 52 PM

Here are backend_16.flac and backend_24.flac that is generated shown in audacity
Screen Shot 2021-07-06 at 1 29 42 PM

So these shows that setting default to 16 bits seems to be a better choice.

@mthrok, I've also added the warning in the latest commit. Please let me know what you think. :)

mthrok (approved):

mthrok:
Please update the PR title to match the change.

PR #1583, opened 2021-06-17
Replace deprecated AutoNonVariableTypeMode

dgenzel2:
Following https://github.com/pytorch/pytorch/blob/master/docs/cpp/source/notes/inference_mode.rst#migration-guide-from-autononvariabletypemode I figured that the proper replacement is AutoDispatchBelowADInplaceOrView, but I am not 100% sure, so please confirm.

Resolves #1522

PR #1605, opened 2021-06-23
Update unit test CUDA to 10.2

mthrok:
The CUDA 10.1 was dropped from binary distribution in April.
Due to this #1583 fails to compile in GPU tests.
Update this so that the latest nightly is pulled correctly.
However latest nightly is exhibiting the issue with CMake-based build and GPU tests fail for another reason.
I will apply the hot fix separately, so that the hot fix is easy to revert later.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -447,7 +447,7 @@ jobs:
     resource_class: gpu.small
     environment:
       <<: *environment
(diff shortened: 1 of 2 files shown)


mthrok:
Merging this as this is blocking all the PRs.

PR #1606, opened 2021-06-23
[HOT FIX] Temporarily pin nightly version on Linux GPU unit test

mthrok:
When hot fix was applied in #1598, it was only applied to CPU jobs.
This is because I looked at the nightly build, saw green lights on GPU unit test jobs, 
and thought that GPU jobs were somehow recovered.

It turned out, however, that GPU unit tests were fetching CUDA 10.1 binaries from April.
Because the PyTorch core dropped binary distribution for CUDA 10.1.

The unit test job was updated in #1605 to use CUDA 10.2. The GPU tests are now
picking up the latest nightly and failing because of broken CMake integration as expected.

This PR applies the same hot fix as in #1598 to GPU unit tests, so that they pass.


NicolasHug (approved):
stamping

PR #1608, opened 2021-06-23
[HOT FIX] Pin nightly version in Windows unit test CI

mthrok:
Similar to #1598 and #1606, pin the nightly version on Windows CI while the upstream issue persists.


NicolasHug (approved):
stamping

PR #1610, opened 2021-06-24
Remove reuse_logits_for_grads option for RNNTL

carolineechen:
following discussion in [#1532](https://github.com/pytorch/audio/pull/1532#discussion_r647660456) and #1536, we are removing the `reuse_logits_for_grad` option for RNNT loss, as this results in modifying inputs in place and is not the usual behavior of a loss function

### DIFF
--- a/test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
+++ b/test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
@@ -17,14 +17,12 @@ class RNNTLossTest:
         self, data, ref_costs, ref_gradients, atol=1e-6, rtol=1e-2
     ):
         logits_shape = data["logits"].shape
-        for reuse_logits_for_grads in [False, True]:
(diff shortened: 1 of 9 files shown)


test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
```diff
@@ -18,26 +18,23 @@ def _test_costs_and_gradients(
...
-                        for t in range(gradients.shape[1]):
-                            for u in range(gradients.shape[2]):
-                                np.testing.assert_allclose(
-                                    gradients[b, t, u],
-                                    ref_gradients[b, t, u],
-                                    atol=atol,
-                                    rtol=rtol,
-                                    err_msg=f"failed on b={b}, t={t}/T={T}, u={u}/U={U}",
-                                )
+        with self.subTest():
```
vincentqb, line 21: nit: subTest is no longer needed since there is only one test to run

vincentqb (approved):
LGTM

Issue #1574, opened 2021-06-14
🚀 Feature Proposal: Linear Frequency Cepstral Coefficients (LFCC)


Joool:
## 🚀 Feature
Add linear frequency cepstral coefficients (think MFCC but with a liner filterbank instead of mel) as a transform.

## Motivation
I implemented linear frequency cepstral coefficients (LFCC) for a recent project based on the torchaudio MFCC implementation. If this is of interest to the project, I would happily commit them.

## Pitch
LFCCs are, for example, used in automatic speaker verification. This would also ease experimentation in other contexts.

## Additional context
This is related to #942 and #943. These PRs propose changes that might help people build different cepstral coefficients by chaining together different other transforms. My solution would be an end-to-end solution, i.e., add a new transform similar to MFCCs. I see value in both approaches and would like some input on this.

mthrok:
Hi @Joool 

Thanks for the proposal. I think idea-wise it is a nice extension.
Couple of questions
- Is there a paper that we can learn the effect / usage of this?
- `torchaudio.transforms.MFCC` is composed of `Spectrogram`, `MelScale` and `DCT transformation`. Is LFCC simply replacing `MelScale`? or did you have to do something more to make it work with the current interface?

Joool:
Hi,

> * Is there a paper that we can learn the effect / usage of this?

Sorry forgot to link a paper: 
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.717.6743&rep=rep1&type=pdf

See Figure 1 for a comparison.

> * `torchaudio.transforms.MFCC` is composed of `Spectrogram`, `MelScale` and `DCT transformation`. Is LFCC simply replacing `MelScale`? or did you have to do something more to make it work with the current interface?

Yes, I basically replaced the matrix creation of MelScale to use a linear scale filter bank instead:

https://github.com/pytorch/audio/blob/c5d0390c8d9a0356d350bfced1c2152ae308e9b9/torchaudio/transforms.py#L280-L283

I would add a new function for that and I would also propose to rename `create_fb_matrix` to something which makes clear it creates a mel filterbank.
#943 goes in a similar direction, but instead wants to change the spectrogram API.

Additionally, spafe offers an implementation in numpy which could be used for differential testing:

https://github.com/SuperKogito/spafe/blob/master/spafe/fbanks/linear_fbanks.py

However, that would add a new dependency.

mthrok:
Hi @Joool

[...]

Issue #943, opened 2020-10-09
Custom spectrogram filterbanks


dvisockas:
## 🚀 Feature
It would be great to have an ability to create non-mel and not linear filterbanking. 

## Motivation
Even though mel-scale ceptstral coefficients are (probably) the most popular, there exists many other ways to filterbank the frequencies of spectra - CQCC, Opus filterbanks, BFCC, etc. I was thinking that it would be nice if the `Spectrogram` transform would accept a filterbank argument and filter the spectrogram accordingly.

## Pitch
Extension of `Spectrogram` API to accept `filter_bank` argument (with a fallback to current implementation). Maybe it could be a tensor of shape (num_banks, filter) where filter would consist of (min_freq, max_freq) with a triangular filterbank for the start?

## Additional context

Somewhat relates to #942

vincentqb:
Thanks for the suggestion :) This relates to [comment](https://github.com/pytorch/audio/pull/593#discussion_r417068529). Thoughts? Would #593 meet your needs?

dvisockas:
#593 is related to mel filter_banks as `create_fb_matrix` calculates mel filterbanks. I was thinking that it would be nice to somehow have an API that would enable users to create non_mel spectrograms. It is a simple operation of `spectrogram * filter_bank`, but implementing non-mel filterbanks is rather tedious because one has to recreate `F.create_fb_matrix` themselves

vincentqb:
Yes, that would be useful, thanks for suggesting :) How about something similar to [AmplitudeToDB](https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py#L173) then?

Issue #1058, opened 2020-11-26
MelSpectrogram inconsistency with librosa melspectrogram


SolomidHero:
Hello! I am excited with this framework a lot and its ability to make transformations on gpu.

**Problem:**
`transforms.Spectrogram` (with power 1.) (which is real) output equals to absolute value of `librosa.stft` (which is complex) with equal parameters.

Here is spectrograms for my example audio (really close results):
<img width="686" alt="Screenshot 2020-11-26 at 22 23 12" src="https://user-images.githubusercontent.com/28161779/100386302-0aa6c180-3036-11eb-9eef-36afd6b03ed6.png">
<img width="678" alt="Screenshot 2020-11-26 at 22 24 45" src="https://user-images.githubusercontent.com/28161779/100386388-42156e00-3036-11eb-9b15-fbbbe8f79e30.png">

Next step is to get melspectrogram using `transforms.MelScale` (on `Spectrogram` with power 1) and `librosa.feature.melspectrogram` (actually power is 1., this argument not in use) (using previous spectrogram). And here we can't get the same result:
- in both steps only matmul takes place
- in `transforms.MelScale` tensors with real values multiplicated, in `librosa.feature.melspectrogram` gives us multiplication of complex based matrices, thus in the result we can get absolutely different values
- also quite misleading use of `power` in `transforms.Spectrogram` (don't need in `librosa.stft`)

And the result (differs not only in some fields, but in scale too):
<img width="679" alt="Screenshot 2020-11-26 at 22 45 46" src="https://user-images.githubusercontent.com/28161779/100387450-2fe8ff00-3039-11eb-8c78-98d2f39f6d76.png">
<img width="678" alt="Screenshot 2020-11-26 at 22 47 18" src="https://user-images.githubusercontent.com/28161779/100387552-6888d880-3039-11eb-914f-f20b06bb289c.png">

eldrin:
Hi! I also happened to have a similar issue with MFCC. After some digging, I found librosa use `slaney` normalization for the mel-filterbank creation as the default, while torchaudio is no normalization by default. (no option to feed normalization setup at the `MelSpec`/`MelSpectrogram` construction.

https://github.com/pytorch/audio/blob/fb3ef9ba427acd7db3084f988ab55169fab14854/torchaudio/transforms.py#L243-L245

After specifying the normalization to `slaney` I got a reasonably comparable result.

[...]

PR #1611, opened 2021-06-25
Lfcc

Joool:
This PR addresses  #1574 and adds Linear Frequency Cepstral Coefficients (LFCCs).

### Changes

- refactored `torchaudio.functional. create_fb_matrix`; pulled out the creation of the filterbank into a new method called `_create_triangular_filterbank `.
- added `torchaudio.functional.linear_fbanks`, which creates a linear triangular filterbank, using the method above.
- added `torchaudio.transforms.LFCC`.

### Possible Alternatives

ATM LFCC is essentially a merge of MFCC and MelScale into one class but only using a linear instead of a mel filterbank.
This could potentially be refactored into having a generic cepstral coefficient class, where the scaling/filterbank can be part of the API.
#943 goes into a similar direction. 

### Open Problem

I have used the filterbank creation in `create_fb_matrix` for the implementation. This method does not create a perfect filterbank (notice that not all filter have a magnitude of 1):

lin_bank

Which is consistent with the behaviour of the mel filterbank torchaudio creates: 

mel_bank

I did some digging and apparently this is related to librosa maintaining API compatibility with Slaney's auditory toolbox. See https://github.com/librosa/librosa/issues/1030 and https://github.com/librosa/librosa/issues/1173. 

I'm unsure if this is a desired behaviour, so I would like some advice here.
The script for creating the above plots can be found [here](https://gist.github.com/Joool/b3b1b1acb09d035366de18335867c7d3).


test/torchaudio_unittest/transforms/transforms_test.py
```diff
@@ -163,6 +163,45 @@ def test_mfcc(self):
 
         self.assertTrue(torch_mfcc_norm_none.allclose(norm_check))
 
+    def test_lfcc(self):
+        audio_orig = self.waveform.clone()
+        audio_scaled = self.scale(audio_orig)  # (1, 16000)
```
mthrok, line 168: Instead of `self.waveform` and `self.scale`, please use `common_utils.get_whitenoise` or something similar.
We are trying to remove these as they create unnecessary couplings among tests and cause confusion to developers.

test/torchaudio_unittest/transforms/transforms_test.py
```diff
@@ -163,6 +163,45 @@ def test_mfcc(self):
...
+                                                              n_filter=n_filter,
+                                                              n_lfcc=n_lfcc,
+                                                              norm=None)
+        torch_lfcc_norm_none = lfcc_transform_norm_none(audio_scaled)  # (1, 40, 321)
+
+        norm_check = torch_lfcc.clone()
+        norm_check[:, 0, :] *= math.sqrt(n_filter) * 2
+        norm_check[:, 1:, :] *= math.sqrt(n_filter / 2) * 2
+
+        self.assertTrue(torch_lfcc_norm_none.allclose(norm_check))
```
mthrok, line 203: Could you split these tests into three different methods? That way even if the first one fails, the other two are separately executed.
Also please turn these comments `# check ...` into docstring in each method, so that they are shown as test description when listing tests i.e. `pytest test --collect-only`.

mthrok: Instead of `assertTrue(Tensor.allclose(...))`, can you use `self.assertEqual`?
That's the standard we use for comparing Tensors in our test code base.

mthrok:
Hi @Joool 

Thanks for the PR. I looked at the implementation and the code itself looks good.

Regarding the issue with imperfect filter bank. I skimmed through the discussion and at the first glance, I am not sure what is the best approach. I will take a detailed look, but at the same time, the following description in MFCC makes the situation more complicated.

> This is not the textbook implementation, but is implemented here to
> give consistency with librosa.

Having the compatibility rule makes it easy to dismiss the issue you raise, but the justification of the rule itself is something hard to convince, as librosa itself says the compatibility with the old audio toolbox is less important these days. Yet, there is also demand for compatibility with `librosa` https://github.com/pytorch/audio/issues/1058#issuecomment-779634174). Personally, if I would prefer textbook implementation, but changing the implementation is very sensitive matter. In fact, we were once told that someone switched from librosa's mel-spactrogram to torchaudio's and the training diverged.

Having said that, do you have an opinion here?

test/torchaudio_unittest/transforms/transforms_test.py
Joool: Addressed in 40ddf05.

test/torchaudio_unittest/transforms/transforms_test.py
Joool: Addressed in 40ddf05.

Joool: Since the test are a one-to-one port of the mfcc tests, I will open a pull request and refactor that one accordingly.

Joool:
Hi @mthrok 

> This is not the textbook implementation, but is implemented here to
> give consistency with librosa.

I though the comment referred to the fact that we use DB-scaling instead of log-scaling (textbook for cepstral coefficients)?

https://github.com/pytorch/audio/blob/284bd1014ad29fac9b3a6336a086681af6380923/torchaudio/transforms.py#L586-L590

Also discussed [here](https://stackoverflow.com/a/60944846). I'm unsure why librosa choses to use the db-scaling (and a quick Google search did not provide me with an answer). Since it only scales the spectrograms, I would expect that training would eventually achieve similar results on both, but that is just my intuition.

> Regarding the issue with imperfect filter bank. I skimmed through the discussion and at the first glance, I am not sure what is the best approach. 

I had a similar response. Since librosa is a fairly popular library, I think keeping the defaults similar is a fair choice. This also makes differential testing between the two implementations easier.

nateanl:
Thanks @Joool for the plots. I agreed that keeping consistency with librosa is a fair decision. 
In terms of the normalization of the filter bank, we can also add the ``np.inf`` option as added in [librosa](https://github.com/librosa/librosa/pull/1175/files#diff-260ae01b06042016564c9ee1694681c899dffdcb1452b5b2c2ee8217066b3cf0R222) (in a separate PR).

torchaudio/transforms.py
```diff
@@ -595,6 +596,108 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+
+        if self.log_lf:
+            log_offset = 1e-6
+            specgram = torch.log(specgram + log_offset)
+        else:
+            specgram = self.amplitude_to_DB(specgram)
+
+        # (..., channel, n_filter, time).transpose(...) dot (n_filter, n_lfcc)
+        # -> (..., channel, time, n_lfcc).transpose(...)
+        lfcc = torch.matmul(specgram.transpose(-2, -1), self.dct_mat).transpose(-2, -1)
```
nateanl, line 697: The formula can be refactored by torch.einsum, for example:
```
lfcc = torch.einsum('...cnt,nl->...ctl', specgram,self.dct_mat)
```

torchaudio/transforms.py
```diff
@@ -595,6 +596,108 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        """
+        specgram = self.Spectrogram(waveform)
+
+        # adopted from mel scale
+        # pack batch
+        shape = specgram.size()
+        specgram = specgram.reshape(-1, shape[-2], shape[-1])
+
+        specgram = torch.matmul(specgram.transpose(1, 2), self.filter_mat)
+        specgram = specgram.transpose(1, 2)
```
nateanl, line 684: These could also be refactored by ``torch.einsum``
```
specgram = torch.einsum('...ft,fg->...gt', specgram, self.filter_mat)
```

torchaudio/transforms.py
Joool: Addressed in 75da309. Since my implementation is simply adapted from MFCC, I also refactored that in 04e2305. If you want we can split that off into a seperate PR.

torchaudio/transforms.py
Joool: Addressed in 75da309. Since my implementation is simply adapted from MelScale, I also refactored that in 04e2305. If you want we can split that off into a seperate PR.

nateanl: Thanks! I think it's fine to refactor those in the same PR.
It looks like the file changes contains others' commits. Could you rebase the PR by clicking the edit button above, change the ``master`` branch to a random one, and change it back? In this way we can filter those committed changes in this PR.

Joool: Sorry, apparently I messed up a rebase. I can open a new (cleaned) PR if you like.

test/torchaudio_unittest/transforms/transforms_test.py
```diff
@@ -180,6 +180,65 @@ def test_mfcc_norms(self):
...
+        n_lfcc = 40
+        n_filter = 128
+        lfcc_transform = torchaudio.transforms.LFCC(sample_rate=sample_rate,
+                                                    n_filter=n_filter,
+                                                    n_lfcc=n_lfcc,
+                                                    norm='ortho')
+        torch_lfcc = lfcc_transform(audio)  # (1, 40, 81)
+        self.assertEqual(torch_lfcc.dim(), 3)
+        self.assertEqual(torch_lfcc.shape[1], n_lfcc)
+        self.assertEqual(torch_lfcc.shape[2], 81, torch_lfcc.shape[2])
```
nateanl, line 198: ``torch_lfcc.shape[2]`` is duplicated. Let's change it to:
```
self.assertEqual(torch_lfcc.shape[2], 81)
```

Joool: Resolved in 51aced6.

torchaudio/transforms.py
```diff
@@ -557,12 +550,104 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        # (..., freq, time) dot (freq, n_filter) -> (..., n_filter, time)
+        specgram = torch.einsum("...ft,fg->...gt", specgram, self.filter_mat)
+
+        if self.log_lf:
+            log_offset = 1e-6
+            specgram = torch.log(specgram + log_offset)
+        else:
+            specgram = self.amplitude_to_DB(specgram)
+
+        # (..., channel, n_filter, time) dot (n_filter, n_lfcc) -> (..., channel, n_lfcc, time)
```
nateanl, line 646: nit: ``channel`` may not be necessary when we deal with single-channel audios. Removing it makes the einsum simpler.

nateanl (approved):
LGTM! 
I think we can merge it after we addressed all nits. 
- Remove the ``channel`` in dct computation.
- Add a auto_grad test for lfcc, similar to the test of [mfcc](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/transforms/autograd_test_impl.py#L105)

carolineechen (commented):
Thanks for working on this! Left a few minor comments as well

docs/source/functional.rst
```diff
@@ -26,6 +26,11 @@ create_fb_matrix
 
 .. autofunction:: create_fb_matrix
 
+linear_fbanks
+----------------
```
carolineechen, line 30:
```suggestion
linear_fbanks
-------------
```

torchaudio/transforms.py
```diff
@@ -557,12 +550,104 @@ def forward(self, waveform: Tensor) -> Tensor:
         else:
             mel_specgram = self.amplitude_to_DB(mel_specgram)
 
-        # (..., channel, n_mels, time).transpose(...) dot (n_mels, n_mfcc)
-        # -> (..., channel, time, n_mfcc).transpose(...)
-        mfcc = torch.matmul(mel_specgram.transpose(-2, -1), self.dct_mat).transpose(-2, -1)
+        # (..., channel, n_mels, time) dot (n_mels, n_mfcc) -> (..., channel, n_nfcc, time)
+        mfcc = torch.einsum("...cnt,nl->...clt", mel_specgram, self.dct_mat)
```
mthrok, line 554: @nateanl What is the performance implication of `einsum`?

torchaudio/transforms.py
```diff
@@ -557,12 +550,104 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+        self.n_lfcc = n_lfcc
+        self.dct_type = dct_type
+        self.norm = norm
+        self.top_db = 80.0
+        self.amplitude_to_DB = AmplitudeToDB('power', self.top_db)
+
+        if speckwargs is not None:
+            self.Spectrogram = Spectrogram(**speckwargs)
+        else:
+            self.Spectrogram = Spectrogram()
```
mthrok, line 609: nit: I think that the following would express the intent better.

```python
speckwargs = speckwargs or {}
self.Spectrogram = Spectrogram(**speckwargs)
```

torchaudio/transforms.py
nateanl: The current version of ``einsum`` should be faster than matmul. Check: https://github.com/pytorch/pytorch/issues/32591#issuecomment-856936116

torchaudio/transforms.py
Joool: b2376ef

torchaudio/functional/functional.py
```diff
@@ -443,6 +463,42 @@ def create_fb_matrix(
...
+        n_filter: int,
+        sample_rate: int,
+) -> Tensor:
+    r"""Creates a linear triangular filterbank.
+
+    Args:
+        n_freqs (int): Number of frequencies to highlight/apply
+        f_min (float): Minimum frequency (Hz)
+        f_max (float): Maximum frequency (Hz)
+        n_mels (int): Number of (linear) triangular filter
```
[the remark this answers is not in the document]
Joool: fc3c1ad

torchaudio/functional/functional.py
```diff
@@ -376,6 +377,31 @@ def _mel_to_hz(mels: Tensor, mel_scale: str = "htk") -> Tensor:
...
 
 
+def _create_triangular_filterbank(
+        all_freqs: Tensor,
+        f_pts: Tensor,
+) -> Tensor:
+    """Create a triangular filter bank.
+
+    Args:
+        all_freqs (Tensor): STFT freq points.
```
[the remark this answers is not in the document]
Joool: fc3c1ad

docs/source/functional.rst
Joool: fc3c1ad. I think the suggested change would have added an additional `linear_fbank`.

torchaudio/transforms.py
Joool: In 0210807 since I missed this comment beforehand -.-

torchaudio/transforms.py
Joool: I thought so as well, but for a simple [test script](https://gist.github.com/Joool/6c3b24cdbb9a915bf7943714f4d46b23) (uses white noise as audio), the old implementation does outperform `einsum` on my machine:

```
PyTorch version 1.10.0.dev20210617+cu111
Torchaudio version 0.10.0a0+bac32ec

CPU:
Old implementation: 0.2768233451060951
Einsum implementation: 0.33643896970897913

GPU:
Old implementation: 0.23426104336977005
Einsum implementation: 0.30553871812298894

```

Joool: The results are similar when I use the `torch.benchmark` ([script](https://gist.github.com/Joool/b7ed70e4c9b3c001473c398254b80687)):

```
PyTorch version 1.10.0.dev20210617+cu111
Torchaudio version 0.10.0a0+bac32ec
CPU:
     Old | 35.8   
     Einsum | 42.0

GPU:
      Old | 24
      Einsum | 31.6

```

nateanl: I have the same result by trying your script and change the ``duration`` to 5.
```
PyTorch version 1.9.0
Torchaudio version 0.10.0a0+0210807
CPU:
Old implementation: 5.004824341
Einsum implementation: 5.26490228
```
Based on that, we should change it back to the old version to get a more effective performance. 🙂

mthrok (commented):
How close is this implementation to librosa?
If it is comparable, then it's nice to have [a test for librosa compatibility](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/functional/librosa_compatibility_test_impl.py).

Joool:
No worries. Librosa does not have an LFCC implementation (afaik).

torchaudio/transforms.py
Joool: Done in f07fb93. I additionally removed the unnecessary batch unpacking which was present in the old implementation.

Old:
https://github.com/pytorch/audio/blob/108a32d92cd77733841d6c61ab91f8e3ce995f35/torchaudio/transforms.py#L285-L294

New:
```
mel_specgram = torch.matmul(specgram.transpose(-1, -2), self.fb).transpose(-1, -2)
```

Which is slightly more efficient (I assume since you do not reshape as much):
```
PyTorch version 1.10.0.dev20210617+cu111
Torchaudio version 0.10.0a0+bac32ec
CPU:
    Old |                   35.8
    Einsum |                41.5
    Old simplified |        30.9 

GPU:
    Old |                   24
    Einsum |                30.3
    Old simplified |        18

```

mthrok (approved):

nateanl:
Thanks for your contribution! @Joool  I have merged LFCC to the master.

PR #776, opened 2020-07-13
Offer pretrained wavernn model

jimchen90:
Offer WaveRNN model with pretrained weights (8 bits waveform mode) on LJSpeech using the method as [torchvision example](https://github.com/pytorch/vision/blob/master/torchvision/models/alexnet.py#L52).

Related to #735  #749 
[Internal](https://drive.google.com/drive/folders/10XmuONMpff1zZIh-qc2pJgnHRKztpMSJ?usp=sharing)

PR #1612, opened 2021-06-25
Add pretrained weights for wavernn

yangarbiter:
Closes #776.

Offers pertained weight for WaveRNN with 8 bits waveform mode and trained on LJSpeech.

Following the convention [here](https://github.com/pytorch/vision/blob/master/torchvision/models/alexnet.py#L53).

### DIFF
--- a/torchaudio/models/_utils.py
+++ b/torchaudio/models/_utils.py
@@ -1,0 +1,8 @@
+try:
+    from torch.hub import load_state_dict_from_url
+except ImportError:
+    from torch.utils.model_zoo import load_url as load_state_dict_from_url
+
+__all__ = [
+    'load_state_dict_from_url',
+]
--- a/torchaudio/models/_utils.py
+++ b/torchaudio/models/_utils.py
@@ -1,8 +1,0 @@
-try:
-    from torch.hub import load_state_dict_from_url
-except ImportError:
-    from torch.utils.model_zoo import load_url as load_state_dict_from_url
-
-__all__ = [
-    'load_state_dict_from_url',
-]
--- a/torchaudio/models/wavernn.py
+++ b/torchaudio/models/wavernn.py
@@ -1,4 +1,4 @@
-from typing import List, Tuple, Any
+from typing import List, Tuple, Dict, Any
 
 import torch
 from torch import Tensor
@@ -13,13 +13,26 @@ __all__ = [
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
-    "wavernn_10k_epochs_8bits_ljspeech",
+    "get_pretrained_wavernn",
 ]
 
 
-model_urls = {
-    'wavernn_10k_epochs_8bits_ljspeech': 'https://download.pytorch.org/models/'
-                                         'audio/wavernn_10k_epochs_8bits_ljspeech.pth',
+model_config_and_urls: Dict[str, Tuple[str, Dict[str, Any]]] = {
+    'wavernn_10k_epochs_8bits_ljspeech': (
+        'https://download.pytorch.org/models/audio/wavernn_10k_epochs_8bits_ljspeech.pth',
+        {
+            'upsample_scales': [5, 5, 11],
+            'n_classes': 2 ** 8,  # n_bits = 8
+            'hop_length': 275,
+            'n_res_block': 10,
+            'n_rnn': 512,
+            'n_fc': 512,
+            'kernel_size': 5,
+            'n_freq': 80,
+            'n_hidden': 128,
+            'n_output': 128
+        }
+    )
 }
 
 
@@ -336,35 +349,25 @@ class WaveRNN(nn.Module):
         return x.unsqueeze(1)
 
 
-def _wavernn(arch: str, pretrained: bool, progress: bool, **kwargs: Any) -> WaveRNN:
-    model = WaveRNN(**kwargs)
-    if pretrained:
-        state_dict = load_state_dict_from_url(model_urls['wavernn'],
-                                              progress=progress)
-        model.load_state_dict(state_dict)
-    return model
+def get_pretrained_wavernn(checkpoint_name: str, progress: bool = True) -> WaveRNN:
+    r"""Get pretrained WaveRNN model.
 
+    Here are the available checkpoints:
 
-def wavernn_10k_epochs_8bits_ljspeech(pretrained: bool = True, progress: bool = True, **kwargs: Any) -> WaveRNN:
-    r"""WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
-    The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
+    - wavernn_10k_epochs_8bits_ljspeech
+
+        WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
+        The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
 
     Args:
-        pretrained (bool): If True, returns a model pre-trained on LJSpeech
-        progress (bool): If True, displays a progress bar of the download to stderr
+        checkpoint_name (str): The name of the checkpoint to load.
+        progress (bool): If True, displays a progress bar of the download to stderr.
     """
-    n_bits = 8
-    configs = {
-        'upsample_scales': [5, 5, 11],
-        'n_classes': 2 ** n_bits,
-        'hop_length': 275,
-        'n_res_block': 10,
-        'n_rnn': 512,
-        'n_fc': 512,
-        'kernel_size': 5,
-        'n_freq': 80,
-        'n_hidden': 128,
-        'n_output': 128
-    }
-    configs.update(kwargs)
-    return _wavernn("wavernn_10k_epochs_8bits_ljspeech", pretrained=pretrained, progress=progress, **configs)
+    if checkpoint_name in model_config_and_urls:
+        url, configs = model_config_and_urls[checkpoint_name]
+        model = WaveRNN(**configs)
+        state_dict = load_state_dict_from_url(url, progress=progress)
+        model.load_state_dict(state_dict)
+        return model
+    else:
+        raise ValueError("The model_name `{}` is not supported.".format(checkpoint_name))
--- a/torchaudio/models/wavernn.py
+++ b/torchaudio/models/wavernn.py
@@ -3,10 +3,7 @@ from typing import List, Tuple, Dict, Any
 import torch
 from torch import Tensor
 from torch import nn
-try:
-    from torch.hub import load_state_dict_from_url
-except ImportError:
-    from torch.utils.model_zoo import load_url as load_state_dict_from_url
+from torch.hub import load_state_dict_from_url
 
 
 __all__ = [
@@ -15,7 +12,7 @@ __all__ = [
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
-    "get_pretrained_wavernn",
+    "wavernn",
 ]
 
 
@@ -351,25 +348,24 @@ class WaveRNN(nn.Module):
         return x.unsqueeze(1)
 
 
-def get_pretrained_wavernn(checkpoint_name: str, progress: bool = True) -> WaveRNN:
+def wavernn(checkpoint_name: str) -> WaveRNN:
     r"""Get pretrained WaveRNN model.
 
-    Here are the available checkpoints:
-
-    - wavernn_10k_epochs_8bits_ljspeech
-
-        WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
-        The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
-
     Args:
-        checkpoint_name (str): The name of the checkpoint to load.
-        progress (bool): If True, displays a progress bar of the download to stderr.
+        checkpoint_name (str): The name of the checkpoint to load. Available checkpoints:
+
+            - ``"wavernn_10k_epochs_8bits_ljspeech"``:
+
+                WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
+                The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
     """
     if checkpoint_name not in _MODEL_CONFIG_AND_URLS:
-        raise ValueError("The checkpoint_name `{}` is not supported.".format(checkpoint_name))
+        raise ValueError(
+            f"Unexpected checkpoint_name: '{checkpoint_name}'. "
+            f"Valid choices are; {list(_MODEL_CONFIG_AND_URLS.keys())}")
 
     url, configs = _MODEL_CONFIG_AND_URLS[checkpoint_name]
     model = WaveRNN(**configs)
-    state_dict = load_state_dict_from_url(url, progress=progress)
+    state_dict = load_state_dict_from_url(url, progress=False)
     model.load_state_dict(state_dict)
     return model
--- a/torchaudio/models/wavernn.py
+++ b/torchaudio/models/wavernn.py
@@ -357,7 +357,9 @@ def wavernn(checkpoint_name: str) -> WaveRNN:
             - ``"wavernn_10k_epochs_8bits_ljspeech"``:
 
                 WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
-                The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
+                The model is trained using the default parameters and code of the
+                `examples/pipeline_wavernn/main.py
+                <https://github.com/pytorch/audio/tree/master/examples/pipeline_wavernn>`_.
     """
     if checkpoint_name not in _MODEL_CONFIG_AND_URLS:
         raise ValueError(
--- a/torchaudio/models/wavernn.py
+++ b/torchaudio/models/wavernn.py
@@ -3,8 +3,10 @@ from typing import List, Tuple, Dict, Any
 import torch
 from torch import Tensor
 from torch import nn
-
-from ._utils import load_state_dict_from_url
+try:
+    from torch.hub import load_state_dict_from_url
+except ImportError:
+    from torch.utils.model_zoo import load_url as load_state_dict_from_url
 
 
 __all__ = [
@@ -17,7 +19,7 @@ __all__ = [
 ]
 
 
-model_config_and_urls: Dict[str, Tuple[str, Dict[str, Any]]] = {
+_MODEL_CONFIG_AND_URLS: Dict[str, Tuple[str, Dict[str, Any]]] = {
     'wavernn_10k_epochs_8bits_ljspeech': (
         'https://download.pytorch.org/models/audio/wavernn_10k_epochs_8bits_ljspeech.pth',
         {
@@ -363,11 +365,11 @@ def get_pretrained_wavernn(checkpoint_name: str, progress: bool = True) -> WaveR
         checkpoint_name (str): The name of the checkpoint to load.
         progress (bool): If True, displays a progress bar of the download to stderr.
     """
-    if checkpoint_name in model_config_and_urls:
-        url, configs = model_config_and_urls[checkpoint_name]
-        model = WaveRNN(**configs)
-        state_dict = load_state_dict_from_url(url, progress=progress)
-        model.load_state_dict(state_dict)
-        return model
-    else:
-        raise ValueError("The model_name `{}` is not supported.".format(checkpoint_name))
+    if checkpoint_name not in _MODEL_CONFIG_AND_URLS:
+        raise ValueError("The checkpoint_name `{}` is not supported.".format(checkpoint_name))
+
+    url, configs = _MODEL_CONFIG_AND_URLS[checkpoint_name]
+    model = WaveRNN(**configs)
+    state_dict = load_state_dict_from_url(url, progress=progress)
+    model.load_state_dict(state_dict)
+    return model
--- a/torchaudio/models/wavernn.py
+++ b/torchaudio/models/wavernn.py
@@ -1,8 +1,11 @@
-from typing import List, Tuple
+from typing import List, Tuple, Any
 
 import torch
 from torch import Tensor
 from torch import nn
+
+from ._utils import load_state_dict_from_url
+
 
 __all__ = [
     "ResBlock",
@@ -10,7 +13,14 @@ __all__ = [
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
+    "wavernn_10k_epochs_8bits_ljspeech",
 ]
+
+
+model_urls = {
+    'wavernn_10k_epochs_8bits_ljspeech': 'https://download.pytorch.org/models/'
+                                         'audio/wavernn_10k_epochs_8bits_ljspeech.pth',
+}
 
 
 class ResBlock(nn.Module):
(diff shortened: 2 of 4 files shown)


torchaudio/models/wavernn.py
```diff
@@ -1,18 +1,29 @@
...
     "MelResNet",
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
+    "wavernn",
 ]
 
 
+model_urls = {
+    # FIXME Replace URL by final one once determined
```
vincentqb, line 22: nit: please remove the comment since the URL has been updated.

torchaudio/models/wavernn.py
```diff
@@ -324,3 +335,21 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+def wavernn(pretrained: bool = True, progress: bool = True, **kwargs: Any):
+    r"""WaveRNN model based on the implementation from
+    `fatchord <https://github.com/fatchord/WaveRNN>`_.
+    Args:
+        pretrained (bool): If True, returns a model pre-trained on LJSpeech
+        progress (bool): If True, displays a progress bar of the download to stderr
+    """
+    kwargs.setdefault("upsample_scales", [5, 5, 11])
+    kwargs.setdefault("n_classes", 30)
+    kwargs.setdefault("n_output", 128)
```
vincentqb, line 349: Let's add those as [default](https://github.com/pytorch/audio/blob/master/torchaudio/models/wavernn.py#L230) to the main WaveRNN class instead, with the other defaults.

I'm assuming you took the defaults from [here](https://github.com/pytorch/audio/blob/master/examples/pipeline_wavernn/main.py#L83).

PetrochukM:
Just as a reminder, this is a real person's voice and identity, specifically Linda Johnson's voice. Do you think we should get in writing her consent to publish her voice before it's publicly accessible to everyone? I think that's the right thing to do...

mthrok:
Hi @PetrochukM 

Thanks for bringing up the issue again.
@vincentqb Can you address these concerns?

@PetrochukM I am curious to learn your opinion on publishing the pre-trained model for vocoder.
Would you think that publishing the model is risky and would not do if you were a maintainer of the repo?

vincentqb:
@dongreenberg -- can you comment here? following [internal: april 21](https://fb.quip.com/ZbB5AqwjiQtr
)

vincentqb:
nit: flake8 :)

torchaudio/models/wavernn.py
```diff
@@ -324,3 +333,35 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
 
         # bring back channel dimension
         return x.unsqueeze(1)
+
+
+def wavernn(pretrained: bool = True, progress: bool = True, **kwargs: Any) -> WaveRNN:
```
vincentqb, line 338: let's stay closer to the convention set in [here](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py): we have a helper function `_wavernn` that passes the kwargs to WaveRNN, and a particular one called `wavernn_10k_epochs_8bits_ljspeech`

torchaudio/models/wavernn.py
```diff
@@ -1,18 +1,27 @@
...
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
+    "wavernn",
 ]
 
 
+model_urls = {
+    'wavernn': 'https://download.pytorch.org/models/audio/wavernn_10k_epochs_8bits_ljspeech.pth',
+}
```
vincentqb, line 22: in case this line is too long:
```
model_urls = {
    'wavernn_10k_epochs_8bits_ljspeech': (
           'https://download.pytorch.org/models/audio/'
           'wavernn_10k_epochs_8bits_ljspeech.pth'
     ),
}
```

PetrochukM:
> @PetrochukM I am curious to learn your opinion on publishing the pre-trained model for vocoder.
> Would you think that publishing the model is risky and would not do if you were a maintainer of the repo?

I think it's OKAY as long as the voice actor(s) have given their permission to publish their voice. 

I think it'd be really cool if `torchaudio` standardized asking voice actor(s) for consent, and including a short consent document along with the published model.

torchaudio/models/wavernn.py
```diff
@@ -324,3 +334,37 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+        'hop_length': 275,
+        'n_res_block': 10,
+        'n_rnn': 512,
+        'n_fc': 512,
+        'kernel_size': 5,
+        'n_freq': 80,
+        'n_hidden': 128,
+        'n_output': 128
+    }
+    configs.update(kwargs)
```
vincentqb, line 369: to follow the convention in [here](https://github.com/pytorch/vision/blob/master/torchvision/models/resnet.py), we should have `kwargs.update(configs)` or otherwise just update the dictionary directly.

torchaudio/models/wavernn.py
```diff
@@ -324,3 +334,37 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+def _wavernn(arch: str, pretrained: bool, progress: bool, **kwargs: Any) -> WaveRNN:
+    model = WaveRNN(**kwargs)
+    if pretrained:
+        state_dict = load_state_dict_from_url(model_urls['wavernn'],
+                                              progress=progress)
+        model.load_state_dict(state_dict)
+    return model
+
+
+def wavernn_10k_epochs_8bits_ljspeech(pretrained: bool = True, progress: bool = True, **kwargs: Any) -> WaveRNN:
```
mthrok, line 348: Note: `pretrained: bool = True` is not future proof.
In case there is a newer pretrained parameters, and we want to be able to have both parameters available, this signature makes it difficult to have data backward compatibility, friendly UX for picking the right data, and maintainability. It is better to define some ID which uniquely determines which pretrained data it should fetch.

mthrok: I feel that arbitral `kwargs` and loading pretrained model (`pretrained = True`) do not work well together.
There should be a check that prevents users from providing `kwargs` that is not compatibile with pretrained model in the case `pretrained=True`, or simply separate the interface for fetching pretrained model and building un-trained model.

mthrok: What does `10k_epochs` mean when `pretrained = False`?

torchaudio/models/wavernn.py
```diff
@@ -324,3 +334,37 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
 
         # bring back channel dimension
         return x.unsqueeze(1)
+
+
+def _wavernn(arch: str, pretrained: bool, progress: bool, **kwargs: Any) -> WaveRNN:
```
mthrok, line 339: Looking at the code, there is no point of having this as a helper function. Performing this directly in `wavernn_10k_epochs_8bits_ljspeech` function is more readable.

Note that torchvision's resnet adopts this pattern because it has many factory functions that only differ in configuration, and [Don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) is a common theme in software engineering.

torchaudio/models/wavernn.py
yangarbiter: We are actually following the convention from [torchvision](https://github.com/pytorch/vision/blob/master/torchvision/models/vgg.py). In the future, if there are new model (or new pretrained weights), we can just add a new function. In their example, they have separate functions for newer pretrained parameters Vgg11, Vgg13, Vgg16, ... and so on.

I also like what you have suggested, which seems to be changing the `pretrained` parameter into taking a string that matches corresponding weight URL. But the downside is that it is different from torchvision, and people maybe more familiar with their convention.

> What does `10k_epochs` mean when `pretrained = False`?
good point, `10k_epochs` don't mean anything when `pretrained = False`.

@vincentqb Please let me know your comment on which interface to take, the one that follows torchvision or using the `pretrained` parameter to take in some ID for each weight URL. Also please let me know if I missed anything :)

mthrok: Notice that `torchvision`'s code has some legacy/historic constraints and IIRC, @datumbox had some problem with it. There was a model code change that originated from the upstream PyTorch core, but there was no way to adopt the change without making BC-breaking. If we copy them as-is, we will go through the same issue in future. We can take the inspiration from `torchvision`, but sticking to it is a different concern.

yangarbiter: I've changed the interface to (get_pretrained_wavernn)[https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R352].

Please let me know what you think. :)

torchaudio/models/wavernn.py
yangarbiter: I think this should be resolved with the new interface.
Thanks.

vincentqb:
@PetrochukM -- thanks again for raising those concerns :) The previous discussion is in #776 and the author of the dataset @keithito [commented here](https://github.com/pytorch/audio/pull/776#issuecomment-660832473) that he has personally corresponded with Linda, and confirmed that she has been very supportive of having her recordings used as the basis of a public domain speech dataset. based on this, we will go ahead and publish the pre-trained weights. however, anyone using such pre-trained models should consult their own lawyers ahead of time, in a similar fashion to the notice given [here](https://github.com/pytorch/audio#disclaimer-on-datasets). please do let us know if you have any other concerns.
 
cc @dongreenberg

PetrochukM:
@vincentqb Thanks for addressing my concerns. The Linda Johnson dataset is now 4 years old, so I'm worried that her comments were made a long time ago. I'm worried that this dataset has been used much more widely than originally intended. 

Would it be okay if got some more clarification on the correspondence between Linda and Keith?

mthrok:
> @vincentqb Thanks for addressing my concerns!
> 
> The Linda Johnson dataset is now 4 years old (before Tacotron-2 was even published), so I'm worried that her comments were made a long time ago. I'm worried that this dataset has been used much more widely than originally intended.
> 
> Would it be okay if got some more clarification on the correspondence between Linda and Keith?

@vincentqb Both the comment of @keithito and the internal document you are pointing are about the copy right of the data set (and the derived copy right of a model trained with the dataset). To me it looks different from the points and concerns @PetrochukM is bringing up. I do not think we should make a rushed decision to make it available as this seems very sensitive matter.

torchaudio/models/_utils.py
```diff
@@ -0,0 +1,8 @@
+try:
```
mthrok, line 1: Since this `_utils` is only used by WaveRNN and we do not have a plan to use this in other place, can you place this in `wavernn.py` directly?

torchaudio/models/wavernn.py
```diff
@@ -1,18 +1,41 @@
...
     "ResBlock",
     "MelResNet",
     "Stretch2d",
     "UpsampleNetwork",
     "WaveRNN",
+    "get_pretrained_wavernn",
 ]
 
 
+model_config_and_urls: Dict[str, Tuple[str, Dict[str, Any]]] = {
```
mthrok, line 20: Since this is a private global object, please prefix it with underscore and capitalize all.

torchaudio/models/wavernn.py
```diff
@@ -324,3 +347,27 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+    - wavernn_10k_epochs_8bits_ljspeech
+
+        WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
+        The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
+
+    Args:
+        checkpoint_name (str): The name of the checkpoint to load.
+        progress (bool): If True, displays a progress bar of the download to stderr.
+    """
+    if checkpoint_name in model_config_and_urls:
```
mthrok, line 366: 1. Typically, argument validation is performed at the beginning of the function, so that the rest of the function logic can assume that they are valid.
2. When the logic in `if` clause is longer than `else` (or `else` is a few line), flipping the the condition will make the code more readable.
3. `else` is unnecessary.

```python
def get_pretrained_wavernn(...):
    if checkpoint_name not in _MODEL_CONFIG_AND_URLS:
        raise ValueError(...)

    url, configs = _MODEL_CONFIG_AND_URLS[checkpoint_name]
    ...
    return model
```

torchaudio/models/wavernn.py
```diff
@@ -324,3 +347,27 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+        url, configs = model_config_and_urls[checkpoint_name]
+        model = WaveRNN(**configs)
+        state_dict = load_state_dict_from_url(url, progress=progress)
+        model.load_state_dict(state_dict)
+        return model
+    else:
+        raise ValueError("The model_name `{}` is not supported.".format(checkpoint_name))
```
mthrok, line 373: The argument is `checkpoint_name` so I think it's better to stick to it.

mthrok (commented):
Overall, it looks good.

torchaudio/models/_utils.py
yangarbiter: What I was originally thinking is that if we want to add pertained weights for [Tacotron2](https://github.com/pytorch/audio/pull/1621) in the future, we may want to use the similar function. But since this is only an internal function, the cost of moving this function back to _utils.py should be low in the future. So I am okey with either way. Do you still think it is better to keep this function in `wavernn.py`?

torchaudio/models/wavernn.py
yangarbiter: Thanks for the comment.
It is fixed [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R22)

torchaudio/models/wavernn.py
yangarbiter: Thanks for the comment, it is addressed [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R371).

torchaudio/models/wavernn.py
yangarbiter: Thanks for the comment.
It is addressed [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R371).

torchaudio/models/_utils.py
yangarbiter: I've removed this file for now. We can decide whether to put it back in the future if needed.

vincentqb (approved):
LGTM, but @mthrok do you have any other feedback? 

i'll also let @mthrok and @dongreenberg follow-up on [comment](https://github.com/pytorch/audio/pull/1612#issuecomment-877451268).

dongreenberg:
Closing the loop on this. We had an internal review and had our legal team analyze the license to see whether this is in the scope of the license, which they deem it to be.

torchaudio/models/wavernn.py
```diff
@@ -324,3 +349,27 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
 
         # bring back channel dimension
         return x.unsqueeze(1)
+
+
+def get_pretrained_wavernn(checkpoint_name: str, progress: bool = True) -> WaveRNN:
```
mthrok, line 354: Can we simply call this `wavernn`?

I am not a fun of progress bar as it often messes up the log when running the code in an non-interactive way, while providing almost no benefit. Do you think it is mission-critical?

torchaudio/models/wavernn.py
```diff
@@ -324,3 +349,27 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+    r"""Get pretrained WaveRNN model.
+
+    Here are the available checkpoints:
+
+    - wavernn_10k_epochs_8bits_ljspeech
```
mthrok, line 359: Why not put this under `checkpoint_name` section? I think that is more suitable place.

https://262737-90321822-gh.circle-artifacts.com/0/docs/models.html#get-pretrained-wavernn

torchaudio/models/wavernn.py
yangarbiter: > Can we simply call this wavernn?
Sure!

For the progress bar, I am okay with removing them. With the model size and the average internet speed, downloading the model won't let most of the users feel like the program is dead. But it may still provide a bit of function for people with slow internet.

yangarbiter: I've removed the ``progressbar`` for now [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R370). Please let me know if you think we should add it back and set. the default to False instead.

torchaudio/models/wavernn.py
yangarbiter: I've moved it under the checkpoint_name [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R358)

torchaudio/models/wavernn.py
```diff
@@ -1,18 +1,43 @@
-from typing import List, Tuple
+from typing import List, Tuple, Dict, Any
 
 import torch
 from torch import Tensor
 from torch import nn
+try:
+    from torch.hub import load_state_dict_from_url
+except ImportError:
+    from torch.utils.model_zoo import load_url as load_state_dict_from_url
```
mthrok, line 9: Is it necessary to try `torch.hub` first?
`from torch.utils.model_zoo import load_url` seems to suffice.

mthrok: Looking at the document, the code has been moved to `torch.hub.load_state_dict_from_url`.
`torchaudio` only supports the matching PyTorch that is released at the same time, so only `from torch.hub import load_state_dict_from_url` should work.

yangarbiter: Thank you for pointing it out, I didn't realize that.
It's addressed [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R6)

torchaudio/models/wavernn.py
```diff
@@ -324,3 +346,24 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+    Args:
+        checkpoint_name (str): The name of the checkpoint to load. Available checkpoints:
+
+            - wavernn_10k_epochs_8bits_ljspeech:
+
+                WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
+                The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
+    """
+    if checkpoint_name not in _MODEL_CONFIG_AND_URLS:
+        raise ValueError("The checkpoint_name `{}` is not supported.".format(checkpoint_name))
```
mthrok, line 363: When validating a value and there is a (small number of) finite set of valid values, listing them out is more user-friendly.

Imagine that I tried to pass `wavernn_10k_epochs_8bits_ljspeech` but misspelled `wavernn_10k_epochs_8bits_ljspeeck`. If the error message only tells me it's invalid, then I have to search for the documentation to see what is correct. If the error message also tells what are the valid choices, then I can copy-paste the valid ones from the error message and retry at instant.

`not supported` is correct but sounds like it's planned to be supported, and I think `unexpected` is more regularly used.

`str.format` method is fine, but typically, f-string is more readable and the code becomes shorter.

```suggestion
        raise ValueError(
            f"Unexpected checkpoint_name: '{checkpoint_name}'. "
            f"Valid choices are; {list(_MODEL_CONFIG_AND_URLS.keys())}")
```

torchaudio/models/wavernn.py
```diff
@@ -324,3 +346,24 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
         return x.unsqueeze(1)
+
+
+def wavernn(checkpoint_name: str) -> WaveRNN:
+    r"""Get pretrained WaveRNN model.
+
+    Args:
+        checkpoint_name (str): The name of the checkpoint to load. Available checkpoints:
+
+            - wavernn_10k_epochs_8bits_ljspeech:
```
mthrok, line 357:
```suggestion
            - ``"wavernn_10k_epochs_8bits_ljspeech"``
```

torchaudio/models/wavernn.py
yangarbiter: Thanks for pointing it out. These are definitely better designs.
I've fixed them [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R363).

torchaudio/models/wavernn.py
yangarbiter: Thanks for point it out. I've fixed them [here](https://github.com/pytorch/audio/pull/1612/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R357).

PetrochukM:
I wanted to check in. Are y'all going to publish Linda Johnson's voice without asking her for explicit and informed permission?

torchaudio/models/wavernn.py
```diff
@@ -324,3 +346,26 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+def wavernn(checkpoint_name: str) -> WaveRNN:
+    r"""Get pretrained WaveRNN model.
+
+    Args:
+        checkpoint_name (str): The name of the checkpoint to load. Available checkpoints:
+
+            - ``"wavernn_10k_epochs_8bits_ljspeech"``:
+
+                WaveRNN model trained with 10k epochs and 8 bits depth waveform on the LJSpeech dataset.
+                The model is trained using the default parameters and code of the examples/pipeline_wavernn/main.py.
```
mthrok, line 360: nit: Maybe adding hyperlink. https://github.com/pytorch/audio/tree/master/examples/pipeline_wavernn

mthrok (approved):

dongreenberg:
> I wanted to check in. Are y'all going to publish Linda Johnson's voice for the public to use without asking her for explicit and informed permission?

Most definitely not.

torchaudio/models/wavernn.py
datumbox: @mthrok Apologies for not seeing this earlier. I see that the final PR adopts a different approach that the one proposed in this commit. Indeed even in TorchVision we are in the process of assessing better support of versioning. I'll get to it once I'm back.

Out of curiosity, I see that the adopted solution does not permit for creating a model of specific configuration WITHOUT pre-trained weights. Is this something you would like to support on the future? Aka load the weight with specific architecture config/params but without loading pre-trained weights.

mthrok: >  Is this something you would like to support on the future?

@datumbox I am open to it if there is a feature request with valid use cases.
It feels to me that, it is just about passing the fixed configuration to the constructor of `WaveRNN` class, adding that feature to this factory function is nothing but a syntax sugar. So it's nice if we have a good justification for the syntax sugar first. In the case of torchvision's ResNet, there are bunch of different yet popular configurations across different tasks, so I think it's well justified. For WaveRNN, I am not sure a.t.m. And I think it's not too late to wait for someone with expertise in TTS domain to show up and advise us what are the common use cases.

datumbox: Sounds good. Can't comment on what's best on the Audio domain as my knowledge is limited. I'll take your input into account for the proposed solution on vision to see if we can converge to a cross domain solution.

mthrok: Cool. just FYI in case of TTS or STT, we always want to have the possibility of multiple pre-trained models for the same architecture because in simple case STT models are trained per language.

BTW: If I were to add that feature now, I think I would make a signature like the following;

```pyton
def wave_rnn(checkpoint=None, config=None) -> Module:
    if checkpoint is not None and config is None:
        # return pre-trained
    if checkpoint is None and config is not None:
        # return un-trained
    raise ValueError("One of (and only one of) `checkpoint` or `config` must be given.")
 ```

datumbox: There are pros/cons with having different params for checkpoint and config. It's a lengthy topic I don't want to open here cause it's going to get buried. I think we should catch up once I am back to discuss options and get your thoughts.

PR #1615, opened 2021-06-28
Remove fused_log_softmax option from RNNT Loss

carolineechen:
Note: merge after #1610

Remove `fused_log_softmax` option and keep the behavior of the default `True` option. `fused_log_softmax=False` option was only relevant when `reuse_logits_for_grads=True`, but this behavior is no longer supported after #1610.


vincentqb (approved):
LGTM

PR #1616, opened 2021-06-28
Remove gradient variable from RNNT Loss Python code

carolineechen:
`gradients` variable is not accessed in Python code


vincentqb (approved):
LGTM

PR #855, opened 2020-08-04
Add tacotron model

jimchen90:
This is to add the first part of tacotron2 model.

PR #1620, opened 2021-06-30
Improve RNNT Loss docstrings

carolineechen:
contains changes from #1610 #1615, merge after

PR #1621, opened 2021-07-06
Add Tacotron2 model

yangarbiter:
Closes #855.

Porting Tacotron2 from [Nvidia](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/Tacotron2/tacotron2/model.py).

Here are the differences between Nvidia's and this version:

- These functions and classes are moved into tacotron2.py instead of separated files.
```
from tacotron2_common.layers import ConvNorm, LinearNorm
from tacotron2_common.utils import to_gpu, get_mask_from_lengths
```

- The forward function of the original Encoder is not torchscriptable due to [this line](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/Tacotron2/tacotron2/model.py#L218). But removing it does not seem to affect the result. So the current version of forward for Encoder is torchscriptable.

- New tests added.

The tests are similar to the ones in #855, which tests the output dimension of Encoder, Decoder, and Tacotron2. I've also created separate tests for GPU.

I've also added a torchscript consistency test for the Decoder. ~~Encoder and Tacotron2 are not torchscriptable because they've called the `flatten_parameters` function and it is not torchscriptable (https://github.com/pytorch/pytorch/issues/46375).~~ After checking with @gibiansky, `flatten_parameters` can be moved to __init__ so now Encoder and Tacotron2 are torchscriptable.

- Some variables are renamed to match [torchaudio conventions](https://github.com/pytorch/audio#conventions)

CC @vincentqb @mthrok @carolineechen. It would be great if you can leave some comments. Thanks in advance.


test/torchaudio_unittest/models/tacotron2/model_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+        torch.random.manual_seed(40)
+        ts_output = ts_func(*tensors)
+
+        self.assertEqual(ts_output, output)
+
+
+class TestEncoder(TorchaudioTestCase, TorchscriptConsistencyMixin):
+
+    def _get_model(self, encoder_n_convolutions=3, encoder_embedding_dim=512, encoder_kernel_size=5):
```
mthrok, line 31: There seems to be no benefit of making this into a method. How about calling `_Encoder` directly?

torchaudio/models/__init__.py
```diff
@@ -19,4 +20,5 @@
     'wav2vec2_base',
     'wav2vec2_large',
     'wav2vec2_large_lv60k',
+    '_Tacotron2',
```
mthrok, line 23: Why exporting the class suffixed with underscore?

torchaudio/models/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+from typing import Tuple, List, Optional, Union
+
+import torch
+from torch import nn
+from torch import Tensor
+from torch.nn import functional as F
+
+
+__all__ = [
+    "_Tacotron2",
```
mthrok, line 38: It looks strange to have the class marked private in `__all__`.

torchaudio/models/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+        >>> input = torch.rand(10, 20, 30)
+        >>> output = encoder(input)  # shape: (10, 30, 512)
+    """
+
+    def __init__(self,
+                 encoder_n_convolutions: int,
+                 encoder_embedding_dim: int,
+                 encoder_kernel_size: int,
+                 ) -> None:
+        super(_Encoder, self).__init__()
```
mthrok, line 346: This is Python 2 style. `super().__init__()` should suffice.

test/torchaudio_unittest/models/tacotron2/model_test.py
```diff
@@ -0,0 +1,202 @@
...
+        """Validate the output dimensions of a Encoder.
+        """
+
+        x = torch.rand(n_batch, encoder_embedding_dim, n_seq, device=device, dtype=dtype)
+        input_lengths = torch.ones(n_batch, device=device, dtype=torch.int32) * n_seq
+        out = model(x, input_lengths)
+
+        assert out.size() == (n_batch, n_seq, encoder_embedding_dim)
+
+    @parameterized.expand([(32, 64, 512, torch.float32, torch.device('cpu'))])
```
mthrok, line 44: Parameterized here is redundant thus confusing. Can you simply define them inside of the method?

test/torchaudio_unittest/models/tacotron2/model_test.py
```diff
@@ -0,0 +1,202 @@
...
+    def test_cpu_encoder_output(self, n_batch, n_seq, encoder_embedding_dim, dtype, device):
```
mthrok, line 45: Please write a short script of what is "intended" in this test.
For example; "A should meet condition B".

Running `pytest --collect-only` will give this one-liner as the summary of the tests.

test/torchaudio_unittest/models/tacotron2/model_test.py
```diff
@@ -0,0 +1,202 @@
...
+        model = self._get_model(encoder_embedding_dim=encoder_embedding_dim).to(device).eval()
+        self._output_shape_test(model, n_batch, encoder_embedding_dim, n_seq, device, dtype)
+
+    @parameterized.expand([(32, 64, 512, torch.float32, torch.device('cuda'))])
+    @skipIfNoCuda
+    def test_gpu_encoder_output(self, n_batch, n_seq, encoder_embedding_dim, dtype, device):
```
mthrok, line 51: Please split CPU tests and CUDA tests into separate files. Currently we do not run CUDA tests in fb infra and skipping will trigger unnecessary alerts.

test/torchaudio_unittest/models/tacotron2/model_test.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def _output_shape_test(self,
+                           model,
+                           n_mels,
+                           n_batch,
+                           encoder_embedding_dim,
+                           n_seq,
+                           device,
+                           dtype):
+        """Validate the output dimensions of a Decoder.
```
mthrok, line 102: Can you replace "validate" with something more descriptive?

What conditions output dimensions should be satisfying?

mthrok (commented):
Had a quick review.

docs/source/refs.bib
```diff
@@ -38,6 +38,14 @@ @misc{kalchbrenner2018efficient
...
       primaryClass={cs.SD}
 }
+@inproceedings{shen2018natural,
+      title={Natural TTS Synthesis by Conditioning WaveNet on Mel Spectrogram Predictions},
+      author={Shen, Jonathan and Pang, Ruoming and Weiss, Ron J and Schuster, Mike and Jaitly, Navdeep and Yang, Zongheng and Chen, Zhifeng and Zhang, Yu and Wang, Yuxuan and Skerrv-Ryan, Rj and others},
+      year={2017},
+      eprint={1712.05884},
+      archivePrefix={arXiv},
+      primaryClass={cs.CL}
+}
```
mthrok, line 48: Looks like `docs/source/models.rst` is not updated.
But I cannot tell if `Tacotoron2` class is public or private.
If it's public, it has to be added to `models.rst`.

torchaudio/models/tacotron2.py
yangarbiter: For now, we want to keep it private because we only have the model in this PR but no training/inference pipeline yet. After several more PRs when we have everything merged, we will change Tacotron2 from private to public class.

docs/source/refs.bib
yangarbiter: We plan to make it public after we include the training pipeline in the examples. If this is the case, should I remove this for now?

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: I've added more comments for each of the test functions (here)[https://github.com/pytorch/audio/pull/1621/files#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR29], but I am not sure why they didn't show up on my `pytest --coolect-only`

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: Thanks for the comments!
I've moved them inside the method.

torchaudio/models/tacotron2.py
yangarbiter: Fixed in the new commit.
Thanks.

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: Fixed in the new commit.
Thanks.

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: I've split it into test/torchaudio_unittest/models/tacotron2/model_test_cpu_test.py and test/torchaudio_unittest/models/tacotron2/model_test_gpu_test.py
Thanks.

torchaudio/models/tacotron2.py
yangarbiter: Fixed in the new commit.
Thanks!

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: I've changed it to 
```python
        """Feed tensors with specific shape into Tacotron2 Decoder and validate
        that it outputs with a tensor with expected shape.
        """
```
Please let me know if this is clear enough or more detailed comments should be written.
Thanks.

torchaudio/models/__init__.py
yangarbiter: Previously I use underscore to represent prototype.
I've moved Tacotron2 to the `torchaudio.prototype` directory to avoid confusion.
Please let me know if this is the right place to put at this stage.
Thanks.

torchaudio/models/tacotron2.py
yangarbiter: Same as above https://github.com/pytorch/audio/pull/1621#discussion_r666244850.
I've moved Tacotron2 to the `torchaudio.prototype` directory to avoid confusion.
Please let me know if this is the right place to put at this stage.
Thanks.

test/torchaudio_unittest/models/tacotron2/model_test.py
yangarbiter: offline chat with @mthrok. `pytest --collect-only -v` will do the job.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+        decoder_rnn_dim (int, optional): Number of units in decoder LSTM (Default: ``1024``).
+        prenet_dim (int, optional): Number of ReLU units in prenet layers (Default: ``256``).
+        max_decoder_steps (int, optional): Maximum number of output mel spectrograms (Default: ``2000``).
+        gate_threshold (float, optional): Probability threshold for stop token (Default: ``0.5``).
+        p_attention_dropout (float, optional): Dropout probability for attention LSTM (Default: ``0.1``).
+        p_decoder_dropout (float, optional): Dropout probability for decoder LSTM (Default: ``0.1``).
+        postnet_embedding_dim (int, optional): Postnet embedding dimension (Default: ``512``).
+        postnet_kernel_size (int, optional): Postnet kernel size (Default: ``5``).
+        postnet_n_convolutions (int, optional): Number of postnet convolutions (Default: ``5``).
+        decoder_no_early_stopping (bool, optional): Stop decoding when all samples are finished (Default: ``False``).
```
vincentqb, line 746:
```suggestion
        mask_padding (bool, optional): Use mask padding (Default: ``False``).
        n_mel (int, optional): Number of mel bins (Default: ``80``).
        n_symbol (int, optional): Number of symbols for the input text (Default: ``148``).
        n_frame_per_step (int, optional): Number of frames processed per step, only 1 is supported (Default: ``1``).
        symbol_embedding_dim (int, optional): Input embedding dimension (Default: ``512``).
        encoder_n_convolution (int, optional): Number of encoder convolutions (Default: ``3``).
        encoder_kernel_size (int, optional): Encoder kernel size (Default: ``5``).
        encoder_embedding_dim (int, optional): Encoder embedding dimension (Default: ``512``).
        decoder_rnn_dim (int, optional): Number of units in decoder LSTM (Default: ``1024``).
        decoder_max_step (int, optional): Maximum number of output mel spectrograms (Default: ``2000``).
        decoder_dropout (float, optional): Dropout probability for decoder LSTM (Default: ``0.1``).
        decoder_early_stopping (bool, optional): Continue decoding after all samples are finished (Default: ``True``).
        attention_rnn_dim (int, optional): Number of units in attention LSTM (Default: ``1024``).
        attention_hidden_dim (int, optional): Dimension of attention hidden representation (Default: ``128``).
        attention_location_n_filter (int, optional): Number of filters for attention model (Default: ``32``).
        attention_location_kernel_size (int, optional): Kernel size for attention model (Default: ``31``).
        attention_dropout (float, optional): Dropout probability for attention LSTM (Default: ``0.1``).
        prenet_dim (int, optional): Number of ReLU units in prenet layers (Default: ``256``).
        postnet_n_convolution (int, optional): Number of postnet convolutions (Default: ``5``).
        postnet_kernel_size (int, optional): Postnet kernel size (Default: ``5``).
        postnet_embedding_dim (int, optional): Postnet embedding dimension (Default: ``512``).
        gate_threshold (float, optional): Probability threshold for stop token (Default: ``0.5``).
```

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+                with shape (n_batch, n_mels, mel_specgram_lengths.max()).
+            mel_specgram_postnet (Tensor): Mel spectrogram after Postnet
+                with shape (n_batch, n_mels, mel_specgram_lengths.max()).
+            stop_token (Tensor): The output for stop token at each time step
+                with shape (n_batch, mel_specgram_lengths.max()).
+            alignment (Tensor): Sequence of attention weights from the decoder.
+                with shape (n_batch, mel_specgram_lengths.max(), text_lengths.max()).
+        """
+
+        text_lengths, mel_specgram_lengths = text_lengths.data, mel_specgram_lengths.data
```
vincentqb, line 839: why is `.data` needed here? in general, we should never need to call `.data`.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+                                gate_threshold, p_attention_dropout,
+                                p_decoder_dropout,
+                                not decoder_no_early_stopping)
+        self.postnet = _Postnet(n_mels, postnet_embedding_dim,
+                                postnet_kernel_size,
+                                postnet_n_convolutions)
+
+    def _parse_output(self,
+                      outputs: Tuple[Tensor, Tensor, Tensor, Tensor],
+                      output_lengths: Tensor) -> Tuple[Tensor, Tensor, Tensor, Tensor]:
```
vincentqb, line 798: since this is a single purpose function, i'd recommend moving this code in place, with a comment.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,851 @@
...
+        forcing mode, which is generally used for training.
+
+        The input ``text`` should be padded with zeros to length text_lengths.max().
+        The input ``mel_specgram`` should be padded with zeros to length mel_specgram_lengths.max().
+
+        Args:
+            text (Tensor): The input text to Tacotron2 with shape (n_batch, text_lengths.max()).
+            text_lengths (Tensor): The length of each text with shape (n_batch).
+            mel_specgram (Tensor): The target mel spectrogram with shape (n_batch, n_mels, mel_specgram_lengths.max()).
+            mel_specgram_lengths (Tensor): The length of each mel spectrogram with shape (n_batch).
```
vincentqb, line 826:
```suggestion
            text (Tensor): The input text to Tacotron2 with shape (n_batch, max of text_lengths).
            text_lengths (Tensor): The length of each text with shape (n_batch).
            mel_specgram (Tensor): The target mel spectrogram with shape (n_batch, n_mel, max of mel_specgram_lengths).
            mel_specgram_lengths (Tensor): The length of each mel spectrogram with shape (n_batch).
```
or follow convention [here](https://github.com/pytorch/audio/pull/1620/commits/34cf5210ddfae17c89e1f013b261abb65b4b2e63) for instance. thoughts?

torchaudio/prototype/tacotron2.py
yangarbiter: Addressed int the new commit.
Thanks.

torchaudio/prototype/tacotron2.py
yangarbiter: addressed in the new [commit](https://github.com/pytorch/audio/pull/1621/commits/5fccd01181b8a5d5aa3e33f0f0102c27bd154223).
Thanks.

torchaudio/prototype/tacotron2.py
yangarbiter: addressed in the new [commit](https://github.com/pytorch/audio/pull/1621/commits/5fccd01181b8a5d5aa3e33f0f0102c27bd154223).
Thanks.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,847 @@
...
+from torch import Tensor
+from torch.nn import functional as F
+
+
+__all__ = [
+    "Tacotron2",
+]
+
+
+class _LinearNorm(torch.nn.Module):
```
mthrok, line 42: I understand that this is being ported from Nvidia's code but I wonder if we can change this to a function. After all, what this class does is instantiate a class from PyTorch and initialize it with Xavier formula. This can be simply replaced with

```python
def _get_linear_norm(...):
    linear = torch.nn.Linear(...)
    torch.nn.init.xavier_uniform_(linear.weight, ...)
    return linear
```

Of course, the resulting model structure is slightly different and if you have a pre-trained model already, then a model surgery will be necessary to use it.
But in my opinion, it is more maintainable if there is less custom layers.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,847 @@
...
+
+        torch.nn.init.xavier_uniform_(
+            self.conv.weight,
+            gain=torch.nn.init.calculate_gain(w_init_gain))
+
+    def forward(self, signal):
+        return self.conv(signal)
+
+
+def get_mask_from_lengths(lengths: Tensor) -> Tensor:
```
mthrok, line 100: nit: since other non-public interfaces are prefixed with underscore, this function looks public API. (I am guessing it is not)

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,847 @@
...
+        r"""Gets all zeros frames to use as the first decoder input.
+
+        Args:
+            memory (Tensor): Encoder outputs (n_batch, text_lengths.max(), encoder_embedding_dim)
+
+        Returns:
+            decoder_input (Tensor): all zeros frames (n_batch, text_lengths.max(), n_mel * n_frames_per_step)
+        """
+
+        B = memory.size(0)
```
mthrok, line 476: Can we replace `B` with something more readable? something like `batch_size`.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,847 @@
...
+            decoder_hidden (Tensor): Hidden state of the decoder LSTM with shape (n_batch, decoder_rnn_dim).
+            decoder_cell (Tensor): Hidden state of the decoder LSTM with shape (n_batch, decoder_rnn_dim).
+            attention_weights (Tensor): Attention weights with shape (n_batch, text_lengths.max()).
+            attention_weights_cum (Tensor): Cumulated attention weights with shape (n_batch, text_lengths.max()).
+            attention_context (Tensor): Context vector with shape (n_batch, encoder_embedding_dim).
+            processed_memory (Tensor): Processed encoder outputs
+                with shape (n_batch, text_lengths.max(), attention_hidden_dim).
+        """
+        B = memory.size(0)
+        MAX_TIME = memory.size(1)
```
mthrok, line 505: Can we use lower case `max_time`?
All capitalized looks like this is global variable.

yangarbiter: Thanks for pointing it out. It's addressed [here](https://github.com/pytorch/audio/pull/1621/commits/68b127377f8f8f8b6ae64c06ff4f6de885335d57#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR502).

torchaudio/prototype/tacotron2.py
yangarbiter: Thanks for pointing it out. It's addressed [here](https://github.com/pytorch/audio/pull/1621/commits/68b127377f8f8f8b6ae64c06ff4f6de885335d57#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR501).

torchaudio/prototype/tacotron2.py
yangarbiter: Thanks for pointing it out. It's addressed [here](https://github.com/pytorch/audio/pull/1621/commits/68b127377f8f8f8b6ae64c06ff4f6de885335d57#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR97)

torchaudio/prototype/tacotron2.py
yangarbiter: I've changed it [this way](https://github.com/pytorch/audio/pull/1621/files#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR42). Please let me know how you think :).

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
```diff
@@ -0,0 +1,171 @@
...
+
+        torch.random.manual_seed(40)
+        ts_output = ts_func(*tensors)
+
+        self.assertEqual(ts_output, output)
+
+
+class Tacotron2EncoderTests(TestBaseMixin, TorchscriptConsistencyMixin):
+
+    def _get_model(self, encoder_embedding_dim):
```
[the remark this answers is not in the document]
yangarbiter: Thanks for pointing it out!
It is addressed [here](https://github.com/pytorch/audio/pull/1621/commits/68b127377f8f8f8b6ae64c06ff4f6de885335d57#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR32)

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
```diff
@@ -0,0 +1,171 @@
...
+        x = torch.rand(n_batch, encoder_embedding_dim, n_seq, device=self.device, dtype=self.dtype)
+        input_lengths = torch.ones(n_batch, device=self.device, dtype=torch.int32) * n_seq
+        out = model(x, input_lengths)
+
+        assert out.size() == (n_batch, n_seq, encoder_embedding_dim)
+
+
+class Tacotron2DecoderTests(TestBaseMixin, TorchscriptConsistencyMixin):
+
+    def _get_model(self, n_mel=80, encoder_embedding_dim=512):
```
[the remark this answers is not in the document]
yangarbiter: Thanks for pointing it out!
It is addressed [here](https://github.com/pytorch/audio/pull/1621/commits/68b127377f8f8f8b6ae64c06ff4f6de885335d57#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR59).

torchaudio/prototype/tacotron2.py
vincentqb: Similarly, please do the same for `_ConvNorm` and replace by `_get_conv_norm` with a similar structure.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,844 @@
...
+            )
+
+        self.convolutions.append(
+            nn.Sequential(
+                _ConvNorm(postnet_embedding_dim, n_mel,
+                          kernel_size=postnet_kernel_size, stride=1,
+                          padding=int((postnet_kernel_size - 1) / 2),
+                          dilation=1, w_init_gain='linear'),
+                nn.BatchNorm1d(n_mel))
+        )
```
vincentqb, line 300: Combining these blocks is more readable. 

```python
for i in range(postnet_n_convolution):
  in_channels = n_mel if i == 0 else postnet_embedding_dim
  out_channels = n_mel if i == postnet_n_convolution - 1 else postnet_embedding_dim
  init_gain = 'linear' if i == postnet_n_convolution - 1 else 'tanh'
  self.convolutions.append(
            nn.Sequential(
                _ConvNorm(in_channels, out_channels,
                          kernel_size=postnet_kernel_size, stride=1,
                          padding=int((postnet_kernel_size - 1) / 2),
                          dilation=1, w_init_gain='linear'),
                nn.BatchNorm1d(n_mel)
             )
        )
```

I'd also prefer a way of removing the sequential, but giving how it is used below, I'd be ok keeping it as is.

torchaudio/prototype/tacotron2.py
```diff
@@ -0,0 +1,844 @@
...
+    """
+
+    def __init__(self,
+                 encoder_embedding_dim: int,
+                 encoder_n_convolution: int,
+                 encoder_kernel_size: int,
+                 ) -> None:
+        super().__init__()
+
+        convolutions = []
```
vincentqb, line 345: please create `nn.ModuleList` directly here, and this will remove the line where the list is converted to ModuleList.

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
```diff
@@ -0,0 +1,257 @@
...
+        r"""Feed tensors with specific shape to Tacotron2 Decoder and validate
+        that it outputs with a tensor with expected shape.
+        """
+        n_batch, n_mel, n_seq, encoder_embedding_dim, n_time_steps = (
+            16,
+            80,
+            200,
+            256,
+            150,
+        )
```
vincentqb, line 133: nit: I recommend: 
```suggestion
        n_batch = 16
        n_mel = 80
        n_seq = 200
        encoder_embedding_dim = 256
        n_time_steps = 150
```

torchaudio/prototype/tacotron2.py
vincentqb: see also [torchaudio naming convention](https://github.com/pytorch/audio/blob/master/README.md#conventions)

torchaudio/prototype/tacotron2.py
yangarbiter: Thanks. Addressed [here](https://github.com/pytorch/audio/pull/1621/files#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR323).

torchaudio/prototype/tacotron2.py
yangarbiter: Thanks. Addressed [here](https://github.com/pytorch/audio/pull/1621/files#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR392)

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
yangarbiter: Thanks. Addressed [here](https://github.com/pytorch/audio/pull/1621/files#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR126).

yangarbiter:
Following @vincentqb suggestion, I've also run the code through [black](https://github.com/psf/black).

vincentqb (approved):
LGTM, @mthrok do you have any other feedback?

torchaudio/prototype/tacotron2.py
yangarbiter: Added [here](https://github.com/pytorch/audio/pull/1621/files#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR64)

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
```diff
@@ -0,0 +1,238 @@
...
+
+        model = _get_tacotron2_model(n_mels).to(self.device)
+        inputs = self._get_inputs(
+            n_mels, n_batch, max_mel_specgram_length, max_text_length
+        )
+        mel_out, mel_out_postnet, gate_outputs, _ = model(*inputs)
+
+        mel_out.sum().backward(retain_graph=True)
+        mel_out_postnet.sum().backward(retain_graph=True)
+        gate_outputs.sum().backward()
```
mthrok, line 238: Note: If we decide to land loss function in library, then we can use it here to mimic the expected use case more closely.

mthrok (approved):

PR #1622, opened 2021-07-07
Fix CircleCI environment

carolineechen:
add expecttest as dependency


mthrok (approved):
LGTM given the tests pass.

I wonder if we stop using `torch._internal` and use the new functional `torch.testing.assert_close`, then we do not need to install `expecttest`. Reading the description of `expecttest`, I get the impression that it's only relevant to PyTorch developers.

PR #1626, opened 2021-07-12
Remove skipIfRocm from soundfile.save_test test_fileobj_flac

yangarbiter:
Allow the test `soundfile.save_test::TestFileObject:test_fileobj_flac` to run in the Rocm stack.


mthrok (approved):
Looks good.

Can you update the description (and commit message) to tell why?

The test was broken from the beginning and for a reason unrelated to ROCm, but was disabled for ROCm in #1411. #1604 must fixed the test so that we are re-enabling it.

PR #1627, opened 2021-07-13
Add CMUDict dataset

yangarbiter:
Adding the [CMUDict dataset](http://www.speech.cs.cmu.edu/cgi-bin/cmudict?stress=-s&in=) for mapping English words into phonemes. Currently, we have two intended usage: 
(1) This can be used for the Tacotron2 pipeline for training a TTS model base on phonemes instead of characters.
(2) This can also be used to train a model that [maps each word to phonemes](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/43264.pdf).

Question:
How should we deal with the punctuation? For example,
```
!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T
"CLOSE-QUOTE  K L OW1 Z K W OW1 T
"DOUBLE-QUOTE  D AH1 B AH0 L K W OW1 T
"END-OF-QUOTE  EH1 N D AH0 V K W OW1 T
```

### DIFF
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -33,6 +33,14 @@ CMUARCTIC
   :special-members: __getitem__
 
 
+CMUDICT
+~~~~~~~~~
+
+.. autoclass:: CMUDICT
+  :members:
+  :special-members: __getitem__
+
+
 COMMONVOICE
 ~~~~~~~~~~~
 
--- a/docs/source/datasets.rst
+++ b/docs/source/datasets.rst
@@ -33,10 +33,10 @@ CMUARCTIC
   :special-members: __getitem__
 
 
-CMUDICT
+CMUDict
 ~~~~~~~~~
 
-.. autoclass:: CMUDICT
+.. autoclass:: CMUDict
   :members:
   :special-members: __getitem__
 
--- a/test/torchaudio_unittest/datasets/cmudict_test.py
+++ b/test/torchaudio_unittest/datasets/cmudict_test.py
@@ -1,6 +1,5 @@
 import os
 from pathlib import Path
-import string
 
 from torchaudio.datasets import CMUDict
 
@@ -10,45 +9,158 @@ from torchaudio_unittest.common_utils import (
 )
 
 
-def get_mock_dataset(root_dir):
+def get_mock_dataset(root_dir, return_punc=False):
     """
     root_dir: directory to the mocked dataset
     """
-    mocked_dictionary = [
+    header = [
         ";;; # CMUdict  --  Major Version: 0.07",
         ";;; ",
         ";;; # $HeadURL$",
+    ]
+
+    puncs = [
         "!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T",
         "\"CLOSE-QUOTE  K L OW1 Z K W OW1 T",
-        "AALIYAH  AA2 L IY1 AA2",
-        "AALSETH  AA1 L S EH0 TH",
-        "AAMODT  AA1 M AH0 T",
-        "AANCOR  AA1 N K AO2 R",
-        "AARDEMA  AA0 R D EH1 M AH0",
-        "AARDVARK  AA1 R D V AA2 R K",
+        "#HASH-MARK  HH AE1 M AA2 R K",
+        "%PERCENT  P ER0 S EH1 N T",
+        "&AMPERSAND  AE1 M P ER0 S AE2 N D",
+        "'END-INNER-QUOTE  EH1 N D IH1 N ER0 K W OW1 T",
+        "(BEGIN-PARENS  B IH0 G IH1 N P ER0 EH1 N Z",
+        ")CLOSE-PAREN  K L OW1 Z P ER0 EH1 N",
+        "+PLUS  P L UH1 S",
+        ",COMMA  K AA1 M AH0",
+        "--DASH  D AE1 SH",
+        "!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T",
+        "/SLASH  S L AE1 SH",
+        ":COLON  K OW1 L AH0 N",
+        ";SEMI-COLON  S EH1 M IY0 K OW1 L AH0 N",
+        "?QUESTION-MARK  K W EH1 S CH AH0 N M AA1 R K",
+        "{BRACE  B R EY1 S",
+        "}CLOSE-BRACE  K L OW1 Z B R EY1 S",
+        "...ELLIPSIS  IH2 L IH1 P S IH0 S",
     ]
+
+    punc_outputs = [
+        "!",
+        "\"",
+        "#",
+        "%",
+        "&",
+        "'",
+        "(",
+        ")",
+        "+",
+        ",",
+        "--",
+        "!",
+        "/",
+        ":",
+        ";",
+        "?",
+        "{",
+        "}",
+        "...",
+    ]
+
+    words = [
+        "3-D  TH R IY1 D IY2",
+        "'BOUT  B AW1 T",
+        "'CAUSE  K AH0 Z",
+        "'TWAS  T W AH1 Z",
+        "A  AH0",
+        "B  B IY1",
+        "C  S IY1",
+        "D  D IY1",
+        "E  IY1",
+        "F  EH1 F",
+        "G  JH IY1",
+        "H  EY1 CH",
+        "I  AY1",
+        "J  JH EY1",
+        "K  K EY1",
+        "L  EH1 L",
+        "M  EH1 M",
+        "N  EH1 N",
+        "O  OW1",
+        "P  P IY1",
+        "Q  K Y UW1",
+        "R  AA1 R",
+        "S  EH1 S",
+        "T  T IY1",
+        "U  Y UW1",
+        "V  V IY1",
+        "X  EH1 K S",
+        "Y  W AY1",
+        "Z  Z IY1",
+    ]
+
     mocked_symbols = [
-        "AA0", "AA1", "AA2", "AH0", "AO2", "D", "EH0", "EH2", "IY1", "K", "L", "M",
-        "N", "P", "OW1", "OY2", "R", "S", "SH", "T", "TH", "W", "V", "Z",
+        "AA1",
+        "AA2",
+        "AE1",
+        "AE2",
+        "AH0",
+        "AH1",
+        "AY1",
+        "B",
+        "CH",
+        "D",
+        "EH1",
+        "EH2",
+        "ER0",
+        "EY1",
+        "F",
+        "G",
+        "HH",
+        "IH0",
+        "IH1",
+        "IY0",
+        "IY1",
+        "IY2",
+        "JH",
+        "K",
+        "L",
+        "M",
+        "N",
+        "OW1",
+        "OY2",
+        "P",
+        "R",
+        "S",
+        "SH",
+        "T",
+        "TH",
+        "UH1",
+        "UW0",
+        "UW1",
+        "V",
+        "W",
+        "Y",
+        "Z",
     ]
 
     dict_file = os.path.join(root_dir, "cmudict-0.7b")
     symbol_file = os.path.join(root_dir, "cmudict-0.7b.symbols")
 
-    with open(dict_file, "w") as txt:
-        txt.write("\n".join(mocked_dictionary))
+    with open(dict_file, "w") as fileobj:
+        for section in [header, puncs, words]:
+            for line in section:
+                fileobj.write(line)
+                fileobj.write("\n")
 
     with open(symbol_file, "w") as txt:
         txt.write("\n".join(mocked_symbols))
 
     mocked_data = []
-    for ent in mocked_dictionary:
-        if len(ent) and ent[:3] == ";;;":
-            continue
+
+    if return_punc:
+        for i, ent in enumerate(puncs):
+            _, phones = ent.split("  ")
+            mocked_data.append((punc_outputs[i], phones.split(" ")))
+
+    for ent in words:
         word, phones = ent.split("  ")
-        if word[0] in string.punctuation:
-            # for punctuation, we only keep the puncuation itself
-            word = word[0]
         mocked_data.append((word, phones.split(" ")))
 
     return mocked_data
@@ -56,42 +168,51 @@ def get_mock_dataset(root_dir):
 
 class TestCMUDict(TempDirMixin, TorchaudioTestCase):
     root_dir = None
+    root_punc_dir = None
     samples = []
+    punc_samples = []
 
     @classmethod
     def setUpClass(cls):
-        cls.root_dir = cls.get_base_temp_dir()
+        cls.root_dir = os.path.join(cls.get_base_temp_dir(), "normal")
+        os.mkdir(cls.root_dir)
         cls.samples = get_mock_dataset(cls.root_dir)
+        cls.root_punc_dir = os.path.join(cls.get_base_temp_dir(), "punc")
+        os.mkdir(cls.root_punc_dir)
+        cls.punc_samples = get_mock_dataset(cls.root_punc_dir, return_punc=True)
 
     def _test_cmudict(self, dataset):
         """Test if the dataset is reading the mocked data correctly."""
         n_item = 0
         for i, (word, phones) in enumerate(dataset):
-            if dataset.exclude_punctuations == True:
-                assert word not in string.punctuation
-                # because two punctuations in the beginning are excluded
-                i = i + 2
             expected_word, expected_phones = self.samples[i]
             assert word == expected_word
             assert phones == expected_phones
             n_item += 1
-        if dataset.exclude_punctuations:
-            assert n_item == (len(self.samples) - 2)
-        else:
-            assert n_item == len(self.samples)
+        assert n_item == len(self.samples)
+
+    def _test_punc_cmudict(self, dataset):
+        """Test if the dataset is reading the mocked data with punctuations correctly."""
+        n_item = 0
+        for i, (word, phones) in enumerate(dataset):
+            expected_word, expected_phones = self.punc_samples[i]
+            assert word == expected_word
+            assert phones == expected_phones
+            n_item += 1
+        assert n_item == len(self.punc_samples)
 
     def test_cmuarctic_path_with_punctuation(self):
-        dataset = CMUDict(Path(self.root_dir), exclude_punctuations=False)
-        self._test_cmudict(dataset)
+        dataset = CMUDict(Path(self.root_punc_dir), exclude_punctuations=False)
+        self._test_punc_cmudict(dataset)
 
     def test_cmuarctic_str_with_punctuation(self):
-        dataset = CMUDict(self.root_dir, exclude_punctuations=False)
+        dataset = CMUDict(self.root_punc_dir, exclude_punctuations=False)
+        self._test_punc_cmudict(dataset)
+
+    def test_cmuarctic_path(self):
+        dataset = CMUDict(Path(self.root_punc_dir), exclude_punctuations=True)
         self._test_cmudict(dataset)
 
     def test_cmuarctic_str(self):
-        dataset = CMUDict(self.root_dir)
-        self._test_cmudict(dataset)
-
-    def test_cmuarctic_path(self):
-        dataset = CMUDict(Path(self.root_dir))
+        dataset = CMUDict(self.root_punc_dir, exclude_punctuations=True)
         self._test_cmudict(dataset)
--- a/test/torchaudio_unittest/datasets/cmudict_test.py
+++ b/test/torchaudio_unittest/datasets/cmudict_test.py
@@ -1,0 +1,97 @@
+import os
+from pathlib import Path
+import string
+
+from torchaudio.datasets import CMUDict
+
+from torchaudio_unittest.common_utils import (
+    TempDirMixin,
+    TorchaudioTestCase,
+)
+
+
+def get_mock_dataset(root_dir):
+    """
+    root_dir: directory to the mocked dataset
+    """
+    mocked_dictionary = [
+        ";;; # CMUdict  --  Major Version: 0.07",
+        ";;; ",
+        ";;; # $HeadURL$",
+        "!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T",
+        "\"CLOSE-QUOTE  K L OW1 Z K W OW1 T",
+        "AALIYAH  AA2 L IY1 AA2",
+        "AALSETH  AA1 L S EH0 TH",
+        "AAMODT  AA1 M AH0 T",
+        "AANCOR  AA1 N K AO2 R",
+        "AARDEMA  AA0 R D EH1 M AH0",
+        "AARDVARK  AA1 R D V AA2 R K",
+    ]
+    mocked_symbols = [
+        "AA0", "AA1", "AA2", "AH0", "AO2", "D", "EH0", "EH2", "IY1", "K", "L", "M",
+        "N", "P", "OW1", "OY2", "R", "S", "SH", "T", "TH", "W", "V", "Z",
+    ]
+
+    dict_file = os.path.join(root_dir, "cmudict-0.7b")
+    symbol_file = os.path.join(root_dir, "cmudict-0.7b.symbols")
+
+    with open(dict_file, "w") as txt:
+        txt.write("\n".join(mocked_dictionary))
+
+    with open(symbol_file, "w") as txt:
+        txt.write("\n".join(mocked_symbols))
+
+    mocked_data = []
+    for ent in mocked_dictionary:
+        if len(ent) and ent[:3] == ";;;":
+            continue
+        word, phones = ent.split("  ")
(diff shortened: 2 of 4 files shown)


torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _parse_dictionary(self, lines: List[str]):
+        _alt_re = re.compile(r'\([0-9]+\)')
+        cmudict: List[Tuple[str, List[str]]] = list
+        for line in lines:
+            if len(line) and line[:3] != ";;;":  # ignore comments
+                parts = line.strip().split('  ')
+                # if a word have multiple pronunciations, there will be (number) appended to it
+                # for example, DATAPOINTS(1)
+                word = re.sub(_alt_re, '', parts[0])
```
mthrok, line 71: If I am reading this regex correctly, this removes the number.
I guess the comment, "there will be (number) appended" is referring to the original form, and I found it confusing. Can you, instead, explain what is done (and possibly why)?

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+            tuple: The corresponding word and phonemes ``(word, [phonemes])``.
+
+        """
+        return self._dictionary[n]
+
+    def __len__(self) -> int:
+        return len(self._dictionary)
+
+    def get_symbols(self) -> List[str]:
+        return self._symbols
```
mthrok, line 92: What would happen if I do this?

```python
cmu_dict = CMUDict(root)
cmu_dict.get_symbols().append("foo")
print(cmu_dict.get_symbols())
```

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+                 url_symbols: str = "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b.symbols",
+                 download: bool = False) -> None:
+
+        # Get string representation of 'root' in case Path object is passed
+        root = os.fspath(root)
+        if not os.path.isdir(root):
+            os.mkdir(root)
+
+        if download:
+            if not os.path.isdir(root):
```
mthrok, line 45: Not sure the logic here is correct.

In L41 and L42, `root` directory is created if it is not a directory, and in L45 it will perform download if `root` is not a directory.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+    """
+
+    def __init__(self,
+                 root: Union[str, Path],
+                 url: str = "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b",
+                 url_symbols: str = "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b.symbols",
+                 download: bool = False) -> None:
+
+        # Get string representation of 'root' in case Path object is passed
+        root = os.fspath(root)
```
mthrok, line 40: I think it is cleaner if we convert `str` to `Path` object.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+
+_CHECKSUMS = {
+    "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b":
+    "825f4ebd9183f2417df9f067a9cabe86",
+    "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b.symbols":
+    "385e490aabc71b48e772118e3d02923e",
+}
+
+
+class CMUDICT(Dataset):
```
mthrok, line 18: Can we call this `CMUDict`? I do not see why this has to be all capital. This is not a convention or a good practice to follow either.

torchaudio/datasets/cmudict.py
mthrok: In Tedlium dataset, it uses `property` and `copy` to avoid in-place modification.

https://github.com/pytorch/audio/blob/d1d6dbc64a1dd7506ce6b2ff2f9969ae13a6b52d/torchaudio/datasets/tedlium.py#L180-L192

torchaudio/datasets/cmudict.py
yangarbiter: I actually like CMUDict better.
But I saw that many of the previous dataset names like [CommonVoice is called COMMONVOICE](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/commonvoice.py#L32), [LJSpeech is called LJSPEECH](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/ljspeech.py#L20) and so on.

mthrok: Yeah, let's call it `CMUDict`. That's more natural.

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    mocked_data = []
+    for ent in mocked_dictionary:
+        mocked_data.append((ent.split("  ")[0], [ent.split("  ")[1].split(" ")]))
+
+    return mocked_data
+
+
+class TestCMUARCTIC(TempDirMixin, TorchaudioTestCase):
+    backend = "default"
```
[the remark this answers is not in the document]
yangarbiter: Thanks for the feedback, I've removed it [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R56). I've also corrected the Test name to `TestCMUDict`.

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,72 @@
...
+
+    with open(dict_file, "w") as txt:
+        txt.write("\n".join(mocked_dictionary))
+
+    with open(symbol_file, "w") as txt:
+        txt.write("\n".join(mocked_symbols))
+
+    mocked_data = []
+    for ent in mocked_dictionary:
+        mocked_data.append((ent.split("  ")[0], [ent.split("  ")[1].split(" ")]))
```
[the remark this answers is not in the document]
yangarbiter: Thanks for the feedback, I've fixed this [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R48).

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,72 @@
...
+    mocked_symbols = [
+        "AA0", "AA1", "AA2", "AH0", "AO2", "D", "EH0", "IY1", "K", "L", "M",
+        "N", "R", "S", "T", "TH", "V"
+    ]
+
+    dict_file = os.path.join(root_dir, "cmudict-0.7b")
+    symbol_file = os.path.join(root_dir, "cmudict-0.7b.symbols")
+
+    with open(dict_file, "w") as txt:
+        txt.write("\n".join(mocked_dictionary))
```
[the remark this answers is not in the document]
yangarbiter: I've added the comments as well as two additional punctuations [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R18)

yangarbiter:
> from offline discussion
> 
> > How should we deal with the punctuation?
> 
> Add an option to include/exclude and default to exclude them as the primal use case is for TTS, in which punctuations are not pronounced independently.

I've added the `exclude_punctuations` parameter [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R38).

torchaudio/datasets/cmudict.py
yangarbiter: Thanks for the suggestion, I've covert `root` [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R42).

torchaudio/datasets/cmudict.py
yangarbiter: Good point! Now it will call the copy method before returning ([here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R102)).

torchaudio/datasets/cmudict.py
yangarbiter: A more detailed explanation is [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R78). Please let me know if it is clear.
Thanks.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+        with open(os.path.join(self._root_path, basename), "r") as text:
+            self._dictionary = self._parse_dictionary(text.readlines())
+
+        self._available_words = list(self._dictionary.keys())
+
+    def _parse_dictionary(self, lines: List[str]):
+        _alt_re = re.compile(r'\([0-9]+\)')
+        cmudict: List[Tuple[str, List[str]]] = list
+        for line in lines:
+            if len(line) and line[:3] != ";;;":  # ignore comments
```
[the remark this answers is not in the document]
yangarbiter: Thanks! It is much clearer [this way](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R67).

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,92 @@
...
+                parts = line.strip().split('  ')
```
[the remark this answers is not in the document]
yangarbiter: I've changed this [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R70) and instead, I call them word and phones to be consistent.
Thanks for the suggestion!

docs/source/datasets.rst
```diff
@@ -33,6 +33,14 @@ CMUARCTIC
   :special-members: __getitem__
 
 
+CMUDICT
```
mthrok, line 36:
```suggestion
CMUDict
```

docs/source/datasets.rst
```diff
@@ -33,6 +33,14 @@ CMUARCTIC
...
+~~~~~~~~~
+
+.. autoclass:: CMUDICT
```
mthrok, line 39:
```suggestion
.. autoclass:: CMUDict
```

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,105 @@
...
+        Returns:
+            tuple: The corresponding word and phonemes ``(word, [phonemes])``.
+
+        """
+        return self._dictionary[n]
+
+    def __len__(self) -> int:
+        return len(self._dictionary)
+
+    def get_symbols(self) -> List[str]:
```
mthrok, line 104: 1. Is `List[str]` the right data structure? I am getting a feeling that `set` is more appropriate, and if that's the case, we can use `frozenset` so that we do not need to copy.
2. Can we use `@property` access? If the underlying symbols are frozen, it is more Pythonic to access it as a property, rather than getter/setter approach.

mthrok: For 1, alternative way to think is; do we need to iterate the symbols?

yangarbiter: Currently in the dataset, we don't need to iterate the symbols. But in the normal usage, we will need to translate each character into a number, that is where we need to iterate the symbols.

The symbol here is very similar the the symbol [here](https://github.com/pytorch/audio/blob/master/examples/pipeline_tacotron2/text/text_preprocessing.py#L65). And we will need to map each symbol (character) to a number like [here](https://github.com/pytorch/audio/blob/master/examples/pipeline_tacotron2/text/text_preprocessing.py#L85) and [here](https://github.com/pytorch/audio/blob/master/examples/pipeline_tacotron2/text/text_preprocessing.py#L85). So we actually need their order.

mthrok: So in client code, the ordering is used for indexing, but the question is, the ordering in client code has to be natively originated from the `CMUDict` class?

For example, I see a mapping "ABE -> EY1 B" in the dictionary. But the class `CMUDict` does not define/care to which index `EY1` or `B` should be mapped. `CMUDict` only defines G2P mapping and the set of phoneme symbols. It appears to me that how to index the symbols is user's choice. If we use `List[str]`, then we are giving that this is the order that the symbols should be used.

yangarbiter: No, it does not need to be natively originated from the `CMUDict` class. The users can always turn the set into a list and sort it in the way they like. 

And yes, I agree that using `List[str]` is similar to giving the user instruction on the order of the symbols (another way of looking at it is that we are preserving the order information from the original dataset). But is there a possible downside of having a default ordering of the symbols?

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,97 @@
...
+
+    with open(symbol_file, "w") as txt:
+        txt.write("\n".join(mocked_symbols))
+
+    mocked_data = []
+    for ent in mocked_dictionary:
+        if len(ent) and ent[:3] == ";;;":
```
mthrok, line 46: In tests, generally speaking, it is better to avoid the same logic implemented in the library, and keep the test logic as simple as possible. (Note: we use tests to debug the library, but finding a logical error in the test code is much harder when the test simply reports "pass".)

1. Splitting the `mocked_dictionary` into header, punctuations, and non-punctuations, the preparation logic becomes cleaner.
2. It is ideal to avoid parsing the returned `mocked_data` for exclusion of punctuation.

So I suggest to change the helper function to support exclusion of punctuations, making it the responsibility of helper function to return the proper mock data.

```python
def get_mock_dataset(root_dir, return_punc=False):
    header = [
        ";;; # CMUdict  --  Major Version: 0.07",
        ...
    ]

    puncs = [
        "!EXCLAMATION-POINT EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T",
        ...
    ]

    words = [
        "AALIYAH  AA2 L IY1 AA2",
        ...
    ]
    ...
    with open(dict_file, "w") as fileobj:
        for section in [header, punch, words]:
            for line in section:
                fileobj.write(line)
                fileobj.write("\n")
    ...

    data_to_include = puncs + words if return_punc else words
    mocked_data = []
    for ent in data_to_include:
        word, phones = ent.split("  ")
        mocked_data.append((word, phones.split(" ")))
    return mocked_data
```

then in the test class, we can generate two mock datasets, one with punctuation and one without punctuations, and use the different `root_dir` value for different test.
(also, the use of `if dataset.exclude_punctuations` should be avoided in test logic to keep it simple.)

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def _parse_dictionary(self, lines: List[str]):
+        _alt_re = re.compile(r'\([0-9]+\)')
+        cmudict: List[Tuple[str, List[str]]] = list()
+        for line in lines:
+            if not line or line.startswith(';;;'):  # ignore comments
+                continue
+
+            word, phones = line.strip().split('  ')
+            if word[0] in string.punctuation:
```
mthrok, line 74: The content of `string.punctuation` is as follow;

```
$ python -c 'import string;print(string.punctuation)'
!"#$%&'()*+,-./:;<=>?@[\]^_`{|}~
```

while the `word[0]` can take the following values.
The implementation looks correct. I feel safer if the mock dataset include at least one line example par the following unique first character.

```
$ grep -v ';;;' cmudict-0.7b | LC_CTYPE=C LANG=C cut -c-1 | sort -u
!
"
#
%
&
'
(
)
+
,
-
.
/
3
:
;
?
A
B
C
D
E
F
G
H
I
J
K
L
M
N
O
P
Q
R
S
T
U
V
W
X
Y
Z
{
}
```

mthrok: BTW `string.punctuation` is of `str` type. When performing `in` operation a lot, it is better to covert it to `set`.

mthrok:
```
for char in `grep -v ';;;' cmudict-0.7b | LC_CTYPE=C LANG=C cut -c-1 | sort -u`; grep -m1 "^${char}" cmudict-0.7b;
!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T
"CLOSE-QUOTE  K L OW1 Z K W OW1 T
#HASH-MARK  HH AE1 M AA2 R K
%PERCENT  P ER0 S EH1 N T
&AMPERSAND  AE1 M P ER0 S AE2 N D
'ALLO  AA2 L OW1
(BEGIN-PARENS  B IH0 G IH1 N P ER0 EH1 N Z
)CLOSE-PAREN  K L OW1 Z P ER0 EH1 N
+PLUS  P L UH1 S
,COMMA  K AA1 M AH0
--DASH  D AE1 SH
;;; # CMUdict  --  Major Version: 0.07
/SLASH  S L AE1 SH
3-D  TH R IY1 D IY2
:COLON  K OW1 L AH0 N
;;; # CMUdict  --  Major Version: 0.07
?QUESTION-MARK  K W EH1 S CH AH0 N M AA1 R K
A  AH0
B  B IY1
C  S IY1
D  D IY1
E  IY1
F  EH1 F
G  JH IY1
H  EY1 CH
I  AY1
J  JH EY1
K  K EY1
L  EH1 L
M  EH1 M
N  EH1 N
O  OW1
P  P IY1
Q  K Y UW1
R  AA1 R
S  EH1 S
T  T IY1
U  Y UW1
V  V IY1
W  D AH1 B AH0 L Y UW0
X  EH1 K S
Y  W AY1
Z  Z IY1
{BRACE  B R EY1 S
}CLOSE-BRACE  K L OW1 Z B R EY1 S
```

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,170 @@
...
+        cmudict: List[Tuple[str, List[str]]] = list()
+        for line in lines:
+            if not line or line.startswith(';;;'):  # ignore comments
+                continue
+
+            word, phones = line.strip().split('  ')
+            if word in _PUNCTUATIONS:
+                if self.exclude_punctuations:
+                    continue
+                else:
```
mthrok, line 134: I think `else` and the resulting indentation is redundant here.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,170 @@
...
+                    # !EXCLAMATION-POINT -> !
+                    # --DASH -> --
+                    # ...ELLIPSIS -> ...
+                    if word[:3] == "...":  # everything ing  _PUNCTUATIONS has length > 3
```
mthrok, line 138: `l[:3]` means "up to the third element", so no need to worry about if word is always longer than 3.
Rather, I think `if word.startswith("...")` is more readable.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,170 @@
...
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import download_url
+
+_CHECKSUMS = {
+    "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b":
+    "825f4ebd9183f2417df9f067a9cabe86",
+    "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b.symbols":
+    "385e490aabc71b48e772118e3d02923e",
+}
+_PUNCTUATIONS = set([
```
mthrok, line 15: Nice use of `set`. 👍

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,213 @@
...
+        "+PLUS  P L UH1 S",
+        ",COMMA  K AA1 M AH0",
+        "--DASH  D AE1 SH",
+        "!EXCLAMATION-POINT  EH2 K S K L AH0 M EY1 SH AH0 N P OY2 N T",
+        "/SLASH  S L AE1 SH",
+        ":COLON  K OW1 L AH0 N",
+        ";SEMI-COLON  S EH1 M IY0 K OW1 L AH0 N",
+        "?QUESTION-MARK  K W EH1 S CH AH0 N M AA1 R K",
+        "{BRACE  B R EY1 S",
+        "}CLOSE-BRACE  K L OW1 Z B R EY1 S",
```
mthrok, line 40: Can you add ellipses case? I think it was missing from my comment, as I assumed that all the punctuations are one character long.

test/torchaudio_unittest/datasets/cmudict_test.py
```diff
@@ -0,0 +1,213 @@
...
+        "!",
+        "/",
+        ":",
+        ";",
+        "?",
+        "{",
+        "}",
+    ]
+
+    words = [
```
mthrok, line 64: Can you add the special cases you found? `'BOUT` and such.

test/torchaudio_unittest/datasets/cmudict_test.py
yangarbiter: Thanks for pointing this out, I've added it [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R41).

test/torchaudio_unittest/datasets/cmudict_test.py
yangarbiter: Sure, I've added three of these examples [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R68-R70).

torchaudio/datasets/cmudict.py
yangarbiter: Thanks for pointing this out! I've removed it [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R134)

torchaudio/datasets/cmudict.py
yangarbiter: Agree! This is way better then what I've written. I've fixed them [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R137-R139)

torchaudio/datasets/cmudict.py
yangarbiter: Now I've changed to using a [explicit list of punctuation](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R15) to filter out the punctuations. Could you also help me check if I missed any of the punctuations?
Thanks!

test/torchaudio_unittest/datasets/cmudict_test.py
yangarbiter: Thanks for the suggestion! I've rewritten the testing part with this logic [here](https://github.com/pytorch/audio/pull/1627/files#diff-017895ceaff7f853256bf11da184c3d6873e5322b9ee3c0051208639aecaa3d2R12).

torchaudio/datasets/cmudict.py
yangarbiter: Thanks for pointing it out. I've fixed the logics [here](https://github.com/pytorch/audio/pull/1627/files#diff-50ee900e0897ac7528dccd9bdd36d5f0c7cb2d45310aefb46a3bf281f73d88c2R99-R111). It now creates a directory if root is not a directory and only download the data when the directory is there.

torchaudio/datasets/cmudict.py
```diff
@@ -0,0 +1,170 @@
...
+        basename = os.path.basename(url)
+        basename_symbols = os.path.basename(url_symbols)
+
+        with open(os.path.join(self._root_path, basename_symbols), "r") as text:
+            self._symbols = [line.strip() for line in text.readlines()]
+
+        with open(os.path.join(self._root_path, basename), "r") as text:
+            self._dictionary = self._parse_dictionary(text.readlines())
+
+    def _parse_dictionary(self, lines: List[str]):
```
mthrok, line 123: nit: since this function uses `self` only once, can we move this to a plain function?
Also, 

```python

def _parse_fidtionary(src: Iterable[str], exclude_punctuations: bool) -> List[str]:
    ...
```

mthrok (approved):
LGTM

torchaudio/datasets/cmudict.py
yangarbiter: Thanks for the suggestion, this is fixed in [this commit](https://github.com/pytorch/audio/pull/1627/commits/4fa648fe192a9dd178e427d6730cebadb9d8db4f).

PR #1629, opened 2021-07-14
Add PitchShift to functional and transform

nateanl:
Refers to #438


torchaudio/functional/functional.py
```diff
@@ -1488,3 +1489,75 @@ def edit_distance(seq1: Sequence, seq2: Sequence) -> int:
...
+    """
+    Shift the pitch of a waveform by ``n_steps`` steps.
+
+    Args:
+        waveform (Tensor): The input waveform of shape `(..., time)`.
+        sample_rate (float): Sample rate of `waveform`.
+        n_steps (int): The (fractional) steps to shift ``waveform``.
+        bins_per_octave (int): The number of steps per octave.
+        n_fft (int): Size of FFT, creates ``n_fft // 2 + 1`` bins
+        hop_length (int): Length of hop between STFT windows. (
```
carolineechen, line 1513: reverse the ordering of `hop_length` and `win_length` to match arg list ordering

torchaudio/functional/functional.py
```diff
@@ -1488,3 +1489,75 @@ def edit_distance(seq1: Sequence, seq2: Sequence) -> int:
...
+            Default: ``win_length // 4``)
+        win_length (int): Window size. (Default: ``n_fft``)
```
carolineechen, line 1515: nit: `(int or None, optional)` for Optional typed parameters

carolineechen: something like `Window size. If None, then ``n_fft`` is used. (Default: ``None``)` is more accurate for these docstrings

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -422,6 +422,16 @@ def test_simple_case_edit_distance(self, seq1, seq2, distance):
         assert F.edit_distance(seq1, seq2) == distance
         assert F.edit_distance(seq2, seq1) == distance
 
+    @nested_params(
+        [-4, -2, 0, 2, 4],
+    )
+    def test_pitch_shift_shape(self, n_steps):
```
carolineechen, line 428: It would be good to have a unit test that tests for accuracy in addition to just the shape -- is there a reason you removed the reversible test from previous commits? 

Other potential ideas include
- generating and using a simple example as a baseline to check for correctness
- comparing against librosa's implementation if relevant (see `librosa_compatibility_test_impl.py`), but I am not sure if this would pass since intermediate functions (like resample) may differ and not provide close enough results

nateanl: I tried to increase 4 and decrease 4 steps of pitch and listened to the output in my jupyter notebook. Apparently there is  signal distortion after two pitch_shift operations (for both librosa and torchaudio). Thus I removed the reversible test.

Not sure if it will achieve identical output as librosa, phase_vocoder does, but stft and resample are not certain.

torchaudio/functional/functional.py
```diff
@@ -1488,3 +1489,76 @@ def edit_distance(seq1: Sequence, seq2: Sequence) -> int:
...
+    window: Optional[Tensor] = None,
+) -> Tensor:
+    """
+    Shift the pitch of a waveform by ``n_steps`` steps.
+
+    Args:
+        waveform (Tensor): The input waveform of shape `(..., time)`.
+        sample_rate (float): Sample rate of `waveform`.
+        n_steps (int): The (fractional) steps to shift ``waveform``.
+        bins_per_octave (int, optional): The number of steps per octave.
```
carolineechen, line 1511: can you also add `(Default: ___)` for the other parameters with default values?

torchaudio/functional/functional.py
```diff
@@ -1488,3 +1489,76 @@ def edit_distance(seq1: Sequence, seq2: Sequence) -> int:
...
+    window: Optional[Tensor] = None,
+) -> Tensor:
+    """
+    Shift the pitch of a waveform by ``n_steps`` steps.
+
+    Args:
+        waveform (Tensor): The input waveform of shape `(..., time)`.
+        sample_rate (float): Sample rate of `waveform`.
+        n_steps (int): The (fractional) steps to shift `waveform`.
+        bins_per_octave (int, optional): The number of steps per octave (Default: ``12``).
```
mthrok, line 1511: Help me understand this, so at the end, the pitch is `n_steps/bins_per_octave` octave higher (or lower if `n_steps` is negative)?

test/torchaudio_unittest/functional/functional_impl.py
mthrok: > Apparently there is signal distortion after two pitch_shift operations (for both librosa and torchaudio). Thus I removed the reversible test.

Curious. Was the signal distorted everywhere or was it boundary effect?

> It would be good to have a unit test that tests for accuracy in addition to just the shape

For accuracy, one way I can think of is to use sine wave (with `get_sinusoid`) and check if the resulting signal is similar to a sine wave of modified frequency.

mthrok (approved):
The implementation looks good. Can you paste the screen shot of before and after just in case?
The accuracy test is something we want to have and continue improving. But it might be difficult to achieve and we would like to be able to come back and refer that the initial implementation works.

torchaudio/functional/functional.py
nateanl: Correct. positive ``n_steps`` makes the pitch higher.

nateanl:
I tried to apply librosa's pitch_shift to a sine wave array and reverse it. Here are the code and plots:
```
%matplotlib inline
import librosa
import numpy as np
t = np.arange(0.0, 20, 0.005)
sig = np.sin(2*np.pi*t)
sr = 8000
sig_shift = librosa.effects.pitch_shift(sig, sr,n_steps=4)
sig_reverse = librosa.effects.pitch_shift(sig_shift,sr, n_steps=-4)

fig, (ax1, ax2, ax3) = plt.subplots(3, 1, figsize=(14,10))
ax1.plot(sig)
ax2.plot(sig_shift)
ax3.plot(sig_reverse)
fig.tight_layout()
```
image

According to the plot, librosa itself can't reverse the signal.

carolineechen:
> I tried to apply librosa's pitch_shift to a sine wave array and reverse it. Here are the code and plots:

Could you also show the results of this using the torchaudio implementation from this PR?

nateanl:
Sure. Here is the plot by using torchaudio. The difference is that ``n_fft`` in librosa is set to 2048 by default. To get identical plots, we need to manually set the parameters here. 
```
%matplotlib inline
import librosa
import torch
import numpy as np
t = np.arange(0.0, 20, 0.005)
sig = np.sin(2*np.pi*t)
sr = 8000
sig_shift = pitch_shift(torch.tensor(sig), sr,n_steps=4, n_fft=2048)
sig_reverse = pitch_shift(sig_shift,sr, n_steps=-4, n_fft=2048)

fig, (ax1, ax2, ax3) = plt.subplots(3, 1, figsize=(14,10))
ax1.plot(sig)
ax2.plot(sig_shift)
ax3.plot(sig_reverse)
fig.tight_layout()
```
image

carolineechen (approved):
Thanks for the plots, LGTM!

PR #1631, opened 2021-07-16
Remove torchaudio._internal.fft module

hwangjeff:
`torchaudio._internal.fft` was originally added to account for the introduction of module `torch.fft`, when `torch.fft` could refer to either a module or function. now that `torch.fft` refers unambiguously to a module, we remove `torchaudio._internal.fft` here.


mthrok (approved):
Looks good. Thanks!

PR #1635, opened 2021-07-20
Update and move convention section to CONTRIBUTING.md

yangarbiter:
- Move convention section from README.md to CONTRIBUTING.md
- Changed `n_mel` to `n_mels` since all related variables in [functional](https://github.com/pytorch/audio/blob/master/torchaudio/functional/functional.py) and (transforms)[https://github.com/pytorch/audio/blob/master/torchaudio/transforms.py] are called `n_mels`.
- For `n_freq`, there are two exceptions ([1](https://github.com/pytorch/audio/blob/f5dbb0021b7c6b4681c55e9ebe9b601faa1fd4da/torchaudio/functional/functional.py#L380) and [2](https://github.com/pytorch/audio/blob/9a0e70ea6699e311c8f4e0c7a52d7042ce16b8b9/test/torchaudio_unittest/functional/librosa_compatibility_test_impl.py#L99)) calling it `n_freqs`.
- `min_freq` and `max_freq` are changed to `f_min` and `f_max` are used [here](https://github.com/pytorch/audio/blob/f5dbb0021b7c6b4681c55e9ebe9b601faa1fd4da/torchaudio/transforms.py) and no other function uses `min_freq` or `max_freq`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -129,6 +129,39 @@ make html
 
 The built docs should now be available in `docs/build/html`
 
+## Conventions
+
+As a good software development practice, we try to stick to existing variable
+names and shape (for tensors).
+The following are some of the conventions that we follow.
+
+- We use an ellipsis "..." as a placeholder for the rest of the dimensions of a
+  tensor, e.g. optional batching and channel dimensions. If batching, the
+  "batch" dimension should come in the first diemension.
+- Tensors are assumed to have "channel" dimension coming before the "time"
+  dimension. The bins in frequency domain (freq and mel) are assumed to come
+  before the "time" dimension but after the "channel" dimension. These
+  ordering makes the tensors consistent with PyTorch's dimensions.
+- For size names, the prefix `n_` is used (e.g. "a tensor of size (`n_freq`,
+  `n_mels`)") whereas dimension names do not have this prefix (e.g. "a tensor of
+  dimension (channel, time)")
+
+Here are some of the examples of commonly used variables with thier names,
+meanings, and shapes (or units):
+
+* `waveform`: a tensor of audio samples with dimensions (..., channel, time)
+* `sample_rate`: the rate of audio dimensions (samples per second)
+* `specgram`: a tensor of spectrogram with dimensions (..., channel, freq, time)
+* `mel_specgram`: a mel spectrogram with dimensions (..., channel, mel, time)
+* `hop_length`: the number of samples between the starts of consecutive frames
+* `n_fft`: the number of Fourier bins
+* `n_mels`, `n_mfcc`: the number of mel and MFCC bins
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -129,6 +129,44 @@ make html
...
+* `hop_length`: the number of samples between the starts of consecutive frames
+* `n_fft`: the number of Fourier bins
+* `n_mels`, `n_mfcc`: the number of mel and MFCC bins
+* `n_freq`: the number of bins in a linear spectrogram
+* `f_min`: the lowest frequency of the lowest band in a spectrogram
+* `f_max`: the highest frequency of the highest band in a spectrogram
+* `win_length`: the length of the STFT window
+* `window_fn`: for functions that creates windows e.g. `torch.hann_window`
+
+Transforms expect and return the following dimensions.
```
mthrok, line 155: Let's remove the transforms. This is not a convention. The same thing is explained in the doc, and there should be only one source of truth. Also this is not necessarily relevant to new contributors.

CONTRIBUTING.md
```diff
@@ -129,6 +129,44 @@ make html
...
+* `specgram`: a tensor of spectrogram with dimensions (channel, freq, time)
+* `mel_specgram`: a mel spectrogram with dimensions (channel, mel, time)
+* `hop_length`: the number of samples between the starts of consecutive frames
+* `n_fft`: the number of Fourier bins
+* `n_mels`, `n_mfcc`: the number of mel and MFCC bins
+* `n_freq`: the number of bins in a linear spectrogram
+* `f_min`: the lowest frequency of the lowest band in a spectrogram
+* `f_max`: the highest frequency of the highest band in a spectrogram
+* `win_length`: the length of the STFT window
+* `window_fn`: for functions that creates windows e.g. `torch.hann_window`
```
mthrok, line 153: Looking at the code, the explanation, and your comment I think only `waveform`, and `sample_rate` is standard and relevant across the code base. If there are only one or two usage, then it's not really convention or standard. And when a new contributor add a new feature, and when we review it, as a good software development practice, we naturally try to stick to existing variable names. So I doubt we need this.
Rather, this list of convention is rather confusing and limiting because IIRC there are cases where `waveform` have batch dimension in addition to channel and time, which seems like not following the convention.
How about removing this?

CONTRIBUTING.md
yangarbiter: I agree with this. If I need to know the input or output, I would just go to the documentation website instead of here.

CONTRIBUTING.md
yangarbiter: I actually think we should keep it. I think this short table can help for a quick reference. I've updated the batch dimension problem and rewrite this paragraph a bit.
Please let me know how you think about this new version.
Thanks.

mthrok (approved):

PR #1571, opened 2021-06-11
[BC-Breaking] Remove lazy behavior from MelScale

Patil2099:
Fixes #1569
Closes #851

Issue #1569, opened 2021-06-11
Remove lazy behavior from MelScale


mthrok:
Address https://github.com/pytorch/audio/issues/1510

In 0.9.0 we are deprecating the lazy behavior of MelScale because it can make an invalid TorchScript object and it does not align with the design of torchaudio. Now in master branch, we can remove the implementation.

See [CONTRIBUTING.md](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md) for setting up the development environment.

## Steps
1.  Replace [`n_stft: Optional[int] = None`](https://github.com/pytorch/audio/blob/6f9c14bec787d952c2aa1b6a13d11509c7a38bed/torchaudio/transforms.py#L258) with `n_stft: int = 201`. 
2. Remove [the branches](https://github.com/pytorch/audio/blob/6f9c14bec787d952c2aa1b6a13d11509c7a38bed/torchaudio/transforms.py#L271-L282) for `n_stft = None`
3. Remove [`__prepare_script__`](https://github.com/pytorch/audio/blob/6f9c14bec787d952c2aa1b6a13d11509c7a38bed/torchaudio/transforms.py#L285-L297) method
4. Remove [the case `numel() == 0`](https://github.com/pytorch/audio/blob/6f9c14bec787d952c2aa1b6a13d11509c7a38bed/torchaudio/transforms.py#L312-L319)
5. Remove [test](https://github.com/pytorch/audio/blob/6f9c14bec787d952c2aa1b6a13d11509c7a38bed/test/torchaudio_unittest/transforms/torchscript_consistency_impl.py#L62-L65)

## Tests

Run test `cd test && pytest torchaudio_unittest.transforms -vk MelScale` and make changes to failures.

Patil2099:
I would love to work on this @mthrok

mthrok:
Resolved by #1636

PR #1636, opened 2021-07-21
Remove lazy behavior from MelScale

hwangjeff:
Rebases #1571; addresses #1569:

"In 0.9.0 we are deprecating the lazy behavior of MelScale because it can make an invalid TorchScript object and it does not align with the design of torchaudio. Now in master branch, we can remove the implementation."

### DIFF
--- a/test/torchaudio_unittest/transforms/batch_consistency_test.py
+++ b/test/torchaudio_unittest/transforms/batch_consistency_test.py
@@ -33,13 +33,14 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         self.assertEqual(computed, expected)
 
     def test_batch_MelScale(self):
-        specgram = torch.randn(2, 31, 2786)
+        n_stft = 31
+        specgram = torch.randn(2, n_stft, 2786)
 
         # Single then transform then batch
-        expected = torchaudio.transforms.MelScale()(specgram).repeat(3, 1, 1, 1)
+        expected = torchaudio.transforms.MelScale(n_stft=n_stft)(specgram).repeat(3, 1, 1, 1)
 
         # Batch then transform
-        computed = torchaudio.transforms.MelScale()(specgram.repeat(3, 1, 1, 1))
+        computed = torchaudio.transforms.MelScale(n_stft=n_stft)(specgram.repeat(3, 1, 1, 1))
 
         # shape = (3, 2, 201, 1394)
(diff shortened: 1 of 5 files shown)


test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -33,13 +33,14 @@ def test_batch_Resample(self):
         self.assertEqual(computed, expected)
 
     def test_batch_MelScale(self):
-        specgram = torch.randn(2, 31, 2786)
+        n_stft = 31
+        specgram = torch.randn(2, n_stft, 2786)
```
mthrok, line 37: Can you use the default value of `n_stft=201`, so that we can catch an error for default value first if that happens in the future?
Also not sure where the 2786 is coming from, but I think we can use something shorter, like 256.

test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -33,13 +33,14 @@ def test_batch_Resample(self):
...
 
         # Single then transform then batch
-        expected = torchaudio.transforms.MelScale()(specgram).repeat(3, 1, 1, 1)
+        expected = torchaudio.transforms.MelScale(n_stft=n_stft)(specgram).repeat(3, 1, 1, 1)
```
mthrok, line 40: Note: (follow-up, not the scope of this PR) For batch consistency test, use of repeated does not catch some edge case. https://github.com/pytorch/audio/pull/1315
We should update the logic at some point.

test/torchaudio_unittest/transforms/transforms_test.py
```diff
@@ -56,7 +56,7 @@ def test_AmplitudeToDB(self):
 
     def test_melscale_load_save(self):
         specgram = torch.ones(1, 1000, 100)
-        melscale_transform = transforms.MelScale()
+        melscale_transform = transforms.MelScale(n_stft=1000)
```
mthrok, line 59: Instead of setting `n_stft`, can we change the shape of `specgram` to `1, 201, 100`?
This is more of a personal preference but `1000` is somehow conspicuous.

mthrok (approved):
Looks good. Left minor comments on test improvements.

I know these are not directly related to the PR, but otherwise we wouldn't be improving the tests.

PR #1637, opened 2021-07-21
Add an inference example for wavernn

yangarbiter:
This example takes in an example from LJSPEECH and transform it into a mel spectrogram. The inference function of WaveRNN should be able to transform the mel spectrogram back to it's original waveform.

The implementation if ported from [here](https://github.com/fatchord/WaveRNN/blob/master/models/fatchord_version.py#L169).

@mthrok we may want to discuss whether to put [the inference function](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R181) into the core library since in the current form, it is not torchscriptable.

### DIFF
--- a/examples/pipeline_wavernn/wavernn_inference_wrapper.py
+++ b/examples/pipeline_wavernn/wavernn_inference_wrapper.py
@@ -48,8 +48,8 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
         Where each h is a vector of conditioning channels
         Eg: timesteps=2, overlap=1 with x.size(1)=10
         folded = [[h1, h2, h3, h4],
-                [h4, h5, h6, h7],
-                [h7, h8, h9, h10]]
+                  [h4, h5, h6, h7],
+                  [h7, h8, h9, h10]]
 
         Args:
             x (tensor): Upsampled conditioning channels with shape (1, timesteps, channel).
@@ -79,7 +79,7 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
         for i in range(n_folds):
             start = i * (timesteps + overlap)
             end = start + timesteps + 2 * overlap
-            folded[i] = x[:, :, start:end]
+            folded[i] = x[0, :, start:end]
 
         return folded
 
@@ -160,8 +160,14 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
                             f"Valid choices are 'both', 'before' and 'after'.")
         return padded
 
-    def infer(self, specgram: Tensor, loss_name: str = "crossentropy", mulaw: str = True,
-            batched: bool = True, timesteps: int = 11000, overlap: int = 550) -> Tensor:
+    @torch.jit.export
+    def infer(self,
+              specgram: Tensor,
+              loss_name: str = "crossentropy",
+              mulaw: bool = True,
+              batched: bool = True,
+              timesteps: int = 11000,
+              overlap: int = 550) -> Tensor:
         r"""Inference function for WaveRNN.
 
         Based on the implementation from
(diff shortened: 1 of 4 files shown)


mthrok:
The summary of offline discussion
---

> @mthrok we may want to discuss whether to put [the inference function](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R181) into the core library

Extract the model data handling part from the current `infer` function to see if we can make the proposed `infer` method simple and minimalistic. (Things like mulaw and batch folding should be handled outside of model method)

yangarbiter:
> The summary of offline discussion
> 
> > @mthrok we may want to discuss whether to put [the inference function](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R181) into the core library
> 
> Extract the model data handling part from the current `infer` function to see if we can make the proposed `infer` method simple and minimalistic. (Things like mulaw and batch folding should be handled outside of model method)

If appears that the inference function is not yet jitable due to [`torch.distributions.Categorical`](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R248) being unjitable (https://github.com/pytorch/pytorch/issues/18094)

So for future improvement, we want to move 
1. mulaw decoding,
2. batch folding, and
3. categorical sampling
outside the inference method

mthrok:
> If appears that the inference function is not yet jitable due to [`torch.distributions.Categorical`](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R248) being unjitable ([pytorch/pytorch#18094](https://github.com/pytorch/pytorch/issues/18094))
> 

The document says `torch.distributions.Categorical` is equivalent to `torch.multinomial`. Can you try that?

yangarbiter:
> > If appears that the inference function is not yet jitable due to [`torch.distributions.Categorical`](https://github.com/pytorch/audio/pull/1637/files#diff-2f1a2a7054ae6468415baac0a5a6c4d7e40a36350942d65c8ba0ba3b469c47a4R248) being unjitable ([pytorch/pytorch#18094](https://github.com/pytorch/pytorch/issues/18094))
> 
> The document says `torch.distributions.Categorical` is equivalent to `torch.multinomial`. Can you try that?

Thanks a lot for the suggestion! It works!

examples/pipeline_wavernn/wavernn_inference_wrapper.py
```diff
@@ -0,0 +1,259 @@
...
+            padded[:, :, pad:pad + t] = x
+        elif side == 'after':
+            padded[:, :, :t] = x
+        else:
+            raise ValueError(f"Unexpected side: '{side}'. "
+                            f"Valid choices are 'both', 'before' and 'after'.")
+        return padded
+
+    @torch.jit.export
+    def infer(self,
```
mthrok, line 164: I think you can call this method `forward` in the wrapper, and no need to decorate `torch.jit.export`.

mthrok (approved):
LGTM. The next step will be to move the core inference function to WaveRNN model.

examples/pipeline_wavernn/wavernn_inference_wrapper.py
yangarbiter: Addressed in #1650

PR #1625, opened 2021-07-10
Add Tacotron2 loss function

yangarbiter:
The main Tacotron2 model is in #1621. Since that PR is already too large, I'll put the Tacotron2 loss and it's tests in the example directory temporary. We should discuss more on where to put this loss function (likely the same file as Tacotron2's model).

PR #1639, opened 2021-07-22
Add text preprocessing utilities for TTS pipeline

yangarbiter:
Some of the preprocessing functions from [here](https://github.com/keithito/tacotron).

### DIFF
--- a/examples/pipeline_tacotron2/text/numbers.py
+++ b/examples/pipeline_tacotron2/text/numbers.py
@@ -1,0 +1,71 @@
+"""Modified from https://github.com/keithito/tacotron"""
+
+import inflect
+import re
+
+
+_inflect = inflect.engine()
+_comma_number_re = re.compile(r'([0-9][0-9\,]+[0-9])')
+_decimal_number_re = re.compile(r'([0-9]+\.[0-9]+)')
+_pounds_re = re.compile(r'£([0-9\,]*[0-9]+)')
+_dollars_re = re.compile(r'\$([0-9\.\,]*[0-9]+)')
+_ordinal_re = re.compile(r'[0-9]+(st|nd|rd|th)')
+_number_re = re.compile(r'[0-9]+')
+
+
+def _remove_commas(m: re.Match) -> str:
+    return m.group(1).replace(',', '')
+
+
+def _expand_decimal_point(m: re.Match) -> str:
+    return m.group(1).replace('.', ' point ')
+
+
+def _expand_dollars(m: re.Match) -> str:
+    match = m.group(1)
+    parts = match.split('.')
+    if len(parts) > 2:
+        return match + ' dollars'  # Unexpected format
+    dollars = int(parts[0]) if parts[0] else 0
+    cents = int(parts[1]) if len(parts) > 1 and parts[1] else 0
+    if dollars and cents:
+        dollar_unit = 'dollar' if dollars == 1 else 'dollars'
+        cent_unit = 'cent' if cents == 1 else 'cents'
+        return '%s %s, %s %s' % (dollars, dollar_unit, cents, cent_unit)
+    elif dollars:
+        dollar_unit = 'dollar' if dollars == 1 else 'dollars'
+        return '%s %s' % (dollars, dollar_unit)
+    elif cents:
+        cent_unit = 'cent' if cents == 1 else 'cents'
+        return '%s %s' % (cents, cent_unit)
+    else:
+        return 'zero dollars'
+
+
+def _expand_ordinal(m: re.Match) -> str:
+    return _inflect.number_to_words(m.group(0))
+
+
+def _expand_number(m: re.Match) -> str:
+    num = int(m.group(0))
+    if num > 1000 and num < 3000:
+        if num == 2000:
+            return 'two thousand'
+        elif num > 2000 and num < 2010:
+            return 'two thousand ' + _inflect.number_to_words(num % 100)
+        elif num % 100 == 0:
+            return _inflect.number_to_words(num // 100) + ' hundred'
+        else:
+            return _inflect.number_to_words(num, andword='', zero='oh', group=2).replace(', ', ' ')
+    else:
+        return _inflect.number_to_words(num, andword='')
+
+
+def normalize_numbers(text: str) -> str:
+    text = re.sub(_comma_number_re, _remove_commas, text)
+    text = re.sub(_pounds_re, r'\1 pounds', text)
+    text = re.sub(_dollars_re, _expand_dollars, text)
+    text = re.sub(_decimal_number_re, _expand_decimal_point, text)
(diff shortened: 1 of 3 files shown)


examples/pipeline_tacotron2/text/numbers.py
```diff
@@ -0,0 +1,71 @@
...
+_inflect = inflect.engine()
+_comma_number_re = re.compile(r'([0-9][0-9\,]+[0-9])')
+_decimal_number_re = re.compile(r'([0-9]+\.[0-9]+)')
+_pounds_re = re.compile(r'£([0-9\,]*[0-9]+)')
+_dollars_re = re.compile(r'\$([0-9\.\,]*[0-9]+)')
+_ordinal_re = re.compile(r'[0-9]+(st|nd|rd|th)')
+_number_re = re.compile(r'[0-9]+')
+
+
+def _remove_commas(m):
```
nateanl, line 16: nit: Let's add the type of the parameters ``m``

examples/pipeline_tacotron2/text/numbers.py
```diff
@@ -0,0 +1,71 @@
...
+    return m.group(1).replace(',', '')
+
+
+def _expand_decimal_point(m):
+    return m.group(1).replace('.', ' point ')
+
+
+def _expand_dollars(m):
```
nateanl, line 24: same here

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -0,0 +1,60 @@
...
+        sent (str): The input sentence to convert to a sequence.
+
+      Returns:
+        List of integers corresponding to the symbols in the sentence.
+    '''
+    sent = unidecode(sent)  # convert to ascii
+    sent = sent.lower()  # lower case
+    sent = normalize_numbers(sent)  # expand numbers
+    for regex, replacement in _abbreviations:  # expand abbreviations
+        sent = re.sub(regex, replacement, sent)
```
nateanl, line 57: Could you test the functionality of the method? I tried with some toy string but it didn't work:
```
sent = 'dr lee`
sent = unidecode(sent)  # convert to ascii
sent = sent.lower()  # lower case
for regex, replacement in _abbreviations:  # expand abbreviations
    sent = re.sub(regex, replacement, sent)
print(sent)
```
Then I got the same ``dr lee`` , not ``doctor lee``.

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -0,0 +1,60 @@
...
+    sent = re.sub(_whitespace_re, ' ', sent)  # collapse whitespace
+
+    return [_symbol_to_id[s] for s in sent if s in _symbol_to_id]
```
nateanl, line 60: So the preprocessing will ignore the characters that are not in ``symbol_to_id``. Where is the preprocessing used in tacotron?

nateanl (changes requested):

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: Yes, it will ignore characters that are not in it. You can only encode a finite set of symbols. This preprocessing will be used to preprocess the input text before sending into Tacotron2. You want to encode each character into a number.

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: For the current version, you'll have to go with `dr. lee` to get `docter lee`.

examples/pipeline_tacotron2/text/text_preprocessing.py
nateanl: I see. I guess it's only used in tacotron2 since other TTS model focus on the vocoder. If so, we can just put it here, otherwise we can put it as a general TTS function.

examples/pipeline_tacotron2/text/numbers.py
yangarbiter: Thanks for pointing it out!
I've added them here.

examples/pipeline_tacotron2/text/numbers.py
yangarbiter: Thanks for pointing it out!
I've added them on all functions in numbers.py.

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: BTW, I've also added some tests at `examples/pipeline_tacotron2/text/test_text.py`.

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: Yes, for now it is only for Tacotron2. But I doubt this will go into the core library as this is a text preprocessing function and torchaudio is an audio/signal processing library.

nateanl: It can be a general function if it's a standard processing method in TTS. Since it's Tacotron specific let's keep it here :)

nateanl (approved):
Thanks. LGTM!

examples/pipeline_tacotron2/text/numbers.py
```diff
@@ -0,0 +1,71 @@
+"""Modified from https://github.com/keithito/tacotron"""
```
mthrok, line 1: It's MIT license. Can add the copy right license notice?
https://github.com/keithito/tacotron/blob/master/LICENSE

yangarbiter: Thanks for pointing it out, I've uploaded it in `examples/pipeline_tacotron2/text/LICENSE`.
Thanks.

mthrok: Actually, can you copy-paste the LICENSE here directly?
That way, it is clearer which files are from `keithito/tacotron` and when we add something new to the same directory, it will be ambiguous which one belongs to the MIT license.

mthrok: Same goes to `text_preprocessing.py`.

mthrok: Similar to https://github.com/pytorch/audio/pull/1625/files#diff-446b770bff29efd0f406a17635f004c65a136ace539bd3d4eea923e14de8915dR1

yangarbiter: Thanks for pointing it out, I've updated `examples/pipeline_tacotron2/text/numbers.py` and `examples/pipeline_tacotron2/text/text_preprocessing.py`.

mthrok (approved):

PR #1642, opened 2021-07-27
Add tacotron2 training script

yangarbiter:
This ports the training script from [here](https://github.com/NVIDIA/DeepLearningExamples/blob/5a8521ee05934170ace0243f62e81cd7d7774fbd/PyTorch/SpeechSynthesis/Tacotron2/train.py).

Here are the things that are different:
- The melspectrogram are also changed to using torchaudio's implemenation
- The data loading part has been changed to using torchaudio's dataset
- tensorboard support are added
- Multi-gpu codes are ported from apex to torch.distributed

### DIFF
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -1,1 +1,36 @@
 This is an example pipeline for text-to-speech using Tacotron2.
+
+
+## Install required packages
+
+Required packages
+```bash
+pip install librosa tqdm inflect
+```
+
+To use tensorboard
+```bash
+pip install tensorboard pillow
+```
+
+## Training Tacotron2
+
+The training of Tacotron2 can be invoked with the following command.
+
+```python
+python train.py --learning-rate 1e-3 --epochs 1501 --anneal-steps 500 1000 1500 --anneal-factor 0.1 \
+    --batch-size 96 --weight-decay 1e-6 --grad-clip 1.0 --text-preprocessor character \
+    --logging-dir ./logs --checkpoint-path ./ckpt.pth --dataset-path ./
+```
+
+The training script will use all GPUs that is available, please set the
+environment variable `CUDA_VISIBLE_DEVICES` if you don't want all GPUs to be used.
+The newest checkpoint will be saved to `./ckpt.pth` and the checkpoint with the best validation
+loss will be saved to `./best_ckpt.pth`.
+The training log will be saved to `./logs/train.log` and the tensorboard results will also
+be in `./logs`.
+
+If `./ckpt.pth` already exist, this script will automatically load the file and try to continue
+training from the checkpoint.
+
+This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -17,10 +17,19 @@ pip install tensorboard pillow
 
 The training of Tacotron2 can be invoked with the following command.
 
-```python
-python train.py --learning-rate 1e-3 --epochs 1501 --anneal-steps 500 1000 1500 --anneal-factor 0.1 \
-    --batch-size 96 --weight-decay 1e-6 --grad-clip 1.0 --text-preprocessor character \
-    --logging-dir ./logs --checkpoint-path ./ckpt.pth --dataset-path ./
+```bash
+python train.py \
+    --learning-rate 1e-3 \
+    --epochs 1501 \
+    --anneal-steps 500 1000 1500 \
+    --anneal-factor 0.1 \
+    --batch-size 96 \
+    --weight-decay 1e-6 \
+    --grad-clip 1.0 \
+    --text-preprocessor character \
+    --logging-dir ./logs \
+    --checkpoint-path ./ckpt.pth \
+    --dataset-path ./
 ```
 
 The training script will use all GPUs that is available, please set the
--- a/examples/pipeline_tacotron2/datasets.py
+++ b/examples/pipeline_tacotron2/datasets.py
@@ -134,7 +134,6 @@ def text_mel_collate_fn(batch: Tuple[Tensor, Tensor],
         n_frames_per_step (int): The number of frames to advance every step.
 
     Returns:
-    text (Tensor): .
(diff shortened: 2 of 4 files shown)


examples/pipeline_tacotron2/utils.py
```diff
@@ -0,0 +1,99 @@
...
+                           text_processor: Callable[[str], List[int]]) -> Tuple[Tensor, Tensor]:
+    d = []
+    for text in texts:
+        d.append(torch.IntTensor(text_processor(text)[:]))
+
+    text_padded, input_lengths = pad_sequences(d)
+    return text_padded, input_lengths
+
+
+def get_text_preprocessor(preprocessor_name: str) -> Tuple[List[str], Callable[[str], List[int]]]:
```
mthrok, line 91: nit: Normally, this kind of helper logic should be added when it becomes necessary. (in this case when adding "phoneme") because priorities change. but we can keep this for the sake of moving forward.

examples/pipeline_tacotron2/utils.py
```diff
@@ -0,0 +1,99 @@
...
+import logging
+import os
+import shutil
+from typing import List, Tuple, Callable
+
+import torch
+from torch import Tensor
+
+
+def save_checkpoint(state, is_best, filename):
```
mthrok, line 37: We do not need this logic. This was added in wavernn because our training cluster had 24 hours limit and we needed to resume multiple times. However using other logics like prefixing the filename with epoch, one does not need 'temp' mechanism.

examples/pipeline_tacotron2/train.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+def main(args):
+    logger.info("Start time: {}".format(str(datetime.now())))
+
+    torch.manual_seed(0)
+    random.seed(0)
+
+    os.environ['MASTER_ADDR'] = 'localhost'
+    os.environ['MASTER_PORT'] = '17778'
```
mthrok, line 460: This will always use the same configuration and depending on user's port configuration, it might not be usable.
Typically, this kind of configuration has priorities like 1. set if from CLI, 2. resort to env ver 3. use default values.

examples/pipeline_tacotron2/train.py
```diff
@@ -0,0 +1,485 @@
...
+
+    logger.info(f"# available GPUs: {device_counts}")
+
+    # download dataset is not already downloaded
+    if args.dataset == 'ljspeech':
+        if not os.path.exists(os.path.join(args.dataset_path, 'LJSpeech-1.1')):
+            from torchaudio.datasets import LJSPEECH
+            LJSPEECH(root=args.dataset_path, download=True)
+
+    if device_counts == 1:
```
mthrok, line 472: Organizing the main workflow into layers of operation will make the script more readable, for example;

Just an illustration

```python

def main():
    model = _init_modules()
    dataset = _init_dataset()
    with _init_distributed():  # <- initialize / finalize distributed training config.
        _main_train_loop()  # <- this helper is agnostic to distributed training config
```

You can look into source separation example for the actual implementation of modularization of training steps.

mthrok, line 472: Organizing the main workflow into layers of operation will make the script more readable, for example;

Just an illustration

```python

def main():
    model = _init_modules()
    dataset = _init_dataset()
    with _init_distributed():  # <- initialize / finalize distributed training config.
        _main_train_loop()  # <- this helper is agnostic to distributed training config
```

You can look into source separation example for the actual implementation of modularization of training steps.

examples/pipeline_tacotron2/train.py
```diff
@@ -0,0 +1,485 @@
...
+    writer.add_figure("trn/mel_out", fig, epoch)
+    fig = plt.figure()
+    ax = plt.gca()
+    ax.imshow(mel_out_postnet[0].cpu().numpy())
+    writer.add_figure("trn/mel_out_postnet", fig, epoch)
+    writer.add_image("trn/gate_out", torch.tile(gate_out[:1], (10, 1)), epoch, dataformats="HW")
+    writer.add_image("trn/alignment", alignment[0], epoch, dataformats="HW")
+
+
+def train(rank, world_size, args):
```
mthrok, line 261: This function does a lot of work, extracting the sub routines would improve the readability.

https://en.wikipedia.org/wiki/Single-responsibility_principle

mthrok (approved):

examples/pipeline_tacotron2/datasets.py
```diff
@@ -0,0 +1,172 @@
...
+    """The collate function padding and adjusting the data based on `n_frames_per_step`.
+    Modified from https://github.com/NVIDIA/DeepLearningExamples
+
+    Args:
+        batch (tuple of two tensors): the first tensor is the mel spectrogram with shape
+            (n_batch, n_mels, n_frames), the second tensor is the text with shape (n_batch, ).
+        n_frames_per_step (int): The number of frames to advance every step.
+
+    Returns:
+    text (Tensor): .
```
mthrok, line 137:
```suggestion
    text (Tensor):
```

yangarbiter: Thanks for pointing it out, its addressed [here](https://github.com/pytorch/audio/pull/1642/commits/a270c524ff3f2d49f8554b334f2bba8cd4eb3858)

examples/pipeline_tacotron2/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    r"""Save the model to a temporary file first, then copy it to filename,
+    in case signals interrupt the torch.save() process.
+    """
+
+    if filename == "":
```
mthrok, line 42: nit : This kind of branching is making the function silently fails, which often is a cause of subtle bug.

examples/pipeline_tacotron2/README.md
```diff
@@ -1 +1,36 @@
...
+```bash
+pip install tensorboard pillow
+```
+
+## Training Tacotron2
+
+The training of Tacotron2 can be invoked with the following command.
+
+```python
+python train.py --learning-rate 1e-3 --epochs 1501 --anneal-steps 500 1000 1500 --anneal-factor 0.1 \
```
mthrok, line 21: nit: can you place each option into a dedicated line? It's easier to read that way.

Issue #1564, opened 2021-06-10
Add example codes for documentation


harishsdev:
## 📚 Documentation

I have observed that Example codes are missing from Documentation for the below mentioned Classes from torchaudio, Can I Raise PR for it

Group A

- [x] Spectrogram  PR:https://github.com/pytorch/audio/pull/1566
- [x] GriffinLim  PR:https://github.com/pytorch/audio/pull/1671/
- [ ] AmplitudeToDB
- [ ] MelScale
- [ ] InverseMelScale
- [ ] MFCC
- [x] MuLawEncoding  PR:https://github.com/pytorch/audio/pull/1586
- [x] MuLawDecoding  PR:https://github.com/pytorch/audio/pull/1684
- [ ] SpectralCentroid

Group B

- [x] Resample PR:https://github.com/pytorch/audio/pull/1644
- [x] ComplexNorm PR:https://github.com/pytorch/audio/pull/1658
- [x] ComputeDeltas
- [ ] TimeStretch
- [x] Fade

Group C

- [ ] FrequencyMasking
- [ ] TimeMasking
- [x] Vol
- [x] SlidingWindowCmn
- [x] Vad

mthrok:
Hi @harishsdev 

Yes, this will be a great addition. Feel free to add PRs. However, please split PRs for each class, (make one PR per class) so that it is easier to review.

Also, some of them are covered in tutorial, so please have a look it as well. 
https://pytorch.org/tutorials/beginner/audio_preprocessing_tutorial.html#feature-extractions

mthrok:
I will make this issue open for contribution.

harishsdev:
I am working on Group A

harishsdev:
@mthrok please help to review

vivekkumar7089:
After following the instructions given in `contributing.md` for development and installation, when I am trying to build docs locally I am getting this error after running `make html` 
```
Running Sphinx v2.4.4
checking for /home/vivek_kumar/OpenSource/audio/docs/source/refs.bib in bibtex cache... not found
parsing bibtex file /home/vivek_kumar/OpenSource/audio/docs/source/refs.bib... parsed 10 entries
loading intersphinx inventory from https://docs.python.org/objects.inv...
loading intersphinx inventory from https://docs.scipy.org/doc/numpy/objects.inv...
loading intersphinx inventory from https://pytorch.org/docs/stable/objects.inv...
[...]

PR #1644, opened 2021-07-28
added example for Resampler

radioactive11:
Added example for [Resample](https://pytorch.org/audio/stable/transforms.html#torchaudio.transforms.Resample) as mentioned in [this issue](https://github.com/pytorch/audio/issues/1564)

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -643,6 +643,11 @@ class Resample(torch.nn.Module):
             cached as ``torch.float64``. If you use resample with lower precision, then instead of providing this
             providing this argument, please use ``Resample.to(dtype)``, so that the kernel generation is still
             carried out on ``torch.float64``.
+
+    Example
+        >>> waveform, sample_rate = torchaudio.load('test.wav', normalization=True)
+        >>> resampler = transforms.Resample(sample_rate, sample_rate/10)
+        >>> waveform = resampler(waveform)
     """
 
     def __init__(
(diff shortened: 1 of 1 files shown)


torchaudio/transforms.py
```diff
@@ -643,6 +643,11 @@ class Resample(torch.nn.Module):
             cached as ``torch.float64``. If you use resample with lower precision, then instead of providing this
             providing this argument, please use ``Resample.to(dtype)``, so that the kernel generation is still
             carried out on ``torch.float64``.
+
+    Example
+        >>> waveform, sample_rate = torchaudio.load('test.wav', normalization=True)
```
nateanl, line 648: nit: ``normalization`` is deprecated in ``torch.load``. Let's change it to ``normalize``

torchaudio/transforms.py
```diff
@@ -643,6 +643,11 @@ class Resample(torch.nn.Module):
...
+        >>> resampler = transforms.Resample(sample_rate, sample_rate/10)
```
nateanl, line 649: Rename ``resampler`` as ``transform`` so that we have all examples consistent.

nateanl (changes requested):
Thanks for the PR! We can merge it after addressing those comments.

torchaudio/transforms.py
radioactive11: Sure thing!

torchaudio/transforms.py
radioactive11: Thank you for the suggestion!

nateanl (approved):

nateanl:
Thanks for your contribution @radioactive11! I have merged it.

PR #1648, opened 2021-07-29
Add Tacotron2 inference method

yangarbiter:
Tacotron2 inference method ported from [nvidia](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/Tacotron2/tacotron2/model.py).
Here are the things that are changed:
- Implemented the `infer` method for Tacotron2
- jit consistency test and output shape test for this `infer` method
- [This line is changed](https://github.com/pytorch/audio/pull/1648#issue-699091717) because `decoder_early_stopping` is by default True and it is an error from the last commit.
- Some of the variable names are also renamed for consistency ([here](https://github.com/pytorch/audio/pull/1648/files#diff-8b8ec57192b9aaeb99a9ec5e709c977771b31b49fe7ddc6704c2128cf161a00eR642) and [here](https://github.com/pytorch/audio/pull/1648/files#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR137))

### DIFF
--- a/test/torchaudio_unittest/models/tacotron2/model_test_impl.py
+++ b/test/torchaudio_unittest/models/tacotron2/model_test_impl.py
@@ -9,21 +9,28 @@ from torchaudio_unittest.common_utils import (
 class TorchscriptConsistencyMixin(TempDirMixin):
     r"""Mixin to provide easy access assert torchscript consistency"""
 
-    def _assert_torchscript_consistency(self, model, tensors):
+    def _assert_torchscript_consistency(self, model, tensors, is_inference=False):
         path = self.get_temp_path("func.zip")
         torch.jit.script(model).save(path)
         ts_func = torch.jit.load(path)
 
         torch.random.manual_seed(40)
-        output = model(*tensors)
+        if not is_inference:
+            output = model(*tensors)
+        else:
+            output = model.infer(*tensors)
 
         torch.random.manual_seed(40)
-        ts_output = ts_func(*tensors)
+        if not is_inference:
+            ts_output = ts_func(*tensors)
+        else:
+            ts_output = ts_func.infer(*tensors)
 
         self.assertEqual(ts_output, output)
 
 
 class Tacotron2EncoderTests(TestBaseMixin, TorchscriptConsistencyMixin):
+
     def test_tacotron2_torchscript_consistency(self):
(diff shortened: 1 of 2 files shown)


nateanl (approved):
LGTM

test/torchaudio_unittest/models/tacotron2/model_test_impl.py
```diff
@@ -9,21 +9,28 @@
 class TorchscriptConsistencyMixin(TempDirMixin):
     r"""Mixin to provide easy access assert torchscript consistency"""
 
-    def _assert_torchscript_consistency(self, model, tensors):
+    def _assert_torchscript_consistency(self, model, tensors, is_inference=False):
```
mthrok, line 12: Instead of adding `is_inference`, can your wrap your model with `torch.nn.Module` and define `forward` that redirect to `infer` method?

yangarbiter: Thanks for the suggestion.
I've remove `is_inference` and use the wrapper implemented [here](https://github.com/pytorch/audio/pull/1648/files#diff-001793fb337f116c66a3e7f3869a18012a618ab079df93db381fec0f45b63d9eR9-R18).

torchaudio/prototype/tacotron2.py
```diff
@@ -947,3 +1070,35 @@ def forward(
             gate_outputs.masked_fill_(mask[:, 0, :], 1e3)
 
         return mel_specgram, mel_specgram_postnet, gate_outputs, alignments
+
+    @torch.jit.export
+    def infer(self, text: Tensor, text_lengths: Tensor) -> Tuple[Tensor, Tensor, Tensor]:
+        r"""Using Tacotron2 for inference. This is generally used for generating mel spectrograms.
```
mthrok, line 1076: Can you describe what this function does and/or how this is expected to be used?

mthrok (approved):

torchaudio/prototype/tacotron2.py
yangarbiter: Thanks for the suggestion, it's added [here](https://github.com/pytorch/audio/pull/1648/commits/e2315340dee97e812ce5cd066c0c6de119608cd0).

Issue #942, opened 2020-10-09
InverseSpectrogram transform


dvisockas:
## 🚀 Feature
I'd like to propose adding of `InverseSpectrogram` transform to `torchaudio`

## Motivation
`istft` is already present in `torchaudio.functional` and if users want to build a cepstrum, they have to wrap the functional method into a torch module themselves.

## Pitch

Adding a wrapper for `functinal.istft`, similar to how `Spectrogram` transform wraps `F.spectrogram`

## Additional context

This thought arose from the fact that I was thinking about quick ways to implement various non-mel scale cepstral coefficients in pytorch (LFCC, CQCC, LPCC, BFCC, etc). Implementing all the various filterbanks seemed like it would overload the library. Adding the transform would alleviate the complexity needed to extract various cepstral coefficients.

mthrok:
Hi @dvisockas 

Thanks for the proposal. Could you help me understand your proposal? The proposal is about Transform of Inverse of Spectrogram and which uses `istft`. Is that correct? I am asking this because `istft` was moved to PyTorch core and the one in `torchaudio` has been removed in master branch.

I think this is a good addition. Having a set of functionalities that can achieve roundtrip between time dimension and frequency dimension.

dvisockas:
Precisely that. Basically a transform that wraps `torch.istft`. 

The ability to make an easy roundrip between time and frequency is the main thing why I am proposing this.

vincentqb:
Thanks for the idea :) Can you explain how the algorithm differ from [Griffin-Lim](https://github.com/pytorch/audio/blob/master/torchaudio/functional.py#L102)?

dvisockas:
Griffin-Lim takes a magnitude-only spectrogram and tries to recover the phase. `istft` takes both magnitude and phase tensor as an input. If you want to go into the frequency domain without discarding the phase you can pass `power=None` to the `Spectrogram` transform and you will get a magnitude-phase tensor. If you then do some transforms and want to go back to the time domain, you could pass only the magnitude part to the `Griffin-Lim`, but the phase would have to be reconstructed which is both slower and less accurate.

vincentqb:
Good point :) Yes, that sounds like a good addition.

To test, how close are the original waveform and the reconstructed waveform? Since stft and istft are close, I would expect Spectrogram and InverseSpectrogram to be close. Is that correct? Otherwise, do you have a reference to match? I don't see it in librosa.

dvisockas:
[...]

PR #1652, opened 2021-07-29
Add InverseSpectrogram to transforms and functional

nateanl:
Solves #942. This PR is a replacement of #1593


torchaudio/functional/functional.py
```diff
@@ -135,6 +136,102 @@ def spectrogram(
...
+            Default: ``True``
+        return_complex (bool, optional):
+            This value is ignored, and a real-valued output is always returned. It is provided only
+            for compatibility with the spectrogram function.
+            Default: ``False``
+
+    Returns:
+        Tensor: Dimension (..., time). Least squares estimation of the original signal.
+    """
+    if power is not None:
```
mthrok, line 189: Let's remove the `power` option as the arguments are not completely symmetric to `Spectrogram`.

test/torchaudio_unittest/transforms/transforms_test_impl.py
```diff
@@ -82,3 +82,28 @@ def test_resample_cache_dtype(self, resampling_method, dtype):
...
+        param(n_fft=300, center=True, onesided=True),
+        param(n_fft=400, center=True, onesided=False),
+        param(n_fft=400, center=True, onesided=False),
+        param(n_fft=300, center=True, onesided=False, return_complex=False),
+        param(n_fft=400, hop_length=10),
+        param(n_fft=800, win_length=400, hop_length=20),
+        param(n_fft=800, win_length=400, hop_length=20, normalized=True),
+        param(),
+        param(n_fft=400, pad=32),
+        #   these do not work - cause runtime error :(
```
mthrok, line 96: Can you add the link to https://github.com/pytorch/pytorch/issues/62323?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -79,6 +79,41 @@ def func(tensor):
...
+            ws = 400
+            hop = 200
+            pad = 0
+            window = torch.hann_window(ws, device=tensor.device, dtype=torch.view_as_real(tensor).dtype)
+            power = None
+            normalize = False
+            return F.inverse_spectrogram(tensor, length, pad, window, n_fft, hop, ws, power, normalize)
+
+        waveform = common_utils.get_whitenoise(sample_rate=8000, duration=0.05)
+        tensor = F.spectrogram(waveform, window=torch.hann_window(400, device=waveform.device, dtype=waveform.dtype),
```
mthrok, line 95: It's ideal if each unit test uses only the function to be tested. Say if a bug is introduced in `F.spectrogram`, then this test might also fail, which is a false signal.

There is [`get_spectrogram`](https://github.com/pytorch/audio/blob/16f3b2f943237f883956e82e3d71cf99622226b9/test/torchaudio_unittest/common_utils/data_utils.py#L115) function, which practically does the same thing as `F.spectrogarm`.

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -76,6 +76,27 @@ def test_spectrogram(self, kwargs):
...
+        ({'pad': 0, 'normalized': True, 'power': None, 'return_complex': True}, ),
+        ({'pad': 3, 'normalized': True, 'power': None, 'return_complex': True}, ),
+        ({'pad': 0, 'normalized': False, 'power': None}, ),
+        ({'pad': 3, 'normalized': False, 'power': None}, ),
+        ({'pad': 0, 'normalized': True, 'power': None}, ),
+        ({'pad': 3, 'normalized': True, 'power': None}, ),
+    ])
+    def test_inverse_spectrogram(self, kwargs):
+        # create a realistic input:
+        transform = T.Spectrogram(**kwargs)
```
mthrok, line 91: Similarly, it's ideal to not use `T.Spectrogram`, but we need to update the interface of [`get_spectrogram`](https://github.com/pytorch/audio/blob/9078c0b99e642943c193ee22b266f47fc21eda31/test/torchaudio_unittest/common_utils/data_utils.py#L115) for that.

torchaudio/functional/functional.py
```diff
@@ -135,6 +136,102 @@ def spectrogram(
...
+        win_length=win_length,
+        window=window,
+        center=center,
+        normalized=False,
+        onesided=onesided,
+        length=length + 2 * pad if length is not None else None,
+        return_complex=False,
+    )
+
+    if pad > 0:
```
mthrok, line 225: Isn't this `pad_mode`?

mthrok (approved):
I realized that `F.spectrogram` and `T.Spectrogram` have `pad` and `pad_mode` arguments and they are confusing.

Before merging, please make sure that these arguments in the `InverseSpectrogram` are consistent with them.

nateanl:
> I realized that `F.spectrogram` and `T.Spectrogram` have `pad` and `pad_mode` arguments and they are confusing.
> 
> Before merging, please make sure that these arguments in the `InverseSpectrogram` are consistent with them.

`pad_mode` is used in ``torch.stft`` but not used in ``torch.istft``. `pad` is used in both ``F.spectrogram`` and ``F.inversespectrogram``. Shall we remove the ``pad_mode`` in ``F.inversespectrogram``?

torchaudio/functional/functional.py
```diff
@@ -135,6 +136,95 @@ def spectrogram(
...
+        # remove padding from front and back
+        waveform = waveform[:, pad:-pad]
```
hwangjeff, line 220: here, `pad` is applied when it's larger than 0. in the call to `torch.istft` above, it's applied when `length` is not `None`. is this discrepancy expected?

nateanl:
After offline discussion with @hwangjeff, we found ``pad`` is a unique parameter compared with librosa. To address the discrepancy issue. There are two possible solutions:
- simply add a conditional statement for length in unpadding:
```
if length is not None and pad > 0:
        # remove padding from front and back
        waveform = waveform[:, pad:-pad]
```
- make length a required parameter. The reason is we assume people don't expect a waveform with a different length after InverseSpectrogram. For example:
waveform -> Spectrogram() -> InverseSpectrogram() -> waveform_inverse
However, if people create a complex spectrogram by themselves and only want to use InverseSpectrogram, do they still expect an length in the parameter? 
cc @mthrok

torchaudio/functional/functional.py
```diff
@@ -135,6 +136,95 @@ def spectrogram(
...
+        normalized (bool): Whether the stft output was normalized by magnitude
+        center (bool, optional): whether the waveform was padded on both sides so
+            that the :math:`t`-th frame is centered at time :math:`t \times \text{hop\_length}`.
+            Default: ``True``
+        pad_mode (string, optional): controls the padding method used when
+            :attr:`center` is ``True``. This parameter is provided for compatibility with the
+            spectrogram function and is not used. Default: ``"reflect"``
+        onesided (bool, optional): controls whether spectrogram was done in onesided mode.
+            Default: ``True``
+        return_complex (bool, optional):
```
mthrok, line 177: I think we can get rid of `return_complex`, for the same reason as we are removing the `power` argument. We cannot achieve the complete symmetry in function arguments, and complex-ness can be detected at run time.

torchaudio/functional/functional.py
```diff
@@ -135,6 +136,95 @@ def spectrogram(
...
+) -> Tensor:
+    r"""Create an inverse spectrogram or a batch of inverse spectrograms from the provided
+    complex-valued spectrogram.
+
+    Args:
+        spectrogram (Tensor): Complex tensor of audio of dimension (..., freq, time).
+            Alternatively, the tensor can be pseudo-complex, i.e. a real tensor of
+            dimension (..., freq, time, 2), if return_complex is False, although this
+            behavior is deprecated.
+        length (int): The amount to trim the signal by (i.e. the original signal length).
```
mthrok, line 161:
```suggestion
        length (int, optional): The amount to trim the signal by (i.e. the original signal length).
```

torchaudio/functional/functional.py
mthrok: @nateanl Is this resolved? The current code looks okay to me. My understanding is that `pad` is effective only when `length` is not `None`.

torchaudio/functional/functional.py
mthrok: I found the expression `The amount to trim the signal by` somewhat confusing. Can we simply say `the output length`?

torchaudio/functional/functional.py
```diff
@@ -135,6 +136,95 @@ def spectrogram(
...
+            Default: None (whole signal).
+        pad (int): Two sided padding of signal
```
mthrok, line 163: Can we note that this argument is only effective when a valid `length` is provided?

mthrok:
> After offline discussion with @hwangjeff, we found `pad` is a unique parameter compared with librosa. To address the discrepancy issue. There are two possible solutions:
> 
> * simply add a conditional statement for length in unpadding:
> 
> ```
> if length is not None and pad > 0:
>         # remove padding from front and back
>         waveform = waveform[:, pad:-pad]
> ```
> 
>  * make length a required parameter. The reason is we assume people don't expect a waveform with a different length after InverseSpectrogram. For example:
>   waveform -> Spectrogram() -> InverseSpectrogram() -> waveform_inverse
>   However, if people create a complex spectrogram by themselves and only want to use InverseSpectrogram, do they still expect an length in the parameter?
>   cc @mthrok

As a user, I expect I do not need to worry about the length (or any other parameters) if I am doing roundtrip with `Spectrogram` and `InversSpectrogram`.
In a case where I have a model that generates spectrogram and want to use InverseSpectrogram for constructing waveform, then I think it's reasonable that I have to give some information on the resulting waveform, and providing `length` makes sense. 

Looking at the current code, optional `length` parameter makes sense to me. Do you have some other form of API proposal?

nateanl:
> > After offline discussion with @hwangjeff, we found `pad` is a unique parameter compared with librosa. To address the discrepancy issue. There are two possible solutions:
> > 
> > * simply add a conditional statement for length in unpadding:
> > 
> > ```
> > if length is not None and pad > 0:
> >         # remove padding from front and back
> >         waveform = waveform[:, pad:-pad]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * make length a required parameter. The reason is we assume people don't expect a waveform with a different length after InverseSpectrogram. For example:
> >   waveform -> Spectrogram() -> InverseSpectrogram() -> waveform_inverse
> >   However, if people create a complex spectrogram by themselves and only want to use InverseSpectrogram, do they still expect an length in the parameter?
> >   cc @mthrok
> 
> As a user, I expect I do not need to worry about the length (or any other parameters) if I am doing roundtrip with `Spectrogram` and `InversSpectrogram`.
> In a case where I have a model that generates spectrogram and want to use InverseSpectrogram for constructing waveform, then I think it's reasonable that I have to give some information on the resulting waveform, and providing `length` makes sense.
> 
> Looking at the current code, optional `length` parameter makes sense to me. Do you have some other form of API proposal?

I agree with the current version so that people can either use it directly or pass a specified length to generate the waveform.

mthrok:
@nateanl Look like we forgot to add InverseSpectrogram to the documentation. I will file an issue for it.

PR #1653, opened 2021-07-30
Add melscale_fbanks and deprecate create_fb_matrix

Joool:
This commit addresses the third part of issue [#1574](https://github.com/pytorch/audio/issues/1574#issuecomment-862672386) by providing a clearer naming scheme for the creation of the melscale filter bank.

Since this is a public API change I assume there must be some kind of warning for users relying on this function (not yet implemented).
However, I assume the most common use case of the function is inside the MFCC transform.
These users do not need to take any action and the warning would probably cause more confusion than good.

Any thoughts?


nateanl:
Thanks for the PR @Joool! 
For the coming release cycle. I think we can include only the warning part to notify users that we are changing ``torchaudio.functional.create_fb_matrix`` to ``torchaudio.functional.melscale_fbanks`` in the "next release". 
 After we finish the coming release. We can merge this PR to master then.

mthrok:
@Joool @nateanl 

We make deprecate a feature, we take (at least) one release cycle to give users warning, and provide an alternative. 
In this case, removing `create_fb_matrix` is BC-breaking but adding `melscale_fbanks` is not BC-breaking.
When deprecating `create_fb_matrix`, we want to provide the alternative at the same time, so we should add `melscale_fbanks`, and in the deprecation message of `create_fb_matrix`, we should tell users to use `melscale_fbanks` instead.
As you pointed out, `create_fb_matrix` is use d in other transforms, we also have to change the any functions that uses `create_fb_matrix` to `melscale_fbanks`.

Here is the gist of the procedure.
In this PR;
1. Rename `create_fb_matrix ` to `melscale_fbanks`. (as already done, but documentations should have both `create_fb_matrix` and `mescale_fbanks`)
2. Add `create_fb_matrix` interface, which only issues warnings and pass all the parameters to `melscale_fbanks`. (So that only those who use `create_fb_matrix` explicitly will be warned.)
3. Tag this PR and commit title as `[BC-Breaking]`.
After the next release
1. Remove `create_fb_matrix`.

nateanl:
@mthrok  So we will have two PRs for the change. 
For this PR, we keep both ``create_fb_matrix`` and ``melscale_fbanks``, but include warnings in ``create_fb_matrix`` and redirect the params to ``melscale_fbanks``.
The next PR will be tagged as [BC-Breaking] since it removes ``create_fb_matrix``.

mthrok:
> @mthrok So we will have two PRs for the change.
> For this PR, we keep both `create_fb_matrix` and `melscale_fbanks`, but include warnings in `create_fb_matrix` and redirect the params to `melscale_fbanks`.
> The next PR will be tagged as [BC-Breaking] since it removes `create_fb_matrix`.

Oops, and yes that's right I will fix my previous comment.

Joool:
This should do it then?

mthrok:
The tests are not running. Let me close and re-open the PR to see if it helps.

mthrok:
> This should do it then?

@Joool Yes. LGTM.

mthrok (approved):

nateanl (approved):
LGTM!

test/torchaudio_unittest/functional/librosa_compatibility_test_impl.py
```diff
@@ -76,7 +76,7 @@ def test_griffinlim(self, momentum):
         [param(norm=n) for n in [None, 'slaney']],
         [param(mel_scale=s) for s in ['htk', 'slaney']],
     )
-    def test_create_fb(self, n_mels=40, sample_rate=22050, n_fft=2048,
+    def test_create_mel_fb(self, n_mels=40, sample_rate=22050, n_fft=2048,
                        fmin=0.0, fmax=8000.0, norm=None, mel_scale="htk"):
```
nateanl, line 80: Oops, just realized it fails the flake8 check. Could you fix the indentation?

Joool: Sorry about that. I really need to fix my linting/autoformatting (my other projects use different rules so I simply disabled it -.-).

PR #1654, opened 2021-07-30
Add an inference example for Tacotron2

yangarbiter:
An example for doing Tacotron2 inferencing.
#1642 #1648 should be merged before this can work.

### DIFF
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -146,11 +146,12 @@ python train.py \
 ## Text-to-speech pipeline
 
 Here we present an example of how to use Tacotron2 to generate audio from text.
-For the text-to-speech pipeline, with an input text, 
-Tacotron2 will first generate the corresponding mel spectrogram.
-After the mel spectrograms is generated, a vocoder will take in the mel spectrograms
-and convert them in to a waveform.
-Currently, there are three vocoders being supported, which includes the
+The text-to-speech pipeline goes as follows:
+1. text preprocessing: encoder the text into list of symbols (the symbols can represent characters, phonemes, etc.)
+2. spectrogram generation: after retrieving the list of symbols, we feed this list to a Tacotron2 model and the model
+will output the mel spectrogram.
+3. time-domain conversion: when the mel spectrogram is generated, we need to convert it into audio with a vocoder.
+Currently, there are three vocoders being supported in this script, which includes the
 [WaveRNN](https://pytorch.org/audio/stable/models/wavernn.html),
 [Griffin-Lim](https://pytorch.org/audio/stable/transforms.html#griffinlim), and
 [Nvidia's WaveGlow](https://pytorch.org/hub/nvidia_deeplearningexamples_tacotron2/).
@@ -195,6 +196,26 @@ python inference.py --checkpoint-path ${model_path} \
     --output-path "./outputs.wav"
 ```
 
+To use torchaudio pretrained models, please see the following example command.
+For Tacotron2, we use the checkpoint named `"tacotron2_english_phonemes_1500_epochs_wavernn_ljspeech"`, and
+for WaveRNN, we use the checkpoint named `"wavernn_10k_epochs_8bits_ljspeech"`.
+See https://pytorch.org/audio/stable/models.html for more checkpoint options for Tacotron2 and WaveRNN.
+
+```bash
+python inference.py \
+    --checkpoint-path tacotron2_english_phonemes_1500_epochs_wavernn_ljspeech \
+    --wavernn-checkpoint-path wavernn_10k_epochs_8bits_ljspeech \
+    --vocoder wavernn \
+    --n-fft 2048 \
+    --mel-fmin 40 \
+    --mel-fmax 11025 \
+    --input-text "Hello world!" \
+    --text-preprocessor english_phonemes \
+    --phonimizer DeepPhonemizer \
+    --phoimizer-checkpoint ./en_us_cmudict_forward.pt \
+    --cmudict-root ./ \
+    --output-path "./outputs.wav"
+```
 
 #### Griffin-Lim's algorithm as the Vocoder
 
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -141,3 +141,95 @@ python train.py \
     --checkpoint-path ./english_phonemes_wavernn_ckpt.pth \
     --dataset-path ./
 ```
+
+
+## Text-to-speech pipeline
+
+Here we present an example of how to use Tacotron2 to generate audio from text.
+For the text-to-speech pipeline, with an input text, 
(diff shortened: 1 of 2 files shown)


examples/pipeline_tacotron2/README.md
```diff
@@ -141,3 +141,95 @@ python train.py \
     --checkpoint-path ./english_phonemes_wavernn_ckpt.pth \
     --dataset-path ./
 ```
+
+
+## Text-to-speech pipeline
+
+Here we present an example of how to use Tacotron2 to generate audio from text.
+For the text-to-speech pipeline, with an input text, 
+Tacotron2 will first generate the corresponding mel spectrogram.
```
mthrok, line 150: nit: for those who do not know how TTS works, giving a brief description of process (text preprocessing, feature extraction, spectrogram generation, time-domain conversion) would be nice. (if not done so)
Maybe it is better suited for tutorial.

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,294 @@
...
+import argparse
+import os
+import random
+import sys
+
+import torch
+import torchaudio
+import numpy as np
+from torchaudio.prototype.tacotron2 import Tacotron2
+from torchaudio.models.wavernn import _MODEL_CONFIG_AND_URLS
```
mthrok, line 11: Can you put this into `_parse_args` so that it will be (somewhat) hidden?

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,294 @@
...
+        if args.jit:
+            vocoder = torch.jit.script(vocoder)
+
+        waveform = vocoder(mel_specgram.cpu())
+
+    torchaudio.save(args.output_path, waveform, args.sample_rate)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description='TTS Generator')
```
mthrok, line 290: What I often do when writing a script is;
1. Add a module-level docstring.
2. Refer the docstring `__doc__` from `ArgumentParser(description=__doc__)`.
That way, one can write detailed docstring which is accessible from CLI as well.

(Also I would put the construction of parer in `parse_args` as well.)

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,294 @@
...
+
+def main(args):
+    torch.manual_seed(0)
+    random.seed(0)
+    np.random.seed(0)
+
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    n_symbols = len(get_symbol_list(args.text_preprocessor))
+    text_preprocessor = partial(
```
mthrok, line 186: Similar to https://github.com/pytorch/audio/pull/1668/files#r686174534, turning this into a class will make the code more maintainable.

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,294 @@
...
+        "--wavernn-batch-overlap",
+        default=550,
+        type=int,
+        help="The overlapping time steps between batches. Only used when batch inference is used",
+    )
+
+    return parser
+
+
+def unwrap_distributed(state_dict):
```
mthrok, line 164: I feel we should update the pre-trained model, so that the structure is unwrapped, if that's not BC-breaking. Thoughts?

yangarbiter: For this part, it is used to load the checkpoints after the user train a model with `train.py`. So I think we can keep it this way (assuming the user trained the model from scratch themselfves.

To load our pertained model build in the library, I'll create a separate argument for it.

examples/pipeline_tacotron2/inference.py
yangarbiter: I'll submit a separate PR for this. I've created an issue #1710 to keep track of this.

examples/pipeline_tacotron2/inference.py
yangarbiter: Thanks for the reminder, I've move the parser into `parse_args` and wrote a simple module docstring.

examples/pipeline_tacotron2/inference.py
yangarbiter: Good suggestion! I've moved it into `parse_args` [here](https://github.com/pytorch/audio/pull/1654/files#diff-190b61efc2a83079d005a5da48a9fde8b28a937a8a653a8b4ede3b792f742384R31)

examples/pipeline_tacotron2/README.md
yangarbiter: I've updated the README [here](https://github.com/pytorch/audio/pull/1654/files#diff-a17fd0fdb11c8336af63efc614215c0ef98e67d810b77c5c5e8d87a012ca64f1R148-R157) and hopefully its a bit clearer.

And yes, when the tutorial is ready we should definitely add the link here.

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,320 @@
...
+    if args.jit:
+        tacotron2 = torch.jit.script(tacotron2)
+
+    sequences, lengths = prepare_input_sequence([args.input_text],
+                                                text_processor=text_preprocessor)
+    sequences, lengths = sequences.long().to(device), lengths.long().to(device)
+    with torch.no_grad():
+        mel_specgram, _, _ = tacotron2.infer(sequences, lengths)
+
+    if args.vocoder == "nvidia_waveglow":
```
mthrok, line 235: This `if ~ elif ~ else` clause is huge, which makes it difficult to follow the flow. Extracting each logic into helper functions help.

examples/pipeline_tacotron2/inference.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if args.jit:
+            raise ValueError("Vocoder option `nvidia_waveglow is not jittable.")
+
+        with torch.no_grad():
+            waveform = waveglow.infer(mel_specgram).cpu()
+
+    elif args.vocoder == "wavernn":
+        from torchaudio.models import wavernn
+        sys.path.append(os.path.join(os.path.dirname(__file__), "../pipeline_wavernn"))
```
mthrok, line 249: Not in the scope of this PR, but I think putting components for wavernn and tacotoron2 under the same directory, say `examples/tts` would allow to organize the code better.

mthrok (approved):

examples/pipeline_tacotron2/inference.py
yangarbiter: I've created [an issue](https://github.com/pytorch/audio/issues/1712) to keep track of this.

examples/pipeline_tacotron2/inference.py
yangarbiter: I've created an issue [here](https://github.com/pytorch/audio/issues/1713) for future improvement.

PR #1658, opened 2021-07-31
Added example for ComplexNorm

radioactive11:
Added example for [ComplexNorm](https://pytorch.org/audio/stable/transforms.html#complexnorm) as mentioned in this [issue](https://github.com/pytorch/audio/issues/1564)


torchaudio/transforms.py
```diff
@@ -780,6 +780,11 @@ class ComplexNorm(torch.nn.Module):
 
     Args:
         power (float, optional): Power of the norm. (Default: to ``1.0``)
+
+    Example
+        >>> complex_tensor = ... #  Tensor shape of (…, complex=2)
+        >>> transform = transforms.ComplexNorm(power=2)
+        >>> complex_tensor = transform(complex_tensor)
```
nateanl, line 787:
```suggestion
        >>> complex_norm = transform(complex_tensor)
```

nateanl (changes requested):
Hi @radioactive11! Sorry for my late reply. Just fix the nit and we can merge it to main branch.

torchaudio/transforms.py
radioactive11: No problems @nateanl. I have made the changes as per your suggestion :)

nateanl (approved):
Thanks. LGTM!

Issue #1662, opened 2021-08-02
`torchaudio.models` is not imported on top-level


mthrok:
Par #1660. The following should work.

```python
import torchaudio

model = torchaudio.models.XXX(...)
```

nateanl:
Shall we add models in ``__init__.py``?

mthrok:
> Shall we add models in `__init__.py`?

Yes!

mthrok:
addressed by #1664

PR #1664, opened 2021-08-02
Fix models import

nateanl:
This solves #1662.

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -4,6 +4,7 @@ from torchaudio import (
     compliance,
     datasets,
     functional,
+    models,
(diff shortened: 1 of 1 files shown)


torchaudio/__init__.py
```diff
@@ -25,6 +25,7 @@
     'compliance',
     'datasets',
     'functional',
+    'models',
```
mthrok, line 28: Don't we need to add the actual import statement at L3?

nateanl: Yep, sorry I ignored it. Now it's added.

mthrok (approved):

PR #1665, opened 2021-08-02
Fix encoding of CMUDict data reading

yangarbiter:
It appears that there will be an error for the current `torchaudio.datasets.CMUDict` due to using the wrong data loading encoding. According to [this](https://github.com/NVIDIA/DeepLearningExamples/blob/master/PyTorch/SpeechSynthesis/Tacotron2/tacotron2/text/cmudict.py#L23), we should use `latin-1` as the encoding method for loading CMUDict data.


mthrok (approved):
LGTM.

`chardetect` also reports Latin-1.
https://en.wikipedia.org/wiki/ISO/IEC_8859-1

```
chardetect cmudict-0.7b
cmudict-0.7b: ISO-8859-1 with confidence 0.73
```

PR #1668, opened 2021-08-03
Add phoneme text preprocessing for Tacotron2

yangarbiter:
Currently, it supports using [DeepPhonemizer](https://github.com/as-ideas/DeepPhonemizer) as the phonemizer since it is easy to install.

### DIFF
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -13,7 +13,7 @@ To use tensorboard
 pip install tensorboard pillow
 ```
 
-## Training Tacotron2
+## Training Tacotron2 with character as the input
 
 The training of Tacotron2 can be invoked with the following command.
 
@@ -43,3 +43,16 @@ If `./ckpt.pth` already exist, this script will automatically load the file and 
 training from the checkpoint.
 
 This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
+
+## Training Tacotron2 with phoneme as the input
+
+#### Dependencies
+
+If you want to use [DeepPhonemizer](https://github.com/as-ideas/DeepPhonemizer) as
+the phonemizer, please install with the following command (the code is tested with version 0.0.15).
+
+```bash
+pip install deep-phonemizer==0.0.15
+```
+
+Then download the model weights from [their website](https://public-asai-dl-models.s3.eu-central-1.amazonaws.com/DeepPhonemizer/en_us_cmudict_forward.pt).
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -5,7 +5,7 @@ This is an example pipeline for text-to-speech using Tacotron2.
 
 Required packages
 ```bash
-pip install librosa tqdm inflect
+pip install librosa tqdm inflect joblib
 ```
 
 To use tensorboard
@@ -13,7 +13,7 @@ To use tensorboard
 pip install tensorboard pillow
 ```
 
-## Training Tacotron2 with character as the input
+## Training Tacotron2 with character as input
 
 The training of Tacotron2 can be invoked with the following command.
 
@@ -26,7 +26,7 @@ python train.py \
     --batch-size 96 \
     --weight-decay 1e-6 \
     --grad-clip 1.0 \
-    --text-preprocessor character \
+    --text-preprocessor english_characters \
     --logging-dir ./logs \
     --checkpoint-path ./ckpt.pth \
     --dataset-path ./
@@ -44,7 +44,7 @@ training from the checkpoint.
 
 This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
 
-## Training Tacotron2 with phoneme as the input
+## Training Tacotron2 with phoneme as input
 
 #### Dependencies
 
@@ -56,3 +56,25 @@ pip install deep-phonemizer==0.0.15
 ```
 
 Then download the model weights from [their website](https://public-asai-dl-models.s3.eu-central-1.amazonaws.com/DeepPhonemizer/en_us_cmudict_forward.pt).
+
+#### Running training script
+
+The training of Tacotron2 with english phonemes as input can be invoked with the following command.
+
+```bash
+python train.py \
+    --learning-rate 1e-3 \
+    --epochs 1501 \
+    --anneal-steps 500 1000 1500 \
+    --anneal-factor 0.1 \
+    --batch-size 96 \
+    --weight-decay 1e-6 \
+    --grad-clip 1.0 \
+    --text-preprocessor english_phonemes \
+    --phonemizer DeepPhonemizer \
+    --phonemizer-checkpoint ./en_us_cmudict_forward.pt \
+    --cmudict-root ./ \
+    --logging-dir ./logs \
+    --checkpoint-path ./english_phonemes_ckpt.pth \
+    --dataset-path ./
+```
--- a/examples/pipeline_tacotron2/README.md
+++ b/examples/pipeline_tacotron2/README.md
@@ -44,18 +44,47 @@ training from the checkpoint.
 
 This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
 
+To train the Tacotron2 model to work with the [pretrained wavernn](https://pytorch.org/audio/main/models.html#id10)
+with checkpoint_name `"wavernn_10k_epochs_8bits_ljspeech"`, please run the following command instead.
+
+```bash
+python train.py
+    --learning-rate 1e-3 \
+    --epochs 1501 \
+    --anneal-steps 500 1000 1500 \
+    --anneal-factor 0.1 \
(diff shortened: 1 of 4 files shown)


examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+                     symbol_list: Union[str, List[str]] = "english_characters",
+                     phonemizer: Optional[str] = "DeepPhonemizer",
+                     checkpoint: Optional[str] = "./en_us_cmudict_forward.pt",
+                     cmudict_root: Optional[str] = "./") -> List[int]:
     r'''Converts a string of text to a sequence of IDs corresponding to the symbols in the text.
 
-      Args:
-        sent (str): The input sentence to convert to a sequence.
+        Args:
+            sent (str): The input sentence to convert to a sequence.
```
nateanl, line 122: the datatype of sent should be ``Iterable``?

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+            symbol_list (str or List of string, optional): When the input is a string, available options include
```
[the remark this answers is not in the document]
yangarbiter: I think the optional in the docstring and optional in types means two different things (correct me if I am wrong).
Optional in types means that it can take in the value `None`. While optional in docstring means that it has default value.

examples/pipeline_tacotron2/README.md
```diff
@@ -42,4 +42,39 @@ be in `./logs`.
...
 
-This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
\ No newline at end of file
+This command takes around 36 hours to train on 8 NVIDIA Tesla V100 GPUs.
+
+## Training Tacotron2 with phoneme as input
+
+#### Dependencies
+
+If you want to use [DeepPhonemizer](https://github.com/as-ideas/DeepPhonemizer) as
```
mthrok, line 51: Since this section is about the use of phoneme and there is no alternative prepared at the moment, the wording can be improved.

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+        # ['HH', 'AH', 'L', 'OW', ' ', 'W', 'ER', 'L', 'D', '!']
+        ret = [r.replace("[", "").replace("]", "") for r in ret]
+
+        return ret
+    else:
+        raise ValueError(f"The `phonemizer` {phonemizer} is not supported."
+                         "Supported `symbol_list` includes `'DeepPhonemizer'`.")
+
+
+def text_to_sequence(sent: str,
```
mthrok, line 114: If you want to get rid of the global variable, you can turn this into a class and perform the initialization in the constructor and reuse the resource.

mthrok: Also, having a different class for different operation mode (here, character vs phoneme) and preparing a factory function that instantiates the appropriate class will make the code easier to read.

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+
     sent = unidecode(sent)  # convert to ascii
     sent = sent.lower()  # lower case
     sent = normalize_numbers(sent)  # expand numbers
     for regex, replacement in _abbreviations:  # expand abbreviations
         sent = re.sub(regex, replacement, sent)
     sent = re.sub(_whitespace_re, ' ', sent)  # collapse whitespace
 
+    if isinstance(symbol_list, list):
+        symbols = symbol_list.copy()
```
mthrok, line 151: At a glance, this `copy` seems unnecessary. Please verify it.

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+    elif isinstance(symbol_list, str):
+        symbols = get_symbol_list(symbol_list, cmudict_root=cmudict_root)
+        if symbol_list == "english_phonemes":
+            if phonemizer is None:
+                raise ValueError(f"`phonemizer` should not be None when symbol_list is {symbol_list}.")
+            if checkpoint is None:
+                raise ValueError(f"`checkpoint` should not be None when symbol_list is {symbol_list}.")
+            if cmudict_root is None:
+                raise ValueError(f"`cmudict_root` should not be None when symbol_list is {symbol_list}.")
```
mthrok, line 160: This validation is better performed at the top of the function so that processing logic is more readable.

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,102 @@
...
+def text_to_sequence(sent: str,
+                     symbol_list: Union[str, List[str]] = "english_characters",
+                     phonemizer: Optional[str] = "DeepPhonemizer",
+                     checkpoint: Optional[str] = "./en_us_cmudict_forward.pt",
+                     cmudict_root: Optional[str] = "./") -> List[int]:
     r'''Converts a string of text to a sequence of IDs corresponding to the symbols in the text.
 
-      Args:
-        sent (str): The input sentence to convert to a sequence.
+        Args:
```
mthrok, line 121: Please verify if the indentation is correct.

mthrok (approved):

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: I've changed it back to `str`. Thanks for the reminder!

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: I'll refactor it in a separate PR.
Thanks for the suggestion.

examples/pipeline_tacotron2/README.md
yangarbiter: I've rewrite the description [here](https://github.com/pytorch/audio/pull/1668/files#diff-a17fd0fdb11c8336af63efc614215c0ef98e67d810b77c5c5e8d87a012ca64f1R76).
Thanks.

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: That's right, I've removed the copy [here](https://github.com/pytorch/audio/pull/1668/files#diff-2b633e14bbda0eeaa9126220bbcc9782404e90d7cf7a0a4b2ab1be29dda27e56R158)

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: Thanks for the reminder, I've moved them to the beginning of the function [here](https://github.com/pytorch/audio/pull/1668/files#diff-2b633e14bbda0eeaa9126220bbcc9782404e90d7cf7a0a4b2ab1be29dda27e56R141-R149).

examples/pipeline_tacotron2/text/text_preprocessing.py
yangarbiter: I've changed the indentation back to the correct one [here](https://github.com/pytorch/audio/pull/1668/files#diff-2b633e14bbda0eeaa9126220bbcc9782404e90d7cf7a0a4b2ab1be29dda27e56R119-R141)

examples/pipeline_tacotron2/text/text_preprocessing.py
```diff
@@ -63,23 +64,103 @@
...
+    Returns:
         List of integers corresponding to the symbols in the sentence.
+
+    Examples:
+        >>> text_to_sequence("hello world!", "english_characters")
+        [19, 16, 23, 23, 26, 11, 34, 26, 29, 23, 15, 2]
+        >>> text_to_sequence("hello world!", "english_phonemes")
+        [54, 20, 65, 69, 11, 92, 44, 65, 38, 2]
     '''
+    if symbol_list == "english_phonemes":
```
mthrok, line 142: nit1: when writing a error message, rather than stating "what it should NOT be", explaining "what it should be" is more user-friendly.
For example: `threshold value cannot be 0.5` -> `thireshold value must be smaller than 0.4`.

nit2: You can also combine these all three checks with `any` but it is rather subjective which one is more readable.

```python
if any(param is None for param in [phonemizer, checkpoint, cmudict_root]):
    raise ValueError(
        "When `symbol_list` is 'english_phonemes', "
        "all of `phonemizer`, `checkpoint`, and `cmudict_root` must be provided.")
```

yangarbiter: Thank you for the suggestion.
I would prefer nit2 in this case.
It's addressed [here](https://github.com/pytorch/audio/pull/1668/files#diff-2b633e14bbda0eeaa9126220bbcc9782404e90d7cf7a0a4b2ab1be29dda27e56R143-R146).

PR #967, opened 2020-10-18
Add basic filtfilt implementation

v-goncharenko:
Resolves #789 

I also could add tests for scipy compatibility as well.

P.S. sorry for late PR

Issue #789, opened 2020-07-16
🚀 Feature Request: Add Filtfilt (scipy.signal.filtfilt equivalent)


v-goncharenko:
## 🚀 Feature

Add utility `filtfilt` to `torchaudio.functional`. This utility applies `lfilt` twice: in forward and backward direction, so phase of resulting signal is preserved. This function exist for example in `scipy` (see [their docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.filtfilt.html)) and is frequently used in signal processing.

## Motivation

I frequently need this function, so I've already wrote and tested it, so I would make a PR to include it to core library.

## Pitch

Add bidirectional filtering function `filtfilt` commonly used in signal processing.

mthrok:
Hi @v-goncharenko 

Thanks for the suggestion. I do not have anything against adding this to `torchaudio`, especially if it's well tested. @vincentqb your thoughts?

If your existing code is publicly available, can I take a look at it?
Adding code to `torchaudio` will likely to have different requirements for testing. You can read about it [here](https://github.com/pytorch/audio/tree/master/test#torchaudio-test-suite). This case, we will need to add new type of test (spicy compatibility).

mthrok:
Question: the `spicy` doc says 

> The function `sosfiltfilt` (and filter design using `output='sos'`) should be preferred over `filtfilt` for most filtering tasks, as second-order sections have fewer numerical problems.

What's your take on this one? I am not familiar with these functions so would like to hear your opinion.

vincentqb:
> Thanks for the suggestion. I do not have anything against adding this to `torchaudio`, especially if it's well tested. @vincentqb your thoughts?

Yup, I'm ok with having this in torchaudio too @mthrok :) As usual, testing will need to ensure that the implementation support batch, torchscript, and correctness. It may be a good idea to have some testing around numerical stability given @mthrok's observation.

@v-goncharenko, please feel free to open a pull request for this!

v-goncharenko:
> Adding code to torchaudio will likely to have different requirements for testing. You can read about it here. This case, we will need to add new type of test (spicy compatibility).

My tests aimed to check scipy compatibility. And of course I'm ready to write some more if needed.

> What's your take on this one? I am not familiar with these functions so would like to hear your opinion.

Well, this requires `sosfilt` to be implemented, but now we have only `lfilter`, so this is topic for another issue and PR.

> @v-goncharenko, please feel free to open a pull request for this!

[...]

PR #1681, opened 2021-08-04
Add basic filtfilt implementation

hwangjeff:
Builds upon #967; addresses #789


torchaudio/functional/filtering.py
```diff
@@ -642,6 +642,34 @@ def equalizer_biquad(
...
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of ``(n_order + 1)``.
+            Lower delays coefficients are first, e.g. ``[b0, b1, b2, ...]``.
+            Must be same size as a_coeffs (pad with 0's as necessary).
+        clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
+
+    Returns:
+        Tensor: Waveform with dimension of ``(..., time)``.
+    """
+    forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, clamp)
+    backward_filtered = lfilter(forward_filtered.flip(-1), a_coeffs, b_coeffs, clamp).flip(-1)
```
nateanl, line 669: I moved the ``.flip(-1)`` to the end to make it more consistent 🙂
```suggestion
    forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, clamp).flip(-1)
    backward_filtered = lfilter(forward_filtered, a_coeffs, b_coeffs, clamp).flip(-1)
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -106,6 +106,84 @@ def test_lfilter_9th_order_filter_stability(self):
...
+        )
+
+        # Extend waveform in each direction, preserving periodicity.
+        padded_waveform = torch.cat((waveform[:-1], waveform, waveform[1:]))
+
+        output_waveform = F.filtfilt(padded_waveform, a_coeffs, b_coeffs)
+
+        # Remove padding from output waveform; confirm that result
+        # closely matches waveform_k0.
+        self.assertEqual(output_waveform[samples - 1: 2 * samples - 1], waveform_k0, atol=1e-3, rtol=1e-3)
```
nateanl, line 185: nit: ``output_waveform[samples - 1 : 2 * samples - 1]``

Could ``atol`` and ``rtol`` be smaller than 1e-3, or it will throw an error if we decrease it?

nateanl (commented):
The impl LGTM. We could also add test codes for autograd and torchscript.

torchaudio/functional/filtering.py
hwangjeff: i feel like the semantics of the original make more sense — `forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, clamp)` is indeed the forward-filtered waveform

test/torchaudio_unittest/functional/functional_impl.py
hwangjeff: the linter likes it without the space

the maximum error is around 1e-4

hwangjeff:
> The impl LGTM. We can also add test codes for autograd and torchscript.

ah i totally missed these — thanks. will add

torchaudio/functional/filtering.py
nateanl: Yea I'm convinced 🙂. Yours is better than scipy impl

test/torchaudio_unittest/functional/functional_impl.py
nateanl: let's try it with 1e-4 to see if it passes the test. As ``lfilter`` has a lower threshold (1e-5) for testing.

docs/source/functional.rst
```diff
@@ -152,6 +151,11 @@ equalizer_biquad
 
 .. autofunction:: equalizer_biquad
 
+filtfilt
+----------------
```
mthrok, line 155: Strange thing about RST format, the length of the horizontal line is supposed to be same as the title.

```suggestion
--------
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -106,6 +106,84 @@ def test_lfilter_9th_order_filter_stability(self):
...
         self.assertEqual(yhat, y, atol=1e-4, rtol=1e-5)
 
+    def test_filtfilt_simple(self):
+        """
+        Check that, for an arbitrary signal, applying filtfilt with filter coefficients
+        corresponding to a pure delay filter imparts no time delay.
+        """
+
+        torch.random.manual_seed(42)
+        waveform = torch.rand(2, 44100 * 1, dtype=self.dtype, device=self.device)
```
mthrok, line 116: Can you use `get_whitenoise` or `get_sinusoid`?
Also can you use smaller sample rate like 8000 / 16000? (<- this has nothing to do with the quality of the test but we often see timeout in fbcode)

yoyololicon:
@hwangjeff Thanks for the contribution.

Just want to remind you that we have added some new feature in `lfilter` through 8094751f5536a13e60e7e1f19305d0bf4eb8cc22 and aa0dd03b07b9f20b4026e9652e9cdd543278ed8a, which introduced a new `batching` parameter and some changes in the docstring. You might need to reflect the changes on `filtfilt` as well.

torchaudio/functional/filtering.py
```diff
@@ -642,6 +642,34 @@ def equalizer_biquad(
...
+            Must be same size as b_coeffs (pad with 0's as necessary).
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of ``(n_order + 1)``.
+            Lower delays coefficients are first, e.g. ``[b0, b1, b2, ...]``.
+            Must be same size as a_coeffs (pad with 0's as necessary).
+        clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
+
+    Returns:
+        Tensor: Waveform with dimension of ``(..., time)``.
+    """
+    forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, clamp)
```
yoyololicon, line 668: I'm thinking that whether should `clamp` option also be applied in the first pass. 
If we only want to avoid outrange value, this can just applied in the second pass of `lfilter`.

hwangjeff: makes sense. thanks for flagging — updated the logic

test/torchaudio_unittest/functional/functional_impl.py
hwangjeff: it fails with 1e-4, unfortunately

docs/source/functional.rst
hwangjeff: thanks, fixed

test/torchaudio_unittest/functional/functional_impl.py
hwangjeff: sure, done

hwangjeff:
added autograd, batch consistency, and torchscript tests; revised logic and comments to account for new batching/filterbanks logic introduced for `lfilter`

torchaudio/functional/filtering.py
```diff
@@ -642,6 +642,44 @@ def equalizer_biquad(
...
+                                    a_coeffs[i], b_coeffs[i], clamp=clamp, batching=False)``. (Default: ``True``)
+
+    Returns:
+        Tensor: Waveform with dimension of either ``(..., num_filters, time)`` if ``a_coeffs`` and ``b_coeffs``
+                are 2D Tensors, or ``(..., time)`` otherwise.
+    """
+    forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, batching=batching)
+    backward_filtered = lfilter(
+        forward_filtered.flip(-1), a_coeffs, b_coeffs, clamp=clamp, batching=batching
+    ).flip(-1)
```
yoyololicon, line 679: After the first pass of lfilter, the waveform shape might have changed because of using filterbanks, if this case happened, the second pass should be done in batching mode:
```suggestion
    forward_filtered = lfilter(waveform, a_coeffs, b_coeffs, clamp=False, batching=batching)
    if forward_filtered.shape != waveform.shape:
        batching = True
    backward_filtered = lfilter(
        forward_filtered.flip(-1), a_coeffs, b_coeffs, clamp=clamp, batching=batching
    ).flip(-1)
```

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -232,3 +232,30 @@ def test_lfilter(self):
...
+        num_order = 3
+        torch.manual_seed(296)
+
+        waveform = common_utils.get_whitenoise(sample_rate=8000, n_channels=4)
+        a_coeffs = torch.rand(num_filters, num_order + 1)
+        b_coeffs = torch.rand(num_filters, num_order + 1)
+
+        self.assert_batch_consistency(
+            F.filtfilt, waveform, a_coeffs, b_coeffs, batching=False
+        )
```
yoyololicon, line 261: These type of tests might not be neccessary. 
I recoommend only test the batching behavior of filter coefficients, like how `test_lfilter` do.

yoyololicon (commented):
Hi @hwangjeff 
Please let me know your thoughts on my comments.

torchaudio/functional/filtering.py
nateanl: I'm curious about the conditional logic here. If we set ``batching`` to ``False`` in filtfilt, will the shape of ``forward_filtered`` still be different to ``waveform``?

yoyololicon: It can still be different if the coefficients are 2D Tensor, thus operate in filterbanks mode, result in one extra dimension.

nateanl: I checked the lfilter impl. If ``waveform`` is 1D, ``a_coeffs`` is 2D, ``batching`` has to be ``False``, otherwise it will throw an error like ``assert waveform.ndim > 1``. And the output is 2D.
If ``batching`` is ``True``, then ``waveform`` has to be 2D, so is ``a_coeffs``, and the output is also 2D.
In the first case the output shape can be different from waveform. That makes sense to me.

nateanl (approved):
LGTM!

hwangjeff:
@yoyololicon thanks a lot for the review — made the changes you suggested

yoyololicon:
Besides the gpu test error, I have no other questions.

hwangjeff:
yeah i'll look into the test failures

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -78,6 +78,47 @@ def test_lfilter_batching(self):
                           [0.7, 0.2, 0.6]])
         self.assert_grad(F.lfilter, (x, a, b))
 
+    def test_filtfilt_a(self):
+        torch.random.manual_seed(2434)
+        x = get_whitenoise(sample_rate=22050, duration=0.01, n_channels=2)
```
mthrok, line 83: For the GPU test error, I think something like `get_whitenoise(..., device=self.device, dtype=self.dtype)` would do.

torchaudio/functional/filtering.py
```diff
@@ -642,6 +642,48 @@ def equalizer_biquad(
...
+        a_coeffs (Tensor): denominator coefficients of difference equation of dimension of either
+                                1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
+                                Lower delay coefficients are first, e.g. ``[a0, a1, a2, ...]``.
+                                Must be same size as b_coeffs (pad with 0's as necessary).
+        b_coeffs (Tensor): numerator coefficients of difference equation of dimension of either
+                                1D with shape ``(num_order + 1)`` or 2D with shape ``(num_filters, num_order + 1)``.
+                                Lower delay coefficients are first, e.g. ``[b0, b1, b2, ...]``.
+                                Must be same size as a_coeffs (pad with 0's as necessary).
+        clamp (bool, optional): If ``True``, clamp the output signal to be in the range [-1, 1] (Default: ``True``)
+        batching (bool, optional): Effective only when coefficients are 2D. If ``True``, then waveform should be at
```
mthrok, line 667: I personally would like not to expose `batching` option for now. (and pass `batching=True` to `lfilter`)
`batching=False` option in `lfilter` is very unique feature, not following the general convention of DL frameworks. So first I would like to gauge the popularity/usefulness in `lfilter` then adopt it to the other functions. From software perspective, it's easy to add a public interface but very difficult to remove it.

@yoyololicon What do you think?

yoyololicon: @mthrok 
`lfilter` and `filtfilt` are very similar functions so it looks natural to me to sync the changes together.
But I understand your concern,  I'm ok with not adopt this parameter on `filtfilt` first.
I do consider extend the batching/filterbanks feature on biquad filters as well in the future.

test/torchaudio_unittest/common_utils/data_utils.py
```diff
@@ -75,7 +75,8 @@ def get_whitenoise(
     tensor.clamp_(-1.0, 1.0)
     if not channels_first:
         tensor = tensor.t()
-    return convert_tensor_encoding(tensor, dtype)
+    tensor = convert_tensor_encoding(tensor, dtype)
+    return tensor.to(device=device)
```
hwangjeff, line 79: seems like this was unintentionally removed in a previous refactoring job — added it back in

hwangjeff: nvm — it breaks other tests. will address elsewhere

hwangjeff:
removed `batching` option per discussion above and fixed tests

mthrok (approved):
LGMT.

PR #1682, opened 2021-08-05
[Fbsync] Synchronize extension initialization mechanism with fbcode

mthrok:
ref: D29989551

note: The changes in `test/torchaudio_unittest/__init__.py ` in already commited on main branch but somehow it keeps showing up in the PR change. I am not sure how to remove it from this PR but as long as GH detects no conflict, it should be fine.


nateanl (approved):
LGTM. Just curious will it include all commits to main branch? Or you could clean the history in the PR.

mthrok:
> LGTM. Just curious will it include all commits to main branch? Or you could clean the history in the PR.

The commit history is that of `fbsync` branch, which has a different commits due to the treatment in facebook's code base but supposedly has the same code as `main` branch. When some change is introduced from facebook's internal system, it will be automatically pushed to `fbsync`, and we merge the `fbsync` branch to `main` branch to propagate it. It is indeed strange that every time we merge from `fbsync`, it contains all the history, but from the practical view point, it does not matter much.

PR #1685, opened 2021-08-06
[docs] Update sphinx to 3.5.4

nateanl:


mthrok:
Thanks for doing this;

Screen Shot 2021-08-06 at 11 44 45

The rendering is broken. I think this is caused by https://github.com/pytorch/pytorch_sphinx_theme/issues/115 . We have to decide to keep the current version or remove the type annotation from documentation like PyTorch core if we really want to update to 3.5.4. However, I am not aware of the necessity to do so.

nateanl:
I suggest waiting for the solution from PyTorch side and keeping this PR open until everything is settled. Since 3.5.4 is already hard-coded into PyTorch, it's good to follow that way (when the rendering is fixed).

mthrok:
@nateanl the issue in pytorch doc theme has been resolved. Let's retry and see how it goes.

nateanl:
This is how it looks now:
image

PR #1688, opened 2021-08-09
Update version setup.py for LTS

msaroufim:
To test

```
conda install -y pytorch=1.8.1 torchvision=0.9.1 torchaudio=0.8.1 torchtext=0.9.1 cpuonly -c pytorch-lts
```

Then run `python -c "import torchaudio; print(torchaudio.__version__)" `

Output should be `0.8.1`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ ROOT_DIR = Path(__file__).parent.resolve()
 
 
 # Creating the version file
-version = '0.8.0a0'
+version = '0.8.1'
 sha = 'Unknown'
 
 try:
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -12,7 +12,7 @@
 
 
 # Creating the version file
-version = '0.8.0a0'
+version = '0.8.1'
```
seemethere, line 15: This version shouldn't really matter since the version gets set by the git tag anyways through use of the `BUILD_VERSION` variable

seemethere (approved):

mthrok:
Merged par the direct request by @msaroufim.

PR #1693, opened 2021-08-10
Add tacotron2 pretrained models

yangarbiter:
Add tacotron2 pretrained models

### DIFF
--- a/torchaudio/prototype/tacotron2.py
+++ b/torchaudio/prototype/tacotron2.py
@@ -27,17 +27,75 @@
 
 import warnings
 from math import sqrt
-from typing import Tuple, List, Optional, Union
+from typing import Tuple, List, Optional, Union, Any, Dict
 
 import torch
 from torch import nn
 from torch import Tensor
 from torch.nn import functional as F
+from torch.hub import load_state_dict_from_url
 
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


torchaudio/prototype/tacotron2.py
```diff
@@ -1105,3 +1163,60 @@ def infer(self, text: Tensor, text_lengths: Tensor) -> Tuple[Tensor, Tensor, Ten
         alignments = alignments.unfold(1, n_batch, n_batch).transpose(0, 2)
 
         return mel_outputs_postnet, mel_specgram_lengths, alignments
+
+
+def tacotron2(checkpoint_name: str) -> Tacotron2:
+    r"""Get pretrained Tacotron2 model.
+
+    Args:
+        checkpoint_name (str): The name of the checkpoint to load. Available checkpoints:
```
mthrok, line 1172: I just realized that there is no documentation for Tactoron2. Can you add that so that we can see how this docstring is converted to documentation?

mthrok (approved):

torchaudio/prototype/tacotron2.py
yangarbiter: I'll address this in a separate PR.
Thanks.

PR #1702, opened 2021-08-12
Fix tacotron2 pretrained model

yangarbiter:
I found that there is some problem with loading tacotron2 pretrained model.
@mthrok should we fix it as seen in this PR, or should we remap the checkpoint and re-upload the checkpoint files?

### DIFF
--- a/torchaudio/prototype/tacotron2.py
+++ b/torchaudio/prototype/tacotron2.py
@@ -1219,5 +1219,7 @@ def tacotron2(checkpoint_name: str) -> Tacotron2:
     url, configs = _MODEL_CONFIG_AND_URLS[checkpoint_name]
     model = Tacotron2(**configs)
     state_dict = load_state_dict_from_url(url, progress=False)
-    model.load_state_dict(state_dict)
+    state_dict = state_dict['state_dict']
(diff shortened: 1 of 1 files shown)


torchaudio/prototype/tacotron2.py
```diff
@@ -1218,6 +1218,8 @@ def tacotron2(checkpoint_name: str) -> Tacotron2:
 
     url, configs = _MODEL_CONFIG_AND_URLS[checkpoint_name]
     model = Tacotron2(**configs)
-    state_dict = load_state_dict_from_url(url, progress=False)
-    model.load_state_dict(state_dict)
+    state_dict = load_state_dict_from_url(url, progress=False, map_location=torch.device('cpu'))
+    state_dict = state_dict['state_dict']
+    new_state_dict = {k.replace("module.", ""): v for k, v in state_dict.items()}
+    model.load_state_dict(new_state_dict)
```
nateanl, line 1224: Is there a parameter to check the parameters in the state_dict match the ones in the model?

yangarbiter: I think by default, `model.load_state_dict` will check, therefore the original version gives out errors.

yangarbiter:
Fixed by re-uploading the correct checkpoint file.

Issue #1651, opened 2021-07-29
Move the WaveRNN inference wrapper to infer method


yangarbiter:
As discussed in #1637, we want to support infer method for WavrRNN natively. To do so, we need to move 1. mulaw decoding, 2. batch folding, and 3. categorical sampling outside the forward method of the WaveRNNInferenceWrapper.

PR #1704, opened 2021-08-13
Refactor WaveRNN infer and move to codebase

yangarbiter:
As discussed in #1651 

The training part of the WaveRNN seems to support two loss functions -- CrossEntropy and MolLoss
(https://github.com/pytorch/audio/blob/main/examples/pipeline_wavernn/losses.py). Currently the inference code only supports the model trained with CrossEntropy since there is no pretrained model trained with MolLoss to test it out yet.

### DIFF
--- a/examples/pipeline_wavernn/inference.py
+++ b/examples/pipeline_wavernn/inference.py
@@ -22,8 +22,9 @@ def parse_args():
         help="If used, the model and inference function is jitted."
     )
     parser.add_argument(
-        "--loss", default="crossentropy", choices=["crossentropy"],
-        type=str, help="The type of loss the pretrained model is trained on.",
+        "--sampling-mode", default="multinomial", choices=["multinomial"],
+        type=str, help="The sampling mode used for inference. "
+                       "If the model is trained with cross entropy loss, 'multinomial' should be used.",
     )
     parser.add_argument(
         "--no-batch-inference", default=False, action="store_true",
@@ -39,11 +40,11 @@ def parse_args():
         help="Select the WaveRNN checkpoint."
     )
     parser.add_argument(
-        "--batch-timesteps", default=11000, type=int,
+        "--batch-timesteps", default=100, type=int,
         help="The time steps for each batch. Only used when batch inference is used",
     )
     parser.add_argument(
-        "--batch-overlap", default=550, type=int,
+        "--batch-overlap", default=5, type=int,
         help="The overlapping time steps between batches. Only used when batch inference is used",
     )
     args = parser.parse_args()
@@ -79,13 +80,13 @@ def main(args):
 
     with torch.no_grad():
         output = wavernn_inference_model(mel_specgram.to(device),
-                                         loss_name=args.loss,
+                                         sampling_mode=args.sampling_mode,
                                          mulaw=(not args.no_mulaw),
                                          batched=(not args.no_batch_inference),
                                          timesteps=args.batch_timesteps,
                                          overlap=args.batch_overlap,)
 
-    torchaudio.save(args.output_wav_path, output.reshape(1, -1), sample_rate=sample_rate)
+    torchaudio.save(args.output_wav_path, output, sample_rate=sample_rate)
 
 
 if __name__ == "__main__":
--- a/examples/pipeline_wavernn/inference.py
+++ b/examples/pipeline_wavernn/inference.py
@@ -22,11 +22,6 @@ def parse_args():
         help="If used, the model and inference function is jitted."
     )
     parser.add_argument(
-        "--sampling-mode", default="multinomial", choices=["multinomial"],
-        type=str, help="The sampling mode used for inference. "
-                       "If the model is trained with cross entropy loss, 'multinomial' should be used.",
-    )
-    parser.add_argument(
         "--no-batch-inference", default=False, action="store_true",
         help="Don't use batch inference."
     )
@@ -80,7 +75,6 @@ def main(args):
 
     with torch.no_grad():
         output = wavernn_inference_model(mel_specgram.to(device),
-                                         sampling_mode=args.sampling_mode,
                                          mulaw=(not args.no_mulaw),
                                          batched=(not args.no_batch_inference),
                                          timesteps=args.batch_timesteps,
--- a/examples/pipeline_wavernn/wavernn_inference_wrapper.py
+++ b/examples/pipeline_wavernn/wavernn_inference_wrapper.py
@@ -157,7 +157,6 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
 
     def forward(self,
                 specgram: Tensor,
-                sampling_mode: str = "multinomial",
                 mulaw: bool = True,
                 batched: bool = True,
                 timesteps: int = 100,
@@ -167,10 +166,11 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
         Based on the implementation from
         https://github.com/fatchord/WaveRNN/blob/master/models/fatchord_version.py.
 
+
+        Currently only supports multinomial sampling.
+
         Args:
             specgram (Tensor): spectrogram of size (n_mels, n_time)
-            sampling_mode (str): The sampling method used to generate the waveform.
-                Available `loss_name` includes `'multinomial'`. (Default; ``'multinomial```)
             mulaw (bool): Whether to perform mulaw decoding (Default: ``True``).
             batched (bool): Whether to perform batch prediction. Using batch prediction
                 will significantly increase the inference speed (Default: ``True``).
@@ -192,7 +192,7 @@ class WaveRNNInferenceWrapper(torch.nn.Module):
 
         n_bits = int(torch.log2(torch.ones(1) * self.wavernn_model.n_classes))
 
-        output = self.wavernn_model.infer(specgram, sampling_mode=sampling_mode).cpu()
+        output = self.wavernn_model.infer(specgram).cpu()
 
         if mulaw:
             output = normalized_waveform_to_bits(output, n_bits)
--- a/examples/pipeline_wavernn/wavernn_inference_wrapper.py
+++ b/examples/pipeline_wavernn/wavernn_inference_wrapper.py
@@ -21,18 +21,12 @@
 # *****************************************************************************
 
 
-from typing import List
-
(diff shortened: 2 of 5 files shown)


torchaudio/models/wavernn.py
```diff
@@ -347,6 +355,68 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+
+            logits = self.fc3(x)
+
+            posterior = F.softmax(logits, dim=1)
+
+            x = torch.multinomial(posterior, 1).float()
+            x = bits_to_normalized_waveform(x, n_bits)
+            output.append(x)
+
+        return torch.stack(output).transpose(0, 1).transpose(1, 2)
```
nateanl, line 418: nit: is (1, 2, 0) the final dimension for the output?
```suggestion
        return torch.stack(output).permute(1, 2, 0)
```

torchaudio/models/wavernn.py
```diff
@@ -347,6 +355,68 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
         # bring back channel dimension
         return x.unsqueeze(1)
 
+    @torch.jit.export
+    def infer(self, specgram: Tensor) -> Tensor:
+        r"""Inference method of WaveRNN.
+
+        This function currently only supports multinomial sampling, which assumes the
+        network is trained on crossentropy loss.
```
nateanl, line 363:
```suggestion
        network is trained on cross_entropy loss.
```

examples/pipeline_wavernn/wavernn_inference_wrapper.py
```diff
@@ -166,8 +160,8 @@ def forward(self,
                 loss_name: str = "crossentropy",
                 mulaw: bool = True,
                 batched: bool = True,
-                timesteps: int = 11000,
-                overlap: int = 550) -> Tensor:
+                timesteps: int = 100,
+                overlap: int = 5) -> Tensor:
```
nateanl, line 164: nit: update the default values in the doc.

Also could you update the ``shape`` term in the docs? So that we can keep it in consistent with CONTRIBUTING.md.

torchaudio/models/wavernn.py
```diff
@@ -347,6 +355,68 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+
+        Args:
+            specgram (Tensor): The input spectrogram to the WaveRNN with shape (n_batch, n_freq, n_time).
```
nateanl, line 366:
```suggestion
            specgram (Tensor): The input spectrogram to the WaveRNN of size (n_batch, n_freq, n_time).
```

torchaudio/models/wavernn.py
```diff
@@ -347,6 +355,68 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+
+        Return:
+            waveform (Tensor): The inferred waveform with shape (n_batch, 1, n_time).
```
nateanl, line 369:
```suggestion
            waveform (Tensor): The inferred waveform of size (n_batch, 1, n_time).
```

examples/pipeline_wavernn/wavernn_inference_wrapper.py
```diff
@@ -21,18 +21,12 @@
 # *****************************************************************************
 
 
-from typing import List
-
 from torchaudio.models.wavernn import WaveRNN
 import torch
-import torch.nn.functional as F
 import torchaudio
```
nateanl, line 26: Not sure if it follows Google's style check. I need your opinion on this to help me understand it.

```
import torch
from torch import Tensor
import torchaudio
from torchaudio.models.wavernn import WaveRNN
```
Or it's like 
```
import torch
import torchaudio
from torch import Tensor
from torchaudio.models.wavernn import WaveRNN
```

nateanl (approved):
LGTM overall. Just need to address the nits then we can merge it.

examples/pipeline_wavernn/inference.py
```diff
@@ -39,11 +39,11 @@ def parse_args():
         help="Select the WaveRNN checkpoint."
     )
     parser.add_argument(
-        "--batch-timesteps", default=11000, type=int,
+        "--batch-timesteps", default=100, type=int,
```
mthrok, line 42: What are the impacts of these changes?

torchaudio/models/wavernn.py
```diff
@@ -35,6 +36,13 @@
 }
 
 
+def bits_to_normalized_waveform(label: torch.Tensor, bits: int) -> torch.Tensor:
```
mthrok, line 39: Should this be public?

examples/pipeline_wavernn/inference.py
yangarbiter: When we want to use batch prediction, we would like to fold the signal on time domain first and then do the batch inference. This argument affects the folding length of the signal. For example:

original signal: [[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]]
if this value set to 5, it would be 
[[0, 1, 2, 3, 4]
 [5, 6, 7, 8, 9]]
if this value set to 3, it would be 
[[0, 1, 2],
 [3, 4, 5],
 [7, 8, 9]]

We need to make this change because in the original implementation, the upsampling is applied first and then we perform the folding (see [here](https://github.com/pytorch/audio/pull/1704/files#diff-9b05c2c6426592ffa5157949e3e853bb3eae33276b34e837e0bae9751e10adcdL191-L200)). However, to move the folding of the signal outside of the infer method, we would need to fold first and then do the upsampling. To fold before the upsampling, the number of signals will be a lot lesser.

I think this change of the implementation will affect the output waveform a little, but I couldn't hear the difference before and after the change.

examples/pipeline_wavernn/wavernn_inference_wrapper.py
yangarbiter: Thanks for the reminder, I've changed all `with shape` to `of size`.

torchaudio/models/wavernn.py
yangarbiter: No, it shouldn't. I've changed it to a private function [here](https://github.com/pytorch/audio/pull/1704/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R39-R43).

examples/pipeline_wavernn/wavernn_inference_wrapper.py
yangarbiter: Thanks for the reminder, I've changed all `with shape` to `of size`.

yangarbiter:
I've also made an additional change [here](https://github.com/pytorch/audio/pull/1704/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R359) so that in the future, we can easily support difference sampling mode. We may want this interface because the `pipeline_wavernn` supports the MolLoss and if a model is trained with that loss, it requires a different sampling method.

torchaudio/models/wavernn.py
yangarbiter: Fixed

torchaudio/models/wavernn.py
yangarbiter: Fixed

examples/pipeline_wavernn/wavernn_inference_wrapper.py
yangarbiter: This part I am not sure since it doesn't trigger flake8's warning.
@mthrok what is our style for sorting the imports?

torchaudio/models/wavernn.py
yangarbiter: Thanks for the suggestion, I've fixed it [here](https://github.com/pytorch/audio/pull/1704/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R425).

torchaudio/models/wavernn.py
yangarbiter: Thanks for the comment, I've fixed it [here](https://github.com/pytorch/audio/pull/1704/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R363)

examples/pipeline_wavernn/wavernn_inference_wrapper.py
mthrok: Currently, we do not have a style for sorting imports for historical reason.
So, we resorted to just make sure flake8 does not complain.

Probably it's a good time to start linting as a part of https://fburl.com/jjsby41j

torchaudio/models/wavernn.py
```diff
@@ -35,6 +36,13 @@
 }
 
 
+def _bits_to_normalized_waveform(label: torch.Tensor, bits: int) -> torch.Tensor:
```
mthrok, line 39: Since this function is used by only one place, why don't we hardcode it in the call site, with the docsrtings?

mthrok (approved):

torchaudio/models/wavernn.py
yangarbiter: Thanks for the suggestion, I've fixed it [here](https://github.com/pytorch/audio/pull/1704/files#diff-7f4ad259da182d3840fe4c75dd0c0b7b9fe30f31503df0eff87b1e5c9378cac4R412)

torchaudio/models/wavernn.py
```diff
@@ -347,6 +348,76 @@ def forward(self, waveform: Tensor, specgram: Tensor) -> Tensor:
...
+            x = x + h2[0]
+            x = torch.cat([x, a3_t], dim=1)
+            x = F.relu(self.fc1(x))
+
+            x = torch.cat([x, a4_t], dim=1)
+            x = F.relu(self.fc2(x))
+
+            logits = self.fc3(x)
+
+            if sampling_mode == "multinomial":
```
mthrok, line 407: Since, there is only one `sampling_mode`, we can get rid of it.

yangarbiter: I've removed the sampling mode option and opened an issue to keep track of the support for MOLLoss (#1721).

PR #1707, opened 2021-08-13
RNNT Loss remove null grads

carolineechen:
Previously, forward calls to `RNNTLoss` or `rnnt_loss` would fail with a `cannot call get_autograd_meta() on undefined tensor` error when `requires_grad=False`. This PR resolves this issue so that forward calls can be made even when the logits do not have `requires_grad=True`.

### DIFF
--- a/test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
+++ b/test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
@@ -27,6 +27,12 @@ class RNNTLossTest:
         loss = rnnt_loss(logits, targets, logit_lengths, target_lengths)
         loss.backward()
 
+    def test_basic_forward_no_grad(self):
+        rnnt_loss = RNNTLoss()
+        logits, targets, logit_lengths, target_lengths = get_basic_data(self.device)
+        logits.requires_grad_(False)
+        rnnt_loss(logits, targets, logit_lengths, target_lengths)
+
     def test_costs_and_gradients_B1_T2_U3_D5_fp32(self):
         data, ref_costs, ref_gradients = get_B1_T2_U3_D5_data(
             dtype=torch.float32,
--- a/torchaudio/csrc/rnnt/cpu/compute.cpp
+++ b/torchaudio/csrc/rnnt/cpu/compute.cpp
@@ -87,10 +87,7 @@ std::tuple<torch::Tensor, c10::optional<torch::Tensor>> compute(
   torch::Tensor costs = torch::empty(
       options.batchSize_ * options.nHypos_,
       torch::TensorOptions().device(logits.device()).dtype(logits.dtype()));
-  c10::optional<torch::Tensor> gradients = c10::nullopt;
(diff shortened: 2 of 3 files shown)


torchaudio/csrc/rnnt/cpu/compute.cpp
```diff
@@ -87,10 +87,7 @@ std::tuple<torch::Tensor, c10::optional<torch::Tensor>> compute(
   torch::Tensor costs = torch::empty(
       options.batchSize_ * options.nHypos_,
       torch::TensorOptions().device(logits.device()).dtype(logits.dtype()));
-  c10::optional<torch::Tensor> gradients = c10::nullopt;
-  if (logits.requires_grad()) {
-    gradients = torch::zeros_like(logits);
-  }
+  c10::optional<torch::Tensor> gradients = torch::zeros_like(logits);
```
mthrok, line 90: Here, Tensor with the same number of elements as `logits` is initialized, even when `requires_grad=False`.

Looking at all the cases, the storage of the tensor is passed to `Compute` function.
Can we handle  the case `c10::optional<torch::Tensor> gradients == nullopt` in `Compute` function, ideally, without allocating `gradient` tensor?

mthrok: It is eventually passed to https://github.com/pytorch/audio/blob/8a347b62cf5c907d2676bdc983354834e500a282/torchaudio/csrc/rnnt/cpu/cpu_kernels.h#L373

and it seems to me not easy to fix this function. So I think the current approach of quick fix is good enough. Maybe we can file the issue for the further improvement (handle `gradient==nullptr` in `ComputeGradient`) in the future.

mthrok (approved):

mthrok:
@carolineechen Could we have a simple test to make sure that passing Tensor with `requires_grad=False` to RNN-T loss works?

test/torchaudio_unittest/rnnt/rnnt_loss_impl.py
```diff
@@ -27,6 +27,12 @@ def test_basic_backward(self):
         loss = rnnt_loss(logits, targets, logit_lengths, target_lengths)
         loss.backward()
 
+    def test_basic_forward_no_grad(self):
```
mthrok, line 30: Can we give the description of the context?
I think for future maintainer, this test looks insignificant so it's better to give explanation why this test matters.
Something like

```
In early stage, `RNNTLoss` caused segmentation fault when `logits` have `requires_grad = False`.
This test makes sure that this segmentation fault is caught properly.

See also https://github.com/pytorch/audio/pull/1707
```

carolineechen: I've added a comment in the following PR [here](https://github.com/pytorch/audio/pull/1711/commits/59366cc616f4adced46c3eed11367023e5cbc0d3#diff-cf61df2ded7de3af992d120c6b1530bddb6333ad62e825ce4c9251dafe9afd1dR459)

PR #1711, opened 2021-08-17
Move RNNTL out of prototype

carolineechen:
Move RNNT Loss out of prototype folder to make it available as a beta feature
- move loss implementation into functional/transforms
- move loss unit tests into corresponding torchaudio test files
- refactor the utils/tests as appropriate (RNNT Loss is only valid for float32/16)
- set default to build RNNT and include description (previously not built unless explicitly stated)

Follow ups
- add subsections for transforms docs


carolineechen:
@mthrok do you have any thoughts on where to put `rnnt_loss` in the functional documentation or if we should create a new category? it doesn't seem to fit under the existing categories [here](https://pytorch.org/audio/main/functional.html)

mthrok:
> @mthrok do you have any thoughts on where to put `rnnt_loss` in the functional documentation or if we should create a new category? it doesn't seem to fit under the existing categories [here](https://pytorch.org/audio/main/functional.html)

You can create a new one.

docs/source/transforms.rst
```diff
@@ -171,6 +171,12 @@ Transforms are common audio transforms. They can be chained together using :clas
 
   .. automethod:: forward
 
+:hidden:`RNNTLoss`
```
mthrok, line 174: This makes me wonder if transforms should have subsections similar to functionals doc.

test/torchaudio_unittest/common_utils/__init__.py
```diff
@@ -31,6 +31,15 @@
     load_params,
     nested_params
 )
+from .rnnt_utils import (
```
mthrok, line 34: I have a mixed feeling about this imports. `rnnt_utils` is specific to RNN-T loss, but the APIs live in multiple modules (`functional` and `transforms`), so putting it in `common_utils` makes sense.

Looking at the individual functions imported bellow, their names are generic, so from the perspective of developer working on other modules, they are somewhat confusing.

Can we maybe just import `rnn_utils` here and individual tests access them as the following?

```python
from torchaudio_unittest.common_utils import rnn_utils

def test_foo(...):
    rnn_utils.get_basic_data(...)
```

docs/source/transforms.rst
nateanl: How about creating a loss category? We have Si-SDR and RNNTLoss so far, maybe we could add more under this subsection.

carolineechen: I've created a new `Loss` section in [functional docs](https://304212-90321822-gh.circle-artifacts.com/0/docs/functional.html) for `rnnt_loss` in this PR, but the [transforms docs](https://304212-90321822-gh.circle-artifacts.com/0/docs/transforms.html) is lacking these similar subsections and instead treats each transform independently. I think it makes sense to add subsections for transforms, especially since the number of transforms is growing, and we can follow up on this in a separate PR.

mthrok: Yes, we can add subsection in transforms doc in a separate PR.

mthrok: > How about creating a loss category? We have Si-SDR and RNNTLoss so far, maybe we could add more under this subsection.

I am not confident with the implementation so, if someone ports it, we need to double-check the implementation.

mthrok (approved):
Assuming that loss code and tests are moved properly, the changes look good.

test/torchaudio_unittest/common_utils/rnnt_utils.py
```diff
@@ -1,9 +1,174 @@
 import unittest
 import random
 import torch
-from torchaudio.prototype.rnnt_loss import RNNTLoss
-
-from .numpy_transducer import NumpyTransducerLoss
```
nateanl, line 6: The test code in [autograd_impl](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/rnnt/autograd_impl.py#L15) uses ``NumpyTransducerLoss``,  we should also update that.

test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
```diff
@@ -155,3 +155,22 @@ def test_PitchShift(self):
             T.PitchShift(sample_rate=sample_rate, n_steps=n_steps),
             waveform
         )
+
+
+class TransformsFloat32Only(TestBaseMixin):
+    def test_rnnt_loss(self):
+        def func(tensor):
+            return T.RNNTLoss(tensor, targets, logit_lengths, target_lengths)
```
nateanl, line 163: Is this method necessary?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -692,3 +692,21 @@ def func(tensor):
 
         tensor = torch.view_as_complex(torch.randn(2, 1025, 400, 2))
         self._assert_consistency_complex(func, tensor, test_paseudo_complex)
+
+
+class FunctionalFloat32Only(TestBaseMixin):
```
nateanl, line 697: Does the test in ``torchscript_consistency_impl`` also need ``FunctionalFloat32Only``?

carolineechen: I'm not too sure what you mean here, can you elaborate?

test/torchaudio_unittest/transforms/torchscript_consistency_impl.py
carolineechen: Good catch, will remove

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
nateanl: Yea I saw the tests in other file like auto_grad_impl use ``FunctionalFloat32Only``, so I'm wondering if we also add it here.

nateanl: Oops, It's the test itself. I'm confusing myself lol. Never mind.

nateanl (approved):

PR #1714, opened 2021-08-17
Move Tacotron2 out of prototype

yangarbiter:
Move Tacotron2 out of prototype folder to make it available as a beta feature

- move Tacotron2 into models
- correct imports from `torchaudio.prototype.tacotron2` to `torchaudio.models`


mthrok (commented):
The change looks good to me so far.
Are there any tests that has to be moved?

mthrok (approved):
LGTM

yangarbiter:
> The change looks good to me so far.
> Are there any tests that has to be moved?

The tests are already in the `test/torchaudio_unittest/models/` directory so no need to move it again.
Thanks.

PR #1716, opened 2021-08-17
Fix Tacotron2 inference example

yangarbiter:
- the `wavernn_vocode` called on line 327 is missing the `wavernn_checkpoint_name` argument, this PR fixes this
- also refactored the function call `griffin_lim_vocode` on line 338 so that it is more readable


mthrok (approved):
Cannot tell if this is the right signature from the PR but otherwise it looks good. Token approval.

yangarbiter:
> Cannot tell if this is the right signature from the PR but otherwise it looks good. Token approval.

Thanks for the reminder, I've rerun the speech generation script for all three vocoders and they worked.

Issue #1687, opened 2021-08-09
Detection of libsndfile


mthrok:
Currently, `torchaudio` treats `soundfile` as an optional dependency, and it assumes either `soundfile` is properly installed or it is not installed at all. However, there is a third state where `soundfile` is installed but the backing library `libsndfile` is not installed, and in this case, `import torchaudio` fails.

For regular users, the issue is easily resolved with installing `libsndfile`, but the error message is not intuitively clear and for users who were installing a package that depends on torchaudio while the environment has the incomplete installation of `soundfile`, this is not a great experience to debug.

```

  ...More stack trace here

  File "/***/lib/python3.8/site-packages/torchaudio/__init__.py", line 15, in <module>
    from torchaudio.backend import (
  File "/***/lib/python3.8/site-packages/torchaudio/backend/__init__.py", line 2, in <module>
    from . import utils
  File "/***/lib/python3.8/site-packages/torchaudio/backend/utils.py", line 7, in <module>
    from . import (
  File "/***/lib/python3.8/site-packages/torchaudio/backend/soundfile_backend.py", line 15, in <module>
    import soundfile
  File "/***/lib/python3.8/site-packages/soundfile.py", line 142, in <module>
    raise OSError('sndfile library not found')
OSError: sndfile library not found
```

We can improve this by actually performing `import soundfile` if `soundfile` is installed.

## Steps

1. In [module_utils.py](https://github.com/pytorch/audio/blob/main/torchaudio/_internal/module_utils.py), add variable `is_soundfile_available` and function `requires_soundfile` (Notice that we already have similar functions for other modules.)
2. Initialize the `is_soundfile_available` depending on the result of the following logic
  1. `is_module_available('soundfile')` returns `False` -> `False`
  2. If the step 1 succeeds, try `import soundfile` and if it succeed assign `True`, else `False`.
3. Implement `requires_soundfile` using `is_soundfile_available`, following the existing `requires_XXX` function.
3. In [`soundfile_backend.py`](https://github.com/pytorch/audio/blob/main/torchaudio/backend/soundfile_backend.py), replace `is_module_available("soundfile")` with `is_soundfile_available` and `requires_module("soundfile")` with `requires_soundfile`

## Test / repro

- In your conda environment, install `soundfile` then uninstall `libsndfile`.
- Check that the you get a stack trace similar to the above description.
- Implement the above

## Additional Context

https://github.com/huggingface/transformers/issues/13034

mthrok:
[...]

PR #1718, opened 2021-08-18
Add additional guard to libsndfile

bowangbj:
Summary:

Add additional check of libsndfile when soundfile is needed.
Address https://github.com/pytorch/audio/issues/1687
No functional change

Test steps:
1. Setup
Install soundfile: $ pip install soundfile
delete libsndfile: $ conda remove libsndfile

2. Repro the error
(audio) [bowangbj@devgpu006.pnb1 ~/local/audio] python
Python 3.9.6 (default, Jul 30 2021, 16:35:19) 
[GCC 7.5.0] :: Anaconda, Inc. on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import torchaudio
/data/users/bowangbj/audio/torchaudio/_internal/module_utils.py:87: UserWarning: Failed to import soundfile. 'soundfile' backend is not available.
  warnings.warn("Failed to import soundfile. 'soundfile' backend is not available.")
$conda install -c conda-forge libsndfile

3. Fix the error
Follow error msg above to install libsndfile, and import torchaudio, it will succeed.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -76,6 +76,30 @@ def requires_kaldi():
             return wrapped
     return decorator
 
+def is_soundfile_available():
+    if not is_module_available('soundfile'): return False
+    try:
+        import soundfile
+    except OSError as os_error:
+        if str(os_error).find("sndfile library not found") != -1:
+            raise RuntimeError("""soundfile requires libsndfile to be installed. Try install it via:
+$conda install -c conda-forge libsndfile """)
+            return False
+        else:
+            raise os_error
+    return True
+
+def requires_soundfile():
+    if is_soundfile_available():
+        def decorator(func):
+            return func
+    else:
+        def decorator(func):
+            @wraps(func)
+            def wrapped(*args, **kwargs):
+                raise RuntimeError(f'{func.__module__}.{func.__name__} requires soundfile')
+            return wrapped
+    return decorator
 
 def is_sox_available():
     return is_module_available('torchaudio._torchaudio') and torch.ops.torchaudio.is_sox_available()
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -80,22 +80,19 @@ def requires_kaldi():
 def _check_soundfile_importable():
     if not is_module_available('soundfile'):
         return False
-    else:
-        try:
-            import soundfile    # noqa: F401
-            return True
-        except OSError:
-            raise RuntimeError("""Failed to import soundfile, most likely it's
-because the required dependency libsndfile not installed yet, try install it:
-$conda install -c conda-forge libsndfile """)
-            return False
+    try:
+        import soundfile    # noqa: F401
+        return True
+    except Exception:
+        warnings.warn("Failed to import soundfile. 'soundfile' backend is not available.")
+        return False
 
 
 _is_soundfile_importable = _check_soundfile_importable()
 
 
 def is_soundfile_available():
-    return is_module_available('soundfile') and _is_soundfile_importable
+    return _is_soundfile_importable
 
 
 def requires_soundfile():
--- a/torchaudio/_internal/module_utils.py
+++ b/torchaudio/_internal/module_utils.py
@@ -95,7 +95,7 @@ _is_soundfile_importable = _check_soundfile_importable()
 
 
 def is_soundfile_available():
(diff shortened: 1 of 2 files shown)


torchaudio/_internal/module_utils.py
```diff
@@ -76,6 +76,30 @@ def wrapped(*args, **kwargs):
             return wrapped
     return decorator
 
+def is_soundfile_available():
+    if not is_module_available('soundfile'): return False
+    try:
+        import soundfile
+    except OSError as os_error:
+        if str(os_error).find("sndfile library not found") != -1:
```
bowangbj, line 84: Note os_error.error_no is None thus I was trying to use string match to check. pls let me know in case there is a better approach

mthrok: I think we can simply catch `Exception` and we do not need to check the detail of the exception.
The reason is that, from the perspective of `torchaudio` library, we only want to know if `soundfile` could be imported without error or not. If it was imported, then we enable some features in `torchaudio` and otherwise disable the features. We do not want to propagate the exception caught during `import soundfile` (we can add a warning if we want though), because for `torchaudio` it's optional dependency, and otherwise `import torchaudio` will stop working.

Because of the above reasons, it is not necessary for us to inspect the reason `import soundfile` failed. If `import soundfile` is necessary for client code, then we can expect that client code will run into a issue after `import torchaudio`, either they try to do `import soundfile` by themselves or try to use `torchaudio`'s functionality that requires `soundfile`. So it will not be that inconvenient for users even if we do not check what kind of exception was raised during `import soundfile`.

torchaudio/_internal/module_utils.py
```diff
@@ -77,6 +77,34 @@ def wrapped(*args, **kwargs):
     return decorator
 
 
+def is_soundfile_available():
```
mthrok, line 80: Since this function will be called multiple times, I think it's better to avoid repeating `try:import~except:~` in it.

I think something like the following should do.

```python
def _check_soundfile_importable():
    try:
        import soundfile
        return True
    except Exception:
        return False


_is_soundfile_importable = _check_soundfile_importable()


def is_soundfile_available():
    return is_module_available('soundfile') and _is_soundfile_importable
```

mthrok (commented):
Hi @bowangbj 

Thanks for the PR.

torchaudio/_internal/module_utils.py
bowangbj: Great point, modified the code to take OSError only, also updated error msg a little bit to make it precise.

torchaudio/_internal/module_utils.py
bowangbj: Great point, updated, I still need to check module_available in _check_soundfile_importable b/c we wanted to avoid import if module not exist.

bowangbj:
> Hi @bowangbj
> 
> Thanks for the PR.

Thanks Moto for your super quick review, resolved all your cmts, PTAL

torchaudio/_internal/module_utils.py
```diff
@@ -77,6 +77,40 @@ def wrapped(*args, **kwargs):
     return decorator
 
 
+def _check_soundfile_importable():
+    if not is_module_available('soundfile'):
+        return False
+    else:
```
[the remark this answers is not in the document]
bowangbj: Good point. removed.

torchaudio/_internal/module_utils.py
```diff
@@ -77,6 +77,40 @@ def wrapped(*args, **kwargs):
...
+        try:
+            import soundfile    # noqa: F401
+            return True
+        except OSError:
+            raise RuntimeError("""Failed to import soundfile, most likely it's
+because the required dependency libsndfile not installed yet, try install it:
+$conda install -c conda-forge libsndfile """)
```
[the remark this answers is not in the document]
bowangbj: Great point, updated as suggested

bowangbj (commented):
Thanks Moto for the super quick and great review. Resolved all your comments, PTAL

Nit: you can find the diffs (from prev version by selecting the last commits like https://github.com/pytorch/audio/pull/1718/commits/76c25b40b66252e34d8a996c336b25dc53814f43 ) :)

mthrok (approved):
LGTM. Thanks!

PR #1720, opened 2021-08-20
Add sections to transforms docs

carolineechen:
Following discussion from https://github.com/pytorch/audio/pull/1711#discussion_r691372056, add subsections in pytorch docs for transforms similar to functional

### DIFF
--- a/docs/source/functional.rst
+++ b/docs/source/functional.rst
@@ -197,12 +197,11 @@ treble_biquad
 
 .. autofunction:: treble_biquad
 
-
-vad
----
-
 :hidden:`Feature Extractions`
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


mthrok (approved):
LGTM

I feel like `compute_deltas` can be in a utility section, like MelScale conversion, but at the same time quick search finds some article describes delta as feature. https://wiki.aalto.fi/display/ITSP/Deltas+and+Delta-deltas

nateanl:
deltas and delta-deltas can be used as features to acoustic model training. For example: https://github.com/kaldi-asr/kaldi/blob/cbed4ff688a172a7f765493d24771c1bd57dcd20/egs/wsj/s5/steps/train_mono.sh#L64

nateanl (approved):

PR #1729, opened 2021-08-27
[ROCM] fix build error

micmelesse:
fixed a bug where USE_CUDA path was being activated on ROCM


micmelesse:
@malfet I addressed your comments and made sure that use_cuda flag is not set on rocm dockers to fix the build errors

micmelesse:
@malfet I am not sure why the tests are stalling. It has been a few days.

mthrok:
Build and tests are all green. I guess the web hook was not handled properly between CircleCI and GitHub.

https://app.circleci.com/pipelines/github/pytorch/audio/7161/workflows/b6bbb273-5f7a-41a1-ae88-b9b0bf5b899c
https://app.circleci.com/pipelines/github/pytorch/audio/7161/workflows/22d3f7f3-6f96-407a-84ca-d65cd710eb83

mthrok:
@malfet It looks to me this is ready to merge.

micmelesse:
@malfet  ping

CMakeLists.txt
```diff
@@ -60,6 +53,21 @@ option(BUILD_RNNT "Enable RNN transducer" ON)
...
+# check that USE_CUDA and USE_ROCM are not set at the same time
+if(USE_CUDA AND USE_ROCM)
+  message(FATAL "CUDA and ROCm are mutually exclusive")
+endif()
+
+if(USE_ROCM) 
+  # Find the HIP package, set the HIP paths, load the HIP CMake.
+  include(cmake/LoadHIP.cmake)
+  if(NOT PYTORCH_FOUND_HIP)
+    set(USE_ROCM OFF)
```
malfet, line 68: Nit, don't you want to raise an error here? Similarly, how TorchAudio should not compile if USE_CUDA is true, but CUDA could not be found?

malfet (approved):

CMakeLists.txt
mthrok: @carolineechen Can you address this? Let's make sure that build fails if CUDA is unavailable while `USE_CUDA=1`

carolineechen: @mthrok @malfet the default behavior without setting `USE_CUDA` is whether or not CUDA is available, so I believe one of the reasons for this variable is to be able to force CUDA during the build even if it can not be found. This way, we can still run code with CUDA functionalities at runtime if CUDA is available then, otherwise it falls back to CPU behavior anyways. Let me know if you have any thoughts on this or if this shouldn’t be the case

mthrok: @carolineechen I think there are multiple layers of logics to handle CUDA build.

1. to successfully build CUDA, one needs CUDA development environment and PyTorch binary with CUDA-capabilities.
2. User set `USE_CUDA` environment variable when calling `python setup.py build_ext`
3. There is a default value for `USE_CUDA`, which is dynamically determined by the environment configuration. (noted in 1)

Considering these, the following is what I think is the expected build behavior


| `USE_CUDA` env var | CUDA dev env |     Build result    |
|:------------------:|:------------:|:-------------------:|
|        unset       |      N/A     |  Success (CPU only) |
|        unset       |   Available  | Success (with CUDA) |
|        False       |      N/A     |  Success (CPU only) |
|        False       |   Available  |  Success (CPU only) |
|        True        |      N/A     |         **Fail**        | 
|        True        |   Available  | Success (with CUDA) |

What @malfet is proposing is to make sure to fail if the user requested to build with CUDA-capability but the build process cannot find the necessary tools.

mthrok: IIUC, `torch.cuda.is_available()` can only detect the runtime environment. So maybe we should not use the dynamic default when `USE_CUDA` is not set.

micmelesse: @malfet  Sorry for the late reply. I will add that in a pr.

mthrok: @malfet We looked at this and I think the current way is fine.

Our build process uses `find_package(Torch required)`, and I believe it will fail if PyTorch has CUDA capability but no CUDA development environment is available.

PR #1735, opened 2021-08-30
Cleanup meta.yaml

malfet:
Neither cmake nor setup tools are needed to run torchaudio
And `cpuonly` package is not needed to run tests


malfet:
Actually, it's a bit bizzare, but they are...

PR #1737, opened 2021-08-31
Fix CUDA build logic for _torchaudio.so

malfet:
It's wrong to depend on `${TORCH_LIBRARIES}` as it pulls in explicit
`libcuda.so.1` dependency, which violates the assumption that GPU
accelerated libraries should be loadable with no NVIDIA drivers installed

Instead, make it depend on `torch` target, which includes all necessary
Torch C++ API dependences


mthrok (approved):
Stamp

cc @seemethere

PR #1740, opened 2021-08-31
Fix WaveRNN training example

yangarbiter:
We saw that in [this commit](https://github.com/pytorch/audio/commit/e061b268d284a6f43acc44314b83ec6ffe58e91a#diff-a473ab6313da91f82a155edc91015cd96b4867cff38d9dd63bf75dfd411281b3), we ported mel spectrogram from librosa to torchaudio. However this port misses a parameter norm.

[This internal link](https://colab.research.google.com/drive/18d1xOp3x4EEgsa_OIZRYaKB7k_Mig12m?usp=sharing) shows that by adding the `norm='slaney'` parameter to `torchaudio.transforms.MelSpectrogram`, we are able to maintain the original transform (`torchaudio.transforms.Spectrogram` +  `LinearToMel`).

### DIFF
--- a/examples/pipeline_wavernn/main.py
+++ b/examples/pipeline_wavernn/main.py
@@ -272,6 +272,7 @@ def main(args):
             n_mels=args.n_freq,
             f_min=args.f_min,
             mel_scale='slaney',
+            norm='slaney',
             **melkwargs,
         ),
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Token approval.

PR #1744, opened 2021-09-01
Put output tensor on proper device in `get_whitenoise()`

jayleverett:
Put the random audio data generated by `get_whitenoise()` utility function onto the proper device, rather than defaulting to cpu.

Additionally, fix any test failures that arise from this change.


jayleverett:
Intent is to trigger Circle CI tests so that I can see which ones will fail. It's expected that there will be failing tests initially

hwangjeff (approved):

PR #1745, opened 2021-09-01
Update the version of fairseq used for testing model

mthrok:
As a preparation for the HuBERT model integration.


nateanl (approved):
LGTM as the wave2vec2.0 tests all pass based on the new fairseq commit.

carolineechen (approved):

PR #1752, opened 2021-09-02
Split extension into non-Python and Python wrapper libraries

mthrok:
This PR splits the extension module (`_torchaudio.so`) into non-Python part (`libtorchaudio.so`) and Python wrapper around it (`_torchaudio.so`).

Overview of changes
- cmake always builds libtorchaudio (the core implementation)
  - `BUILD_LIBTORCHAUDIO` option is removed from CMake.

The new assumptions around the library discoverability
- In regular OSS workflow (`pip`/`conda`-based binary installation), both `libtorchaudio.so` and `_torchaudio.so` are present.
    In this case,`libtorchaudio.so` has to be loaded manually with `torch.ops.load_library` and/or `torch.classes.load_library` otherwise importing `_torchaudio` would not be able to resolve the symbols defined in `libtorchaudio.so`.
- When `torchaudio` is deployed with PEX format (single zip file)
  - We expect that`libtorchaudio.so` exists as a file in some search path configured by client code.
  - `_torchaudio.so` is still importable and because we do not know where `libtorchaudio.so` will exist, we will let the dynamic loader resolve the dependency from `_torchaudio.so` to `libtorchaudio`, which should work as long as `libtorchaudio.so` is in a library search path (search path is not modifiable from already-running Python process).

### DIFF
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -43,7 +43,10 @@ _TORCH_CUDA_ARCH_LIST = os.environ.get('TORCH_CUDA_ARCH_LIST', None)
 
 
 def get_ext_modules():
-    return [Extension(name='torchaudio._torchaudio', sources=[])]
+    return [
+        Extension(name='torchaudio.libtorchaudio', sources=[]),
+        Extension(name='torchaudio._torchaudio', sources=[]),
+    ]
 
 
 # Based off of
@@ -57,6 +60,10 @@ class CMakeBuild(build_ext):
         super().run()
 
     def build_extension(self, ext):
+        if ext.name == 'torchaudio.libtorchaudio':
+            # libtorchaudio is built as a part of _torchaudio
+            return
+
         extdir = os.path.abspath(
             os.path.dirname(self.get_ext_fullpath(ext.name)))
 
@@ -76,7 +83,7 @@ class CMakeBuild(build_ext):
             f"-DBUILD_KALDI:BOOL={'ON' if _BUILD_KALDI else 'OFF'}",
             f"-DBUILD_RNNT:BOOL={'ON' if _BUILD_RNNT else 'OFF'}",
             "-DBUILD_TORCHAUDIO_PYTHON_EXTENSION:BOOL=ON",
-            "-DBUILD_LIBTORCHAUDIO:BOOL=OFF",
+            "-DBUILD_LIBTORCHAUDIO:BOOL=ON",
             f"-DUSE_ROCM:BOOL={'ON' if _USE_ROCM else 'OFF'}",
             f"-DUSE_CUDA:BOOL={'ON' if _USE_CUDA else 'OFF'}",
         ]
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -83,7 +83,6 @@ class CMakeBuild(build_ext):
             f"-DBUILD_KALDI:BOOL={'ON' if _BUILD_KALDI else 'OFF'}",
             f"-DBUILD_RNNT:BOOL={'ON' if _BUILD_RNNT else 'OFF'}",
             "-DBUILD_TORCHAUDIO_PYTHON_EXTENSION:BOOL=ON",
-            "-DBUILD_LIBTORCHAUDIO:BOOL=ON",
             f"-DUSE_ROCM:BOOL={'ON' if _USE_ROCM else 'OFF'}",
             f"-DUSE_CUDA:BOOL={'ON' if _USE_CUDA else 'OFF'}",
         ]
--- a/build_tools/setup_helpers/extension.py
+++ b/build_tools/setup_helpers/extension.py
@@ -56,12 +56,17 @@ class CMakeBuild(build_ext):
         try:
             subprocess.check_output(['cmake', '--version'])
         except OSError:
-            raise RuntimeError("CMake is not available.")
+            raise RuntimeError("CMake is not available.") from None
         super().run()
 
     def build_extension(self, ext):
-        if ext.name == 'torchaudio.libtorchaudio':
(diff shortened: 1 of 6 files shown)


build_tools/setup_helpers/extension.py
```diff
@@ -43,7 +43,10 @@ def _get_build(var, default=False):
 
 
 def get_ext_modules():
-    return [Extension(name='torchaudio._torchaudio', sources=[])]
+    return [
+        Extension(name='torchaudio.libtorchaudio', sources=[]),
```
carolineechen, line 47: are there situations in which only the `libtorchaudio` C++ extension will be needed instead of both? I'm trying to understand if there's a reason why we're splitting the extension into separate C++ and Python components or if this is simply a refactor.

torchaudio/csrc/CMakeLists.txt
```diff
@@ -119,31 +137,12 @@ if (BUILD_TORCHAUDIO_PYTHON_EXTENSION)
     set_target_properties(_torchaudio PROPERTIES LINK_FLAGS "-undefined dynamic_lookup")
   endif()
 
-  if (BUILD_SOX)
```
carolineechen, line 122: I don't think this really affects the end result, but just curious why several components originally under the `_torchaudio.so` section are being moved to the `libtorchaudio` build

carolineechen (commented):
overall looks like the code changes are fine/preserve correctness, but had a few questions

torchaudio/csrc/CMakeLists.txt
mthrok: Now `libtorchaudio` contains the implementation related to those definitions, such as (CUDA, Kaldi, libsox except file-like object I/O) meanwhile `_libtorchaudio` only contains libsox file-like object I/O. So these compiler definitions (`USE_CUDA`, `INCLUDE_KALDI` and `INCLUDE_SOX`) need to be defined in `libtorchaudio`. (and propagated to `_torchaudio` via `PUBLIC` keyword so that we only need to set them in one place).

Now that I think of it, probably, it's better to prefix these compiler definitions with `TORCHAUDIO_`.

build_tools/setup_helpers/extension.py
mthrok: > are there situations in which only the libtorchaudio C++ extension will be needed instead of both?

In C++ examples and mobile deployment, where they do not want to bring the whole Python library.

>  I'm trying to understand if there's a reason why we're splitting the extension into separate C++ and Python components or if this is simply a refactor.

It's somewhat of both but aiming for distributing `libtorchaudio` as a part of our Python package binary, so that integration with C++ code become easier, similar to PyTorch core does. Right now, when one wants to run torchaudio's custom implementation in C++ via TorchScript, they have to build `libtorchaudio` from scratch. By splitting the library and extracting the Python portion to `_torchaudio`, one can simply reuse the `libtorchaudio`.
Ideally, we should also provide `findTorchAudio.cmake` which is analogous to PyTorch core's `torch.utils.cmake_prefix_path`, to provide the same experience as PyTorch core, but that's a future work.

torchaudio/csrc/CMakeLists.txt
```diff
@@ -155,20 +154,11 @@ if (BUILD_TORCHAUDIO_PYTHON_EXTENSION)
 
   target_link_libraries(
     _torchaudio
-    torch
```
nateanl, line 158: Is it okay to remove pytorch?

build_tools/setup_helpers/extension.py
carolineechen: from this PR implementation, it looks like `libtorchaudio` can not be built as a standalone but only by default as part of the `_torchaudio` Python extension (see [here](https://github.com/pytorch/audio/pull/1752/files#diff-dc5272c8001918aecaca882d26285022fa5179d9fd7c2268805a263525269179R63)), so for the first point (C++/mobile), is it possible to only bring in the C++ extension without the Python extension?

torchaudio/csrc/CMakeLists.txt
carolineechen: `torch` is now linked to `libtorchaudio` from this [change](https://github.com/pytorch/audio/pull/1752/files#diff-12a57bc8e7ba6a5dd32aa566c805dd6c6514f46d26d5f53d0bf4400d557eaecbR67) so this should be okay

build_tools/setup_helpers/extension.py
mthrok: `libtorchaudio` is always built. Check out the resulting [CMakeLists.txt](https://github.com/mthrok/audio/blob/split-lib/torchaudio/csrc/CMakeLists.txt) and notice that the libtorchaudio portion is not guarded by any if clause. On contrary, `_torchaudio.so` is built only when `BUILD_TORCHAUDIO_PYTHON_EXTENSION=1`, which is set by `setup.py`.

torchaudio/csrc/CMakeLists.txt
mthrok: Yup so the library dependency graph looks like the following

Untitled

Not sure if we need to explicitly add `libtorch` and `libsox` to `_torchaudio` in this case, because they will be handled either way.

build_tools/setup_helpers/extension.py
```diff
@@ -57,6 +60,10 @@ def run(self):
         super().run()
 
     def build_extension(self, ext):
+        if ext.name == 'torchaudio.libtorchaudio':
+            # libtorchaudio is built as a part of _torchaudio
```
carolineechen, line 64: per offline discussion, improve comment to indicate that the main purpose of this if statement is to avoid building extension multiple times (as opposed to specifically not building it when called with `libtorchaudio`)

carolineechen (approved):

mthrok:
thanks!

PR #1755, opened 2021-09-07
Update the way to access libsox global config

mthrok:
Preparation for splitting `libtorchaudio` and `_torchaudio`.

When two libraries are compiled separately, and each code does
`#include <sox.h>` independently, two copies of libsox's static global
variables (`sox_globals_t`) are created.

Our code should be referring to the same instance. To achieve this,
`_torchaudio` should be accessing the global variable defined in
`libtorchaudio` via the custom utility functions, and it should not
directly use `sox_get_globals`.

### DIFF
--- a/torchaudio/csrc/pybind/sox/effects.cpp
+++ b/torchaudio/csrc/pybind/sox/effects.cpp
@@ -48,14 +48,16 @@ std::tuple<torch::Tensor, int64_t> apply_effects_fileobj(
   // the whole audio data are fetched.
   //
   // * This can be changed with `torchaudio.utils.sox_utils.set_buffer_size`.
-  const auto capacity = [](&){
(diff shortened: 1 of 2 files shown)


mthrok:
@hwangjeff 

In https://github.com/pytorch/audio/pull/1752, the vorbis large header test was failing.

https://app.circleci.com/pipelines/github/pytorch/audio/7223/workflows/8ad268f4-b970-4128-87b8-60c7af957cca/jobs/320861

After the investigation, it turned out that when splitting the compilation of `_torchaudio` into `libtorchaudio` and `_torchaudio`, the direct use of `sox_get_globals` cause the problem.

hwangjeff (approved):

carolineechen (approved):

PR #1756, opened 2021-09-08
Simplify code using torch.normal

datumbox:
Should be covered by `test_dither_GPDF()`


datumbox:
requires more work and better understanding of the algo that I have to make the chance, I'll close this for now.

mthrok:
@datumbox What was the issue with your approach?

datumbox:
The tests were failing but it's unclear to me why as I  don't know enough about the specific technique. It's worth looking at the test execution to see the error and tell me what you think

Issue #1754, opened 2021-09-07
Add a training recipe of ConvTasNet on LibriMix dataset and provide pre-trained weights.


nateanl:
## Motivation
In ``torchaudio.models``, we supported ConvTasNet, a convolutional network for speech separation that outperforms the ideal ratio masks. However, the pre-trained weights are missing that makes it difficult to validate the performance of the model.

Thanks to [Asteroid](https://github.com/asteroid-team/asteroid), we have plenty of speech separation models and utility functions to reproduce the results🚀 🚀 . It will be great to add a training recipe for ConvTasNet into torchaudio so that we make sure the model structure is good to use. Also users will benefit from it by: 1) using the pre-trained weights directly provided by torchaudio; 2) use the training recipe as a baseline for their research. 

## Pitch

- Add a LibriMix dataset. Different from wsj0-mix, LibriMix is based on LibriSpeech and it's purely open-source. It'll attract more researchers to develop their methods and compare with the ConvTasNet baseline.
- Add a training recipe for ConvTasNet. This includes a PyTorch-Lightning trainer, a training pipeline and an evaluation step.
- After we achieves an identical performance as Asteroid, we can provide a pre-trained weight to let users directly use it instead of training the model from scratch.

Again, shout out to Asteroid people @mpariente @popcornell @JorisCos @jonashaag for building an amazing library for speech separation and enhancement!

nateanl:
Finished by #1757 #2411. #2440, and #2675

PR #1757, opened 2021-09-08
[WIP] Add a training recipe for ConvTasNet on LibriMix dataset.

nateanl:
The PR addresses #1754 
Summary:
- Add a LibriMix dataset.
- Add a PyTorch Lightning trainer for ConvTasNet.
- Add a training pipeline to reproduce the results.

TODO:
- Put the Si-SDR and SDR improvements to the results.
- Provide the pre-trained weights.
- Add ``torch.hub`` API.


nateanl:
I will create another PR for loading pre-trained weights into ConvTasNet

examples/source_separation/utils/dataset/librimix.py
```diff
@@ -0,0 +1,73 @@
...
+
+import torch
+from torch.utils.data import Dataset
+
+import torchaudio
+
+SampleType = Tuple[int, torch.Tensor, List[torch.Tensor]]
+
+
+class LibriMix(Dataset):
```
carolineechen, line 12: This dataset is also used in some superb tasks and I think would be worth supporting in `torchaudio/datasets` instead of the `examples` folder, where we can create a wrapper class in the examples directory if necessary for running source separation. Thoughts on this? This could be a follow up that doesn't need to be addressed in this PR

nateanl: Yeah sounds great. Let's make a list of the datasets we want to add for s3prl and have a team discussion?

torchaudio/models/conv_tasnet.py
```diff
@@ -138,6 +139,10 @@ def __init__(
         self.output_conv = torch.nn.Conv1d(
             in_channels=num_feats, out_channels=input_dim * num_sources, kernel_size=1,
         )
+        if msk_activate == "sigmoid":
+            self.mask_activate = torch.nn.Sigmoid()
+        elif msk_activate == "relu":
+            self.mask_activate = torch.nn.ReLU()
```
mthrok, line 145: Could you add `else` clause so that it will fail here if there is typo?

examples/source_separation/utils/dataset/librimix.py
```diff
@@ -0,0 +1,73 @@
...
+    r"""Create the LibriMix dataset.
+
+    Args:
+        root (str or Path): the path to the directory where the dataset is stored.
+        num_speakers (int, optional): The number of speakers, which determines the directories
+            to traverse. The Dataset will traverse ``s1`` to ``sN`` directories to collect
+            N source audios. (Default: 2)
+        sample_rate (int, optional): sample rate of audio files. If any of the audio has a
+            different sample rate, raises ``ValueError``. (Default: 8000)
```
mthrok, line 21: I know this is coming from my implementation of WSJ dataset, but if we are moving this implementation to the library, then maybe we should change this behavior to resample to the target sample rate, or let the client code resample it. 

The reason why I took this approach back then was that `torchaudio.transform.Resample` was so slow, and WSJ0Mix dataset generates sample rate of 8k and 16k Hz. So I assumed that just sticking to 8k and 16k Hz should be fine.

I think it's fine now.

examples/source_separation/conv_tasnet/README.md
```diff
@@ -13,15 +13,15 @@ For the usage, please checkout the [source separation README](../README.md).
 The default training/model configurations follow the best non-causal implementation from the paper. (causal configuration is not implemented.)
```
mthrok, line 13: IIRC, the following configurations are from paper setting. Since you are changing it, can you update the above statement as well?

examples/source_separation/eval.py
```diff
@@ -0,0 +1,102 @@
...
+            sdr, sir, sar, _ = mir_eval.separation.bss_eval_sources(src[0], est[0])
+            sdr_mix, sir_mix, sar_mix, _ = mir_eval.separation.bss_eval_sources(src[0], mix[0])
+            results += torch.tensor([
+                sdr.mean() - sdr_mix.mean(),
+                sisdri,
+                sir.mean() - sir_mix.mean(),
+                sar.mean() - sar_mix.mean()
+            ])
+    results /= len(data_loader)
+    print("SDR improvement: ", results[0].item())
```
mthrok, line 29: Can you use `logging` module for logging?

mthrok: oh this is eval, then it's okay if it does not take a long time.

examples/source_separation/utils/dataset/librimix.py
nateanl: Yep LibriMix also has both 8k and 16k Hz versions. Most of the papers use the 8k Hz dataset. So I think it's fine.

examples/source_separation/utils/dataset/librimix.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.files = [p.name for p in self.mix_dir.glob("*wav")]
+        self.files.sort()
+
+    def _load_audio(self, path) -> torch.Tensor:
+        waveform, sample_rate = torchaudio.load(path)
+        if sample_rate != self.sample_rate:
+            raise ValueError(
+                f"The dataset contains audio file of sample rate {sample_rate}. "
+                "Where the requested sample rate is {self.sample_rate}."
```
hwangjeff, line 47: f-string; also maybe swapping in "but" for "where" would clarify the problem, e.g.

```suggestion
                f", but the requested sample rate is {self.sample_rate}."
```

examples/source_separation/README.md
```diff
@@ -8,105 +8,23 @@ This directory contains reference implementations for source separations. For th
 
 ### Overview
 
-To training a model, you can use [`train.py`](./train.py). This script takes the form of
-`train.py [parameters for distributed training] -- [parameters for model/training]`
+To training a model, you can use [`lightning_train.py`](./lightning_train.py). This script takes the form of
+`ligntning_train.py [parameters]`
```
hwangjeff, line 12:
```suggestion
`lightning_train.py [parameters]`
```

"h" and "n" do look rather similar...

there are other instances of this below

torchaudio/models/conv_tasnet.py
```diff
@@ -102,6 +102,7 @@ def __init__(
         num_hidden: int,
         num_layers: int,
         num_stacks: int,
+        msk_activate: str,
```
hwangjeff, line 105: how about calling this `mask_activate` to match the attribute name below and adding a comment to the args list above?

torchaudio/models/conv_tasnet.py
```diff
@@ -194,6 +201,7 @@ def __init__(
         msk_num_hidden_feats: int = 512,
         msk_num_layers: int = 8,
         msk_num_stacks: int = 3,
+        msk_activate: str = "sigmoid",
```
hwangjeff, line 204: and a comment for this one as well

torchaudio/models/conv_tasnet.py
nateanl: This aligns with other ``msk_xx`` arguments. If we replace all arguments then it'll be BC-Breaking. Let's fix it so far and make another PR for renaming.

mthrok (approved):

PR #1768, opened 2021-09-15
Add MVDR beamforming tutorial to example directory

nateanl:
The tutorial shows how to apply MVDR module on multi-channel audios by using either single-channel or multi-channel Time-Frequency masks.

Check the [Colab demo](https://colab.research.google.com/github/nateanl/audio/blob/mvdr/examples/beamforming/MVDR_tutorial.ipynb)


mthrok:
BTW, just to double-check, the speech audio example in use, it is not anything offensive or controversial, right?

examples/beamforming/MVDR_tutorial.ipynb
```diff
@@ -0,0 +1,576 @@
...
+        "-----------\n",
+        "\n",
+        "The multi-channel audio example is selected from [ConferencingSpeech](https://github.com/ConferencingSpeech/ConferencingSpeech2021) dataset. \n",
+        "\n",
+        "```\n",
+        "original filename: SSB07200001\\#noise-sound-bible-0038\\#7.86_6.16_3.00_3.14_4.84_134.5285_191.7899_0.4735\\#15217\\#25.16333303751458\\#0.2101221178590021.wav\n",
+        "```\n",
+        "\n",
+        "Note:\n",
+        "- You need to use the nightly torchaudio in order to use the MVDR and InverseSpectorgram transform module.\n",
```
carolineechen, line 53: nit: `InverseSpectorgram` spelling

carolineechen (commented):
+1 on Moto's comment on the input audio, or if it's possible to include in torchaudio so we would not need the additional github dependency

examples/beamforming/MVDR_tutorial.ipynb
nateanl: Nice catch. I will correct it.

nateanl:
The dataset is under Apache 2.0 license, I agree with Caroline, we can put the wavs to torchaudio. Some questions: 1) The wav files are ~3MB, it doesn't affect much on downloading the repo, right? 2) Should we also include a separate License in the directory? 

> The documentation only list the available methods. Can we have simple description on how each solution solves the problem?

Yes, I will add the formula for three solutions in a separate PR.

> the speech audio example in use, it is not anything offensive or controversial, right?

The content is "High-end screen products are still in short supply", so that should be fine.

mthrok:
> The dataset is under Apache 2.0 license, I agree with Caroline, we can put the wavs to torchaudio. Some questions: 1) The wav files are ~3MB, it doesn't affect much on downloading the repo, right? 2) Should we also include a separate License in the directory?

I did "Download as zip" from the top page of the repo and currently the zip file is 2.9 MB, so adding 3MB file will double.

PR #1769, opened 2021-09-16
Add HuBERT model architectures

mthrok:
This PR adds HuBERT model architectures for 
 - `base` (pre-training)
 - `large` (pre-training / fine-tuning)
 - `xlarge` (pre-training / fine-tuning)

Since the internal components are same as `Wav2Vec2Model`, it reuses the existing modules..
With these models, it is possible to 
- import the pre-trained model published by `fairseq` and TorchScript it.
- finetune the existing model for downstream task.


nateanl (commented):
Looks good overall. For training HuBERT I might need additional layers on top of HuBERT encoder. Do you think it's good to add them in the model definition, or in the training recipe?

mthrok:
Offline discussion: Will split factory functions for models fine-tuned for ASR task and pre-trainig.

torchaudio/models/wav2vec2/components.py
```diff
@@ -397,6 +401,28 @@ def forward(
 
         return x
 
+    def get_intermediate_outputs(
+            self,
+            x: Tensor,
+            indices: Optional[int] = None,
```
carolineechen, line 407: this should be `Optional[List[int]]` instead as we expect a list of indices, not a single number. `len(indices)` would raise an error if `indices` is of type `int`

torchaudio/models/wav2vec2/components.py
```diff
@@ -397,6 +401,28 @@ def forward(
...
+    ) -> List[Tensor]:
+        if indices is None:
+            indices = list(range(len(self.layers)))
+        assert indices is not None
+        if len(indices) == 0:
+            return []
+        ret: List[Tensor] = []
+
+        x = self._preprocess(x)
+        for i, layer in enumerate(self.layers):
```
carolineechen, line 418: this way of iterating through layers does not ensure that the order of the resulting list of layers corresponds to the order of the indices, since things are being added to the return list in order of increasing layer index. ex/ `indices=[2,1]` would return `[layer1, layer2]` as opposed to `[layer2, layer1]`

with the current implementation, the check for `i in indices` should precede the layer masking computation to avoid unnecessary computation

nateanl: To achieve the purpose of correct ordering, we can improve the current impl by
```
for i in range(max(indices) + 1):
    layer = self.layers[i]
    x = layer(x, attention_mask)
    ret.append(x)
return ret[indices]
```

torchaudio/models/wav2vec2/components.py
mthrok: The changes related to feature extraction has been moved to #1776 due to its BC-breaking change. I will address it there.

torchaudio/models/wav2vec2/components.py
mthrok: Thanks, the current signature also allows the duplication of the layers.
I think it would be easier to resort to the `fairseq`'s signature where you provide the index of the last layer you need and return all the results from the previous layers. That way we do not need to worry about the order and duplication. We only need to worry if the index user provided is within the number of existing layers.

Let's continue this on  #1776

torchaudio/models/wav2vec2/model.py
```diff
@@ -6,20 +6,7 @@
...
-            Feature extractor that extracts feature vectors from raw audio Tensor.
-
-        encoder (torch.nn.Module):
-            Encoder that converts the audio features into the sequence of probability
-            distribution (in negative log-likelihood) over labels.
-    """
+class _BaseModel(Module):
     def __init__(
             self,
             feature_extractor: Module,
```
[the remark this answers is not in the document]
mthrok: Addressed in #1776

torchaudio/models/wav2vec2/model.py
```diff
@@ -100,6 +84,44 @@ def forward(
...
+        encoder (torch.nn.Module):
+            Encoder that converts the audio features into the sequence of probability
+            distribution (in negative log-likelihood) over labels.
+
+        aux (torch.nn.Module or None, optional):
+            Auxiliary module. If provided, the output from encoder is passed to this module.
+    """
+
+
+class HubertModel(_BaseModel):
```
mthrok, line 106: @nateanl 

I started to doubt the value of adding a separate `HubertModel` class. It adds maintenance cost, but practically, it is same as `Wav2Vec2Model`. With the factory functions bellow and pre-trained weights later, I think the existing `Wav2Vec2Model` suffices.

nateanl: Good point. I personally agree that the factory functions are good for now since
1) I think people would try the default configurations of HuBERT (Base, Large, etc) instead of building their own version. Once there is evidence that there is need for hyper-tuning the model, we can provide a more flexible solution then.
2) HuBERT is difficult to train in terms of time, especially for XLarge model. Based on it, the pretrained API is more likely to be used.

mthrok: Okay, I removed the `HubertModel` class. The fact that I did not need to update the unittests I wrote is also a strong evidence that the class was not necessary in the first place.

torchaudio/models/wav2vec2/utils/import_fairseq.py
```diff
@@ -127,9 +130,10 @@ def import_fairseq_model(original: Module) -> Wav2Vec2Model:
```
nateanl, line 130: Why there's an underscore after `fairseq`_. ?

nateanl (approved):
LGTM. Just some nits to be addressed. 
I'm wondering what kind of API is for loading weights file into torchaudio's model. In ``torchaudio/models/wave2vec2/utils/import_fairseq.py`` there is ``import_fairseq_model`` method, but it seems like you have to create fairseq's model first, then transform to torchaudio version?

torchaudio/models/wav2vec2/utils/import_fairseq.py
mthrok: This is a hyperlink syntax of RST format. See `fairseq` is linked to the repo. https://pytorch.org/audio/master/models.html#import-fairseq-model

mthrok:
> LGTM. Just some nits to be addressed. I'm wondering what kind of API is for loading weights file into torchaudio's model. In `torchaudio/models/wave2vec2/utils/import_fairseq.py` there is `import_fairseq_model` method, but it seems like you have to create fairseq's model first, then transform to torchaudio version?

@nateanl 

Yes, the existing way is to build it via `fairseq`. I will follow up with the pretrained weights soon. Then one can load pretrained weights directly within torchaudio.

PR #1771, opened 2021-09-16
Move MVDR and PSD modules to transforms

nateanl:


torchaudio/functional/functional.py
```diff
@@ -1798,3 +1798,23 @@ def rnnt_loss(
         return costs.sum()
 
     return costs
+
+
+def _get_mat_trace(input: torch.Tensor, dim1: int = -1, dim2: int = -2) -> torch.Tensor:
```
mthrok, line 1803: This function seems not used in `functional`. In that case, isn't `transforms.py` a better place to have?

mthrok (commented):
Looks good. Can you update `docs/source/transform.rst` so that new modules show up?

torchaudio/functional/functional.py
nateanl: Sure. When PyTorch supports this method, it will be removed from torchaudio anyway. I will move it to transforms then.

nateanl:
Updated the documentation in [2467bdc](https://github.com/nateanl/audio/commit/2467bdc1500051d6c681480c6537f6c08960de77)

torchaudio/transforms.py
```diff
@@ -1488,3 +1490,436 @@ def forward(
...
+        eps (float, optional): a value added to the denominator in mask normalization. (Default: 1e-15)
+    """
+
+    def __init__(self, multi_mask: bool = False, normalize: bool = True, eps: float = 1e-15):
+        super().__init__()
+        self.multi_mask = multi_mask
+        self.normalize = normalize
+        self.eps = eps
+
+    def forward(self, X: torch.Tensor, mask: Optional[torch.Tensor] = None):
```
mthrok, line 1530: For spectrogram input, please use `specgram ` as a variable name.

https://github.com/pytorch/audio/blob/main/CONTRIBUTING.md#conventions
(note that this convention note has some space for argument, `waveform` and `specgram`, are fairly enforced)

PR #1772, opened 2021-09-16
Refactor batch consistency test in transforms

nateanl:
The current batch consistency tests don't consider the inter-batch operation. If the transform contains inter-batch operation then the test should be failed. In the current implementation the inputs are repeated, hence the inter-batch operation may not be reflected.

### DIFF
--- a/test/torchaudio_unittest/transforms/batch_consistency_test.py
+++ b/test/torchaudio_unittest/transforms/batch_consistency_test.py
@@ -11,35 +11,44 @@ class TestTransforms(common_utils.TorchaudioTestCase):
 
     """Test suite for classes defined in `transforms` module"""
     def test_batch_AmplitudeToDB(self):
-        spec = torch.rand((2, 6, 201))
+        spec = torch.rand((3, 2, 6, 201))
 
         # Single then transform then batch
-        expected = torchaudio.transforms.AmplitudeToDB()(spec).repeat(3, 1, 1)
+        expected = []
+        for i in range(3):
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Looks mostly good. Please check my comments.

test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -1,222 +1,181 @@
...
-        waveform = common_utils.get_whitenoise(sample_rate=8000, duration=1, n_channels=2)
+        waveform = common_utils.get_whitenoise(sample_rate=8000, duration=1, n_channels=6)
+        waveform = waveform.reshape(3, 2, -1)
 
         # Single then transform then batch
-        waveform_encoded = torchaudio.transforms.MuLawEncoding()(waveform)
-        expected = waveform_encoded.unsqueeze(0).repeat(3, 1, 1)
+        expected = []
+        for i in range(3):
+            expected.append(T.MuLawEncoding()(waveform[i]))
```
hwangjeff, line 74: how about a list comprehension instead? e.g.
```suggestion
        expected = [T.MuLawEncoding()(waveform[i]) for i in range(3)]
```

nateanl: Thanks. I addressed it in https://github.com/pytorch/audio/pull/1772/commits/86d05b02bb78aa073a83785e269417b5b100fb95

PR #1776, opened 2021-09-18
[BC-Breaking] Update `extract_features` of Wav2Vec2Model

mthrok:
Originally, `extract_features` method was returning the result from
passing the input waveform to convolutional feature extractor module.

The features commonly used in downstream tasks are outputs from intermediate
layers of transformer block in encoder.

This commit update the behavior of `extract_features` to allow selectively
retrieve such features.

The previous behavior can be achieved with
`Wav2Vec2Model.feature_extractor(waveform)`.

### DIFF
--- a/torchaudio/models/wav2vec2/components.py
+++ b/torchaudio/models/wav2vec2/components.py
@@ -377,17 +377,21 @@ class Transformer(Module):
         self.dropout = nn.Dropout(dropout)
         self.layers = layers
 
-    def forward(
-            self,
-            x: Tensor,
-            attention_mask: Optional[Tensor] = None,
-    ):
+    def _preprocess(self, x: Tensor):
         x = x + self.pos_conv_embed(x)
 
         if self.layer_norm_first:
             x = self.layer_norm(x)
 
         x = self.dropout(x)
+        return x
+
+    def forward(
+            self,
+            x: Tensor,
+            attention_mask: Optional[Tensor] = None,
+    ):
+        x = self._preprocess(x)
         for layer in self.layers:
             if not (self.training and torch.rand(1).item() <= self.layer_drop):
                 x = layer(x, attention_mask)
@@ -396,6 +400,29 @@ class Transformer(Module):
             x = self.layer_norm(x)
 
         return x
+
+    def get_intermediate_outputs(
+            self,
+            x: Tensor,
+            attention_mask: Optional[Tensor] = None,
+            indices: Optional[List[int]] = None,
+    ) -> List[Tensor]:
+        if indices is None:
+            indices = list(range(len(self.layers)))
+        assert indices is not None
+
+        ret: List[Tensor] = []
+        if len(indices) == 0:
+            return ret
+
+        x = self._preprocess(x)
+        max_ind = max(indices)
+        for i, layer in enumerate(self.layers):
+            if i <= max_ind:
+                x = layer(x, attention_mask)
+                if i in indices:
+                    ret.append(x)
+        return ret
 
 
 class Encoder(Module):
@@ -410,11 +437,11 @@ class Encoder(Module):
         self.transformer = transformer
         self.readout = readout
 
-    def forward(
+    def _preprocess(
             self,
             features: Tensor,
             lengths: Optional[Tensor] = None,
-    ) -> Tensor:
+    ) -> Tuple[Tensor, Optional[Tensor]]:
         x = self.feature_projection(features)
 
         mask: Optional[Tensor] = None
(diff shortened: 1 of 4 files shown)


mthrok:
@nateanl Here is the proposal to address https://github.com/pytorch/audio/pull/1769#discussion_r711115336

torchaudio/models/wav2vec2/model.py
```diff
@@ -29,29 +30,38 @@ def __init__(
         self.feature_extractor = feature_extractor
         self.encoder = encoder
 
+    @torch.jit.export
     def extract_features(
```
nateanl, line 34: As offline discussed, we are thinking if users will want to fine-tune the Wave2Vec model for their down-stream task. If they just want the embeddings of Wave2Vec, this should be fine, we just need to set the model to eval mode to avoid training it. However, if they want to re-train the weights of some layers, it's good to add a method to freeze the unused layers and make ``requires_grad`` be True for the used layers.

torchaudio/models/wav2vec2/components.py
```diff
@@ -397,6 +401,29 @@ def forward(
 
         return x
 
+    def get_intermediate_outputs(
+            self,
+            x: Tensor,
+            attention_mask: Optional[Tensor] = None,
+            indices: Optional[List[int]] = None,
```
nateanl, line 408: The default value of indices can be the last layer? But it looks difficult to set such value for models of different sizes..

torchaudio/models/wav2vec2/components.py
```diff
@@ -426,11 +453,28 @@ def forward(
...
+            self,
+            features: Tensor,
+            lengths: Optional[Tensor] = None,
+    ) -> Tensor:
+        x, mask = self._preprocess(features, lengths)
         x = self.transformer(x, attention_mask=mask)
         x = self.readout(x)
         return x
 
+    def extract_feature(
```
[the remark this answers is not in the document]
mthrok: I share the sentiment of confusion, however the following is my thoughts

1. This method is the one attached on `Encoder` class, which will be used by `Wave2Vec2Model` class. The expected usage is via `Wav2Vec2Model.extract_features`.
2. This class is already called `Encoder`, so renaming the other module to `Encoder` might be more confusing.
3. Since the original `fairseq` call the other module `FeatureExtractor`, I think it is better to keep it that way.
4. The word `encode[r]` is overloaded.
5. From the viewpoint of encoder, the fact that the internal implementation uses transformer is better abstracted away.

> it's plural

Yes, let me fix that one.

torchaudio/models/wav2vec2/components.py
mthrok: The current default is to return the all layers.

```python
model = Wav2Vec2Model(...)
features, _ = model.extract_features(waveform, length, indices=None)

# features is a list of tensors, which consists of outputs from all the intermediate layers in transformer block.
```

I am open to change it to other formats, and here are the design criteria I would like to keep,
1. Since it is computationally heavy, we would like the interface to be flexible enough to extract features from multiple layers at the same time.
2. Once the output from last layer of the desired layers is extracted, stop the computation there. (no need to go through the rest of transformer layers.)

`fairseq` returns dictionary with `x` represents the last layer output, and `layer_resutls` contain the list of intermediate outputs.

The last layer output can be fetched like,

```python
features, _ = model.extract_features(waveform, length, indices=None)
last_layers = features[-1]
```

What do you think?

nateanl: I agree with the design criteria. To save the computational time, we may add a parameter like ``last_only=True`` which only extract the last layer's output by default. The logic can be
```
if indices is None:
    if last_only is True:
        # return last layer output
    else:
        # return all layer outputs
else:
    # return desired layer outputs

```

torchaudio/models/wav2vec2/components.py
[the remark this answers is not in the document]
nateanl: Got it. From users' viewpoint, they will only touch ``Wav2Vec2Model.extract_features`` and ``Wav2Vec2Model.forward`` methods. So this should be fine.

mthrok:
@nateanl @carolineechen I updated the code and changed the signatures to `num_layers: Optional[int]=None`.

By default, it will return outputs from all the layers in intermediate layers in transformer block of the encoder.
`num_layers` parameter will allow it to limit to the first `n` layers.

nateanl (approved):
LGTM. nit: the dropout function can be replaced by nn.Dropout [here](https://github.com/nateanl/audio/blob/main/torchaudio/models/wav2vec2/components.py#L276)

mthrok:
> LGTM. nit: the dropout function can be replaced by nn.Dropout [here](https://github.com/nateanl/audio/blob/main/torchaudio/models/wav2vec2/components.py#L276)

Nice suggestion. I think it's better to use the object-oriented form (`gelu` and `softmax` are other possiblities) as much as possible because, the op shows up as model architecture that way. Let's follow up.

PR #1781, opened 2021-09-21
Fix HF model integration

mthrok:
Previously, when testing wav2vec models from HF transformers, all the model were
instantiated as `Wav2Vec2ForCTC` class, while some of them were supposed to be
`Wav2Vec2Model`.

Fixing this revealed that model importer cannot correctly handle `Wav2Vec2Model` import.

This PR fixes these issues.

### DIFF
--- a/test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
+++ b/test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
@@ -221,4 +221,3 @@ class TestHFIntegration(TorchaudioTestCase):
         reloaded.load_state_dict(imported.state_dict())
         reloaded.eval()
         self._test_recreate(imported, reloaded, config)
-        
--- a/test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
+++ b/test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
@@ -39,14 +39,14 @@ HF_LARGE_LV60_SELF_960H = _load_config('facebook', 'wav2vec2-large-960h-lv60-sel
 HF_LARGE_XLSR_DE = _load_config('facebook', 'wav2vec2-large-xlsr-53-german')
 
 # Config and corresponding factory functions
-HF_CONFIGS = parameterized.expand([
-    # pretrained
+PRETRAIN_CONFIGS = parameterized.expand([
     (HF_BASE, wav2vec2_base),
     (HF_LARGE, wav2vec2_large),
     (HF_LARGE_LV60, wav2vec2_large_lv60k),
     (HF_LARGE_XLSR_53, wav2vec2_large_lv60k),
     (HF_BASE_10K_VOXPOPULI, wav2vec2_base),
-    # finetuned
+], name_func=_name_func)
+FINETUNE_CONFIGS = parameterized.expand([
     (HF_BASE_960H, wav2vec2_base),
     (HF_LARGE_960H, wav2vec2_large),
     (HF_LARGE_LV60_960H, wav2vec2_large_lv60k),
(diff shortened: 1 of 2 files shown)


torchaudio/models/wav2vec2/utils/import_huggingface.py
```diff
@@ -32,11 +32,17 @@ def _get_config(cfg):
 
 
 def _build(config, original):
+    if original.__class__.__name__ == 'Wav2Vec2ForCTC':
+        wav2vec2 = original.wav2vec2
+    else:
+        wav2vec2 = original
+
     imported = _get_model(**config)
```
mthrok, line 40: Note to reviewer

`original` here is an instance of either `Wav2Vec2Model` or `Wav2Vec2ModelForCTC`.
`Wav2Vec2ModelForCTC` instance contains an instance of `Wav2Vec2Model` as `wav2vec2` attribute.

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
```diff
@@ -194,3 +203,22 @@ def test_recreate(self, config, factory_func):
...
+    def test_recreate_pretrain(self, config, factory_func):
+        """Imported models can be recreated via a factory function without Hugging Face transformers."""
+        imported = import_huggingface_model(self._get_model(config)).eval()
+        reloaded = factory_func(num_out=imported.encoder.readout.out_features)
+        reloaded.load_state_dict(imported.state_dict())
+        reloaded.eval()
+        self._test_recreate(imported, reloaded, config)
+
+    @FINETUNE_CONFIGS
+    def test_recreate_finetune(self, config, factory_func):
```
mthrok, line 217: At this moment, `test_recreate_pretrain` and `test_recreate_finetune` contain the exact same logic, but they will be different in the upcoming change. Since all the other tests are separated for fine-tuned models and pre-training models, I also split it here.

torchaudio/models/wav2vec2/utils/import_huggingface.py
```diff
@@ -32,11 +32,17 @@ def _get_config(cfg):
 
 
 def _build(config, original):
+    if original.__class__.__name__ == 'Wav2Vec2ForCTC':
```
hwangjeff, line 35: possible to perform the comparison on classes directly rather than string class names?

mthrok: We could, but the intention here is to avoid `import transformers` inside of `torchaudio` library, so that we do not have `transformers` as an optional dependency, taking advantage of duck-typing.

Of course the expectation is that `transformers` is installed and imported on client code, so it should be practically no problem to do it.

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
```diff
@@ -142,15 +144,22 @@ def test_import(self, config, _):
...
 
-        reloaded = factory_func(num_out=imported.encoder.readout.out_features)
-        reloaded.load_state_dict(imported.state_dict())
-        reloaded.eval()
+    @FINETUNE_CONFIGS
+    def test_import_finetune(self, config, _):
+        """wav2vec2 models from HF transformers can be imported and yields the same results"""
+        original = self._get_model(config).eval()
+        imported = import_huggingface_model(original).eval()
+        self._test_import_pretrain(original.wav2vec2, imported, config)
```
carolineechen, line 159: for my understanding, is it the case that all finetune models only contains models of type `Wav2Vec2ForCTC`? and are pretrained models always `Wav2Vec2Model` or can it be either?

mthrok: My understanding is that, fine-tunes models are of `Wav2Vec2ForCTC` class.

For pretraining model, it seems that it can be either `Wav2Vec2Model` class or `Wav2Vec2ForCTC` class, and if it's later case, then it comes with an additional, untrained Liner module.

torchaudio/models/wav2vec2/utils/import_huggingface.py
hwangjeff: got it — sounds reasonable

hwangjeff (approved):

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
mthrok: @carolineechen Merging the PR, but feel free to continue on dicussion.

carolineechen: yup np, the PR lgtm!

PR #1746, opened 2021-09-02
Standardize optional types in docstrings

carolineechen:
standardize optional type representation in docstrings:
- type `T` with default value: `T, optional`
- type `Optional[T]`: `T or None`
- type `Optional[T]` with default value: `T or None, optional`

PR #1782, opened 2021-09-21
[BC-Breaking] Move fine-tune specific module out of wav2vec2 encoder

mthrok:
[BC-Breaking] Move fine-tune specific module out of wav2vec2 encoder

Previously, the Linear module (called `readout`), used for an ASR fine-tuning task, was placed in encoder module. Conceptually, the encoder has nothing to do with a module specific to fine-tuning / downstreaming task.

The problems here are that;
1. encoder can be also used in pretraining phase, which does not require such a module.
2. The choice of Linear module is arbitral, and it is inconvenient for users to have hard-coded module structure in encoder.

Therefore this commit moves the Linear module out the encoder, and places it as `aux` attribute of `Wav2Vec2Model`. (as a result `Wav2Vec2Model` has `feature_extractor`, `encoder` and `aux` moduels)

An alternative approach is to define another module and place `Wav2Vec2Model` and aux module along each other. But that will introduce a new class we need to maintain, and since the use of `aux` is only expected for loading the pre-trained parameters published by `fairseq` (and it's variations from HF), and it is not general enough for downstream adoptations, where there will be a bunch of different more complicated models. (i.e. s3prl)

So based on the minimalistic approach, we put them inside of `Wav2Vec2Model`.

### DIFF
--- a/torchaudio/models/wav2vec2/model.py
+++ b/torchaudio/models/wav2vec2/model.py
@@ -20,15 +20,20 @@ class Wav2Vec2Model(Module):
         encoder (torch.nn.Module):
             Encoder that converts the audio features into the sequence of probability
             distribution (in negative log-likelihood) over labels.
+
+        aux (Optional[torch.nn.Module]):
+            Auxiliary module. If provided, the output from encoder is passed to this module.
     """
     def __init__(
             self,
             feature_extractor: Module,
             encoder: Module,
+            aux: Optional[Module] = None,
     ):
         super().__init__()
         self.feature_extractor = feature_extractor
         self.encoder = encoder
+        self.aux = aux
 
     @torch.jit.export
     def extract_features(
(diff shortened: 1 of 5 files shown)


carolineechen (approved):
LGTM

torchaudio/models/wav2vec2/model.py
```diff
@@ -20,15 +20,20 @@ class Wav2Vec2Model(Module):
         encoder (torch.nn.Module):
             Encoder that converts the audio features into the sequence of probability
             distribution (in negative log-likelihood) over labels.
+
+        aux (Optional[torch.nn.Module]):
```
hwangjeff, line 24: nit: should we follow the scheme described at #1746? e.g.

```suggestion
        aux (torch.nn.Module or None, optional):
```

hwangjeff (approved):

nateanl (approved):
LGTM

torchaudio/models/wav2vec2/model.py
mthrok: good catch.

PR #1783, opened 2021-09-21
[BC-Breaking] Split pretraining and finetuning factory functions

mthrok:
Previously, factory functions of wav2vec2 only generated the architecture
for the fine-tuning architecture used in wav2ve2 paper for ASR task.
That is, pre-training architecture + Linear module, and it did not
provide a straightforward way to generate architectures for pre-training.

The goal of the original implementation was to allow the inference of
wav2vec2 in non-Python environment via TorchScript. Now we would like to
expand it to pre-training/fine-tuning and HuBERT model as well.

Therefore, we need to have factory functions for both pre-training and
fine-tuning. This commit introduces new factory functions and separate
functions for pre-training and fine-tuning.

1. New functions for ASR fine-tuning.

We introdcue `wav2vec2_asr_XXX` functions which generates the architecture
used for the fine-tuning task in wav2vec2 paper. *1

2. Re-purpse the old functions

The existing functions, `wav2vec2_XXX`, now generates the architecture with
pre-trainig module only. (no Linear module)

Note
*1 This architecture is just one way to define architecture for fine-tuning
and it is not universal definition. The new `wav2vec2_asr_XXX` functions are
designed to provide these specific fine-tuning configuration and they are not
meant to support generic architecture for downstream task.

### DIFF
--- a/test/torchaudio_unittest/models/wav2vec2/model_test.py
+++ b/test/torchaudio_unittest/models/wav2vec2/model_test.py
@@ -37,8 +37,7 @@ finetune_factory_funcs = parameterized.expand([
 
 
 class TestWav2Vec2Model(TorchaudioTestCase):
-    def _smoke_test(self, device, dtype):
-        model = wav2vec2_asr_base(num_out=32)
+    def _smoke_test(self, model, device, dtype):
         model = model.to(device=device, dtype=dtype)
         model = model.eval()
 
@@ -54,12 +53,18 @@ class TestWav2Vec2Model(TorchaudioTestCase):
 
     @parameterized.expand([(torch.float32, ), (torch.float64, )])
     def test_cpu_smoke_test(self, dtype):
-        self._smoke_test(torch.device('cpu'), dtype)
+        model = wav2vec2_base()
+        self._smoke_test(model, torch.device('cpu'), dtype)
+        model = wav2vec2_asr_base(num_out=32)
+        self._smoke_test(model, torch.device('cpu'), dtype)
 
     @parameterized.expand([(torch.float32, ), (torch.float64, )])
     @skipIfNoCuda
     def test_cuda_smoke_test(self, dtype):
-        self._smoke_test(torch.device('cuda'), dtype)
+        model = wav2vec2_base()
+        self._smoke_test(model, torch.device('cuda'), dtype)
+        model = wav2vec2_asr_base(num_out=32)
+        self._smoke_test(model, torch.device('cuda'), dtype)
 
     def _feature_extractor_test(self, model):
         batch_size, num_frames = 3, 1024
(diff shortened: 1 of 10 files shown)


torchaudio/models/wav2vec2/model.py
```diff
@@ -138,34 +138,53 @@ def _get_model(
...
+    [:footcite:`baevski2020wav2vec`] for fine-tuning for ASR task.
 
     Args:
         num_out: int
             The number of output labels.
 
     Returns:
-        Wav2Vec2Model: The resulting model.
-
-    Example - Reload fine-tuned model from Hugging Face:
```
mthrok, line 158: I plan to update the example when adding pre-trained weights

hwangjeff (approved):

test/torchaudio_unittest/models/wav2vec2/model_test.py
```diff
@@ -19,16 +22,23 @@ def _name_func(testcase_func, i, param):
...
+    (wav2vec2_asr_base, ),
+    (wav2vec2_asr_large, ),
+    (wav2vec2_asr_large_lv60k, ),
+], name_func=_name_func)
+
+
 class TestWav2Vec2Model(TorchaudioTestCase):
     def _smoke_test(self, device, dtype):
-        model = wav2vec2_base(num_out=32)
+        model = wav2vec2_asr_base(num_out=32)
```
carolineechen, line 41: should we also do a smoke test for the pretrained w2v2 model?

carolineechen (approved):
Changes look good overall! Could you take a look at the unittest failure before merging? Also I'm curious as to what the plan is to warn users or publicize these BC breaking changes, and if you can also add the BC-breaking label to these PRs?

nateanl (approved):
LGTM

mthrok:
> Could you take a look at the unittest failure before merging? 

Interesting it is only failing on macOS, but failing on all Python versions.

`E   AssertionError: False is not true : Tensors failed to compare as equal!With rtol=1.3e-06 and atol=1e-05, found 1 element(s) (out of 6144) whose difference(s) exceeded the margin of error (including 0 nan comparisons). The greatest difference was 1.0967254638671875e-05 (-0.12493154406547546 vs. -0.12494251132011414), which occurred at index (1, 1, 169). `

I guess this is caused by some library implementation detail in PyTorch core. Since the difference is very small, I will increase the tolerance for macOS with notes.

> Also I'm curious as to what the plan is to warn users or publicize these BC breaking changes,

All the BC-braking changes related to wav2vec2 is for the same goal of supporting pretraining and fine-tuning.  I plan to add a section about it in the upcoming release note, with migration note, then link to each PR.

> add the BC-breaking label to these PRs?

Sure. thanks for the reminder.

PR #1777, opened 2021-09-18
Enable audio windows cuda tests

mszhanyi:
Enable `unittest_windows_gpu`.
1. add `cuda_install` (runtime) step.
2. add update cuda driver step.
3. add set cuda envs.
4. enable  `TORCHAUDIO_TEST_FORCE_CUDA=1`

These 4 changes have to work together to pass `unittest_windows_gpu`

closes #1565

PR #1787, opened 2021-09-23
Enable audio build on Windows with CUDA

mszhanyi:
With this PR, audio could be build on Windows with CUDA now.
It's from https://github.com/pytorch/audio/pull/1777
I've verified it locally.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,8 +26,17 @@ if(env_cxx_standard GREATER -1)
       WARNING "C++ standard version definition detected in environment variable."
       "PyTorch requires -std=c++14. Please remove -std=c++ settings in your environment.")
 endif()
+
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_C_STANDARD 11)
+
+# https://developercommunity.visualstudio.com/t/VS-16100-isnt-compatible-with-CUDA-11/1433342
+if(MSVC)
+  if(USE_CUDA)
+    set(CMAKE_CXX_STANDARD 17)
+  endif()
+endif()
+
 
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 set(CMAKE_POSITION_INDEPENDENT_CODE ON)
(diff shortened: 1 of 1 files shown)


mszhanyi:
@mthrok 
Did you miss removing something related in this commit.
https://github.com/pytorch/audio/commit/0f82217914c84f02ceab3d35ffdf2c2296836eda#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a
image

without this, there's one error locally.
image
cc @malfet

CMakeLists.txt
```diff
@@ -26,9 +26,18 @@ if(env_cxx_standard GREATER -1)
...
 endif()
+
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_C_STANDARD 11)
 
+# https://developercommunity.visualstudio.com/t/VS-16100-isnt-compatible-with-CUDA-11/1433342
+if(MSVC)
+  if(USE_CUDA)
+    set(CMAKE_CXX_STANDARD 17)
+  endif()
```
malfet, line 37: But that wouldn't work when TorchAudio is compiled with CUDA-10.2, would it?

malfet (approved):
LGTM if CI is green

mthrok:
> @mthrok
> Did you miss removing something related in this commit.
> [0f82217#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a](https://github.com/pytorch/audio/commit/0f82217914c84f02ceab3d35ffdf2c2296836eda#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20a)
> image
> 
> without this, there's one error locally.
> image
> cc @malfet

The deletion itself is as-intended. The option was removed because `libtorchaudio` should be built all the time. `libtorchaudio` is where all the C++ implementation that do not require `libpython` dependency is placed.

What I might have missed is to check if the extension `.dll` is/should be recognized by `setuptools`.

The logic there is that;

1. `python setup.py build_ext` invokes the compilation of extension module, expecting `torchaudio._torchaudio` and `torchaudio.lib.libtorchaudio`.
2. cmake builds `libtorchaudio` and `_torchaudio`.
For Linux and macOS they are `libtorchaudio.so` and `_torchaudio.so`. For Windows `_torchaudio` has to have `pyd` extension but I am not sure about `libtorchaudio` as it is C++ library without any Python component.
3. `python setup.py install` picks up the compiled modules from build directory and place it in the installation location.
Your error indicates that `setuptools` expect the `libtorchaudio` to have extension `.pyd` but, currently CMake produces `libtorchaudio.dll` thus `setuptools` cannot find `libtorchaudio`.

I am not sure what's the right fix for Windows, but one way it might work is to set the extension of `libtorchaudio` to `pyd` as well. You can try adding [this](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/csrc/CMakeLists.txt#L130-L132) for `libtorchaudio` [here](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/csrc/CMakeLists.txt#L92), then change [this line](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/_extension.py#L14) from `dll` to `pyd`.

CMakeLists.txt
mszhanyi: It would.
https://app.circleci.com/pipelines/github/pytorch/audio/7396/workflows/058b3c2c-8421-40c2-915c-a8441deac091/jobs/339245.
It's a Visual Studio Issue. I verified that another workaround is to downgrade the Visual Studio to 16.8.5 which PyTorch Builder is using.
But I think user experience might better that the user doesn't need to select the compiler with the specific version.

mszhanyi:
> I am not sure what's the right fix for Windows, but one way it might work is to set the extension of `libtorchaudio` to `pyd` as well. You can try adding [this](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/csrc/CMakeLists.txt#L130-L132) for `libtorchaudio` [here](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/csrc/CMakeLists.txt#L92), then change [this line](https://github.com/pytorch/audio/blob/c69955c6dcff7840f69ba9be188b900e480c9633/torchaudio/_extension.py#L14) from `dll` to `pyd`.

Thanks @mthrok . It works now. I've submitted another PR https://github.com/pytorch/audio/pull/1788 for these change.
Once #1788 is merged, It can  be merged.

CMakeLists.txt
mszhanyi: The root cause is a regression in VS. https://github.com/microsoft/STL/issues/2020

PR #1790, opened 2021-09-24
[doc] Fix return type of wav2vec2 model

mthrok:
## Before

Screen Shot 2021-09-24 at 17 48 55

## After

Screen Shot 2021-09-24 at 17 49 10


carolineechen (approved):

mthrok:
Does "optional output" sound like it is conditionally included in returned values?

PR #1791, opened 2021-09-25
[CircleCI] Update GPU resource class

malfet:
s/gpu.small/gpu.nvidia.medium/


mthrok:
It seems that the new image is not as powerful as previous one. The tests keep failing. Should we increase the size?

atalman (approved):

PR #1799, opened 2021-09-28
Add HuBERT pre-trained weight support

mthrok:
This PR adds 5 pre-trained weight support to HuBERT factory functions.
The pre-trained models were taken from https://github.com/pytorch/fairseq/tree/main/examples/hubert#pre-trained-and-fine-tuned-asr-models.


test/integration_tests/hubert_test.py
```diff
@@ -0,0 +1,26 @@
+import torchaudio
```
mthrok, line 1: Note to reviewer: This is an integration test, which is not a part of unit test as it depends on external resources. (network and the actual pre-trained model)

We want to run this test suite occasionally (say once a day on `main` branch), which I have not set up yet but will follow up.

torchaudio/models/wav2vec2/model.py
```diff
@@ -387,98 +454,203 @@ def hubert_large() -> Wav2Vec2Model:
...
     Returns:
         HuBERT: The resulting model.
     """
-    return _get_model(
+    checkpoints = ["fairseq_xlarge_ll60k"]
+    if checkpoint is not None:
+        if checkpoint not in checkpoints:
+            raise ValueError(f'Expected one of {checkpoints}. Found: "{checkpoint}".')
+
+    model = _get_model(
```
mthrok, line 564: Note to reviewer: Turned out that the following configurations are wrong.

torchaudio/models/wav2vec2/model.py
```diff
@@ -330,16 +362,55 @@ def wav2vec2_asr_large_lv60k(num_out: int) -> Wav2Vec2Model:
     )
 
 
-def hubert_base() -> Wav2Vec2Model:
+def hubert_base(
+        *,
+        checkpoint: Optional[str] = None,
+        dl_kwargs: Dict[str, Any] = None,
```
hwangjeff, line 368:
```suggestion
        dl_kwargs: Optional[Dict[str, Any]] = None,
```

also applies to functions below

hwangjeff (approved):

torchaudio/models/wav2vec2/model.py
mthrok: Good catch. Addressed!

mthrok:
Replaced by #1821

PR #1802, opened 2021-09-29
Replace issue templates with new issue forms

NicolasHug:
This PR introduces the new issue forms that replace issue templates.

This is similar to what was done in torchvision https://github.com/pytorch/vision/pull/4299, you can see the end result here: https://github.com/pytorch/vision/issues/new/choose (click on one of the "Get Started" buttons)

The main new thing is that we can enforce some of the fields to be filled, especially for bug reports. It's also a much cleaner GUI for users IMHO, and we can provide better examples and instructions.

There is still a "blank" template available.

I removed the "Questions" form: we pretend we close these issues but we actually don't. I replaced it with a direct link to https://discuss.pytorch.org. Since we still have a "blank" template, I think this  covers all previous use-cases properly.

We don't need to have the exact same templates across all projects, so **please feel free to directly push to this PR to address your own comments and to tweak the templates as you please** :)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,60 @@
...
+      # All necessary imports at the beginning
+      import torch
+      import torchvision
+      from torchvision.ops import nms
+
+      # A succinct reproducing example trimmed down to the essential parts:
+      N = 5
+      boxes = torch.rand(N, 4)  # Note: the bug is here, we should enforce that x1 < x2 and y1 < y2!
+      scores = torch.rand(N)
+      nms(boxes, scores, iou_threshold=.9)
```
NicolasHug, line 27: This is still vision-specific, I guess it'd be better to have a torchaudio snippet here. Perhaps @mthrok will be more inspired than I am for this?

nateanl: I have an example for the template. 
```python
    import torch
    from torchaudio.transforms import Spectrogram, InverseSpectrogram
    # A succinct reproducing example trimmed down to the essential parts:
    theta = torch.linspace(0, 300 * 3.14, 8000)
    waveform = torch.sin(theta).unsqueeze(0)
    n_fft = 400
    center = False
    spec = Spectrogram(n_fft=n_fft, center=center, power=None, return_complex=True)(waveform)
    recon = InverseSpectrogram(n_fft=n_fft, center=center)(spec)  # Note: the bug is here, the window and center settings must satisfy NOLA check.
    print('recon:', recon.shape)
```

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,62 @@
...
+  attributes:
+    value: >
+      #### Before submitting a bug, please make sure the issue hasn't been already addressed by searching through [the existing and past issues](https://github.com/pytorch/audio/issues?q=is%3Aissue+sort%3Acreated-desc+).
+- type: textarea
+  attributes:
+    label: 🐛 Describe the bug
+    description: |
+      Please provide a clear and concise description of what the bug is.
+
+      If relevant, add a minimal example so that we can reproduce the error by running the code. It is very important for he snippet to be as succinct (minimal) as possible, so please take time to trim down any irrelevant code to help us debug efficiently. We are going to copy-paste your code and we expect to get the same result as you did: avoid any external data, and include the relevant imports, etc. For example:
```
mthrok, line 15:
```suggestion
      If relevant, add a minimal example so that we can reproduce the error by running the code. It is very important for the snippet to be as succinct (minimal) as possible, so please take time to trim down any irrelevant code to help us debug efficiently. We are going to copy-paste your code and we hope to get the same result as you did: avoid any external data, and include the relevant imports, etc. For example:
```

IMHO, `expect` is a bit strong and I am afraid this might scar some users away because it might suggest that we will not respond if it's the snippet is runnable.
Of course, the complete repro code is nice and helpful and often times, a good gist is also as efficient as well.
We cannot always expect the complete repro code for various reasons, yet we still would like to see an issue filed than not filed at all.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,62 @@
...
+      theta = torch.linspace(0, 300 * 3.14, 8000)
+      waveform = torch.sin(theta).unsqueeze(0)
+      n_fft = 400
+      center = False
+      spec = Spectrogram(n_fft=n_fft, center=center, power=None, return_complex=True)(waveform)
+      recon = InverseSpectrogram(n_fft=n_fft, center=center)(spec)  # Note: the bug is here, the window and center settings must satisfy NOLA check.
+      print('recon:', recon.shape)
+      ```
+
+      If the code is too long (hopefully, it isn't), feel free to put it in a public gist and link it in the issue: https://gist.github.com.
```
mthrok, line 32:
```suggestion
      If the code and or the traceback is too long, feel free to put it in a public gist and link it in the issue: https://gist.github.com.
```

.github/ISSUE_TEMPLATE/documentation.yml
```diff
@@ -0,0 +1,20 @@
+name: 📚 Documentation
+description: Report an issue related to https://pytorch.org/audio/stable/index.html
+
+body:
+- type: textarea
+  attributes:
+    label: 📚 The doc issue
+    description: >
+      A clear and concise description of what content in https://pytorch.org/audio/stable/index.html is an issue. If this has to do with the general https://pytorch.org website, please file an issue at https://github.com/pytorch/pytorch.github.io/issues/new/choose instead. If this has to do with https://pytorch.org/tutorials, please file an issue at https://github.com/pytorch/tutorials/issues/new.
```
mthrok, line 9:
```suggestion
      A description of what content in https://pytorch.org/audio/stable/index.html is an issue. If this has to do with the general https://pytorch.org website, please file an issue at https://github.com/pytorch/pytorch.github.io/issues/new/choose instead. If this has to do with https://pytorch.org/tutorials, please file an issue at https://github.com/pytorch/tutorials/issues/new.
```

mthrok (approved):

Issue #733, opened 2020-06-19
doc for bg_iterator et al


vincentqb:
The [documentation](https://pytorch.org/audio/datasets.html) for torchaudio does not include `bg_iterator` and `disk_cache_iterator` from [code](https://github.com/pytorch/audio/blob/master/torchaudio/datasets/utils.py). Thanks to [comment](https://discuss.pytorch.org/t/audio-dataset-load-large-file-into-memory-in-background/85943) to point this out :)

vincentqb:
Example usage with any iterator:
```
for i in bg_iterator(some_iterator):
    # do stuff
```
In particular with `data_loader`:
```
for datapoint in bg_iterator(data_loader):
    # training
```

faroit:
I evaluated the use of `bg_iterator` in my pipeline and got almost no speed up regardless which loading backend I used. 

@vincentqb Can you share your experience when to expect an improvement using it especially in combination with the number of workers?

faroit:
furthermore together with torch 1.7 I now get deadlocks when using `bg_iterator`... simular to [this report](https://discuss.pytorch.org/t/possible-deadlock-training-example-gets-stuck/94221). Is there anything that could have caused this?

mthrok:
@vincentqb Do you want to look into this?

mthrok:
@AzizCode92 found a bug in `bg_iterator`. It hangs when loading fails. https://github.com/pytorch/audio/pull/1126#issuecomment-751279803

So the `bg_iterator` is
1. no speed up
2. not documented
3. buggy

I do not think this utility is providing any value, and this is a pure technical debt. I suggest we remove it from the library.

PR #1809, opened 2021-09-30
Deprecate data utils

mthrok:
Now that torchdata is public, deprecating the utilities.

### DIFF
--- a/torchaudio/datasets/utils.py
+++ b/torchaudio/datasets/utils.py
@@ -236,7 +236,7 @@ class _DiskCache(Dataset):
         return len(self.dataset)
 
 
-@deprecated('Please migrate to torchdata.', version='0.11')
+@deprecated('', version='0.11')
 def diskcache_iterator(dataset: Dataset, location: str = ".cached") -> Dataset:
     return _DiskCache(dataset, location)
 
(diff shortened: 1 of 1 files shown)


carolineechen (approved):

nateanl (approved):

torchaudio/datasets/utils.py
```diff
@@ -234,6 +236,7 @@ def __len__(self) -> int:
         return len(self.dataset)
 
 
+@deprecated('Please migrate to torchdata.', version='0.11')
```
carolineechen, line 239: actually, could we replace this with a better message? [torchdata](https://github.com/pytorch/data) is still a prototype depending on pytorch nightly currently

carolineechen (changes requested):

torchaudio/datasets/utils.py
mthrok: In fact, there is no need to replace this because these utilities do not provide any value to end user.
Specifically, `bg_iterator` is reported to cause deadlock https://github.com/pytorch/audio/issues/733#issuecomment-736092030
So they should not be used at all. It is not documented either, so I bet no one is using it.
Since this is not a public API, and not documented I actually think the right way is to simply delete them. (won't be BC-breaking)
This was a sort of compromise. If the current message is not helpful, it will be strange to say `'Please stop using'` either. We can just delete them too. What do you think?

carolineechen: Is there any downside in not including an explicit message and just simply saying that it will be deprecated in v0.11? That way, we still provide warning to anyone who may be using the functions (even if unlikely), but do not provide a misleading/unnecessary message.

PR #1814, opened 2021-09-30
Set seed value and slightly loosen atol for hubert_xlarge model

mthrok:
Originally, the small numerical discrepancy was observed only on macOS CI.
Now it is also sporadically observed on Linux CI as well.
Thus, loosing atol (1.0e-5 -> 1.3e-5).


mthrok:
Incorporated in #1811

PR #1819, opened 2021-10-02
Add DR-VCTK dataset

### LINKED ISSUE #1792: Add dataset Device Recorded VCTK (Small subset version)
## 🚀 Feature

VCTK audio dataset is very huge 10GB,can we add small subset version of it 
Device Recorded VCTK (Small subset version)
https://datashare.ed.ac.uk/handle/10283/3038

kingyiusuen:
Closes #1792

### DIFF
--- a/test/torchaudio_unittest/datasets/dr_vctk_test.py
+++ b/test/torchaudio_unittest/datasets/dr_vctk_test.py
@@ -1,0 +1,136 @@
+import os
+from pathlib import Path
+
+from torchaudio.datasets import DR_VCTK
+
+from torchaudio_unittest.common_utils import (
+    TempDirMixin,
+    TorchaudioTestCase,
+    get_whitenoise,
+    save_wav,
+)
+
+
+_SUBSETS = ["train", "test"]
+_CONDITIONS = ["clean", "device-recorded"]
+_SOURCES = ["DR-VCTK_Office1_ClosedWindow", "DR-VCTK_Office1_OpenedWindow"]
+_SPEAKER_IDS = range(226, 230)
+_CHANNEL_IDS = range(1, 6)
+
+
+def get_mock_dataset(root_dir):
+    """
+    root_dir: root directory of the mocked data
+    """
+    mocked_samples = {}
+
+    dataset_dir = os.path.join(root_dir, "DR-VCTK")
+    os.makedirs(dataset_dir, exist_ok=True)
+
+    config_dir = os.path.join(dataset_dir, "configurations")
+    os.makedirs(config_dir, exist_ok=True)
+
+    sample_rate = 16000
+    seed = 0
+
+    for subset in _SUBSETS:
+        mocked_samples[subset] = []
+
+        for condition in _CONDITIONS:
+            audio_dir = os.path.join(dataset_dir, f"{condition}_{subset}set_wav_16k")
+            os.makedirs(audio_dir, exist_ok=True)
+
+        config_filepath = os.path.join(config_dir, f"{subset}_ch_log.txt")
+        with open(config_filepath, "w") as f:
+            if subset == "train":
+                f.write("\n")
+            f.write("File Name\tMain Source\tChannel Idx\n")
+
+            for speaker_id in _SPEAKER_IDS:
+                utterance_id = 1
+                for source in _SOURCES:
+                    for channel_id in _CHANNEL_IDS:
+                        filename = f"p{speaker_id}_{utterance_id:03d}.wav"
+                        f.write(f"{filename}\t{source}\t{channel_id}\n")
+
+                        data = {}
+                        for condition in _CONDITIONS:
+                            data[condition] = get_whitenoise(
+                                sample_rate=sample_rate,
+                                duration=0.01,
+                                n_channels=1,
+                                dtype='float32',
+                                seed=seed
+                            )
+                            audio_dir = os.path.join(dataset_dir, f"{condition}_{subset}set_wav_16k") 
+                            audio_file_path = os.path.join(audio_dir, filename)
+                            save_wav(audio_file_path, data[condition], sample_rate)
+                            seed += 1
+
+                        sample = (
+                            data[_CONDITIONS[0]],
+                            sample_rate,
+                            data[_CONDITIONS[1]],
+                            sample_rate,
+                            "p" + str(speaker_id),
+                            f"{utterance_id:03d}",
+                            source,
+                            channel_id,
+                        )
+                        mocked_samples[subset].append(sample)
+                        utterance_id += 1
+
+    return mocked_samples
+
+
+class TestDRVCTK(TempDirMixin, TorchaudioTestCase):
+    backend = 'default'
+
+    root_dir = None
+    samples = []
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        cls.samples = get_mock_dataset(cls.root_dir)
+
+    def _test_dr_vctk(self, dataset, subset):
+        num_samples = 0
+        for i, (
+            waveform_clean,
+            sample_rate_clean,
+            waveform_dr,
+            sample_rate_dr,
+            speaker_id,
+            utterance_id,
+            source,
+            channel_id,
+        ) in enumerate(dataset):
+            self.assertEqual(waveform_clean, self.samples[subset][i][0], atol=5e-5, rtol=1e-8)
+            assert sample_rate_clean == self.samples[subset][i][1]
+            self.assertEqual(waveform_dr, self.samples[subset][i][2], atol=5e-5, rtol=1e-8)
+            assert sample_rate_dr == self.samples[subset][i][3]
+            assert speaker_id == self.samples[subset][i][4]
+            assert int(utterance_id) == self.samples[subset][i][5]
+            assert source == self.samples[subset][i][6]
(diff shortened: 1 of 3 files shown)


torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,111 @@
...
+                    download_url(url, root, hash_value=checksum, hash_type="md5")
+                extract_archive(archive, self._path)
+
+        if not os.path.isdir(self._path):
+            raise RuntimeError(
+                "Dataset not found. Please use `download=True` to download it."
+            )
+
+        self._config = self._load_config(self._config_filepath)
+        self._walker = sorted(self._config)
```
mthrok, line 64: `_walker` is a misnomer, originated from legacy implementation. Can you call it something else?

mthrok (commented):
Hi @kingyiusuen 

Thanks for working on this all the way through writing test as well. I think most of the code was derived from existing ones and there are some improvements we can make on top of it. Please check my comments.

Also, I tried `dataset = torchaudio.datasets.DR_VCTK('../datasets/', download=True)` and got the following error.

```
  File "/torchaudio/foo.py", line 4, in <module>
    dataset = torchaudio.datasets.DR_VCTK('../datasets/', download=True)
  File "/torchaudio/torchaudio/datasets/dr_vctk.py", line 69, in __init__
    self._config = self._load_config(self._config_filepath)
  File "/torchaudio/torchaudio/datasets/dr_vctk.py", line 77, in _load_config
    with open(filepath) as f:
FileNotFoundError: [Errno 2] No such file or directory: '/datasets/DR-VCTK/configurations/train_ch_log.txt'
```

Looks like there are intermediate directories `DR-VCTK/DR-VCTK` created. Could you check if that's expected?

```
% tree ../datasets/ 
../datasets/
├── DR-VCTK
│   └── DR-VCTK
│       └── DR-VCTK
│           ├── README.txt
│           ├── clean_testset_wav_16k
│           │   ├── p232_001.wav
│           │   ├── p232_002.wav
│           │   ├── p232_003.wav
│           │   ├── p232_005.wav
...
│           │   ├── p287_421.wav
│           │   ├── p287_422.wav
│           │   ├── p287_423.wav
│           │   └── p287_424.wav
│           └── objective-measures
│               ├── test.txt
│               └── train.txt
└── DR-VCTK.zip

9 directories, 24799 files
```

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,111 @@
...
+        # Get string representation of 'root' in case Path object is passed
+        root = os.fspath(root)
+
+        archive = os.path.join(root, "DR-VCTK.zip")
+        self._subset = "train" if train else "test"
+        self._path = os.path.join(root, "DR-VCTK")
+        self._clean_audio_dir = os.path.join(self._path, f"clean_{self._subset}set_wav_16k")
+        self._noisy_audio_dir = os.path.join(self._path, f"device-recorded_{self._subset}set_wav_16k")
+        self._config_filepath = os.path.join(self._path, "configurations", f"{self._subset}_ch_log.txt")
+
```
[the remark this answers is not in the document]
kingyiusuen: The logic for verifying the checksum seems to be embedded in `download_url()`?

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,109 @@
...
+        train (bool): If True, creates dataset from training set, otherwise creates from test set.
+            (default: ``True``).
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        download: bool = False,
+        url: str = URL,
+        folder_in_archive: Union[str, Path] = FOLDER_IN_ARCHIVE,
```
mthrok, line 39: There is no need for `folder_in_archive` to be configurable. Especially, the default behavior verifies the the checksum of the downloaded downloaded archive, meaning that the intermediate directories are always there.

Can you remove it and hardcode it?

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,109 @@
...
+        self._clean_audio_dir = self._path / f"clean_{self._subset}set_wav_16k"
+        self._noisy_audio_dir = self._path / f"device-recorded_{self._subset}set_wav_16k"
+        self._config_filepath = self._path / "configurations" / f"{self._subset}_ch_log.txt"
+
+        if not self._path.is_dir():
+            if not archive.is_file():
+                if not download:
+                    raise RuntimeError("Dataset not found. Please use `download=True` to download it.")
+                checksum = _CHECKSUMS.get(url, None)
+                download_url(url, root, hash_value=checksum, hash_type="md5")
```
mthrok, line 58: Can you move the verification of checksum before `extract_archive`?
So that the verification is performed even when the code is not downloading but just extracting.

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,109 @@
...
+from torch.utils.data import Dataset
+
+import torchaudio
+from torchaudio.datasets.utils import (
+    download_url,
+    extract_archive,
+)
+
+
+URL = "https://datashare.ed.ac.uk/bitstream/handle/10283/3038/DR-VCTK.zip"
```
mthrok, line 14: Can prefix this with underscode? URL might change anytime, so it's safer to let client code not use it.

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,109 @@
...
+class DR_VCTK(Dataset):
+    """Create a dataset for Device Recorded VCTK (Small subset version).
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found.
+        download (bool):
+            Whether to download the dataset if it is not found at root path. (default: ``False``).
+        url (str): The URL to download the dataset from.
+            (default: ``"https://datashare.ed.ac.uk/bitstream/handle/10283/3038/DR-VCTK.zip"``)
+        train (bool): If True, creates dataset from training set, otherwise creates from test set.
```
mthrok, line 30: In case there is a version update with an addition of new subset in the future, can we change the signature to be more extendable one?
I think `"subset (str)"` with valid value of `"train"` and `"test"` is straight forward choice.

test/torchaudio_unittest/datasets/dr_vctk_test.py
```diff
@@ -0,0 +1,135 @@
...
+_CHANNEL_IDS = range(1, 6)
+
+
+def get_mock_dataset(root_dir):
+    """
+    root_dir: root directory of the mocked data
+    """
+    mocked_samples = {}
+
+    dataset_dir = Path(root_dir) / dr_vctk.FOLDER_IN_ARCHIVE
```
mthrok, line 26: Can you hard code `FOLDER_IN_ARCHIVE`? so that the test can detect the change if the library implementation accidentally changes it.

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,116 @@
...
+        self._noisy_audio_dir = self._path / f"device-recorded_{self._subset}set_wav_16k"
+        self._config_filepath = self._path / "configurations" / f"{self._subset}_ch_log.txt"
+
+        if not self._path.is_dir():
+            if not archive.is_file():
+                if not download:
+                    raise RuntimeError("Dataset not found. Please use `download=True` to download it.")
+                checksum = _CHECKSUMS.get(url, None)
+                download_url(url, root)
+            with open(archive, "rb") as file_obj:
```
mthrok, line 61: The variable `checksum` has to be defined right before performing the check.
In the current way, if the constructor is called when the archive exists but the directory does not, then it will fail.
Also, this will fail if custom URL is provided, as the result of `checksum = _CHECKSUMS.get(url, None)` is `None`.

The URL parameter is useful if the client code has some network restriction, but since we expect the archive to be same even if it is downloaded from somewhere different (say, re-host) can we simply use the same checksum?

If a user really wants to use archive that does not have the same checksum, they can extract it by themselves (but keeping the same directory structure) an the whole logic here is skipped.

Finally, since the logic around it is getting long, can you put this in a helper method?

torchaudio/datasets/dr_vctk.py
```diff
@@ -0,0 +1,116 @@
...
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found.
+        download (bool):
+            Whether to download the dataset if it is not found at root path. (default: ``False``).
+        url (str): The URL to download the dataset from.
+            (default: ``"https://datashare.ed.ac.uk/bitstream/handle/10283/3038/DR-VCTK.zip"``)
+        subset (str): The subset to use. Can be one of ``"train"`` and ``"test"``. (default: ``"train"``).
+    """
+
+    def __init__(
```
mthrok, line 32: Can we reorder the signature so that more used ones come earlier, and we can still add positional arguments?

```python
def __init__(self, root, subset, *, download, url):
```

mthrok (commented):
Hi @kingyiusuen 

Thank you so much for the quick response. It looks good but I have one more request (sorry 😣) to change, which is rather important.

torchaudio/datasets/dr_vctk.py
mthrok: The asterisk to make `download` and `url` keyword-only argument is missing.

mthrok (approved):
Looks good. Thank you for your contribution and patience for the PR.

PR #1821, opened 2021-10-04
Add HUBERT_BASE and HUBERT_ASR_LARGE pretrained models

mthrok:
This commit adds `torchaudio.models.HUBERT_BASE` and `torchaudio.models.HUBERT_ASR_LARGE`.
Which is an access point for retrieving pre-trained models published in `fairseq`, https://github.com/pytorch/fairseq/tree/main/examples/hubert#pre-trained-and-fine-tuned-asr-models.

And provides supplemental information for using the weights, such as labels (for ASR).

The new API looks like the following;
This API is different from the approach to put the pre-trained weight support in factory function like [`tacotoron2`](https://pytorch.org/audio/main/models.html#id4).
The reasoning behind this is that
1. when using pre-trained weight, typically it requires, other objects/information that are related to the data processing.
2. Separating factory function from pre-trained model loading, and building pre-trained model API on top of factory functions achieve nice [separation of concern](https://en.wikipedia.org/wiki/Separation_of_concerns) and make the responsibility of each component (Model object defines the computation graph and customizability, factory function make it easy to build model for specific configuration, then pre-trained model API makes it easy to retrieve all the information necessary for using the pre-trained weight).

In the succeeding PRs, I will add more pre-trained weights from HuBERT paper and wav2vec 2.0 paper.
Also we can update `tacotron2` to align with this API.

```python
>>> import torchaudio
>>>
>>> # Build the model and load pretrained weight.
>>> model = torchaudio.models.HUBERT_BASE.get_model()
Downloading:
100%|███████████████████████████████| 360M/360M [00:06<00:00, 60.6MB/s]
>>> # Extract acoustic features
>>> waveform, sample_rate = torchaudio.load('my_speech.mp3')
>>> features, _ = model.extract_features(waveform)
```

```python
>>> import torchaudio
>>>
>>> # Build the model and load pretrained weight.
>>> model = torchaudio.models.HUBERT_ASR_LARGE.get_model()
Downloading:
100%|███████████████████████████████| 1.18G/1.18G [00:17<00:00, 73.8MB/s]
>>> # Check the corresponding labels of the output.
>>> labels = torchaudio.models.HUBERT_ASR_LARGE.labels
>>> print(labels)
('<s>', '<pad>', '</s>', '<unk>', '|', 'E', 'T', 'A', 'O', 'N', 'I', 'H', 'S', 'R', 'D', 'L', 'U', 'M', 'W', 'C', 'F', 'G', 'Y', 'P', 'B', 'V', 'K', "'", 'X', 'J', 'Q', 'Z')
>>> # Infer the label probability distribution
>>> waveform, sample_rate = torchaudio.load('my_speech.mp3')
>>> emissions, _ = model(waveform)
>>> # Pass emission to decoder
>>> # `ctc_decode` is for illustration purpose only
>>> transcripts = ctc_decode(emissions, labels)
```


hwangjeff (approved):

docs/source/refs.bib
```diff
@@ -1,3 +1,59 @@
...
+      archivePrefix={arXiv},
+      primaryClass={cs.CL}
+}
+@inproceedings{Gales2014SpeechRA,
+  title={Speech recognition and keyword spotting for low-resource languages: Babel project research at CUED},
+  author={Mark John Francis Gales and Kate Knill and Anton Ragni and Shakti Prasad Rath},
+  booktitle={SLTU},
+  year={2014}
+}
+@misc{ardila2020common,
```
carolineechen, line 15: are these all being used with the current changes, or added in advance in preparation for future pretrained weights?

docs/source/models.rst
```diff
@@ -130,11 +130,26 @@ hubert_ft_xlarge
...
 
-.. currentmodule:: torchaudio.models.wav2vec2.utils
+Pre-trained Models
+------------------
+
+.. autoclass:: Wav2Vec2PretrainedModelBundle
+
+   .. automethod:: get_model
+
+   .. autoproperty:: labels
```
carolineechen, line 140: can you check if this is the expected docs format you had in mind? https://357491-90321822-gh.circle-artifacts.com/0/docs/models.html#torchaudio.models.Wav2Vec2PretrainedModelBundle.labels

docs/source/refs.bib
mthrok: It's the later case.

docs/source/models.rst
mthrok: This is what I had in mind, however, now I realize that it's missing type annotations. Let me work on it. Thanks for pointing.

mthrok: Updated https://357902-90321822-gh.circle-artifacts.com/0/docs/models.html#torchaudio.models.Wav2Vec2PretrainedModelBundle.labels

mthrok:
> LGTM. Do you also plan to add HuBERT_ASR_XLARGE?

@nateanl I do. I excluded others to make this PR smaller while introducing fixtures.

carolineechen (approved):

PR #1817, opened 2021-09-30
torchaudio: torch.quantization -> torch.ao.quantization

z-a-f:
Summary:
This changes the imports in the `torchaudio` to include the new import locations.

```
codemod -d pytorch/audio --extensions py 'torch.quantization' 'torch.ao.quantization'
```

Differential Revision: D31302450

PR #1823, opened 2021-10-05
[fbsync] torchaudio: torch.quantization -> torch.ao.quantization

mthrok:
Cherry-pick dc17160ac88404a6afc8f25fa6b83c560617146f which landed in `fbsync` branch via D31302450. (Corresponding export is in #1817)

Closes #1817


nateanl (approved):
LGTM. Could you add [fbsync] in the title when merging, so that we can easily find it for next iteration of fbsync?

mthrok:
> LGTM. Could you add [fbsync] in the title when merging, so that we can easily find it for next iteration of fbsync?

Good point. Done.

PR #1827, opened 2021-10-05
Add pretrained weights from wav2vec2.0 and XLSR papers

mthrok:
Add pretrained weights from https://github.com/pytorch/fairseq/tree/main/examples/wav2vec#pre-trained-models
- Wav2Vec 2.0 Base / Large / Large (LV-60)
- XLSR-53

PR #1828, opened 2021-10-06
[Cherry-picked 0.10] Doc touch up

mthrok:
After #1827, this commit adds the list of available pre-trained on the right side bar.
I liked the section presented in the before, but I think listing is more important and I do not know a way to have both.

**Before**
Screen Shot 2021-10-06 at 10 42 11 AM

**After**
Screen Shot 2021-10-06 at 10 41 11 AM


mthrok:
@nateanl  I found a better alternative.

Screen Shot 2021-10-06 at 11 11 45 PM

PR #1836, opened 2021-10-07
[Cherry-picked 0.10] Add license to pre-trained model doc

mthrok:


mthrok:
> It looks like all the sources are now pointing at the HuBERT pretrained/fine tuned models, even the sources for the wav2vec2 models. Those should point towards https://github.com/pytorch/fairseq/blob/ce6c9eeae163ac04b79539c78e74f292f29eaa18/examples/wav2vec/README.md#pre-trained-models instead.

Good catch. Fixed.

PR #1861, opened 2021-10-12
[Cherry-picked 0.10] Use int in pitch shift resample

carolineechen:
torchaudio resample will throw an error for non-integer sampling rates starting release 0.10 (see #1857), since the internal computation casts the sampling rate to an integer anyways. this PR moves the int cast to outside of the resample function used in pitch shift to avoid this error, but produces no functional changes to the method

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -1690,7 +1690,7 @@ def pitch_shift(
 
     Args:
         waveform (Tensor): The input waveform of shape `(..., time)`.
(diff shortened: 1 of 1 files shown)


carolineechen:
> While I was doing a quick spot check, I learned that `//` still returns float type. (which should not be a problem as `pitch_shift` is not TorchScript-able.)

This does not affect the result of the computation, but updating to use `int(a / b)` to keep it consistent with expected resample parameter type, thanks for noting!

PR #1900, opened 2021-10-19
Clean up setup.py

mthrok:
Clean up `setup.py`

1. Introduce `if __name__ == '__main__'`
2. Split the main execution logic and helper function, making the role of Git more explicitly expressed.
3. Get rid of global variables


mszhanyi:
Just curious, Could audio uses BuildExtension of torch.utils.cpp_extension?

https://github.com/pytorch/audio/blob/7a0d866e34af88a3f785dc8b38130e635824bcf9/setup.py#L113
https://github.com/pytorch/vision/blob/39d052c924304935999980716141786f38e43f8d/setup.py#L477

PR #1939, opened 2021-10-27
Run integration tests on CI

mthrok:
This PR adds integration test CI jobs.
It also adds custom `--use-tmp-hub-dir` option so that pre-trained weights (some of them are larger than 1GB) are deleted after each test. This should avoid the failure where CI runs out of the disk space.

Since we expect the integration job to download huge files, the test is only ran with Python 3.8.

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,0 +1,33 @@
+name: Integration Test
+
+on:
+  pull_request:
+    branches: [ main ]
+
+  workflow_dispatch:
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: [ 3.8 ]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v2
+      with:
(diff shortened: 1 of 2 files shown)


test/integration_tests/wav2vec2_pipeline_test.py
```diff
@@ -61,7 +61,6 @@ def test_pretraining_models(bundle):
 )
 def test_finetune_asr_model(
         bundle,
-        lang,
```
mthrok, line 64: This is not requirement for the CI job, but it's not really necessary either so I removed it.

nateanl:
The lfcc test failed before running the integration test on linux_gpu_py3.8. And I couldn't find the integration_tests on linux_cpu_py3.8. Is it supposed to be run on the two machines?

mthrok:
> The lfcc test failed before running the integration test on linux_gpu_py3.8. And I couldn't find the integration_tests on linux_cpu_py3.8. Is it supposed to be run on the two machines?

The integration workflow is independent of the existing unit test and it's on GitHub Action.
https://github.com/pytorch/audio/runs/4072462385?check_suite_focus=true

Right now, it is only CPU. I wonder if we should add GPU as well...

nateanl (approved):
LGTM.

PR #1943, opened 2021-10-28
Add script to collect PRs between commits

carolineechen:
as part of work to make compilation of release notes easier, this script will collect PRs between commits (previous release, new version commit) and output the relevant data (title, PR #, labels) in a json file that can easily be processed in a Jupyter notebook to organize commit categorization.

script adapted from [torchvision's](https://github.com/pytorch/vision/blob/main/scripts/release_notes/retrieve_prs_data.py)

### DIFF
--- a/scripts/retrieve_prs.py
+++ b/scripts/retrieve_prs.py
@@ -1,0 +1,145 @@
+import json
+import locale
+import os
+import re
+import sys
+import argparse
+import subprocess
+from collections import namedtuple
+from os.path import expanduser
+
+import requests
+
+Features = namedtuple(
+    "Features",
+    [
+        "title",
+        "pr_number",
+        "labels",
+    ],
+)
+
+
+def run(command):
+    """Returns (return-code, stdout, stderr)"""
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    output, err = p.communicate()
+    rc = p.returncode
+    enc = locale.getpreferredencoding()
+    output = output.decode(enc)
+    err = err.decode(enc)
+    return rc, output.strip(), err.strip()
+
+
+def commit_title(commit_hash):
+    cmd = f"git log -n 1 --pretty=format:%s {commit_hash}"
+    ret, out, err = run(cmd)
+    return out if ret == 0 else None
+
+
+def parse_pr_number(commit_hash, title):
+    regex = r"(#[0-9]+)"
+    matches = re.findall(regex, title)
+    if len(matches) == 0:
+        print(f"[{commit_hash}: {title}] Could not parse PR number, ignoring PR")
+        return None
+    if len(matches) > 1:
+        print(f"[{commit_hash}: {title}] Got two PR numbers, using the last one")
+        return matches[-1][1:]
+    return matches[0][1:]
+
+
+def get_ghstack_token():
+    pattern = "github_oauth = (.*)"
+    with open(expanduser("~/.ghstackrc"), "r+") as f:
+        config = f.read()
+    matches = re.findall(pattern, config)
+    if len(matches) == 0:
+        raise RuntimeError("Can't find a github oauth token")
+    return matches[0]
+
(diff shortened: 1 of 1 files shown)


mthrok:
@carolineechen Can we move this into existing `tools` directory instead of `scripts`?

scripts/retrieve_prs.py
```diff
@@ -0,0 +1,145 @@
...
+
+def run(command):
+    """Returns (return-code, stdout, stderr)"""
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    output, err = p.communicate()
+    rc = p.returncode
+    enc = locale.getpreferredencoding()
+    output = output.decode(enc)
+    err = err.decode(enc)
+    return rc, output.strip(), err.strip()
```
mthrok, line 31: @carolineechen Looking at how the returned items are used, I think it is redundant to return `rc` and `err.strip()`. The error check can be directly performed in this function, then there is no need to return return code and stderr.

Also this can be simplified with the use of `check_output`. like the following (except ascii decoding part)

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/setup.py#L15-L19

scripts/retrieve_prs.py
```diff
@@ -0,0 +1,145 @@
...
+    return matches[0]
+
+
+token = get_ghstack_token()
+headers = {"Authorization": f"token {token}"}
+
+
+def run_query(query):
+    request = requests.post("https://api.github.com/graphql", json={"query": query}, headers=headers)
+    if request.status_code == 200:
```
mthrok, line 68: 1. `requests.post` return `Response` object. Calling it `request` is misnomer.
2. Instead of raising exception, I think it is more appropriate to use `Response.raise_for_status()` method for checking the outcome.

scripts/retrieve_prs.py
```diff
@@ -0,0 +1,145 @@
...
+    rc, commits, err = run(cmd)
+    assert rc == 0, err
+
+    log_lines = commits.split("\n")
+    hashes, titles = zip(*[log_line.split(" ", 1) for log_line in log_lines])
+    return hashes, titles
+
+
+def _parse_args(args):
+    parser = argparse.ArgumentParser()
```
mthrok, line 121: Can you put a usage in module docstring and refer to it from CLI, so that help message can be viewed from both code and CLI?

See (`description=__doc__,`) https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/tools/convert_fairseq_models.py

scripts/retrieve_prs.py
```diff
@@ -0,0 +1,145 @@
...
+            print(f"{idx} / {len(hashes)}")
+
+    data = {commit: features._asdict() for commit, features in data.items()}    
+    with open(args.file, "w") as f:
+        json.dump(data, f)
+
+
+if __name__ == "__main__":
+    # Usage: python scripts/release_notes/retrieve_prs.py tags/v0.10.0 \
+    # 18685a517ae68353b05b9a0ede5343df31525c76 --file data.json
```
mthrok, line 144: This comment can be put into module level documentation.

scripts/retrieve_prs.py
```diff
@@ -0,0 +1,145 @@
...
+    _main(_parse_args(sys.argv[1:]))
```
mthrok, line 145: `sys.argv[1:]` should be replaced with `None`.

mthrok (commented):
@carolineechen 

Can you follow up on my comments?

scripts/retrieve_prs.py
nateanl: I found the [conv-tasnet pipeline](https://github.com/pytorch/audio/blob/8a347b62cf5c907d2676bdc983354834e500a282/examples/source_separation/train.py#L176) also uses ``argv[1:]`` in main method, shall we also replace that?

Issue #1883, opened 2021-10-15
Deprecate data utils


mthrok:
PyTorch core has download function and `torch.hub.download_url_to_file`.
Torchaudio can use it for dataset download and does not need to maintain its own `torchaudio.datasets.utils.download_url`.

In addition to that, there seems to be something wrong with `torchaudio.datasets.utils.download_url`.
It fails to download multiple of huge files. I think we should get rid of `torchaudio.datasets.utils.download_url` and `torchaudio.datasets.utils.stream_url`.

### Phase 1
1. Replace the usages of `torchaudio.datasets.utils.download_url` in `torchaudio` with `torch.hub.download_url_to_file` and [`torchaudio.datasets.utils.validate_file`](https://github.com/pytorch/audio/blob/e885204ea0d343b8bddd7e829b9e7b29e7c5a9bb/torchaudio/datasets/utils.py#L116)
2. Add [deprecation warning](https://github.com/pytorch/audio/blob/e885204ea0d343b8bddd7e829b9e7b29e7c5a9bb/torchaudio/_internal/module_utils.py#L42) to `torchaudio.datasets.utils.download_url` and `torchaudio.datasets.utils.stream_url`, saying it will be removed in `v0.12`.
### Phase 2
Once the `v0.11` is released and the main branch is bumped to `v0.12`, we can remove these functions.

## Note

For setting up the development environment, please refer to [CONTRIBUTING.md](https://github.com/pytorch/audio/blob/master/CONTRIBUTING.md) and [torchaudio_unittest](https://github.com/pytorch/audio/blob/master/test/torchaudio_unittest/README.md)

radioactive11:
@mthrok can I work on this issue?

mthrok:
Hi @radioactive11 

Sorry for the late reply. The phase 1 is completed and we need to wait until the branch cut of v0.11 to perform the phase 2. I will update this once the branch cut happens.

PR #1945, opened 2021-10-29
Deprecate data utils

harishsdev:
Deprecate data utils

PR for https://github.com/pytorch/audio/issues/1883

### DIFF
--- a/torchaudio/datasets/cmuarctic.py
+++ b/torchaudio/datasets/cmuarctic.py
@@ -148,7 +148,7 @@ class CMUARCTIC(Dataset):
             if not os.path.isdir(self._path):
                 if not os.path.isfile(archive):
                     checksum = _CHECKSUMS.get(url, None)
-                    download_url(url, root, hash_value=checksum, hash_type="md5")
+                    download_to_url(url, root, hash_prefix=checksum)
                 extract_archive(archive)
 
         self._text = os.path.join(self._path, self._folder_text, self._file_text)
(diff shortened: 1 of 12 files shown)


torchaudio/datasets/cmuarctic.py
```diff
@@ -148,7 +148,7 @@ def __init__(self,
             if not os.path.isdir(self._path):
                 if not os.path.isfile(archive):
                     checksum = _CHECKSUMS.get(url, None)
-                    download_url(url, root, hash_value=checksum, hash_type="md5")
+                    download_url_to_file(url, root, hash_prefix=checksum)
```
mthrok, line 151: `download_url_to_file` can only handle SHA256 hash, therefore the MD5 sum cannot be passed to the function. 
Either the checksum needs to be updated to SHA256 or validation has to be performed separately after the download.

torchaudio/datasets/utils.py
```diff
@@ -52,67 +52,6 @@ def stream_url(url: str,
             pbar.update(len(chunk))
 
 
-def download_url(url: str,
```
mthrok, line 55: We cannot delete the function until the next release. The `deprecated` decorator has to be put on this `download_url`.

torchaudio/datasets/yesno.py
```diff
@@ -44,6 +44,11 @@ def __init__(
 
         self._parse_filesystem(root, url, folder_in_archive, download)
 
+
+@_mod_utils.deprecated(
+    "donwload_url_to_file will be dericated in next version that is v0.12"
+    "version=0.12"
+)
```
mthrok, line 51: This decorator should be added on the implementation of `download_url_to_file`.
Also, `version` has to be keyword argument, not string.

mthrok (changes requested):
Hi @harishsdev 

Thanks for working on this. Please refer to my comments.

krishnakalyan3:
@harishsdev please let me know if you will continue to work on this issue. Else I would want to take a stab at this.

harishsdev:
@krishnakalyan3 please continue,im no more working on this isssue

PR #1951, opened 2021-11-01
Refactor collecting-PR script for release note

nateanl:
refer to https://github.com/pytorch/vision/pull/4626

### DIFF
--- a/tools/retrieve_prs.py
+++ b/tools/retrieve_prs.py
@@ -1,12 +1,18 @@
+"""
+Usage: python scripts/release_notes/retrieve_prs.py tags/v0.10.0 \
+    18685a517ae68353b05b9a0ede5343df31525c76 --file data.json
+"""
+import argparse
 import json
 import re
-import sys
-import argparse
 import subprocess
 from collections import namedtuple
 from os.path import expanduser
+from pathlib import Path
 
 import requests
+
+ROOT_DIR = Path(__file__).parent.resolve()
 
 Features = namedtuple(
     "Features",
@@ -57,11 +63,12 @@ headers = {"Authorization": f"token {token}"}
 
 
 def run_query(query):
-    request = requests.post("https://api.github.com/graphql", json={"query": query}, headers=headers)
-    if request.status_code == 200:
-        return request.json()
-    else:
-        raise Exception("Query failed to run by returning code of {}. {}".format(request.status_code, query))
+    response = requests.post("https://api.github.com/graphql", json={"query": query}, headers=headers)
+    try:
+        response.raise_for_status()
+        return response.json()
+    except requests.exceptions.HTTPError as e:
+        print(e)
 
 
 def gh_labels(pr_number):
@@ -109,7 +116,10 @@ def get_commits_between(base_version, new_version):
 
 
 def _parse_args(args):
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(
+        description=__doc__,
+        formatter_class=argparse.RawTextHelpFormatter,
+    )
     parser.add_argument("base_version", type=str, help="starting tag or commit (exclusive)")
     parser.add_argument("new_version", type=str, help="final tag or commit (inclusive)")
     parser.add_argument("--file", type=str, default="data.json", help="output json file")
@@ -131,6 +141,4 @@ def _main(args):
 
 
 if __name__ == "__main__":
-    # Usage: python scripts/release_notes/retrieve_prs.py tags/v0.10.0 \
-    # 18685a517ae68353b05b9a0ede5343df31525c76 --file data.json
-    _main(_parse_args(sys.argv[1:]))
+    _main(_parse_args())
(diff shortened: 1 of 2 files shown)


mthrok:
Looking at the code it is offending the bandit, it looks like the `subprocess.Popen(command, ...)` is causing the issue.

Before disabling the bandit, can you try change the following commands into list of string?
I think then, we can disable `shell=False`.

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/scripts/retrieve_prs.py#L35

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/scripts/retrieve_prs.py#L105

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/scripts/retrieve_prs.py#L111

mthrok:
The `git` command should work just fine without `shell=True`, like the following

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/setup.py#L84-L86

tools/retrieve_prs.py
```diff
@@ -1,15 +1,19 @@
+"""
+Usage: python scripts/release_notes/retrieve_prs.py tags/v0.10.0 \
```
carolineechen, line 2:
```suggestion
Usage: python tools/release_notes/retrieve_prs.py tags/v0.10.0 \
```

carolineechen:
thanks for looking into and working on this!

tools/retrieve_prs.py
```diff
@@ -20,21 +24,16 @@
...
+        return subprocess.check_output(cmd, cwd=ROOT_DIR).decode('ascii').strip()
+    except Exception:
+        return None
 
 
 def commit_title(commit_hash):
-    cmd = f"git log -n 1 --pretty=format:%s {commit_hash}"
-    ret, out, err = run(cmd)
-    return out if ret == 0 else None
+    cmd = ['git', 'log', '-n', '1', f'--pretty=format:%s {commit_hash}']
```
mthrok, line 35: I think this is different from what it used to be.
```suggestion
    cmd = ['git', 'log', '-n', '1', '--pretty=format:%s', f'{commit_hash}']
```

tools/retrieve_prs.py
```diff
@@ -64,11 +63,12 @@ def get_ghstack_token():
...
-    if request.status_code == 200:
-        return request.json()
-    else:
-        raise Exception("Query failed to run by returning code of {}. {}".format(request.status_code, query))
+    response = requests.post("https://api.github.com/graphql", json={"query": query}, headers=headers)
+    try:
+        response.raise_for_status()
+        return response.json()
+    except requests.exceptions.HTTPError as e:
+        print(e)
```
mthrok, line 71: No need of `try ~ except ` here. Just plain

```python
response.raise_for_status()
return response.json()
```

tools/retrieve_prs.py
```diff
@@ -1,15 +1,19 @@
+"""
```
mthrok, line 1: Can you give the description of what this script does?

tools/retrieve_prs.py
```diff
@@ -20,21 +24,16 @@
...
-    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-    output, err = p.communicate()
-    rc = p.returncode
-    enc = locale.getpreferredencoding()
-    output = output.decode(enc)
-    err = err.decode(enc)
-    return rc, output.strip(), err.strip()
+def _run_cmd(cmd):
+    try:
+        return subprocess.check_output(cmd, cwd=ROOT_DIR).decode('ascii').strip()
```
mthrok, line 29: I do not think `cwd` necessary.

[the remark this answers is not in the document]
mthrok: I actually think simply using `'utf8'` is as good as `getprefferedencoding` approach. If we want to be error safe, for each subprocess invocation, we need to detect the correct encoding.

tools/retrieve_prs.py
```diff
@@ -1,13 +1,18 @@
+"""Collect the PRs between the current and previous releases and
```
carolineechen, line 1:
```suggestion
"""Collect the PRs between two specified tags or commits and
```

tools/retrieve_prs.py
```diff
@@ -109,7 +109,10 @@ def get_commits_between(base_version, new_version):
 
 
 def _parse_args(args):
```
mthrok, line 111:
```suggestion
def _parse_args(args=None):
```

tools/retrieve_prs.py
```diff
@@ -19,10 +24,7 @@
 
 
 def _run_cmd(cmd):
-    try:
-        return subprocess.check_output(cmd).strip()
-    except Exception:
-        return None
+    return subprocess.check_output(cmd).decode('utf-8').strip()
```
mthrok, line 27: Note: This would fail if the commit message is not encoded with `utf-8`. (i.e. if user configures their git with non-uff commit message https://www.git-tower.com/help/guides/faq-and-tips/faq/encoding/windows)

The script needs to handle such case, while logging the failure and keep processing the rest of commits.

mthrok (approved):

tools/retrieve_prs.py
nateanl: Will follow-up the case in the future PR.

PR #1952, opened 2021-11-01
Clean up integration test

mthrok:
1. Get rid of custom download function (use `torch.hub.download_url_to_file`)
2. Remove `lang` from test function argument

### DIFF
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -1,5 +1,4 @@
 import torch
-import requests
 import pytest
 
 
@@ -33,9 +32,18 @@ def ctc_decoder():
 
 
 _FILES = {
-    'en': 'Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.flac',
-    'es': '20130207-0900-PLENARY-7-es_20130207-13_02_05_5.flac',
-    'fr': '20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac',
+    'en': (
+        'Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.flac',
+        '5005909deb42b30b5f95e0d7109455300de467f8750ff8b3515b1477c31eac24',
+    ),
+    'es': (
+        '20130207-0900-PLENARY-7-es_20130207-13_02_05_5.flac',
+        '293888e3bd729b946e4b32babc0c9bb6aa9c7d190a8a52663a1d772ea8026fea',
+    ),
+    'fr': (
+        '20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac',
+        '0ef184b9098936f1ca7f215395ba60ad9e90163482d5f49898f33b653ec64f33',
+    ),
 }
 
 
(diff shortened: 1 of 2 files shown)


test/integration_tests/wav2vec2_pipeline_test.py
```diff
@@ -61,7 +61,6 @@ def test_pretraining_models(bundle):
 )
 def test_finetune_asr_model(
         bundle,
-        lang,
```
hwangjeff, line 64: do we also need to remove the corresponding column in the parameters above?

test/integration_tests/wav2vec2_pipeline_test.py
```diff
@@ -61,7 +61,6 @@ def test_pretraining_models(bundle):
...
         expected,
         sample_speech,
         ctc_decoder,
```
hwangjeff, line 66: where do these come from?

test/integration_tests/conftest.py
```diff
@@ -33,23 +32,29 @@ def ctc_decoder():
 
 
 _FILES = {
-    'en': 'Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.flac',
-    'es': '20130207-0900-PLENARY-7-es_20130207-13_02_05_5.flac',
-    'fr': '20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac',
+    'en': (
+        'Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.flac',
+        '5005909deb42b30b5f95e0d7109455300de467f8750ff8b3515b1477c31eac24',
```
hwangjeff, line 37: might be nice to cite how these hashes were generated

test/integration_tests/wav2vec2_pipeline_test.py
mthrok: We cannot remove the one above. PyTest's `fixture` mechanism is very different from what a normal Python code would behave. Here is the gist of what happens according to my understanding.
1. When PyTest purses test code, it inspects the function argument
2. If the argument name matches a pre-defined fixture, it will be substituted with predefined fixture.
3. If an argument name matches with an argument name of a known fixture, then PyTest will bind it.

So, what happens here is that, PyTest runner, at import time, notices that `lang` is defined, and there is a fixture called `sample_speech(lang, ...)`, which takes `lang` as argument, so it automatically calls `smple_speech` function with `lang` parameter and give the resulting object to test function.
As I have not really used PyTest, this was very confusing behavior, and which is why I forgot to clean up the code and ended up leaving the `lang` argument there.

test/integration_tests/wav2vec2_pipeline_test.py
mthrok: It's defined in `conftest.py`, and PyTest will substitute it automatically. 

https://github.com/pytorch/audio/blob/184466a904c45d3a7e84629bff647b874f5448d5/test/integration_tests/conftest.py#L30-L32

test/integration_tests/conftest.py
mthrok: Adding comment that it's SHA256 hash would suffice?

hwangjeff: sounds good

nateanl (approved):

PR #1956, opened 2021-11-02
Add wav2vec2 ASR English pretrained model from voxpopuli

mthrok:
Add English ASR from VoxPopuli.


torchaudio/pipelines/_wav2vec2/utils.py
```diff
@@ -67,6 +67,38 @@ def _get_de_labels():
...
+        "w",
+        "y",
+        "b",
+        "v",
+        "k",
+        "x",
+        "j",
+        "q",
+        "z",
+    )
```
nateanl, line 99: wondering if the difference from ``_get_en_labels`` is just the lower case? Maybe we can define a standard label method so that the outputs of all english models can be transformed to that standard labels (e.g. in an alphabetic order in lower case).

mthrok: The order is different. `E T A O N I ...` vs `e t o i a n ...`.
That's an interesting idea, but to do that, we need to re-order the pertained weights on the final layer, which is kinda complicated and I wonder if it's really worth it.

nateanl: Instead of manipulating the model weight, we can build a dictionary that maps each model's labels to a standard one. But that takes the same lengths of code though..

mthrok: I think that is something client code can do on top of the existing API. I do not think that we have a strong reason that we have to do that, or the value propositions of such API.

nateanl: Yep, that makes sense.

nateanl (approved):

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -1033,6 +1033,50 @@ def _get_state_dict(self, dl_kwargs):
...
+        "encoder_dropout": 0.0,
+        "encoder_layer_norm_first": False,
+        "encoder_layer_drop": 0.1,
+        "aux_num_out": 28
+    },
+    _labels=utils._get_vp_en_labels(),
+    _sample_rate=16000,
+    _remove_aux_axis=(1, 2, 3, 31),
+)
+VOXPOPULI_ASR_BASE_10K_EN.__doc__ = """wav2vec 2.0 model with "Base" configuration.
```
carolineechen, line 1068: @mthrok fyi this doc format is inconsistent from the other ones (missing the extra line breaks between sentences, see [here](https://pytorch.org/audio/main/pipelines.html#torchaudio.pipelines.VOXPOPULI_ASR_BASE_10K_EN))

PR #1958, opened 2021-11-02
[BC-Breaking] Drop pseudo complex support from spectrogram

mthrok:
Following the plan #1337, this PR drops the support for pseudo complex type from `F.spectrogram` and `T.Spectrogram`. It also deprecates the use of `return_complex` argument.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -87,7 +87,7 @@ def spectrogram(
     if return_complex is not None:
         warnings.warn(
             "`return_complex` argument is now deprecated and is not effective."
-            "`torchaudio.functional.spectrogram(power=None)` always return tensor with "
+            "`torchaudio.functional.spectrogram(power=None)` always returns a tensor with "
             "complex dtype. Please remove the argument in the function call."
         )
 
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -53,7 +53,7 @@ def spectrogram(
         center: bool = True,
         pad_mode: str = "reflect",
         onesided: bool = True,
-        return_complex: bool = True,
+        return_complex: Optional[bool] = None,
 ) -> Tensor:
     r"""Create a spectrogram or a batch of spectrograms from a raw audio signal.
     The spectrogram can be either magnitude-only or complex.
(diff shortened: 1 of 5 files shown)


torchaudio/functional/functional.py
```diff
@@ -77,25 +77,18 @@ def spectrogram(
         onesided (bool, optional): controls whether to return half of results to
             avoid redundancy. Default: ``True``
         return_complex (bool, optional):
```
nateanl, line 79:
```suggestion
        return_complex (bool or None, optional):
```

torchaudio/transforms.py
```diff
@@ -65,13 +65,7 @@ class Spectrogram(torch.nn.Module):
         onesided (bool, optional): controls whether to return half of results to
             avoid redundancy (Default: ``True``)
         return_complex (bool, optional):
```
nateanl, line 67:
```suggestion
        return_complex (bool or None, optional):
```

torchaudio/transforms.py
```diff
@@ -108,7 +102,12 @@ def __init__(self,
         self.center = center
         self.pad_mode = pad_mode
         self.onesided = onesided
-        self.return_complex = return_complex
+        if return_complex is not None:
+            warnings.warn(
+                "`return_complex` argument is now deprecated and is not effective."
+                "`torchaudio.transforms.Spectrogram(power=None)` always return tensor with "
```
nateanl, line 108:
```suggestion
                "`torchaudio.transforms.Spectrogram(power=None)` always returns a tensor with "
```

torchaudio/functional/functional.py
```diff
@@ -77,25 +77,18 @@ def spectrogram(
...
     """
-    if power is None and not return_complex:
+    if return_complex is not None:
         warnings.warn(
-            "The use of pseudo complex type in spectrogram is now deprecated."
-            "Please migrate to native complex type by providing `return_complex=True`. "
-            "Please refer to https://github.com/pytorch/audio/issues/1337 "
-            "for more details about torchaudio's plan to migrate to native complex type."
+            "`return_complex` argument is now deprecated and is not effective."
+            "`torchaudio.functional.spectrogram(power=None)` always return tensor with "
```
nateanl, line 90:
```suggestion
            "`torchaudio.functional.spectrogram(power=None)` always returns a tensor with "
```

nateanl (approved):
LGTM! Just need to address the docstring then we can merge it.

carolineechen (approved):

torchaudio/functional/functional.py
mthrok: Here, `None` is used as an indicator to only know if user pass something, so we will keep it as `bool`.

PR #1960, opened 2021-11-02
Fix bandit failure

nateanl:


carolineechen (approved):

scripts/retrieve_prs.py
```diff
@@ -20,21 +18,16 @@
...
-    rc = p.returncode
-    enc = locale.getpreferredencoding()
-    output = output.decode(enc)
-    err = err.decode(enc)
-    return rc, output.strip(), err.strip()
+def _run_cmd(cmd):
+    try:
+        return subprocess.check_output(cmd).strip()
+    except Exception:
+        return None
```
mthrok, line 26: No `try ~ except` here. The purpose of the script is to purse, so if it cannot, it should stop.

mthrok: Also I think you need `decode("utf-8")`

nateanl: Okay, let me address it in the previous PR.

PR #1947, opened 2021-10-29
Add citation information

yangarbiter:
Thanks @nateanl for uploading the [paper](https://arxiv.org/abs/2110.15018) to arxiv.

I've added the bibtex information in this PR.

Please check.

Thanks.

PR #1962, opened 2021-11-02
Add citation information in the documentation

yangarbiter:
Continuing #1947


mthrok:
Can you double check that the rendering is as intended?

https://415138-90321822-gh.circle-artifacts.com/0/docs/index.html

mthrok (approved):

yangarbiter:
Since the author list is long, different screen size may render slightly differently.

Screen Shot 2021-11-02 at 1 24 17 PM

Is there a way to get around this?

nateanl:
Works for me. Maybe it's about the browser width.
image

nateanl (approved):

mthrok:
> Is there a way to get around this?

I do not think so as long as `code` block is used. I think we can live with it.

mthrok:
@nateanl Please do the honor and merge this.

PR #1964, opened 2021-11-02
Update PR label notifier

mthrok:
I found it hard to recall which labels are primary and which are secondary. This PR changes the bot to include that in the message.

```
% python .github/process_commit.py cbf267c357a7574d9382ffcace75089dc2798295
Hi @mthrok

You merged this PR, but the category labels are missing.
Please add a primary label (`bug fix`, `example`, `other`, `prototype`, `new feature`, `BC-breaking`, `deprecation`, `improvement`) and a secondary label (`module: ops`, `style`, `module: tests`, `other`, `build`, `module: docs`, `perf`, `module: models`, `module: datasets`, `module: pipelines`, `module: I/O`).
```

```
% python .github/process_commit.py 8c78273be11f05b8b77ba24f0764829b7b404b7b
```

### DIFF
--- a/.github/process_commit.py
+++ b/.github/process_commit.py
@@ -77,8 +77,8 @@ def _main():
     if not is_properly_labeled:
         print(f"""Hi @{merger}
 
-You merged this PR, but the category labels are missing.
-Please add a primary label ({_get_formatted(PRIMARY_LABELS)}) and a secondary label ({_get_formatted(SECONDARY_LABELS)}).
+You merged this PR, but one or more labels are missing.
+Please include a primary label ({_get_formatted(PRIMARY_LABELS)}) and a secondary label ({_get_formatted(SECONDARY_LABELS)}).
 """)  # noqa: E501
 
 
--- a/.github/process_commit.py
+++ b/.github/process_commit.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 """
 This script finds the merger responsible for labeling a PR by a commit SHA. It is used by the workflow in
 '.github/workflows/pr-labels.yml'. If there exists no PR associated with the commit or the PR is properly labeled,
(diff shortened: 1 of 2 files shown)


.github/process_commit.py
```diff
@@ -60,14 +61,26 @@ def get_pr_merger_and_labels(pr_number: int) -> Tuple[str, Set[str]]:
...
 
     merger, labels = get_pr_merger_and_labels(pr_number)
     is_properly_labeled = bool(PRIMARY_LABELS.intersection(labels) and SECONDARY_LABELS.intersection(labels))
 
     if not is_properly_labeled:
-        print(f"@{merger}")
+        print(f"""Hi @{merger}
+
+You merged this PR, but the category labels are missing.
+Please add a primary label ({_get_formatted(PRIMARY_LABELS)}) and a secondary label ({_get_formatted(SECONDARY_LABELS)}).
```
carolineechen, line 81:
```suggestion
You merged this PR, but one or more labels are missing.
Please include a primary label ({_get_formatted(PRIMARY_LABELS)}) and a secondary label ({_get_formatted(SECONDARY_LABELS)}).
```

carolineechen (approved):
The message suggestion is because it is possible for one label to be added but one to be missing, so this sounds more correct to me but feel free to play around with the message. 

Another possibility if you're interested is to have separate cases for missing primary and secondary labels, where the message only reminds of the missing label category(ies) instead of listing both all the time.

mthrok:
> The message suggestion is because it is possible for one label to be added but one to be missing, so this sounds more correct to me but feel free to play around with the message.

Sounds good.

> Another possibility if you're interested is to have separate cases for missing primary and secondary labels, where the message only reminds of the missing label category(ies) instead of listing both all the time.

I considered that but labels can be removed as well, so I thought it would not not hurt to list all of them all the time. Let's see how we actually feel about that with the new one and follow up.

mthrok:
Okay, the multiple lines are not working. I reverted this in #1965. Will retry.

PR #1968, opened 2021-11-03
Support multi-node training for source separation pipeline

nateanl:


examples/source_separation/README.md
```diff
@@ -68,9 +71,11 @@ python -u \
   "${this_dir}/lightning_train.py" \
   --num-speakers "${num_speakers}" \
   --sample-rate 8000 \
-  --data-dir "${dataset_dir}" \
+  --root-dir "${dataset_dir}" \
```
carolineechen, line 74:
```suggestion
  --root-dir "${root_dir}" \
```

carolineechen (approved):

mthrok (approved):
Looks good

In README, I think we can keep the number of nodes 1.
In my experience, using multiple nodes have diminishing return.
If the script is using 4 GPUs then, using one node is faster.

PR #1969, opened 2021-11-03
Add Colab/Download/Github link similar to tutorials

mthrok:
This commit adds colab/download/source link to tutorials, like in `pytorch/tutorials` repo.

Since the upstream `pytorch-sphinx-theme` does not provide the interface for this, I added a hack to update URL after the document is ready.

The hack might stop working if there is some update in `pytorch-sphinx-theme`.

Screen Shot 2021-11-04 at 1 55 43 PM

### DIFF
--- a/docs/source/_templates/layout.html
+++ b/docs/source/_templates/layout.html
@@ -60,9 +60,8 @@
       $(document).ready(function() {
 	  var downloadNote = $(".sphx-glr-download-link-note.admonition.note");
 	  if (downloadNote.length >= 1) {
-	      var tutorialUrlArray = $("#tutorial-type").text().split('/');
-              tutorialUrlArray[0] = tutorialUrlArray[0] + "_source"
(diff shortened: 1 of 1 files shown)


carolineechen (approved):
stamping, but can you double check that the run on colab works as intended? I'm getting a "Notebook not found" through trying it out on the build artifacts, but not sure if this is some authorization issue on my end or not. Everything else looks good, thanks!

mthrok:
> stamping, but can you double check that the run on colab works as intended? I'm getting a "Notebook not found" through trying it out on the build artifacts, but not sure if this is some authorization issue on my end or not. Everything else looks good, thanks!

Yeah, that's expected, because SG puts notebooks in randomly named directory, and when you are trying on CircleCI, these generated directory only exists in CircleCI but the generated link is pointing torchaudio's gh-page. I tested it by swapping the random string of the directory, and it worked. If it does not, I will follow up.

PR #1970, opened 2021-11-03
Port Audio I/O tutorials

mthrok:
This commit ports Audio I/O tutorials from tutorials repo.

The tutorials are placed under the new `Getting Started` section, and the index page is updated a bit.

*Current*
https://pytorch.org/audio/main/index.html
Screen Shot 2021-11-05 at 12 58 41

*After*
https://426860-90321822-gh.circle-artifacts.com/0/docs/index.html
Screen Shot 2021-11-05 at 12 41 46 PM

Note: I cannot `import boto3` on M1 Mac. I think it's an issue with the binary signature (`killed 9`). So the doc build process will not work on M1 Mac.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -25,8 +25,11 @@ Features described in this documentation are classified by release status:
 
 The :mod:`torchaudio` package consists of I/O, popular datasets and common audio transformations.
 
+Package References
+------------------
+
 .. toctree::
-   :maxdepth: 2
+   :maxdepth: 1
    :caption: Package Reference
 
    torchaudio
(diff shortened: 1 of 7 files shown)


hwangjeff (approved):
going forward, would we have to manually sync tutorials between here and the main tutorials repo?

mthrok:
> going forward, would we have to manually sync tutorials between here and the main tutorials repo?

Yes, so far. Or we can come up with some automation, or get rid of the tutorials on tutorials repo.

mthrok:
Let's see how these goes...

https://github.com/pytorch/tutorials/pull/1740
https://github.com/pytorch/tutorials/pull/1741

PR #1974, opened 2021-11-03
Consolidate network utils

mthrok:
This commit changes all the `torch.hub` network utility functions to
be imported from `torchaudio._internal`, so that later we can replace
the function within fbcode.

Supersedes #1959
See also [D32078361](https://phabricator.intern.facebook.com/D32078361)

### DIFF
--- a/torchaudio/_internal/__init__.py
+++ b/torchaudio/_internal/__init__.py
@@ -1,0 +1,7 @@
+from torch.hub import load_state_dict_from_url, download_url_to_file
+
+
+__all__ = [
+    "load_state_dict_from_url",
+    "download_url_to_file",
+]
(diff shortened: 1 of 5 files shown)


torchaudio/_internal/__init__.py
```diff
@@ -0,0 +1,7 @@
+from torch.hub import load_state_dict_from_url, download_url_to_file
```
nateanl, line 1: wondering why this is different from the diff. Is `_internal` directory not synced?

nateanl (approved):

torchaudio/_internal/__init__.py
carolineechen: looks like the linked diff corresponds to this PR, which has not been merged/synced https://github.com/pytorch/audio/pull/1959/files

carolineechen (approved):

torchaudio/_internal/__init__.py
mthrok: Yup, the diff and #1959 is not merged. I decided to land the minimal change here first.

Issue #1991, opened 2021-11-06
MVDR module can't work in distributed training


nateanl:
### 🐛 Describe the bug

NCCL currently only supports floating point and integer dtype. (See https://github.com/pytorch/pytorch/issues/45760)
The MVDR module forces to use ``cdouble`` dtype to improve the numerical robustness, however, it will cause an runtime error when integrating it to a neural network:
```python
NCCL Backend doesn't support torch.complexdouble data type
```
To solve it, the temporary workaround is relax the dtype constraint by allowing MVDR to use ``cfloat`` dtype.

### Versions

The issue is general to the current PyTorch and torchaudio version.

mthrok:
This makes me wonder ... Should we keep using pseudo complex type? #1337

nateanl:
I tested ``view_as_complex``, it actually has the same functionality. But I made it work by converting the waveform to ``torch.float`` after MVDR beamforming.
So the current MVDR indeed works in distributed training, just need to convert the output to floating point precision. I'm wondering how to check the dtype of gradient in MVDR to make sure it's double instead of float?

nateanl:
The PyTorch that works is built from source. Let me try with a stable version to see how it is like.

nateanl:
I think this issue is solved in latest PyTorch. Using ``torch.cdouble`` inside MVDR works fine within distributed training. I only need to convert it to ``torch.float``after InverseSpectrogram.

After discussion with @fmassa, I'm thinking adding ``torch.cfloat`` is also a good option, if the performance is not degraded. I will verify it by using ``torch.cfloat`` in the tutorial.

nateanl:
Close it since MVDR works in latest PyTorch in distributed training.

PR #1992, opened 2021-11-07
Relax mvdr dtype constraint

nateanl:
To make MVDR work in the neural network distributed training. Address https://github.com/pytorch/audio/issues/1991

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -1665,9 +1665,9 @@ class MVDR(torch.nn.Module):
             (Default: ``False``)
 
     Note:
-        The MVDR Module requires the input STFT to be double precision (``torch.complex128`` or ``torch.cdouble``),
-        to improve the numerical stability. You can downgrade the precision to ``torch.float`` after generating the
-        enhanced waveform for ASR joint training.
+        It's recommended to use double precision (``torch.complex128`` or ``torch.cdouble``) for the input STFT
+        to improve the numerical stability. However, the current NCCL doesn't support double precision for
+        distributed training. The data type is ``torch.complex64`` or ``torch.cfloat`` as the current workaround.
 
     Note:
         If you use ``stv_evd`` solution, the gradient of the same input may not be identical if the
@@ -1948,9 +1948,10 @@ class MVDR(torch.nn.Module):
             raise ValueError(
                 f"Expected at least 3D tensor (..., channel, freq, time). Found: {specgram.shape}"
             )
-        if specgram.dtype != torch.cdouble:
+        if specgram.dtype not in [torch.cfloat, torch.cdouble]:
             raise ValueError(
-                f"The type of ``specgram`` tensor must be ``torch.cdouble``. Found: {specgram.dtype}"
+                f"The type of ``specgram`` tensor must be ``torch.cfloat`` or ``torch.cdouble``.\
+                    Found: {specgram.dtype}"
             )
 
         if mask_n is None:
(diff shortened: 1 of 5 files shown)


fmassa:
Can you run this function in some audio data for both cfloat and cdouble and compare the outputs? This way we will know if running the operator in cfloat can lead to diverging results in some cases due to numerical issues.

nateanl:
The ``stv_power`` solution of MVDR gets ``nan`` outputs if using ``torch.cfloat`` dtype. I need to debug it.
Using ``torch.cdouble``:
```
ref_channel:  15.035907456985868
stv_evd:  16.563734673832553
stv_power:  17.820481909929907
```

Using ``torch.cfloat``:
```
ref_channel:  15.040328979492188
stv_evd:  16.56372833251953
stv_power:  nan
```

nateanl:
The reason why the value becomes ``nan`` is the overflow of cfloat tensor.
```python
a = torch.tensor([-2.7791e+12+2.3049e+11j], dtype=torch.cfloat)
b = torch.tensor([4.5982e+09+512.j], dtype=torch.cfloat)
c = torch.tensor([9.7214e+21], dtype=torch.cfloat)
print(a * b / c.real)
```
@fmassa @mthrok  Do you have any idea to address it?

nateanl:
Addressed the overflow issue by switching the division and multiplication:
```
a / c.real * b
```
The ``stv_power`` solution still downgrades a lot (from 17.82 to 8.32). 
```
ref_channel:  15.040328979492188
stv_evd:  16.56372833251953
stv_power:  8.320789337158203
```

@ Emrys365 Have you had similar experience when using floating point dtype?

Emrys365:
> @Emrys365 Have you had similar experience when using floating point dtype?

Yes! I think this is a common observation in several works, e.g. [Section 4.4 | Heymann et al; 2019](https://ris.uni-paderborn.de/download/12875/15334/ICASSP_2019_Heymann_Paper.pdf) and [Section 2.2 (4) | Zhang et al; 2021](https://arxiv.org/pdf/2102.11525.pdf).

The higher precision is likely to improve the stability in operations such as matrix inverse (this is also involved in the power method).

nateanl:
After offline discussion with @Emrys365, a good workaround is allowing user to use either ``cfloat`` or ``cdouble`` for the input, and the MVDR module will implicitly convert the spectrogram to ``cdouble`` for computation, the output is ``cfloat`` so that users don't need to worry about manually converting the dtype.

nateanl:
The Si-SDR score looks fine now:
```
ref_channel:  15.035911560058594
stv_evd:  16.563732147216797
stv_power:  17.820480346679688
```

mthrok:
@nateanl So my understanding is that, the original issue about `cdouble` not supported in distributed training can be worked around by keeping the use of `cdouble` inside a module. Is that correct?

torchaudio/transforms.py
```diff
@@ -1948,10 +1948,13 @@ def forward(
             raise ValueError(
                 f"Expected at least 3D tensor (..., channel, freq, time). Found: {specgram.shape}"
             )
-        if specgram.dtype != torch.cdouble:
+        if specgram.dtype not in [torch.cfloat, torch.cdouble]:
```
mthrok, line 1951:
```suggestion
        if !specgram.is_complex():
```

nateanl:
@mthrok Actually when I installed the PyTorch from the source, the ``cdouble`` tensor works in distributed training. So the current design is good in terms of functionality.

Here I just want to improve the usability of this module, by allowing user to use the default ``cfloat`` dtype and don't need to worry about the conversion. However, the test shows converting to ``cfloat`` fails to pass the gradcheck, which means the module is very sensitive to the precision of data.

nateanl:
To guarantee the MVDR module passes gradient check, we should keep the current design (i.e., force the users to use ``torch.cdouble`` for input tensor and let users decide whether to convert it to ``torch.cfloat`` for joint training).

In this way, users are aware that double precision is essential to MVDR module, and there's risk of losing gradient precision if they convert the output to floating point precision.

PR #1993, opened 2021-11-07
Update to xavier_uniform and avoid legacy data.uniform_ initialization

### LINKED ISSUE #1820: Use `torch.nn.init` in Tacotron2 weight initialization
In https://github.com/pytorch/audio/blob/13b2349af91fd02810c8813919a5ae08e617919e/torchaudio/models/tacotron2.py#L1054-L1056
we can probably replace the manual weight initialization with `torch.nn.init.xavier_uniform_`. This will make the code a bit easier to understand, and will avoid using the legacy `.data` attribute in a Variable / Tensor.

krishnakalyan3:
Fixes #1820

@mthrok 
I have also noticed that we use `.fill_` and was wondering if we could replace with [this](https://pytorch.org/docs/stable/generated/torch.Tensor.fill_.html).  
https://github.com/pytorch/audio/blob/main/torchaudio/models/wavernn.py#L177 

and when I tried doing that this is the error
```
----> 1 conv.weight.fill_(1)

RuntimeError: a leaf Variable that requires grad is being used in an in-place operation.
```

### DIFF
--- a/torchaudio/models/tacotron2.py
+++ b/torchaudio/models/tacotron2.py
@@ -986,7 +986,7 @@ class Tacotron2(nn.Module):
         self.embedding = nn.Embedding(n_symbol, symbol_embedding_dim)
         std = sqrt(2.0 / (n_symbol + symbol_embedding_dim))
         val = sqrt(3.0) * std
-        self.embedding.weight.data.uniform_(-val, val)
+        torch.nn.init.xavier_uniform_(embedding.weight)
         self.encoder = _Encoder(
             encoder_embedding_dim, encoder_n_convolution, encoder_kernel_size
         )
--- a/torchaudio/models/tacotron2.py
+++ b/torchaudio/models/tacotron2.py
@@ -26,7 +26,6 @@
 # *****************************************************************************
 
 import warnings
-from math import sqrt
 from typing import Tuple, List, Optional, Union
 
 import torch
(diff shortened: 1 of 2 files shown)


nateanl:
Hi @krishnakalyan3, thanks for working on this.

>  if we could replace with [this](https://pytorch.org/docs/stable/generated/torch.Tensor.fill_.html).
https://github.com/pytorch/audio/blob/main/torchaudio/models/wavernn.py#L177

You can use ``torch.nn.init.constant_(tensor, val)`` to initialize the ``conv.weight``.

krishnakalyan3:
I am really not sure why the grad tests and the flake8 tests are failing.

nateanl (approved):
Hi @krishnakalyan3, the PR looks good to me, just need to address the comment then we can merge it. Thanks!

mthrok:
Hi @krishnakalyan3 

Thanks for the help. Can we split the PR for the WaveRNN update?
The two changes seem to me a bit different.

krishnakalyan3:
Will do!

krishnakalyan3:
Thanks for the feedback @mthrok and @nateanl. I have made the necessary changes. Do we know why the CI Circle builds are failing?.

nateanl:
Hi @krishnakalyan3, the unittest failures are not related with the PR.
According to the style check, removing lines with ``math.sqrt`` should resolve it.

nateanl (approved):
LGTM, thanks! I will merge it later today.

nateanl:
@krishnakalyan3 There seems some issue when importing the PR to fbcode. Let me close it and re-open it. Sorry for the inconvenience.

PR #2000, opened 2021-11-09
Add bucketize sampler and dataset for HuBERT Base model training pipeline

nateanl:
The PR adds PyTorch Lightning based training script for HuBERT Base model. There are two iterations of pre-training and 1 iteration of ASR fine-tuning on LibriSpeech dataset.


examples/hubert/dataset/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+        self.exp_dir = Path(exp_dir)
+        tsv_dir = self.exp_dir / "tsv"
+        label_dir = self.exp_dir / "label"
+        f_list, ind_list, len_list = self._get_lists(
+            tsv_dir,
+            dataset,
+            phase_type,
+            min_sample,
+            max_sample
+        )
```
ejguan, line 127: I am wondering about the size of these returned lists? Just want to give a heads up that multiprocessing with python list would cause N times memory usage. N represents the number of workers.

Using NumPy array or Tensor would help sharing data across processes without copy.

examples/hubert/dataset/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+        self.f_list, self.ind_list, self.len_list = f_list, ind_list, len_list
+        self.labels = self._load_label(label_dir, dataset, phase_type)
```
ejguan, line 129: Ditto.

examples/hubert/dataset/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+        waveform, length, label = sample
+        waveforms.append(waveform)
+        lengths.append(length)
+        labels.append(label)
+
+    data = torch.zeros(len(batch), max(lengths))
+    for i in range(len(waveforms)):
+        data[i][0:waveforms[i].shape[1]] = waveforms[i][0]
+    lengths = torch.Tensor(lengths)
+    labels = torch.nn.utils.rnn.pad_sequence(labels, batch_first=True)
```
ejguan, line 213: Would this op be stateless? I mean we would not do backpropogation here.

examples/hubert/dataset/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+        max_token_count: Optional[int] = None,
+        batch_size: Optional[int] = None
+    ) -> None:
+        if max_token_count is not None and batch_size is not None:
+            raise AssertionError(
+                "The ``max_token_count`` and ``batch_size`` can't be both set."
+            )
+        self.data_source = data_source
+        self.max_token_count = max_token_count
+        self.buckets = self._get_buckets(self.data_source, num_buckets)
```
ejguan, line 44: Same comment on List with multiprocessing.

examples/hubert/dataset/dataset.py
```diff
@@ -0,0 +1,214 @@
...
+    Union,
+)
+from torch import Tensor
+import random
+import torch
+import torchaudio
+from torch.utils.data import Dataset, Sampler
+
+
+class BucketizeSampler(Sampler):
```
ejguan, line 16: Even though this would not cause any problems, do you want to make `BucketizeSampler` inheriting `BatchSampler`?

ejguan (commented):
In general, looks good to me. I have a few comments/suggestions below.

examples/hubert/dataset/dataset.py
nateanl: Good point. The size is actually quite large here. I will refactor it by using Tensor.

examples/hubert/dataset/dataset.py
nateanl: Could you elaborate on the stateless? The intention here is to compute the NCE loss batch-wise and apply masking afterwards.

examples/hubert/dataset/dataset.py
ejguan: Emm. The __init__ is also varied between `BucketizeSampler` and `Sampler` right? It will give users hints about not using it as a sampler as the argument of DataLoader.

nateanl: I see. I have verified that changing `Sampler` to `BatchSampler` doesn't affect the running. Also I don't need to set the batch_size in the dataloader.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,317 @@
...
+        self.labels = self._load_labels(label_dir, dataset, phase_type)
+
+    def __len__(self):
+        return len(self.f_list)
+
+    def _get_lists(
+        self,
+        tsv_dir: Path,
+        dataset: str,
+        phase_type: str,
```
carolineechen, line 155: would `subset` be a better name for this?

nateanl: Sounds good. Let me replace with it.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+            buckets[k] = torch.as_tensor(buckets[k], dtype=torch.int)
+        return buckets
+
+    def __iter__(self) -> Iterator[List[int]]:
+        iter_list = []
+        total_len = 0
+        batch = []
+        len_list = self.data_source.len_list
+        if self.max_token_count:
+            for k in self.buckets.keys():
```
ejguan, line 89: Do you also want to do a permutation of keys here to randomize the bucket?

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+                    index = self.buckets[k][i]
+                    if total_len + len_list[index] > self.max_token_count:
+                        iter_list.append(batch)
+                        batch = [index]
+                        total_len = len_list[index]
+                    else:
+                        batch.append(index)
+                        total_len += len_list[index]
+        else:
+            for i in range(self.buckets[k].size(0)):
```
ejguan, line 100: Please correct me if I'm wrong. Where does `k` come from in this clause?

ejguan (approved):
It looks good to me in terms of the implementation of sampler and dataset. Just having several questions about the internal logic of sampler. I won't block you because of my clarification questions.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+            dataset (str): The dataset for training. Options: [``librispeech``, ``librilight``].
+            subset (str): The subset of the dataset. Options: [``train``, ``valid``].
+
+        Returns:
+            (np.array): The numpy arrary that contains the labels for each audio file.
+        """
+        with open(label_dir / f"{dataset}_{subset}.pt") as f:
+            labels = [line.rstrip() for line in f]
+            labels = [[int(ele) for ele in labels[i].split()] for i in self.ind_list]
+        return np.asarray(labels, dtype=object)
```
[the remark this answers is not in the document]
nateanl: Okay, sounds good.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def __getitem__(self, index):
+        waveform = self._load_audio(index)
+        length = waveform.shape[1]
+        label = torch.Tensor(self.labels[index])
+        return (waveform, label, length)
+
+
+class collate_fn_hubert:
```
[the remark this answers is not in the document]
nateanl: You are right. It should be capital case since it's a class

examples/hubert/dataset/hubert_dataset.py
nateanl: Oops, I didn't test the batch size option. I will fix it.

examples/hubert/dataset/hubert_dataset.py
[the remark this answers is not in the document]
ejguan: How about `np.int`? IIRC, to avoid copy, the data structure needs to share a single ref_count. Like numpy array or Tensor, the whole data structure uses one ref_count. For python array, every single element within it is going to have a separate ref_count.
Basically, we want python only copy the pointer toward the data, rather than all data in multiprocessing.

examples/hubert/dataset/hubert_dataset.py
nateanl: Good question. I tested by shuffling the bucket keys. If I choose a high ``bucket_size`` and ``crop`` option instead of ``padding``, it's very likely that the mini-batch will have different number of tokens. So I will keep the keys sorted here, but shuffle the indices inside each bucket.

examples/hubert/dataset/hubert_dataset.py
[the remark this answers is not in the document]
nateanl: I see. I tried with ``np.int`` but it doesn't work if the lists have different lengths. Padding may work, but that may increase the memory usage.
Do you think ``np.empty`` with the max_length will help save the memory?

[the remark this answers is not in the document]
ejguan: How about not splitting string here and using `np.string_`? -> `self.labels` is going to be list of string.
At the time that `__getitem__` is invoked, split each string to integer list and create a tensor from the list.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+        with open(label_dir / f"{dataset}_{subset}.pt") as f:
+            labels = [line.rstrip() for line in f]
+            labels = [[int(ele) for ele in labels[i].split()] for i in self.ind_list]
+        return np.asarray(labels, dtype=object)
+
+    def __getitem__(self, index):
+        waveform = self._load_audio(index)
+        length = waveform.shape[1]
+        label = torch.Tensor(self.labels[index])
```
ejguan, line 230: BTW, pytorch prefers using `torch.tensor()` to create Tensor rather than `torch.Tensor`

ejguan (approved):
LGTM.

carolineechen (approved):

PR #2003, opened 2021-11-10
Add Emformer RNN-T model

hwangjeff:
Adds streaming-capable recurrent neural network transducer (RNN-T) model that uses Emformer for its transcription network. Includes two factory functions — one that allows for building a custom model, and one that builds a preconfigured base model.

### DIFF
--- a/test/torchaudio_unittest/prototype/rnnt_test_impl.py
+++ b/test/torchaudio_unittest/prototype/rnnt_test_impl.py
@@ -1,0 +1,327 @@
+import torch
+from torchaudio_unittest.common_utils import TestBaseMixin, torch_script
+from torchaudio.prototype.rnnt import emformer_rnnt_model
+
+
+class RNNTTestImpl(TestBaseMixin):
+    def _get_input_config(self):
+        return {
+            "batch_size": 8,
+            "max_input_length": 61,
+            "num_symbols": self._get_model_config()["num_symbols"],
+            "max_target_length": 23,
+            "input_dim": self._get_model_config()["input_dim"],
+            "right_context_length": self._get_model_config()["right_context_length"],
+            "encoding_dim": self._get_model_config()["encoding_dim"],
+            "joiner_max_input_length": 61
+            // self._get_model_config()["time_reduction_stride"],
+            "segment_length": self._get_model_config()["segment_length"],
+            "time_reduction_stride": self._get_model_config()["time_reduction_stride"],
+        }
+
+    def _get_model_config(self):
+        return {
+            "input_dim": 80,
+            "encoding_dim": 128,
+            "num_symbols": 256,
+            "segment_length": 16,
+            "right_context_length": 4,
+            "time_reduction_input_dim": 128,
+            "time_reduction_stride": 4,
+            "transformer_num_heads": 4,
+            "transformer_ffn_dim": 64,
+            "transformer_num_layers": 3,
+            "transformer_dropout": 0.0,
+            "transformer_activation": "relu",
+            "transformer_left_context_length": 30,
+            "transformer_max_memory_size": 0,
+            "transformer_weight_init_scale_strategy": "depthwise",
+            "transformer_tanh_on_mem": True,
+            "symbol_embedding_dim": 64,
+            "num_lstm_layers": 2,
+            "lstm_layer_norm": True,
+            "lstm_layer_norm_epsilon": 1e-3,
+            "lstm_dropout": 0.0,
+        }
+
+    def _get_model(self):
+        return (
+            emformer_rnnt_model(**self._get_model_config())
+            .to(device=self.device, dtype=self.dtype)
+            .eval()
+        )
+
+    def _get_transcriber_input(self):
+        input_config = self._get_input_config()
+        batch_size = input_config["batch_size"]
+        max_input_length = input_config["max_input_length"]
+        input_dim = input_config["input_dim"]
+        right_context_length = input_config["right_context_length"]
+
+        input = torch.rand(
+            batch_size, max_input_length + right_context_length, input_dim
+        ).to(device=self.device, dtype=self.dtype)
+        lengths = torch.randint(1, max_input_length + 1, (batch_size,)).to(
+            device=self.device, dtype=torch.int32
+        )
+        return input, lengths
+
+    def _get_transcriber_streaming_input(self):
+        input_config = self._get_input_config()
+        batch_size = input_config["batch_size"]
+        segment_length = input_config["segment_length"]
+        input_dim = input_config["input_dim"]
+        right_context_length = input_config["right_context_length"]
+
+        input = torch.rand(
+            batch_size, segment_length + right_context_length, input_dim
+        ).to(device=self.device, dtype=self.dtype)
+        lengths = torch.randint(
+            1, segment_length + right_context_length + 1, (batch_size,)
+        ).to(device=self.device, dtype=torch.int32)
+        return input, lengths
+
+    def _get_predictor_input(self):
+        input_config = self._get_input_config()
+        batch_size = input_config["batch_size"]
+        num_symbols = input_config["num_symbols"]
+        max_target_length = input_config["max_target_length"]
+
+        input = torch.randint(0, num_symbols, (batch_size, max_target_length)).to(
+            device=self.device, dtype=torch.int32
+        )
+        lengths = torch.randint(1, max_target_length + 1, (batch_size,)).to(
+            device=self.device, dtype=torch.int32
+        )
+        return input, lengths
+
+    def _get_joiner_input(self):
+        input_config = self._get_input_config()
+        batch_size = input_config["batch_size"]
+        joiner_max_input_length = input_config["joiner_max_input_length"]
+        max_target_length = input_config["max_target_length"]
+        input_dim = input_config["encoding_dim"]
+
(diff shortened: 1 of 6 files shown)


test/torchaudio_unittest/prototype/rnnt_test_impl.py
```diff
@@ -0,0 +1,331 @@
...
+            "max_input_length": 61,
+            "num_symbols": self._get_model_config()["num_symbols"],
+            "max_target_length": 23,
+            "input_dim": self._get_model_config()["input_dim"],
+            "right_context_length": self._get_model_config()["right_context_length"],
+            "encoding_dim": self._get_model_config()["encoding_dim"],
+            "joiner_max_input_length": 61
+            // self._get_model_config()["time_reduction_stride"],
+            "segment_length": self._get_model_config()["segment_length"],
+            "time_reduction_stride": self._get_model_config()["time_reduction_stride"],
```
mthrok, line 19: Looks like the result of `self._get_model_config()` can be cached and re-used.

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+        lstm_layer_norm=lstm_layer_norm,
+        lstm_layer_norm_epsilon=lstm_layer_norm_epsilon,
+        lstm_dropout=lstm_dropout,
+    )
+    joiner = _Joiner(encoding_dim, num_symbols)
+    return _RNNT(transcriber, predictor, joiner)
+
+
+def emformer_rnnt_base() -> _RNNT:
+    r"""Builds basic version of Emformer RNN-T model.
```
mthrok, line 758: Is there a corresponding name referred in the original paper?
In https://arxiv.org/abs/2010.10759, I see architectures referred as `EM-24L` and `EM-36L`.

I assume that these factory functions follow the pattern of Wav2Vec2 module. In the case of wav2vec2, the model architectures are referred to as base/large etc in the paper.

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+
+    Returns:
+        _RNNT:
```
mthrok, line 761: Since the returned model is part of public API, `_RNNT` needs to be public as well.

mthrok (commented):
Could you add the new models to the documentation?

https://github.com/pytorch/audio/blob/main/docs/source/prototype.rst

torchaudio/prototype/rnnt.py
hwangjeff: no, there isn't

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+        T_max = num_frames // self.stride
+
+        output = input.reshape(B, T_max, D * self.stride)
+        output = output.contiguous()
+        return output, lengths
+
+
+class _CustomLSTM(torch.nn.Module):
+    r"""Custom long-short-term memory (LSTM) block that additionally applies
+    linear projection and layer normalization to internal nodes.
```
nateanl, line 58: Would be great to describe the details in the `_CustomLSTM`. If I understand correctly, the difference than conventional LSTM is that the bias of layer `x2g` (if layer norm is applied) and `p2g` are removed, The input gate `i_t` and cell gate  `c_t` are layer-normed?

Technically it's not `additionally linear projection` since LSTM also uses linear weights (but with bias) for input gates and hidden states, but an equation or a description will fully explain it.

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+            outputs.append(h)
+
+        output = torch.stack(outputs, dim=0)
+        state = [h, c]
+
+        return output, state
+
+
+class _Transcriber(torch.nn.Module):
+    r"""Recurrent neural network transducer (RNN-T) transcription network.
```
nateanl, line 138: This is more like an Emformer component of RNN-T transcription network :)?

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+                    layer_norm=lstm_layer_norm,
+                    layer_norm_epsilon=lstm_layer_norm_epsilon,
+                )
+                for idx in range(num_lstm_layers)
+            ]
+        )
+        self.linear = torch.nn.Linear(symbol_embedding_dim, output_dim)
+        self.output_layer_norm = torch.nn.LayerNorm(output_dim)
+
+        self.lstm_dropout = lstm_dropout
```
nateanl, line 329: A module-level should be more standard? 
`self.dropout = torch.nn.Dropout (p=lstm_dropout)`

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+        input_tb = input.permute(1, 0)
+        embedding_out = self.embedding(input_tb)
+        input_layer_norm_out = self.input_layer_norm(embedding_out)
+
+        lstm_out = input_layer_norm_out
+        state_out: List[List[torch.Tensor]] = []
+        for layer_idx, lstm in enumerate(self.lstm_layers):
+            lstm_out, lstm_state_out = lstm(
+                lstm_out, None if state is None else state[layer_idx]
+            )
```
nateanl, line 372: The logic here doesn't look correct. `lstm_state_out` is not updated as `lstm_out`?

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+            target_encodings (torch.Tensor): target encoding sequences, with shape `(B, U, D)`.
+            target_lengths (torch.Tensor): with shape `(B,)` and i-th element representing
+                valid sequence length of i-th batch element in ``target_encodings``.
+
+        Returns:
+            (torch.Tensor, torch.Tensor, torch.Tensor):
+                torch.Tensor
+                    joint network output, with shape `(B, T, U, D)`.
+                torch.Tensor
+                    output input lengths, with shape `(B,)` and i-th element representing
```
nateanl, line 423:
```suggestion
                    output lengths, with shape `(B,)` and i-th element representing
```

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+                    number of valid elements along dim 1 for i-th batch element in joint network output.
+                torch.Tensor
+                    output target lengths, with shape `(B,)` and i-th element representing
+                    number of valid elements along dim 2 for i-th batch element in joint network output.
+        """
+        joint_encodings = (
+            input_encodings.unsqueeze(2).contiguous()
+            + target_encodings.unsqueeze(1).contiguous()
+        )
+        relu_out = torch.nn.functional.relu(joint_encodings)
```
nateanl, line 433: Same here, we can use Module for ReLU :)

torchaudio/prototype/rnnt.py
hwangjeff: yeah the language could be more precise — will revise

torchaudio/prototype/rnnt.py
hwangjeff: for now, yeah. i anticipate once we add in other transformer flavors that we can generalize this beyond emformer

torchaudio/prototype/rnnt.py
hwangjeff: can you elaborate on this?

torchaudio/prototype/rnnt.py
hwangjeff: i added the "input" bit with the intent to clarify that these output lengths correspond to `input`. perhaps i can change `input` to something like `utterance` or `source` to reduce the ambiguity? what do you think?

torchaudio/prototype/rnnt.py
nateanl: The `lstm_out` of the last layer is used as the input for the next layer, while ``lstm_state_out`` is not used in this way. I'm wondering if it's expected.

torchaudio/prototype/rnnt.py
nateanl: Yeah that's a good idea.

torchaudio/prototype/rnnt.py
hwangjeff: got it. yeah this is expected — we maintain state independently on a per-layer basis and provide each layer with the state it generated in the previous timestep

torchaudio/prototype/rnnt.py
```diff
@@ -0,0 +1,786 @@
...
+            source_encodings.unsqueeze(2).contiguous()
+            + target_encodings.unsqueeze(1).contiguous()
+        )
+        relu_out = self.relu(joint_encodings)
+        output = self.linear(relu_out)
+        return output, source_lengths, target_lengths
+
+
+class RNNT(torch.nn.Module):
+    r"""Recurrent neural network transducer (RNN-T) model.
```
mthrok, line 439: https://435773-90321822-gh.circle-artifacts.com/0/docs/prototype.html#torchaudio.prototype.RNNT

Can you add a note to redirect users to the factory functions? Also the doctoring referring to private classes is strange, can you replace them with `torch.nn.Module`? (I would change type annotation as well but that might break TorchScript, so that's fine.)

ref: https://pytorch.org/audio/stable/models.html#wav2vec2model

mthrok (approved):

PR #2004, opened 2021-11-11
Improve MVDR stability

nateanl:
Division first, multiplication second. This helps avoid the value overflow issue. It also helps the ``stv_evd`` solution pass the gradient check.

### DIFF
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -278,8 +278,7 @@ class AutogradTestMixin(TestBaseMixin):
         "ref_channel",
         # stv_power test time too long, comment for now
         # "stv_power",
-        # stv_evd will fail since the eigenvalues are not distinct
-        # "stv_evd",
+        "stv_evd",
     ])
     def test_mvdr(self, solution):
         transform = T.MVDR(solution=solution)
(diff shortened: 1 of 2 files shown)


nateanl:
cc @Emrys365 @popcornell This may also be useful for ESPNet and Asteroid.

nateanl:
Looks like linux_gpu test fails due to ``stv_evd`` solution running for a long time. But the CPU tests show that the ``stv_evd`` solution passes gradient check.

Comment it to move forward.

test/torchaudio_unittest/transforms/autograd_test_impl.py
```diff
@@ -276,10 +276,9 @@ def test_psd_with_mask(self, multi_mask):
 
     @parameterized.expand([
         "ref_channel",
-        # stv_power test time too long, comment for now
+        # stv_power and stv_evd test time too long, comment for now
         # "stv_power",
-        # stv_evd will fail since the eigenvalues are not distinct
-        # "stv_evd",
+        "stv_evd",
```
hwangjeff, line 281: should this be commented out per the comment above?

nateanl: Oh you are right lol. Not sure what I was doing here.

PR #2006, opened 2021-11-11
Fix style checks in examples/tutorials

nateanl:


examples/tutorials/audio_data_augmentation_tutorial.py
```diff
@@ -10,6 +10,11 @@
 # with the following.
 # !pip install torchaudio
 
+from IPython.display import Audio, display
+import matplotlib.pyplot as plt
+import requests
+import os
+import math
```
mthrok, line 17: The imports are separated intentionally. Can we work around this?

examples/tutorials/audio_data_augmentation_tutorial.py
```diff
@@ -22,149 +27,157 @@
...
 SAMPLE_RIR_PATH = os.path.join(_SAMPLE_DIR, "rir.wav")
 
-SAMPLE_WAV_SPEECH_URL = "https://pytorch-tutorial-assets.s3.amazonaws.com/VOiCES_devkit/source-16k/train/sp0307/Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.wav"
+SAMPLE_WAV_SPEECH_URL = "https://pytorch-tutorial-assets.s3.amazonaws.com/VOiCES_devkit/"\
+    "source-16k/train/sp0307/Lab41-SRI-VOiCES-src-sp0307-ch127535-sg0042.wav"
 SAMPLE_WAV_SPEECH_PATH = os.path.join(_SAMPLE_DIR, "speech.wav")
 
-SAMPLE_NOISE_URL = "https://pytorch-tutorial-assets.s3.amazonaws.com/VOiCES_devkit/distant-16k/distractors/rm1/babb/Lab41-SRI-VOiCES-rm1-babb-mc01-stu-clo.wav"
+SAMPLE_NOISE_URL = "https://pytorch-tutorial-assets.s3.amazonaws.com/VOiCES_devkit/"\
+    "distant-16k/distractors/rm1/babb/Lab41-SRI-VOiCES-rm1-babb-mc01-stu-clo.wav"
```
mthrok, line 57: Did you use tool or do it manually? Normally, URL-like strings are exempted from formatter line length rules.

examples/tutorials/forced_alignment_tutorial.py
```diff
@@ -179,84 +180,85 @@ def get_trellis(emission, tokens, blank_id=0):
...
+        # Return token index and time index in non-trellis coordinate.
+        path.append(Point(j - 1, t - 1, prob))
+
+        # 3. Update the token
+        if changed > stayed:
+            j -= 1
+            if j == 0:
+                break
+    else:
+        raise ValueError('Failed to align')
```
nateanl, line 246: @mthrok The indentation here might be wrong, but aligning it with ``if`` will throw ValueError. So I kept it as a follow-up.

examples/tutorials/audio_data_augmentation_tutorial.py
nateanl: Sure. I will revert it to the original version.

examples/tutorials/audio_data_augmentation_tutorial.py
nateanl: I did it manually, to solve D32280635

examples/tutorials/forced_alignment_tutorial.py
mthrok: `for ~ else` is the syntax unique to Python. It will be triggered if the loop does not break, which should never happen here.

See https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops.

examples/tutorials/audio_data_augmentation_tutorial.py
mthrok: Okay, please use `black` or apply the internal linter.
Considering the number of lines changed, manual change is error-prune and hard to verify.

Long URLs should be exempted from line lengths. Please use  `# noqa: E501` at the end of each line that's too long. See for example 

https://github.com/pytorch/audio/blob/e6dd7fd3ef3e07df92b0f46180b3d5888e59c014/torchaudio/pipelines/_wav2vec2/impl.py#L41-L45

nateanl:
Here is the demo page: https://433128-90321822-gh.circle-artifacts.com/0/docs/tutorials/speech_recognition_pipeline_tutorial.html

mthrok (approved):

PR #2008, opened 2021-11-12
Add script for classifying release note PRs

carolineechen:
To generate release notes:
1.`python tools/release_notes/retrieve_prs.py 1d3dcdbd70fa53cf45a55334a70f9fc46c7e4bdb 8202c7983a90fa9cc58f0e3ba49fa6148ce18be8` retrieves PRs between the commits into specified output file (default is data.json)
2. `python tools/release_notes/classify_prs.py` organizes PRs based on labels into 

Sample rendered [notebook](https://gist.github.com/carolineechen/73c81fb2b12434e4dfe55271986aadb0)

Adapted from [torchvision's](https://github.com/pytorch/vision/blob/main/scripts/release_notes/classify_prs.py)


tools/release_notes/classify_prs.py
```diff
@@ -0,0 +1,83 @@
...
+                row[col_name] = label
+                break
+
+
+# In[5]:
+
+
+get_labels("primary_label", primary_labels_mapping.keys())
+get_labels("secondary_label", secondary_labels_mapping.keys())
+df.tail(5)
```
mthrok, line 64: Is `tail(5)` here necessary? It seems already printed in the previous block.

mthrok (approved):

tools/release_notes/classify_prs.py
carolineechen: this block extracts and adds the primary and secondary columns to the df, so calling tail again will additionally show these 2 columns as well. this is only relevant if running and displaying on a notebook, and not when running the script as it's not being printed

PR #2021, opened 2021-11-21
oscillator_bank op

0x00b1:


mthrok:
Hi @0x00b1 

Thanks for the contribution. Could you share a resource to learn about the underlying algorithms and use cases, so that we can study the nature of the op?

0x00b1:
Hi, @mthrok! The Wikipedia [article](https://en.wikipedia.org/wiki/Additive_synthesis) on additive synthesis is fantastic.

mthrok:
> Hi, @mthrok! The Wikipedia [article](https://en.wikipedia.org/wiki/Additive_synthesis) on additive synthesis is fantastic.

Thanks. Can this function generate synthetic sounds like the bell in the article? If so that's cool and that's something we want to add in tutorial / examples. Also that will help me understand the use case.
https://pytorch.org/audio/main/tutorials/audio_data_augmentation_tutorial.html

0x00b1:
@mthrok Yep! The unit test does just that. Try playing the output.

mthrok:
Hi @0x00b1 

I finally had some time to try this `oscillatro_bank` op and I could generate a bell-like sound, following the configuration from https://computermusicresource.com/Simple.bell.tutorial.html.

https://user-images.githubusercontent.com/855818/144518573-0da131bb-d13d-468d-8d6b-8910a196ab64.mp4

```python

import torch
import torchaudio
import torchaudio.functional as F

import math

sample_rate = 16000

def _get_envelop(attack_time, attack_level, decay, duration, sample_rate):
    # Create envelop
    # no sustain or release
    total_length = int(duration * sample_rate)
    attack_length = int(sample_rate * attack_time)
    decay_length = total_length - attack_length

    # decay formula:
    # f(t) = a * exp(-b *t)
    # f(0) = a == attack_level
    # f(duration) = a * exp(-b * duration) == (1 - decay) * attack_level
    # b == -ln(1 - decay) / duration.
    # f(t) = attack_level * exp(ln(1-decay) * t / duration)

    attack_region = torch.linspace(0, attack_level, attack_length)
    t = torch.linspace(0, decay_length / sample_rate, decay_length)
    decay_region = attack_level * torch.exp(math.log(1 - decay) * t / duration)
    return torch.cat([attack_region, decay_region])

envelop = _get_envelop(
    attack_time=0.002,
    attack_level=0.5,
    decay=0.998,
    duration=6.0,
    sample_rate=sample_rate,
)[None, ...]


freq0 = 350.0
freq_multipliers = [
    0.56,
    0.92,
    1.19,
    1.71,
    2.,
    2.74,
    3,
    3.76,
    4.07,
]


frequencies, amplitudes = [], []
for i, mult in enumerate(freq_multipliers):
    frequencies.append(torch.full(envelop.shape, freq0 * mult))
    amplitudes.append(envelop * (0.5 ** i))
frequencies = torch.cat(frequencies)
amplitudes = torch.cat(amplitudes)

frequencies = frequencies.T[None, ...]
amplitudes = amplitudes.T[None, ...]
print('freqs:', frequencies.shape)
for i in range(len(freq_multipliers)):
    print(f'freqs[{i}]:', frequencies[0, ::, i])
print('amps:', amplitudes.shape)
for i in range(amplitudes.size(2)):
    print(f'amps[{i}]:', amplitudes[0, ::, i])

waveform = F.oscillator_bank(
    frequencies,
    amplitudes,
    sample_rate
)

print(amplitudes)
print(waveform.shape)
torchaudio.save(waveform.wav', waveform, sample_rate)

```

</details>

The following is my thoughts while trying it out.
1. From the perspective of library function, I do not think there is much advantage of requiring separate `batch` dimension and `sinusoid`.
2. There are couple of `bank` operators in torchaudio, which generates series of filters. I feel that `oscillator` bank can do the same and that is more primitive and suitable for library implementation. 
3. Looking at the test code and my trial script, the current implementation requires users to use `torch.ones` many times, which suggests to me that if the function takes the parameters underneath these `frequencies` and `amplitudes` and populate the tensor for users, the user experience is improved.
  - Is it a common use case to consider frequencies that change across time? Looking at the test code, that seems not to be the case, if so, it is tedious to have users make frequency matrix, where column values are same. 
4. In `torchaudio`, generally, the dimension order is time last.

Following these, I think the following signature provides a somewhat simpler UX.

```python

def oscillator_bank(envelops, frequencies, sample_rate) -> Tensor:
    """
    Args:
        envelops (Tensor): shape of (num_envelops, num_time_frames) 
        frequencies (Tensor): shape of (num_frequencies, )
        sample_rate (float):
    Returns:
        Tensor: shape of (num_envelops, num_frequencies, num_time_frames)
    """
```

and if users want to mix the resulting sinusoid, they can do `matmul` (or we can provide another API that does this)

What do you think?

0x00b1:
@mthrok Great suggestions! I'll have a commit with the changes soon.

PR #2024, opened 2021-11-22
Relax dtype for MVDR

nateanl:
Allow users to use `torch.cfloat` dtype input for MVDR module. It internally convert the spectrogram into `torch.cdouble` and output the tensor with the original dtype of the spectrogram.


nateanl:
> LGTM.
> 
> This looks like backward-compatible to me as `cdouble` will keep working. Can you double check?

Sure. The auto_grad test passed on my local machine. The reason why it failed last time is that I hard coded the output to ``cfloat``. 

I will merge after the unit tests finish and pass.

PR #2028, opened 2021-11-23
Add RNN-T beam search decoder

hwangjeff:
Adds beam search decoder for RNN-T implementation ``torchaudio.prototype.RNNT`` that is TorchScript-able and supports both streaming and non-streaming inference.


torchaudio/prototype/rnnt_decoder.py
```diff
@@ -0,0 +1,342 @@
...
+
+import torch
+
+from .rnnt import RNNT
+
+
+__all__ = ["RNNTBeamSearch"]
+
+
+class Hypo(NamedTuple):
```
mthrok, line 11: Since this is a public API, could you add docstring?
It might require `:ivar` to document attributes.
cf. https://github.com/pytorch/audio/blob/05ae795ae51c5f72ef7747cd3c06fadccff97e5a/torchaudio/backend/common.py#L1

torchaudio/prototype/rnnt_decoder.py
```diff
@@ -0,0 +1,342 @@
...
+class RNNTBeamSearch(torch.nn.Module):
+    r"""Beam search decoder for RNN-T model.
+
+    Args:
+        model (RNNT): RNN-T model to use.
+        blank (int): index of blank token in vocabulary.
+        temperature (float, optional): temperature to apply to joint network output.
+            Larger values yield more uniform samples. (Default: 1.0)
+        hypo_sort_key (Callable[[Hypo], float], optional): callable that computes a score for a given
+            hypothesis to rank hypotheses by. (Default: _default_hypo_sort_key)
```
mthrok, line 80: 1. Referring to private function in doctoring should be avoided.
2. I think it is better to explain the default behavior in the doctoring.

Therefore I suggest to use the signature, like `hypo_sort_key: Optional[Callable[[Hypo], float]] = None`, and explain the default sort behavior in plain expression. (I think it is average score?)

torchaudio/prototype/rnnt_decoder.py
```diff
@@ -0,0 +1,342 @@
...
+        enc_out, _ = self.model.transcribe(input, length)
+        return self._search(enc_out, None, beam_width)
+
+    @torch.jit.export
+    def infer(
+        self,
+        input: torch.Tensor,
+        length: torch.Tensor,
+        state: Optional[List[List[torch.Tensor]]],
+        hypo: Optional[Hypo],
```
mthrok, line 311: Without a default value of `= None`, isn't this mandatory argument?

torchaudio/prototype/rnnt_decoder.py
mthrok: nit: Also, I feel it's better to spell out `Hypothesis`, but it's subjective.

torchaudio/prototype/rnnt_decoder.py
hwangjeff: yeah it is — changed `state` and `hypo` to kwargs

torchaudio/prototype/rnnt_decoder.py
```diff
@@ -0,0 +1,361 @@
...
+        state: Optional[List[List[torch.Tensor]]] = None,
+        hypothesis: Optional[Hypothesis] = None,
+    ) -> Tuple[List[Hypothesis], List[List[torch.Tensor]]]:
+        r"""Performs beam search for the given input sequence in streaming mode.
+
+        T: number of frames;
+        D: feature dimension of each frame.
+
+        Args:
+            input (torch.Tensor): sequence of input frames, with shape (1, T, D).
```
mthrok, line 338: Is the first dimension something naturally originated from the RNN-T model output?
Or is it there to ease the following hypothesis generation process?

If it's for the later, and if users have to do extra `.unsqueeze(0)` in the client code, then I think it's more straightforward to accept 2D input. otherwise it's fine.

hwangjeff: good point — this is more a consequence of the rnn-t transcription network expecting an input of shape (B, T, D), but we don't have to make users adhere to this. perhaps accept both (1, T, D) and (T, D)?

mthrok: sounds good to me.

torchaudio/prototype/rnnt_decoder.py
```diff
@@ -0,0 +1,377 @@
...
+        assert input.dim() == 2 or (
+            input.dim() == 3 and input.shape[0] == 1
+        ), "input must be of shape (T, D) or (1, T, D)"
+        if input.dim() == 2:
+            input = input.unsqueeze(0)
+
+        assert length.shape == () or length.shape == (
+            1,
+        ), "length must be of shape () or (1,)"
+        if input.dim() == 0:
```
csukuangfj, line 325: In which case is `input.dim() == 0`?

PR #2030, opened 2021-11-23
Update script for getting PR merger and labels

carolineechen:
The previous way of detecting the merger and labels given a commit hash no longer works with ShipIt, as PRs are closed and not merged and are not associated with a commit hash. To work around this, update the script to get the merger (pulled by: ) and PR number from the commit hash message, and then collect labels from the corresponding PR.


mthrok (approved):
Stamp

hwangjeff (approved):

PR #2038, opened 2021-11-27
Refactor the library loading mechanism

mthrok:
(This is a part of refactor series, followed up by #2039 and #2040.
The goal is to make it easy to add a new library artifact alongside with `libtorchudio`, as in https://github.com/pytorch/audio/pull/2048/commits/4ced990849e60f6d19e87ae22819b04d1726648e #2048 .)

We plan to add prototype/beta third party library integrations,
which could be unstable. (segfault, missing dynamic library dependencies etc...)

If we add such integrations into the existing libtorchaudio,
in the worst case, it will prevent users from just `import torchaudio`.

Instead, we would like to separate the prototype/beta integrations
into separate libraries, so that such issues would not impact all users but
users who attempt to use these prototytpe/beta features.

Say, a prototype feature `foo` is added in `torchaudio.prototype.foo`.
The following initialization procedure will achieve the above mechanism.

1. Place the library file `libtorchaudio_foo` in `torchaudio/lib`.
2. In `torchaudio.prototype.foo.__init__.py`, load the `libtorchaudio_foo`.

Note:
The approach will be slightly different for fbcode, because of how buck deploys
C++ libraries and standardized environment, but the code change here is still
applicable.

### DIFF
--- a/torchaudio/_extension.py
+++ b/torchaudio/_extension.py
@@ -24,6 +24,7 @@ def _load_lib(lib: str):
         torch.ops.load_library(path)
         torch.classes.load_library(path)
 
+
 def _init_extension():
     if not _mod_utils.is_module_available('torchaudio._torchaudio'):
         warnings.warn('torchaudio C++ extension is not available.')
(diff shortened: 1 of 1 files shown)


nateanl (approved):
#stamp

hwangjeff (approved):

PR #2039, opened 2021-11-27
Clean up libtorchaudio customization logic

mthrok:
(See #2038 description for the overall goal.)
This PR cleans up CMake customization logic for `libtorchaudio`.

It introduces base variables LIBTORCHAUDIO_INCLUDE_DIRS,
LIBTORCHAUDIO_LINK_LIBRARIES and LIBTORCHAUDIO_COMPILE_DEFINITIONS,
which are respectively used when calling `target_include_directories`,
`target_link_libraries` and `target_compile_definitions`.

The customization logic only modifies these variables.

The original implementation called these functions multiple times
(once par customization logic) and it is getting difficult to understand
the customization logic.

### DIFF
--- a/torchaudio/csrc/CMakeLists.txt
+++ b/torchaudio/csrc/CMakeLists.txt
@@ -10,6 +10,23 @@ set(
   utils.cpp
   )
 
+set(
+  LIBTORCHAUDIO_INCLUDE_DIRS
+  ${PROJECT_SOURCE_DIR}
+  )
+
+set(
+  LIBTORCHAUDIO_LINK_LIBRARIES
+  torch
+  ${TORCHAUDIO_THIRD_PARTIES}
+  )
+
+set(
+  LIBTORCHAUDIO_COMPILE_DEFINITIONS)
+
+#------------------------------------------------------------------------------#
+# START OF CUSTOMIZATION LOGICS
+#------------------------------------------------------------------------------#
 if(BUILD_RNNT)
   list(
     APPEND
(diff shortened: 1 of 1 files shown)


nateanl (approved):

torchaudio/csrc/CMakeLists.txt
```diff
@@ -59,45 +112,21 @@ set_target_properties(libtorchaudio PROPERTIES PREFIX "")
...
-
 target_link_libraries(
   libtorchaudio
-  torch
-  ${TORCHAUDIO_THIRD_PARTIES}
+  ${LIBTORCHAUDIO_LINK_LIBRARIES}
+  )
+target_compile_definitions(
+  libtorchaudio
+  PRIVATE ${LIBTORCHAUDIO_COMPILE_DEFINITIONS}
```
hwangjeff, line 123: what are the consequences of making the scope of `INCLUDE_SOX` and `INCLUDE_KALDI` `PRIVATE` rather than `PUBLIC` as before?

hwangjeff (approved):

torchaudio/csrc/CMakeLists.txt
mthrok: The direct impact is that, previously `_torchaudio` was inheriting these definitions automatically, so they did not need to have the definition again.

The indirect impact I can think of is that if third party project is using this CMakeLists to integrate `libtorchaudio`, and they customize the build process of their application alongside `libtorchaudio`, using the same compiler definition `INCLUE_SOX` and `INCLUDE_KALDI`,  then they have to add the same compile definition by them selves, as it will be no longer visible.

My justification is that, since the third party library has control over whether to include these dependencies, they have enough information to do so by themselves. Plus, even if they are customizing their application in similar manner as `libtorchaudio`, they can pick arbitral names for their compiler definitions. There is no obligation that they have to pick the same definition. So I do not think they need to be public. If we want to make them public, then I think we need to prefix the compiler definition with `LIBTORCHAUDIO_` to reduce the likelihood of name collision.

PR #1986, opened 2021-11-05
Add ffmpeg integration

mthrok:
This PR adds prototype streaming API based on ffmpeg.
Currently, the Python surface is minimum implementation.
It needs refinement.

- [x] Handles audio
   - [x] Can change the sampling rate on the fly
   - [x] Can change the sample format on the fly.
       - [x] (Implemented and tested) `uint8`, `int16`, `int32`, and `float32`
       - [ ] (Implemented but not tested) `int64` and `float64`.
   - [ ] ~Can change the resampling algorithm~ Covered via custom filter
- [x] Handles video
   - [x] Can change the image resolution on the fly.
       - [ ] ~Can change the rescaling algorithm.~ Covered via custom filter
   - [x] Can change the frame rate on the fly.
       - [ ] ~Can change the algorithm for changing the frame rate.~ Covered via custom filter
- [x] Handles multiple input streams.
- [x] Handles multiple output streams.
   - [x] Can duplicate input streams for different output configurations.
- [x] avdevice integration
- [x] Seek mechanism

- [x] Figure out the way to build / package.
  - [x] Split the binary `libtorchaudio` and `libtorchaudio_ffmpeg`.
  - [x] If binding dynamically, figure out the best way to detect installed ffmpeg across platforms. (`pkg-config` seems straightforward.)
- [ ] Add API similar to existing `torchaudio.load` function for simple use cases.
  - [x] Add the function definition
  - [ ] Add the parameters

PR #2043, opened 2021-11-28
Add FilterGraph class

mthrok:

Add FilterGraph class that is responsible for handling AVFilterGraph structure and the application of filters.

Needs to be imported after #2042.


carolineechen (approved):
stamp to unblock

PR #2044, opened 2021-11-28
Add Buffer class

mthrok:

Add Buffer class that is responsible for converting `AVFrame` to `Tensor`.
Note: The API to retrieve the buffered Tensors is tentative.

Needs to be imported after #2043.

### DIFF
--- a/torchaudio/csrc/ffmpeg/buffer.cpp
+++ b/torchaudio/csrc/ffmpeg/buffer.cpp
@@ -1,0 +1,163 @@
+#include <torchaudio/csrc/ffmpeg/buffer.h>
+#include <stdexcept>
+#include <vector>
+
+namespace torchaudio {
+namespace ffmpeg {
+
+Buffer::Buffer(AVMediaType type) : media_type(type) {}
+
+namespace {
+torch::Tensor convert_audio_tensor(AVFrame* pFrame) {
+  // ref: https://ffmpeg.org/doxygen/4.1/filter__audio_8c_source.html#l00215
+  AVSampleFormat format = static_cast<AVSampleFormat>(pFrame->format);
+  int num_channels = pFrame->channels;
+  int bps = av_get_bytes_per_sample(format);
+
+  // Note
+  // FFMpeg's `nb_samples` represnts the number of samples par channel.
+  // This corresponds to `num_frames` in torchaudio's notation.
+  // Also torchaudio uses `num_samples` as the number of samples
+  // across channels.
+  int num_frames = pFrame->nb_samples;
+
+  int is_planar = av_sample_fmt_is_planar(format);
+  int num_planes = is_planar ? num_channels : 1;
+  int plane_size = bps * num_frames * (is_planar ? 1 : num_channels);
+  std::vector<int64_t> shape = is_planar
+      ? std::vector<int64_t>{num_channels, num_frames}
+      : std::vector<int64_t>{num_frames, num_channels};
+
+  torch::Tensor t;
+  uint8_t* ptr = NULL;
+  switch (format) {
+    case AV_SAMPLE_FMT_U8:
+    case AV_SAMPLE_FMT_U8P: {
+      t = torch::empty(shape, torch::kUInt8);
+      ptr = t.data_ptr<uint8_t>();
+      break;
+    }
+    case AV_SAMPLE_FMT_S16:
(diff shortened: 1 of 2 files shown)


torchaudio/csrc/ffmpeg/buffer.h
```diff
@@ -0,0 +1,24 @@
...
+#include <torch/torch.h>
+#include <torchaudio/csrc/ffmpeg/ffmpeg.h>
+#include <deque>
+
+namespace torchaudio {
+namespace ffmpeg {
+
+class Buffer {
+  std::deque<torch::Tensor> chunks;
+  AVMediaType media_type;
```
nateanl, line 11: Shall we add reference to the variables?
```suggestion
  std::deque<torch::Tensor>& chunks;
  AVMediaType& media_type;
```

torchaudio/csrc/ffmpeg/buffer.cpp
```diff
@@ -0,0 +1,163 @@
...
+  }
+  return torch::cat(tmp, 0);
+}
+
+void Buffer::push_frame(AVFrame* frame) {
+  switch (media_type) {
+    case AVMEDIA_TYPE_AUDIO:
+      push_audio_frame(frame);
+      break;
+    case AVMEDIA_TYPE_VIDEO:
```
nateanl, line 154: Where are ``AVMEDIA_TYPE_AUDIO`` and ``AVMEDIA_TYPE_VIDEO`` defined? Or it's taken care  by ffmpeg?

torchaudio/csrc/ffmpeg/buffer.cpp
```diff
@@ -0,0 +1,163 @@
...
+namespace torchaudio {
+namespace ffmpeg {
+
+Buffer::Buffer(AVMediaType type) : media_type(type) {}
+
+namespace {
+torch::Tensor convert_audio_tensor(AVFrame* pFrame) {
+  // ref: https://ffmpeg.org/doxygen/4.1/filter__audio_8c_source.html#l00215
+  AVSampleFormat format = static_cast<AVSampleFormat>(pFrame->format);
+  int num_channels = pFrame->channels;
```
nateanl, line 14: I feel like ``num_channels`` can be defined when initializing the Buffer object? As each Buffer object will receive signal from one Streamer, which fixed the num_channels after initialization.

torchaudio/csrc/ffmpeg/buffer.cpp
```diff
@@ -0,0 +1,163 @@
...
+    ptr += plane_size;
+  }
+  if (is_planar)
+    t = t.t();
+  return t;
+}
+} // namespace
+
+void Buffer::push_audio_frame(AVFrame* pFrame) {
+  chunks.push_back(convert_audio_tensor(pFrame));
```
nateanl, line 86: Do we want to check the size of ``chunks``  so that if it exceeds the buffer size, we might want to pop out the last chunk?

torchaudio/csrc/ffmpeg/buffer.h
mthrok: It's just an enum, so it's cheap and safer to copy.

torchaudio/csrc/ffmpeg/buffer.cpp
mthrok: Yeah, let me follow-up this one. I am still figuring out a safe way to process chunks of different rates at the same time. (sample rate / frame rate)

torchaudio/csrc/ffmpeg/buffer.cpp
mthrok: https://ffmpeg.org/doxygen/4.1/group__lavu__misc.html#ga9a84bba4713dfced21a1a56163be1f48

torchaudio/csrc/ffmpeg/buffer.cpp
mthrok: I agree with you, but unfortunately the shape information is not necessarily available at the time Buffer object is initialized, due to a filter graph operation inserted before the buffer, which can change number of audio channels, image size, image channels and rate. The structure passed around does not have the explicit information of these.

Let me figure this out later. along with the buffer size concern you brought up.

carolineechen (approved):

PR #2047, opened 2021-11-28
Add ffmpeg prototype bindings

mthrok:

Add `Streamer` TorchBind.

Needs to be imported after #2046.

### DIFF
--- a/torchaudio/csrc/ffmpeg/prototype.cpp
+++ b/torchaudio/csrc/ffmpeg/prototype.cpp
@@ -1,0 +1,329 @@
+#include <torch/script.h>
+#include <torchaudio/csrc/ffmpeg/streamer.h>
+
+namespace torchaudio {
+namespace ffmpeg {
+
+namespace {
+
+std::map<std::string, std::string> convert_dict(
+    const c10::optional<c10::Dict<std::string, std::string>>& option) {
+  std::map<std::string, std::string> opts;
+  if (option) {
+    for (auto& it : option.value()) {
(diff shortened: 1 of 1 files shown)


torchaudio/csrc/ffmpeg/prototype.cpp
```diff
@@ -0,0 +1,329 @@
...
+    std::string, // codec long name
+    std::string, // format name
+    int64_t, // bit_rate
+    // Audio
+    double, // sample_rate
+    int64_t, // num_channels
+    // Video
+    int64_t, // width
+    int64_t, // height
+    double // frame_rate
```
hwangjeff, line 50: would this be better represented as e.g. a struct? there are many fields to keep track of

hwangjeff (approved):

torchaudio/csrc/ffmpeg/prototype.cpp
mthrok: This is the part where the data is converted from C++ to Python.
We would need to register a custom class, then add retrieval methods (or functions).

Adding a class is not a bullet-proof. https://github.com/pytorch/pytorch/issues/65000
Then, the class would not register property, so we need to add `getter` methods, calling them from Python resulted in Java-like code.

An alternative could be to use dictionary format, but that in that case, the key names are not strictly-typed, so it can be still error-prune. Although it has a potential to forward-compatible if the Python function using this protocol is ever be TorchScript-ed.

PR #2048, opened 2021-11-28
Add a switch to build ffmpeg binding

mthrok:
This PR adds `BUILD_FFMPEG` switch to torchaudio build process so that features related to ffmpeg are built.
The flag is false by default, so no CI jobs or development flow are affected.

This is because handling the dependencies around ffmpeg is a bit tricky.
Currently, the CMake file uses `pkg-config` to find an ffmpeg installation in the system.
This works fine for both conda-based installation and system-managed installation (like `apt`).

In subsequent PRs, I will find a solution that works for local development and binary distributions.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -14,6 +14,26 @@ if (BUILD_SOX)
 endif()
 
 ################################################################################
+# ffmpeg
+################################################################################
+if (BUILD_FFMPEG)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(LIBAV REQUIRED IMPORTED_TARGET
+    # requires ffmpeg>=4.1
+    libavdevice>=58
+    libavfilter>=7
+    libavformat>=58
+    libavcodec>=58
+    libswresample>=3
+    libswscale>=3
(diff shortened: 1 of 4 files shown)


third_party/CMakeLists.txt
```diff
@@ -13,6 +13,26 @@ if (BUILD_SOX)
   list(APPEND TORCHAUDIO_THIRD_PARTIES libsox)
 endif()
 
+################################################################################
+# ffmpeg
+################################################################################
+if (BUILD_FFMPEG)
+  find_package(PkgConfig REQUIRED)
+  pkg_check_modules(LIBAV REQUIRED IMPORTED_TARGET
+    # requires ffmpeg>=4.1
```
nateanl, line 22: Is the comment intended?

mthrok: Yes. ffmpeg *package* is composed of multiple libraries, each of which has diff version numbers. https://ffmpeg.org/download.html#release_4.1

Without a comment It is hard to see which version of ffmpeg package is required here.

hwangjeff (approved):

nateanl (approved):

PR #2056, opened 2021-12-02
[DO NOT MERGE]Appying ufmt (black + usort) to torchaudio

jdsgomes:
runs `ufmt format torchaudio test` and adds the corresponding configs

The plan of action is the following:
- Import and merge this PR to fbsource (as well test that the config is compatible to the internal formatting tool) 
- Apply the internal formatting tool in a separate diff (internal only)
- Import and merge the [PR](https://github.com/pytorch/audio/pull/2057) to fix the flake8 errors 

cc @hwangjeff @NicolasHug


mthrok:
Hi @jdsgomes 

Once this is ready, please make sure that the formatting is consistent with fbcode.
You can do so by `Import to fbsource` and remove this line https://fburl.com/code/wk0vdz2b.

pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[tool.usort]
+
+first_party_detection = false
+
+[tool.black]
+
+line-length = 120
+target-version = ["py36"]
+
+[tool.ufmt]
```
NicolasHug, line 10: I would personally suggest to ignore the sphinx-gallery examples and to not format them, as we do in torchvision. But this is up to audio's maintainers to decide :)

NicolasHug (commented):
Thanks @jdsgomes , this looks good!!

The CI check seems to work as expected. I would suggest to fix the flake8 issues on a later PR, but you could also do it here if you want.

I see a bunch of `E203 whitespace before ':'` errors. In torchvision, we ignore these - and a bunch of others, see our `setup.cfg`: https://github.com/pytorch/vision/blob/main/setup.cfg

You might want to do something similar!

pyproject.toml
jdsgomes: I. didn't find the `gallery` folder initially but I believe the equivalent here is  `examples/tutorials`. I decided to exclude the entire examples folder but happy to take other suggestions.

jdsgomes:
> Thanks @jdsgomes , this looks good!!
> 
> The CI check seems to work as expected. I would suggest to fix the flake8 issues on a later PR, but you could also do it here if you want.
> 
> I see a bunch of `E203 whitespace before ':'` errors. In torchvision, we ignore these - and a bunch of others, see our `setup.cfg`: https://github.com/pytorch/vision/blob/main/setup.cfg
> 
> You might want to do something similar!

That is the plan [here](https://github.com/pytorch/audio/pull/2057)

jdsgomes:
> Once this is ready, please make sure that the formatting is consistent with fbcode.

Can I import with approval here?

mthrok:
@jdsgomes 

This commit is very big, and it's hard to what configuration is changed.
Can you split it into a commit about the config change and the actual format change?
For the actual commit, it's ideal to start from fbcode and export the diff here to see that the config matches with github, then land it within fbcode.

jdsgomes:
> I might have missed it, but I don't see a CI check for the formatting? I think the `stylecheck_py3.6` job only checks for flake8 and clang-format at the moment

Well spotted. We are missing the circle CI configs will add them now to test and then create separate PRs for config and code formatting.

mthrok:
@jdsgomes I realized that I made a critical typo in my previous comment, so let me re-explain the suggested approach.

1. Create a PR that includes only the configuration changes related to OSS. Things around the CircleCI and GitHub are OSS exclusive, so we can do the review on GitHub. The important thing here is to make the format check a warning instead of error.
2. Create a diff that enable formatting in fbcode, format the audio codebase within fbcode, then export the diff on GitHub. If a format warning is still issued, consider it as configuration error and resolve it in a separate PR. Repeat this process until we see no format warning. Once there is no format warning on CI, then land the diff.
3. Once the formatting diff is landed, convert the format warning in OSS CI to error.

The key idea is that we want to have only one commit for formatting and the formatting to be exactly same as how it is in fbcode.

jdsgomes:
This is being handled in the internal side now

PR #2061, opened 2021-12-03
Centroid_wip

jacobjwebber:
It was mentioned in #1059 that it might be good to support a narrow range of frequencies to find the spectral centroid. This is an attempt at that.

Apologies if this is not correct, it is my first attempt at a PR for pytorch/audio!


torchaudio/functional/functional.py
```diff
@@ -1155,14 +1159,27 @@ def spectral_centroid(
...
+        max_freq (float, optional): Specify a maximum frequency to include in centroid calculation
 
     Returns:
         Tensor: Dimension `(..., time)`
     """
+    nyquist = sample_rate // 2
+    fft_bins = 1 + n_fft // 2
     specgram = spectrogram(waveform, pad=pad, window=window, n_fft=n_fft, hop_length=hop_length,
                            win_length=win_length, power=1., normalized=False)
     freqs = torch.linspace(0, sample_rate // 2, steps=1 + n_fft // 2,
```
mthrok, line 1172:
```suggestion
    freqs = torch.linspace(0, nyquist, steps=fft_bins,
```

torchaudio/functional/functional.py
```diff
@@ -1155,14 +1159,27 @@ def spectral_centroid(
...
                            device=specgram.device).reshape((-1, 1))
+    
+    min_freq_index = int(round((min_freq * fft_bins) / nyquist)) if min_freq is not None else 0
+    max_freq_index = int(round((max_freq * fft_bins) / nyquist)) if max_freq is not None else fft_bins
+
+    if min_freq is not None or max_freq is not None:
+        assert min_freq_index < max_freq_index
```
mthrok, line 1179: Instead of checking the value here, the check on the given arguments should happen at the beginning of the argument.

torchaudio/functional/functional.py
```diff
@@ -1155,14 +1159,27 @@ def spectral_centroid(
...
     """
+    nyquist = sample_rate // 2
+    fft_bins = 1 + n_fft // 2
     specgram = spectrogram(waveform, pad=pad, window=window, n_fft=n_fft, hop_length=hop_length,
                            win_length=win_length, power=1., normalized=False)
     freqs = torch.linspace(0, sample_rate // 2, steps=1 + n_fft // 2,
                            device=specgram.device).reshape((-1, 1))
+    
+    min_freq_index = int(round((min_freq * fft_bins) / nyquist)) if min_freq is not None else 0
+    max_freq_index = int(round((max_freq * fft_bins) / nyquist)) if max_freq is not None else fft_bins
```
mthrok, line 1176: IIRC TorchScript does not handle ternary operator. Can you run `pytest test/torchaudio_unittest/transforms/torchscript_consistency_cpu_test.py -k Centroid` and see if it works?

mthrok (commented):
Hi @jacobjwebber

Thanks for the contribution. I think I understand the gist of the new logic, but can you add a test code that explains the desired behavior?

mthrok:
CircleCI is not getting triggered. Let me close and reopen the PR to see if it helps.

torchaudio/functional/functional.py
jacobjwebber: Hi, this test seems to pass on my system. Should I still swap out the ternary?

jacobjwebber:
Hi @mthrok, sorry for taking a little while to get back to you.

I added a test in batch_consistency_test.py for the new behaviour -- I am not sure if this is the right place.

The basic idea of why you would want this is that you might want the centroid of a specific range of frequencies. I used this recently because I only wanted the centroid of voiced frequencies.

A downside of accepting this is that it adds features that aren't present in librosa. I think this transform was initially intended to clone a function from that library

torchaudio/functional/functional.py
jacobjwebber: Sorry, I don't fully understand this. Would you mind explaining?

PR #2064, opened 2021-12-06
Add HuBERT pretrain model to enable training from scratch

nateanl:
- Add three factory functions:`hubert_pretrain_base`, `hubert_pretrain_large`, and `hubert_pretrain_xlarge`, to enable the HuBERT model to train from scratch.
- Add `num_classes` argument to `hubert_pretrain_base` factory function because the base model has two iterations of training, the first iteration the `num_cluster` is 100, in the second iteration `num_cluster` is 500.
- The model takes `waveforms`, `labels`, and `lengths` as inputs
- The model generates the last layer of transformer embedding, `logit_m`, `logit_u` as the outputs.


mthrok:
I understand the necessity and the direction, but I have some concerns about how the API is extended.

The naming around wav2vec2 is somewhat tricky, because the novelty of the paper wav2vec2 is not about the model architecture but about the training method. This is the reason why the documentation of [`Wav2Vec2Model`](https://pytorch.org/audio/stable/models.html#wav2vec2model) says `Encoder model used in wav2vec 2.0`, (not *proposed*) and why I decided not to add `HubertModel` class when adding the support for HuBERT.

Now, inevitably, the HuBERT model structure is different for pre-training, in a way that does not apply to wav2vec2.0. Similarly, wav2vec2.0 model structure is different for pre-training. So I think we should extend the API in a way we can do the same for wav2vec2.0 pre-training model, even though we do not plan to add wav2vec2.0 pre-training at the moement.

So, here is my suggestion for factory functions, why don't we add a new set of factory functions, and keep the existing ones intact?

```python
hubert_pretraining(<arch params>, <config params>)
  - hubert_pretraining_base(<config params>)
  - hubert_pretraining_large(<config params>)
  - hubert_pretraining_xlarge(<config params>)
```

I will look into the component change later to see if these factory functions can reuse the model classes.

nateanl:
Thanks for looking into it. Now the HuBERT module has those extra arguments:
```
mask_generator: Module,
mask_embedding: Parameter,
label_embeddings: Parameter,
final_proj: Module,
```
where the first three modules are for pre-training only, the `final_proj` is for masked prediction of HuBERT pre-training if `aux` is None, or it's for asr prediction if `aux` is not None.

Another function I plan to add is the fine-tuning mode, if `aux` is not None, when training the model the encoder has no gradients like the [impl in fairseq](https://github.com/pytorch/fairseq/blob/main/fairseq/models/hubert/hubert_asr.py#L312-L317)
```python
      ft = self.freeze_finetune_updates <= self.num_updates

      with torch.no_grad() if not ft else contextlib.ExitStack():
          x, padding_mask = self.w2v_model.extract_features(**w2v_args)

          if tbc:
              # B x T x C -> T x B x C
              x = x.transpose(0, 1)
```

mthrok:
Could you separate the factory functions for new pre-training models from the existing one?
I do not think there is much advantage in mixing them together.
The `Union` return type is a code smell. For example, if all the parameters for pretraining is provided by user but the user forgot to provide `train=True`, what should function do?

So my suggestion is to
1. Rename `hubert_model` to `hubert_pretrain_model`.
2. Do not modify the existing `wav2vec2_XXX` functions as their responsibilities are to generate model architecture without components for pretraining module.

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,23 +116,162 @@ def forward(
...
+    """  # noqa: E501
+    def __init__(
+        self,
+        feature_extractor: Module,
+        encoder: Module,
+        mask_generator: Module,
+        mask_embedding: Parameter,
+        label_embeddings: Parameter,
+        final_proj: Module,
+        aux: Optional[Module] = None,
```
mthrok, line 153: Since this model is designed for pretraining, aux module should not be included.

torchaudio/models/wav2vec2/model.py
```diff
@@ -27,10 +27,10 @@ class Wav2Vec2Model(Module):
             Auxiliary module. If provided, the output from encoder is passed to this module.
     """  # noqa: E501
     def __init__(
-            self,
```
mthrok, line 30: Could you revert the unrelated changes?

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,23 +116,162 @@ def forward(
...
+            audio_lengths (Tensor or None, optional):
+                Indicates the valid length of each audio in the batch.
+                Shape: `(batch, )`.
+                When the ``waveforms`` contains audios with different durations,
+                by providing ``lengths`` argument, the model will compute
+                the corresponding valid output lengths and apply proper mask in
+                transformer attention layer.
+                If ``None``, it is assumed that all the audio in ``waveforms``
+                have valid length. Default: ``None``.
+            label_lengths (Tensor or None, optional):
```
mthrok, line 210: This argument seems to be missing.

torchaudio/models/wav2vec2/model.py
nateanl: Sure, will do. Sorry I didn't notice the indentations when I was reviewing the PR. Do you think we should fix them in a separate PR?

torchaudio/models/wav2vec2/model.py
nateanl: Actually this model is good for pre-training and fine-tuning. If `aux` has value, then only the `aux` layer will have gradient during the training.

torchaudio/models/wav2vec2/model.py
nateanl: Thanks. I will fix it. It seems `label_lengths` is not necessary, using `audio_lengths` helps creates the padding mask for labels.

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,6 +116,125 @@ def forward(
         return x, lengths
 
 
+class HuBERTModel(Wav2Vec2Model):
```
mthrok, line 119: Looking at how `forward` is implemented, I do not see the necessity for this model to inherit `Wav2Vec2Model`.

Also I am wondering if this model should inherit `Wav2Vec2Model` or compose it. https://en.wikipedia.org/wiki/Composition_over_inheritance

I think turning it into composition would allow clearer group of parameter file.
Say we change the model architecture so that an instance of `Wav2Vec2Model` is placed as an attribute of `HuBERTModel`. Then it is easy to extract the related parts of the state dict for fine-tuning.

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,6 +116,125 @@ def forward(
...
+        mask_generator: Module,
+        mask_embedding: Parameter,
+        label_embeddings: Parameter,
+        final_proj: Module,
+    ):
+        super().__init__(feature_extractor, encoder)
+        self.feature_extractor = feature_extractor
+        self.encoder = encoder
+        self.mask_generator = mask_generator
+        self.mask_embedding = mask_embedding
```
mthrok, line 155: I am aware that with the use of factory functions, users do not need to build `Parameter` by themselves, but it appears somewhat unconventional to have `Parameter` as a constructor arguments.

Looking at how `mask_embedding` is used, it is only passed to `mask_generator`. Can't `mask_embedding` be part of `mask_generator`?

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,6 +116,125 @@ def forward(
...
+        self.label_embeddings = label_embeddings
+        self.final_proj = final_proj
+
+    def _compute_pred(
```
mthrok, line 159: `_compute_pred` and `_get_padding_mask` do not use `self` meaning that the logic can be extracted outside of the class. Can you extract them?

torchaudio/models/wav2vec2/model.py
```diff
@@ -284,6 +403,206 @@ def wav2vec2_model(
...
+            to encoder layers. In encoder layer, two layer norms are applied after self
+            attention, before and after feed forward.
+
+            This option corresponds to ``layer_norm_first`` from ``fairseq``.
+
+        encoder_layer_drop (float):
+            Probability to drop each encoder layer during training.
+
+            This option corresponds to ``layerdrop`` from ``fairseq``.
+
```
mthrok, line 554: Looks like the docstring misses the newly introduced arguments, such as `mask_prob` and the rest.

torchaudio/models/wav2vec2/model.py
```diff
@@ -284,6 +403,206 @@ def wav2vec2_model(
...
+    Returns:
+        Wav2Vec2Model:
```
mthrok, line 556:
```suggestion
        HubertModel:
```

torchaudio/models/wav2vec2/model_utils.py
```diff
@@ -0,0 +1,131 @@
+# Originally from fairseq:
+# https://github.com/pytorch/fairseq/blob/0dfd6b624081fc4e1c72fc74ae0cd2de199c334c/fairseq/data/data_utils.py#L393-L517
+from typing import Optional, Tuple
+import numpy as np
```
mthrok, line 4: `torchaudio` does not list NumPy as dependency so this needs a workaround. The best solution is to replace the use of NumPy with equivalent PyTorch.

mthrok (commented):
I think we need to update `__init__.py` and documentation as well.

torchaudio/models/wav2vec2/model.py
nateanl: Sounds good. I feel like we should use composition for the existing ``Wav2Vec2Model`` too, by removing ``aux`` layer from it. 
In this way, we have five basic model class: 
```python
Wav2Vec2Model
HuBERTPretrainModel(wav2vec2: Wav2Vec2Model, mask_generator: Module, etc)
Wav2Vec2PretrainModel(wav2vec2: Wav2Vec2Model, extra_training_args)
HuBERTASRModel(wav2vec2: Wav2Vec2Model, aux: Module)
Wav2Vec2ASRModel(wav2vec2: Wav2Vec2Model, aux: Module)
```
The benefit is that when we finish the pre-training of wav2vec2 or HuBERT model, we can easily load the state dictionary by
```python
model = hubert_asr_base()
checkpoint = torch.load(PATH)
model.wav2vec2.load_state_dict(checkpoint["wav2vec2"])
```
I found that in Wav2Vec2Bundle, the model is generated by ``wav2vec2_model``, and ``wav2vec2_base`` or ``huber_base`` is not used in pretrained pipelines. This means we can keep the same API without BC-Breaking. If users want to initialize the model and train from scratch, then they will use ``hubert_pretrain_base`` and ``hubert_asr_base`` because they contain the basic wav2vec2model and other training parameters they need.
What do you think?

nateanl: In ``hubert_pretrain_model`` or ``hubert_asr_model``, we can initialize the model like this:
```python
def hubert_pretrain_model():
    wav2vec2 = wav2vec2_model()
    mask_generator = ...
    final_proj = ...
    return HuBERTPreTrainModel(
        wav2vec2=wav2vec2,
        mask_generator=mask_generator,
        final_proj=final_proj
    )
def hubert_asr_model():
    wav2vec2 = wav2vec2_model()
    aux = ...
    return HuBERTASRModel(
        wav2vec2=wav2vec2,
        aux=aux,
    )
```

torchaudio/models/wav2vec2/model.py
nateanl: Sure, let me add ``mask_embedding`` into ``mask_generator``, but for ``label_embeddings``, maybe I can create another module class named ``LogitsGenerator`` to include it?

torchaudio/models/wav2vec2/model.py
nateanl: Yep, I didn't fix the docstring yet. Let me do that.

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,6 +116,125 @@ def forward(
...
+
+        final_proj (torch.nn.Module):
+    """  # noqa: E501
+    def __init__(
+        self,
+        feature_extractor: Module,
+        encoder: Module,
+        mask_generator: Module,
+        mask_embedding: Parameter,
+        label_embeddings: Parameter,
```
fmassa, line 148: Instead of passing the raw parameters, why not pass a `nn.Embedding` layer instead?

torchaudio/models/wav2vec2/model.py
```diff
@@ -116,6 +116,125 @@ def forward(
...
+        proj_x = self.final_proj(x)
+        mask_m = torch.logical_and(~padding_mask, mask)
+        mask_u = torch.logical_and(~padding_mask, ~mask_m)
+        proj_x_m = proj_x[mask_m]
+        label_m = labels[mask_m]
+        logit_m = self._compute_pred(proj_x_m, label_m, self.label_embeddings)
+
+        proj_x_u = proj_x[mask_u]
+        label_u = labels[mask_u]
+        logit_u = self._compute_pred(proj_x_u, label_u, self.label_embeddings)
```
fmassa, line 234: Looking at the implementation of Wav2Vec2, it seems that it could be used without modifications (or with minimal modifications) to be able to handle HuBERT.

1 - The `_get_padding_mask`, and `mask_generator` can be moved inside `encoder`. This can be achieve either with a wrapper class around the `encoder`, or by extending the original encoder used in `Wav2Vec2`.
2 - the `final_proj` and the logic that follows (including `_compute_pred` can be contained in an new `nn.Module` which will be passed as the `aux` of the model.

Here is some example snippets that illustrates my points:

```python
# as is
class Wav2Vec2(...):
   def __init__(self, feature_extractor, encoder, aux=None):
       ...

# either we extend the original Encoder
# or we wrap it
class EncoderWithMask(...):
    def forward(self, input):
        mask = generate_mask(input)
        out = self.encoder(input, mask)
        # we return the mask, and it's up to you to
        # decide if you want to potentially support that
        # return value in the Wav2Vec2
        return out, mask
# create a new Aux class that will
# be passed to the wav2vec2 model
class MyAux(nn.Module):
    def __init__(self, ...):
        self.final_proj = nn.Linear(...)
        self.label_embedding = nn.Embedding(...)
    
    def forward(self, input):
       x, mask = input
       out = self.final_proj(x)
       ... = self.label_embedding(targets)
       ... = self_compute_pred(...)
        return ...

# instantiate the HuBERT models
def hubert_pretrain_model(...):
    return Wav2Vec2(feature_extractor, EncoderWithMask(...), aux=MyAux(...))



```

nateanl: @fmassa Thanks for the great input! The design looks reasonable, as the HuBERT model can be initialized the same way as Wav2Vec2. 
One issue is that for pre-training, the ``labels`` is necessary in the forward method, so we can't reuse the same forward method in Wav2Vec2. I was thinking about moving the label-related computation to loss function, however, there are some trainable parameters such as ``mask_embedding`` and ``label_embeddings``, if moving them to loss function, how to add them to the optimizer is tricky then.

torchaudio/models/wav2vec2/model.py
```diff
@@ -117,7 +117,258 @@ def forward(
...
+        layer_norm_first=encoder_layer_norm_first,
+        layer_drop=encoder_layer_drop,
+    )
+    aux = None
+    if aux_num_out is not None:
+        aux = torch.nn.Linear(in_features=encoder_embed_dim, out_features=aux_num_out)
+    return Wav2Vec2Model(feature_extractor, encoder, aux)
+
+
+def hubert_pretrain_model(
```
mthrok, line 371: Just to be safe, could you put all the new functions to the end of this module? `git` is not interpreting correctly.

docs/source/models.rst
```diff
@@ -98,6 +105,21 @@ hubert_xlarge
 
 .. autofunction:: hubert_xlarge
 
+hubert_pretrain_base
```
mthrok, line 108: Is `hubert_pretrain_model` missing?

mthrok (approved):

PR #2076, opened 2021-12-17
Add static build of KenLM

mthrok:
Add KenLM and its dependencies required for static build (`zlib`, `bzip2`, `lzma` and `boost-thread`).

The KenLM and its dependencies are build but since no corresponding code on torchaudio side is changed, the resulting torchaudio extension module is not changed. (therefore, as long as build process passes on CI this PR should be good to go.)


CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ endif()
 option(BUILD_SOX "Build libsox statically" ON)
 option(BUILD_KALDI "Build kaldi statically" ON)
 option(BUILD_RNNT "Enable RNN transducer" ON)
+option(BUILD_KENLM "Build KenLM statically" ON)
```
carolineechen, line 62: do we want this to be supported by default?

carolineechen (approved):

CMakeLists.txt
mthrok: Yes, since we intend to include this in the next release with prototype. In the past, when the release engineering team manually released a binary for M1 mac, they did not know about BUILD_SOX switch, and it caused additional issue.
So I think it's safer to enable by default. Plus, this way if something is wrong, we will receive issue report from users building from source.

PR #2077, opened 2021-12-17
Introduce helper function to define extension

mthrok:
Similar to #2040 this commit refactor the part of the CMakeLists.txt
which defines extension module so that second extension can be added
easily.


carolineechen (approved):
LGTM

PR #2078, opened 2021-12-17
Add FL Decoder / KenLM integration to build process

mthrok:
After all the C++ code from #2072 are added, this commit will enable decoder/KenLM integration in the build process.


torchaudio/csrc/CMakeLists.txt
```diff
@@ -126,6 +126,51 @@ define_library(
...
+  define_library(
+    libtorchaudio_decoder
+    "${LIBTORCHAUDIO_DECODER_SOURCES}"
+    "${PROJECT_SOURCE_DIR}"
+    "${LIBTORCHAUDIO_DECODER_DEPS}"
+    "${LIBTORCHAUDIO_COMPILE_DEFINITIONS};${LIBTORCHAUDIO_DECODER_DEFINITIONS}"
+    )
+endif()
+
+# TODO: Add libtorchaudio_decoder
```
carolineechen, line 173: what is necessary for adding libtorchaudio_decoder?

torchaudio/csrc/CMakeLists.txt
```diff
@@ -126,6 +126,51 @@ define_library(
...
   )
 
+
+################################################################################
+# libtorchaudio_decoder.so
+################################################################################
+if (BUILD_CTC_DECODER)
+  set(
+    LIBTORCHAUDIO_DECODER_SOURCES
+    decoder/src/decoder/Decoder.h
```
carolineechen, line 136: we should be able to remove the `.h` header files from this list

carolineechen (approved):

torchaudio/csrc/CMakeLists.txt
mthrok: Thanks for pointing. Let me follow-up on this in the subsequent PR so that I can land this without alerting the fbcode's change-after-approve system during the weekend.

torchaudio/csrc/CMakeLists.txt
mthrok: The definition of `TORCHAUDIO_LIBRARY` that is expected to be used from C++ integration via CMake. To make the decoder available in C++, we need to add `libtorchaudio_decoder` to it. However, the current way is not entirely correct, so we need to revise it.

PR #2079, opened 2021-12-17
Add C++ files for CTC decoder bindings

carolineechen:
part of #2072 -- splitting up the PR for easier review

Add C++ files for binding CTC decoder functionality for Python

Note: the code here will not be compiled until the build process is changed

### DIFF
--- a/torchaudio/csrc/decoder/bindings/_decoder.cpp
+++ b/torchaudio/csrc/decoder/bindings/_decoder.cpp
@@ -1,0 +1,111 @@
+/*
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT-style license found in
+ * https://github.com/flashlight/flashlight/blob/d385b2150872fd7bf106601475d8719a703fe9ee/LICENSE
+ */
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "torchaudio/csrc/decoder/src/decoder/LexiconDecoder.h"
+#include "torchaudio/csrc/decoder/src/decoder/lm/KenLM.h"
+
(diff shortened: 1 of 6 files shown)


torchaudio/csrc/decoder/bindings/pybind.cpp
```diff
@@ -0,0 +1,140 @@
+#include <torch/extension.h>
+
+#include <torchaudio/csrc/decoder/bindings/_decoder.cpp>
+#include <torchaudio/csrc/decoder/bindings/_dictionary.cpp>
+
+PYBIND11_MODULE(_torchaudio_decoder, m) {
+#ifdef BUILD_CTC_DECODER
```
mthrok, line 7: I think we can remove this directive since we are creating a library file separate from `libtorchaudio`.

torchaudio/csrc/decoder/bindings/cmake/Buildpybind11.cmake
```diff
@@ -0,0 +1,19 @@
+include(ExternalProject)
```
mthrok, line 1: We no longer need any of the `*.cmake` files.

torchaudio/csrc/decoder/bindings/_decoder.cpp
```diff
@@ -0,0 +1,111 @@
+/*
```
mthrok, line 1: I think `_decoder.cpp`, `_dictionary.cpp` and `pybind.cpp` can be combined and placed directly at `torchaudio/csrc/decoder/*`. it looks strange that `pybind.cpp` incldues two other cpp files.

mthrok (approved):

mthrok:
note: style check is failing, so let's follow up on that one as well.

PR #2082, opened 2021-12-18
Update audio augmentation tutorial

mthrok:
1. Reorder Audio display so that audios are playable from browser in doc
2. Add link to function documentations

https://470342-90321822-gh.circle-artifacts.com/0/docs/tutorials/audio_data_augmentation_tutorial.html

### DIFF
--- a/examples/tutorials/audio_data_augmentation_tutorial.py
+++ b/examples/tutorials/audio_data_augmentation_tutorial.py
@@ -6,10 +6,6 @@ Audio Data Augmentation
 ``torchaudio`` provides a variety of ways to augment audio data.
 """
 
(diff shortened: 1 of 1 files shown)


examples/tutorials/audio_data_augmentation_tutorial.py
```diff
@@ -243,12 +238,24 @@ def get_noise_sample(*, resample=None):
 # audio. Doesn’t it sound more dramatic?
```
carolineechen, line 238: (minor) imo this note would make a bit more sense after the displayed plots and audio output

carolineechen (approved):

PR #2087, opened 2021-12-20
Revert linting exemptions introduced in #2071

jdsgomes:
After discussing with Moto Hira, we decided to revert linting exemptions
introduced previously in order to keep the entire audio project as formatted
as possible, to reduce the time we spend on formatting discussion.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,7 +128,7 @@ jobs:
           command: pre-commit run --all-files || true
       - run:
           name: Required lint modifications
-          when: on_fail
+          when: always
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -128,7 +128,7 @@ jobs:
           command: pre-commit run --all-files || true
       - run:
           name: Required lint modifications
-          when: on_fail
```
mthrok, line 131: This change needs to happen from `config.yml.in`.

mthrok (approved):

PR #2089, opened 2021-12-20
Add Python CTC decoder API

carolineechen:
Part of #2072 -- splitting up PR for easier review

This PR adds Python decoder API and basic README


torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,224 @@
+import torch
+import itertools as it
+from typing import List, Optional, Dict
+from collections import namedtuple
+
+import torchaudio
+
+try:
+    torchaudio._extension._load_lib('libtorchaudio_decoder')
```
mthrok, line 9: Can you move the initialization of "libtorchaudio_decoder" to "__init.py__"?
Then wrap it with try except claus. After that in ctc_decoder.py, no need to try except the import statement of _torchaudio_decodrt

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,224 @@
...
+        Returns:
+            List: tokens corresponding to the input IDs
+        """
+        return [self.tokens_dict.get_entry(idx.item()) for idx in idxs]
+
+
+def kenlm_lexicon_decoder(
+    lexicon_file: str,
+    tokens_file: str,
+    kenlm_file: str,
```
mthrok, line 162: I think tokens can be more flexible than requiring a file. Tokens should be able to be directly fed as list of str, so that it can accept the value from `w2v2bundle.get_labels()`.

Similarly I feel that lexicon argument can be a regular dict as well. Looking at the implementation of `_Dictionary`, it needs reverse look up which can be constructed from regular dict if I'm not mistaken. Not requiring a file gives better flexibility at testing but I'm not entirely sure if that's best for UX. Let's follow-up on this one as we add pretrained model support and write tutorials.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,224 @@
...
+        self.nbest = nbest
+        self.decoder_options = decoder_options
+        self.lexicon = lexicon
+        self.word_dict = word_dict
+        self.lm = kenlm
+        self.tokens_dict = tokens_dict
+
+        self.vocab_size = tokens_dict.index_size()
+
+        self.unk_word = word_dict.get_index("<unk>")
```
mthrok, line 70: I think this unk token should be parameterized as well. `UNK` is another popular one.

mthrok (approved):

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: I will follow up on both comments in a future PR, since the way this is currently being done would require C++ changes to the `_Dictionary` binding or perhaps transitioning to use a different Python/PyTorch based Dictionary

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+    def _get_tokens(self, idxs: torch.IntTensor) -> torch.LongTensor:
+        idxs = (g[0] for g in it.groupby(idxs))
+        idxs = filter(lambda x: x != self.blank, idxs)
+        return torch.LongTensor(list(idxs))
+
+    def __call__(
+        self,
+        emissions: torch.FloatTensor,
+        lengths: Optional[torch.Tensor] = None
+    ) -> List[List[Dict]]:
```
hwangjeff, line 100:
```suggestion
    ) -> List[List[Hypothesis]]:
```

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+            lexicon (dict): lexicon mapping of words to spellings
+            word_dict (_Dictionary): dictionary of words
+            tokens_dict (_Dictionary): dictionary of tokens
+            kenlm (_KenLM): n-gram KenLM language model
+            decoder_options (_LexiconDecoderOptions): parameters used for beam search decoding
+            blank_token (str): token corresopnding to blank
+            sil_token (str): token corresponding to silence
+            unk_word (str): word corresponding to unknown
+        """
+
```
hwangjeff, line 55: a few of these instance variables aren't referenced outside of the initializer, e.g. `self.decoder_options`, `self.lexicon`. can they be removed?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+                    score: hypothesis score
+                    words: list of decoded words
+        """
+        B, T, N = emissions.size()
+        if lengths is None:
+            lengths = torch.full((B, ), T)
+
+        hypos = []
+        for b in range(B):
+            emissions_ptr = emissions.data_ptr() + 4 * b * emissions.stride(0)
```
hwangjeff, line 123: what does the 4 correspond to?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+    ) -> List[List[Dict]]:
+        """
+        Args:
+            emissions (FloatTensor): tensor of shape `(batch, frame, num_tokens)` storing sequences of
+                probability distribution over labels; output of acoustic model
+            lengths (Tensor or None, optional): tensor of shape `(batch, )` storing the valid length of
+                in time axis of the output Tensor in each batch
+
+        Returns:
+            List[Hypothesis]:
```
hwangjeff, line 109: and here

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+                    )
+                    for result in nbest_results
+                ]
+            )
+
+        return hypos
+
+    def idxs_to_tokens(self, idxs: torch.LongTensor) -> List:
+        """
+        Map raw token IDs into correponding tokens
```
hwangjeff, line 143:
```suggestion
        Map raw token IDs into corresponding tokens
```

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+            If None, it is set to the total number of tokens (Default: None)
+        beam_threshold (float): threshold for pruning hypothesis (Default: 50)
+        lm_weight (float): weight of lm (Default: 2)
+        word_score (float): word insertion score (Default: 0)
+        unk_score (float): unknown word insertion score (Default: -inf)
+        sil_score (float): silence insertion score (Default: 0)
+        log_add (bool): whether or not to use logadd when merging hypotheses (Default: False)
+        blank_token (str): token corresponding to blank
+        sil_token (str): token corresponding to silence
+
```
hwangjeff, line 190: can add defaults and an entry for arg `unk_word`

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -0,0 +1,221 @@
...
+    unk_word: str = "<unk>",
+) -> KenLMLexiconDecoder:
+    """
+    Builds Ken LM CTC Lexicon Decoder with given parameters
+
+    Args:
+        lexicon (str): lexicon file containing the possible words
+        tokens (str): file containing valid tokens
+        kenlm (str): file containing languge model
+        nbest (int): number of best decodings to return (Default: 1)
```
hwangjeff, line 178: lines starting here should have "optional"

torchaudio/prototype/ctc_decoder/ctc_decoder.py
mthrok: 4 bytes for float, I assume.

mthrok:
Tests for Windows are failing. We need to decide on the Windows protocol.
1. Simply remove the `torchaudio.prototype.ctc_decoder` from Windows package
2. Raise an error on early on stating that ctc_decoder is not supported on Windows.
3. Make it query-able to see if the installed package has ctc_decoder

torchaudio/prototype/ctc_decoder/__init__.py
```diff
@@ -0,0 +1,15 @@
+import torchaudio
+
+try:
+    torchaudio._extension._load_lib('libtorchaudio_decoder')
+    from .ctc_decoder import KenLMLexiconDecoder, kenlm_lexicon_decoder
+except ImportError:
+    raise ImportError(
+        "flashlight decoder bindings are required to use this functionality. "
+        "Please set BUILD_CTC_DECODER=1 when building from source."
+    )
```
mthrok, line 10: Since when `libtorchaudio_decoder` is available, `_torchaudio_decoder` is also available, it is sufficient to wrap `torchaudio._extension._load_lib('libtorchaudio_decoder')` with try except. `from .ctc_decoder import KenLMLexiconDecoder, kenlm_lexicon_decoder` can stay outside of the `try ~ except ~`. (In general, the smaller the statements wrapped with `try ~ except`, the better)

torchaudio/prototype/ctc_decoder/__init__.py
```diff
@@ -0,0 +1,15 @@
+import torchaudio
+
+try:
+    torchaudio._extension._load_lib('libtorchaudio_decoder')
+    from .ctc_decoder import KenLMLexiconDecoder, kenlm_lexicon_decoder
+except ImportError:
```
mthrok, line 6: Please reuse the error context.

```python
except ImportError as err:
    raise ImportError("...") from err
```

torchaudio/_internal/module_utils.py
```diff
@@ -123,3 +123,11 @@ def wrapped(*args, **kwargs):
                 raise RuntimeError(f'{func.__module__}.{func.__name__} requires sox')
             return wrapped
     return decorator
+
+
+def is_ctc_decoder_available():
```
mthrok, line 128: I have a mixed feeling about having this function inside of `torchaudio`.

- Importing `torchaudio.prototype.ctc_decoder` might cause an unrecoverable effect like segfault
- Having this function here might encourage the use in library or other place, which goes against the purpose of logically separating the initialization of `torchaudio.prototype.ctc_decoder`, giving users control when that module should be initialized, for the fear of unrecoverable error.
- Yet, I am somewhat confident that's such error would not happen.

It's probably safer to move this inside of test module. What do you think?

test/torchaudio_unittest/assets/decoder/lexicon.txt
```diff
@@ -0,0 +1,2 @@
+a a |
```
mthrok, line 1: Could you add 
1. words that are composed of multiple of tokens?
2. words that share common ancestor tokens?

I think it's a bit easier to read if words are different from tokens.
 
```
foo f o o |
bar b a r |
foobar f o o b a r |
```

torchaudio/prototype/ctc_decoder/__init__.py
carolineechen: [`_load_lib`](https://github.com/pytorch/audio/blob/main/torchaudio/_extension.py#L23) does not actually throw an error if the library does not exist (it will just load nothing) -- should we change this behavior to throw an error or warning if called with a nonexistent library?

mthrok: > [`_load_lib`](https://github.com/pytorch/audio/blob/main/torchaudio/_extension.py#L23) does not actually throw an error if the library does not exist (it will just load nothing) -- should we change this behavior to throw an error or warning if called with a nonexistent library?

Oops. You are right. This one is tricky, we want to allow the skipping behavior as well for the sake of PEX-format compatibility, which deploys `libtorchaudio*.so` separately. We want to have the same mechanism for prototype CTC decoder.

Okay, so it's more appropriate to wrap the both imports. Thanks for the clarification.

torchaudio/csrc/decoder/README.md
```diff
@@ -0,0 +1,36 @@
+# Flashlight Decoder Binding
+CTC Decoder with KenLM and lexicon support based on [flashlight](https://github.com/flashlight/flashlight) decoder implementation
+and fairseq [KenLMDecoder](https://github.com/pytorch/fairseq/blob/fcca32258c8e8bcc9f9890bf4714fa2f96b6b3e1/examples/speech_recognition/new/decoders/flashlight_decoder.py#L53)
+Python wrapper
+
+## Setup
+### Build torchaudio with decoder support
+```
+BUILD_CTC_DECODER=1 python setup.py develop
```
mthrok, line 9: nit: `BUILD_CTC_DECODER=1` by default now.

PR #2090, opened 2021-12-20
Add parameter p to TimeMasking

hwangjeff:
Adds parameter `p` to `TimeMasking` to allow for enforcing an upper bound on the proportion of time steps that it can mask. This behavior is consistent with the specifications provided in the SpecAugment paper (https://arxiv.org/abs/1904.08779).

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -712,11 +712,11 @@ def _get_mask_param(mask_param: int, p: float, axis_length: int) -> int:
 
 
 def mask_along_axis_iid(
-        specgrams: Tensor,
-        mask_param: int,
-        mask_value: float,
-        axis: int,
-        p: float = 1.0,
+    specgrams: Tensor,
+    mask_param: int,
+    mask_value: float,
+    axis: int,
+    p: float = 1.0,
 ) -> Tensor:
     r"""
     Apply a mask along ``axis``. Mask will be applied from indices ``[v_0, v_0 + v)``, where
@@ -765,11 +765,11 @@ def mask_along_axis_iid(
 
 
 def mask_along_axis(
-        specgram: Tensor,
-        mask_param: int,
-        mask_value: float,
-        axis: int,
-        p: float = 1.0,
+    specgram: Tensor,
+    mask_param: int,
+    mask_value: float,
+    axis: int,
+    p: float = 1.0,
 ) -> Tensor:
     r"""
     Apply a mask along ``axis``. Mask will be applied from indices ``[v_0, v_0 + v)``, where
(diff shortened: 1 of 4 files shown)


carolineechen (commented):
Could you also add a unit test similar to [this](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/functional/functional_impl.py#L338) checking that no more than `max_mask_p` percent is masked?

torchaudio/transforms.py
```diff
@@ -1083,17 +1083,25 @@ class _AxisMasking(torch.nn.Module):
...
+    def __init__(self, mask_param: int, axis: int, iid_masks: bool, max_mask_p: float) -> None:
 
         super(_AxisMasking, self).__init__()
         self.mask_param = mask_param
         self.axis = axis
         self.iid_masks = iid_masks
+        self.max_mask_p = max_mask_p
 
-    def forward(self, specgram: Tensor, mask_value: float = 0.) -> Tensor:
+    def _get_mask_param(self, axis_length: int) -> int:
```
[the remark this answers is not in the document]
hwangjeff: yeah for sure — will change

torchaudio/functional/functional.py
```diff
@@ -762,15 +771,22 @@ def mask_along_axis_iid(
...
     dtype = specgrams.dtype
 
     value = torch.rand(specgrams.shape[:2], device=device, dtype=dtype) * mask_param
     min_value = torch.rand(specgrams.shape[:2], device=device, dtype=dtype) * (specgrams.size(axis) - value)
 
     # Create broadcastable mask
-    mask_start = min_value[..., None, None]
-    mask_end = (min_value + value)[..., None, None]
+    mask_start = min_value.long()[..., None, None]
+    mask_end = (min_value.long() + value.long())[..., None, None]
```
hwangjeff, line 789: made this fix to ensure consistency with `mask_along_axis`; without it, the resulting number of masked columns is not guaranteed to be strictly less than `mask_param`, and `test_mask_along_axis_iid` fails for certain random seeds

torchaudio/functional/functional.py
```diff
@@ -754,6 +762,7 @@ def mask_along_axis_iid(
         mask_param (int): Number of columns to be masked will be uniformly sampled from [0, mask_param]
         mask_value (float): Value to assign to the masked columns
         axis (int): Axis to apply masking on (2 -> frequency, 3 -> time)
+        p (float, optional): maximum proportion of columns that can be masked. (Default: 1.0)
```
mthrok, line 765: Should update the description to include `p`?

```
Apply a mask along ``axis``. Mask will be applied from indices ``[v_0, v_0 + p * v)``, where
    ``v`` is sampled from ``uniform(0, mask_param)``, and ``v_0`` from ``uniform(0, max_v - v)``.
```

carolineechen (approved):

PR #2093, opened 2021-12-22
Add pretrained Emformer RNN-T streaming ASR inference pipeline

hwangjeff:
Adds pretrained Emformer RNN-T inference pipeline that's capable of performing streaming and non-streaming ASR.

Includes demo script that uses pipeline to alternately perform streaming and non-streaming ASR on LibriSpeech test samples (video below).

https://user-images.githubusercontent.com/8345689/147590753-d5126557-d575-4551-8dfe-5977276cb4ad.mov

### DIFF
--- a/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
+++ b/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
@@ -26,28 +26,28 @@ def cli_main():
 
     for idx in range(10):
         sample = dataset[idx]
+        waveform = sample[0].squeeze()
 
-        with torch.no_grad():
-            waveform = sample[0].squeeze()
-
-            # Streaming decode.
-            state, hypothesis = None, None
-            for idx in range(0, len(waveform), num_samples_segment):
-                segment = waveform[idx: idx + num_samples_segment_right_context]
-                segment = torch.nn.functional.pad(segment, (0, num_samples_segment_right_context - len(segment)))
+        # Streaming decode.
+        state, hypothesis = None, None
+        for idx in range(0, len(waveform), num_samples_segment):
+            segment = waveform[idx: idx + num_samples_segment_right_context]
+            segment = torch.nn.functional.pad(segment, (0, num_samples_segment_right_context - len(segment)))
+            with torch.no_grad():
                 features, length = streaming_feature_extractor(segment)
                 hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
-                hypothesis = hypos[0]
-                transcript = token_processor(hypothesis.tokens)
-                if transcript:
-                    print(transcript, end=" ", flush=True)
-            print()
+            hypothesis = hypos[0]
+            transcript = token_processor(hypothesis.tokens)
+            if transcript:
+                print(transcript, end=" ", flush=True)
+        print()
 
-            # Non-streaming decode.
+        # Non-streaming decode.
+        with torch.no_grad():
             features, length = feature_extractor(waveform)
             hypos = decoder(features, length, 10)
-            print(token_processor(hypos[0].tokens))
-            print()
+        print(token_processor(hypos[0].tokens))
+        print()
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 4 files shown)


examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,50 @@
...
+from torchaudio.prototype.rnnt_pipeline import EMFORMER_RNNT_BASE_LIBRISPEECH
+
+
+SAMPLES_PER_CHUNK = 640
+
+
+def cli_main():
+    parser = ArgumentParser()
+    parser.add_argument(
+        "--librispeech_path", type=pathlib.Path, help="Path to LibriSpeech datasets.",
```
mthrok, line 14: I think this one needs `required=True`.

examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,50 @@
+from argparse import ArgumentParser
+import pathlib
+import torch
+import torchaudio
+from torchaudio.prototype.rnnt_pipeline import EMFORMER_RNNT_BASE_LIBRISPEECH
+
+
+SAMPLES_PER_CHUNK = 640
```
mthrok, line 8: Is this configureable at inference time? If it's tied to the way model is trained, I think it should be retrievable from `RNNTBundle`.

torchaudio/prototype/rnnt_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+_BASE_PIPELINES_URL = "https://download.pytorch.org/torchaudio/pipeline-assets"
+
+
+def _download_asset(asset_path: str):
+    dst_path = f"{os.getcwd()}/_assets/{asset_path}"
+    if not os.path.exists(dst_path):
+        os.makedirs(f"{os.getcwd()}/_assets", exist_ok=True)
+        download_url_to_file(f"{_BASE_PIPELINES_URL}/{asset_path}", dst_path)
+    else:
+        sys.stderr.write(f"{asset_path} found at {dst_path}; skipping download.\n")
```
mthrok, line 28: The other pipelines do not print anything when the file is already present.
They print message only when download is necessary, so I think we do not need to print that.

torchaudio/prototype/rnnt_pipeline.py
```diff
@@ -0,0 +1,149 @@
...
+
+__all__ = []
+
+
+_BASE_MODELS_URL = "https://download.pytorch.org/torchaudio/models"
+_BASE_PIPELINES_URL = "https://download.pytorch.org/torchaudio/pipeline-assets"
+
+
+def _download_asset(asset_path: str):
+    dst_path = f"{os.getcwd()}/_assets/{asset_path}"
```
mthrok, line 23: Can you use [`torch.hub.get_dir()`](https://github.com/pytorch/pytorch/blob/12afe2bb84f56780c299a46fbde34639720ba99c/torch/hub.py#L245-L261) as the base directory path, so that it will be configurable via environment variable. 

See also

https://github.com/pytorch/audio/blob/1b17b0116548f60ccb55ef97d09afa02bca7aa26/torchaudio/pipelines/_tts/utils.py#L174-L180

mthrok, line 23: The use of `/` is not portable to Windows. Can you use `os.path` or `pathlib`?

torchaudio/prototype/rnnt_pipeline.py
```diff
@@ -0,0 +1,384 @@
...
+                Decoded text sequence.
+        """
+        filtered_hypo_tokens = [
+            token_index for token_index in tokens[1:] if token_index not in self.post_process_remove_list
+        ]
+        return self.sp_model.decode(filtered_hypo_tokens)
+
+
+@dataclass
+class RNNTBundle:
```
mthrok, line 159: Can you hide the constructor signature by overwriting it with empty signature?
If the first line of the docstring looks like module signature, then Sphinx will use it.

https://github.com/pytorch/audio/blob/7bf04d1e9e366c90918a00003c7efa64ca8b70ae/torchaudio/pipelines/_wav2vec2/impl.py#L14-L17

becomes 

https://pytorch.org/audio/stable/pipelines.html#torchaudio.pipelines.Wav2Vec2Bundle

mthrok (approved):
LGTM! This is exciting addition.

PR #2105, opened 2021-12-26
ffmpeg-prototype-v16

mthrok:
Summary: Add streaming API using ffmpeg

Differential Revision: D33324389


torchaudio/csrc/ffmpeg/buffer.cpp
```diff
@@ -5,8 +5,27 @@
 namespace torchaudio {
 namespace ffmpeg {
 
-Buffer::Buffer(AVMediaType type) : media_type(type) {}
+Buffer::Buffer(int frames_par_chunk, int num_chunks)
```
carolineechen, line 8: should be `per`, not `par`

```suggestion
Buffer::Buffer(int frames_per_chunk, int num_chunks)
```

PR #2106, opened 2021-12-28
Add ASR CTC inference tutorial

carolineechen:
demonstrate usage of the CTC beam search decoder w/ lexicon constraint and KenLM support, on a LibriSpeech sample and using a pretrained wav2vec2 model

rendered: https://485200-90321822-gh.circle-artifacts.com/0/docs/tutorials/asr_inference_with_ctc_decoder_tutorial.html

follow-ups: 
- incorporate `nbest`
- demonstrate customizability of different beam search parameters

### DIFF
--- a/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
+++ b/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
@@ -45,17 +45,6 @@ import IPython
 
 
 ######################################################################
-# Helper Functions
-# ~~~~~~~~~~~~~~~~
-# 
-
-def download_file(url, file):
-  if not os.path.exists(file):
-    with open(file, "wb") as f:
-      f.write(requests.get(url).content)
-
-
-######################################################################
 # Acoustic Model and Data
 # ~~~~~~~~~~~~~~~~~~~~~~~
 # 
@@ -75,11 +64,12 @@ acoustic_model = bundle.get_model()
 # We will load a sample from the LibriSpeech test-other dataset.
 # 
 
-speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.flac"
-speech_file = "_assets/speech.flac"
+hub_dir = torch.hub.get_dir()
 
-os.makedirs("_assets", exist_ok=True)
-download_file(speech_url, speech_file)
+speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.wav"
+speech_file = f"{hub_dir}/speech.wav"
+
+torch.hub.download_url_to_file(speech_url, speech_file)
 
 IPython.display.Audio(speech_file)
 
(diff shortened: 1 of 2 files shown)


examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -0,0 +1,247 @@
...
+import requests
+import IPython
+
+
+######################################################################
+# Helper Functions
+# ~~~~~~~~~~~~~~~~
+# 
+
+def download_file(url, file):
```
mthrok, line 52: This might be coming from the other tutorials but I think we can simply use `torch.hub.download_url_to_file`.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -0,0 +1,247 @@
...
+# We will load a sample from the LibriSpeech test-other dataset.
+# 
+
+speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.flac"
+speech_file = "_assets/speech.flac"
+
+os.makedirs("_assets", exist_ok=True)
+download_file(speech_url, speech_file)
+
+IPython.display.Audio(speech_file)
```
mthrok, line 84: Today's browsers normally won't play flac file. Can you replace this with WAV?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -0,0 +1,247 @@
...
+# 
+
+
+######################################################################
+# Tokens
+# ^^^^^^
+# 
+# The tokens are the possible symbols that the acoustic model can predict,
+# including the blank and silent symbols.
+#
```
mthrok, line 116: Similar to lexicon bellow, can you show a part of the contents of token file?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -0,0 +1,247 @@
...
+# Now that we have the data, acoustic model, and decoder, we can perform
+# inference. Recall the transcript corresponding to the waveform is
+# ``"when it was the seven hundred and eighteenth night"``
+# 
+
+emission, _ = acoustic_model(waveform)
+
+beam_search_result = beam_search_decoder(emission)
+beam_search_transcript = " ".join(beam_search_result[0][0].words).lower().strip()
+beam_search_transcript
```
mthrok, line 236:
```suggestion
print(beam_search_transcript)
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -0,0 +1,247 @@
...
+
+greedy_result = greedy_decoder(emission[0])
+greedy_transcript = greedy_result.replace('|', ' ').lower().strip()
+greedy_transcript
```
mthrok, line 240:
```suggestion
print(greedy_transcript)
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -75,11 +64,10 @@ def download_file(url, file):
 # We will load a sample from the LibriSpeech test-other dataset.
 # 
 
-speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.flac"
-speech_file = "_assets/speech.flac"
+speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.wav"
+speech_file = "/tmp/speech.wav"
```
mthrok, line 68: Can you use [`torch.hub.get_dir`](https://pytorch.org/docs/stable/hub.html#torch.hub.get_dir)? So that configurable default temporary directory will be selected.

Hardcoding `/tmp` would leave artifacts when running locally.

mthrok (approved):
Looks good. As a follow-up idea, we can add more detail on how the decoding result change based on decoding parameters.

Also we should show n-best result, which is something Greedy Decoder cannot provide.

PR #2108, opened 2021-12-28
Update prototype documentations

mthrok:
### Change list

* Split the documentation of prototypes
* Add a new API reference section dedicated for prototypes.
* Hide the signature of KenLMLexiconDecoder constructor. (cc @carolineechen )
  * https://489516-90321822-gh.circle-artifacts.com/0/docs/prototype.ctc_decoder.html#torchaudio.prototype.ctc_decoder.KenLMLexiconDecoder
* Hide the signature of RNNT constructor. (cc @hwangjeff )
  * https://489516-90321822-gh.circle-artifacts.com/0/docs/prototype.rnnt.html#torchaudio.prototype.RNNT
* Tweak CTC tutorial
  * Replace hyperlinks to API reference with backlinks
  * Add `progress=False` to download

### Follow-up

RNNT decoder and CTC decode returns their own `Hypothesis` classes. When I tried to add Hypothesis of CTC decode to the documentation, the build process complains that it's ambiguous.
I think the Hypothesis classes can be put inside of each decoder. (if TorchScript supports it) or make the name different, but in that case the interface of each Hypothesis has to be generic enough.

### Before

https://pytorch.org/audio/main/prototype.html

Screen Shot 2021-12-28 at 1 05 53 PM

### After

https://489516-90321822-gh.circle-artifacts.com/0/docs/prototype.html

Screen Shot 2021-12-28 at 8 37 35 PM

https://489516-90321822-gh.circle-artifacts.com/0/docs/prototype.rnnt.html

Screen Shot 2021-12-28 at 8 38 27 PM

https://489516-90321822-gh.circle-artifacts.com/0/docs/prototype.ctc_decoder.html

Screen Shot 2021-12-28 at 8 39 04 PM


docs/source/prototype.rst
```diff
@@ -1,99 +1,17 @@
...
-  .. automethod:: idxs_to_tokens
-
-
-kenlm_lexicon_decoder
-~~~~~~~~~~~~~~~~~~~~~
-
-.. currentmodule:: torchaudio.prototype.ctc_decoder
-
-.. autoclass:: kenlm_lexicon_decoder
+they are are at an early stage for feedback and testing.
```
carolineechen, line 5:
```suggestion
they are at an early stage for feedback and testing.
```

docs/source/index.rst
```diff
@@ -48,7 +48,17 @@ API References
    compliance.kaldi
    kaldi_io
    utils
+
+Prototype API References
```
carolineechen, line 52: curious if there's some reason we're making prototypes it's own larger section as opposed to just regrouping within the existing `torchaudio.prototypes` section similar to how https://pytorch.org/audio/main/backend.html is structured?

carolineechen (commented):
just another note regarding the followup -- currently the `Hypothesis` return values in the CTC decoder redirect to the RNNT Hypothesis

carolineechen (commented):
another note regarding the followup -- currently the `Hypothesis` in ctc_decoder return types redirects to the RNNT Hypothesis as well

docs/source/index.rst
mthrok: It's somewhat technical and somewhat subjective.

Sphinx-gallery is not flexible enough to nicely render structure with more than two depth. I tried using `.. include:` directive to put all the contents in one page, then the `Reference` did not work, so I ended up using `.. toctree:` directive which created the front page for `torchaudio.prototype`, instead of embedding the contents in one single page.

Then, I stopped and gave some thoughts. 

The biggest difference between modules in `prototype` and others is that each module in `prototype` can contain APIs on different level of abstraction by grouped by topics, rather than functionalities. For example, `rnnt` now has APIs for models and pipelines. (`torchaudio.backend` module is somewhat elaborated, but after all, all the components listed there compose the I/O functionality.) So if we put everything in one page under `torchaudio.prototypes` then that page has categories for topics and features beneath them, which does not have a good, universal way to organize. So when I looked at the new index for the prototype features, I thought that would make sense as well.

Now regarding the new `Prototype API References` index on the left bar, I was originally listing them after the library modules, but having multiple of `torchaudio.prototype.XXX` is cluttering the whole index.

mthrok:
> another note regarding the followup -- currently the `Hypothesis` in ctc_decoder return types redirects to the RNNT Hypothesis as well

We should followup on this once we empty the `torchaudio.prototype` namespace.

carolineechen (approved):

nateanl (approved):

docs/source/prototype.rnnt.rst
```diff
@@ -0,0 +1,110 @@
...
+
+emformer_rnnt_model
+~~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: emformer_rnnt_model
+
+Decoder Classes
+---------------
+
+RNNTBeamSear
```
hwangjeff, line 60:
```suggestion
RNNTBeamSearch
```

docs/source/prototype.rnnt.rst
```diff
@@ -0,0 +1,110 @@
...
+
+
+Emformer
+~~~~~~~~
+
+.. autoclass:: Emformer
+
+  .. automethod:: forward
+
+  .. automethod:: infer
```
hwangjeff, line 26: i would like to later split these out into a separate module (tentatively `torchaudio.prototype.nn`) to indicate that they are generic modules rather than models or rnnt-specific components

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -68,7 +68,7 @@
 speech_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/8461-258277-0000.wav"
 speech_file = f"{hub_dir}/speech.wav"
 
-torch.hub.download_url_to_file(speech_url, speech_file)
+torch.hub.download_url_to_file(speech_url, speech_file, progress=False)
```
hwangjeff, line 71: why set `progress=False`? seems beneficial for the user to get visual feedback that their code execution is making progress rather than hanging silently

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -24,6 +24,14 @@
 
 
 class KenLMLexiconDecoder:
+    """torchaudio.prototype.ctc_decoder.KenLMLexiconDecoder()
+
+    Note:
+        To build the decoder, please use the factory function.
```
hwangjeff, line 30:
```suggestion
        To build the decoder, please use the factory function
```

hwangjeff (approved):

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
mthrok: Hmm... when running, it is indeed beneficial. Meanwhile, while browsing the tutorials on browser, I found them rather distracting or unnecessarily. KenLM part adds 300 lines (though only dozens of lines are visible at a time)

https://pytorch.org/audio/main/tutorials/asr_inference_with_ctc_decoder_tutorial.html#kenlm

docs/source/prototype.rnnt.rst
mthrok: Sure. I am open to re-organizing them.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -115,7 +115,7 @@
 
 token_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/tokens-w2v2.txt"
 token_file = f"{hub_dir}/token.txt"
-torch.hub.download_url_to_file(token_url, token_file)
+torch.hub.download_url_to_file(token_url, token_file, progress=False)
```
mthrok, line 118:
```suggestion
torch.hub.download_url_to_file(token_url, token_file)
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -139,7 +139,7 @@
 
 lexicon_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/lexicon-librispeech.txt"
 lexicon_file = f"{hub_dir}/lexicon.txt"
-torch.hub.download_url_to_file(lexicon_url, lexicon_file)
+torch.hub.download_url_to_file(lexicon_url, lexicon_file, progress=False)
```
mthrok, line 142:
```suggestion
torch.hub.download_url_to_file(lexicon_url, lexicon_file)
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -154,19 +154,18 @@
 
 kenlm_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/4-gram-librispeech.bin"
 kenlm_file = f"{hub_dir}/kenlm.bin"
-torch.hub.download_url_to_file(kenlm_url, kenlm_file)
+torch.hub.download_url_to_file(kenlm_url, kenlm_file, progress=False)
```
mthrok, line 157:
```suggestion
torch.hub.download_url_to_file(kenlm_url, kenlm_file)
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
mthrok, line 71:
```suggestion
torch.hub.download_url_to_file(speech_url, speech_file)
```

mthrok (commented):
reverting `progress=False`.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
hwangjeff: ah i see didn't realize it rendered like that on the tutorial page. that being said, the removal still looks good to me

PR #2110, opened 2021-12-29
Reorganize RNN-T components in prototype module

hwangjeff:
Regroup RNN-T components under `torchaudio.prototype.models` and `torchaudio.prototype.pipelines`.

Updated docs: https://492831-90321822-gh.circle-artifacts.com/0/docs/index.html


mthrok:
When graduating these modules out of prototype, we need to be careful to where we want to land it.
The introduction of the new `nn` requires us to think what are the definitions/standards/differences among `functional`, `transforms`, `models` and `nn` are.

PyTorch core does not have `models` and they keep things under `nn`. We see [components consisting Transformer under `nn`](https://pytorch.org/docs/stable/nn.html#transformer-layers), but also the `Transformer` class itself is described as *Model*.

I think `Emformer` and `Conformer` are analogous to `Transformer`, even though from the viewpoint of RNN-Transducer, it is a part of the whole transducer model. We put models under `torchaudio.models` module.

In general, it is easy to add a new user-facing interface, but much harder to remove one. Having similar submodules confuse users as well.

Having said that, I think it's safer to use `torchaudio.prototype.models` rather than `torchaudio.prototype.nn`.
What do you think?

hwangjeff:
that's fair — i do think `models` is less precise, but i agree that it'd be confusing to have both `models` and `nn`. i'll use `torchaudio.prototype.models`.

should i go ahead and use `torchaudio.prototype.pipelines` for RNNTBundle and EMFORMER_RNNT_BASE_LIBRISPEECH while i'm at it? and, would the decoder classes belong best in `torchaudio.prototype.models`?

mthrok:
> should i go ahead and use `torchaudio.prototype.pipelines` for RNNTBundle and EMFORMER_RNNT_BASE_LIBRISPEECH while i'm at it? and, would the decoder classes belong best in `torchaudio.prototype.models`?

`pipelines` makes sense as well. The decoders are new and unique to audio domain, so at this moment I am not sure. RNN-T Decoder is related to NN so I feel it's understandable if it were placed under models, but n-gram-based decoder is not NN, and it is more about beam **search**, so it would be strange to put it under models, even though it integrates with language **model**.

docs/source/prototype.models.rst
```diff
@@ -0,0 +1,81 @@
...
+
+Model Factory Functions
+-----------------------
+
+emformer_rnnt_base
+~~~~~~~~~~~~~~~~~~
+
+.. autofunction:: emformer_rnnt_base
+
+emformer_rnnt_model
```
nateanl, line 54: I feel like ``emformer_rnnt_model`` could be above ``emformer_rnnt_base``, as the base model is based on ``emformer_rnnt_model``.

nateanl (approved):
LGTM. Just some minor suggestion.

mthrok (approved):

carolineechen (approved):

PR #2112, opened 2021-12-29
Allow token list as CTC decoder input

carolineechen:
Additionally accept list of tokens as CTC decoder input. This makes it possible to directly pass in something like `bundles.get_labels()` into the decoder factory function instead of requiring a separate tokens file.

### DIFF
--- a/test/torchaudio_unittest/prototype/ctc_decoder_test.py
+++ b/test/torchaudio_unittest/prototype/ctc_decoder_test.py
@@ -1,4 +1,5 @@
 import torch
+from parameterized import parameterized
 from torchaudio_unittest.common_utils import (
     TempDirMixin,
     TorchaudioTestCase,
@@ -9,21 +10,24 @@ from torchaudio_unittest.common_utils import (
 
 @skipIfNoCtcDecoder
 class CTCDecoderTest(TempDirMixin, TorchaudioTestCase):
-    def _get_decoder(self):
+    def _get_decoder(self, tokens=None):
         from torchaudio.prototype.ctc_decoder import kenlm_lexicon_decoder
 
         lexicon_file = get_asset_path("decoder/lexicon.txt")
-        tokens_file = get_asset_path("decoder/tokens.txt")
         kenlm_file = get_asset_path("decoder/kenlm.arpa")
+
+        if tokens is None:
+            tokens = get_asset_path("decoder/tokens.txt")
 
         return kenlm_lexicon_decoder(
             lexicon=lexicon_file,
-            tokens=tokens_file,
+            tokens=tokens,
             kenlm=kenlm_file,
         )
 
-    def test_construct_decoder(self):
-        self._get_decoder()
+    @parameterized.expand([(get_asset_path("decoder/tokens.txt"),), (["-", "|", "f", "o", "b", "a", "r"],)])
+    def test_construct_decoder(self, tokens):
+        self._get_decoder(tokens)
 
     def test_shape(self):
         B, T, N = 4, 15, 10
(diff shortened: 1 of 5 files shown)


test/torchaudio_unittest/prototype/ctc_decoder_test.py
```diff
@@ -9,21 +10,24 @@
...
-    def _get_decoder(self):
+    def _get_decoder(self, tokens=None):
         from torchaudio.prototype.ctc_decoder import kenlm_lexicon_decoder
 
         lexicon_file = get_asset_path("decoder/lexicon.txt")
-        tokens_file = get_asset_path("decoder/tokens.txt")
         kenlm_file = get_asset_path("decoder/kenlm.arpa")
 
+        if tokens == None:
+            tokens = get_asset_path("decoder/tokens.txt")
```
hwangjeff, line 20: is this branch necessary?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -177,7 +177,7 @@ def kenlm_lexicon_decoder(
 
     Args:
         lexicon (str): lexicon file containing the possible words
-        tokens (str): file containing valid tokens
+        tokens (str or List[str]): file or list containing valid tokens
```
hwangjeff, line 180: not in the scope of the pr, but is there some reference that we can cite to help users understand how tokens files should be formatted?

hwangjeff (approved):

test/torchaudio_unittest/prototype/ctc_decoder_test.py
carolineechen: this is just to set the default behavior if the `tokens` parameter is not passed in to this function, as in the `test_shape` test, without having a function call being used in the function header

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: good point, this is somewhat shown in the [tutorial](https://pytorch.org/audio/main/tutorials/asr_inference_with_ctc_decoder_tutorial.html#tokens) but could definitely be improved. I'm thinking either in some README section (w/ docstrings linking to it), directly in the docstrings, or in more detail in a tutorial if you have any thoughts?

mthrok: I think we can get rid of loading from the file-path.
`List[str]` is simple to construct, users can format data such way after loading data from a file in a different format.
I am looking at the changes to tests, but keeping the pass-like type increase the test cost.
Specifically, I think `test_construct_decoder` is getting redundant yet not enough without `test_index_to_tokens`.

carolineechen: there is a case from flashlight [README](https://github.com/flashlight/flashlight/blob/9193b92e86538128664c1cf22785f4f5778c8312/flashlight/app/asr/README.md#token-dictionary) that can not be done using `List[str]`, where they mention `If two tokens are on the same line in the tokens file, they are mapped to the same index for training/decoding`. I'm not sure of specific use cases of this but may be good to have the option open?

nateanl (approved):
LGTM

torchaudio/prototype/ctc_decoder/ctc_decoder.py
mthrok: Huh🤔 so in the case of ASR, this is to allow one phoneme to be mapped to different tokens. 

Okay, then let's keep it.

mthrok (approved):

torchaudio/prototype/ctc_decoder/ctc_decoder.py
hwangjeff: > I'm thinking either in some README section (w/ docstrings linking to it), directly in the docstrings, or in more detail in a tutorial if you have any thoughts?

yeah that all seems reasonable. i think if the format is simple enough to describe concisely, it'd be ideal to include it directly in the docstrings for ease of access. otherwise, linking to a readme sounds good

re: the file loading — i guess if the file format is standardized and something users are accustomed to, it'd be helpful to allow for directly loading files of that format. otherwise, it seems like pulling out the file loading logic would yield a more minimalist/flexible solution. the one-index-to-many-tokens case wouldn't necessarily require file loading — we could account for it with a list of lists or dictionary input

Issue #2051, opened 2021-11-29
Dropping Python 3.6 Support


mthrok:
torchaudio will drop Python 3.6 support soon, following PyTorch core.

https://github.com/pytorch/pytorch/issues/69005

---

### 1. Switch stylecheck job
- [x] Move stylecheck test to newer Python. (This is already doable)
  1. Change the following code so that the job is carried out on some other Python version. (considering the next work item, I think it's okay to hard code Python version for this specific job)
    https://github.com/pytorch/audio/blob/fd3c957399d85c003d5d80859f8153b8e8a7f0c1/.circleci/regenerate.py#L233-L242
  2. Run `.circleci/regenerate.py` and check-ing the new `.circleci/config.yml`.

### 2. Remove `3.6` from the test matrix
- [x] Remove `3.6` from the test matrix.
  1. Update this line
     https://github.com/pytorch/audio/blob/fd3c957399d85c003d5d80859f8153b8e8a7f0c1/.circleci/regenerate.py#L24
  2. Run `.circleci/regenerate.py` and check-ing the new `.circleci/config.yml`.

### 3. Update documentation
- [ ] Update [the dependencies section of README](https://github.com/pytorch/audio/blob/main/README.md#dependencies)
- [ ] Update the description in [`setup.py`](https://github.com/pytorch/audio/blob/fd3c957399d85c003d5d80859f8153b8e8a7f0c1/setup.py#L163-L166)

wernerchao:
Working on item (2).

wernerchao:
Working on item (3)

PR #2119, opened 2021-12-31
Drop support for python3.6 as per task T109096383. Item 2 on issue 2051.

wernerchao:
As per item 2 on [issue 2051](https://github.com/pytorch/audio/issues/2051), dropping support for python 3.6.
Removed 3.6 from test matrix and ran `.circleci/regenerate.py `.


mthrok (approved):
Looks good. Thanks!

PR #2124, opened 2022-01-02
Update ffmpeg discovery logic

mthrok:
Update ffmpeg discovery logic

Previously the build process used pkg-config to locate
an installation of ffmpeg, which does not work well Windows/CentOS.

This commit update the discovery process to use the custom
FindFFMPEG.cmake adopted from Kitware/VTK repository with addition of
conda environment.

 The custom discovery logic can support Windows and CentOS.

### DIFF
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -17,30 +17,11 @@ endif()
 # ffmpeg
 ################################################################################
 if (BUILD_FFMPEG)
-  if(WIN32)
-    message(FATAL_ERROR "FFmpeg integration is not supported on Windows.")
-    # TODO
-    # Since pkg-config is not well-supported on Windows (for example, ffmpeg
-    # installed from conda-forge does not include `.pc` medata file),
-    # to support windows, we need to fill interface manually.
-    #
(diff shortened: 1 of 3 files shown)


carolineechen (approved):

third_party/CMakeLists.txt
```diff
@@ -17,30 +17,11 @@ endif()
...
-    # requires ffmpeg>=4.1
-    libavdevice>=58
-    libavfilter>=7
-    libavformat>=58
-    libavcodec>=58
-    libswresample>=3
-    libswscale>=3
-    libavutil>=56
-    )
+  set(FFMPEG_FIND_COMPONENTS avdevice avfilter avformat avcodec avutil)
```
carolineechen, line 20: is there a reason why swresample and swscale are removed?

mthrok: `swresample` and `swscale` implements algorithms on lawer-level.
`avfilter` library wraps them and provides high-level functions that we use.

`swresample` and `swscale` are required at runtime, but not during the build time as we do not include their header directly.

PR #2126, opened 2022-01-03
Add device arg to Resample class

Sorrow321:
This PR adds the ability to specify the device on which resampling would compute in Resample class.

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -895,6 +895,7 @@ class Resample(torch.nn.Module):
         rolloff: float = 0.99,
         beta: Optional[float] = None,
         *,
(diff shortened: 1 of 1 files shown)


mthrok:
Hi @Sorrow321 

To run the resample in GPU, it is advised to use `.to` as in CUDA semantics https://pytorch.org/docs/stable/notes/cuda.html

```python
resampler = torchaudio.transforms.Resample(...)
resampler.to(torch.device('cuda'))
```

Sorrow321:
Ok, thanks! Then my PR isn't useful. Maybe consider adding some hint to the docs, because it's not obvious that it can be used in this way

PR #2128, opened 2022-01-04
Add feature mean square value to HuBERT Pretrain model output

nateanl:
In [Fairseq](https://github.com/pytorch/fairseq/blob/main/examples/hubert/config/pretrain/hubert_base_librispeech.yaml#L48), the training applies additional penalty loss besides the cross-entropy losses. This PR adds the feature's mean square value to the model output to support such penalty loss.


mthrok (approved):
Stamp.

Note: Since `HuBERTPretrainModel` is not released yet, this PR is not `BC-Breaking`.

PR #2129, opened 2022-01-04
Add training recipe for hubert_pretrain_base model

nateanl:


examples/hubert/lightning.py
```diff
@@ -0,0 +1,127 @@
...
+from pytorch_lightning import LightningModule
+from torch import Tensor
+from torch.optim.optimizer import Optimizer
+from torch.utils.data import DataLoader
+
+
+Batch = Tuple[Tensor, Tensor, Tensor]
+
+
+class PolynomialDecayLRScheduler(torch.optim.lr_scheduler._LRScheduler):
```
hwangjeff, line 16: is the following correct?
- for training iterations in [0, `warmup_updates`], we maintain a constant learning rate of `base_lr / warmup_updates`
- in (`warmup_updates`, `max_updates`), we linearly decrease learning rate from ~`base_lr` to 0
- `max_updates` and beyond, we hold the learning rate at 0

if so, what does "polynomial" refer to? and is it expected that the learning rate jumps from `base_lr / warmup_updates` to roughly `base_lr` after the `warmup_updates`-th iteration?

nateanl: This LRScheduler is from [fairseq](https://github.com/pytorch/fairseq/blob/fcca32258c8e8bcc9f9890bf4714fa2f96b6b3e1/fairseq/optim/lr_scheduler/polynomial_decay_schedule.py#L74-L89). The logic should be like this:
- from [0, warmup_updates] steps, increase the learning rate from initial lr to lr * num_step / warmup_updates
- from [warmup_updates, max_updates], decrease the learning rate by the formula.
- from [max_updaes, ], hold the learning rate at 0.

nateanl:
> I think that the commit history got messed up.

Oops, let me fix that.

examples/hubert/lightning.py
nateanl: hmm, the impl doesn't seem correct. num_step / warm_updates is smaller than 1, how can they call it warmup then...

examples/hubert/lightning.py
```diff
@@ -0,0 +1,139 @@
...
+        verbose: bool = False,
+    ):
+        self.warmup_updates = warmup_updates
+        self.warmup_factor = 1.0 / self.warmup_updates
+        self.max_updates = max_updates
+        super().__init__(optimizer, last_epoch=last_epoch, verbose=verbose)
+
+    def get_lr(self):
+        if self._step_count <= self.warmup_updates:
+            return [self.warmup_factor * base_lr for base_lr in self.base_lrs]
```
hwangjeff, line 34: in that case, should this line be amended? e.g.

```suggestion
            return [base_lr * self._step_count / self.warmup_updates for base_lr in self.base_lrs]
```

examples/hubert/lightning.py
hwangjeff: also, perhaps we should call this something other than "polynomial" since it isn't applying a polynomial function to `pct_remaining` as in the original implementation

examples/hubert/lightning.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def get_lr(self):
+        if self._step_count <= self.warmup_updates:
+            return [self._step_count / self.warmup_updates * base_lr for base_lr in self.base_lrs]
+        elif self.optimizer._step_count >= self.max_updates:
+            return [0.0 for _ in self.base_lrs]
+        else:
+            lrs = []
+            for base_lr in self.base_lrs:
+                pct_remaining = (self.max_updates - self._step_count) / (self.max_updates - self.warmup_updates)
```
hwangjeff, line 39: this quantity doesn't depend on `base_lr`, so we can move it out and perhaps use a list comprehension to be consistent with the previous cases

examples/hubert/lightning.py
```diff
@@ -0,0 +1,137 @@
...
+        learning_rate: float,
+        betas: Tuple[float, float],
+        eps: float,
+        weight_decay: float,
+        warmup_updates: int,
+        max_updates: int,
+    ):
+        super().__init__()
+
+        if model_name == "huebrt_pretrain_base":
```
hwangjeff, line 64:
```suggestion
        if model_name == "hubert_pretrain_base":
```

and similarly for the other `model_name` cases

examples/hubert/lightning.py
```diff
@@ -0,0 +1,137 @@
...
+        verbose: bool = False,
+    ):
+        self.warmup_updates = warmup_updates
+        self.max_updates = max_updates
+        super().__init__(optimizer, last_epoch=last_epoch, verbose=verbose)
+
+    def get_lr(self):
+        if self._step_count <= self.warmup_updates:
+            return [self._step_count / self.warmup_updates * base_lr for base_lr in self.base_lrs]
+        elif self.optimizer._step_count >= self.max_updates:
```
hwangjeff, line 34: might want to check on whether `self._step_count` and `self.optimizer._step_count` differ in any way, and then stick to using just one throughout this function

examples/hubert/lightning.py
```diff
@@ -0,0 +1,137 @@
...
+        elif model_name == "huebrt_pretrain_xlarge":
+            self.model = torchaudio.models.hubert_pretrain_xlarge()
+        else:
+            raise ValueError(f"Unsupported model name: {model_name}")
+
+        self.loss = hubert_loss
+        self.optimizer = torch.optim.Adam(
+            self.model.parameters(), lr=learning_rate, betas=betas, eps=eps, weight_decay=weight_decay
+        )
+        self.lr_scheduler = PolynomialDecayLRScheduler(self.optimizer, warmup_updates, max_updates)
```
hwangjeff, line 77:
```suggestion
        self.lr_scheduler = LinearDecayLRScheduler(self.optimizer, warmup_updates, max_updates)
```

examples/hubert/train.py
```diff
@@ -0,0 +1,166 @@
...
+        max_updates=args.max_updates,
+    )
+    trainer.fit(model)
+
+
+def cli_main():
+    parser = ArgumentParser()
+    parser.add_argument(
+        "--model-name",
+        default="huebrt_pretrain_base",
```
hwangjeff, line 65:
```suggestion
        default="hubert_pretrain_base",
```

etc for below

examples/hubert/lightning.py
nateanl: I tested the `_step_count` of optimizer and lr_scheduler in [this script](https://gist.github.com/nateanl/6dbae8b49320f87b1452680f01a044f5). `self._step_count` makes more sense because the optimizer's step_count starts with 0. If we use `optimizer._step_count`, the first learning rate will be 0.

examples/hubert/train.py
```diff
@@ -0,0 +1,166 @@
+import pathlib
```
mthrok, line 1: Since this is a script, can you add shebang line and docsrtring, the reuse the docsrting for help message of CLI?

```python
#!/usr/bin/env python3
"""docsrting"""
```

then

```
class _Formatter(ArgumentDefaultsHelpFormatter, RawDescriptionHelpFormatter):
    # https://stackoverflow.com/a/18462760
    pass

parser = ArgumentParser(
    description=__doc__,
    formatter_class=_Formatter,
)
```

examples/hubert/train.py
```diff
@@ -0,0 +1,166 @@
...
+        eps=args.eps,
+        weight_decay=args.weight_decay,
+        warmup_updates=args.warmup_updates,
+        max_updates=args.max_updates,
+    )
+    trainer.fit(model)
+
+
+def cli_main():
+    parser = ArgumentParser()
```
mthrok, line 62: Since the are many arguments, adding default value manually is not future-proof.
It should use `ArgumentDefaultsHelpFormatter`. (see my previous comment)

examples/hubert/train.py
```diff
@@ -0,0 +1,166 @@
...
+        betas=args.betas,
+        eps=args.eps,
+        weight_decay=args.weight_decay,
+        warmup_updates=args.warmup_updates,
+        max_updates=args.max_updates,
+    )
+    trainer.fit(model)
+
+
+def cli_main():
```
mthrok, line 61: main function should be showing the gist of the execution flow.

Can you extract the argument parse logic outside of main function and put the content of `run_train` in main?

examples/hubert/train.py
nateanl: Sure, will update it.

nateanl:
@mthrok Looks like the commit history messed up😅, I will close this PR and address the comments in a new one.

PR #2130, opened 2022-01-04
Add librispeech inference script

carolineechen:
add script for running CTC beam search decoder on librispeech dataset with torchaudio pretrained wav2vec2 models

### DIFF
--- a/examples/asr/librispeech_ctc_decoder/README.md
+++ b/examples/asr/librispeech_ctc_decoder/README.md
@@ -11,7 +11,7 @@ Sample command
 python inference.py \
     --librispeech_path ./librispeech/ \
     --split test-other \
-    --model FT_960H \
+    --model WAV2VEC2_ASR_BASE_960H \
     --beam-size 1500 \
     --lm-weight 1.74 \
     --word-score 0.52
--- a/examples/asr/librispeech_ctc_decoder/README.md
+++ b/examples/asr/librispeech_ctc_decoder/README.md
@@ -1,0 +1,18 @@
+# Speech Recognition Inference with CTC Beam Search Decoder
+
+This is an example inference script for running decoding on the LibriSpeech dataset and wav2vec 2.0 models, using a CTC beam search decoder that supports lexicon constraint and language model integration. The language model used is a 4-gram KenLM trained on the LibriSpeech dataset.
+
+## Usage
+Additional command line parameters and information can is available with the `--help` option.
+
+Sample command
+
+```
+python inference.py \
+    --librispeech_path ./librispeech/ \
+    --split test-other \
+    --model FT_960H \
+    --beam-size 1500 \
+    --lm-weight 1.74 \
+    --word-score 0.52
+```
--- a/examples/asr/librispeech_ctc_decoder/README.md
+++ b/examples/asr/librispeech_ctc_decoder/README.md
@@ -16,3 +16,12 @@ python inference.py \
     --lm-weight 1.74 \
     --word-score 0.52
 ```
+
+## Results
+The table below contains WER results for various pretrained models on the LibriSpeech test-other split, using a beam size of 1500, and language model weight and word insertion scores taken from Table 7 of [wav2vec 2.0](https://arxiv.org/pdf/2006.11477.pdf).
+
+|                                                                                          Model |     WER |
+|:----------------------------------------------------------------------------------------------:|--------:|
+| [WAV2VEC2_ASR_BASE_10M](https://pytorch.org/audio/main/pipelines.html#wav2vec2-asr-base-10m)   |   0.1591|
+| [WAV2VEC2_ASR_BASE_100H](https://pytorch.org/audio/main/pipelines.html#wav2vec2-asr-base-100h) |   0.0807|
+| [WAV2VEC2_ASR_BASE_960H](https://pytorch.org/audio/main/pipelines.html#wav2vec2-asr-base-960h) |   0.0615|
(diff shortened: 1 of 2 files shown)


examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,126 @@
...
+
+    dataset = torchaudio.datasets.LIBRISPEECH(args.librispeech_path, url=args.split, download=False)
+
+    total_edit_distance = 0
+    total_length = 0
+    for idx, sample in enumerate(dataset):
+        waveform, _, transcript, _, _, _ = sample
+        transcript = transcript.strip().lower().strip()
+
+        emission, _ = model(waveform)
```
nateanl, line 60: Will it be better to run the model inference under torch.inference_mode?
```python
with torch.inference_mode():
```

examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,126 @@
...
+
+import torch
+import torchaudio
+from torchaudio.prototype.ctc_decoder import kenlm_lexicon_decoder
+
+
+bundle_map = {
+    "FT_10M": torchaudio.pipelines.WAV2VEC2_ASR_BASE_10M,
+    "FT_100H": torchaudio.pipelines.WAV2VEC2_ASR_BASE_100H,
+    "FT_960H": torchaudio.pipelines.WAV2VEC2_ASR_BASE_960H,
```
mthrok, line 12: Instead of defining a closed set of the supported models, making it open will reduce the maintenance cost.

For example, we can use `bundle = getattr(torchaudio.pipelines, 'model')`.

examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,119 @@
...
+        with torch.inference_mode():
+            emission, _ = model(waveform)
+        results = decoder(emission)
+        hypothesis = " ".join(results[0][0].words).lower().strip()
+
+        total_edit_distance += torchaudio.functional.edit_distance(transcript.split(), hypothesis.split())
+        total_length += len(transcript.split())
+
+        if idx % 100 == 0:
+            print(f"Processed elem {idx}; WER: {total_edit_distance / total_length}")
```
mthrok, line 62: Can we use `logging` module and add timestamp? So that users can have some idea on how long it will take, or look back the log and see how long it took.

examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,119 @@
...
+    for idx, sample in enumerate(dataset):
+        waveform, _, transcript, _, _, _ = sample
+        transcript = transcript.strip().lower().strip()
+
+        with torch.inference_mode():
+            emission, _ = model(waveform)
+        results = decoder(emission)
+        hypothesis = " ".join(results[0][0].words).lower().strip()
+
+        total_edit_distance += torchaudio.functional.edit_distance(transcript.split(), hypothesis.split())
```
mthrok, line 58: `hypothesis` is `join`ed previously and `split` here. This could produce a wrong result.
Since the part to write out to file is removed, I think it's redundant to `join`.

examples/asr/librispeech_ctc_decoder/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Usage
+Additional command line parameters and information can is available with the `--help` option.
+
+Sample command
+
+```
+python inference.py \
+    --librispeech_path ./librispeech/ \
+    --split test-other \
+    --model FT_960H \
```
mthrok, line 14:
```suggestion
    --model WAV2VEC2_ASR_BASE_960H \
```

examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,124 @@
+import argparse
+import logging
+from typing import Optional
+
+import torch
+import torchaudio
+from torchaudio.prototype.ctc_decoder import kenlm_lexicon_decoder
+
+
+logging.basicConfig(format="%(asctime)s %(message)s", level=logging.INFO, datefmt="%Y-%m-%d %H:%M:%S")
```
mthrok, line 10: Since this is a standalone script, this is fine, but in general, the initialization of logger module should be delayed until the control flow reaches the entrypoint. That way, logger can be customized based on the CLI options.

```python
def _main():
    args = _parse_args()
    _init_logger(args.debug)
    run_inference(args)

def _parse_args():
    parser = ArgumentParser(...)
    parser.add_argument('--debug', action='store_true')
    ...
    return parser.parse_args()

def _init_logger(debug):
    fmt = '<more metadata>' if debug else 'simple'
    level = logging.DEBUG if debug else logging.INFO
    logging.basicConfig(format=fmt, level=level)

if __name__ == '__main__':
    _main()
```

examples/asr/librispeech_ctc_decoder/inference.py
```diff
@@ -0,0 +1,124 @@
...
+        total_edit_distance += torchaudio.functional.edit_distance(transcript.split(), hypothesis.split())
+        total_length += len(transcript.split())
+
+        if idx % 100 == 0:
+            logger.info(f"Processed elem {idx}; WER: {total_edit_distance / total_length}")
+    logger.info(f"Final WER: {total_edit_distance / total_length}")
+
+
+def cli_main():
+    parser = argparse.ArgumentParser(
```
mthrok, line 72: Can you extract the parser part? See my previous comment.

mthrok (approved):

PR #2141, opened 2022-01-05
Add parameter usage to CTC inference tutorial

carolineechen:
Add explanation and demonstration of different beam search decoder parameters.
Additionally use a better sample audio file and load in with token list instead of tokens file.

### DIFF
--- a/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
+++ b/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
@@ -153,6 +153,9 @@ torch.hub.download_url_to_file(lexicon_url, lexicon_file)
 # the binarized ``.bin`` LM can be used, but the binary format is
 # recommended for faster loading.
 #
+# The language model used in this tutorial is a 4-gram KenLM trained using
+# `LibriSpeech <http://www.openslr.org/11>`__.
+#
 
 kenlm_url = "https://pytorch.s3.amazonaws.com/torchaudio/tutorial-assets/ctc-decoding/4-gram-librispeech.bin"
 kenlm_file = f"{hub_dir}/kenlm.bin"
@@ -163,12 +166,10 @@ torch.hub.download_url_to_file(kenlm_url, kenlm_file)
 # Construct Beam Search Decoder
 # -----------------------------
 #
-# The decoder can be constructed using the ``kenlm_lexicon_decoder``
-# factory function from ``torchaudio.prototype.ctc_decoder``. In addition
-# to the previously mentioned components, it also takes in various beam
-# search decoding parameters and token/word parameters. The full list of
-# parameters can be found
-# :py:func:`here <torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder>`.
+# The decoder can be constructed using the factory function
+# :py:func:`kenlm_lexicon_decoder <torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder>`.
+# In addition to the previously mentioned components, it also takes in various beam
+# search decoding parameters and token/word parameters.
 #
 
 from torchaudio.prototype.ctc_decoder import kenlm_lexicon_decoder
@@ -195,6 +196,8 @@ beam_search_decoder = kenlm_lexicon_decoder(
 # basic greedy decoder.
 #
 
+from typing import List
+
 
 class GreedyCTCDecoder(torch.nn.Module):
     def __init__(self, labels, blank=0):
@@ -202,19 +205,19 @@ class GreedyCTCDecoder(torch.nn.Module):
         self.labels = labels
         self.blank = blank
 
-    def forward(self, emission: torch.Tensor) -> str:
-        """Given a sequence emission over labels, get the best path string
+    def forward(self, emission: torch.Tensor) -> List[str]:
+        """Given a sequence emission over labels, get the best path
         Args:
           emission (Tensor): Logit tensors. Shape `[num_seq, num_label]`.
 
         Returns:
-          str: The resulting transcript
+          List[str]: The resulting transcript
         """
         indices = torch.argmax(emission, dim=-1)  # [num_seq,]
         indices = torch.unique_consecutive(indices, dim=-1)
         indices = [i for i in indices if i != self.blank]
         joined = "".join([self.labels[i] for i in indices])
-        return joined.replace("|", " ").lower().strip().split()
+        return joined.replace("|", " ").strip().split()
 
 
 greedy_decoder = GreedyCTCDecoder(tokens)
@@ -253,7 +256,7 @@ print(f"WER: {greedy_wer}")
 #
 
 beam_search_result = beam_search_decoder(emission)
-beam_search_transcript = " ".join(beam_search_result[0][0].words).lower().strip()
+beam_search_transcript = " ".join(beam_search_result[0][0].words).strip()
 beam_search_wer = torchaudio.functional.edit_distance(actual_transcript, beam_search_result[0][0].words) / len(
     actual_transcript
 )
@@ -287,13 +290,14 @@ print(f"WER: {beam_search_wer}")
 #
 
 
-def get_transcript_and_time(decoder, emission):
+def print_decoded(decoder, emission, param, param_value):
     start_time = time.monotonic()
     result = decoder(emission)
     decode_time = time.monotonic() - start_time
 
     transcript = " ".join(result[0][0].words).lower().strip()
-    return transcript, decode_time
+    score = result[0][0].score
+    print(f"{param} {param_value:<3}: {transcript} (score: {score:.2f}; {decode_time:.4f} secs)")
 
 
 ######################################################################
@@ -307,7 +311,7 @@ def get_transcript_and_time(decoder, emission):
 #
 
 for i in range(3):
-    transcript = " ".join(beam_search_result[0][i].words).lower().strip()
+    transcript = " ".join(beam_search_result[0][i].words).strip()
     score = beam_search_result[0][i].score
     print(f"{transcript} (score: {score})")
 
@@ -340,8 +344,7 @@ for beam_size in beam_sizes:
         word_score=WORD_SCORE,
     )
 
-    transcript, decode_time = get_transcript_and_time(beam_search_decoder, emission)
-    print(f"beam size {beam_size:<3}: {transcript} ({decode_time:.4f} secs)")
+    print_decoded(beam_search_decoder, emission, "beam size", beam_size)
 
 
 ######################################################################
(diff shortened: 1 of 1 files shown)


examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -161,26 +162,27 @@
...
-# :py:func:`torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder`
-# factory function.
-# In addition to the previously mentioned components, it also takes in
-# various beam search decoding parameters and token/word parameters.
+# The decoder can be constructed using the ``kenlm_lexicon_decoder``
+# factory function from ``torchaudio.prototype.ctc_decoder``. In addition
+# to the previously mentioned components, it also takes in various beam
+# search decoding parameters and token/word parameters. The full list of
+# parameters can be found
+# `here <https://pytorch.org/audio/main/prototype.html#kenlm-lexicon-decoder>`__.
```
mthrok, line 170:
```suggestion
# :py:func:`here <torchaudio.prototype. kenlm_lexicon_decoder>`.
```

Since once this tutorial is released, then we will have multiple versions of this tutorials, so I think it's better to use interlink.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
 #
 # Now that we have the data, acoustic model, and decoder, we can perform
 # inference. Recall the transcript corresponding to the waveform is
-# ``"when it was the seven hundred and eighteenth night"``
+# ``"i really was very much afraid of showing him how much shocked i was at some parts of what he said"``
```
mthrok, line 227: I think the code-block highlights this better. The following is the comparison of double back-ticks, single back-tick and code block.

```

``"i really was very much afraid of showing him how much shocked i was at some parts of what he said"``

`"i really was very much afraid of showing him how much shocked i was at some parts of what he said"`

.. code::

    i really was very much afraid of showing him how much shocked i was at some parts of what he sai\
d
```

Screen Shot 2022-01-05 at 1 36 32 PM

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
 #
 
+actual_transcript = "i really was very much afraid of showing him how much shocked i was at some parts of what he said"
```
mthrok, line 230: This variable is always used as `actual_transcript.split()`, so I think you can `split()` here.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
 emission, _ = acoustic_model(waveform)
 
+
+######################################################################
+# The greedy decoder give the following result.
+#
+
+greedy_result = greedy_decoder(emission[0])
+greedy_transcript = greedy_result.replace("|", " ").lower().strip()
```
mthrok, line 239: I think `.replace("|", " ").lower().strip()` and `.split()` applied later can be put inside of `GreedyDecoder.forward` method.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
+print(f"Transcript: {beam_search_transcript}")
+print(f"WER: {beam_search_wer}")
+
 
 ######################################################################
-# Using the greedy decoder:
+# We see that the transcript with the lexicon-constrained beam search
+# decoder produces a more accurate result consisting of real words, while
+# the greedy decoder can predict incorrectly spelled words like “affrayd”
+# and “shoktd”.
```
mthrok, line 266: Nice!

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
-print(greedy_transcript)
+
+######################################################################
+# Beam Search Decoder Parameters
+# ------------------------------
+#
+# In this section, we go a little bit more in depth about some different
+# parameters and tradeoffs. For the full list of customizable parameters,
+# please refer to the
+# `documentation <https://pytorch.org/audio/main/prototype.ctc_decoder.html#torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder>`__.  # noqa
```
mthrok, line 277: Similar to the previous one, use of `:py:func:` should is more future-proof.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
+# and “shoktd”.
+#
 
-greedy_result = greedy_decoder(emission[0])
-greedy_transcript = greedy_result.replace("|", " ").lower().strip()
-print(greedy_transcript)
+
+######################################################################
+# Beam Search Decoder Parameters
+# ------------------------------
```
mthrok, line 272: not the scope of this PR but can we show how scores are computed? 

Like the one from [this](https://docs.google.com/presentation/d/1y9IHcy8F8rywgoTyvQbcsnlzzMbCSbPT2rW58pKGxms/edit) or cite a publication if there is as the linked one is more on the side of the implementation detail.

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +224,212 @@ def forward(self, emission: torch.Tensor) -> str:
...
+
+
+######################################################################
+# additional parameters
+# ~~~~~~~~~~~~~~~~~~~~~
+#
+# Additional parameters that can be optimized include the following -
+# ``word_score``: score to add when word finishes - ``unk_score``: unknown
+# word appearance score to add - ``sil_score``: silence appearance score
+# to add - ``log_add``: whether to use log add for lexicon Trie smearing
```
mthrok, line 434: Should this be list layout?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -36,7 +36,7 @@
 # working with
 #
 
-import os
+from time import time
```
mthrok, line 39: Since it's used for measuring the execution time, should use [`monotonic`](https://docs.python.org/3/library/time.html#time.monotonic).

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
carolineechen: sounds good, will leave this as a followup

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -210,10 +213,11 @@ def forward(self, emission: torch.Tensor) -> str:
         indices = torch.argmax(emission, dim=-1)  # [num_seq,]
         indices = torch.unique_consecutive(indices, dim=-1)
         indices = [i for i in indices if i != self.blank]
-        return "".join([self.labels[i] for i in indices])
+        joined = "".join([self.labels[i] for i in indices])
+        return joined.replace("|", " ").lower().strip().split()
```
mthrok, line 217: Looking at `replace("|", " ")`, I wonder if we should add an argument to [`get_labels()`](https://pytorch.org/audio/main/pipelines.html#torchaudio.pipelines.Wav2Vec2ASRBundle.get_labels) to customize this. I did not add this because I was looking the labels from written language perspective and not all the languages have spaces in their writing system. (i.e. Japanese) Meanwhile, in the decoding mechanism this is called silence token and silence is common to all the spoken languages.

mthrok, line 217: Can you update the return type of the signature and docstring?

mthrok: I think `lower` is redundant as the `tokens` passed to the constructor is already lowered.

mthrok (commented):
While at it, could you add a note on how KenLM model is trained?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -161,26 +163,27 @@
...
-# :py:func:`torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder`
-# factory function.
-# In addition to the previously mentioned components, it also takes in
-# various beam search decoding parameters and token/word parameters.
+# The decoder can be constructed using the ``kenlm_lexicon_decoder``
+# factory function from ``torchaudio.prototype.ctc_decoder``. In addition
+# to the previously mentioned components, it also takes in various beam
+# search decoding parameters and token/word parameters. The full list of
+# parameters can be found
+# :py:func:`here <torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder>`.
```
mthrok, line 171: nit: I think instead of waiting to provided the link to the document at the end end of section (`The full list of parameters ...`), we can link it in the first mention of the function.
(But you might have chosen the current working to work around the formatter warning.) 

```
# The decoder can be constructed using the :py:func:`torchaudio.prototype.ctc_decoder.kenlm_lexicon_decoder`.
# In addition to the previously mentioned components, it also takes in various beam
# search decoding parameters and token/word parameters.
```

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +228,215 @@ def forward(self, emission: torch.Tensor) -> str:
...
+######################################################################
+# beam size token
+# ~~~~~~~~~~~~~~~
+#
+# The ``beam_size_token`` parameter corresponds to the number of tokens to
+# consider for expanding each hypothesis at the decoding step. Exploring a
+# larger number of next possible tokens increases the range of potential
+# hypotheses at the cost of computation. This parameter is especially
+# helpful for lexicon-free decoders, where the number of possible next
+# tokens is not already partially constrained by the lexicon.
```
mthrok, line 358: >  This parameter is especially helpful for lexicon-free decoders, 
> where the number of possible next tokens is not already
> partially constrained by the lexicon.

Although this is informative and educative, we do not support a lexicon-free decoder (yet?), so maybe not as useful to users?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -222,28 +228,215 @@ def forward(self, emission: torch.Tensor) -> str:
...
+    transcript, decode_time = get_transcript_and_time(beam_search_decoder, emission)
+    print(f"beam size token {beam_size_token:<3}: {transcript} ({decode_time:.4f} secs)")
+
+
+######################################################################
+# beam threshold
+# ~~~~~~~~~~~~~~
+#
+# The ``beam_threshold`` parameter is used to prune the stored hypotheses
+# set at each decoding step, removing hypotheses whose scores are greater
```
mthrok, line 383: Isn't the higher the better?
https://514632-90321822-gh.circle-artifacts.com/0/docs/tutorials/asr_inference_with_ctc_decoder_tutorial.html#beam-size-token

Also, can you also show scores for each example after `nbest`? I was trying to check the score of other examples.

mthrok (approved):

PR #2143, opened 2022-01-05
Add HuBERT-feature support in preprocessing of HuBERT training

nateanl:
The HuBERT base model requires two iterations of training.
- In the first iteration, the MFCC features of the waveforms are applied to K-Means clustering to generate the labels. The labels are then used for masked-prediction training of the HuBERT base model.
- In the second iteration, instead of MFCC feature, the outputs of the intermediate layer in the first iteration's HuBERT model are used for K-Means clustering. Then the new labels are used for training a HuBERT base model from scratch.

This PR adds support for extracting HuBERT features for K-Means model training and prediction. Users can select which layer's output to generate labels.

### DIFF
--- a/examples/hubert/utils/feature_utils.py
+++ b/examples/hubert/utils/feature_utils.py
@@ -141,7 +141,11 @@ def dump_features(
             Options: [``torch.device('cpu')``, torch.device('cuda')``].
         feature_type (str, optional): The type of the desired feature. Options: [``mfcc``, ``hubert``].
             (Default: ``mfcc``)
-        sample_rate (int, optional): The sample rate of the audio. (Default: 16000)
+        layer_index (int, optional): The index of transformer layers in ``torchaudio.models.HuBERTPretrainModel``
+            for extracting features. Only active when ``feature_type`` is set to ``hubert``. (Default: ``6``)
+        checkpoint_path(Path or None, optional): The checkpoint path of ``torchaudio.models.HuBERTPretrainModel``.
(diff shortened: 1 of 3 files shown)


examples/hubert/utils/feature_utils.py
```diff
@@ -85,6 +125,8 @@ def dump_features(
     num_rank: int,
     device: torch.device,
     feature_type: str = "mfcc",
+    layer_index: int = 6,
+    checkpoint_path: Optional[Path] = None,
```
carolineechen, line 129: can you add these variables to the function docstrings?

examples/hubert/utils/feature_utils.py
```diff
@@ -99,7 +141,11 @@ def dump_features(
             Options: [``torch.device('cpu')``, torch.device('cuda')``].
         feature_type (str, optional): The type of the desired feature. Options: [``mfcc``, ``hubert``].
             (Default: ``mfcc``)
-        sample_rate (int, optional): The sample rate of the audio. (Default: 16000)
+        layer_index (int, optional): The index of transformer layers in ``torchaudio.models.HuBERTPretrainModel``
```
carolineechen, line 144: would it make more sense for this to be an `Optional[int]` type if it's unused if `feature_type` is not `hubert`?

carolineechen (approved):

examples/hubert/utils/feature_utils.py
nateanl: Good point, will update it.

carolineechen (approved):

PR #2144, opened 2022-01-05
[Example] Fix BucketizeSampler docstring

erip:
HuBERTDataSet has a `len_list` attribute that `Dataset`s don't have.


nateanl:
@erip Thanks for the PR! As we discussed in https://github.com/pytorch/text/issues/1474, let's make the `BucketizeSampler` generic to any Dataset. To make that, could you:
- Replace the `data_source` to `lengths`
- Add `shuffle` argument to decide whether to shuffle the samples in the buckets.
- Add `max_len` and `min_len` to filter out samples that are `> max_len` or `< min_len`. 

I will make adjustments to the `HuBERTDataset` as a follow-up after this PR. Thanks!

PR #2147, opened 2022-01-06
[Example] abstracts BucketizeSampler to be usable outside of HuBERT example.

erip:
This PR:

- Replaces the `data_source` with `lengths`
- Adds a `shuffle` argument to decide whether to shuffle the samples in the buckets.
- Add `max_len` and `min_len` to filter out samples that are > max_len or < min_len.

cc @nateanl

### DIFF
--- a/examples/hubert/dataset/hubert_dataset.py
+++ b/examples/hubert/dataset/hubert_dataset.py
@@ -53,7 +53,7 @@ class BucketizeSampler(BatchSampler):
         filtered_length_idx = [(length, i) for i, length in enumerate(lengths) if min_len <= length <= max_len]
         if len(filtered_length_idx) == 0:
             raise AssertionError("``lengths`` cannot be empty after filtering.")
-        sorted_filtered_length_idx = list(sorted(filtered_length_idx, key=lambda x: x[0]))
+        sorted_filtered_length_idx = sorted(filtered_length_idx, key=lambda x: x[0])
         self.lengths = [e[0] for e in sorted_filtered_length_idx]
         self.indices = [e[1] for e in sorted_filtered_length_idx]
         self.max_token_count = max_token_count
--- a/examples/hubert/dataset/hubert_dataset.py
+++ b/examples/hubert/dataset/hubert_dataset.py
@@ -20,12 +20,18 @@ class BucketizeSampler(BatchSampler):
     """Buketize sampler for data with different lengths to reduce number of paddings.
 
     Args:
-        data_source (Dataset): The dataset to sample
+        lengths (List[int]): The lengths of the examples in the dataset.e
         num_buckets (int): The number of buckets to split the data samples.
+        min_len (int: The minimum example lengths to keep.
+            (Default: 0)
+        max_len (int or None, optional): The maximum example lengths to keep. Inferred if not provided.
+            (Default ``None``)
         max_token_count (int or None, optional): The max number of tokens in one mini-batch.
             (Default: ``None``)
         batch_size (int or None, optional): The number of samples in one mini-batch.
              (Default: ``None``)
+        shuffle (bool): Whether to shuffle buckets for non-monotonic length sampling.
+             (Default True)
 
     Note: If ``max_token_count`` is not ``None``, the ``batch_size`` couldn't be set since
         the lengths of samples are unknown, the batch size may be different for different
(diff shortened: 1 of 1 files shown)


examples/hubert/dataset/hubert_dataset.py
```diff
@@ -34,65 +33,77 @@ class BucketizeSampler(BatchSampler):
...
+
+        if not (0 <= min_len <= max_len):
+            raise AssertionError("``min_len`` should be non-negative and smaller than ``max_len``")
         if max_token_count is not None and batch_size is not None:
             raise AssertionError("The ``max_token_count`` and ``batch_size`` can't be both set.")
-        self.data_source = data_source
+        # Filter out examples which are outside the bounds of [min_len, max_len]
+        self.lengths = [length for length in lengths if min_len <= length <= max_len]
+        if len(self.lengths) == 0:
+            raise AssertionError("``lengths`` cannot be empty after filtering.")
```
nateanl, line 54: ``lengths`` needs to be sorted and a new indices list is used for batching. Although bucketization already gathered the samples with similar lengths, sorting the lengths can help reduce the gap when transferring to a new bucket.

For example:
batch_size = 16
batches = [0:8] is from bucket 0
bucket[8:16] is from bucket 100
The length different will be large.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -1,31 +1,30 @@
...
-from torch.utils.data import Dataset, BatchSampler
+from torch.utils.data import BatchSampler, Dataset
 
 
 class BucketizeSampler(BatchSampler):
     """Buketize sampler for data with different lengths to reduce number of paddings.
 
     Args:
-        data_source (Dataset): The dataset to sample
+        lengths (List[int]): The lengths of the examples in the dataset.e
```
nateanl, line 16:
```suggestion
        lengths (List[int]): The lengths of the samples in the dataset.
```

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -34,65 +33,77 @@ class BucketizeSampler(BatchSampler):
...
 
-    def _get_buckets(self, data_source: Dataset, num_buckets: int) -> Dict[int, Tensor]:
+    def _get_buckets(self, lengths: List[int], num_buckets: int, min_len: int, max_len: int) -> Dict[int, Tensor]:
         """Generate buckets based on the dataset.
         Args:
-            data_source (Dataset): The dataset object to bucketize.
+            lengths (List[int]): The lengths of the examples in the dataset.e
             num_buckets (int): The number of buckets.
+            min_int (int): The lower bound of the evenly spaced length intervals to determine bucket width.
+            max_int (int): The upper bound of the evenly spaced length intervals to determine bucket width.
```
nateanl, line 66:
```suggestion
            min_len (int): The lower bound of the evenly spaced length intervals to determine bucket width.
            max_len (int): The upper bound of the evenly spaced length intervals to determine bucket width.
```

nateanl (commented):
LGTM! Could you address the comments and update the PR? Thanks.

examples/hubert/dataset/hubert_dataset.py
erip: Oops, good catch. :-)

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -34,65 +33,78 @@ class BucketizeSampler(BatchSampler):
...
+            max_len = max(lengths)
+
+        if not (0 <= min_len <= max_len):
+            raise AssertionError("``min_len`` should be non-negative and smaller than ``max_len``")
         if max_token_count is not None and batch_size is not None:
             raise AssertionError("The ``max_token_count`` and ``batch_size`` can't be both set.")
-        self.data_source = data_source
+        # Filter out examples which are outside the bounds of [min_len, max_len]
+        # sort to minimize gap when bucketizing.
+        self.lengths = list(sorted(length for length in lengths if min_len <= length <= max_len))
```
nateanl, line 53: The indices needs to be updated since `lengths` order is changed.
```python
length_id = [(length, i) for i, length in enumerate(lengths) if min_len <= length <= max_len]
sorted_length_id = list(sorted(length_id, key = lambda x : x[0]))
self.lengths = ...
self.indices = ...
```

erip: I see. So the `for i, _ in enumerate(len_list)` from earlier code should just be `for i in self.indices`?

erip: Maybe to make it equivalent it would be `for i, _ in zip(self.indices, self.lengths)`, but the zip and ignore is basically a no-op.

nateanl: Exactly. ``self.indices`` makes more sense. We just need to make sure the index matches the one from the original dataset.

erip: I have something like this:

```patch
+        filtered_lengths = torch.tensor([length for length in lengths if min_len <= length <= max_len])
+        if len(filtered_lengths) == 0:
             raise AssertionError("``lengths`` cannot be empty after filtering.")
+        self.length_idx = torch.argsort(filtered_lengths)
+        self.lengths = filtered_lengths[length_idx]
         self.max_token_count = max_token_count
         self.batch_size = batch_size
-        self.buckets = self._get_buckets(self.lengths, num_buckets, min_len, max_len)
+        self.buckets = self._get_buckets(self.reordered_lengths, self.length_idx, num_buckets, min_len, max_len)
```

where `_get_buckets`s only code change is that the enumeration is replaced by iteration over `self.indices`. Basically it's keeping track of the original place of the example in the dataset. This seems pretty close to my implementation -- sorry it's taking a bit of time for this to click. 😅

nateanl: Thanks. When we filter out the lengths, the indices already changed. So we can sort the lengths first to get the indices, then filter the lengths and indices, what do you think?

erip: Yup, realized that right after posting this diff. OK, I'll make these changes quickly and push it. Looks like CI failures are unrelated, too...

nateanl (approved):
LGTM! Thanks for working on it.

nateanl:
cc @ejguan This PR will be helpful when designing the bucketizeDataPipe.

erip:
Lingering TODO is to fix the `HuBERTDataSet` to use this new interface. Do you want to take that @nateanl or shall I?

nateanl:
Sure. I will do it. Testing the Dataset requires running the preprocessing, but since I have done that part it's easy to me.

PR #2149, opened 2022-01-07
Exclude numpy version that has security issue

nateanl:


docs/requirements-tutorials.txt
```diff
@@ -3,4 +3,4 @@ deep-phonemizer
 boto3
 pandas
 librosa
-numpy<=1.20
```
mthrok, line 6: Since PyTorch requires `>=1.11`, let's just do `numpy >= 1.21` and see if the `build_doc` job succeeds.

mthrok, line 6: Let's keep this as-is until `numba` supports `numpy >=1.21`.
Theoretically, this is only for building documentation, and it should not affect the requirements of a service using torchaudio.

nateanl: Sure. Let's wait the news from numba

PR #2150, opened 2022-01-07
[Example] Refactor BucketizeBatchSampler and HuBERTDataset

nateanl:
- Rename `BucketizeSampler` to `BucketizeBatchSampler`
- Fix bugs in `BucketizeBatchSampler`
- Adjust HuBERTDataset based on the latest `BucketizeBatchSampler`.


nateanl:
Here is the document page for ``BucketizeBatchSampler``: https://519728-90321822-gh.circle-artifacts.com/0/docs/prototype.datasets.html

torchaudio/prototype/datasets/bucketize_sampler.py
```diff
@@ -0,0 +1,141 @@
...
+        bucket_ids = torch.bucketize(torch.tensor(lengths), torch.tensor(boundaries))
+        for i in indices:
+            bucket_id = bucket_ids[self.index2iter[i]]
+            if bucket_id in buckets:
+                buckets[bucket_id].append(i)
+            else:
+                buckets[bucket_id] = [i]
+        for k in buckets:
+            buckets[k] = torch.as_tensor(buckets[k], dtype=torch.int)
+            if self.shuffle:
```
hwangjeff, line 94: as the caller, i might expect samples to be reshuffled whenever i request a new iterator, rather than shuffled just once during initialization. this behavior would also match that of the default samplers. what do you think?

torchaudio/prototype/datasets/bucketize_sampler.py
```diff
@@ -0,0 +1,141 @@
...
+                        total_len += 1
+
+        for batch in iter_list:
+            yield batch
+
+    def __len__(self):
+        if self.batch_size:
+            return len(self.lengths) // self.batch_size
+        else:
+            return sum(self.lengths) // self.max_token_count
```
hwangjeff, line 141: this doesn't hold in general

for example, if the underlying dataset comprises 10 samples of length 6 each, and `max_token_count` is 10, `__len__` would return 6, but the actual number of batches returned would be 10

torchaudio/prototype/datasets/bucketize_sampler.py
```diff
@@ -0,0 +1,141 @@
...
+            for k in self.buckets.keys():
+                for i in range(self.buckets[k].size(0)):
+                    index = self.buckets[k][i]
+                    if total_len == self.batch_size:
+                        iter_list.append(batch)
+                        batch = [index]
+                        total_len = 1
+                    else:
+                        batch.append(index)
+                        total_len += 1
```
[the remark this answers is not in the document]
hwangjeff: also, i think this might drop the last batch — should perhaps use the same logic as above to add the last partial batch to `iter_list` and update `__len__` to return the ceiling of `len(self.lengths) / self.batch_size`

torchaudio/prototype/datasets/bucketize_sampler.py
nateanl: Good point, I will move the shuffle in ``__iter__`` method.

torchaudio/prototype/datasets/bucketize_sampler.py
```diff
@@ -0,0 +1,141 @@
...
+        if self.max_token_count is not None:
+            for k in self.buckets.keys():
+                for i in range(self.buckets[k].size(0)):
+                    index = self.buckets[k][i]
+                    if len(batch) == 0 and self.lengths[self.index2iter[int(index)]] > self.max_token_count:
+                        raise ValueError(
+                            f"The length of the sample is {self.lengths[self.index2iter[int(index)]]}, "
+                            f"which is greater than ``max_token_count``: {self.max_token_count}."
+                            "Make sure you set the correct ``max_len`` and ``max_token_count``."
+                        )
```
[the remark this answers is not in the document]
nateanl: You mean detect the maximum of ``lengths`` is smaller than ``max_token_count`` when ``max_token_count`` is given? Makes sense, will correct it.

torchaudio/prototype/datasets/bucketize_sampler.py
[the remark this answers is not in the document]
nateanl: Got it. I will add ``drop_last`` argument and a note saying this argument is only valid when ``batch_size`` is given.

torchaudio/prototype/datasets/bucketize_sampler.py
nateanl: Yep, but I don't know how to estimate the correct number of batches, especially when ``shuffle`` is True.

hwangjeff: yeah i don't think it's possible. ideally this shouldn't return a numerical value to reflect that. what do callers expect?

torchaudio/prototype/datasets/bucketize_sampler.py
```diff
@@ -0,0 +1,141 @@
...
+            raise AssertionError("One of ``max_token_count`` or ``batch_size`` must be set.")
+        # Filter out samples which are outside the bounds of [min_len, max_len]
+        filtered_length_idx = [(length, i) for i, length in enumerate(lengths) if min_len <= length <= max_len]
+        if len(filtered_length_idx) == 0:
+            raise AssertionError("``lengths`` cannot be empty after filtering.")
+        # sort to minimize gap when bucketizing.
+        sorted_filtered_length_idx = sorted(filtered_length_idx, key=lambda x: x[0])
+        self.lengths = [e[0] for e in sorted_filtered_length_idx]
+        self.indices = [e[1] for e in sorted_filtered_length_idx]
+        self.index2iter = {e: i for i, e in enumerate(self.indices)}
```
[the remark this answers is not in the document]
hwangjeff: hmm i might be misunderstanding the purpose. why is sorting necessary?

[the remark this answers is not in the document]
nateanl: Sorting the samples by the lengths will minimize the paddings. Then for the samples in each bucket, we can shuffle them to introduce some randomness with marginal zero paddings, since they have similar lengths.

torchaudio/prototype/datasets/bucketize_sampler.py
nateanl: Ideally for each epoch they expect to iterate all the data once. So the lengths should reflect the number of batches in each iteration.
Is it possible to update the ``__len__`` of the sampler in the ``__iter__`` method?

torchaudio/prototype/datasets/bucketize_sampler.py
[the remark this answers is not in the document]
hwangjeff: yeah, i see the purpose in minimizing padding. but, downstream, you are using `bucketize` to assign bucket indices to the samples by length — sorting the lengths beforehand doesn't influence the assignments, and shuffling would undo the sorting within each bucket anyway

anyway, i see that you've moved shuffling into `__iter__`, so it isn't an issue now. re: the first point — i think it would be cleaner to remove `index2iter` and directly bucketize `lengths` so that you don't have to remap indices

torchaudio/prototype/datasets/bucketize_sampler.py
hwangjeff: hmm i'd rather not, since the length isn't guaranteed to be stable and we have no control over the order in which the user calls `__iter__` and `__len__`. `__len__` should simply be invalidated somehow in cases where we're batching by token count and shuffling

torchaudio/prototype/datasets/bucketize_sampler.py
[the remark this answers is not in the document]
nateanl: Oh I see. My assumption is sorting also guarantees the buckets are sorted, the if one bucket has been iterated and the batch is not filled, the samples from the next bucket will have similar length.
After second thought, this can be solved by sorting the keys in self.buckets

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -78,56 +89,59 @@ def _get_buckets(
...
+                        batch.append(self.indices[index])
                         total_len += self.lengths[index]
-        else:
-            for k in self.buckets.keys():
-                for i in range(self.buckets[k].size(0)):
-                    index = self.buckets[k][i]
+                    else:
+                        iter_list.append(batch)
+                        batch = []
+                        total_len = 0
```
hwangjeff, line 123: here, we'd be incorrectly skipping adding the current sample to a batch

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -78,56 +89,59 @@ def _get_buckets(
...
-            for k in self.buckets.keys():
-                for i in range(self.buckets[k].size(0)):
-                    index = self.buckets[k][i]
-                    if total_len > self.max_token_count:
-                        iter_list.append(batch)
-                        batch = [index]
-                        total_len = self.lengths[index]
-                    else:
-                        batch.append(index)
+        for k in sorted(self.buckets):
```
hwangjeff, line 113: we can sort the buckets just once in the initializer

examples/hubert/dataset/hubert_dataset.py
hwangjeff, line 123: i think we can simplify the logic and make it less error-prone by consolidating both batching cases, e.g., roughly,

```
max_batch_size = self.max_token_count if self.max_token_count else self.batch_size
[...]
sample_length = self.lengths[index] if self.max_token_count else 1
if total_len + sample_length <= max_batch_size:
    batch.append(self.indices[index])
    total_len += sample_length
else:
    iter_list.append(batch)
    batch = [self.indices[index]]
    total_len = sample_length
```

then, we wouldn't have to maintain separate code branches conditioned on `self.max_token_count`

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -78,56 +89,59 @@ def _get_buckets(
...
                     if total_len == self.batch_size:
                         iter_list.append(batch)
-                        batch = [index]
+                        batch = [self.indices[index]]
                         total_len = 1
                     else:
-                        batch.append(index)
+                        batch.append(self.indices[index])
                         total_len += 1
+            if len(batch) > 0 and (self.drop_last or self.max_token_count):
```
hwangjeff, line 132: `self.drop_last` being `True` should result in the last batch being dropped if it's partially filled and we're batching by sample count — i don't think this is captured properly here

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -78,56 +89,59 @@ def _get_buckets(
...
 
     def __len__(self):
-        return len(self.data_source)
+        if self.batch_size:
+            if self.drop_last:
+                return len(self.lengths) // self.batch_size
+            else:
+                return (len(self.lengths) + self.batch_size - 1) // self.batch_size
+        else:
+            return None
```
hwangjeff, line 144: is this a valid return value, i.e. dataloaders know how to handle it?

nateanl: In the new commit, the `self.iter_list` is initialized before `__iter__` to make sure the length is valid. Then every time the `iter_list` is updated, so will be the value returned by `__len__` method.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -48,26 +54,30 @@ def __init__(
...
+            ), "The  ``max_token_count`` must be greater than or equal to the maximum value of ``lengths``."
         # Filter out samples which are outside the bounds of [min_len, max_len]
-        # sort to minimize gap when bucketizing.
         filtered_length_idx = [(length, i) for i, length in enumerate(lengths) if min_len <= length <= max_len]
         if len(filtered_length_idx) == 0:
             raise AssertionError("``lengths`` cannot be empty after filtering.")
-        sorted_filtered_length_idx = sorted(filtered_length_idx, key=lambda x: x[0])
-        self.lengths = [e[0] for e in sorted_filtered_length_idx]
-        self.indices = [e[1] for e in sorted_filtered_length_idx]
+        self.lengths = [e[0] for e in filtered_length_idx]
```
nateanl, line 67: I feel like we still need to sort the lengths for one time. If `shuffle=False`, then we will generate the batch from the sorted index list, to minimize the padding.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -77,57 +88,48 @@ def _get_buckets(
...
+        if self.shuffle:
+            for k in self.buckets:
+                self.buckets[k] = self.buckets[k][torch.randperm(self.buckets[k].size(0))]
+            self._update_iter_list()
+
+        return iter(self.iter_list)
 
     def __len__(self):
-        return len(self.data_source)
+        return len(self.iter_list)
```
hwangjeff, line 132: i had commented before:

"i'd rather not, since the length isn't guaranteed to be stable and we have no control over the order in which the user calls `__iter__` and `__len__`"

for example, if a user calls `__len__` first and uses the returned length to guide how they access the iterator returned by `__iter__`, they're going to hit problems because the returned length will no longer correspond to the iterator

what would justify including this?

hwangjeff (approved):

mthrok (approved):

PR #2153, opened 2022-01-12
Use postponed evaluation to remove signature override

mthrok:
Python 3.7 introduced postponed evaluation of annotation. https://docs.python.org/3/whatsnew/3.7.html

Enabling this allows the code to use the exported name in annotation.
As a result we can directly annotate functions with exported name, and this way, Sphinx displays the output class in exported name without manually overriding the signature.

### DIFF
--- a/torchaudio/models/wav2vec2/model.py
+++ b/torchaudio/models/wav2vec2/model.py
@@ -1,6 +1,9 @@
+from __future__ import annotations
+
 from typing import Optional, Tuple, List
 
 import torch
+import torchaudio
 from torch import Tensor
 from torch.nn import Module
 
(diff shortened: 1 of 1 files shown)


torchaudio/models/wav2vec2/model.py
```diff
@@ -222,11 +226,8 @@ def wav2vec2_model(
     encoder_layer_norm_first: bool,
     encoder_layer_drop: float,
     aux_num_out: Optional[int],
-) -> Wav2Vec2Model:
-    # Overriding the signature so that the return type is correct on Sphinx
-    """wav2vec2_model(extractor_mode: str, extractor_conv_layer_config: Optional[List[Tuple[int, int, int]]], extractor_conv_bias: bool, encoder_embed_dim: int, encoder_projection_dropout: float, encoder_pos_conv_kernel: int, encoder_pos_conv_groups: int, encoder_num_layers: int, encoder_num_heads: int, encoder_attention_dropout: float, encoder_ff_interm_features: int, encoder_ff_interm_dropout: float, encoder_dropout: float, encoder_layer_norm_first: bool, encoder_layer_drop: float, aux_num_out: Optional[int]) -> torchaudio.models.Wav2Vec2Model
-
-    Build a custom Wav2Vec2Model
+) -> torchaudio.models.Wav2Vec2Model:
```
carolineechen, line 229: The [docs](https://555551-90321822-gh.circle-artifacts.com/0/docs/models.html#wav2vec2-model) still render this as `torchaudio.models.wav2vec2.model.Wav2Vec2Model`

nateanl (approved):
LGTM. Could you resolve the comment and merge conflict? Thanks.

PR #2156, opened 2022-01-14
allow Tacotron2 decoding batch_size 1 examples

popcornell:
it seems to me that the current Tacotron2 model does not allow for decoding batch size 1 examples: 
e.g. following code fails. I may have a fix for that. 

```python
if __name__ == "__main__":
    max_length = 400
    n_batch = 1
    hdim = 32
    dec = _Decoder(
        encoder_embedding_dim=hdim,
        n_mels = hdim,
        n_frames_per_step = 1,
        decoder_rnn_dim = 1024,
        decoder_max_step = 2000,
        decoder_dropout = 0.1,
        decoder_early_stopping = True,
        attention_rnn_dim = 1024,
        attention_hidden_dim = 128,
        attention_location_n_filter = 32,
        attention_location_kernel_size = 31,
        attention_dropout = 0.1,
        prenet_dim = 256,
        gate_threshold = 0.5)

    inp = torch.rand((n_batch, max_length, hdim))
    lengths = torch.tensor([max_length]).expand(n_batch).to(inp.device, inp.dtype)
    dec(inp, torch.rand((n_batch, hdim, max_length)), lengths)[0]
    dec.infer(inp, lengths)[0]
```


nateanl:
cc @yangarbiter @mthrok Could you review the fix? Thanks!

mthrok:
Hi @popcornell 

What is the equivalent of the snippet you posted, in terms of the public API (Tacotron2 method)?

As far as I know, the inference (which uses `Tacotron2.infer` instead of `Tacotron2.forward`) works on batch size = 1. https://pytorch.org/audio/main/tutorials/tacotron2_pipeline_tutorial.html

Does the issue happen when using forward method?

popcornell:
This also fails for me, without the PR fix. 
Inference is fine. So not a big deal really because nobody will train with batch size 1 (but may use that for some quick testing). 

```python
max_length_tokens = 20
n_batch = 1
max_length_mel = 100


taco = Tacotron2()
inp = torch.rand((n_batch, max_length_tokens)).long()
lengths = torch.tensor([max_length_tokens]).expand(n_batch).to(inp.device, inp.dtype)

mel_spec = torch.rand((n_batch, 80, max_length_mel))
mel_spec_lens = torch.tensor([max_length_mel]).expand(n_batch).to(inp.device, inp.dtype)
taco.forward(inp, lengths, mel_spec, mel_spec_lens)
```

nateanl:
Checked the original implementation from [Nvidia](https://github.com/NVIDIA/tacotron2/blob/master/model.py#L410), this bug fix LGTM. Thanks for detecting it and work on the fix!

nateanl (approved):

PR #2160, opened 2022-01-18
Update PR labeling workflow

carolineechen:
update the labeling reminder to be triggered when PRs are closed rather than merged, because of the transition of merging through fbcode

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -1,9 +1,9 @@
 name: pr-labels
 
 on:
-  push:
-    branches:
-      - main
+  pull_request:
+    types:
+      - closed
 
 jobs:
   is-properly-labeled:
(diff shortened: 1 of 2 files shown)


nateanl (approved):
Thanks for working on it.

.github/workflows/pr-labels.yml
```diff
@@ -1,9 +1,9 @@
 name: pr-labels
 
 on:
-  push:
-    branches:
-      - main
+  pull_request:
+    types:
+      - closed
```
mthrok, line 6: What would happen if I open a test PR that is not meant to be merged and just close it?
Should we check if the PR was closed by the open source bot?

carolineechen: if the PR is closed but not merged, the commit will not correspond to a PR number (which is obtained by parsing the merged commit message) and should not output a merger or generate any response

mthrok: Okay, sounds good.

mthrok (approved):

PR #2164, opened 2022-01-19
Add Streaming API

mthrok:
This PR adds the prototype streaming API.
The implementation is based on ffmpeg libraries.

For the detailed usage, please refer to [the resulting tutorial](https://534376-90321822-gh.circle-artifacts.com/0/docs/tutorials/streaming_api_tutorial.html).

### DIFF
--- a/test/torchaudio_unittest/prototype/io_test.py
+++ b/test/torchaudio_unittest/prototype/io_test.py
@@ -1,0 +1,416 @@
+import torch
+from parameterized import parameterized
+from torchaudio_unittest.common_utils import (
+    TorchaudioTestCase,
+    TempDirMixin,
+    get_asset_path,
+    get_wav_data,
+    save_wav,
+    skipIfNoFFmpeg,
+    nested_params,
+    is_ffmpeg_available,
+    get_image,
+    save_image,
+)
+
+if is_ffmpeg_available():
+    from torchaudio.prototype.io import (
+        Streamer,
+        SourceStream,
+        SourceVideoStream,
+        SourceAudioStream,
+    )
+
+
+def get_video_asset(file="nasa_13013.mp4"):
+    return get_asset_path(file)
+
+
+@skipIfNoFFmpeg
+class StreamerInterfaceTest(TempDirMixin, TorchaudioTestCase):
+    """Test suite for interface behaviors around Streamer"""
+
+    def test_streamer_invalid_input(self):
+        """Streamer constructor does not segfault but raise an exception when the input is invalid"""
+        with self.assertRaises(RuntimeError):
+            Streamer("foobar")
+
+    def test_src_info(self):
+        """`get_src_stream_info` properly fetches information"""
+        s = Streamer(get_video_asset())
+        assert s.num_src_streams == 6
+
+        expected = [
+            SourceVideoStream(
+                media_type="video",
+                codec="h264",
+                codec_long_name="H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10",
+                format="yuv420p",
+                bit_rate=71925,
+                width=320,
+                height=180,
+                frame_rate=25.0,
+            ),
+            SourceAudioStream(
+                media_type="audio",
+                codec="aac",
+                codec_long_name="AAC (Advanced Audio Coding)",
+                format="fltp",
+                bit_rate=72093,
+                sample_rate=8000.0,
+                num_channels=2,
+            ),
+            SourceStream(
+                media_type="subtitle",
+                codec="mov_text",
+                codec_long_name="MOV text",
+                format=None,
+                bit_rate=None,
+            ),
+            SourceVideoStream(
+                media_type="video",
+                codec="h264",
+                codec_long_name="H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10",
+                format="yuv420p",
+                bit_rate=128783,
+                width=480,
+                height=270,
+                frame_rate=29.97002997002997,
+            ),
+            SourceAudioStream(
+                media_type="audio",
+                codec="aac",
+                codec_long_name="AAC (Advanced Audio Coding)",
+                format="fltp",
+                bit_rate=128837,
+                sample_rate=16000.0,
+                num_channels=2,
+            ),
+            SourceStream(
+                media_type="subtitle",
+                codec="mov_text",
+                codec_long_name="MOV text",
+                format=None,
+                bit_rate=None,
+            ),
+        ]
+        for i, exp in enumerate(expected):
+            assert exp == s.get_src_stream_info(i)
+
+    def test_src_info_invalid_index(self):
+        """`get_src_stream_info` does not segfault but raise an exception when input is invalid"""
+        s = Streamer(get_video_asset())
+        for i in [-1, 6, 7, 8]:
+            with self.assertRaises(IndexError):
+                s.get_src_stream_info(i)
+
+    def test_default_streams(self):
+        """default stream is not None"""
+        s = Streamer(get_video_asset())
+        assert s.default_audio_stream is not None
+        assert s.default_video_stream is not None
+
+    def test_default_audio_stream_none(self):
+        """default audio stream is None for video without audio"""
+        s = Streamer(get_video_asset("nasa_13013_no_audio.mp4"))
+        assert s.default_audio_stream is None
+
+    def test_default_video_stream_none(self):
+        """default video stream is None for video with only audio"""
+        s = Streamer(get_video_asset("nasa_13013_no_video.mp4"))
+        assert s.default_video_stream is None
+
+    def test_num_out_stream(self):
+        """num_out_streams gives the correct count of output streams"""
+        s = Streamer(get_video_asset())
+        n, m = 6, 4
+        for i in range(n):
+            assert s.num_out_streams == i
+            s.add_audio_stream(frames_per_chunk=-1)
+        for i in range(m):
+            assert s.num_out_streams == n - i
+            s.remove_stream(0)
+        for i in range(m):
+            assert s.num_out_streams == n - m + i
+            s.add_video_stream(frames_per_chunk=-1)
+        for i in range(n):
+            assert s.num_out_streams == n - i
+            s.remove_stream(n - i - 1)
+        assert s.num_out_streams == 0
+
+    def test_basic_audio_stream(self):
+        """`add_basic_audio_stream` constructs a correct filter."""
+        s = Streamer(get_video_asset())
+        s.add_basic_audio_stream(frames_per_chunk=-1, dtype=None)
+        s.add_basic_audio_stream(frames_per_chunk=-1, sample_rate=8000)
+        s.add_basic_audio_stream(frames_per_chunk=-1, dtype=torch.int16)
+
+        sinfo = s.get_out_stream_info(0)
+        assert sinfo.source_index == s.default_audio_stream
+        assert sinfo.filter_description == ""
+
+        sinfo = s.get_out_stream_info(1)
+        assert sinfo.source_index == s.default_audio_stream
+        assert "aresample=8000" in sinfo.filter_description
+
+        sinfo = s.get_out_stream_info(2)
+        assert sinfo.source_index == s.default_audio_stream
+        assert "aformat=sample_fmts=s16" in sinfo.filter_description
+
+    def test_basic_video_stream(self):
+        """`add_basic_video_stream` constructs a correct filter."""
+        s = Streamer(get_video_asset())
+        s.add_basic_video_stream(frames_per_chunk=-1, format=None)
+        s.add_basic_video_stream(frames_per_chunk=-1, width=3, height=5)
+        s.add_basic_video_stream(frames_per_chunk=-1, frame_rate=7)
+        s.add_basic_video_stream(frames_per_chunk=-1, format="BGR")
+
+        sinfo = s.get_out_stream_info(0)
+        assert sinfo.source_index == s.default_video_stream
+        assert sinfo.filter_description == ""
+
+        sinfo = s.get_out_stream_info(1)
+        assert sinfo.source_index == s.default_video_stream
+        assert "scale=width=3:height=5" in sinfo.filter_description
+
+        sinfo = s.get_out_stream_info(2)
+        assert sinfo.source_index == s.default_video_stream
+        assert "fps=7" in sinfo.filter_description
+
+        sinfo = s.get_out_stream_info(3)
+        assert sinfo.source_index == s.default_video_stream
+        assert "format=pix_fmts=bgr24" in sinfo.filter_description
+
+    def test_remove_streams(self):
+        """`remove_stream` removes the correct output stream"""
+        s = Streamer(get_video_asset())
+        s.add_basic_audio_stream(frames_per_chunk=-1, sample_rate=24000)
+        s.add_basic_video_stream(frames_per_chunk=-1, width=16, height=16)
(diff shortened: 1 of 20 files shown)


torchaudio/csrc/ffmpeg/prototype.cpp
```diff
@@ -99,7 +99,7 @@ int64_t find_best_video_stream(S s) {
   return s->s.find_best_video_stream();
 }
 
-void seek(S s, int64_t timestamp) {
```
mthrok, line 102: bug fix

torchaudio/csrc/ffmpeg/typedefs.h
```diff
@@ -8,9 +8,9 @@ namespace ffmpeg {
 
 struct SrcStreamInfo {
   AVMediaType media_type;
-  const char* codec_name = NULL;
```
mthrok, line 11: bug fix

keeping this NULL caused segmentation fault if a source stream is subtitle.

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+class Streamer:
+    """Fetch and decode audio/video streams chunk by chunk.
+
+    For the detailed usage of this class, please refer to the tutorial.
+
+    Args:
+        src (str): Source. Can be a file path, URL, device identifier or filter expression.
+        format (str, optional):
```
carolineechen, line 141: same for remaining Optional[xxx] typed arguments in this file
```suggestion
        format (str or None, optional):
```

test/torchaudio_unittest/prototype/io_test.py
```diff
@@ -0,0 +1,348 @@
...
+                codec_long_name="MOV text",
+                format=None,
+                bit_rate=None,
+            ),
+        ]
+        for i, exp in enumerate(expected):
+            assert exp == s.get_src_stream_info(i)
+
+    def test_src_info_invalid_index(self):
+        """`get_src_stream_info` should not crash when invalid values are passed"""
```
carolineechen, line 101: should this be should crash or should not crash?

test/torchaudio_unittest/prototype/io_test.py
```diff
@@ -0,0 +1,348 @@
...
+        for i in range(m):
+            assert s.num_out_streams == n - m + i
+            s.add_video_stream(frames_per_chunk=-1)
+        for i in range(n):
+            assert s.num_out_streams == n - i
+            s.remove_stream(n - i - 1)
+        assert s.num_out_streams == 0
+
+    def test_basic_audio_stream(self):
+        """`add_basic_audio_stream` constrcuts and a correct filter."""
```
carolineechen, line 142:
```suggestion
        """`add_basic_audio_stream` constructs a correct filter."""
```

test/torchaudio_unittest/prototype/io_test.py
```diff
@@ -0,0 +1,348 @@
...
+        sinfo = s.get_out_stream_info(1)
+        assert sinfo.source_index == s.default_audio_stream
+        assert "aresample=8000" in sinfo.filter_description
+
+        sinfo = s.get_out_stream_info(2)
+        assert sinfo.source_index == s.default_audio_stream
+        assert "aformat=sample_fmts=s16" in sinfo.filter_description
+
+    def test_basic_video_stream(self):
+        """`add_basic_video_stream` constrcuts and a correct filter."""
```
carolineechen, line 161:
```suggestion
        """`add_basic_video_stream` constructs a correct filter."""
```

test/torchaudio_unittest/prototype/io_test.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def _test_wav(self, path, original, dtype):
+        s = Streamer(path)
+        s.add_basic_audio_stream(frames_per_chunk=-1, dtype=dtype)
+        s.process_all_packets()
+        (output,) = s.pop_chunks()
+        self.assertEqual(original, output)
+
+    @nested_params(
+        ["int16", "uint8", "int32"],  # "float", "double", "int64"]
```
carolineechen, line 237: why are the additional types commented out?

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+        )
+
+    def add_audio_stream(
+        self,
+        frames_per_chunk: int,
+        buffer_chunk_size: int = 3,
+        stream_index: Optional[int] = None,
+        filter_desc: Optional[str] = None,
+    ):
+        """Add output video stream
```
nateanl, line 332:
```suggestion
        """Add output audio stream
```

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+            frames_per_chunk (int): A unit number of frames returned by Streamer as a chunk.
+                If the source stream is exhausted before enough frames are buffered,
+                then the chunk is returned as-is.
+
+            buffer_chunk_size (int): Internal buffer size.
+                When this many chunks are created, but
+                client code does not pop the chunk, if a new frame comes in, the old
+                chunk is dropped.
+
+            stream_index (int, optional): The source audio stream index.
```
nateanl, line 344:
```suggestion
            stream_index (int or None, optional): The source audio stream index.
```

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+        filter_desc: Optional[str] = None,
+    ):
+        """Add output video stream
+
+        Args:
+            frames_per_chunk (int): A unit number of frames returned by Streamer as a chunk.
+                If the source stream is exhausted before enough frames are buffered,
+                then the chunk is returned as-is.
+
+            buffer_chunk_size (int): Internal buffer size.
```
nateanl, line 339:
```suggestion
            buffer_chunk_size (int, optional): Internal buffer size.
```

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
```
nateanl, line 371: Same for the following three arguments.

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+        an output stream, then the data are discarded.
+
+        When the source reaches EOF, then it triggers all the output stream
+        processors to ender drain mode. All the output stream processors
+        flush the pending frames.
+
+        Returns:
+            int:
+                ``0``
+                A packet is processed properly. The caller can keep
```
nateanl, line 413: To make it consistent with ``1`` and ``<0``
```suggestion
                A packet was processed properly. The caller can keep
```

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+                flushed the pending frames. The caller should stop calling
+                this method.
+
+                ``<0``
+                An error has occured.
+        """
+        return torch.ops.torchaudio.ffmpeg_streamer_process_packet(self._s)
+
+    def process_all_packets(self) -> int:
+        """Keep processing the source until it reaches EOF or an error happens.
```
nateanl, line 427:
```suggestion
        """Keep processing the source until it reached EOF or an error happened.
```

test/torchaudio_unittest/prototype/io_test.py
mthrok: I meant a crash in C++ layer, which is not recoverable from Python layer. Let me update the description.

test/torchaudio_unittest/prototype/io_test.py
mthrok: The code supports these, but test fixture (save_wav which is backed by scipy) does not support it so we cannot test.

test/torchaudio_unittest/prototype/io_test.py
```diff
@@ -0,0 +1,348 @@
...
+def get_video_asset(file="nasa_13013.mp4"):
+    return get_asset_path(file)
+
+
+@skipIfNoFFmpeg
+class StreamerInterfaceTest(TempDirMixin, TorchaudioTestCase):
+    """Test suite for interface behaviors around Streamer"""
+
+    def test_streamer_invalid_input(self):
+        """Streamer constructor does not crash when receiving an invalid input"""
```
mthrok, line 34:
```suggestion
        """Streamer constructor does not segfault but raise an exception when receiving an invalid input"""
```

test/torchaudio_unittest/prototype/io_test.py
mthrok, line 101:
```suggestion
        """`get_src_stream_info` does not segfault but raise an exception when input is invalid"""
```

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+        self,
+        frames_per_chunk: int,
+        buffer_chunk_size: int = 3,
+        stream_index: Optional[int] = None,
+        filter_desc: Optional[str] = None,
+    ):
+        """Add output audio stream
+
+        Args:
+            frames_per_chunk (int): A unit number of frames returned by Streamer as a chunk.
```
hwangjeff, line 335: maybe just "Number of frames [...] per chunk" instead of "A unit number of frames [...]"?

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+                yield self.pop_chunks()
+            elif code == 1:
+                while True:
+                    chunks = self.pop_chunks()
+                    if all(c is None for c in chunks):
+                        return
+                    yield chunks
+                return
+            else:
+                return
```
hwangjeff, line 486: should the user be alerted that an unexpected error has occurred?

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+                If omitted, :py:attr:`default_audio_stream` is used.
+
+            sample_rate (int or None, optional): If provided, resample the audio.
+
+            dtype (torch.dtype, optional): If not ``None``, change the output sample precision.
+                If floating point, then the sample value range is
+                `[-1, 1]`.
+        """
+        i = self.default_audio_stream if stream_index is None else stream_index
+        torch.ops.torchaudio.ffmpeg_streamer_add_basic_audio_stream(
```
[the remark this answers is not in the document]
mthrok: > if one output stream advances more quickly than the others, e.g. its `frames_per_chunk`-to-`sample_rate` ratio is higher, are the other output streams guaranteed to be not affected?

Under the circumstance the value of `buffer_chunk_size` is large enough so that no chunk is dropped during the streaming, then the other output streams are not affected.

The media source with multiple streams, like mp4 is composed of the sequence of containers corresponding to the different streams. If one output stream is configured to have higher `frames_per_chunk / sample_rate` value, then when the `Streamer` attempts to fill all output stream buffers, even after the other output streams have enough frames to consist the first chunk, this output stream with higher `frames_per_chunk / sample_rate` value can be still missing frames for the first chunk. Then the `Streamer` has to parse more frames to fill the buffer of this stream. When the `Streamer` parse more frames, inevitably it will also parse the frames of the other streams as well.

Untitled drawing

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+                chunk is dropped.
+
+            stream_index (int or None, optional): The source video stream index.
+                If omitted, :py:attr:`default_video_stream` is used.
+
+            frame_rate (int or None, optional): If provided, change the frame rate.
+
+            width (int or None, optional): If provided, change the image width. Unit: Pixel.
+            height (int or None, optional): If provided, change the image height. Unit: Pixel.
+            format (str, optional): Change the format of image channels. Valid values are,
```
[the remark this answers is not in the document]
mthrok: `None` for these parameters are just placeholder for "missing value that is to be fallback to some (dynamically determined) default value". So `None` is not what we expect / advertise users to pass. 

Consistency is important but I think saying `or None` is more confusing because the code expect that users to pass some meaningful value other than `None` if users are to provide. What do you think? @hwangjeff  @carolineechen

torchaudio/prototype/io/streamer.py
mthrok: Good point and definitely we need to iterate on this but I am not sure what is the best way to communicate the error.

The code that streams [the tutorial MP4](https://download.pytorch.org/torchaudio/tutorial-assets/stream-api/NASAs_Most_Scientifically_Complex_Space_Observatory_Requires_Precision-MP4.mp4) was working fine in NY but now I am facing some issue in Tokyo, maybe because of network (bandwidth?) issue.

```
[tls @ 0x144e041d0] Error in the pull function.
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x147808200] stream 0, offset 0x46f5ac3: partial file
[tls @ 0x1126ccfd0] Error in the pull function.
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x163012000] stream 0, offset 0x59fc887: partial file
[tls @ 0x11262c8c0] Error in the pull function.
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x163012000] stream 0, offset 0x6e951de: partial file
[tls @ 0x12277e830] Error in the pull function.
[mov,mp4,m4a,3gp,3g2,mj2 @ 0x166008600] stream 0, offset 0x37f037f: partial file
[tls @ 0x1126f0ef0] Error in the pull function.
[aac @ 0x16300e200] decode_band_types: Input buffer exhausted before END element found
```

In C++ code, the streaming part does not throw an exception for the sake of performance and recoverability, maybe we should raise an exception on Python layer.

torchaudio/prototype/io/streamer.py
```diff
@@ -0,0 +1,486 @@
...
+    """Width of the video frame in pixel."""
+    height: int
+    """Height of the video frame in pixel."""
+    frame_rate: float
+    """Frame rate."""
+
+
+def _parse_si(i):
+    # - COMMON
+    # media_type: str
```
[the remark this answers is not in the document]
mthrok: Not sure how to apply enum. Do you mean something like naming an index, like `codec_name = 0; info[codec_name]`?

torchaudio/prototype/io/streamer.py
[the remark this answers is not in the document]
mthrok: ah okay, I misunderstood what you meant.

torchaudio/prototype/io/streamer.py
mthrok: @hwangjeff Okay, I changed the semantic on Python layer. Errors that happen during the actual streaming in C++ is notified by return value. The new code convert this into exception so that Python API is more Pythonic.

torchaudio/prototype/io/streamer.py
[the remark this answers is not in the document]
hwangjeff: ah, one more nit: it seems like we have a convention of adding default values for kwargs in docstrings (this applies throughout the module)

torchaudio/prototype/io/streamer.py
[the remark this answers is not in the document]
hwangjeff: i see. i'm not very familiar with ffmpeg — would most users generally know that chunks would be dropped if output streams are configured differently? wondering if it'd make sense to, say, surface a warning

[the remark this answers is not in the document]
mthrok: Okay, added warnings [here](https://github.com/pytorch/audio/pull/2164/commits/5f531fa71d35e85ac8d37057d28e340b68591222).

hwangjeff (approved):

PR #2170, opened 2022-01-22
Add `is_ffmpeg_available` in test

mthrok:
Part of #2164.
To make the tests introduced in #2164 skippable if ffmpeg features are not available,
this commit adds `is_ffmpeg_available`.

The availability of the features depend on two factors;
1. If it was enabled at build.
2. If the ffmpeg libraries are found at runtime.

A simple way (for OSS workflow) to detect these is simply checking if
`libtorchaudio_ffmpeg` presents and can be loaded without a failure.

To facilitate this, this commit changes the
`torchaudio._extension._load_lib` to return boolean result.

### DIFF
--- a/torchaudio/_extension.py
+++ b/torchaudio/_extension.py
@@ -37,9 +37,13 @@ def _load_lib(lib: str) -> bool:
 
     Raises:
         Exception:
-            Any exception thrown when loading the library.
+            If the library file is found, but there is an issue loading the library file,
+            (when underlying `ctype.DLL` throws an exception), this function will pass
+            the exception as-is, instead of catching it and returning bool.
             The expected case is `OSError` thrown by `ctype.DLL` when a dynamic dependency
             is not found.
+            This behavior was chosen because the expected failure case is not recoverable.
+            If a dependency is missing, then users have to install it.
     """
     path = _get_lib_path(lib)
     if not path.exists():
--- a/torchaudio/_extension.py
+++ b/torchaudio/_extension.py
@@ -14,15 +14,34 @@ def _get_lib_path(lib: str):
     return path
 
 
-def _load_lib(lib: str):
+def _load_lib(lib: str) -> bool:
+    """Load extension module
+
+    Note:
+        In case `torchaudio` is deployed with `pex` format, the library file does not
+        exist as a stand alone file.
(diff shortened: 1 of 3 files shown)


torchaudio/_extension.py
```diff
@@ -14,15 +14,34 @@ def _get_lib_path(lib: str):
...
+def _load_lib(lib: str) -> bool:
+    """Load extension module
+
+    Note:
+        In case `torchaudio` is deployed with `pex` format, the library file does not
+        exist as a stand alone file.
+        In this case, we expect that `libtorchaudio` is available somewhere
+        in the search path of dynamic loading mechanism, so that importing
+        `_torchaudio` will have library loader find and load `libtorchaudio`.
+        This is the reason why the function should not raising an error when the library
```
carolineechen, line 26:
```suggestion
        This is the reason why the function should not raise an error when the library
```

torchaudio/_extension.py
```diff
@@ -14,15 +14,34 @@ def _get_lib_path(lib: str):
...
+        file is not found.
+
+    Returns:
+        bool:
+            False if the library file is not found.
+            True if the library file is found AND the library loaded without failure.
```
carolineechen, line 32: To account for the scenario in which the library file is found but the library fails to load correctly:

```suggestion            
            True if the library file is found AND the library loaded without failure.
            False otherwise.
```

carolineechen (approved):

torchaudio/_extension.py
mthrok: Hmm, the suggested change might give an impression that the function still returns when the library is found but not loaded. I would like to make it explicit specification that this function has three branches, 1. the target library is success loaded, 2. the target library was not found (a condition happens in anticipated use case) 3. an issue occurred when loading the library (an unexpected condition, thus we won't try to recover to the normal execution path.)

carolineechen: got it, so the raises Exception case covers case 3 right, when the library file is found but there's a failure when loading. can you mention that in the raises exception section just to make it a little bit more clear?

mthrok: How does it sound now?

carolineechen (approved):

PR #2171, opened 2022-01-25
Remove subsampling and positional embedding logic from Conformer

hwangjeff:
To facilitate experimenting with different strategies, this PR removes the existing subsampling and positional embedding logic from `Conformer`.

### DIFF
--- a/test/torchaudio_unittest/prototype/conformer_test_impl.py
+++ b/test/torchaudio_unittest/prototype/conformer_test_impl.py
@@ -9,10 +9,6 @@ class ConformerTestImpl(TestBaseMixin):
             Conformer(
                 num_layers=4,
                 input_dim=80,
-                conv_channels=64,
-                conformer_layer_input_dim=256,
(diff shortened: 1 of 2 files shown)


mthrok (approved):
Are the removed components added back as independent components?

hwangjeff:
> Are the removed components added back as independent components?

no — there hasn't been a demonstrated need for them, and i haven't tested them in training yet (for my conformer training experiments thus far, i've been doing conventional time reduction and have omitted the sinusoidal embeddings). figured i can add them back in later on after i verify their efficacy

leng-yue:
> > Are the removed components added back as independent components?
> 
> no — there hasn't been a demonstrated need for them, and i haven't tested them in training yet (for my conformer training experiments thus far, i've been doing conventional time reduction and have omitted the sinusoidal embeddings). figured i can add them back in later on after i verify their efficacy

Does that means we don't actually need the relative positional embedding in conformer?

PR #2172, opened 2022-01-25
Add integration test for Emformer RNN-T LibriSpeech pipeline

hwangjeff:
Adds integration test for pretrained ASR pipeline `EMFORMER_RNNT_BASE_LIBRISPEECH`.

### DIFF
--- a/test/integration_tests/rnnt_pipeline_test.py
+++ b/test/integration_tests/rnnt_pipeline_test.py
@@ -4,7 +4,8 @@ from torchaudio.prototype.pipelines import EMFORMER_RNNT_BASE_LIBRISPEECH
 
 
 @pytest.mark.parametrize(
-    "bundle,lang,expected", [(EMFORMER_RNNT_BASE_LIBRISPEECH, "en", "i have that curiosity beside me at this moment")],
+    "bundle,lang,expected",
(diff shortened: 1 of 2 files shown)


test/integration_tests/rnnt_pipeline_test.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@pytest.mark.parametrize(
+    "bundle,lang,expected",
+    [
+        (
+            EMFORMER_RNNT_BASE_LIBRISPEECH,
+            "en",
+            # SentencePiece token ids for "i have that curiosity beside me at this moment"
+            [4096, 18, 135, 60, 3526, 1931, 113, 100, 138, 619],
```
mthrok, line 13: Why not test on the final output from sentence piece?

hwangjeff: i chose to test on just the raw ids because we haven't made `sentencepiece` an explicit dependency

carolineechen: I don't have a strong preference between the two, but one option could to have `sentencepiece` be an optional testing dependency like the ones listed [here](https://github.com/pytorch/audio/blob/main/CONTRIBUTING.md#install-torchaudio)

carolineechen (approved):

nateanl (approved):

PR #2177, opened 2022-01-26
Add RNNTBundle with weights pre-trained on tedlium3 dataset

nateanl:


nateanl:
The model achieves 11.4% WER on dev set and 10.8% WER on the test set of TED-LIUM release3. The training recipe will be added in a separate PR.

mthrok:
> The model achieves 11.4% WER on dev set and 10.8% WER on the test set of TED-LIUM release3. The training recipe will be added in a separate PR.

Do you mean to update the training script? The docsrting says it is trained using the script in example directory.
As a follow up, can you add an integration tests like in #2172?

torchaudio/prototype/pipelines/rnnt_pipeline.py
```diff
@@ -384,3 +384,27 @@ def get_token_processor(self) -> TokenProcessor:
 
     Please refer to :py:class:`RNNTBundle` for usage instructions.
     """
+
+
+EMFORMER_RNNT_BASE_TEDLIUM3 = RNNTBundle(
+    _rnnt_path="emformer_rnnt_base_tedlium3.pt",
+    _rnnt_factory_func=emformer_rnnt_base,
+    _global_stats_path="global_stats_rnnt_tedlium3.json",
+    _sp_model_path="spm_bpe_500_tedlium3.model",
```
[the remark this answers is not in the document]
mthrok: Does it make sense to add `num_symbols` parameter to `emformer_rnnt_base`? I think, even though the `base` is a sort of specification for model architecture, to handle different languages, it is common to change the number of final output tokens.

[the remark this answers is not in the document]
nateanl: I see, the `emformer_rnnt_base` is designed with 4096 symbols. In this case, shall we refactor the current ``emformer_rnnt_base`` with `num_symbols=4096`` as input argument, or a new factory function, like ``emformer_rnnt_tedlium``?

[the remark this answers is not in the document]
hwangjeff: yeah, that seems reasonable

nateanl:
> Do you mean to update the training script? The docsrting says it is trained using the script in example directory. As a follow up, can you add an integration tests like in #2172?

Yeah that's what I want to discuss, shall we create a new recipe in ``tedlium3_emformer_rnnt`` directory, or update the current recipe as there are many overlapped scripts can be reused?

nateanl:
Update: the pipeline achieves 10.8% WER on the dev set and 9.8% WER on the test set.

nateanl:
Docs is available here: https://538871-90321822-gh.circle-artifacts.com/0/docs/pipelines.html#emformer-rnnt-base-tedlium3
Will add the training recipe for `tedlium3_emformer_rnnt` in a separate PR and update the link reference in the doc then.

torchaudio/pipelines/rnnt_pipeline.py
```diff
@@ -385,3 +385,27 @@ def get_token_processor(self) -> TokenProcessor:
...
+    _n_mels=80,
+    _hop_length=160,
+    _segment_length=16,
+    _right_context_length=4,
+)
+EMFORMER_RNNT_BASE_TEDLIUM3.__doc__ = """Pre-trained Emformer-RNNT-based ASR pipeline capable of performing both streaming and non-streaming inference.
+
+    The underlying model is constructed by :py:func:`torchaudio.prototype.models.emformer_rnnt_base`
+    and utilizes weights trained on TED-LIUM Release 3 dataset using training script ``train.py``
+    `here <https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_emformer_rnnt>`__ with default arguments.
```
mthrok, line 408: Is `with default arguments` accurate?

mthrok:
> > Do you mean to update the training script? The docsrting says it is trained using the script in example directory. As a follow up, can you add an integration tests like in #2172?
> 
> Yeah that's what I want to discuss, shall we create a new recipe in `tedlium3_emformer_rnnt` directory, or update the current recipe as there are many overlapped scripts can be reused?

I think it should go to the same directory and we can remove the dataset name from the directory name.

torchaudio/pipelines/rnnt_pipeline.py
hwangjeff, line 408: do you want to prepare and merge the training recipe first? then we can update the docstring here accordingly

hwangjeff:
also, should this be first merged as a prototype feature rather than to `torchaudio.pipelines`? @mthrok what do you think?

torchaudio/pipelines/rnnt_pipeline.py
nateanl: Oops, will change that. Thanks.

nateanl:
>I think it should go to the same directory and we can remove the dataset name from the directory name.

Actually there are some logic changes that are different from librispeech pipeline. For example, the way to ignore the audios that have `ignore_time_segment_in_scoring` as the transcription label, dealing with the ``<unk>`` token in sentencepiece model training and mapping it to real `unk` index in the rnnt training. I feel it makes more clear to put all the training scripts into a separate folder.

torchaudio/pipelines/rnnt_pipeline.py
nateanl: Sure, will prepare it today.

hwangjeff (commented):
it'd be helpful to include a streaming asr demo as well, similar to https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py. this can be done in a separate pr

that aside, i'd like to get confirmation from the others on whether it's appropriate to directly land this in `torchaudio.pipelines` versus `torchaudio.prototype.pipelines`. what's your reasoning?

torchaudio/prototype/pipelines/rnnt_pipeline.py
```diff
@@ -0,0 +1,28 @@
...
+    _segment_length=16,
+    _right_context_length=4,
+)
+EMFORMER_RNNT_BASE_TEDLIUM3.__doc__ = """Pre-trained Emformer-RNNT-based ASR pipeline capable of performing both streaming and non-streaming inference.
+
+    The underlying model is constructed by :py:func:`torchaudio.models.emformer_rnnt_base`
+    and utilizes weights trained on TED-LIUM Release 3 dataset using training script ``train.py``
+    `here <https://github.com/pytorch/audio/tree/main/examples/asr/tedlium3_emformer_rnnt>`__ with ``num_symbols=501``.
+
+    Please refer to :py:class:`RNNTBundle` for usage instructions.
```
hwangjeff, line 27: can you update this reference so that it links to `torchaudio.pipelines.RNNTBundle`? e.g. as in https://pytorch.org/audio/main/pipelines.html#emformer-rnnt-base-librispeech

hwangjeff (approved):
lg

mthrok:
> it'd be helpful to include a streaming asr demo as well, similar to https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py. this can be done in a separate pr
> 
> that aside, i'd like to get confirmation from the others on whether it's appropriate to directly land this in `torchaudio.pipelines` versus `torchaudio.prototype.pipelines`. what's your reasoning?

@nateanl @hwangjeff though I need to try it, first, I think it’s okay to move the pretrained weights out of prototype. The result looks reasonable and the API / backward compatibility should not be an issue. (It will have the same maintenance cost as LibriSpeech bundle only if we change the interface. By design, bundles allow the model architecture in library to evolve while the model architecture used in bundle to be fixed.

mthrok:
Oh but we need an integration test for bundles like in wav2vec2

PR #2181, opened 2022-01-27
Add psd and mvdr methods to functional & Refactor PSD and MVDR module in transforms

nateanl:
- Provide following methods in `torchaudio.functional`:
 `compute_power_spectral_density_matrix`
 `compute_mvdr_weights_souden`
 `compute_mvdr_weights_rtf`
 `compute_rtf_evd`
 `compute_rtf_power`
 `apply_beamforming`
- Add doc strings for the methods above.
- Refactor the current `torchaudio.transforms.PSD` and `torchaudio.transforms.MVDR` by using the added methods.


nateanl:
@Emrys365 @popcornell @boeddeker Would you like to review the algorithms in the methods? Thanks!

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+    """
+    w, v = torch.linalg.eig(psd_s)  # (..., freq, channel, channel)
+    _, indices = torch.max(w.abs(), dim=-1, keepdim=True)
+    indices = indices.unsqueeze(-1)
+    rtf = v.gather(-1, indices.expand(psd_s.shape[:-1] + (1,)))  # (..., freq, channel, 1)
+    return rtf
+
+
+def compute_rtf_power(psd_s: Tensor, psd_n: Tensor, reference_channel: Union[int, Tensor]) -> Tensor:
+    r"""Estimate the relative transfer function (RTF) or the steering vector by the power method.
```
boeddeker, line 1859: I miss an iteration argument or is power method not the same as power iteration?

boeddeker (commented):
I added some comments to the core code. The code feels somehow familiar, it is not too far away from our.

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+    mask: Optional[Tensor] = None,
+    multi_mask: bool = False,
+    normalize: bool = True,
+    eps: float = 1e-10,
+) -> Tensor:
+    """Compute cross-channel power spectral density (PSD) matrix.
+
+    Args:
+        specgram (Tensor): multi-channel complex-valued STFT matrix.
+            Tensor of dimension `(..., channel, freq, time)`
```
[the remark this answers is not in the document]
nateanl: The shape is to align with other methods in torchaudio: https://github.com/pytorch/audio/blob/main/CONTRIBUTING.md?plain=1#L226

But your point makes sense, for the solution, we can keep the argument ``(..., channel, freq, time)`` and do transpose internally to reduce memory.

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+        reg (float, optional): regularization factor (Default: 1e-8)
+        eps (float, optional): a value to avoid the correlation matrix is all-zero (Default: ``1e-8``)
+
+    Returns:
+        torch.Tensor: regularized matrix (..., channel, channel)
+    """
+    # Add eps
+    C = mat.size(-1)
+    eye = torch.eye(C, dtype=mat.dtype, device=mat.device)
+    with torch.no_grad():
```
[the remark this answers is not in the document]
nateanl: This is adopted from https://github.com/espnet/espnet/blob/master/espnet2/enh/layers/beamformer.py#L587
My guess is the trace works as normalization. @Emrys365 any thoughts on that?

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
         return costs.sum()
 
     return costs
+
+
+def _compute_mat_trace(input: torch.Tensor, dim1: int = -1, dim2: int = -2) -> torch.Tensor:
```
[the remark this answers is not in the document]
nateanl: Yes, this is a temporary solution, will refactor it when https://github.com/pytorch/pytorch/pull/62714 is merged to PyTorch.

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+def compute_rtf_evd(psd_s: Tensor) -> Tensor:
+    r"""Estimate the relative transfer function (RTF) or the steering vector by eigenvalue decomposition.
+
+    Args:
+        psd_s (Tensor): The covariance matrix of target speech.
+            Tensor of dimension `(..., freq, channel, channel)`
+
+    Returns:
+        Tensor: The estimated RTF of target speech.
+            Tensor of dimension `(..., freq, channel, 1)`
```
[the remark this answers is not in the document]
nateanl: Oh, this is to fit the need of einsum, will change it.

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+            psd_n.size()[:-3] + psd_n.size()[-1:], device=psd_n.device, dtype=psd_n.dtype
+        )  # (..., channel)
+        ref_vector[..., reference_channel].fill_(1)
+    else:
+        ref_vector = reference_channel
+    phi = torch.linalg.solve(psd_n, psd_s)  # psd_n.inv() @ psd_s
+    rtf = torch.einsum("...fec,...c->...fe", [phi, ref_vector])
+    rtf = rtf.unsqueeze(-1)  # (..., freq, channel, 1)
+    rtf = torch.matmul(phi, rtf)
+    rtf = torch.matmul(psd_s, rtf)
```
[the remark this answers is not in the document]
nateanl: Could you share the codepoint of your implementation? Not sure if it's the pb_bss or other ones.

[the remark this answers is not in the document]
boeddeker: It's not yet published, because I rewrote the code from `torch_complex` to native `torch` and haven't found the time to cleanup, testing and publish the code.

I copied my code to gist.github.com: 
[beamformer.py](https://gist.github.com/boeddeker/ba7226c0f15bc34a610689a47bfeb6f1)

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+
+    Args:
+        psd_s (Tensor): The covariance matrix of target speech.
+            Tensor of dimension `(..., freq, channel, channel)`
+
+    Returns:
+        Tensor: The estimated RTF of target speech.
+            Tensor of dimension `(..., freq, channel, 1)`
+    """
+    w, v = torch.linalg.eig(psd_s)  # (..., freq, channel, channel)
```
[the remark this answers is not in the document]
nateanl: You are right, it's actually faster than manually finding the eigenvector with max eigenvalue. WiIl update it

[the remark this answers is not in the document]
boeddeker: That is an advantage of `eigh`. When you assume the input is Hermitian, then the eigenvalues are real, and can be ordered.

torchaudio/functional/functional.py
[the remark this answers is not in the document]
nateanl: But I doubt why the speed is faster. I thought tensor operation at one time should be faster than for loop. Or is the for loop in parallel?

torchaudio/functional/functional.py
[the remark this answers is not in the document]
boeddeker: In https://arxiv.org/pdf/2011.15003.pdf we wrote a small section about power iteration.
image
When you have only one iteration, you don't need `psd_n`, because it cancels.

In my implementation, I tried to minimize the memory footprint, to avoid an OOM error, because of the backward graph.

torchaudio/functional/functional.py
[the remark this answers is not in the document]
boeddeker: It can be faster. 
With batched operations you get the advantage, that the python overhead is smaller and the C or C++ code can use vector instructions. But at some point, the python overhead is neglectable and vector instructions cannot further increase the execution time. On the other side, when your data gets too large, it will no longer fit into the L1/L2/L3 cache. So this will slow down your program, because the data between each operation has to be moved between the caches and maybe also the RAM.

In WPE it turned out, that a loop over the frequency is sometimes faster than the batched implementation, especially for longer audio clips. (On the GPU this is different, but I got only a speedup of 3 with CUPY vs CPU and profiled it only once)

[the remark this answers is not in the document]
boeddeker: > The shape is to align with other methods in torchaudio:

I expected that answer. We gave that idea up, because different algorithms/approaches prefer different shapes and some of them get very ugly or slow, when you enforce a particular shape.

I just wanted to mention this, and aligning shapes is a very good argument.

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,274 @@ def rnnt_loss(
...
+    """Compute cross-channel power spectral density (PSD) matrix.
+
+    Args:
+        specgram (Tensor): multi-channel complex-valued STFT matrix.
+            Tensor of dimension `(..., channel, freq, time)`
+        mask (Tensor or None, optional): Time-Frequency mask for normalization.
+            Tensor of dimension `(..., freq, time)` if multi_mask is ``False`` or
+            of dimension `(..., channel, freq, time)` if multi_mask is ``True``.
+            (Default: ``None``)
+        multi_mask (bool, optional): whether to use multi-channel Time-Frequency masks. (Default: ``False``)
```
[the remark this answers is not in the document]
nateanl: I see. I feel like users can do ``mean`` or ``median`` reduction before calling psd method, while ``multi_mask`` can be changed to use different masks for each channel. Does that make sense to you?

torchaudio/functional/functional.py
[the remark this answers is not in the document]
nateanl: Got it, `R_n^-1 R_n` becomes `I`. if the iteration is less than 2, you don't even need to compute `\Phi`, instead `PSD_s  matmul u` should suffice. 

Thanks. I will update the code to improve it.

torchaudio/functional/functional.py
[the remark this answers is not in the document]
boeddeker: I am not sure, if providing an option to multiply different masks to different channels should be supported.
With that operation, you could break the calculation of the correlation between the channels and further the implementation is `sum (mask*obs)@(mask*obs)^H`, while you have currently `sum mask*obs@obs^H`.

Until now, I have only seen papers that do that, but not a paper that shows that it is better.
It's also not difficult to do it manually.

I only wanted to list different options, that, if such an option is provided, the name should be carefully selected to avoid confusions.

torchaudio/functional/functional.py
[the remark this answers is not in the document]
boeddeker: Yes. In my implementation, I combined the one hot vector (line 2) with line 4, so I was able to use getitem instead of a matrix multiplication. While this combination is more efficient, it produces less nice code.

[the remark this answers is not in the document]
nateanl: Addressed in https://github.com/pytorch/audio/pull/2181/commits/07bc84597b3cb9b0b9775634f2bc14d4ddbe3631

torchaudio/transforms.py
```diff
@@ -1689,14 +1657,18 @@ class MVDR(torch.nn.Module):
     PSD matrices of speech and noise, respectively.
```
danpovey, line 1657: You might want to also refer to a paper explaining MVDR?
If ESPNet is refactored in future, your link above may die, an arxiv one may be more permanent.
Although I suppose people can just google it.

torchaudio/transforms.py
```diff
@@ -1730,242 +1703,100 @@ def __init__(
...
-
-    def forward(
-        self, specgram: torch.Tensor, mask_s: torch.Tensor, mask_n: Optional[torch.Tensor] = None
-    ) -> torch.Tensor:
+    def forward(self, specgram: Tensor, mask_s: Tensor, mask_n: Optional[Tensor] = None) -> Tensor:
         """Perform MVDR beamforming.
 
         Args:
-            specgram (torch.Tensor): the multi-channel STF of the noisy speech.
+            specgram (Tensor): the multi-channel STF of the noisy speech.
```
danpovey, line 1788: Elsewhere I think you refer  to STF as STFT?  And perhaps you should specifiy that it is complex? 
(I assume it is..)

torchaudio/transforms.py
```diff
@@ -1730,242 +1703,100 @@ def __init__(
...
                 Tensor of dimension `(..., channel, freq, time)`
-            mask_s (torch.Tensor): Time-Frequency mask of target speech.
+            mask_s (Tensor): Time-Frequency mask of target speech.
                 Tensor of dimension `(..., freq, time)` if multi_mask is ``False``
```
danpovey, line 1791: I think perhaps you should comment on the expected type and/or range of mask_s and mask_n.  E.g. are they expected to be boolean or reals in [0..1]; and are they expected to sum to one?  Also perhaps comment on how
they would typically be obtained.  This may help those (such as myself!) who are not super familiar with 
beamforming issues.
You might also want to specify that mask_s and mask_n cannot be zero everywhere (if indeed that is the case).

torchaudio/transforms.py
```diff
@@ -1730,242 +1703,100 @@ def __init__(
...
-            mask_n (torch.Tensor or None, optional): Time-Frequency mask of noise.
+            mask_n (Tensor or None, optional): Time-Frequency mask of noise.
                 Tensor of dimension `(..., freq, time)` if multi_mask is ``False``
                 or or dimension `(..., channel, freq, time)` if multi_mask is ``True``
                 (Default: None)
 
         Returns:
-            torch.Tensor: The single-channel STFT of the enhanced speech.
+            Tensor: The single-channel STFT of the enhanced speech.
                 Tensor of dimension `(..., freq, time)`
```
danpovey, line 1800: Perhaps specify that it's complex?

danpovey (commented):
Some small comments

torchaudio/transforms.py
nateanl: Sometimes the sum of `mask_s` and `mask_n` will not be one, for example, the [phase-sensitive mask](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.727.7832&rep=rep1&type=pdf) or the masks directly estimated by the neural networks. But you are right, in this implementation they must be real instead of boolean.

Sure, I will update the docstring to reflect those, thanks!

torchaudio/transforms.py
nateanl: Okay, will add a citation to the MVDR paper after the ESPNet link.

torchaudio/transforms.py
nateanl: Oops, it's a typo, should be STFT. Maybe calling it `spectrogram` is better for such description.

boeddeker: https://en.wikipedia.org/wiki/Spectrogram :

> A spectrogram is a visual representation of the [spectrum](https://en.wikipedia.org/wiki/Spectral_density) of [frequencies](https://en.wikipedia.org/wiki/Frequencies) of a signal as it varies with time. When applied to an [audio signal](https://en.wikipedia.org/wiki/Audio_signal), spectrograms are sometimes called sonographs, voiceprints, or voicegrams. When the data are represented in a 3D plot they may be called [waterfall displays](https://en.wikipedia.org/wiki/Waterfall_display).

For me, the `spectrogram` is a real valued signal, e.g. the magnitude or the power of the STFT signal.
According to wikipedia, it is the visualization of a signal, with a time and a frequency axis.
If you want to keep the name, I think it is important to mention, that it is complex values,
because often this name is connected with something real values.

torchaudio/transforms.py
boeddeker: > You might also want to specify that mask_s and mask_n cannot be zero everywhere (if indeed that is the case).

This is partially correct. Normally, the necessary property is, that the mask cannot be zero everywhere for any frequency. So on each frequency you need one non-zero value.
The implementation on the other side, uses some regulations, that the code will not fail.

In my experience, I ensure that the mask has this property and drop the regulations.

nateanl: In neural-network based training, the mask values can be very small for some frequency bin, considering that I will keep the regulation term and relax the non-zero check.

torchaudio/transforms.py
nateanl: This is true, but I do see some papers use `spectrum` to refer the complex valued STFT, like the phase-sensitive mask paper. Since torchaudio support `Spectrogram` module that returns complex Tensor, using the same term makes it more consistent. But yeah, definitely mention the `complex-valued` helps users understand better.

torchaudio/transforms.py
boeddeker: I know. I observed that `mask + 0.01` is a good alternative regulation and has the advantage to be scale independent.
Not sure, but I compared it once and saw a tiny gain for this, so I kept this in my experiments.

torchaudio/transforms.py
boeddeker: So `spectrum` would maybe a better name?
For that, it is at least difficult to find a definition, that it is real or complex valued.

nateanl: > So spectrum would maybe a better name?

Sure, let's use it in the docstring.

nateanl: Addressed in https://github.com/pytorch/audio/pull/2181/commits/4c1a539aefaa9bb27e2f3628f694cf107386001d

torchaudio/transforms.py
nateanl: Addressed in https://github.com/pytorch/audio/pull/2181/commits/4c1a539aefaa9bb27e2f3628f694cf107386001d

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -250,6 +252,59 @@ def test_bandreject_biquad(self, central_freq, Q):
...
+            mask = torch.rand(10, 5)
+        else:
+            mask = None
+        self.assert_grad(F.compute_power_spectral_density_matrix, (specgram, mask))
+
+    def test_compute_mvdr_weights_souden(self):
+        torch.random.manual_seed(2434)
+        channel = 4
+        psd_speech = torch.rand(64, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(64, channel, channel, dtype=torch.cfloat)
```
mthrok, line 274: Autograd tests are one of the most time-consuming ones. I think batch size 3 is suffice.

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -250,6 +252,59 @@ def test_bandreject_biquad(self, central_freq, Q):
...
+
+    def test_compute_mvdr_weights_rtf(self):
+        torch.random.manual_seed(2434)
+        channel = 4
+        rtf = torch.rand(64, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(64, channel, channel, dtype=torch.cfloat)
+        self.assert_grad(F.compute_mvdr_weights_rtf, (rtf, psd_noise, 0))
+
+    # The eigenvector can be different in different runs, expected to fail
+    @expectedFailure
```
mthrok, line 285: Reading https://pytorch.org/docs/stable/generated/torch.linalg.eigh.html, there seems no way to control the randomness. Since this is expected to fail, what value does this test provides?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -644,6 +644,59 @@ def func(tensor):
         tensor = torch.view_as_complex(torch.randn(2, 1025, 400, 2))
         self._assert_consistency_complex(func, tensor)
 
+    def test_compute_power_spectral_density_matrix(self):
+        def func(tensor):
+            return F.compute_power_spectral_density_matrix(tensor)
```
mthrok, line 649: There seems no benefit of defining this `func`.

test/torchaudio_unittest/functional/autograd_impl.py
mthrok: or is this frame length?

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1637,260 @@ def rnnt_loss(
...
+
+    Returns:
+        Tensor: the complex-valued MVDR beamforming weight matrix of dimension (..., freq, channel).
+    """
+    if diagonal_loading:
+        psd_n = _tik_reg(psd_n, reg=diag_eps, eps=eps)
+    numerator = torch.linalg.solve(psd_n, psd_s)  # psd_n.inv() @ psd_s
+    # ws: (..., C, C) / (...,) -> (..., C, C)
+    ws = numerator / (_compute_mat_trace(numerator)[..., None, None] + eps)
+    if isinstance(reference_channel, int):
```
mthrok, line 1760: nit: Would swapping `isinstance` with `torch.jit.isinstance` help support TorchScript?

test/torchaudio_unittest/functional/autograd_impl.py
nateanl: 64 is the frequency dimension. We can reduce it to 5 or 10.

test/torchaudio_unittest/functional/autograd_impl.py
nateanl: This works as a flag showing eigenvalue decomposition itself can have different results on the same matrix.

boeddeker: You can normalize the rtf vector (`rtf / rtf[0].conj()`), then it should work.
If it still does not work, the eigenvalue gap should be larger, i.e. the difference between the largest and second-largest eigenvalue, e.g.:
```python
import numpy as np
channel = 4
samples = 100  # samples >> channel
data = np.random.randn(channel, samples)
cov = data @ data.T.conj()  # Hermetian matrix
eigenvalues, eigenvectors = np.linalg.eigh(cov)
eigenvalues[...] = np.random.uniform(0, 1, size=eigenvalues.shape)  # Change the eigenvalues to be between 0 and 1
eigenvalues[0] = 2  # Set first eigenvalue to be the largest
cov = np.einsum('ij,j,kj->ik', eigenvectors, eigenvalues, eigenvectors.conj())  # Revert the eigh decomposition
print(np.linalg.eigh(cov))
```

boeddeker: Btw. probably the gradient for eig and eigh was derived with the constraint, that you have no gradient in the direction of the length or a phase change, i.e. the objective does not change, when you replace an eigenvector with itself multiplied with any nonzero complex number.

nateanl: IMO the returned eigenvalue can be different even when the eigenvectors are unique. One simple example is `-eigenvalue` and `-eigenvector` become another solution. 
From PyTorch side, they seem to hard-coded check it and throw a runtime error in
https://github.com/pytorch/pytorch/blob/master/torch/csrc/autograd/FunctionsManual.cpp#L2907-L2910

mthrok: What is the source of disorder of the eigenvalues? The reason it fails is essentially multiple runs generate different results. If it is caused by values randomly initialized inside of `torch.linarg.eigh`, then wrapping the tested function with fixing random seed should work.

This is done in transform autograd test as following

https://github.com/pytorch/audio/blob/8e3c6144e88dcbbbed6dec2b47da33fac2278820/test/torchaudio_unittest/transforms/autograd_test_impl.py#L17-L27

nateanl: @mthrok  See the warning notes of [``torch.linalg.eigh``](https://pytorch.org/docs/stable/generated/torch.linalg.eigh.html).

nateanl: Before make the gradcheck work, I need to resolve the runtime error I saw in https://github.com/pytorch/pytorch/blob/master/torch/csrc/autograd/FunctionsManual.cpp#L2907-L2910
Here is a more detailed error message:
```python
RuntimeError: linalg_eigh_backward: The eigenvectors in the complex case are specified up to multiplication by e^{i phi}. The specified loss function depends on this quantity, so it is ill-defined.
```
It should be related with only returning the partial eigenvector. If I return the eigenvalues and eigenvectors, then the error becomes the Jacobian mismatch by gradcheck.

mthrok: If the implementation does not support gradient, then there is no point of having this test with expected failure. It simply adds the cost for testing and maintaining. Especially, since the explanation of why it fails is not on the spot, so it will be difficult for random future maintainers to decide what to do with this test.
If we want to let people know, we can simply put it down on the docstring.

boeddeker: > IMO the returned eigenvalue can be different even when the eigenvectors are unique. One simple example is -eigenvalue and -eigenvector become another solution.

The eigenvalues are unique, and for a PSD matrix the eigenvalue must be non-negative. You only get an issue with duplicated eigenvalues, but then you have another issue (NaN grad). The freedom of the eigenvector is a complex scalar, in your example you selected -1. As far as I know, all implementations yield eigenvectors with l2 norm 1 (except generalized eigh), so the observable freedom is `e^{j angle}`

When you use `eigenvector / eigenvector[0].conj()` as an operation after the eigenvalue decomposition, the output will be unique. In 2017 with chainer and numpy I wrote numeric gradient testcases for the eigenvalue decomposition for beamforming, and they worked, when I considered all realization aspects (non-unique eigh output, eigh does not use all input values, ...).

nateanl:
Split the PR to the following small PRs:
- [ ] https://github.com/pytorch/audio/pull/2227
- [ ] Add ``compute_mvdr_weights_souden``
- [ ] Add ``compute_mvdr_weights_rtf``
- [ ] Add ``compute_rtf_evd``
- [ ] Add ``compute_rtf_power``
- [ ] Add ``apply_beamforming``
- [ ] Refactor ``MVDR`` module

test/torchaudio_unittest/functional/autograd_impl.py
nateanl: @boeddeker  I verified that using ``eigenvector / eigenvector[..., -1:].conj()`` passed the gradcheck. Thanks! In my case the eigenvalues are in ascending order, hence the normalization should rely on the last dimension instead of 0th?

nateanl: Sorry my mistake. I converted the psd matrix to real Tensor by mistake. It still fails the gradcheck now...

boeddeker: > I verified that using eigenvector / eigenvector[..., -1:].conj() passed the gradcheck.

Sorry, I copied my code example from above and renamed the variable.
The correct code is
```python
...
assert rtf.ndim == 1, rtf.shape
reference_channel = 0
rtf = rtf / rtf[reference_channel].conj()
...
```
So I assumed to have a single eigenvector and normalize it by a reference channel. In this way, the eigenvectors are unique.

When I implemented the gradient check, I had issues with inputs like `psd_speech`. 
It was necessary to include something like `psd_speech = 0.5 * (psd_speech + hermitian(psd_speech))` in the checked function, otherwise the changes in the input were not correctly reflected in the output, because eigh ignores values of the matrix (e.g. all values above the diagonal aren't used). In the user code, this is not necessary, because `F.compute_power_spectral_density_matrix` takes care of this.

popcornell: Nevermind, the rtf normalization seems is there already and is True by default.

PR #2183, opened 2022-01-28
Add global stats script and new json for LibriSpeech RNN-T training recipe

hwangjeff:
Adds script for generating global feature statistics along with new feature statistics json for LibriSpeech RNN-T training recipe.


examples/asr/librispeech_emformer_rnnt/global_stats.py
```diff
@@ -0,0 +1,69 @@
...
+        help="Path to LibriSpeech datasets. "
+        "All of 'train-clean-360', 'train-clean-100', and 'train-other-500' must exist.",
+    )
+    parser.add_argument(
+        "--output_path",
+        default=pathlib.Path("global_stats.json"),
+        type=pathlib.Path,
+        help="File to save feature statistics to. (Default: './global_stats.json')",
+    )
+    args = parser.parse_args()
```
mthrok, line 50: nit: I would extract the logic to parse CLI arguments to a helper function. That way the execution flow of the main function becomes easier to follow.

examples/asr/librispeech_emformer_rnnt/global_stats.py
```diff
@@ -0,0 +1,69 @@
...
+        N += M
+
+        if idx % 100 == 0:
+            logger.info(f"Processed {idx}")
+
+    return E_x, (E_x_2 - E_x ** 2) ** 0.5
+
+
+def cli_main():
+    parser = ArgumentParser()
```
mthrok, line 36: Can you add general CLI `description`?
My recommendation is to add module level docstring and reuse it in `description` argument. https://github.com/pytorch/audio/blob/main/examples/libtorchaudio/speech_recognition/parse_librispeech.py

mthrok (approved):

examples/asr/librispeech_emformer_rnnt/global_stats.py
```diff
@@ -0,0 +1,69 @@
...
+        M = scaled_mel_spec.size(0)
+
+        E_x = E_x * (N / (N + M)) + sum / (N + M)
+        E_x_2 = E_x_2 * (N / (N + M)) + sq_sum / (N + M)
+        N += M
+
+        if idx % 100 == 0:
+            logger.info(f"Processed {idx}")
+
+    return E_x, (E_x_2 - E_x ** 2) ** 0.5
```
nateanl, line 32: After checking the formula, the std computation may not be precise. We need to run two rounds of iterations to get the std:
- compute `E_x` after one iteration
- compute the mean of `(x - E_x) ** 2`  in the second iteration
- compute square root of the mean to get std

nateanl: Actually, you are right. The results are identical, which is `(E_x_2 - E_x ** 2) ** 0.5` 🙂

examples/asr/librispeech_emformer_rnnt/global_stats.py
```diff
@@ -0,0 +1,78 @@
...
+
+import torch
+import torchaudio
+from utils import GAIN, piecewise_linear_log, spectrogram_transform
+
+logger = logging.getLogger()
+
+
+def parse_args():
+    parser = ArgumentParser(description=__doc__)
```
mthrok, line 20: nit: You might need `formatter_class=RawTextHelpFormatter`. Try running `--help` and see if the docstring is printed as you intended.

PR #2184, opened 2022-01-28
Add CTC decoder timesteps

carolineechen:
add timesteps field to CTC decoder hypotheses, corresponding to the time step of occurrences of non-blank tokens


test/torchaudio_unittest/prototype/ctc_decoder_test.py
```diff
@@ -64,6 +64,16 @@ def test_shape(self):
...
 
+    def test_timesteps_shape(self):
+        """Each token should correspond with a timestep"""
+        emissions = self._get_emissions()
+        decoder = self._get_decoder()
+
+        results = decoder(emissions)
+        for i in range(emissions.shape[0]):
+            result = results[i][0]
+            self.assertEqual(result.tokens.shape[0], result.timesteps.shape[0])
```
mthrok, line 75: Can we compare the whole shape?
```suggestion
            self.assertEqual(result.tokens.shape, result.timesteps.shape)
```

test/torchaudio_unittest/prototype/ctc_decoder_test.py
```diff
@@ -64,6 +64,16 @@ def test_shape(self):
...
+
```
mthrok, line 76: Can we add a test for timesteps value with fixed input emission?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -26,10 +26,12 @@ class Hypothesis(NamedTuple):
     :ivar torch.LongTensor tokens: Predicted sequence of token IDs
     :ivar List[str] words: List of predicted words
     :ivar float score: Score corresponding to hypothesis
+    :ivar torch.IntTensor timestpeps
```
mthrok, line 29: Could you mention the shape of the tensor here?
(same goes to `tokens` as well, though strictly speaking it's not this PR's scope)

mthrok (approved):
Looks good but kind of hard to see if the output timesteps value really makes sense.

Could you update the tutorial to print out timesteps? (and maybe split audio into words, like in alignment example https://pytorch.org/audio/main/tutorials/forced_alignment_tutorial.html#id4)

test/torchaudio_unittest/prototype/ctc_decoder_test.py
carolineechen: this is doable but slightly more complicated because the emission goes through the decoder and produces the final tokens+timesteps without the intermediate duplicate or blank tokens. do you think testing the internal function `_get_timesteps` (and `_get_tokens`) with an unprocessed token list would be sufficient?

mthrok: > testing the internal function `_get_timesteps` (and `_get_tokens`) with an unprocessed token list

Yeah, that sounds reasonable.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: thoughts on how to describe the shape? they are the post-processed tokens and corresponding timesteps list, where blanks and duplicates have already been filtered and so it's shape is different from emission length

ex/ `['|', 'f', 'o', 'o', '|', 'b', 'a', 'r', '|']`

carolineechen:
> Could you update the tutorial to print out timesteps? (and maybe split audio into words, like in alignment example

will follow up on this on a separate PR and try to merge the feature first, but at a first glance the alignment looks reasonable
Screen Shot 2022-01-31 at 5 42 18 PM

torchaudio/prototype/ctc_decoder/ctc_decoder.py
mthrok: Something like "shape `(L, )`, where `L` is the length of output label sequence."?

mthrok:
> > Could you update the tutorial to print out timesteps? (and maybe split audio into words, like in alignment example
> 
> thinking of following up on this on a separate PR and try to merge the feature first, but at a first glance the alignment looks reasonable 

Nice! This indeed looks reasonable.

mthrok (approved):

PR #2192, opened 2022-02-02
Revise RNN-T pipeline streaming decoding logic

hwangjeff:
Rather than apply SentencePiece's `decode` to directly convert each hypothesis's token id sequence to an output string, we convert each token id sequence to word pieces and then manually join the word pieces ourselves. This allows us to preserve leading whitespaces on output strings and therefore account for word breaks and continuations across token processor invocations, which is particularly useful when performing streaming ASR.

https://user-images.githubusercontent.com/8345689/152093668-11fb775a-bf7b-4b1d-9516-9f8d5a9b6683.mov

Versus the previous behavior visualized in #2093, the scheme here properly constructs words comprising multiple pieces.

### DIFF
--- a/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
+++ b/examples/asr/librispeech_emformer_rnnt/pipeline_demo.py
@@ -41,9 +41,8 @@ def cli_main():
                 features, length = streaming_feature_extractor(segment)
                 hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
             hypothesis = hypos[0]
-            transcript = token_processor(hypothesis.tokens)
-            if transcript:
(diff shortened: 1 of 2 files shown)


mthrok (approved):
LGTM.

I wonder 
* if `\u2581` is universal across different space-separated languages.
* if the `_TokenProcessor` can handle non-latin languages as well.

hwangjeff:
> * if `\u2581` is universal across different space-separated languages.

SentencePiece's [README](https://github.com/google/sentencepiece) seems to suggest so, but I haven't confirmed

> * if the `_TokenProcessor` can handle non-latin languages as well.

Python's `str` can handle unicode, so I suppose the interface should account for non-Latin languages

PR #2195, opened 2022-02-02
Add training recipe of Emformer trained on TED-LIUM release 3 dataset

nateanl:


examples/asr/tedlium3_emformer_rnnt/README.md
```diff
@@ -0,0 +1,31 @@
+# Emformer RNN-T ASR Example for TED-LIUM release 3 dataset
+
+This directory contains sample implementations of training and evaluation pipelines for an on-device-oriented streaming-capable Emformer RNN-T ASR model.
+
+## Usage
+
+### Training
+
+[`train.py`](./train.py) trains an Emformer RNN-T model on TED-LIUM release 3 using PyTorch Lightning. Note that the script expects users to have access to GPU nodes for training and provide paths to the full LibriSpeech dataset and the SentencePiece model to be used to encode targets.
```
carolineechen, line 9:
```suggestion
[`train.py`](./train.py) trains an Emformer RNN-T model on TED-LIUM release 3 using PyTorch Lightning. Note that the script expects users to have access to GPU nodes for training and provide paths to the full TED-LIUM release 3 dataset and the SentencePiece model to be used to encode targets.
```

examples/asr/tedlium3_emformer_rnnt/train_spm.py
```diff
@@ -0,0 +1,74 @@
+"""Traine the SentencePiece model by using the transcripts of TED-LIUM release 3 training set.
```
carolineechen, line 1:
```suggestion
"""Train the SentencePiece model by using the transcripts of TED-LIUM release 3 training set.
```

examples/asr/tedlium3_emformer_rnnt/lightning.py
```diff
@@ -0,0 +1,306 @@
...
+        self.invstddev = torch.tensor(blob["invstddev"])
+
+    def forward(self, input):
+        return (input - self.mean) * self.invstddev
+
+
+def _piecewise_linear_log(x):
+    x[x > math.e] = torch.log(x[x > math.e])
+    x[x <= math.e] = x[x <= math.e] / math.e
+    return x
```
hwangjeff, line 115: we can remove this since we're now using `piecewise_linear_log` from `utils`

examples/asr/tedlium3_emformer_rnnt/lightning.py
```diff
@@ -0,0 +1,306 @@
...
+    def configure_optimizers(self):
+        return (
+            [self.optimizer],
+            [
+                {
+                    "scheduler": self.lr_scheduler,
+                    "monitor": "Losses/val_loss",
+                    "interval": "epoch",
+                },
+                {"scheduler": self.warmup_lr_scheduler, "interval": "step"},
```
hwangjeff, line 259: i need to clean this up in the librispeech recipe: the way things are configured,`self.lr_scheduler` doesn't actually do anything, i.e. only `self.warmup_lr_scheduler` influences the learning rate, so we can remove it

examples/asr/tedlium3_emformer_rnnt/lightning.py
```diff
@@ -0,0 +1,306 @@
...
+            FunctionalModule(lambda x: torch.nn.functional.pad(x, (0, 4))),
+            FunctionalModule(lambda x: x.transpose(1, 2)),
+        )
+
+        self.tedlium_path = tedlium_path
+
+        self.sp_model = spm.SentencePieceProcessor(model_file=sp_model_path)
+        self.blank_idx = self.sp_model.get_piece_size()
+
+    def _extract_labels(self, samples: List):
```
hwangjeff, line 189: maybe a brief comment here would be helpful since it might not be obvious to the reader why we're bothering with these remapping gymnastics

examples/asr/tedlium3_emformer_rnnt/lightning.py
```diff
@@ -0,0 +1,306 @@
...
+        self.warmup_lr_scheduler = WarmupLR(self.optimizer, 10000)
+
+        self.train_data_pipeline = torch.nn.Sequential(
+            FunctionalModule(lambda x: piecewise_linear_log(x * GAIN)),
+            GlobalStatsNormalization(global_stats_path),
+            FunctionalModule(lambda x: x.transpose(1, 2)),
+            torchaudio.transforms.FrequencyMasking(27),
+            torchaudio.transforms.FrequencyMasking(27),
+            TimeMasking(100, 0.2),
+            TimeMasking(100, 0.2),
```
hwangjeff, line 172: another thing to also clean up in the librispeech recipe — replacing this with `torchaudio.transforms.TimeMasking` since it now supports parameter `p`. i can do the swap later, if you'd like

examples/asr/tedlium3_emformer_rnnt/train.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+    parser.add_argument(
+        "--global-stats-path",
+        default=pathlib.Path("global_stats.json"),
+        type=pathlib.Path,
+        help="Path to JSON file containing feature means and stddevs.",
+    )
+    parser.add_argument(
+        "--tedlium-path",
+        type=pathlib.Path,
```
hwangjeff, line 71: and yet another thing to fix in the original as well — `required=True`

hwangjeff (approved):
looks good — just some small things. of these, there are some that were replicated from the librispeech emformer rnnt recipe that should be fixed, but i can address them on both sides later, if you'd like

mthrok:
@hwangjeff @nateanl Can we merge the two recipes for Emformer into one directory?
The duplicated code means increased maintenance cost, and I am not fun of using symlink to reuse the code.

hwangjeff:
@mthrok merging into one directory is fine. there are some other bits that can be refactored and reused from the librispeech recipe. would prefer to at least keep separate lightning modules to accommodate recipe-specific logic, e.g. `<unk>` substring handling in labels and the custom dataset reading wrapper for tedlium, and to allow for tuning each recipe independently

mthrok:
> keep separate lightning modules to accommodate recipe-specific logic, e.g. `<unk>` substring handling in labels and the custom dataset reading wrapper for tedlium, and to allow for tuning each recipe independently

Yeah, that makes sense. Hope that lightning helps modularizing nicely. Ideally, each script, (train, eval, demo) takes the same set of argument to change the behavior. My main concern is that the main training loop is not shared, thus it increases the cost for testing. yet, the utility module is shared as symlink. I know many recipe projects use symlink, but this approach should be the last resort.

PR #2197, opened 2022-02-03
Added example for ComputeDeltas

radioactive11:
Example code for [ComputeDeltas](https://github.com/pytorch/audio/blob/94f4ef0fb4a7bdb58122ddad204ecc9a697c4098/torchaudio/transforms.py#L936) as mentioned in [this issue](https://github.com/pytorch/audio/issues/1564)


torchaudio/transforms.py
```diff
@@ -941,6 +941,12 @@ class ComputeDeltas(torch.nn.Module):
     Args:
         win_length (int, optional): The window length used for computing delta. (Default: ``5``)
         mode (str, optional): Mode parameter passed to padding. (Default: ``'replicate'``)
+
+    Example
+        >>> spectrogram = transforms.Spectrogram()
+        >>> compute_deltas = ComputeDeltas(win_length=5, mode="replicate")
+        >>> original = spectrogram(waveform)
+        >>> delta = compute_deltas(original)
```
nateanl, line 949:
```suggestion
        >>> specgram = torch.randn(2, 257, 100, dtype=torch.cfloat)
        >>> transform = transforms.ComputeDeltas(win_length=5, mode="replicate")
        >>> delta = transform(specgram)
```

nateanl (changes requested):
Hi @radioactive11, thanks for your contribution! I have added a comment on the example. Looks like your commit history contains other changes, could you clean up your commit history and only maintain the one related to ``ComputeDelats``? Thanks!

torchaudio/transforms.py
radioactive11: @nateanl Thank you for the review! I will make the suggested changes and update the PR.

radioactive11: @nateanl I have made the necessary changes. If you want, I can raise a new, clean PR with only the requested changes.

nateanl: Thanks. Let's make a new PR with just one commit then. I will help merge it when you finish.

PR #2202, opened 2022-02-04
Improve device streaming

mthrok:
This commit adds tutorial for device ASR, and update API for device streaming.

The changes for the interface are
1. Add `timeout` and `backoff` parameters to `process_packet` and `stream` methods.
2. Move `fill_buffer` method to private.

When dealing with device stream, there are situations where the device buffer is not
ready and the system returns `EAGAIN`. In such case, the previous implementation of
`process_packet` method raised an exception in Python layer , but for device ASR, 
this is inefficient. A better approach is to retry within C++ layer in blocking manner.
The new `timeout` parameter serves this purpose.

### DIFF
--- a/examples/tutorials/device_asr.py
+++ b/examples/tutorials/device_asr.py
@@ -1,0 +1,330 @@
+"""
+Device ASR with Emformer RNN-T
+==============================
+
+**Author** : `Moto Hira <moto@fb.com>`__, `Jeff Hwang <jeffhwang@fb.com>`__.
+
+This tutorial shows how to use Emformer RNN-T and streaming API
+to perform speech recognition on a streaming device input, i.e. microphone
+on laptop.
+
+.. note::
+
+   This tutorial requires prototype Streaming API and ffmpeg>=4.1.
+
+   Prototype features are not part of binary releases, but available in
+   nightly build. Please refer to https://pytorch.org for installing
+   nightly build.
+
+   If you are using Anaconda Python distribution,
+   ``conda install -c anaconda ffmpeg`` will install
+   the required libraries.
+
+.. note::
+
+   This tutorial was tested on MacBook Pro and Dynabook with Windows 10.
+
+   This tutorial does NOT work on Google Colab because the server running
+   this tutorial does not have a microphone that you can talk to.
+"""
+
+######################################################################
+# 1. Overview
+# -----------
+#
+# We use streaming API to fetch audio from audio device (microphone)
+# chunk by chunk, then run inference using Emformer RNN-T.
+#
+# For the basic usage of the streaming API and Emformer RNN-T
+# please refer to
+# `Media Stream API tutorial <./streaming_api_tutorial.html>`__ and
+# `Online ASR with Emformer RNN-T <./online_asr_tutorial.html>`__.
+#
+
+######################################################################
+# 2. Checking the supported devices
+# ---------------------------------
+#
+# Firstly, we need to check the devices that Streaming API can access,
+# and figure out the arguments (``src`` and ``format``) we need to pass
+# to :py:func:`~torchaudio.prototype.io.Streamer` class.
+#
+# We use ``ffmpeg`` command for this. ``ffmpeg`` abstracts away the
+# difference of underlying hardware implementations, but the expected
+# value for ``format`` vary across OS and each ``format`` defines
+# different syntax for ``src``.
+#
+# The details of supported ``format`` values and ``src`` syntax can
+# be found in https://ffmpeg.org/ffmpeg-devices.html.
+#
+# For macOS, the following command will list the available devices.
+#
+# .. code::
+#
+#    $ ffmpeg -f avfoundation -list_devices true -i dummy
+#    ...
+#    [AVFoundation indev @ 0x126e049d0] AVFoundation video devices:
+#    [AVFoundation indev @ 0x126e049d0] [0] FaceTime HD Camera
+#    [AVFoundation indev @ 0x126e049d0] [1] Capture screen 0
+#    [AVFoundation indev @ 0x126e049d0] AVFoundation audio devices:
+#    [AVFoundation indev @ 0x126e049d0] [0] ZoomAudioDevice
+#    [AVFoundation indev @ 0x126e049d0] [1] MacBook Pro Microphone
+#
+# We will use the following values for Streaming API.
+#
+# .. code::
+#
+#    Streamer(
+#        src = ":1",  # no video, audio from device 1, "MacBook Pro Microphone"
+#        format = "avfoundation",
+#    )
+
+######################################################################
+#
+# For Windows, ``dshow`` device should work.
+#
+# .. code::
+#
+#    > ffmpeg -f dshow -list_devices true -i dummy
+#    ...
+#    [dshow @ 000001adcabb02c0] DirectShow video devices (some may be both video and audio devices)
+#    [dshow @ 000001adcabb02c0]  "TOSHIBA Web Camera - FHD"
+#    [dshow @ 000001adcabb02c0]     Alternative name "@device_pnp_\\?\usb#vid_10f1&pid_1a42&mi_00#7&27d916e6&0&0000#{65e8773d-8f56-11d0-a3b9-00a0c9223196}\global"
+#    [dshow @ 000001adcabb02c0] DirectShow audio devices
+#    [dshow @ 000001adcabb02c0]  "... (Realtek High Definition Audio)"
+#    [dshow @ 000001adcabb02c0]     Alternative name "@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\wave_{BF2B8AE1-10B8-4CA4-A0DC-D02E18A56177}"
+#
+# In the above case, the following value can be used to stream from microphone.
+#
+# .. code::
+#
+#    Streamer(
+#        src = "audio=@device_cm_{33D9A762-90C8-11D0-BD43-00A0C911CE86}\wave_{BF2B8AE1-10B8-4CA4-A0DC-D02E18A56177}",
+#        format = "dshow",
+#    )
+#
+
+######################################################################
+# 3. Data acquisition
+# -------------------
+#
+# Streaming audio from microphone requires to properly time acquiring
+# data form hardware. Failing to do so indtorduces discontinuity in
+# data stream.
+#
+# For this reason, we will run the data acquisition in a subprocess.
+#
+# Firstly, we create a helper function that encupsulates the whole
+# process executed in the subprocess.
+#
+# This function initializes the streaming API, acquire data then
+# put it in a queue, which the main process is watching.
+#
+
+import torch
+import torchaudio
+
+
+# The data acquisition process will stop after this number of steps.
+# This eliminates the need of process synchronization and makes this
+# tutorial simple.
+NUM_ITER = 100
+
+
+def stream(q, format, src, segment_length, sample_rate):
+    from torchaudio.prototype.io import Streamer
+
+    print("Building Streamer...")
+    streamer = Streamer(src, format=format)
+    streamer.add_basic_audio_stream(frames_per_chunk=segment_length, sample_rate=sample_rate)
+
+    print(streamer.get_src_stream_info(0))
+    print(streamer.get_out_stream_info(0))
+
+    print("Streaming...")
+    print()
+    stream_iterator = streamer.stream(timeout=-1, backoff=1.0)
+    for _ in range(NUM_ITER):
+        (chunk,) = next(stream_iterator)
+        q.put(chunk)
+
+
+######################################################################
+#
+# The notable difference from the non-device streaming is that,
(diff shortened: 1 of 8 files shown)


examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,166 @@
...
+#
+
+with torch.no_grad():
+    for (chunk,) in stream_iterator:
+        segment = cacher(chunk).T[0]
+        features, length = feature_extractor(segment)
+        hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
+        hypothesis = hypos[0]
+        transcript = token_processor(hypothesis.tokens, lstrip=False)
+        print(transcript, end="", flush=True)
```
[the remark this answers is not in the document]
mthrok: This file does not have `_tutorial` suffix, so Sphinx-gallery does not execute it.
Note that this tutorial requires hardware device, such as microphone, so it is not possible to run inside of CI, which is a VM inside datacenter. Also this tutorial does finish either.

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,166 @@
+"""
+Device ASR with Emformer RNN-T
+==============================
+
+**Author** : `Moto Hira <moto@fb.com>`__, `Jeff Hwang <jeffhwang@fb.com>`__.
+
+This tutorial shows how to use Emformer RNN-T and streaming API
+to perform device speech recognition.
```
hwangjeff, line 8: maybe something like the following would be clearer:

```suggestion
to perform speech recognition on a streaming device input.
```

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,166 @@
...
+
+print(torch.__version__)
+print(torchaudio.__version__)
+
+######################################################################
+# 3. Construct the pipeline
+# -------------------------
+#
+# First, we construct the ASR inference pipeline and fetch the parameters
+# for configuring the stream. For the detail, please refer to
```
hwangjeff, line 48:
```suggestion
# for configuring the stream. For details, please refer to
```

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def __call__(self, chunk: torch.Tensor):
+        right_context = chunk[: self.right_context_size, :]
+        chunk_with_context = torch.cat((self.chunk, right_context))
+        self.chunk = chunk
+        return chunk_with_context
+
+
+######################################################################
+# 4. Configuring mirophone stream
```
hwangjeff, line 95:
```suggestion
# 4. Configuring microphone stream
```

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,166 @@
...
+#
+# Finally, we run the recognition.
+#
+# First, we initialize the stream iterator, context cacher, and
+# state and hypothesis that are used by decoder to carry over the
+# decoding state between inference calls.
+#
+
+stream_iterator = streamer.stream(retry=1)
+cacher = ContextCacher(next(stream_iterator)[0], right_context_size)
```
hwangjeff, line 150: one observation: we read two chunks before running inference on the first chunk, but we actually only need the first chunk plus the right context. the former adds a delay (constant offset) of `chunk length - right context length` to the transcriptions, which in this case is 120ms and would be amplified as chunk length increases

ideally, we'd initialize `ContextCacher` with the first `right context length` amount of data in the stream, and then proceed with processing a chunk on each call as is currently done

torchaudio/prototype/io/streamer.py
```diff
@@ -479,19 +488,30 @@ def fill_buffer(self) -> int:
...
+                return code
         return 0
 
-    def stream(self) -> Iterator[Tuple[Optional[torch.Tensor]]]:
-        """Return an iterator that generates output tensors"""
+    def stream(self, retry: Optional[float] = 10) -> Iterator[Tuple[Optional[torch.Tensor]]]:
+        """Return an iterator that generates output tensors
+
+        Arguments:
+            retry (float or None, optional): See :py:func:`~Streamer.process_packet`.
```
hwangjeff, line 500:
```suggestion
            retry (float or None, optional): See :py:func:`~Streamer.process_packet`. (Default: ``10.0``)
```

torchaudio/prototype/io/streamer.py
```diff
@@ -479,19 +488,30 @@ def fill_buffer(self) -> int:
...
-                if code != 0:
-                    return code
+            code = self.process_packet(retry)
+            if code != 0:
+                return code
         return 0
 
-    def stream(self) -> Iterator[Tuple[Optional[torch.Tensor]]]:
-        """Return an iterator that generates output tensors"""
+    def stream(self, retry: Optional[float] = 10) -> Iterator[Tuple[Optional[torch.Tensor]]]:
```
hwangjeff, line 496:
```suggestion
    def stream(self, retry: Optional[float] = 10.0) -> Iterator[Tuple[Optional[torch.Tensor]]]:
```

torchaudio/prototype/io/streamer.py
```diff
@@ -420,11 +420,11 @@ def remove_stream(self, i: int):
         """
         torch.ops.torchaudio.ffmpeg_streamer_remove_stream(self._s, i)
 
-    def process_packet(self) -> int:
+    def process_packet(self, retry: Optional[float] = 10) -> int:
```
hwangjeff, line 423:
```suggestion
    def process_packet(self, retry: Optional[float] = 10.0) -> int:
```

torchaudio/prototype/io/streamer.py
```diff
@@ -433,6 +433,15 @@ def process_packet(self) -> int:
...
         flush the pending frames.
 
+        Args:
+            retry (float or None, optional): When using a media device such as
+                a microphone, there are cases where the underlying buffer
+                is not ready. Calling this function in such cases would cause the
+                system to report `EAGAIN (resource temporarily unavailable)`.
+                If the value of ``retry`` is ``None``, then the function raises
+                an exception. Otherwise the function keeps retrying after
+                the given amount of time (in milliseconds). Default: ``10``.
```
hwangjeff, line 443:
```suggestion
                the given amount of time (in milliseconds). Default: ``10.0``.
```

torchaudio/prototype/io/streamer.py
```diff
@@ -479,19 +488,30 @@ def fill_buffer(self) -> int:
...
-    def stream(self) -> Iterator[Tuple[Optional[torch.Tensor]]]:
-        """Return an iterator that generates output tensors"""
+    def stream(self, retry: Optional[float] = 10) -> Iterator[Tuple[Optional[torch.Tensor]]]:
+        """Return an iterator that generates output tensors
+
+        Arguments:
+            retry (float or None, optional): See :py:func:`~Streamer.process_packet`.
+
+        Returns:
+            Iterator[Tuple[Optional[torch.Tensor]]]:
```
hwangjeff, line 503: is it the case that each tuple returned by the function contains just one element that's either a tensor or `None`?

mthrok: Yeah, it starts from the list of tuple of tensors, each corresponds to one output stream, then as output streams get exhausted, they start to return None. It will stop when all the output streams get exhausted.

hwangjeff: ah i see. then, we should amend the type hint
```suggestion
            Iterator[Tuple[Optional[torch.Tensor], ...]]:
```

examples/tutorials/device_asr.py
mthrok: Very good point. In fact, I was a bit confused about the fact it is called "right context", as opposed to "left context". "left" would infer that it is a carry over from the past. This terminology had me think in this way.

I think we can initialize the cacher with zero Tensors of `right context length`, then start feeding, append and cache the latest. Let me try this on the existing tutorial.

nateanl (approved):
stamp

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,330 @@
...
+# to the documentation of
+# :py:meth:`~torchaudio.prototype.io.Streamer.stream` method.
+#
+# .. note::
+#
+#    The proper value of ``backoff`` depends on the system configuration.
+#    One way to see if ``backoff`` value is appropriate is to save the
+#    series of acquired chunks as a continuous audio and listen to it.
+#    If ``backoff`` value is too large, then the data stream is discontinuous.
+#    The resulting audio sounds sped up.
```
hwangjeff, line 178: is this because the streamer buffer wouldn't be flushed quickly enough due to the excessively large backoff? so buffer contents end up getting overwritten, and therefore sections of the data stream end up going missing?

torchaudio/prototype/io/streamer.py
```diff
@@ -433,6 +433,39 @@ def process_packet(self) -> int:
...
+
+                Note:
+
+                    The retry behavior is applicable only when the reason is the
+                    unavailable resource. It is not invoked if the reason of failure is
+                    other.
+
+            backoff (float, optional): Time to wait before retrying in milli seconds.
+
+                This optioin is effective only when `timeout` is effective. (not ``None``)
```
hwangjeff, line 464: "option"

torchaudio/prototype/io/streamer.py
```diff
@@ -433,6 +433,39 @@ def process_packet(self) -> int:
...
+
+                When `timeout` is effective, this `backoff` controls how long the function
+                should wait before retry-ing. Default: ``10.0``.
```
hwangjeff, line 467: "retrying"

torchaudio/prototype/io/streamer.py
```diff
@@ -479,19 +512,36 @@ def fill_buffer(self) -> int:
...
+        Arguments:
+            timeout (float or None, optional): See
+                :py:func:`~Streamer.process_packet`. (Default: ``None``)
+
+            backoff (float, optional): See
+                :py:func:`~Streamer.process_packet`. (Default: ``10.0``)
+
+        Returns:
+            Iterator[Tuple[Optional[torch.Tensor], ...]]:
+                Iterator that yields a tuple of chunks that correpond to the output
```
hwangjeff, line 534: "correspond"

torchaudio/csrc/ffmpeg/streamer.cpp
```diff
@@ -229,6 +231,45 @@ int Streamer::process_packet() {
...
+    // https://github.com/FFmpeg/FFmpeg/blob/b0f8dbb0cacc45a19f18c043afc706d7d26bef74/fftools/ffmpeg.c#L3952
+    // https://github.com/FFmpeg/FFmpeg/blob/b0f8dbb0cacc45a19f18c043afc706d7d26bef74/fftools/ffmpeg.c#L4542
+    //
+    // But it does not seem to sleep when running in single thread.
+    // Empirically we observed that the streaming result is worse with sleep.
+    // busy-waiting is not a recommended way to resolve this, but after simple
+    // testing, there wasn't a noticible difference in CPU utility. So we do not
+    // sleep here.
+    //
+    std::this_thread::sleep_for(sleep);
```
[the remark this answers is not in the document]
mthrok: Not anymore. I forgot to update.

Originally, I did not want to introduce `backoff` parameter, so I tested the difference of busy-waiting and back off waiting. CPU usage was 30% different. So I introduced the sleep and put this in a separate process in the example. Thanks for catching. I will remove the comment.

examples/tutorials/device_asr.py
mthrok: Yes, the hardware buffer will be refilled at a rate faster than the torchaudio code acquiring the data.
The data client code receive miss portions, which would sound like skipped, thus sped up.

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,330 @@
...
+# Streaming audio from microphone requires to properly time acquiring
+# data form hardware. Failing to do so indtorduces discontinuity in
+# data stream.
+#
+# For this reason, we will run the data acquisition in a subprocess.
+#
+# Firstly, we create a helper function that encupsulates the whole
+# process executed in the subprocess.
+#
+# This function initializes the streaming API, acquire data then
```
carolineechen, line 120:
```suggestion
# This function initializes the streaming API, acquires data then
```

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,330 @@
...
+# put it in a queue, which the main process is watching.
```
carolineechen, line 121:
```suggestion
# puts it in a queue, which the main process is watching.
```

examples/tutorials/device_asr.py
```diff
@@ -0,0 +1,330 @@
...
+#    this increases the CPU consumption.
+#
+
+######################################################################
+# 4. Building inference pipeline
+# ------------------------------
+#
+# The next step is to create components required for inference.
+#
+# The is the same process as
```
carolineechen, line 190:
```suggestion
This is the same process as
```

PR #2203, opened 2022-02-04
Refactor pipeline_demo.py to support variant EMFORMER_RNNT bundles

nateanl:
We refactored the demo script that can apply RNNT decoding using both `torchaudio.pipelines.EMFORMER_RNNT_BASE_LIBRISPEECH` and `torchaudio.prototype.pipelines.EMFORMER_RNNT_BASE_TEDLIUM3` in both streaming and non-streaming mode. (The first hypothesis prediction is streaming and the second one is non-streaming).

We convert each token id sequence to word pieces and then manually join the word pieces. This allows us to preserve leading whitespaces on output strings and therefore account for word breaks and continuations across token processor invocations, which is particularly useful when performing streaming ASR.

https://user-images.githubusercontent.com/8653221/153627956-f0806f18-3c1c-44df-ac07-ec2def58a0cf.mov


nateanl:
> code-wise what is the difference from the equivalent one from [librispeech_emformer_rnnt](https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_emformer_rnnt)?

It's actually the same, only the pipeline-related part is different.

hwangjeff (commented):
the readme also references `pipeline_demo.py` — can you update it to be consistent with the changes here?

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,114 @@
...
+            features, length = feature_extractor(waveform)
+            hypos = decoder(features, length, 10)
+        print(token_processor(hypos[0].tokens))
+        print()
+
+
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument("--model_type", type=str, choices=[MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3], required=True)
+    parser.add_argument(
```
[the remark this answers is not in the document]
nateanl: addressed in https://github.com/pytorch/audio/pull/2203/commits/a4f6a06071de37b7d3c3d00ccc93911209e39577

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,114 @@
...
+        token_processor = EMFORMER_RNNT_BASE_TEDLIUM3.get_token_processor()
+        feature_extractor = EMFORMER_RNNT_BASE_TEDLIUM3.get_feature_extractor()
+        streaming_feature_extractor = EMFORMER_RNNT_BASE_TEDLIUM3.get_streaming_feature_extractor()
+        hop_length = EMFORMER_RNNT_BASE_TEDLIUM3.hop_length
+        num_samples_segment = EMFORMER_RNNT_BASE_TEDLIUM3.segment_length * hop_length
+        num_samples_segment_right_context = (
+            num_samples_segment + EMFORMER_RNNT_BASE_TEDLIUM3.right_context_length * hop_length
+        )
+    else:
+        raise ValueError(f"Encountered unsupported model type {args.model_type}.")
```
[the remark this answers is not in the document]
nateanl: addressed in https://github.com/pytorch/audio/pull/2203/commits/a4f6a06071de37b7d3c3d00ccc93911209e39577

hwangjeff (commented):
in the video, why do the word pieces for each streaming transcription show up all at once?

nateanl:
>  why do the word pieces for each streaming transcription show up all at once?

I guess it's because it's run on AWS cluster, there may be some delay when printing to the screen.

hwangjeff:
> I guess it's because it's run on AWS cluster, there may be some delay when printing to the screen.

in that case, for the screen capture, can you run the script locally? that way, we can clearly show users what we mean by streaming asr and how responsive it is using the bundles (example: #2192)

nateanl:
Sure. The demo video has been updated @hwangjeff

hwangjeff (approved):
looks good — thanks!

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,94 @@
...
+import torchaudio
+from common import MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3
+from torchaudio.pipelines import EMFORMER_RNNT_BASE_LIBRISPEECH
+from torchaudio.prototype.pipelines import EMFORMER_RNNT_BASE_TEDLIUM3
+
+
+logger = logging.getLogger()
+
+
+def get_dataset(model_type, dataset_path):
```
mthrok, line 15: Instead of repeating the key validation here and there, but these options in a dictionary, sthen pass the keys to `choices` arguments of argpsrse. So that the closed set of available options are defined once and only once.

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,94 @@
...
+        # Non-streaming decode.
+        with torch.no_grad():
+            features, length = feature_extractor(waveform)
+            hypos = decoder(features, length, 10)
+        print(token_processor(hypos[0].tokens))
+        print()
+
+
+def parse_args():
+    parser = ArgumentParser()
```
mthrok, line 69: Please add module level docstring that describes the gist of this script, then pass it to the help description so that it's easy to see what this script does.

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -0,0 +1,94 @@
+import logging
```
mthrok, line 1: Please add shebang line.

PR #2208, opened 2022-02-09
Fix librosa calls

hwangjeff:
Yesterday's release of librosa 0.9.0 made args keyword-only and changed default padding from "reflect" to "zero" for some functions. This PR adjusts callsites in our tutorials and tests accordingly.


mthrok (approved):

hwangjeff:
Actually, there's a bunch of references in our tests too — will just add to this PR.

PR #2212, opened 2022-02-09
Refactor Emformer RNNT recipes

hwangjeff:
Consolidates LibriSpeech and TED-LIUM Release 3 Emformer RNN-T training recipes in a single directory.


examples/asr/emformer_rnnt/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+[`train.py`](./train.py) trains an Emformer RNN-T model using PyTorch Lightning. Note that the script expects users to have access to GPU nodes for training and provide paths to datasets and the SentencePiece model to be used to encode targets. The script also expects a file (--global_stats_path) that contains training set feature statistics; this file can be generated via [`global_stats.py`](./global_stats.py).
+
+### Evaluation
+
+[`eval.py`](./eval.py) evaluates a trained Emformer RNN-T model on a given dataset.
+
+## Model Types
+
+Currently, we have training recipes for the LibriSpeech and TED-LIUM Release 3 datasets.
```
mthrok, line 17: Could you have a brief description on what’s the difference between these models? I know vocab size is one. Is there any other?

mthrok (approved):
I think that demo script can be moved one directory up so that it can handle both librispeech model and tedlium model. It doesn’t have to happen in this PR. @nateanl wojld you like to update #2203 after this one?

examples/asr/emformer_rnnt/README.md
hwangjeff: yeah, that's the only difference in the model architecture. the recipes themselves have some differences that are specific to the datasets — this should be apparent in the lightning modules

nateanl: Sure, will do.

examples/asr/emformer_rnnt/tedlium3/lightning.py
```diff
@@ -46,20 +29,21 @@ def __init__(self, base_dataset, max_token_limit):
...
+        idx_target_lengths = [(idx, length) for idx, length in idx_target_lengths if length != -1]
 
         assert len(idx_target_lengths) > 0
 
         idx_target_lengths = sorted(idx_target_lengths, key=lambda x: x[1])
 
         assert max_token_limit >= idx_target_lengths[-1][1]
 
-        self.batches = _batch_by_token_count(idx_target_lengths, max_token_limit)
+        self.batches = batch_by_token_count(idx_target_lengths, max_token_limit)[:100]
```
nateanl, line 40: Why we choose the first 100 batches here?

nateanl:
Do we want to keep the ``train_sentencepiece.py`` script in each sub directory? Otherwise they may ignore the important details that may impact the model performance.

examples/asr/emformer_rnnt/tedlium3/lightning.py
hwangjeff: good call — this was meant to speed up the training epoch for testing purposes only and should be removed

mthrok:
> Do we want to keep the `train_sentencepiece.py` script in each sub directory? Otherwise they may ignore the important details that may impact the model performance.

@nateanl The difference between librispeech model and tedluim model should be clearly outlined in the top level README. It can be as simple as `the Sentence Piece model also needs to be different, for the detail please refer to the READMe in respective directory` and delegate the detail to the sub directory READMEs.

examples/asr/emformer_rnnt/train.py
```diff
@@ -29,65 +31,68 @@ def run_train(args):
...
+def parse_args():
+    parser = ArgumentParser()
+    parser.add_argument("--model_type", type=str, choices=[MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3], required=True)
     parser.add_argument(
-        "--global-stats-path",
+        "--global_stats_path",
         default=pathlib.Path("global_stats.json"),
         type=pathlib.Path,
         help="Path to JSON file containing feature means and stddevs.",
+        required=True,
```
csukuangfj, line 71: As it has a default value, is it really necessary to specify `required=True` here?
Also, the default value is not correct as that file is inside the recipe dir.

I would suggest removing the default value.

PR #2216, opened 2022-02-11
Add unit tests for Emformer RNN-T LibriSpeech recipe

hwangjeff:
Adds unit tests for Emformer RNN-T LibriSpeech recipe. Also makes changes to recipe to resolve errors with pickling lambda functions in Windows.

### DIFF
--- a/examples/asr/emformer_rnnt/global_stats.py
+++ b/examples/asr/emformer_rnnt/global_stats.py
@@ -11,7 +11,7 @@ from argparse import ArgumentParser, RawTextHelpFormatter
 
 import torch
 import torchaudio
-from common import GAIN, MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3, piecewise_linear_log, spectrogram_transform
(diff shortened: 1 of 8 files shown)


examples/asr/emformer_rnnt/global_stats.py
```diff
@@ -42,7 +42,7 @@ def generate_statistics(samples):
 
     for idx, sample in enumerate(samples):
         mel_spec = spectrogram_transform(sample[0].squeeze()).transpose(1, 0)
-        scaled_mel_spec = piecewise_linear_log(mel_spec * GAIN)
+        scaled_mel_spec = piecewise_linear_log(mel_spec)
```
nateanl, line 45: If we remove the GAIN computation, does it mean the `global_stats.json` needs to be updated as well?

hwangjeff: the scaling by `GAIN` happens within `piecewise_linear_log`, so no

nateanl: Got it.

PR #2218, opened 2022-02-11
Add SentencePiece model training script for LibriSpeech Emformer RNN-T

hwangjeff:
Adds SentencePiece model training script for LibriSpeech Emformer RNN-T example recipe; updates readme with references.


nateanl (commented):
LGTM.

PR #2219, opened 2022-02-11
Add training recipe for Emformer RNNT trained on MuST-C release v2.0 dataset

nateanl:
- Add a MUSTC dataset under examples
- Add a lightning module for MuST-C dataset
- Refactor `train.py`, `eval.py`, and `global_stats.py` scripts

### DIFF
--- a/examples/asr/emformer_rnnt/README.md
+++ b/examples/asr/emformer_rnnt/README.md
@@ -25,12 +25,12 @@ Currently, we have training recipes for the LibriSpeech and TED-LIUM Release 3 d
 
 Sample SLURM command for training:
 ```
-srun --cpus-per-task=12 --gpus-per-node=8 -N 4 --ntasks-per-node=8 python train.py --model_type librispeech --exp_dir ./experiments --dataset_path ./datasets/librispeech --global_stats_path ./global_stats.json --sp_model_path ./spm_bpe_4096.model
+srun --cpus-per-task=12 --gpus-per-node=8 -N 4 --ntasks-per-node=8 python train.py --model-type librispeech --exp-dir ./experiments --dataset-path ./datasets/librispeech --global-stats-path ./global_stats.json --sp-model-path ./spm_bpe_4096.model
 ```
 
 Sample SLURM command for evaluation:
 ```
-srun python eval.py --model_type librispeech --checkpoint_path ./experiments/checkpoints/epoch=119-step=208079.ckpt --dataset_path ./datasets/librispeech  --sp_model_path ./spm_bpe_4096.model --use_cuda
+srun python eval.py --model_type librispeech --checkpoint_path ./experiments/checkpoints/epoch=119-step=208079.ckpt --dataset-path ./datasets/librispeech  --sp-model-path ./spm_bpe_4096.model --use-cuda
 ```
 
 Using the sample training command above along with a SentencePiece model trained on LibriSpeech with vocab size 4096 and type bpe, [`train.py`](./train.py) produces a model with 76.7M parameters (307MB) that achieves an WER of 0.0456 when evaluated on test-clean with [`eval.py`](./eval.py).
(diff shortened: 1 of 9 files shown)


examples/asr/emformer_rnnt/common.py
```diff
@@ -27,11 +28,13 @@ def piecewise_linear_log(x):
...
 
 
-def batch_by_token_count(idx_target_lengths, token_limit):
+def batch_by_token_count(idx_target_lengths, token_limit, max_len=None):
     batches = []
     current_batch = []
     current_token_count = 0
     for idx, target_length in idx_target_lengths:
+        if max_len is not None and target_length > max_len:
+            continue
```
hwangjeff, line 37: given the responsibilities the name `batch_by_token_count` implies, this logic would be better placed elsewhere. how about doing it in `CustomDataset`?

examples/asr/emformer_rnnt/README.md
```diff
@@ -63,3 +63,25 @@ The table below contains WER results for dev and test subsets of TED-LIUM releas
 | test        |       0.098  |
 
 [`tedlium3/eval_pipeline.py`](./tedlium3/eval_pipeline.py) evaluates the pre-trained `EMFORMER_RNNT_BASE_TEDLIUM3` bundle on the dev and test sets of TED-LIUM release 3. Running the script should produce WER results that are identical to those in the above table.
+
+### MuST-C release v2.0
+
+The MuST-C model is configured with a vocabulary size of 500. Consequently, the MuST-C model's last linear layer in the joiner has an output dimension of 501 (500 + 1 to account for the blank symbol). Different from the above two models, we keep the transcriptions cased and punctuated when training the SentencePiece model.
```
hwangjeff, line 72: the current description seems to suggest that we chose to not keep transcriptions cased and punctuated for the other two models, when in actuality we didn't have a choice because the datasets themselves aren't cased and punctuated. we should clarify this, e.g. something like

```suggestion
The MuST-C model is configured with a vocabulary size of 500. Consequently, the MuST-C model's last linear layer in the joiner has an output dimension of 501 (500 + 1 to account for the blank symbol). In contrast to those of the datasets for the above two models, MuST-C's transcripts are cased and punctuated; we preserve the casing and punctuation when training the SentencePiece model.
```

examples/asr/emformer_rnnt/eval.py
```diff
@@ -28,7 +28,25 @@ def run_eval(model):
             total_length += len(actual.split())
             if idx % 100 == 0:
                 logger.info(f"Processed elem {idx}; WER: {total_edit_distance / total_length}")
-    logger.info(f"Final WER: {total_edit_distance / total_length}")
+    logger.info(f"Final WER for {subset} set: {total_edit_distance / total_length}")
+
+
+def run_eval(model, model_type):
+    if model_type == MODEL_TYPE_LIBRISPEECH:
```
hwangjeff, line 35: better to add a default case with exception

examples/asr/emformer_rnnt/mustc/train_spm.py
```diff
@@ -0,0 +1,54 @@
...
+    args = _parse_args()
+    _init_logger(args.debug)
+    text_path = args.mustc_path / "en-de/data/train/txt/train.en"
+
+    spm.SentencePieceTrainer.train(
+        input=text_path,
+        vocab_size=3000,
+        model_prefix="spm_bpe_3000",
+        model_type="bpe",
+        input_sentence_size=100000000,
```
hwangjeff, line 46: intent might be clearer by providing -1 (unless there's special significance around 100000000)

examples/asr/emformer_rnnt/common.py
nateanl: sounds good, will move it.

examples/asr/emformer_rnnt/eval.py
```diff
@@ -46,36 +66,45 @@ def get_lightning_module(args):
...
 
 def parse_args():
     parser = ArgumentParser()
-    parser.add_argument("--model_type", type=str, choices=[MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3], required=True)
     parser.add_argument(
-        "--checkpoint_path",
+        "--model-type", type=str, choices=[MODEL_TYPE_LIBRISPEECH, MODEL_TYPE_TEDLIUM3, MODEL_TYPE_MUSTC], required=True
+    )
+    parser.add_argument(
+        "--checkpoint-path",
```
hwangjeff, line 86: while you're at it, could you also add `required=True` to the args here?

hwangjeff (approved):
lg. as a follow up, can you add unit tests for the lightning module?

mthrok (commented):
The validation of `model_type` values are repeated across the scripts. Can you use dictionary?

PR #2226, opened 2022-02-15
Update online ASR tutorial

mthrok:
https://554729-90321822-gh.circle-artifacts.com/0/docs/tutorials/online_asr_tutorial.html

1. Add figure to explain the caching
2. Fix the initialization of stream iterator

### DIFF
--- a/examples/tutorials/online_asr_tutorial.py
+++ b/examples/tutorials/online_asr_tutorial.py
@@ -72,9 +72,9 @@ token_processor = bundle.get_token_processor()
 
 ######################################################################
 # Streaming inference works on input data with overlap.
-# Emformer RNN-T model treats the newer portion of the input data,
-# denoted as "right context", a preview of future context.
-# In the subsequent inference call, the model expects the main segment
+# Emformer RNN-T model treats the newest portion of the input data
+# as the "right context" — a preview of future context.
+# In each inference call, the model expects the main segment
 # to start from this right context from the previous inference call.
 # The following figure illustrates this.
 #
@@ -120,9 +120,9 @@ print(streamer.get_out_stream_info(0))
 
 ######################################################################
 # As previously explained, Emformer RNN-T model expects input data with
-# overlaps, however, `Streamer` iterates the source media without overlap,
+# overlaps; however, `Streamer` iterates the source media without overlap,
 # so we make a helper structure that caches a part of input data from
-# `Streamer` as right context, then append it to the next input data from
+# `Streamer` as right context and then appends it to the next input data from
 # `Streamer`.
 #
 # The following figure illustrates this.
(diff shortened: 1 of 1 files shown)


examples/tutorials/online_asr_tutorial.py
```diff
@@ -72,7 +72,11 @@
 
 ######################################################################
 # Streaming inference works on input data with overlap.
-# Emformer RNN-T expects right context like the following.
+# Emformer RNN-T model treats the newer portion of the input data,
+# denoted as "right context", a preview of future context.
```
hwangjeff, line 76:
```suggestion
# Emformer RNN-T model treats the newest portion of the input data
# as the "right context" — a preview of future context.
```

examples/tutorials/online_asr_tutorial.py
```diff
@@ -72,7 +72,11 @@
...
+# In the subsequent inference call, the model expects the main segment
```
hwangjeff, line 77:
```suggestion
# In each inference call, the model expects the main segment
```

examples/tutorials/online_asr_tutorial.py
```diff
@@ -115,14 +119,20 @@
 print(streamer.get_out_stream_info(0))
 
 ######################################################################
-# `Streamer` iterate the source media without overlap, so we make a
-# helper structure that caches a chunk and return it with right context
-# appended when the next chunk is given.
+# As previously explained, Emformer RNN-T model expects input data with
+# overlaps, however, `Streamer` iterates the source media without overlap,
```
hwangjeff, line 123:
```suggestion
# overlaps; however, `Streamer` iterates the source media without overlap,
```

examples/tutorials/online_asr_tutorial.py
```diff
@@ -115,14 +119,20 @@
...
+# so we make a helper structure that caches a part of input data from
+# `Streamer` as right context, then append it to the next input data from
```
hwangjeff, line 125:
```suggestion
# `Streamer` as right context and then appends it to the next input data from
```

hwangjeff (approved):
just some nits — looks good

side thought: do you think it'd be helpful to be explicit about how the quality of the transcriptions is as expected partly because the model is trained on librispeech but is used to perform inference on non-librispeech data?

mthrok:
> side thought: do you think it'd be helpful to be explicit about how the quality of the transcriptions is as expected partly because the model is trained on librispeech but is used to perform inference on non-librispeech data?

How would you phrase it?

I gave some thoughts at the very beginning, but if we mention that the sample is out-of-domain data, then the narrative is "it's not that good and the reason is the sample is out-of-domain". So I decided not to mention.

PR #2227, opened 2022-02-15
Add psd method to torchaudio.functional

nateanl:
This PR adds ``psd`` method to ``torchaudio.functional``.
It computes the power spectral density (PSD) matrix of the complex-valued spectrum.
The method also supports normalization of Time-Frequency mask.


torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,38 @@ def rnnt_loss(
         return costs.sum()
 
     return costs
+
+
+def compute_power_spectral_density_matrix(
```
hwangjeff, line 1637: the name seems a little verbose. would something like `psd` suffice? would also be consistent with the existing transform `PSD`

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,38 @@ def rnnt_loss(
...
+    eps: float = 1e-10,
+) -> Tensor:
+    """Compute cross-channel power spectral density (PSD) matrix.
+    Args:
+        specgram (Tensor): Multi-channel complex-valued spectrum.
+            Tensor of dimension `(..., channel, freq, time)`
+        mask (Tensor or None, optional): Real-valued Time-Frequency mask
+            for normalization. Tensor of dimension `(..., freq, time)`
+            (Default: ``None``)
+        normalize (bool, optional): whether normalize the mask along the time dimension.
```
hwangjeff, line 1650:
```suggestion
        normalize (bool, optional): whether to normalize the mask along the time dimension.
```

also, default arg

and a couple more:
- our spectrogram functions accept parameter "normalized" — should we adjust here for consistency's sake?
- we already expect users to construct masks outside the function, and normalization doesn't depend on anything else inside the function. wondering how reasonable it'd be to remove `normalize` entirely

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,38 @@ def rnnt_loss(
...
+    specgram: Tensor,
+    mask: Optional[Tensor] = None,
+    normalize: bool = True,
+    eps: float = 1e-10,
+) -> Tensor:
+    """Compute cross-channel power spectral density (PSD) matrix.
+    Args:
+        specgram (Tensor): Multi-channel complex-valued spectrum.
+            Tensor of dimension `(..., channel, freq, time)`
+        mask (Tensor or None, optional): Real-valued Time-Frequency mask
```
hwangjeff, line 1647:
```suggestion
        mask (Tensor or None, optional): Real-valued time-frequency mask
```

torchaudio/functional/functional.py
nateanl: This is to align the [get_power_spectral_density_matrix](https://github.com/espnet/espnet/blob/master/espnet2/enh/layers/beamformer.py#L25) method in ESPNet and [get_power_spectral_density_matrix](https://github.com/fgnt/pb_bss/blob/master/pb_bss/extraction/beamformer.py#L59) in pb_bss. `compute` sounds more reasonable than `get`.

torchaudio/functional/functional.py
nateanl: That's a good question. Here `normalize` is intended to normalize the `psd` by the sum of the mask values on the time dimension.

Suppose we have an estimated mask by some algorithm, users may want to compute the psd matrix without modifying the mask, like normalizing it manually.

spectrogram has to use `normalized` because `torch.stft` requires such name. In `torch.load` we also have `normalize` argument, so I think it's fine.

torchaudio/functional/functional.py
hwangjeff: we already appear to have a convention of names of functionals being the snake-case of the names of the corresponding transforms, e.g. `torchaudio.functional.inverse_spectrogram` and `torchaudio.transforms.InverseSpectrogram`, and `torchaudio.transforms.PSD` already exists. why would we prefer diverging to align with espnet?

nateanl: `psd` is fine for me.
For the other methods like `compute_mvdr_weights_souden`, there may not be a corresponding module, as the new beamforming module in release 0.12 will use both `compute_mvdr_weights_xx` and `apply_beamforming` to obtain the enhanced spectrogram directly. Do those method names LGTY?

hwangjeff: with a couple of exceptions, it seems like, if no corresponding concise verb exists, functionals are named as nouns. if we want to follow that convention, then `mvdr_weights_xx` would be more appropriate. also, adding `compute_` seems a little redundant since it's implicit that every functional computes something

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -582,6 +583,35 @@ def test_rnnt_loss_costs_and_gradients_random_data_with_numpy_fp32(self):
             ref_costs, ref_gradients = rnnt_utils.compute_with_numpy_transducer(data=data)
             self._test_costs_and_gradients(data=data, ref_costs=ref_costs, ref_gradients=ref_gradients)
 
+    def test_psd(self):
```
mthrok, line 586: Can you add a docstring?

mthrok (approved):

PR #2230, opened 2022-02-15
Add rtf_evd method to torchaudio.functional

nateanl:
This PR adds `rtf_evd` method to `torchaudio.functional`.
The method computes the relative transfer function (RTF) or the steering vector by eigenvalue decomposition.
The input argument is the power spectral density (PSD) matrix of the target speech.

### DIFF
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -365,3 +365,12 @@ class TestFunctional(common_utils.TorchaudioTestCase):
         reference_channel = torch.zeros(batch_size, channel)
         reference_channel[..., 0].fill_(1)
         self.assert_batch_consistency(F.mvdr_weights_rtf, (rtf, psd_noise, reference_channel))
+
+    def test_rtf_evd(self):
+        torch.random.manual_seed(2434)
+        batch_size = 2
+        channel = 4
+        n_fft_bin = 5
+        spectrum = torch.rand(batch_size, n_fft_bin, channel, dtype=torch.cfloat)
+        psd = torch.einsum("...c,...d->...cd", spectrum, spectrum.conj())
+        self.assert_batch_consistency(F.rtf_evd, (psd,))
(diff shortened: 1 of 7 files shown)


test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -241,3 +241,12 @@ def test_filtfilt(self):
...
         self.assertEqual(batchwise_output, itemwise_output)
+
+    def test_compute_rtf_evd(self):
+        torch.random.manual_seed(2434)
+        channel = 4
+        spectrum = torch.rand(2, 5, channel, dtype=torch.cfloat)
+        psd = torch.einsum("...c,...d->...cd", spectrum, spectrum.conj())
+        batchwise_output = F.compute_rtf_evd(psd)
+        itemwise_output = torch.stack([F.compute_rtf_evd(psd[i]) for i in range(2)])
+        self.assertEqual(batchwise_output, itemwise_output)
```
mthrok, line 252: Can you use `assert_batch_consistency`?

mthrok (commented):
We should have a test against some reference, so that when the implementation is changed, we know it produces similar values from the original implementation.

If it is hard to hand-compute reference values, we can use NumPy, so that implementation won't change.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -582,6 +583,15 @@ def test_rnnt_loss_costs_and_gradients_random_data_with_numpy_fp32(self):
             ref_costs, ref_gradients = rnnt_utils.compute_with_numpy_transducer(data=data)
             self._test_costs_and_gradients(data=data, ref_costs=ref_costs, ref_gradients=ref_gradients)
 
+    def test_rtf_evd(self):
```
mthrok, line 586: Please add a dosctring about what is the intended tests. So that we have later clues later, in some unexpected investigation of the code.

mthrok (approved):

PR #2231, opened 2022-02-15
Add rtf_power method to torchaudio.functional

nateanl:
This PR adds ``rtf_power`` method to ``torchaudio.functional``.
The method computes the relative transfer function (RTF) or the steering vector by [the power iteration method](https://onlinelibrary.wiley.com/doi/abs/10.1002/zamm.19290090206). 
[This paper](https://arxiv.org/pdf/2011.15003.pdf) describes the power iteration method in English.
The input arguments are the complex-valued power spectral density (PSD) matrix of the target speech, PSD matrix of noise, int or one-hot Tensor to indicate the reference channel, number of iterations, respectively.

### DIFF
--- a/test/torchaudio_unittest/functional/autograd_impl.py
+++ b/test/torchaudio_unittest/functional/autograd_impl.py
@@ -303,6 +303,36 @@ class Autograd(TestBaseMixin):
         reference_channel[..., 0].fill_(1)
         self.assert_grad(F.mvdr_weights_rtf, (rtf, psd_noise, reference_channel))
 
+    @parameterized.expand(
+        [
+            (1,),
+            (3,),
+        ]
+    )
+    def test_rtf_power(self, n_iter):
+        torch.random.manual_seed(2434)
+        channel = 4
+        n_fft_bin = 5
+        psd_speech = torch.rand(n_fft_bin, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(n_fft_bin, channel, channel, dtype=torch.cfloat)
+        self.assert_grad(F.rtf_power, (psd_speech, psd_noise, 0, n_iter))
+
+    @parameterized.expand(
+        [
+            (1,),
+            (3,),
+        ]
+    )
+    def test_rtf_power_with_tensor(self, n_iter):
+        torch.random.manual_seed(2434)
+        channel = 4
+        n_fft_bin = 5
+        psd_speech = torch.rand(n_fft_bin, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(n_fft_bin, channel, channel, dtype=torch.cfloat)
+        reference_channel = torch.zeros(channel)
+        reference_channel[0].fill_(1)
+        self.assert_grad(F.rtf_power, (psd_speech, psd_noise, reference_channel, n_iter))
+
 
 class AutogradFloat32(TestBaseMixin):
     def assert_grad(
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -374,3 +374,44 @@ class TestFunctional(common_utils.TorchaudioTestCase):
         spectrum = torch.rand(batch_size, n_fft_bin, channel, dtype=torch.cfloat)
         psd = torch.einsum("...c,...d->...cd", spectrum, spectrum.conj())
         self.assert_batch_consistency(F.rtf_evd, (psd,))
+
+    @parameterized.expand(
+        [
+            (1,),
+            (3,),
+        ]
+    )
+    def test_rtf_power(self, n_iter):
+        torch.random.manual_seed(2434)
+        channel = 4
+        batch_size = 2
+        n_fft_bin = 10
+        psd_speech = torch.rand(batch_size, n_fft_bin, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(batch_size, n_fft_bin, channel, channel, dtype=torch.cfloat)
+        kwargs = {
+            "reference_channel": 0,
+            "n_iter": n_iter,
+        }
+        func = partial(F.rtf_power, **kwargs)
+        self.assert_batch_consistency(func, (psd_speech, psd_noise))
+
+    @parameterized.expand(
+        [
+            (1,),
+            (3,),
+        ]
+    )
+    def test_rtf_power_with_tensor(self, n_iter):
+        torch.random.manual_seed(2434)
+        channel = 4
+        batch_size = 2
+        n_fft_bin = 10
+        psd_speech = torch.rand(batch_size, n_fft_bin, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(batch_size, n_fft_bin, channel, channel, dtype=torch.cfloat)
+        reference_channel = torch.zeros(batch_size, channel)
+        reference_channel[..., 0].fill_(1)
+        kwargs = {
+            "n_iter": n_iter,
+        }
+        func = partial(F.rtf_power, **kwargs)
+        self.assert_batch_consistency(func, (psd_speech, psd_noise, reference_channel))
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -738,6 +738,63 @@ class Functional(TestBaseMixin):
         rtf_audio = F.rtf_evd(torch.tensor(psd, dtype=self.complex_dtype, device=self.device))
         self.assertEqual(torch.tensor(rtf, dtype=self.complex_dtype, device=self.device), rtf_audio)
 
+    @parameterized.expand(
+        [
+            (1,),
+            (2,),
+            (3,),
+        ]
+    )
+    def test_rtf_power(self, n_iter):
+        """Verify ``F.rtf_power`` method by numpy implementation.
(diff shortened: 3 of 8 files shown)


test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -250,6 +250,13 @@ def test_bandreject_biquad(self, central_freq, Q):
         Q = torch.tensor(Q)
         self.assert_grad(F.bandreject_biquad, (x, sr, central_freq, Q))
 
+    def test_compute_rtf_power(self):
+        torch.random.manual_seed(2434)
+        channel = 4
+        psd_speech = torch.rand(5, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(5, channel, channel, dtype=torch.cfloat)
```
mthrok, line 257: Please define variables for the repeatedly used constants. This helps the constraints among values. Also giving a good variable name helps readers understand the test logic.

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -241,3 +241,12 @@ def test_filtfilt(self):
...
         self.assertEqual(batchwise_output, itemwise_output)
+
+    def test_compute_compute_rtf_power(self):
+        torch.random.manual_seed(2434)
+        channel = 4
+        psd_speech = torch.rand(2, 129, channel, channel, dtype=torch.cfloat)
+        psd_noise = torch.rand(2, 129, channel, channel, dtype=torch.cfloat)
+        batchwise_output = F.compute_rtf_power(psd_speech, psd_noise, 0)
+        itemwise_output = torch.stack([F.compute_rtf_power(psd_speech[i], psd_noise[i], 0) for i in range(2)])
+        self.assertEqual(batchwise_output, itemwise_output)
```
mthrok, line 252: Could you use `assert_batch_consistency` method?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -644,6 +644,16 @@ def func(tensor):
...
 
+    def test_compute_compute_rtf_power(self):
+        def func(_):
+            channel = 4
+            psd_speech = torch.rand(129, channel, channel, dtype=torch.cfloat)
+            psd_noise = torch.rand(129, channel, channel, dtype=torch.cfloat)
+            return F.compute_rtf_power(psd_speech, psd_noise, 0)
+
+        dummy = torch.rand(1, 1)
+        self._assert_consistency(func, dummy)
```
mthrok, line 655: The use of dummy here seems it does not fulfill the purpose of the test.

Can you modify the `_assert_consistency` method so that it can accept multiple arguments, or pack two tensors with tuple?

test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -644,6 +644,16 @@ def func(tensor):
         tensor = torch.view_as_complex(torch.randn(2, 1025, 400, 2))
         self._assert_consistency_complex(func, tensor)
 
+    def test_compute_compute_rtf_power(self):
```
mthrok, line 647: Can you also add the test case where `reference_channel` is Tensor?

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,35 @@ def rnnt_loss(
...
+        psd_n (Tensor): The complex-valued covariance matrix of noise.
+            Tensor of dimension `(..., freq, channel, channel)`
+        reference_channel (int or Tensor): Indicate the reference channel.
+            If the dtype is ``int``, it represent the reference channel index.
+            If the dtype is ``Tensor``, the dimension is `(..., channel)`.
+        iterations (int): number of iterations in power method. (Default: ``3``)
+    Returns:
+        Tensor: the estimated complex-valued RTF of target speech
+            Tensor of dimension `(..., freq, channel)`
+    """
```
mthrok, line 1653: Can you add `assert` statement to reject invalid value of `iteration`?

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,35 @@ def rnnt_loss(
...
+    phi = torch.linalg.solve(psd_n, psd_s)  # psd_n.inv() @ psd_s
+    if torch.jit.isinstance(reference_channel, int):
+        rtf = phi[..., reference_channel]
+    else:
+        rtf = torch.einsum("...c,...c->...", [phi, reference_channel[..., None, None, :]])
+    rtf = rtf.unsqueeze(-1)  # (..., freq, channel, 1)
+    if iterations >= 2:
+        for _ in range(iterations - 2):
```
mthrok, line 1661: Can you add a comment of why `- 2` so that future maintainers who do not have knowledge of this algorithm knows it's indented and won't suspect it's a bug?

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,43 @@ def rnnt_loss(
...
+    if torch.jit.isinstance(reference_channel, int):
+        rtf = phi[..., reference_channel]
+    elif torch.jit.isinstance(reference_channel, Tensor):
+        reference_channel = reference_channel.to(psd_n.dtype)
+        rtf = torch.einsum("...c,...c->...", [phi, reference_channel[..., None, None, :]])
+    else:
+        raise TypeError(f"Unsupported dtype for reference_channel. Found: {type(reference_channel)}.")
+    rtf = rtf.unsqueeze(-1)  # (..., freq, channel, 1)
+    if iterations >= 2:
+        # rtf = psd_s (psd_n^-1  psd_s)^(iterations - 2) psd_n^-1  psd_s  ref_channel
```
mthrok, line 1666: Where is the missing 2 iterations happening?

In the `else` clause, there is one `matmul` happening and this corresponds to the number of iterations given by client code (== 1).

In the `if` clause, I see `iteration - 1` times of `matmul`. Where is the one more iteration that client code asks?

mthrok: I think `iteration=1` and `iteration=2` return the same result, which does not seem correct.

nateanl: The difference of `iteration 1` and `iteration 2` is the `psd` that rtf multiplies with:
```python
if iteration == 1:
    rtf = psd_n @ psd_n.inv() @ psd_s @ reference_channel
elif iteration == 2:
    rtf = psd_s @ psd_n.inv() @ psd_s @ reference_channel
else:
    rtf = psd_s @ (psd_n.inv() @ psd_s) ** (iteration - 2) @ psd_n.inv() @ psd_s @ reference_channel
```

nateanl: This formula can be simplified to one formula:

```
rtf = (psd_s @ psd_n.inv()) ** (iteration -1) @ psd_s @ reference_channel
```

In Pytorch, computing `psd_n.inv() @ psd_s` is easier to compute `psd_s @ psd_n.inv()`, I think that's why `phi = torch.linalg.solve(psd_n, psd_s) ` is computed first.

mthrok: ah okay, I did not see the difference of `psd_s` and `psd_n` in `torch.matmul`.

So if I understand correctly, then the initial computation of `phi` counts as one iteration, then the final `torch.matmul(psd_s, rtf)` counts as another iteration, which is why `rtf = torch.matmul(phi, rtf)` is performed `iterations - 2` times. 

If the above statement is correct, can you update the comment so that random future maintainers understand why `iterations -2`?  (and please use words instead of formula, so that SWEs without knowledge of beam forming can see and won't wonder why `iterations - 2`.

mthrok: @nateanl The existing APIs use `n_iter`. So can you use the same argument name?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -582,6 +583,51 @@ def test_rnnt_loss_costs_and_gradients_random_data_with_numpy_fp32(self):
...
             self._test_costs_and_gradients(data=data, ref_costs=ref_costs, ref_gradients=ref_gradients)
 
+    @parameterized.expand(
+        [
+            (1,),
+            (2,),
+            (3,),
+        ]
+    )
+    def test_rtf_power(self, iterations):
```
mthrok, line 593: Even though it is simple, please add the docstring. So that random future maintainers get what this test is about, without reading the code. (the same goes to the other test)

mthrok (approved):

torchaudio/functional/functional.py
```diff
@@ -1631,3 +1632,43 @@ def rnnt_loss(
...
+    """
+    assert iterations > 0, "The number of iteration must be greater than 0."
+    phi = torch.linalg.solve(psd_n, psd_s)  # psd_n.inv() @ psd_s
+    if torch.jit.isinstance(reference_channel, int):
+        rtf = phi[..., reference_channel]
+    elif torch.jit.isinstance(reference_channel, Tensor):
+        reference_channel = reference_channel.to(psd_n.dtype)
+        rtf = torch.einsum("...c,...c->...", [phi, reference_channel[..., None, None, :]])
+    else:
+        raise TypeError(f"Unsupported dtype for reference_channel. Found: {type(reference_channel)}.")
```
mthrok, line 1663: To be user-friendly, write what users should do to fix the error, in addition to what went wrong.

Something like `Expected 'int' or 'Tensor' as reference_channel. Found: ...` would save users from time they need to look up the API.

PR #2240, opened 2022-02-15
Add unit tests for PyTorch Lightning modules of emformer_rnnt recipes

nateanl:
- Refactor the current `LibriSpeechRNNTModule`'s unit test.
- Add unit tests for `TEDLIUM3RNNTModule` and `MuSTCRNNTModule`
- Replace the lambda with partial in `TEDLIUM3RNNTModule` to pass the lightning unit test.

### DIFF
--- a/examples/asr/emformer_rnnt/tedlium3/lightning.py
+++ b/examples/asr/emformer_rnnt/tedlium3/lightning.py
@@ -1,4 +1,5 @@
 import os
+from functools import partial
 from typing import List
 
 import sentencepiece as spm
@@ -86,20 +87,20 @@ class TEDLIUM3RNNTModule(LightningModule):
         self.train_data_pipeline = torch.nn.Sequential(
             FunctionalModule(piecewise_linear_log),
             GlobalStatsNormalization(global_stats_path),
-            FunctionalModule(lambda x: x.transpose(1, 2)),
+            FunctionalModule(partial(torch.transpose, dim0=1, dim1=2)),
             torchaudio.transforms.FrequencyMasking(27),
             torchaudio.transforms.FrequencyMasking(27),
             torchaudio.transforms.TimeMasking(100, p=0.2),
             torchaudio.transforms.TimeMasking(100, p=0.2),
-            FunctionalModule(lambda x: torch.nn.functional.pad(x, (0, 4))),
-            FunctionalModule(lambda x: x.transpose(1, 2)),
+            FunctionalModule(partial(torch.nn.functional.pad, pad=(0, 4))),
+            FunctionalModule(partial(torch.transpose, dim0=1, dim1=2)),
         )
         self.valid_data_pipeline = torch.nn.Sequential(
             FunctionalModule(piecewise_linear_log),
             GlobalStatsNormalization(global_stats_path),
-            FunctionalModule(lambda x: x.transpose(1, 2)),
-            FunctionalModule(lambda x: torch.nn.functional.pad(x, (0, 4))),
-            FunctionalModule(lambda x: x.transpose(1, 2)),
+            FunctionalModule(partial(torch.transpose, dim0=1, dim1=2)),
+            FunctionalModule(partial(torch.nn.functional.pad, pad=(0, 4))),
+            FunctionalModule(partial(torch.transpose, dim0=1, dim1=2)),
         )
 
         self.tedlium_path = tedlium_path
@@ -197,8 +198,8 @@ class TEDLIUM3RNNTModule(LightningModule):
     def validation_step(self, batch, batch_idx):
         return self._step(batch, batch_idx, "val")
 
-    def test_step(self, batch, batch_idx):
-        return self._step(batch, batch_idx, "test")
+    def test_step(self, batch_tuple, batch_idx):
+        return self._step(batch_tuple[0], batch_idx, "test")
 
     def train_dataloader(self):
         dataset = CustomDataset(torchaudio.datasets.TEDLIUM(self.tedlium_path, release="release3", subset="train"), 100)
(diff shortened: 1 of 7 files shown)


examples/asr/emformer_rnnt/tedlium3/lightning.py
```diff
@@ -197,8 +198,8 @@ def training_step(self, batch: Batch, batch_idx):
     def validation_step(self, batch, batch_idx):
         return self._step(batch, batch_idx, "val")
 
-    def test_step(self, batch, batch_idx):
-        return self._step(batch, batch_idx, "test")
+    def test_step(self, batch_tuple, batch_idx):
+        return self._step(batch_tuple[0], batch_idx, "test")
```
mthrok, line 202: What is the significance of `[0]` here?
All the other changes appear cosmetic, but this one is different.

test/torchaudio_unittest/example/emformer_rnnt/test_tedlium3_lightning.py
```diff
@@ -0,0 +1,76 @@
...
+            sp_model_path="sp_model_path",
+            global_stats_path="global_stats_path",
+        )
+
+
+@skipIfNoModule("pytorch_lightning")
+@skipIfNoModule("sentencepiece")
+class TestTEDLIUM3RNNTModule(TorchaudioTestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
```
mthrok, line 51: Has to call the same method of the super class first.

```python
super().setUpClass()
```

examples/asr/emformer_rnnt/tedlium3/lightning.py
nateanl: The change aligns with [LibriSpeech recipe](https://github.com/pytorch/audio/blob/main/examples/asr/emformer_rnnt/librispeech/lightning.py#L177-L178)

The reason is that for test_dataloader, the text transcription is required for WER computation, hence the [_test_collate_fn](https://github.com/pytorch/audio/blob/main/examples/asr/emformer_rnnt/librispeech/lightning.py#L138) returns the transcript along with the Batch input.

test/torchaudio_unittest/example/emformer_rnnt/test_mustc_lightning.py
```diff
@@ -0,0 +1,79 @@
...
+
+@skipIfNoModule("pytorch_lightning")
+@skipIfNoModule("sentencepiece")
+class TestMuSTCRNNTModule(TorchaudioTestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        torch.random.manual_seed(31)
+
+    def test_training_step(self):
```
mthrok, line 51: Looks like the implementation of four test methods (`test_training_step`, `test_validation_step`, `test_test_common_step` and `test_test_he_step`) are exactly the same.

Can you find a way to abstract the difference (I assume it's just how lightning module is mocked out) and parameterize, so that we do not need to maintain duplicated code?

test/torchaudio_unittest/example/emformer_rnnt/test_librispeech_lightning.py
```diff
@@ -80,28 +52,33 @@ def get_lightning_module():
...
+                dataloader = lightning_module.test_dataloader()
+                step = lightning_module.test_step
+            else:
+                dataloader = lightning_module.val_dataloader()
+                step = lightning_module
+            batch = next(iter(dataloader))
+            if subset == "forward":
+                step(batch)
+            else:
+                step(batch, 0)
```
hwangjeff, line 84: if we want to parameterize it this way, i'd forego the mapping from "subset", e.g.

```
@parameterized.expand(
    [
        ("training_step", "train_dataloader"),
        # etc
    ]
)
def test_step(step_fname, dataloader_fname):
    with get_lightning_module() as lightning_module:
        dataloader = getattr(lightning_module, dataloader_fname)()
        batch = next(iter(dataloader))
        getattr(lightning_module, step_fname)(batch, 0)
```

hwangjeff:
can you also try out mocking the dataloader to bypass multiprocessing to see whether that mitigates the mac python 3.7 test hanging issue?

nateanl:
@hwangjeff The hanging issue seems to be resolved by mocking the dataloader.

torchaudio/pipelines/rnnt_pipeline.py
```diff
@@ -388,7 +388,7 @@ def get_token_processor(self) -> TokenProcessor:
 
     The underlying model is constructed by :py:func:`torchaudio.models.emformer_rnnt_base`
     and utilizes weights trained on LibriSpeech using training script ``train.py``
-    `here <https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_emformer_rnnt>`__ with default arguments.
+    `here <https://github.com/pytorch/audio/tree/main/examples/asr/emformer_rnnt>`__ with default arguments.
```
hwangjeff, line 391: thanks for fixing these

hwangjeff:
> @hwangjeff The hanging issue seems to be resolved by mocking the dataloader.

great, thanks!

hwangjeff (approved):

mthrok (approved):
What I originally meant was that test code was identical across librispeech/tedlium cases, but It seems parameterizing within the same dataset is also a good option.

mthrok:
> > @hwangjeff The hanging issue seems to be resolved by mocking the dataloader.
> 
> great, thanks!

I think this is worth reporting somewhere in the core.
@VitalyFedyunin @ejguan @NivekT We had an issue where mock dataset hangs inside of dataloader. Was there a dataloader update recently? The issue was observed sporadicly so we are not sure about the trigger yet.

VitalyFedyunin:
No known DataLoader updates

PR #2243, opened 2022-02-16
Refactor unit tests in functional

nateanl:
The PR addresses these issues:
- In autograd tests, the complex dtype is not considered. Add ``torch.cdouble`` support for the Tensors.
- In batch_consistency tests, the ``assert_batch_consistency`` method only accepts single Tensor, this is not applicable to some methods, for example, `lfilter` and `filtfilt` requires three Tensors as the arguments, hence they don't use `assert_batch_consistency` in the tests. Refactor the test to accept a list of Tensors
- Similar in torchscript_consistency tests, the `func` in each test method only accepts one `tensor` as the argument, for the other arguments for `F.xxx` method, they need to be defined inside the `func`. If there is no Tensor argument in the test, they use a `dummy` tensor which is not used anywhere. In this PR, we refactor ``_assert_consistency`` and ``_assert_consistency_complex`` to accept a list of inputs instead of just one `tensor`.

### DIFF
--- a/test/torchaudio_unittest/functional/autograd_impl.py
+++ b/test/torchaudio_unittest/functional/autograd_impl.py
@@ -24,7 +24,7 @@ class Autograd(TestBaseMixin):
         inputs_ = []
         for i in inputs:
             if torch.is_tensor(i):
-                i = i.to(dtype=self.dtype, device=self.device)
+                i = i.to(dtype=torch.cdouble if i.is_complex() else self.dtype, device=self.device)
                 if enable_all_grad:
                     i.requires_grad = True
             inputs_.append(i)
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -26,20 +26,19 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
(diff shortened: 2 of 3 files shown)


mthrok (commented):
Can you add motivation of these changes in description? I am aware of some, but not sure about the others.

nateanl:
Updated the description.

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -24,7 +24,7 @@ def assert_grad(
         inputs_ = []
         for i in inputs:
             if torch.is_tensor(i):
-                i = i.to(dtype=self.dtype, device=self.device)
+                i = i.to(dtype=torch.cdouble if i.is_complex() else self.dtype, device=self.device)
```
mthrok, line 27: `self.complex_dtype` should do.

https://github.com/pytorch/audio/blob/99b5ef5ca0c3a48d8b703d0e1260aafe3334e844/test/torchaudio_unittest/common_utils/case_utils.py#L86-L92

nateanl:
Sure, will do.

nateanl:
Split the PR into three small PRs:
- [ ] https://github.com/pytorch/audio/pull/2244
- [ ] https://github.com/pytorch/audio/pull/2245
- [ ] https://github.com/pytorch/audio/pull/2246

PR #2245, opened 2022-02-16
Refactor batch consistency test in functional

nateanl:
In batch_consistency tests, the `assert_batch_consistency` method only accepts single Tensor, which is not applicable to some methods. For example, `lfilter` and `filtfilt` requires three Tensors as the arguments, hence they don't follow `assert_batch_consistency` in the tests. 
This PR refactors the test to accept a tuple of Tensors which have `batch` dimension. For the other arguments like `int` or `str`, they are given as `*args` after the tuple.

### DIFF
--- a/test/torchaudio_unittest/functional/batch_consistency_test.py
+++ b/test/torchaudio_unittest/functional/batch_consistency_test.py
@@ -26,20 +26,19 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
-    def assert_batch_consistency(self, functional, batch, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
-        n = batch.size(0)
+    def assert_batch_consistency(self, functional, inputs, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
+        n = inputs[0].size(0)
 
         # Compute items separately, then batch the result
         torch.random.manual_seed(seed)
-        items_input = batch.clone()
-        items_result = torch.stack([functional(items_input[i], *args, **kwargs) for i in range(n)])
+        items_input = [[ele[i].clone() for ele in inputs] for i in range(n)]
+        items_result = torch.stack([functional(*items_input[i], *args, **kwargs) for i in range(n)])
 
         # Batch the input and run
         torch.random.manual_seed(seed)
-        batch_input = batch.clone()
-        batch_result = functional(batch_input, *args, **kwargs)
+        batch_input = [ele.clone() for ele in inputs]
+        batch_result = functional(*batch_input, *args, **kwargs)
 
-        self.assertEqual(items_input, batch_input, rtol=rtol, atol=atol)
         self.assertEqual(items_result, batch_result, rtol=rtol, atol=atol)
 
     def test_griffinlim(self):
@@ -54,7 +53,7 @@ class TestFunctional(common_utils.TorchaudioTestCase):
         torch.random.manual_seed(0)
         batch = torch.rand(self.batch_size, 1, 201, 6)
         self.assert_batch_consistency(
-            F.griffinlim, batch, window, n_fft, hop, ws, power, n_iter, momentum, length, 0, atol=5e-5
+            F.griffinlim, (batch,), window, n_fft, hop, ws, power, n_iter, momentum, length, 0, atol=5e-5
         )
 
     @parameterized.expand(
@@ -79,7 +78,7 @@ class TestFunctional(common_utils.TorchaudioTestCase):
                 for frequency in frequencies
             ]
         )
-        self.assert_batch_consistency(F.detect_pitch_frequency, waveforms, sample_rate)
+        self.assert_batch_consistency(F.detect_pitch_frequency, (waveforms,), sample_rate)
 
     def test_amplitude_to_DB(self):
         torch.manual_seed(0)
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -26,20 +26,19 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
-    def assert_batch_consistency(self, functional, batch, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
-        n = batch.size(0)
+    def assert_batch_consistency(self, functional, inputs, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
```
mthrok, line 29: With this number of arguments and the combination of args and kwargs, it's error-prone. I think it's better to use keyword-only.

```suggestion
    def assert_batch_consistency(self, functional, *, inputs, args=None, kwargs=None, atol=1e-8, rtol=1e-5, seed=42):
```



```suggestion
    def assert_batch_consistency(self, functional, *inputs, *, atol=1e-8, rtol=1e-5, seed=42, args=None, kwargs=None):
```

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -26,20 +26,19 @@ class TestFunctional(common_utils.TorchaudioTestCase):
...
+        n = inputs[0].size(0)
```
mthrok, line 30: Needs to check that all the input tensors have the same batch size.

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -216,28 +209,20 @@ def test_compute_kaldi_pitch(self):
...
         signal_length = 2048
         torch.manual_seed(2434)
         x = torch.randn(self.batch_size, signal_length)
         a = torch.rand(self.batch_size, 3)
         b = torch.rand(self.batch_size, 3)
-
-        batchwise_output = F.lfilter(x, a, b, batching=True)
-        itemwise_output = torch.stack([F.lfilter(x[i], a[i], b[i]) for i in range(self.batch_size)])
-
-        self.assertEqual(batchwise_output, itemwise_output)
```
mthrok, line 231: I just realized that this method was not honoring the `self.dtype` and `self.device` configuration. Grad that you pulled up this PR.

test/torchaudio_unittest/functional/batch_consistency_test.py
nateanl: If not using ``kwargs``, how to pass the keywords to `functional`?

test/torchaudio_unittest/functional/batch_consistency_test.py
nateanl: There're other methods that use `torch.rand` without `self.dtype` and `self.device`, shall we also update them in this PR?

mthrok: If they use `self.assert_batch_consistency`, it's fine, otherwise we need to fix it. (but let's do it in a separate PR.)

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -26,20 +26,24 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
-    def assert_batch_consistency(self, functional, batch, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
-        n = batch.size(0)
-
+    def assert_batch_consistency(self, functional, *, inputs, args=None, kwargs=None, atol=1e-8, rtol=1e-5, seed=42):
+        if len(inputs) > 1:
+            self.assertEqual(*[ele.size(0) for ele in inputs])
```
mthrok, line 31: Does assertEqual works for n>2?

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -26,20 +26,24 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
-    def assert_batch_consistency(self, functional, batch, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
-        n = batch.size(0)
-
+    def assert_batch_consistency(self, functional, *, inputs, args=None, kwargs=None, atol=1e-8, rtol=1e-5, seed=42):
```
mthrok, line 29: I am having a second thought about this. Maybe using `partial` is better. What do you think?

nateanl: `partial` may make the `functional` not consistent. If the function doesn't have `kwargs`, then it's a pure `F.xyz` method, otherwise it's a `partial`. I feel like `kwargs` as dictionary is fine.

mthrok: > `partial` may make the `functional` not consistent. 

That should not matter because the test is interested in whether the returned value are consistent. If the function passed belongs to `torchaudio.functional` has no effect on the test logic.

nateanl: That's true. The test logic won't be affected in either case. The advantage of `kwargs` is that the test methods can follow the same pattern.

mthrok: [partial](https://docs.python.org/3/library/functools.html) allows args and kwargs, so it's essentially just moving the argument binding outside of the helper method.
The concern / responsibility of the helper method is the function to be tested and Tensor inputs, so delegating bidding of the other arguments to call site is less error-prone.

test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -26,20 +27,21 @@ class TestFunctional(common_utils.TorchaudioTestCase):
 
     backend = "default"
 
-    def assert_batch_consistency(self, functional, batch, *args, atol=1e-8, rtol=1e-5, seed=42, **kwargs):
-        n = batch.size(0)
-
+    def assert_batch_consistency(self, functional, inputs, atol=1e-8, rtol=1e-5, seed=42):
+        n = inputs[0].size(0)
+        if len(inputs) > 1:
```
mthrok, line 32: `if len(inputs) > 1:` is redundant. `range(1, len(inputs))` will handle the case `len(inputs) == 1`.

mthrok (approved):

PR #2246, opened 2022-02-16
Refactor torchscript consistency test in functional

nateanl:
In torchscript_consistency tests, the `func` in each test method only accepts one `tensor` as the argument, for the other arguments of `F.xyz` method, they need to be defined inside the `func`. If there is no `Tensor` argument in `F.xzy`, the tests use a `dummy` tensor which is not used anywhere. In this PR, we refactor ``_assert_consistency`` and ``_assert_consistency_complex`` to accept a tuple of inputs instead of just one `tensor`.

### DIFF
--- a/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
+++ b/test/torchaudio_unittest/functional/torchscript_consistency_impl.py
@@ -15,90 +15,91 @@ from torchaudio_unittest.common_utils import (
 class Functional(TempDirMixin, TestBaseMixin):
     """Implements test for `functional` module that are performed for different devices"""
 
-    def _assert_consistency(self, func, tensor, shape_only=False):
-        tensor = tensor.to(device=self.device, dtype=self.dtype)
+    def _assert_consistency(self, func, inputs, shape_only=False):
+        inputs_ = []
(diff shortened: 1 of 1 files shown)


test/torchaudio_unittest/functional/torchscript_consistency_impl.py
```diff
@@ -603,46 +579,38 @@ def func(tensor):
...
-
         tensor = common_utils.get_whitenoise(sample_rate=16000)
-        self._assert_consistency(func, tensor)
-        self._assert_consistency(func_beta, tensor)
+        sr1, sr2 = 16000, 8000
+        beta = 6.0
+        lowpass_filter_width = 6
+        rolloff = 0.99
+        self._assert_consistency(F.resample, (tensor, sr1, sr2, lowpass_filter_width, rolloff, "kaiser_window", None))
+        self._assert_consistency(F.resample, (tensor, sr1, sr2, lowpass_filter_width, rolloff, "kaiser_window", beta))
```
mthrok, line 599: Technically out of the scope of this PR but can you use parameterized on `beta`?

carolineechen (approved):

PR #2249, opened 2022-02-17
Update release notes labeling

carolineechen:
- fix retrieve PR script to handle commits with unrecognized/invalid PR numbers, such as in 7b6b2d000023e2aa3365b769866c5f375e0d5fda
- add modifications similar to pytorch's [#71917](https://github.com/pytorch/pytorch/pull/71917), [#72085](https://github.com/pytorch/pytorch/pull/72085)

### DIFF
--- a/.github/process_commit.py
+++ b/.github/process_commit.py
@@ -5,6 +5,8 @@ Note: we only ping the person who pulls the pr, not the reviewers, as the review
 to torchaudio with no labeling responsibility, so we don't want to bother them.
 """
 
+import json
+import os
 import sys
 from typing import Any, Optional, Set, Tuple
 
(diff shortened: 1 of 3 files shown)


mthrok (approved):
Stamp

mthrok (commented):
I am curious, why not making it a CI job and simply make it merge blocking job?

PyTorch core needs more flexibility because of frequent external contributors, but we do not have one.

.github/process_commit.py
```diff
@@ -36,15 +37,18 @@
     "perf",
     "other",
 }
+GITHUB_TOKEN = os.environ.get("GITHUB_TOKEN")
```
mthrok, line 40: Is a maintainer expected to run this script manually and locally?
If so, I think it's better to 
- fail if the env var is not defined, and
- provide `--help` option and the `--help` option to work even when GITHUB_TOKEN is not set.

carolineechen: this script shouldn't be run by maintainers, but by the github workflow when there is a push to main. the github token is set [here](https://github.com/pytorch/audio/blob/e8f7e192d4b47f0ec25b4920bd29ffe216739ebe/.github/workflows/pr-labels.yml#L26)

carolineechen:
> I am curious, why not making it a CI job and simply make it merge blocking job?

Wanted to try resolving the current approach if changes were simple, as most of the scripts and setup are already implemented. I'm not as familiar with implementing the CI job yet, but if this is still not working or if the team prefers (or if you'd like to make the change), happy to use a CI job to replace this instead.

nateanl (approved):

PR #2256, opened 2022-02-24
ci: Limit scope of unittest to one python version

seemethere:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* __->__ #2256

Limits scope of unittesting to one python version for both macOS and
Windows. These types of workflows are particularly expensive and take a
long time so running them on every PR / every push is a bit wasteful
considering the value in signal between different python versions is
probably negligible.

Differential Revision: [D34459626](https://our.internmc.facebook.com/intern/diff/D34459626)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1323,56 +1323,14 @@ workflows:
           cuda_version: cpu
           name: unittest_windows_cpu_py3.7
           python_version: '3.7'
-      - unittest_windows_cpu:
-          cuda_version: cpu
-          name: unittest_windows_cpu_py3.8
-          python_version: '3.8'
-      - unittest_windows_cpu:
-          cuda_version: cpu
-          name: unittest_windows_cpu_py3.9
-          python_version: '3.9'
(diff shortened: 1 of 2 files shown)


malfet (approved):

atalman (approved):

seemethere:
@seemethere has imported this pull request.  If you are a Facebook employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D34459626).

mthrok (approved):
LGTM.

I just wanted to note that there were cases where Python version mattered in macOS. 
It ultimately boils down to how multiprocessing is used.

https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods
> Changed in version 3.8: On macOS, the spawn start method is now the default. The fork start method should be considered unsafe as it can lead to crashes of the subprocess. See [bpo-33725](https://bugs.python.org/issue33725).

However, it is more about how the code is using `subprocessing`, (and it still comes back from time to time) so I agree with the point, not providing much signal.

Issue #2158, opened 2022-01-17
New interface for MVDR beamforming


Emrys365:
### 🚀 The feature

It would be very helpful to provide the following interface for the beamforming module ([torchaudio.transforms.MVDR](https://pytorch.org/audio/master/transforms.html#torchaudio.transforms.MVDR.forward)):
```python
forward(specgram: torch.Tensor, psd_s: torch.Tensor, psd_n: torch.Tensor) → torch.Tensor
```
and maybe add some high-level glue functions that takes the masks as input, but has only a few lines of code.

### Motivation, pitch

The current forward method of `torchaudio.transforms.MVDR` only accepts spectrogram and masks as input, and calculate the PSD matrices internally.

The current design is easy to use mainly for mask-based beamforming, but it may lose the flexibility for the users to:
  * use their own PSD estimates (e.g. mask-based PSDs, neural-network-based PSD estimates in [ADL-MVDR](https://arxiv.org/abs/2008.06994) and [NICE-BEAM](https://arxiv.org/abs/2112.04613))
  * adjust PSD estimates with some additional regularization/stabilization techniques (e.g. PSD diagonal loading and PSD eigenvalue flooring)
     > Because in different setups, a different stabilization may be ideal or preferred. It would be better to allow the user to make the choice.
  * combine with conventional blind beamforming methods
  * etc.

nateanl:
@mthrok What do you think? I feel like the current `torchaudio.transforms.MVDR` can be used as the high-level module, which calls another module with PSD as input arguments in the `forward` method.

Maybe we should rename the current MVDR module to a generalized `MaskBasedBeamforming` module which accepts any kind of beamforming method, similar to [` DNN_Beamformer`](https://github.com/espnet/espnet/blob/master/espnet2/enh/layers/dnn_beamformer.py) in ESPNet, but without trainable parameters. Then everytime we add a new beamforming module, we can add such support in the `MaskBasedBeamforming`.

mthrok:
@nateanl Can you make a diagram of changes you propose with call graph and signatures? (see the following for an inspiration)

![](https://user-images.githubusercontent.com/855818/102725281-ae0c8e80-42e3-11eb-929f-d3848989eafb.png)

nateanl:
Here is the diagram for changing the API of MVDR.
[...]

PR #2262, opened 2022-02-25
Add deprecation warning in torchaudio.transforms.MVDR

nateanl:
As described in https://github.com/pytorch/audio/issues/2280, we will add new API designs to beamforming methods and modules, such as souden-version MVDR and RTF-version MVDR. This PR adds deprecation warning message to the current MVDR module. The deprecation process will be done in release 0.12.

### DIFF
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -1720,6 +1720,13 @@ class MVDR(torch.nn.Module):
         online: bool = False,
     ):
         super().__init__()
+        warnings.warn(
+            "The MVDR module will be deprecated in release 0.12. "
+            "Please refer to https://github.com/pytorch/audio/issues/2158 "
+            "for more details about torchaudio's plan for refactoring beamforming methods and modules. "
+            "For replacement, please check the mvdr_weights_souden and mvdr_weights_rtf "
+            "methods in torchaudio.functional."
+        )
         assert solution in [
             "ref_channel",
             "stv_evd",
--- a/torchaudio/transforms.py
+++ b/torchaudio/transforms.py
@@ -1721,9 +1721,9 @@ class MVDR(torch.nn.Module):
     ):
         super().__init__()
         warnings.warn(
-            "The MVDR module will be deprecated in release 0.12. "
-            "Please refer to https://github.com/pytorch/audio/issues/2158 "
-            "for more details about torchaudio's plan for refactoring beamforming methods and modules. "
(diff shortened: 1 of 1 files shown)


torchaudio/transforms.py
```diff
@@ -1720,6 +1720,13 @@ def __init__(
         online: bool = False,
     ):
         super().__init__()
+        warnings.warn(
+            "The MVDR module will be deprecated in release 0.12. "
```
mthrok, line 1724: Do you mean `removed`?

torchaudio/transforms.py
```diff
@@ -1720,6 +1720,13 @@ def __init__(
...
+            "Please refer to https://github.com/pytorch/audio/issues/2158 "
```
mthrok, line 1725: #2158 is a proposal and discussion. Can you make an issue dedicated for overview, timeline and progress? like https://github.com/pytorch/audio/issues/1337?

mthrok:
Note: If deprecating MVDR in 0.11 and removing it in 0.12, then we need to add a migration guide in the dedicated issue or release note.

The migration guide should be in a table format like 

https://github.com/pytorch/pytorch/releases/tag/v1.10.0

torchaudio/transforms.py
```diff
@@ -1720,6 +1720,13 @@ def __init__(
         online: bool = False,
     ):
         super().__init__()
+        warnings.warn(
+            "The MVDR module is deprecated and will be removed in release 0.12. "
```
mthrok, line 1724: Here are some thoughts around this deprecation.

1. We are constrained to give two release cycles for deprecation and removal for stable features. (MVDR was released as beta features.)
2. I am not sure if we can release the new replacements in the minor release `0.11.1`. And my preference is to postpone until `0.12`.
3. My understanding is that after introducing the new features and refactoring the internal, this MVDR module is in free form. It is not like the existence of this deprecated MVDR module will prevent the development some other features.

So I think we do not need to be eager about removal in 0.12. Once the replacements are introduced, we can leave it there, until we meet the two release cycle deprecation criteria, even though MVDR is not stable feature.

nateanl: After conforming with @Emrys365, we don't have to push the new features to minor release and can still adopt them to ESPNet. So It's fine to postpone them to `0.12`.
About the MVDR module, we can leave it before we add new modules in `transforms`. Then we can come back to this PR and deprecate it.

PR #2033, opened 2021-11-24
[WIP] move set random seed in test cases to setup

krishnakalyan3:

PR #2267, opened 2022-02-27
Improve Unit tests by moving random seeds to setUp

krishnakalyan3:
Following up on #2033.

### DIFF
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -83,6 +83,7 @@ class TestBaseMixin:
     def setUp(self):
         super().setUp()
         set_audio_backend(self.backend)
+        torch.random.manual_seed(2434)
 
     @property
     def complex_dtype(self):
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -83,7 +83,6 @@ class TestBaseMixin:
     def setUp(self):
         super().setUp()
         set_audio_backend(self.backend)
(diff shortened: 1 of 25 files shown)


nateanl (changes requested):
Hi @krishnakalyan3, thanks for the PR! Some test methods still use `seed` when calling `get_whitenoise`. For example, the methods under [datasets test](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/datasets/cmuarctic_test.py#L40). Could you also address them in this PR?

krishnakalyan3:
@nateanl looks like the seed is being [changed](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/datasets/cmuarctic_test.py#L51) in all dataset tests. Do we modify the random seed within the loop for all datasets?.

mthrok:
> Hi @krishnakalyan3, thanks for the PR! Some test methods still use `seed` when calling `get_whitenoise`. For example, the methods under [datasets test](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/datasets/cmuarctic_test.py#L40). Could you also address them in this PR?

@zhangguanheng66 This is intended design. The problem that the original PR was solving is automatically setting the seed of the library code that is run. If seed is not reset at the start of every test run, the test results end up depend on the order the test is executed. Test utilities allowing explicit seed is the right thing to do, otherwise it will implicitly depend on the global seed state, which could interfere the library code behavior.

@krishnakalyan3 the initial state of the PR looked fine. Can you revert the latest commit? We can merge it and follow up for the remaining.

carolineechen:
completed in #2425 and #2436

PR #2275, opened 2022-03-11
Add Pretrained LM Support for Decoder

carolineechen:
add function to download pretrained files for LibriSpeech 3-gram/4-gram KenLM, tests, and updated tutorial

### DIFF
--- a/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
+++ b/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
@@ -178,11 +178,7 @@ print(tokens)
 
 from torchaudio.prototype.ctc_decoder import download_pretrained_files
 
-files = download_pretrained_files(
-    path=hub_dir,
-    dataset="LibriSpeech",
-    lm="4-gram",
-)
+files = download_pretrained_files("librispeech-4-gram")
 
 print(files)
 
--- a/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
+++ b/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
@@ -11,6 +11,21 @@ using CTC loss.
 
 """
 
+######################################################################
+#
+# .. note::
+#
+#    This tutorial requires torchaudio prototype features.
+#
+#    torchaudio prototype features are available on nightly builds.
+#    Please refer to https://pytorch.org/get-started/locally/
+#    for instructions.
+#
+#    The interfaces of prototype features are unstable and subject to
+#    change. Please refer to `the nightly build documentation
+#    <https://pytorch.org/audio/main/>`__ for the up-to-date
+#    API references.
+#
 
 ######################################################################
 # Overview
@@ -21,7 +36,10 @@ using CTC loss.
 # highest scores at each time step. A language model can be incorporated into
 # the scoring computation, and adding a lexicon constraint restricts the
 # next possible tokens for the hypotheses so that only words from the lexicon
-# can be generated. A mathematical formula for the decoder optimization can be
+# can be generated.
+#
+# The underlying implementation is ported from `Flashlight <https://arxiv.org/pdf/2201.12465.pdf>`__'s
+# beam search decoder. A mathematical formula for the decoder optimization can be
 # found in the `Wav2Letter paper <https://arxiv.org/pdf/1609.03193.pdf>`__, and
 # a more detailed algorithm can be found in this `blog
 # <https://towardsdatascience.com/boosting-your-sequence-generation-performance-with-beam-search-language-model-decoding-74ee64de435a>`__.
--- a/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
+++ b/examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
@@ -102,10 +102,9 @@ if sample_rate != bundle.sample_rate:
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
 # Next, we load in our token, lexicon, and KenLM data, which are used by
-# the decoder to predict words from the acoustic model output.
-#
-# Note: this cell may take a couple of minutes to run, as the language
-# model can be large
+# the decoder to predict words from the acoustic model output. Pretrained
+# files for the LibriSpeech dataset can be downloaded through torchaudio,
+# or the user can provide their own files.
 #
 
 
@@ -151,10 +150,6 @@ print(tokens)
 #    ...
 #
 
-lexicon_url = "https://download.pytorch.org/torchaudio/tutorial-assets/ctc-decoding/lexicon-librispeech.txt"
-lexicon_file = f"{hub_dir}/lexicon.txt"
-torch.hub.download_url_to_file(lexicon_url, lexicon_file)
-
 
 ######################################################################
 # KenLM
@@ -169,9 +164,27 @@ torch.hub.download_url_to_file(lexicon_url, lexicon_file)
 # `LibriSpeech <http://www.openslr.org/11>`__.
 #
 
-kenlm_url = "https://download.pytorch.org/torchaudio/tutorial-assets/ctc-decoding/4-gram-librispeech.bin"
-kenlm_file = f"{hub_dir}/kenlm.bin"
-torch.hub.download_url_to_file(kenlm_url, kenlm_file)
+
+######################################################################
+# Downloading Pretrained Files
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Pretrained files for the LibriSpeech dataset can be downloaded using
+# :py:func:`download_pretrained_files <torchaudio.prototype.ctc_decoder.download_pretrained_files>`.
+#
+# Note: this cell may take a couple of minutes to run, as the language
+# model can be large
+#
+
+from torchaudio.prototype.ctc_decoder import download_pretrained_files
+
+files = download_pretrained_files(
+    path=hub_dir,
+    dataset="LibriSpeech",
+    lm="4-gram",
+)
+
+print(files)
 
 
 ######################################################################
@@ -200,9 +213,9 @@ LM_WEIGHT = 3.23
 WORD_SCORE = -0.26
 
 beam_search_decoder = lexicon_decoder(
-    lexicon=lexicon_file,
-    tokens=tokens,
-    lm=kenlm_file,
+    lexicon=files.lexicon,
+    tokens=files.tokens,
+    lm=files.lm,
     nbest=3,
     beam_size=1500,
     lm_weight=LM_WEIGHT,
@@ -267,8 +280,8 @@ emission, _ = acoustic_model(waveform)
 #
 
 greedy_result = greedy_decoder(emission[0])
-greedy_transcript = greedy_result
-greedy_wer = torchaudio.functional.edit_distance(actual_transcript, greedy_transcript) / len(actual_transcript)
+greedy_transcript = " ".join(greedy_result)
+greedy_wer = torchaudio.functional.edit_distance(actual_transcript, greedy_result) / len(actual_transcript)
 
 print(f"Transcript: {greedy_transcript}")
 print(f"WER: {greedy_wer}")
@@ -404,9 +417,9 @@ beam_sizes = [1, 5, 50, 500]
 
 for beam_size in beam_sizes:
     beam_search_decoder = lexicon_decoder(
-        lexicon=lexicon_file,
-        tokens=tokens,
-        lm=kenlm_file,
+        lexicon=files.lexicon,
+        tokens=files.tokens,
+        lm=files.lm,
         beam_size=beam_size,
         lm_weight=LM_WEIGHT,
         word_score=WORD_SCORE,
@@ -430,9 +443,9 @@ beam_size_tokens = [1, 5, 10, num_tokens]
 
 for beam_size_token in beam_size_tokens:
     beam_search_decoder = lexicon_decoder(
-        lexicon=lexicon_file,
-        tokens=tokens,
-        lm=kenlm_file,
+        lexicon=files.lexicon,
+        tokens=files.tokens,
+        lm=files.lm,
         beam_size_token=beam_size_token,
         lm_weight=LM_WEIGHT,
         word_score=WORD_SCORE,
@@ -457,9 +470,9 @@ beam_thresholds = [1, 5, 10, 25]
 
 for beam_threshold in beam_thresholds:
     beam_search_decoder = lexicon_decoder(
-        lexicon=lexicon_file,
-        tokens=tokens,
-        lm=kenlm_file,
+        lexicon=files.lexicon,
+        tokens=files.tokens,
+        lm=files.lm,
         beam_threshold=beam_threshold,
         lm_weight=LM_WEIGHT,
         word_score=WORD_SCORE,
@@ -483,9 +496,9 @@ lm_weights = [0, LM_WEIGHT, 15]
 
 for lm_weight in lm_weights:
     beam_search_decoder = lexicon_decoder(
-        lexicon=lexicon_file,
-        tokens=tokens,
-        lm=kenlm_file,
+        lexicon=files.lexicon,
+        tokens=files.tokens,
+        lm=files.lm,
         lm_weight=lm_weight,
         word_score=WORD_SCORE,
     )
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -73,3 +73,13 @@ def temp_hub_dir(tmpdir, pytestconfig):
         torch.hub.set_dir(tmpdir)
         yield
         torch.hub.set_dir(org_dir)
+
+
+@pytest.fixture()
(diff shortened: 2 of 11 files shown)


torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +264,57 @@ def lexicon_decoder(
         sil_token=sil_token,
         unk_word=unk_word,
     )
+
+
+def download_pretrained_files(
+    path: Union[str, Path],
+    dataset: str,
+    lm: Optional[str] = None,
+) -> PretrainedFiles:
```
mthrok, line 273: The configuration of `dataset` + `lm` seems to be specific to the librispeech cases that we know, and I do not think this is future proof.

Imagine a hypothetical case where we want to add pertained model that is trained on some subset of a dataset. say `librispeech-dev100-3gram`. Do we want to add `subset` parameter? What about some other specifier that we do not know yet?

So I think it's more future proof to use single key for specifying the target model.
That way I think that the API will be more straightforward to users as well.

test/torchaudio_unittest/prototype/ctc_decoder_test.py
```diff
@@ -98,10 +105,28 @@ def test_get_tokens_and_idxs(self):
...
-        decoder = self._get_decoder(tokens)
+        decoder = self._get_decoder(tokens=tokens)
 
         idxs = torch.LongTensor((1, 2, 1, 3, 5))
         tokens = decoder.idxs_to_tokens(idxs)
 
         expected_tokens = ["|", "f", "|", "o", "a"]
         self.assertEqual(tokens, expected_tokens)
+
+    def test_download_pretrained_files(self):
```
mthrok, line 116: Can you either mock the http request or move this into integration tests directory?
Unit tests should not rely on any external resources. (tests should be runnable without internet connection)

torchaudio/prototype/ctc_decoder/ctc_decoder.py
mthrok, line 273: I am wondering if it is necessary to make `path` explicit.
Since the implementation is using hub download feature, how about abstracting away the path where the model is downloaded, like `pipelines` module does?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +263,46 @@ def lexicon_decoder(
...
+def download_pretrained_files(model: str) -> PretrainedFiles:
+    """
+    Retrieves pretrained data files used for CTC decoder.
+
+    Args:
+        model (str): pretrained language model to download
+            Options: ["librispeech-3-gram", "librispeech-4-gram", "librispeech"]
+
+    Returns:
+        Dictionary with the following key mapping:
```
mthrok, line 277:
```suggestion
        Object with the following attributes:
```

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -20,6 +23,9 @@
 __all__ = ["Hypothesis", "LexiconDecoder", "lexicon_decoder"]
 
 
+PretrainedFiles = namedtuple("PretrainedFiles", ["lexicon", "tokens", "lm"])
```
mthrok, line 26: Can we prefix this with underscore so that we can show it is discouraged to use this type.
This return type is introduced merely to facilitate the access, and it is not intended for the public use.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +263,46 @@ def lexicon_decoder(
...
+            lm: path corresponding to downloaded language model, or None if lm not provided
+            lexicon: path corresponding to downloaded lexicon file
+            tokens: path corresponding to downloaded tokens file
+    """
+
+    hub_dir = torch.hub.get_dir()
+    path = f"{hub_dir}/{model}"
+
+    url_prefix = "https://download.pytorch.org/torchaudio/decoder-assets"
+    model_dir = f"{url_prefix}/{model}"
```
mthrok, line 287: 1. I think it is better to extract the logic which handles "key -> route on download" to a helper function.
2. Whether if lm is available (whether `lm` attribute should be `None`) should be handled by the code here. It should not be the result of successful download.


```python

routes = _get_route(model)
token_file, lexicon_file, lm_file = routes

download_url_to_file(<token_uri>, <local_token_path>)
download_url_to_file(<lexicon_uri>, <local_lexicon_path>)
if lm_file is not None:
    download_url_to_file(<lm_uri>, <local_lm_path>)

return ...
```

mthrok, line 287: The variable name `model_dir` is not intuitive.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +263,46 @@ def lexicon_decoder(
...
+    download_url_to_file(f"{model_dir}/lexicon.txt", lexicon_file)
+
+    # Support for decoding without a language model
+    lm_url = f"{model_dir}/lm.bin"
+    response = requests.get(lm_url)
+    if response.status_code == 200:
+        lm_file = f"{path}/lm.bin"
+        download_url_to_file(lm_url, lm_file)
+    else:
+        lm_file = None
```
mthrok, line 302: This does not handle the case where the model exist but the download failed for some network issue.
`lm_file=None` should not be the result of download but the information encoded in this code.

See the comment above.

test/integration_tests/prototype/ctc_decoder_integration_test.py
```diff
@@ -0,0 +1,27 @@
+from parameterized import parameterized
+from torchaudio_unittest.common_utils import skipIfNoCtcDecoder
+from torchaudio_unittest.common_utils.ctc_decoder_utils import is_ctc_decoder_available
```
mthrok, line 3: I am avoiding to reuse the unit test utilities in integration tests for the sake of decoupling.
Can we simply let the tests fail if CTC decode feature is not found?

Also, please use `@pytest.mark.parametrize` for parameterizing in integration. We established the use of `parameterized` in unit tests because the unit tests take OOP approach, and we could not use `pytest` in fbcode in early days.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +269,48 @@ def lexicon_decoder(
         sil_token=sil_token,
         unk_word=unk_word,
     )
+
+
+def _get_filenames(model: str) -> _PretrainedFiles:
+    files = MODEL_FILE_MAP[model]
+    return _PretrainedFiles(lexicon=files.lexicon, tokens=files.tokens, lm=files.lm if files.lm is not None else None)
```
mthrok, line 276: It seems that the re-constructed of `_PretrainedFiles` instance is exactly same as the previous. Is this necessary?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -20,6 +23,15 @@
 __all__ = ["Hypothesis", "LexiconDecoder", "lexicon_decoder"]
 
 
+_PretrainedFiles = namedtuple("PretrainedFiles", ["lexicon", "tokens", "lm"])
+
+MODEL_FILE_MAP = {
```
mthrok, line 28: 1. It is kind of wasteful to instantiate this mapping on module scope, so let's move this inside of `_get_filenames` function.

2. Looking ash the current mapping, the key is not encoded in the values. I see that on L305, the key (`model` variable) is still used to construct the final URL. That part should be consolidated to `_get_filenames`. So that `_get_filenames` function is responsible for providing the all the information to find the data to be downloaded, and the rest of `download_pretrained_files` function only focus on downloading and managing file on local file system.

mthrok (commented):
Getting better. I think the implementation can be tidied a bit more.

Q: What happens if a file exists in a local storage?

As a follow-up, I think it's time for us to write a custom utility function that handles download from https://download.pytorch.org/torchaudio/* into a file with caching mechanism and manage the file location, so that the base url and the prefix of the local file path are handled behind the scene.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +263,54 @@ def lexicon_decoder(
...
+        Object with the following attributes:
+            lm: path corresponding to downloaded language model, or None if lm not provided
+            lexicon: path corresponding to downloaded lexicon file
+            tokens: path corresponding to downloaded tokens file
+    """
+
+    hub_dir = torch.hub.get_dir()
+    download_path = f"{hub_dir}/{model}"
+    if not os.path.isdir(download_path):
+        os.makedirs(download_path)
```
mthrok, line 302: L299-L302 are now handled by `download_asset` function, so we can get rid of it.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -257,3 +263,54 @@ def lexicon_decoder(
...
+        "librispeech-4-gram": _PretrainedFiles(lexicon="lexicon.txt", tokens="tokens.txt", lm="lm.bin"),
+    }
+
+    files = MODEL_FILE_MAP[model]
+    prefix = f"decoder-assets/{model}"
+    return _PretrainedFiles(
+        lexicon=f"{prefix}/{files.lexicon}",
+        tokens=f"{prefix}/{files.tokens}",
+        lm=f"{prefix}/{files.lm}" if files.lm is not None else None,
+    )
```
mthrok, line 281: Since the values of MODEL_FILE_MAP are identical, this function can be reduced to 

```python
def _get_filenames(model: str) -> _PretrainedFiles:
    prefix = f"decoder-assets/{model}"
    return _PretrainedFiles(
         lexicon=f"{prefix}/lexicon.txt",
         tokens=f"{prefix}/tokens.txt",
         lm=None if model is None else f"{prefix}/lm.bin",
)
```

note: The whole purpose of this helper function is to decouple the responsibility of mapping from key to route on download.pytorch.org/torchaudio. But the function itself does not have to use dict type for the mapping as long as the function serves as the correct mapping.

test/integration_tests/prototype/ctc_decoder_integration_test.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.mark.parametrize("model", ["librispeech", "librispeech-3-gram"])
+def test_download_pretrain(model):
+    # smoke test for downloading pretrained files
+    _ = download_pretrained_files(model)
+
+
+@pytest.mark.parametrize("model", ["librispeech", "librispeech-3-gram"])
+def test_decoder_from_pretrained(model):
```
mthrok, line 12: I think `test_download_pretrain` and `test_decoder_from_pretrained` can be merged. `test_download_pretrain` does not seem to add much value over `test_decoder_from_pretrained `, and yet it is hard to decouple the download logic from `test_decoder_from_pretrained`.

mthrok: Can we put `from torchaudio.prototype.ctc_decoder import lexicon_decoder, download_pretrained_files` inside of the test function so that the test will fail at test time instead of collect time, in case the said module is not available.

test/integration_tests/prototype/ctc_decoder_integration_test.py
```diff
@@ -0,0 +1,20 @@
...
+    # smoke test for constructing decoder from pretrained files
+    pretrained_files = download_pretrained_files(model)
+
+    _ = lexicon_decoder(
```
mthrok, line 16: Since we are constructing the decode object, how about actually decoding something?

I think we can upload some emission data to S3 and fetch it for the sake of testing here, similar to 

[`test_finetune_asr_model`](https://github.com/pytorch/audio/blob/7444f56819f679f68eee8bf915bbb74be3da0e40/test/integration_tests/wav2vec2_pipeline_test.py#L84) but starting from emission instead of waveform.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: sure, that works for now, but will probably need to be modified in the future (ex/ pretrained models for NNLM/lexicon free) where the file naming convention will be different

test/integration_tests/prototype/ctc_decoder_integration_test.py
carolineechen: good point, thanks

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -169,9 +164,23 @@
...
+# Pretrained files for the LibriSpeech dataset can be downloaded using
+# :py:func:`download_pretrained_files <torchaudio.prototype.ctc_decoder.download_pretrained_files>`.
+#
+# Note: this cell may take a couple of minutes to run, as the language
+# model can be large
+#
+
+from torchaudio.prototype.ctc_decoder import download_pretrained_files
+
+files = download_pretrained_files("librispeech-4-gram")
```
nateanl, line 181: It seems the method will download the file if the file already exists. Can we update it to check if the file is already there?

test/integration_tests/prototype/ctc_decoder_integration_test.py
carolineechen: sure, I added a function that generates/returns a random emission if that's sufficient

torchaudio/prototype/ctc_decoder/ctc_decoder.py
mthrok: > that would need to be modified in the future (ex/ when adding pretrained models for NNLM or lexicon free) where the naming convention will be different

Yeah, we can change it when we have the spec for the new additions. Until we have the spec (know how these new files will be placed) we need/should not take these in considerations. (this is called YAGNI principle. It is important because otherwise we sometimes cannot agree on how much of future expectation we want to put in the current implementation. And then priorities change and there are cases we end up giving up the said expected updates.)

test/integration_tests/conftest.py
```diff
@@ -73,3 +73,13 @@ def temp_hub_dir(tmpdir, pytestconfig):
         torch.hub.set_dir(tmpdir)
         yield
         torch.hub.set_dir(org_dir)
+
+
+@pytest.fixture()
+def emissions():
```
mthrok, line 79: I would go one step further and  use the actual emission.

1. Pick a license free speech
2. Feed it to wav2vec2 model
3. Save the resulting emission
4. Upload it to download.pytorch.org/test-assets/
5. Fetch it at the test time.

mthrok (approved):

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: "librispeech-3/4-gram" key will load the pretrained files including the corresponding kenlm file, but there's also the option of "librispeech", which will load the lexicon and tokens but not an LM file, as there is no LM associated with just "librispeech". so in that case, we need to differentiate if we return a path for the lm path or `None`

mthrok: Isn't that logic already implemented at `lm=f"{prefix}/lm.bin" if model != "librispeech" else None`?

test/integration_tests/prototype/ctc_decoder_integration_test.py
mthrok: > is there a reason to upload/fetch a sample emission?

That way, we can rest assured that the decode is generating a meaningful result.

torchaudio/prototype/ctc_decoder/ctc_decoder.py
carolineechen: yea that was added after the original comment -- just wanted to make sure hardcoding is fine for now since these are the only ones we are supporting now, sounds good

mthrok: Yup. that one LGTM. Please make sure to add validation on what keys are supported. that way the implementation contains sufficient information for the maintenance.

test/integration_tests/prototype/ctc_decoder_integration_test.py
```diff
@@ -0,0 +1,15 @@
...
+    # smoke test for constructing and running decoder from pretrained files
+    from torchaudio.prototype.ctc_decoder import lexicon_decoder, download_pretrained_files
+
+    pretrained_files = download_pretrained_files(model)
+    decoder = lexicon_decoder(
+        lexicon=pretrained_files.lexicon,
+        tokens=pretrained_files.tokens,
+        lm=pretrained_files.lm,
+    )
+    decoder(emissions)
```
mthrok, line 15: Can you assert the result, say, if the best hypothesis is reasonable reading?

mthrok:
@carolineechen Could you update https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_ctc_decoder/inference.py to use the pre-trained models added in this PR?

PR #2276, opened 2022-03-12
Fix version numbers of torch and torchaudio in 0.11 doc

nateanl:
When building the doc, I didn't build pytorch 1.11 RC from source, hence the version numbers are all `1.10.0`. 
The version number of `torchaudio` still has commit hash in the tutorials, the PR fixes them to the correct ones.


mthrok (commented):
The fact that the download progress are gone does not look right.

Can you download the doc built by CI from the following and update the page with it?

https://app.circleci.com/pipelines/github/pytorch/audio/9862/workflows/407489d0-f05f-460f-9b03-43dafb8807b5/jobs/571024/artifacts

CircleCI does not allow downloading artifact all at once, so some API trickery is necessary. We should add zipped doc to artifact as well.

https://discuss.circleci.com/t/download-all-artifacts-for-a-build-on-circleci/7217

0.11.0/functional.html
```diff
@@ -1495,13 +1495,6 @@ <h3><span class="hidden-section">spectrogram</span><a class="headerlink" href="#
 <dd class="field-odd"><p>Tensor</p>
 </dd>
 </dl>
-<dl>
-<dt>Tutorials using <code class="docutils literal notranslate"><span class="pre">spectrogram</span></code>:</dt><dd><div class="sphx-glr-thumbcontainer" tooltip="torchaudio implements feature extractions commonly used in the audio domain. They are available..."><div class="figure align-default" id="id13">
```
mthrok, line 1499: I was wondering why this should be removed, but now looking at the tutorial, this should not be in place in the first place. 

It should be either `torchaudio.transforms.Spectrogram()` or `torchaudio.functional.spectrogram()`.

https://pytorch.org/audio/stable/tutorials/audio_feature_extractions_tutorial.html#spectrogram

https://github.com/pytorch/audio/blob/4b47412e91c22d75a1d5f4ea5f71e0ceaa26efd7/examples/tutorials/audio_feature_extractions_tutorial.py#L210-L216

mthrok (approved):
Looks good.

Can you make sure that the `0.11.0` directory was removed once before adding back the correct doc files? (instead of just overwriting the files, so that unrelated files from the original commit, if exist, are removed)

nateanl:
Yes, that's how I updated the docs (i.e., `rm -r 0.11.0, mkdir 0.11.0, cp -r ci_docs/* ./0.11.0/`).

PR #2283, opened 2022-03-18
Add download utility specialized for torchaudio

mthrok:
In recent updates, torchaudio added features that download assets/models from
download.pytorch.org/torchaudio.

To reduce the code duplication, the implementations uses utilities from
``torch.hub``, but still, there are patterns repeated in implementing
the fetch mechanism, notably cache and local file path handling.

This commit introduces the utility function that handles
download/cache/local path management that can be used for 
fetching pre-trained model data.

### DIFF
--- a/torchaudio/utils/download.py
+++ b/torchaudio/utils/download.py
@@ -1,0 +1,89 @@
+import hashlib
+import logging
+from pathlib import Path
+from os import PathLike
+from typing import Union
+
+import torch
+
+
+_LG = logging.getLogger(__name__)
+
+
+def _get_local_path(key):
+    path = Path(torch.hub.get_dir()) / 'torchaudio' / Path(key)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    return path
+
+
+def _download(key, path, progress):
+    url = f"https://download.pytorch.org/torchaudio/{key}"
+    torch.hub.download_url_to_file(url, path, progress=progress)
+
+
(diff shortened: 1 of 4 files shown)


torchaudio/utils/download.py
```diff
@@ -0,0 +1,89 @@
...
+        Currently the valid key values are the route on ``download.pytorch.org/torchaudio``,
+        but this is an implementation detail.
+
+    Returns:
+        str: The path to the asset on the local file system.
+    """
+    path = path or _get_local_path(key)
+
+    if path.exists():
+        _LG.info("The local file (%s) exists. Skipping the download.", path)
```
hwangjeff, line 72: is there an advantage to doing this versus using f-strings?

hwangjeff (approved):
lg

carolineechen (approved):
looks good, thanks for adding this

torchaudio/utils/download.py
mthrok: The official advantage is that if a logging is not triggered, (say, the config level is set to WARNING by client code here so that info is not triggered), then the string substitution is not performed, so that we can avoid constructing the log message string.

https://docs.python.org/3/howto/logging.html#optimization

Personally I think it is negligible, but pylint with default configuration complaints this

https://pylint.pycqa.org/en/latest/messages/warning/logging-fstring-interpolation.html

Issue #2274, opened 2022-03-11
Running lexicon decoder on GPU causes crash on notebook


mmdalix:
### 🐛 Describe the bug

My notebook on kaggle crashes every time I try to run lexicon decoder with emissions on GPU, and the container has to be rebooted.
No exceptions or output will be printed.

I'm using it this way:
```
beam_search_decoder = lexicon_decoder(
    lexicon=lexicon_file,
    tokens=tokens,
    lm=kenlm_file,
    nbest=1,
    beam_size=350,
    lm_weight=LM_WEIGHT,
    word_score=WORD_SCORE,
)
beam_search_decoder(emissions.to('cpu:0'))     #if I dont use .to('cpu:0') it will crash

```

### Versions

!pip3 install torch==1.12.0.dev20220305+cu113 torchaudio==0.12.0.dev20220305+cu113 -f https://download.pytorch.org/whl/nightly/cu113/torch_nightly.html

hwangjeff:
Hi @mmdalix, thanks for reporting this issue. I suspect the cause of the exception is a device mismatch between the input and the `lengths` tensor that's being inferred: https://github.com/pytorch/audio/blob/4b47412e91c22d75a1d5f4ea5f71e0ceaa26efd7/torchaudio/prototype/ctc_decoder/ctc_decoder.py#L144

Here, `lengths` is placed on cpu, so when you provide an input that's placed on gpu, we observe a mismatch.

To test this assumption, could you try explicitly supplying a lengths tensor placed on gpu and see if that resolves the mismatch? e.g.
```
B, T, N = emissions.size()
lengths = torch.full((B,), T)
beam_search_decoder(emissions.to('cuda'), lengths.to('cuda'))
```

mthrok:
CTC decoder access memories directly so it expects the tensor to be contingent and on CPU.
We should perhaps add the check for CPU tensor. cc @carolineechen

mmdalix:
> 

I don't have any idea about how it works under the hood. but it would a lot better if it was designed to be GPU-compatible because it will take a long time when I increase the beam_size parameter

mthrok:
I am not CUDA expert but they say that decoding in GPU is not necessarily faster than CPU because of the nature of the search algorithm. Decoding is a causal operation, which is very difficult, if not impossible, to pallalelize in a way that can take advantage of GPU pallalelism.

mthrok:
Now the CTC decoder code properly rejects CPU Tensors. Thanks for the report.

PR #2289, opened 2022-03-24
make sure inputs live on CPU for ctc decoder

xiaohui-zhang:
Addressing the issue https://github.com/pytorch/audio/issues/2274: 
Raise Runtime errors when the input tensors to the CTC decoder are GPU tensors since the CTC decoder only runs on CPU. Also update the data type check to use "raise" rather than "assert". 

---
Pull Request resolved: https://github.com/pytorch/audio/pull/2289
GitHub Author: xiaohui-zhang <xiaohuizhang@fb.com>

### DIFF
--- a/torchaudio/prototype/ctc_decoder/ctc_decoder.py
+++ b/torchaudio/prototype/ctc_decoder/ctc_decoder.py
@@ -135,10 +135,10 @@ class LexiconDecoder:
             List[List[Hypothesis]]:
                 List of sorted best hypotheses for each audio sequence in the batch.
         """
- 
+
         if emissions.dtype != torch.float32:
             raise ValueError('emissions must be float32.')
-  
+
         if emissions.is_cuda:
             raise RuntimeError('emissions must live on CPU.')
 
--- a/torchaudio/prototype/ctc_decoder/ctc_decoder.py
+++ b/torchaudio/prototype/ctc_decoder/ctc_decoder.py
@@ -127,23 +127,27 @@ class LexiconDecoder:
 
         Args:
             emissions (torch.FloatTensor): tensor of shape `(batch, frame, num_tokens)` storing sequences of
-                probability distribution over labels; output of acoustic model. It must lives on CPU.
+                probability distribution over labels; output of acoustic model. It must live on CPU.
             lengths (Tensor or None, optional): tensor of shape `(batch, )` storing the valid length of
-                in time axis of the output Tensor in each batch. It must lives on CPU.
+                in time axis of the output Tensor in each batch. It must live on CPU.
 
         Returns:
             List[List[Hypothesis]]:
                 List of sorted best hypotheses for each audio sequence in the batch.
         """
-        assert emissions.dtype == torch.float32
+        
+        if emissions.dtype != torch.float32:
+            raise ValueError('emissions must be float32.')
+        
+        if emissions.is_cuda:
+            raise Exception('emissions must live on CPU.')
 
-
+        if lengths is not None and lengths.is_cuda:
+            raise Exception('lengths must live on CPU.')
+        
         B, T, N = emissions.size()
         if lengths is None:
             lengths = torch.full((B,), T)
-
-        assert not emissions.is_cuda
-        assert not lengths.is_cuda
         
         float_bytes = 4
         hypos = []
--- a/torchaudio/prototype/ctc_decoder/ctc_decoder.py
+++ b/torchaudio/prototype/ctc_decoder/ctc_decoder.py
@@ -127,15 +127,18 @@ class LexiconDecoder:
 
         Args:
             emissions (torch.FloatTensor): tensor of shape `(batch, frame, num_tokens)` storing sequences of
-                probability distribution over labels; output of acoustic model
+                probability distribution over labels; output of acoustic model. It must lives on CPU.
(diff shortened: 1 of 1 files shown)


torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -129,20 +129,24 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
...
         """
         assert emissions.dtype == torch.float32
 
+
         B, T, N = emissions.size()
         if lengths is None:
             lengths = torch.full((B,), T)
 
+        assert not emissions.is_cuda
+        assert not lengths.is_cuda
```
mthrok, line 148: The logic it self looks good. Couple of suggestions.

1. For a better UX, could you provide an error message that tell users what to do (as opposed to what was wrong)?
2. Please perform the input validation as soon as possible, before any operation is performed. Otherwise the operation before validation could be wasteful
3. This is not a written rule, but `assert` is more fore internal assertion. (although it is used in L140, which I guess was missed at a review time) So can replace it with `if <condition>: raise <something>`?

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -129,20 +129,24 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
 
         Args:
             emissions (torch.FloatTensor): tensor of shape `(batch, frame, num_tokens)` storing sequences of
-                probability distribution over labels; output of acoustic model
+                probability distribution over labels; output of acoustic model. It must lives on CPU.
```
carolineechen, line 132:
```suggestion
                probability distribution over labels; output of acoustic model. It must live on CPU.
```

xiaohui-zhang:
thanks @carolineechen and @mthrok for the thorough review of my first PR!

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -129,20 +129,28 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
...
                 List of sorted best hypotheses for each audio sequence in the batch.
         """
-        assert emissions.dtype == torch.float32
-
+        
+        if emissions.dtype != torch.float32:
+            raise ValueError('emissions must be float32.')
+        
+        if emissions.is_cuda:
+            raise Exception('emissions must live on CPU.')
```
hwangjeff, line 145: it'd be better to raise a more specific exception, e.g. `RuntimeError` at min

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -129,20 +129,28 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
 
         Args:
             emissions (torch.FloatTensor): tensor of shape `(batch, frame, num_tokens)` storing sequences of
-                probability distribution over labels; output of acoustic model
+                probability distribution over labels; output of acoustic model. It must live on CPU.
```
hwangjeff, line 132: perhaps just "CPU tensor of shape `(batch, frame, num_tokens)` storing sequences of probability distribution over labels; output of acoustic model."? sounds a little more concrete than "live on CPU"

and so forth below

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -129,20 +129,28 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
...
+            raise Exception('emissions must live on CPU.')
+
+        if lengths is not None and lengths.is_cuda:
+            raise Exception('lengths must live on CPU.')
+        
         B, T, N = emissions.size()
         if lengths is None:
             lengths = torch.full((B,), T)
-
+
```
hwangjeff, line 153: can remove extra whitespace

mthrok (approved):

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -126,16 +126,24 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
...
             List[List[Hypothesis]]:
                 List of sorted best hypotheses for each audio sequence in the batch.
         """
-        assert emissions.dtype == torch.float32
+
+        if emissions.dtype != torch.float32:
+            raise ValueError("emissions must be float32.")
+
+        if emissions.is_cuda:
+            raise RuntimeError("emissions must live on CPU.")
```
hwangjeff, line 143: similar nit as above

```suggestion
            raise RuntimeError("emissions must be a CPU tensor.")
```

torchaudio/prototype/ctc_decoder/ctc_decoder.py
```diff
@@ -126,16 +126,24 @@ def __call__(self, emissions: torch.FloatTensor, lengths: Optional[torch.Tensor]
...
+
+        if lengths is not None and lengths.is_cuda:
+            raise RuntimeError("lengths must live on CPU.")
```
hwangjeff, line 146: and here
```suggestion
            raise RuntimeError("lengths must be a CPU tensor.")
```

nateanl (approved):

PR #2290, opened 2022-03-25
Add QUESST14 dataset

carolineechen:
implementation adapted from [s3prl](https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/quesst14_dtw/dataset.py)

modifying the s3prl downstream expert to [this](https://github.com/carolineechen/s3prl/commit/adc91a53d581a604f495f3795a865d84aa17f1a5) using this dataset implementation produces the same results as using the original s3prl pipeline

### DIFF
--- a/torchaudio/datasets/quesst14.py
+++ b/torchaudio/datasets/quesst14.py
@@ -30,8 +30,7 @@ class QUESST14(Dataset):
         language (str, optional): Language to get dataset for.
             Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``].
             (default: ``"nnenglish"``)
-        subset (str or None, optional): subset of the dataset to use. Options: [None, "dev", "eval"].
-            None indicates the whole dataset, including dev and eval. (default: ``None``)
+        subset (str): subset of the dataset to use. Options: ["docs", "dev", "eval"].
         download (bool, optional): Whether to download the dataset if it is not found at root path.
             (default: ``False``)
     """
@@ -43,7 +42,7 @@ class QUESST14(Dataset):
         subset: Optional[str] = None,
         download: bool = False,
     ) -> None:
-        assert subset is None or subset in ["dev", "eval"], "`subset` must be one of [None, 'dev', 'eval']"
+        assert subset is None or subset in ["docs", "dev", "eval"], "`subset` must be one of ['docs', 'dev', 'eval']"
 
         assert language is None or language in _LANGUAGES, f"`language` must be None or one of {str(_LANGUAGES)}"
 
@@ -63,22 +62,16 @@ class QUESST14(Dataset):
                 download_url_to_file(URL, archive, hash_prefix=_CHECKSUM)
             extract_archive(archive, root)
 
-        doc_paths = filter_audio_paths(self._path, language, "language_key_utterances.lst")
-        if subset is None:
-            dev_paths = filter_audio_paths(self._path, language, "language_key_dev.lst")
-            eval_paths = filter_audio_paths(self._path, language, "language_key_eval.lst")
-            query_paths = dev_paths + eval_paths
-        else:
-            query_paths = filter_audio_paths(self._path, language, f"language_key_{subset}.lst")
-
-        self.n_docs = len(doc_paths)
-        self.n_queries = len(query_paths)
-        self.data = query_paths + doc_paths
+        if subset == "docs":
+            self.data = filter_audio_paths(self._path, language, "language_key_utterances.lst")
+        elif subset == "dev":
+            self.data = filter_audio_paths(self._path, language, "language_key_dev.lst")
+        elif subset == "eval":
+            self.data = filter_audio_paths(self._path, language, "language_key_eval.lst")
 
     def _load_sample(self, n: int) -> Tuple[torch.Tensor, str]:
         audio_path = self.data[n]
         wav, _ = torchaudio.load(audio_path)
-        wav = wav.squeeze(0)
         return wav, audio_path.with_suffix("").name
 
     def __getitem__(self, n: int) -> Tuple[torch.Tensor, str]:
--- a/torchaudio/datasets/quesst14.py
+++ b/torchaudio/datasets/quesst14.py
@@ -83,7 +83,6 @@ class QUESST14(Dataset):
             [
                 ["channels", "1"],
                 ["rate", "16000"],
-                ["gain", "-3.0"],
             ],
         )
         wav = wav.squeeze(0)
--- a/torchaudio/datasets/quesst14.py
+++ b/torchaudio/datasets/quesst14.py
@@ -4,10 +4,10 @@ from pathlib import Path
 from typing import Tuple, Union, Optional
 
 import torch
+import torchaudio
 from torch.hub import download_url_to_file
 from torch.utils.data import Dataset
 from torchaudio.datasets.utils import extract_archive
-from torchaudio.sox_effects import apply_effects_file
 
 
 URL = "https://speech.fit.vutbr.cz/files/quesst14Database.tgz"
@@ -78,13 +78,7 @@ class QUESST14(Dataset):
 
     def _load_sample(self, n: int) -> Tuple[torch.Tensor, str]:
         audio_path = self.data[n]
-        wav, _ = apply_effects_file(
-            str(audio_path),
-            [
-                ["channels", "1"],
-                ["rate", "16000"],
-            ],
-        )
+        wav, _ = torchaudio.load(audio_path)
         wav = wav.squeeze(0)
         return wav, audio_path.with_suffix("").name
 
--- a/torchaudio/datasets/quesst14.py
+++ b/torchaudio/datasets/quesst14.py
@@ -27,12 +27,11 @@ class QUESST14(Dataset):
 
     Args:
         root (str or Path): Root directory where the dataset's top level directory is found
-        subset (str): The subset to use. Options: [``dev``, ``eval``]
         language (str, optional): Language to get dataset for.
-            Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``]
-        folder_in_archive (str, optional): The top-level directory of the dataset. default: (``"quesst14Database"``)
+            Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``].
+            (default: ``"nnenglish"``)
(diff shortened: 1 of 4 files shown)


torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,116 @@
...
+        folder_in_archive (str, optional): The top-level directory of the dataset. default: (``"quesst14Database"``)
+        subset (str or None, optional): subset of the dataset to use. Options: [None, "dev", "eval"].
+            None indicates the whole dataset, including dev and eval.
+        download (bool, optional): Whether to download the dataset if it is not found at root path.
+            (default: ``False``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
```
mthrok, line 42: I am wondering if we can use a default path for the download location, similar to pretained model. Sure it will be different from the previous datasets, but I often find that it's more convenient that way. Thoughts?

nateanl: The default path will be tricky if we consider Windows (`C:\`), Linux and MacOS (`/usr/datasets/`).

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,116 @@
...
+    "slovak",
+]
+
+
+class QUESST14(Dataset):
+    """Create QUESST14 Dataset
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        subset (str): The subset to use. Options: [``dev``, ``eval``]
```
nateanl, line 30:
```suggestion
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,116 @@
...
+        language (str, optional): Language to get dataset for.
+            Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``]
+        folder_in_archive (str, optional): The top-level directory of the dataset. default: (``"quesst14Database"``)
```
nateanl, line 33:
```suggestion
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,116 @@
...
+        download (bool, optional): Whether to download the dataset if it is not found at root path.
+            (default: ``False``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        language: str = "nnenglish",
+        download: bool = False,
+        subset: Optional[str] = None,
```
nateanl, line 45:
```suggestion
        subset: Optional[str] = None,
        download: bool = False,
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,116 @@
...
+        self.data = query_paths + doc_paths
+
+    def _load_sample(self, n: int) -> Tuple[torch.Tensor, str]:
+        audio_path = self.data[n]
+        wav, _ = apply_effects_file(
+            str(audio_path),
+            [
+                ["channels", "1"],
+                ["rate", "16000"],
+                ["gain", "-3.0"],
```
nateanl, line 86: Curious, is it the standard way of loading QUESST14 dataset? Could you find a reference that applies the same gain value?

carolineechen: per offline discussion, we can remove these applied sox effects for the dataset, and they can just be applied directly by users using torchaudio's resample/gain functionality as necessary. using torchaudio to perform these effects afterwards rather than sox at load time is also faster, on both cpu and cuda

torchaudio/datasets/quesst14.py
carolineechen: this seems out of scope for this PR, but I agree that it can be good for managing consistent dataset paths and allow us to better maintain cached dataset artifacts, and would not negatively affect users as they can still input their own root path if they have preferences. do you have a suggestion on the path location?

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,109 @@
...
+        download (bool, optional): Whether to download the dataset if it is not found at root path.
+            (default: ``False``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        language: str = "nnenglish",
+        download: bool = False,
+        subset: Optional[str] = None,
```
nateanl, line 45:
```suggestion
        subset: Optional[str] = None,
        download: bool = False,
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,109 @@
...
+    """Create QUESST14 Dataset
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        subset (str): The subset to use. Options: [``dev``, ``eval``]
+        language (str, optional): Language to get dataset for.
+            Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``]
+        folder_in_archive (str, optional): The top-level directory of the dataset. default: (``"quesst14Database"``)
+        subset (str or None, optional): subset of the dataset to use. Options: [None, "dev", "eval"].
+            None indicates the whole dataset, including dev and eval.
```
nateanl, line 35:
```suggestion
            None indicates the whole dataset, including dev and eval. (default: ``None``)
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,109 @@
...
+        if subset is None:
+            dev_paths = filter_audio_paths(self._path, language, "language_key_dev.lst")
+            eval_paths = filter_audio_paths(self._path, language, "language_key_eval.lst")
+            query_paths = dev_paths + eval_paths
+        else:
+            query_paths = filter_audio_paths(self._path, language, f"language_key_{subset}.lst")
+
+        self.n_docs = len(doc_paths)
+        self.n_queries = len(query_paths)
+        self.data = query_paths + doc_paths
```
nateanl, line 77: So `self.data` contain both `doc` and subset. I'm wondering is there a case where only subset is used?

nateanl: The docs and dev subsets are used separately, according to [s3prl](https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/quesst14_dtw/expert.py#L84-L87)
`docs` is the audios for retrieval, `dev` and `test` are query audios. It makes more sense to split `docs` as a separate subset. We should also remove `n_docs` in the dataset, as they can be gotten by `len(dataset)` if the dataset is initialized with `docs` subset. Same for `n_queries`.

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,101 @@
...
+
+class QUESST14(Dataset):
+    """Create QUESST14 Dataset
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        language (str, optional): Language to get dataset for.
+            Options: [None, ``albanian``, ``basque``, ``czech``, `nnenglish``, ``romanian``, ``slovak``].
+            (default: ``"nnenglish"``)
+        subset (str): subset of the dataset to use. Options: ["docs", "dev", "eval"].
```
nateanl, line 33:
```suggestion
        subset (str or None, optional): subset of the dataset to use. Options: ["docs", "dev", "eval"].
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,101 @@
...
+        download (bool, optional): Whether to download the dataset if it is not found at root path.
+            (default: ``False``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        language: str = "nnenglish",
```
nateanl, line 41:
```suggestion
        language: Optional[str] = "nnenglish",
```

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,101 @@
...
+    "slovak",
+]
+
+
+class QUESST14(Dataset):
+    """Create QUESST14 Dataset
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        language (str, optional): Language to get dataset for.
```
nateanl, line 30:
```suggestion
        language (str or None, optional): Language to get dataset for.
```

nateanl (approved):
LGTM. Just some nits and we can merge it. Thanks!

torchaudio/datasets/quesst14.py
carolineechen: I'm thinking of keeping subset as a required parameter, and not allowing `None` as it is a bit vague given there is the `docs` subset. It seems `docs` would generally be extracted separately and used differently than the `dev` and `eval` set, so supporting a dataset that could support docs+dev only or dev+eval only etc seems like it could be messy without having additional class variables like len(docs). I think the user can just use `ConcatDataset` if they'd like to merge multiple subsets, thoughts?

nateanl: Yeah that's also what I'm thinking. making it a required parameter is a good idea.

torchaudio/datasets/quesst14.py
```diff
@@ -0,0 +1,101 @@
...
+    "albanian",
+    "basque",
+    "czech",
+    "nnenglish",
+    "romanian",
+    "slovak",
+]
+
+
+class QUESST14(Dataset):
```
mthrok, line 25: @carolineechen Can we have `Quesst14` as the name of the dataset?
I understand that the original name is all capital but there is no benefit going against PEP-8's naming convention. Currently `torchaudio.datasets` has couple of datasets named like that but that's definitely not a standard and we should not make it a standard.

carolineechen: I agree we shouldn't default to all caps for dataset naming as we have done with several datasets in the past, but as this dataset's original name is already all capitalized (["QUESST 2014 Multilingual..."](https://speech.fit.vutbr.cz/software/quesst-2014-multilingual-database-query-by-example-keyword-spotting)), I think it makes sense for this to remain all caps, and to follow the dataset's conventional naming. What do you think?

mthrok: I think this is where the sticking with the PEP-8 naming convention matters.
It is unfortunate that existing datasets violate them and we are not gonna change it for the sake of changing it, but
as we add more datasets, we should stick with PEP-8. 

The reason is that boundary of allowing all caps and not are kinda arbitrary this manner.
This can lead to discussion and disagreement every time we add datasets
 It's like formatting issue, where subjectivity causes disagreement and make the collaboration unproductive.
Sticking with PEP-8 allows us to reason why more clearly. If all caps are required from technical reason, it will be clearer.

PR #2296, opened 2022-03-29
Fix CollateFn in HuBERT pre-training recipe

nateanl:
- When cropping the waveform and corresponding label, we use the formula `torch.div(audio_start - kernel_size * sample_rate, stride * sample_rate, rounding_mode="floor")` to align the audio start and label start indices. However, sometimes the value can be negative, which result in an empty label. The training example will hurt the performance after zero-padding (i.e., the labels are all zero for the input waveform).
This PR fixes the bug by checking if `label_start` is negative, and change it to zero if so.
- If `pad` is True, the `length` should be the length of each waveform instead of the max length. Fix it to make the model ignore the padding component in pre-training.


examples/hubert/dataset/hubert_dataset.py
```diff
@@ -250,7 +250,7 @@ def __init__(
         self.pad = pad
         self.rand_crop = rand_crop
 
-    def __call__(self, batch: Tuple[Tensor, Tensor, int]) -> Tuple[Tensor, Tensor, Tensor]:
+    def __call__(self, batch: List[Tuple[Tensor, Tensor, int]]) -> Tuple[Tensor, Tensor, Tensor]:
```
mthrok, line 253: Is this bugfix or spec change?

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -309,14 +310,20 @@ def _collate_audio_label(
         kernel_size = 25
         stride = 20
         sample_rate = 16  # 16 per millisecond
-        if waveform.shape[1] > audio_size:
-            diff = waveform.size(1) - audio_size
+        audio_start = 0
+        waveform = waveform[0]
+        if waveform.shape[0] > audio_size:
+            diff = waveform.size(0) - audio_size
             audio_start = torch.randint(diff, size=(1,)) if rand_crop else 0
```
mthrok, line 317: If `rand_crop` is falsy value, this does not change the value of `audio_start`, thus the computation of `diff` becomes redundant. I think updating the condition will express the logic better.

```python
if waveform.shape[1] > audio_size and rand_crop:
    diff = waveform.size(0) - audio_size
    audio_start = torch.randint(diff, size=(1,))
```

examples/hubert/dataset/hubert_dataset.py
nateanl: It's a bug fix. The batch should be a list of tuples instead of one tuple.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -276,13 +276,14 @@ def __call__(self, batch: Tuple[Tensor, Tensor, int]) -> Tuple[Tensor, Tensor, T
...
+            labels = torch.nn.utils.rnn.pad_sequence(labels, batch_first=True)
+        else:
+            waveforms = torch.stack(waveforms)
+            labels = torch.stack(labels)
         lengths = torch.tensor(lengths)
-        labels = torch.nn.utils.rnn.pad_sequence(labels, batch_first=True)
-        return data, labels, lengths
+        return waveforms, labels, lengths
 
     def _collate_audio_label(
```
hwangjeff, line 288: "collate" doesn't accurately describe what the function does. perhaps `_crop_audio_label`?

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -309,14 +310,20 @@ def _collate_audio_label(
...
-            waveform = waveform[:, audio_start : audio_start + audio_size]
-            label = label[label_start : label_start + label_size]
-            length = audio_size
+        else:
+            audio_size = waveform.shape[0]
+        label_start = torch.max(
+            torch.div(audio_start - kernel_size * sample_rate, stride * sample_rate, rounding_mode="floor") + 1,
+            torch.tensor([0]),
+        )
+        label_size = torch.div(audio_size - kernel_size * sample_rate, stride * sample_rate, rounding_mode="floor") + 1
```
hwangjeff, line 324: is this correct when `audio_size - kernel_size * sample_rate` is a multiple of `stride * sample_rate`?

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -276,13 +276,14 @@ def __call__(self, batch: Tuple[Tensor, Tensor, int]) -> Tuple[Tensor, Tensor, T
             waveforms.append(waveform)
             lengths.append(length)
             labels.append(label)
-
-        data = torch.zeros(len(batch), audio_size)
-        for i in range(len(waveforms)):
-            data[i][0 : waveforms[i].shape[1]] = waveforms[i][0]
+        if self.pad:
+            waveforms = torch.nn.utils.rnn.pad_sequence(waveforms, batch_first=True)
+            labels = torch.nn.utils.rnn.pad_sequence(labels, batch_first=True)
```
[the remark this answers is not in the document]
nateanl: The logic here is to guarantee that Tensors in the `waveforms` have the same shape so that it can be stacked.

examples/hubert/dataset/hubert_dataset.py
nateanl: Yes, it's correct as there is padding in the convolutional layer of wav2vec2 and hubert models. The way to compute the label_size is from the [length computation in wav2vec2](https://github.com/pytorch/audio/blob/main/torchaudio/models/wav2vec2/components.py#L64).

mthrok: Can you add unit test?

nateanl: sure

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -309,14 +310,20 @@ def _collate_audio_label(
...
-            label_size = torch.div(audio_size - kernel_size * sample_rate, stride * sample_rate, rounding_mode="floor")
-            waveform = waveform[:, audio_start : audio_start + audio_size]
-            label = label[label_start : label_start + label_size]
-            length = audio_size
+        else:
+            audio_size = waveform.shape[0]
+        label_start = torch.max(
+            torch.div(audio_start - kernel_size * sample_rate, stride * sample_rate, rounding_mode="floor") + 1,
+            torch.tensor([0]),
+        )
```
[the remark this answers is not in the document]
hwangjeff: to the point above, why not simply use conventional python arithmetic operations?

examples/hubert/dataset/hubert_dataset.py
mthrok: > is this correct when `audio_size - kernel_size * sample_rate` is a multiple of `stride * sample_rate`?

@nateanl `audio_size` is not precise naming. (it could mean size in bytes from other software engineers.) Can you use `frames` to refer to the number of frames? In torchaudio, the standard is to use `frame` when referring to the length of audio in channel-agnostic way, and `sample` to refer to the number of samples across channels. (i.e. `num_frames` * `num_channels` == `num_samples`)

test/torchaudio_unittest/example/hubert/test_crop_audio_label.py
```diff
@@ -0,0 +1,46 @@
...
+        super().setUpClass()
+        torch.random.manual_seed(31)
+
+    @parameterized.expand(
+        [
+            (400,),
+            (800,),
+        ]
+    )
+    def test_zero_offset(self, num_frames):
```
mthrok, line 21: Could you add description of what "you" intend to test here, so that future maintainers without context can understand it easily.

In general, bugs in tests are harder to detect, so having description of "(under condition C) doing A should produce B", help a lot. Without such description, future maintainers have to 1. first figure out what test does, 2. try to make sense of it, and 3. debug. Which is very difficult.

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -303,6 +304,46 @@ def __getitem__(self, index):
...
+def _crop_audio_label(
+    waveform: Tensor,
+    label: Tensor,
+    length: Tensor,
+    num_frames: int,
+    rand_crop: bool,
+) -> Tuple[Tensor, Tensor, Tensor]:
+    """Collate the audio and label at the same time.
+    Args:
+        waveform (Tensor): The waveform Tensor of dimension `[1, time]`.
```
hwangjeff, line 316: nit:
```suggestion
        waveform (Tensor): The waveform Tensor of dimensions `(1, time)`.
```

test/torchaudio_unittest/example/hubert/test_crop_audio_label.py
```diff
@@ -0,0 +1,60 @@
...
+        """Test _crop_audio_label method with zero frame offset.
+        Given the ``num_frames`` argument, the method returns the first ``num_frames`` samples in the waveform,
+        the corresponding label, and the length of the cropped waveform.
+        The cropped waveform should be identical to the first ``num_frames`` samples of original waveform.
+        The length of the cropped waveform should be identical to ``num_frames``.
+        The shape of label should be identicial to the frame dimension of HuBERT transformer feature.
```
hwangjeff, line 26:
```suggestion
        The shape of label should be identical to the frame dimensions of HuBERT transformer feature.
```

what does "HuBERT transformer feature" mean?

test/torchaudio_unittest/example/hubert/test_crop_audio_label.py
```diff
@@ -0,0 +1,60 @@
...
+    @parameterized.expand(
+        [
+            (400,),
+            (800,),
+        ]
+    )
+    def test_zero_offset(self, num_frames):
+        """Test _crop_audio_label method with zero frame offset.
+        Given the ``num_frames`` argument, the method returns the first ``num_frames`` samples in the waveform,
+        the corresponding label, and the length of the cropped waveform.
```
hwangjeff, line 23: should it be "labels"?

test/torchaudio_unittest/example/hubert/test_crop_audio_label.py
```diff
@@ -0,0 +1,60 @@
...
+            (400,),
+            (800,),
+        ]
+    )
+    def test_rand_crop(self, num_frames):
+        """Test _crop_audio_label method with random frame offset.
+        Given the ``num_frames`` argument, the method returns ``num_frames`` samples in the waveform
+        starting with random offset, the corresponding label, and the length of the cropped waveform.
+        The length of the cropped waveform should be identical to ``num_frames``.
+        The shape of label should be identicial to the frame dimension of HuBERT transformer feature.
```
hwangjeff, line 50: same comment as above

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -339,61 +380,26 @@ def __call__(self, batch: Tuple[Tensor, Tensor, int]) -> Tuple[Tensor, Tensor, T
...
             labels.append(label)
-
-        data = torch.zeros(len(batch), audio_size)
-        for i in range(len(waveforms)):
-            data[i][0 : waveforms[i].shape[1]] = waveforms[i][0]
-        lengths = torch.tensor(lengths)
+        # make sure the shapes are the same if not apply zero-padding
+        if not self.pad:
+            assert all([waveform.shape[0] == waveforms[0].shape[0] for waveform in waveforms])
+            assert all([label.shape[0] == labels[0].shape[0] for label in labels])
```
mthrok, line 401: Can you give error message, so that users hitting this error know what to do to fix the error.

mthrok (approved):

test/torchaudio_unittest/example/hubert/test_crop_audio_label.py
hwangjeff: i see. yeah, dimension should be singular in that case. how about "HuBERT transformer layer output frame dimension" then?

but, if we go with that, "shape" is no longer compatible in the sentence, so we'd have to fix that too

nateanl: okay, sounds good.

Issue #2295, opened 2022-03-29
Build error, cannot fetch zlib-1.2.11.tar.gz


dusty-nv:
### 🐛 Describe the bug

```bash
git clone --recursive -b v0.11.0 https://github.com/pytorch/audio torchaudio
cd torchaudio
BUILD_SOX=1 python3 setup.py install
```

This was working fine up until a day or two ago.  Then it started to not be able to download `zlib-1.2.11.tar.gz`.  I don't believe it's a connection issue on my end because I am not experiencing other network issues.

``` bash
-- Git tag: v0.11.0
-- PyTorch dependency: torch
-- Building version 0.11.0+820b383
 --- Initializing submodules
 --- Initialized submodule
 --- Fetching zlib-1.2.11.tar.gz
Traceback (most recent call last):
  File "setup.py", line 181, in <module>
    _main()
  File "setup.py", line 144, in _main
    _fetch_third_party_libraries()
  File "setup.py", line 128, in _fetch_third_party_libraries
    _fetch_archives(_parse_sources())
  File "setup.py", line 122, in _fetch_archives
    torch.hub.download_url_to_file(url, dest, progress=False)
  File "/usr/local/lib/python3.8/dist-packages/torch/hub.py", line 457, in download_url_to_file
    u = urlopen(req)
  File "/usr/lib/python3.8/urllib/request.py", line 222, in urlopen
    return opener.open(url, data, timeout)
  File "/usr/lib/python3.8/urllib/request.py", line 531, in open
    response = meth(req, response)
  File "/usr/lib/python3.8/urllib/request.py", line 640, in http_response
    response = self.parent.error(
  File "/usr/lib/python3.8/urllib/request.py", line 569, in error
    return self._call_chain(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 502, in _call_chain
    result = func(*args)
  File "/usr/lib/python3.8/urllib/request.py", line 649, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 404: Not Found
```

### Versions

Note:  this build is being done in docker, whereas collect_env.py was run outside docker, so not all versions are showing appropriately.  It is with PyTorch 1.11

```
python3 collect_env.py
Collecting environment information...
PyTorch version: N/A
Is debug build: N/A
CUDA used to build PyTorch: N/A
ROCM used to build PyTorch: N/A

[...]

PR #2297, opened 2022-03-29
Use sourceforge url to fetch zlib

nateanl:
This PR addresses #2295 by updating `zlib`'s url to the one on sourceforge.net.
`zlib` 1.2.11 source code is removed from the official site. According to https://zlib.net, ```Due to the bug fixes, any installations of 1.2.11 should be replaced with 1.2.12.```
sourceforge preserves the older versions thus is more stable. The PR keep 1.2.11 as currently there is no 1.2.12 on sourceforge.

### DIFF
--- a/third_party/zlib/CMakeLists.txt
+++ b/third_party/zlib/CMakeLists.txt
@@ -22,8 +22,8 @@ set(
 ExternalProject_Add(zlib-
   PREFIX ${CMAKE_CURRENT_BINARY_DIR}
   DOWNLOAD_DIR ${ARCHIVE_DIR}
-  URL https://zlib.net/zlib-1.2.11.tar.gz
-  URL_HASH SHA256=c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
+  URL https://zlib.net/zlib-1.2.12.tar.gz
+  URL_HASH SHA256=91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9
   BUILD_BYPRODUCTS ${ZLIB_LIBRARIES}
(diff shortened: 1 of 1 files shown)


third_party/zlib/CMakeLists.txt
```diff
@@ -22,8 +22,8 @@ set(
 ExternalProject_Add(zlib-
   PREFIX ${CMAKE_CURRENT_BINARY_DIR}
   DOWNLOAD_DIR ${ARCHIVE_DIR}
-  URL https://zlib.net/zlib-1.2.11.tar.gz
-  URL_HASH SHA256=c3e5e9fdd5004dcb542feda5ee4f0ff0744628baf8ed2dd5d66f8ca1197cb1a1
+  URL https://zlib.net/zlib-1.2.12.tar.gz
+  URL_HASH SHA256=91844808532e5ce316b3c010929493c0244f3d37593afd6de04f71821d5136d9
```
mthrok, line 26: Although this is correct. Can we actually keep 1.2.11 and instead update the URL to one on sourceforge? i.e. `URL https://sourceforge.net/projects/libpng/files/zlib/1.2.11/zlib-1.2.11.tar.gz`

Here are the reasons.
1. We learned that zlib.net would delete previous releases when there is a new release. This gives concerns because when there is a new release of zlib, then the source build would break again. Due to this, the latest versioned release (0.11) cannot be built from source. If we rely on sourceforge.net, then we can expect that old releases are preserved. 
2. Currently there is no 1.2.12 on sourceforge. https://sourceforge.net/projects/libpng/files/zlib/ We need to wait for them to upload 1.2.12 to source forge.

mthrok (approved):
Looks good. Once landed, can you cherry-pick this to `release/0.11` branch?

PR #2298, opened 2022-03-29
v1.2.11 deleted; https://www.zlib.net/

spital:
Due to the bug fixes, any installations of 1.2.11 should be replaced with 1.2.12.


mthrok:
Thanks @spital for the PR. This has been handled by #2297 and #2300,

PR #2299, opened 2022-03-30
Introduce DistributedBatchSampler

nateanl:
When using customized `batch_sampler`, pytorch_lightning can't wrap the distributed sampler onto it. Hence we provide a `DistributedBatchSampler` that supports `BucketizeBatchSampler` in `ddp` mode.

The `DistributedBatchSampler` assumes `BucketizeBatchSampler.iter_list` is a list of lists, where each sub-list contains a batch of indices. Setting `shuffle` to `True` will shuffle the lists based on `seed` and current `epoch`.

The `shuffle` only happens in the initialization, and won't be changed if user don't reset it. The reason is shuffling `BucketizeBatchSampler` may have a different length than before, do shuffling in ``__iter__`` may result in mismatch between ``__len__`` and the real length value.
Hence users need to set `reload_dataloaders_every_n_epochs=1` in pytorch_lightning's Trainer. Then the value of ``__len__``  and the real length is the same.

### DIFF
--- a/examples/hubert/dataset/hubert_dataset.py
+++ b/examples/hubert/dataset/hubert_dataset.py
@@ -3,9 +3,10 @@ from typing import Dict, Iterator, List, Optional, Tuple, Union
 
 import numpy as np
 import torch
+import torch.distributed as dist
 import torchaudio
 from torch import Tensor
-from torch.utils.data import BatchSampler, Dataset
+from torch.utils.data import BatchSampler, Dataset, DistributedSampler
 
 
 class BucketizeBatchSampler(BatchSampler):
@@ -34,6 +35,10 @@ class BucketizeBatchSampler(BatchSampler):
 
     Note:
         ``drop_last`` is only valid when ``batch_size`` argument is given.
+
+    Note:
+        if ``shuffle`` is True, it will only shuffle the data once. Please set ``reload_dataloaders_every_n_epochs=1``
+        in pytorch_lightning Trainer to enable shuffling every epoch.
     """
 
     def __init__(
@@ -72,7 +77,6 @@ class BucketizeBatchSampler(BatchSampler):
         self.shuffle = shuffle
         self.drop_last = drop_last
         self.buckets = self._get_buckets(self.lengths, num_buckets, min_len, max_len)
-        self.iter_list = []
         self._update_iter_list()
 
     def _get_buckets(self, lengths: List[int], num_buckets: int, min_len: int, max_len: int) -> Dict[int, Tensor]:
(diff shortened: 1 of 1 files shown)


examples/hubert/dataset/hubert_dataset.py
```diff
@@ -121,18 +128,87 @@ def _update_iter_list(self) -> None:
...
+                raise RuntimeError("Requires distributed package to be available")
+            num_replicas = dist.get_world_size()
+        if rank is None:
+            if not dist.is_available():
+                raise RuntimeError("Requires distributed package to be available")
+            rank = dist.get_rank()
+        self.num_replicas = num_replicas
+        self.rank = rank
+        self.shuffle = shuffle
+        self.epoch = 0
```
hwangjeff, line 185: who updates this value?

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -121,18 +128,87 @@ def _update_iter_list(self) -> None:
...
+            self.total_size = len(indices)
+        self.num_samples = self.total_size // self.num_replicas
+        self.subset = indices[self.rank : self.total_size : self.num_replicas]
+        assert len(self.subset) == self.num_samples
+
+    def __iter__(self):
+        return iter(self.subset)
+
+    def __len__(self):
+        return self.total_size // self.num_replicas
```
hwangjeff, line 209:
```suggestion
        return self.num_samples
```

examples/hubert/dataset/hubert_dataset.py
```diff
@@ -101,7 +105,10 @@ def _get_buckets(self, lengths: List[int], num_buckets: int, min_len: int, max_l
         buckets = {k: v for k, v in sorted(buckets.items())}
         return buckets
 
-    def _update_iter_list(self) -> None:
+    def _update_iter_list(self, generator=None) -> None:
```
hwangjeff, line 108: do we need kwarg `generator`?

examples/hubert/dataset/hubert_dataset.py
nateanl: The class inherits from [DistributedSampler](https://github.com/pytorch/pytorch/blob/d74bb42f7ad347181f5ac60d27889739c5a79bf4/torch/utils/data/distributed.py#L125), `epoch` can be updated by `set_epoch()` in pytorch_lightning.

examples/hubert/dataset/hubert_dataset.py
nateanl: I was thinking shuffling `BucketizeBatchSampler` with the same generator guarantees each GPU gets the same `iter_list`, after second thought, `BucketizeBatchSampler` is already initialized when it is passed to `DistributedBatchSampler`. So we don't need to worry about it. Will remove it, thanks.

hwangjeff (approved):

PR #2301, opened 2022-03-31
Add randomness to test sinusoid data

mthrok:
This commit update `get_sinusoid` function in test utility so that
when a multi channel is requested, non-primal channel have randomized
initial phase.

This adds variety in test data. Currently `get_sinusoid` returns identical
waveforms for all the channels. The multi channel support was added just to
mock the input data so that it is easy to test features with multi-channel
inputs.

When working on numerical parity, it is more useful if the raw waveforms
are somewhat different.


test/torchaudio_unittest/common_utils/data_utils.py
```diff
@@ -106,8 +106,13 @@ def get_sinusoid(
...
     pie2 = 2 * 3.141592653589793
     end = pie2 * frequency * duration
-    theta = torch.linspace(0, end, int(sample_rate * duration), dtype=torch.float32, device=device)
-    tensor = torch.sin(theta, out=None).repeat([n_channels, 1])
+    num_frames = int(sample_rate * duration)
+    # Randomize the initial phase. (except the first channel)
+    theta0 = pie2 * torch.randn(n_channels, 1, dtype=torch.float32, device=device)
+    theta0[0, :] = 0
+    theta = torch.linspace(0, end, num_frames, dtype=torch.float32, device=device)
+    theta = theta0.repeat(1, num_frames) + theta.unsqueeze(0)
```
hwangjeff, line 114:
```suggestion
    theta = theta0 + theta
```

should be functionally identical and would probably be more performant

hwangjeff (approved):

PR #2302, opened 2022-03-31
Add LibriLightLimited dataset

nateanl:
The `LibriLightLimited` dataset is created for fine-tuning SSL models, such as Wav2Vec2 and HuBERT. It is a supervised subset of [Libri-Light](https://github.com/facebookresearch/libri-light) dataset. To distinguish the unsupervised subset and the supervised one, it's clearer to put it in a separate dataset class for fine-tuning purpose.
It contains "10 min", "1 hour", "10 hour" splits.


torchaudio/datasets/librispeech_finetune.py
```diff
@@ -0,0 +1,100 @@
...
+import torchaudio
+from torch import Tensor
+from torch.hub import download_url_to_file
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import (
+    extract_archive,
+)
+
+
+FOLDER_IN_ARCHIVE = "librispeech_finetuning"
```
mthrok, line 14: Making this `FOLDER_INARCHIVE` global variable, is a practice that does not have clear reasoning / advantage, so can we simply not use this? (same goes to `folder_in_archive` argument in the constructor.)

torchaudio/datasets/librispeech_finetune.py
```diff
@@ -0,0 +1,100 @@
...
+URL = "https://dl.fbaipublicfiles.com/librilight/data/librispeech_finetuning.tgz"
+CHECKSUM = "5d1efdc777b548194d7e09ba89126e2188026df9fd57aa57eb14408d2b2342af"
```
mthrok, line 16: Could you prefix these with underscore?

torchaudio/datasets/librispeech_finetune.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _get_files(path, split, _ext_audio):
+    if split == "10min":
+        files = sorted(str(p) for p in Path(path).glob("1h/0/*/*/*/*" + _ext_audio))
+    elif split in ["1h", "10h"]:
+        files = [str(p) for p in Path(path).glob("1h/*/*/*/*/*" + _ext_audio)]
```
mthrok, line 23: Does this work with Windows?

torchaudio/datasets/librispeech_finetune.py
```diff
@@ -0,0 +1,100 @@
...
+        waveform,
+        sample_rate,
+        transcript,
+        int(speaker_id),
+        int(chapter_id),
+        int(utterance_id),
+    )
+
+
+class LibriSpeechFineTune(Dataset):
```
mthrok, line 67: Although I understand that this dataset archive was created to facilitate fine-tuning, I am reluctant to call it `FineTune`, because the dataset itself can be used for other purposes.

I do not have an alternative suggestion, but the would it be possible to call something else?

nateanl: ESPnet uses [`LIBRILIGHT_LIMITED `](https://github.com/espnet/espnet/blob/master/egs2/librilight_limited/asr1/) as the dataset name (for the reference).

torchaudio/datasets/librispeech_finetune.py
nateanl: Yeah, seems it passes the unit test.

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,99 @@
...
+        int(speaker_id),
+        int(chapter_id),
+        int(utterance_id),
+    )
+
+
+class LibriLightLimited(Dataset):
+    _ext_txt = ".trans.txt"
+    _ext_audio = ".flac"
+
```
carolineechen, line 69: can you add docstrings here?

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,99 @@
...
+        files = [str(p) for p in Path(path).glob("1h/*/*/*/*/*" + _ext_audio)]
+        if split == "10h":
+            files += [str(p) for p in Path(path).glob("9h/*/*/*/*" + _ext_audio)]
+        files = sorted(files)
+    else:
+        raise ValueError(f"Unsupported split value. Found {split}.")
+    return files
+
+
+def _load_item(file_path: str, ext_audio: str, ext_txt: str) -> Tuple[Tensor, int, str, int, int, int]:
```
carolineechen, line 31: thoughts on importing and reusing `load_librispeech_item` from the librispeech dataset instead of reimplementing the same logic?

nateanl: The difference is that `LibriLightLimited` contains parent folders `1h` and `9h`, which is not compatible with the file path in `load_librispeech_item`. I can refactor `load_librispeech_item` to fit both datasets, but that may make viewing the code inconvenient.

nateanl: The structure of `LibriSpeech` dataset is like `subset/speaker_id/chapter_id/file.flac`, while `LibriLightLimited` dataset looks like `1h/[0-5]/[clean, other]/speaker_id/chapter_id/file.flac` for 1 hour directory, and `9h/[clean, other]/speaker_id/chapter_id/file.flac` for 9 hour directory.

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,111 @@
...
+        root: Union[str, Path],
+        subset: str = "10min",
+        download: bool = False,
+    ) -> None:
+        assert subset in ["10min", "1h", "10h"], "`subset` must be one of ['10min', '1h', '10h']"
+
+        root = os.fspath(root)
+        self._path = os.path.join(root, "librispeech_finetuning")
+        archive = os.path.join(root, "librispeech_finetuning" + ".tgz")
+        if download:
```
carolineechen, line 92: could you throw a runtime error if the dataset can not be found locally and download = False?

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,110 @@
...
+from torch.hub import download_url_to_file
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import extract_archive
+
+
+_URL = "https://dl.fbaipublicfiles.com/librilight/data/librispeech_finetuning.tgz"
+_CHECKSUM = "5d1efdc777b548194d7e09ba89126e2188026df9fd57aa57eb14408d2b2342af"
+
+
+def _get_files(path, subset, _ext_audio):
```
hwangjeff, line 16: why add the leading underscore in `_ext_audio`?

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,110 @@
...
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str, optional): The subset to use. Options: [``10min`, ``1h``, ``10h``]
+            (Default: ``10min``).
+        download (bool, optional):
+            Whether to download the dataset if it is not found at root path. (default: ``False``).
+    """
+
+    _ext_txt = ".trans.txt"
+    _ext_audio = ".flac"
```
hwangjeff, line 77: what's the reasoning behind making these values class variables while making `_URL` and `_CHECKSUM` module-level variables?

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,110 @@
...
+        self,
+        root: Union[str, Path],
+        subset: str = "10min",
+        download: bool = False,
+    ) -> None:
+        assert subset in ["10min", "1h", "10h"], "`subset` must be one of ['10min', '1h', '10h']"
+
+        root = os.fspath(root)
+        self._path = os.path.join(root, "librispeech_finetuning")
+        archive = os.path.join(root, "librispeech_finetuning" + ".tgz")
```
hwangjeff, line 89: can we define a variable for "librispeech_finetuning" and use it in both of these lines?

torchaudio/datasets/librilight_limited.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Load text
+    with open(file_text) as ft:
+        for line in ft:
+            fileid_text, transcript = line.strip().split(" ", 1)
+            if fileid_audio == fileid_text:
+                break
+        else:
+            # Translation not found
+            raise FileNotFoundError("Translation not found for " + fileid_audio)
```
hwangjeff, line 52: f-string

torchaudio/datasets/librilight_limited.py
nateanl: sure, in LibriSpeech's `load_librispeech_item`, `path` is the directory of the desired subset, for example, `/root/dev-clean/`, `fileid` is the filename with format `speaker_id-chapter_id-utterance_id.flac`, with that we can find the location of transcription file, which is `speaker_id-chapter_id.trans.txt'.

In LibriLightLimited, the `path` is not general to every subset. `10min` subset uses `/root/1h/0/...` directory, while `1h` subset uses `/root/1h/`, and `10h` subset uses both `/root/1h/` and `/root/9h/`. And because of the `clean` and `other` subdirectories and `[0-5]` subdirectories in `1h`, we can't reconstruct the path of transcription based on `fileid`.

I can refactor LibriSpeech class so that they can use the same method here, what do you think? Notice if we do that, we need to change the usage in librispeech recipes, like https://github.com/pytorch/audio/blob/main/examples/asr/emformer_rnnt/librispeech/lightning.py#L32

hwangjeff: got it, i think i understand. so, i suppose if we want to use `load_librispeech_item` as is in `__getitem__`, it seems like we could extract the fileid and the dirname preceding the speaker id and chapter id from `file_path`, and then call `load_librispeech_item` on those. would that work?

nateanl: I think so, the `path` argument needs to be tuned per `fileid`, not like in LibriSpeech which can use `self._path` for every file.

torchaudio/datasets/librilight_limited.py
mthrok: The flac part particularly is so that in unittest we can mock the dataset with WAV files.  In test, we do not want to use torchaudio's I/O module because it will make the test depend on not only the dataset implementation but also on I/O module.

Now, without our own I/O module, there aren't many tools that provide nice FLAC support. (PySoundFile can, but it also depends on installation)

So in test, we generate mock data with WAV format and overwrite the audio extension  for the duration of test.

mthrok: I do not see why `.trans.txt` should be class variable.

mthrok (approved):

torchaudio/datasets/librilight_limited.py
nateanl: For `.trans.txt` we can make a separate PR to address for all datasets, such as LibriSpeech and CommonVoice.

torchaudio/datasets/librilight_limited.py
carolineechen: @nateanl hmm ok, so currently this is what happens for quesst14:
-  if neither the folder or archive (.tgz) for the dataset exist, it will either throw a download error or correctly download and extract the files, which is the intended behavior
- if the folder doesn't exist but the archive exists, it will extract the archive files and therefore create the folder

for the second case, you're saying the archive file may not be correct and we extract incorrect files. To get around this, we could raise the download error if `self._path` does not exist, which means the user would download and extract the files themselves, then pass in the root corresponding to it. should we expect the users to extract the files themselves, and take the approach above?

I do think this is an issue we can't really predict, since even if they do have a folder with the correct name, we have no way of telling if the contents of the folder are expected without performing added checks for folder structure/files.

PR #2310, opened 2022-04-01
Fix loading checkpoint in hubert preprocessing

nateanl:
When checkpoint is on GPU device and preprocessing is on CPU, the script will throw an exception error. Fix it to load the model state dictionary into CPU by default.

### DIFF
--- a/examples/hubert/utils/feature_utils.py
+++ b/examples/hubert/utils/feature_utils.py
@@ -114,7 +114,7 @@ def _load_state(model: Module, checkpoint_path: Path) -> Module:
     Returns:
         (Module): The pretrained model.
     """
-    state_dict = torch.load(checkpoint_path)
+    state_dict = torch.load(checkpoint_path, map_location=torch.device("cpu"))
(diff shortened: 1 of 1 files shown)


examples/hubert/utils/feature_utils.py
```diff
@@ -114,7 +114,7 @@ def _load_state(model: Module, checkpoint_path: Path) -> Module:
     Returns:
         (Module): The pretrained model.
     """
-    state_dict = torch.load(checkpoint_path)
+    state_dict = torch.load(checkpoint_path, map_location=torch.device("cpu"))
```
mthrok, line 117: This function is used as the following, 

```python
model = _load_state(model, checkpoint_path) 
model.to(device) 
```

so moving model first and loading the state dict there seems to be as straight forward.

```python
def _load_state(model: module, checkpoint_path: Path, device=torch.device("cpu")):
    state_dict = torch.load(checkpoint_path, map_location=device)
     ...
```

then change the call site

```python
model.to(device) 
_load_state(model, checkpoint_path, device) 
```

mthrok (approved):

PR #2311, opened 2022-04-01
Disable multiprocessing when dumping features in hubert preprocessing

nateanl:
The multi-processing works well on MFCC features. However, it sometimes makes the script hang when dumping HuBERT features. Change it to for-loop resolves the issue.

### DIFF
--- a/examples/hubert/preprocess.py
+++ b/examples/hubert/preprocess.py
@@ -8,7 +8,6 @@ The script includes:
 """
 import logging
 from argparse import ArgumentParser, RawTextHelpFormatter
-from multiprocessing import Pool
 from pathlib import Path
 
 import torch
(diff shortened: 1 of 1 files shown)


mthrok:
> The multi-processing works well on MFCC features. However, it sometimes makes the script hang when dumping HuBERT features. Change it to for-loop resolves the issue.

Sounds like oversubscription issue.
What happens if you set the number of threads to 1?

https://jdhao.github.io/2020/07/06/pytorch_set_num_threads/

nateanl:
That's what I was using. `OMP_NUM_THREADS=1 `
I was using it for training kmeans model with mfcc features, because there is some warning message shows
```
OpenBLAS Warning : Detect OpenMP Loop and this application may hang. Please rebuild the library with USE_OPENMP=1 option.
```
With `OMP_NUM_THREADS=1`, the kmeans method is working.
However, with `OMP_NUM_THREADS=1`, the script still hangs in the dumping hubert feature step.

nateanl:
oops, I misread the option. I used `OMP_NUM_THREADS` instead of `OM_NUM_THREADS`, let me try with it with multi-processing.

mthrok:
`OM_NUM_THREADS` is typo. `OMP_NUM_THREADS` is correct.

mthrok (approved):

PR #2315, opened 2022-04-01
Fix arguments in CTC decoding script

nateanl:
Some arguments in `ArgumentParser` are not used in the `lexicon_decoder`. Fix them to use the ones in the parser.


carolineechen (approved):
LGTM, thanks for catching this!

PR #2319, opened 2022-04-04
Add an option to use Tanh instead of ReLU in RNNT joiner

xiaohui-zhang:
Add an option to use Tanh instead of ReLU in RNNT joiner, which enables better training performance sometimes.

---
Pull Request resolved: https://github.com/pytorch/audio/pull/2319
GitHub Author: xiaohui-zhang <xiaohuizhang@fb.com>

### DIFF
--- a/torchaudio/models/rnnt.py
+++ b/torchaudio/models/rnnt.py
@@ -377,20 +377,20 @@ class _Joiner(torch.nn.Module):
     Args:
         input_dim (int): source and target input dimension.
         output_dim (int): output dimension.
-        joiner_activation (str, optional): activation function to use in the joiner
+        activation (str, optional): activation function to use in the joiner
             Must be one of ("relu", "tanh"). (Default: "relu")
 
     """
 
-    def __init__(self, input_dim: int, output_dim: int, joiner_activation: str = "relu") -> None:
+    def __init__(self, input_dim: int, output_dim: int, activation: str = "relu") -> None:
         super().__init__()
         self.linear = torch.nn.Linear(input_dim, output_dim, bias=True)
-        if joiner_activation == "relu":
+        if activation == "relu":
             self.activation = torch.nn.ReLU()
-        elif joiner_activation == "tanh":
+        elif activation == "tanh":
             self.activation = torch.nn.Tanh()
         else:
-            raise ValueError(f"Unsupported activation {joiner_activation}")
+            raise ValueError(f"Unsupported activation {activation}")
 
     def forward(
         self,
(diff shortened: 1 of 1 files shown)


torchaudio/models/rnnt.py
```diff
@@ -377,12 +377,17 @@ class _Joiner(torch.nn.Module):
     Args:
         input_dim (int): source and target input dimension.
         output_dim (int): output dimension.
+        use_tanh (bool, optional): if ``True``, uses Tanh rather ReLU activation. (Default: ``False``)
+
     """
 
-    def __init__(self, input_dim: int, output_dim: int) -> None:
+    def __init__(self, input_dim: int, output_dim: int, use_tanh: bool = False) -> None:
```
hwangjeff, line 384: instead of bool option `use_tanh`, how about we do a str option `activation` that can take on a value of either `"relu"` or `"tanh"`? this way, if we find later that other activation functions are viable, we can continue adding them without needing to update the option. 

example: https://github.com/pytorch/audio/blob/66185e0074d06330dff4ea84cbface3b83312ec4/torchaudio/models/emformer.py#L343

torchaudio/models/rnnt.py
```diff
@@ -377,12 +377,17 @@ class _Joiner(torch.nn.Module):
...
         super().__init__()
         self.linear = torch.nn.Linear(input_dim, output_dim, bias=True)
-        self.relu = torch.nn.ReLU()
+        if use_tanh:
+            self.activation = torch.nn.Tanh()
+        else:
+            self.activation = torch.nn.ReLU()
```
hwangjeff, line 390: then, if we use `activation`, we can update this to raise an exception in the default case to prevent users from choosing an activation function they didn't intend to choose.

example:
https://github.com/pytorch/audio/blob/66185e0074d06330dff4ea84cbface3b83312ec4/torchaudio/models/emformer.py#L39-L47

torchaudio/models/rnnt.py
```diff
@@ -377,12 +377,20 @@ class _Joiner(torch.nn.Module):
     Args:
         input_dim (int): source and target input dimension.
         output_dim (int): output dimension.
+        joiner_activation (str, optional): activation function to use in the joiner
+            Must be one of ("relu", "tanh"). (Default: "relu")
+
     """
 
-    def __init__(self, input_dim: int, output_dim: int) -> None:
+    def __init__(self, input_dim: int, output_dim: int, joiner_activation: str = "relu") -> None:
```
hwangjeff, line 385: nit: since the class is called `_Joiner`, we can just call the argument `activation` for simplicity

hwangjeff, line 385: nit: since the class is called `_Joiner`, we can just call the argument `activation` for simplicity

hwangjeff (approved):

PR #2320, opened 2022-04-04
Add an option to use GroupNorm instead of BatchNorm in Conformer model

xiaohui-zhang:
gpurun pytest test/torchaudio_unittest/models/conformer
git add ============================= test session starts ==============================
platform linux -- Python 3.7.0, pytest-7.1.1, pluggy-1.0.0
rootdir: /data/home/xiaohuizhang/audio_dev2
collected 4 items

test/torchaudio_unittest/models/conformer/conformer_cpu_test.py ..       [ 50%]
test/torchaudio_unittest/models/conformer/conformer_gpu_test.py ..       [100%]

============================== 4 passed in 10.36s ==============================

### DIFF
--- a/torchaudio/models/conformer.py
+++ b/torchaudio/models/conformer.py
@@ -31,6 +31,7 @@ class _ConvolutionModule(torch.nn.Module):
         num_channels: int,
         depthwise_kernel_size: int,
         bias: bool = False,
+        use_group_norm: bool = False,
         dropout: float = 0.0,
     ) -> None:
         super().__init__()
@@ -55,7 +56,7 @@ class _ConvolutionModule(torch.nn.Module):
                 groups=num_channels,
                 bias=bias,
             ),
-            torch.nn.BatchNorm1d(num_channels),
+            torch.nn.GroupNorm(num_groups=1, num_channels=num_channels) if use_group_norm else torch.nn.BatchNorm1d(num_channels),
             torch.nn.SiLU(),
             torch.nn.Conv1d(
                 num_channels,
@@ -131,6 +132,7 @@ class ConformerLayer(torch.nn.Module):
         num_attention_heads: int,
         depthwise_conv_kernel_size: int,
         dropout: float = 0.0,
+        use_group_norm: bool = False,
     ) -> None:
         super().__init__()
 
@@ -145,6 +147,7 @@ class ConformerLayer(torch.nn.Module):
             num_channels=input_dim,
             depthwise_kernel_size=depthwise_conv_kernel_size,
             bias=True,
+            use_group_norm=use_group_norm,
             dropout=dropout,
         )
 


torchaudio/models/conformer.py
```diff
@@ -131,6 +134,7 @@ def __init__(
         num_attention_heads: int,
         depthwise_conv_kernel_size: int,
         dropout: float = 0.0,
+        use_group_norm: bool = False,
```
hwangjeff, line 137: can you add a corresponding entry to the docstring?

torchaudio/models/conformer.py
```diff
@@ -31,6 +31,7 @@ def __init__(
         num_channels: int,
         depthwise_kernel_size: int,
         bias: bool = False,
+        use_group_norm: bool = False,
```
hwangjeff, line 34: and similarly here?

also noticed that `dropout` is missing a docstring...we can address separately

torchaudio/models/conformer.py
```diff
@@ -22,16 +22,19 @@ class _ConvolutionModule(torch.nn.Module):
...
     """
 
     def __init__(
         self,
         input_dim: int,
         num_channels: int,
         depthwise_kernel_size: int,
-        bias: bool = False,
         dropout: float = 0.0,
+        bias: bool = False,
```
mthrok, line 36: Moving `bias` parameter is BC-breaking. Is this intended?

xiaohui-zhang: sorry what does it mean by BC-breaking? I just wanted to gather bool-type inputs together, after int/float inputs

mthrok: Changing the order of function arguments breaks backward compatibility.

Say, a library, FooLibrary of v1 has a function like the following

```python
foo(a: int, b: str)
```

and in my project code I use it as

```python
result = foo(3, "foo")
```

Now, if the library code is changed like the following in v2 or something,

```python
foo(b: str, a: int)
```

Then simply upgrading my environment from FooLibrary v1 to v2, my project code stops working, unless I change my own code. The kind of changes that force client code change, is backward-incompatible, and it has to be communicated with users (deprecation). In PyTorch, we take at least two-release cycles for deprecating a public API.

mthrok: however the said class is not a public API, rather an internal one, so I guess it's kinda okay. Although I would say it's unnecessary if it's for the sake of cosmetics. @hwangjeff thoughts?

torchaudio/models/conformer.py
```diff
@@ -122,6 +127,8 @@ class ConformerLayer(torch.nn.Module):
         num_attention_heads (int): number of attention heads.
         depthwise_conv_kernel_size (int): kernel size of depthwise convolution layer.
         dropout (float, optional): dropout probability. (Default: 0.0)
+        use_group_norm (bool, optional): use GroupNorm rather than BatchNorm in ConvolutionModule. (Default: False)
+        convolution_first (bool, optional): apply the convolution module ahead of the attention module. (Default: False)
```
hwangjeff, line 131:
```suggestion
        convolution_first (bool, optional): apply the convolution module ahead of the attention module. (Default: ``False``)
```

torchaudio/models/conformer.py
xiaohui-zhang: it's an internal function so it should be ok to change, though I agree it's really a cosmetics change...

hwangjeff (approved):
lg

PR #2330, opened 2022-04-12
Support NV12 format in video decoding

mthrok:
Support NV12 format in Streamer API.

NV12 is a biplanar format with a full sized Y plane followed by a single chroma plane with weaved U and V values.
https://chromium.googlesource.com/libyuv/libyuv/+/HEAD/docs/formats.md#nv12-and-nv21

The original UV plane is smaller than Y plane, so in this implmentation,
UV plane is upsampled to match the size of Y plane.

### DIFF
--- a/torchaudio/csrc/ffmpeg/buffer.cpp
+++ b/torchaudio/csrc/ffmpeg/buffer.cpp
@@ -2,6 +2,10 @@
 #include <stdexcept>
 #include <vector>
 
+#ifdef USE_CUDA
+#include <c10/cuda/CUDAStream.h>
+#endif
+
 namespace torchaudio {
 namespace ffmpeg {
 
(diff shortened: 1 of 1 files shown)


torchaudio/csrc/ffmpeg/buffer.cpp
```diff
@@ -164,6 +168,94 @@ void AudioBuffer::push_frame(AVFrame* frame) {
...
+                     .layout(torch::kStrided)
+                     .device(torch::kCPU);
+
+  torch::Tensor y = torch::empty({1, height, width, 1}, options);
+  {
+    uint8_t* tgt = y.data_ptr<uint8_t>();
+    uint8_t* src = pFrame->data[0];
+    int linesize = pFrame->linesize[0];
+    for (int h = 0; h < height; ++h) {
+      memcpy(tgt, src, width);
```
hwangjeff, line 187: based on my limited understanding from scanning docs: `width` represents the pixel count per row, and `memcpy` would interpret it as the number of bytes to copy. so, the assumption seems to be that each pixel is 8 bits, but, for nv12, each pixel is 12 bits. how does this work?

mthrok: 12 bits par pixel is a statistic of the amount of information used for pixel, but the actual memory alignment is a different story.

In the case of NV12, the data is stored in two different planes, Y and UV.
The Y plane is full-sized, meaning it has the same number of pixel as the actual image.

```
FFmpeg work memory

<--------linesize-------->
<----width---->
YYYYYYYYYYYYYYY----------- ^  
YYYYYYYYYYYYYYY----------- |  
  ...                      | height 
YYYYYYYYYYYYYYY----------- ^

Y : valid pixel
- : working memory region (no valid data)
```
So in this particular for-loop, memcpy copy only valid pixels 
while moving the pointer to the next line, in different rate among
source and destination. (the destination Tensor is size of `width x height`)

Now, UV plane is different. First, each U pixels and V pixels have one quarter of pixels. 
Both of them `width / 2 x height / 2` number of pixels.
This is why on average, One pixel has 12 bits. (8 bits for Y and 8 / 4 == 2 bits for U and same for V).

In the case of NV12, UV plane is interweaved so the memory alignment looks like 
it's `width / 2 x height` size image, with a specific linesize.

```
<---------linesize--------->
<----width---->
UVUVUVUVUVUVUV------------- ^  
UVUVUVUVUVUVUV------------- |  
  ...                       | height / 2
UVUVUVUVUVUVUV------------- ^

U : valid pixel of U
V : valid pixel of V
- : working memory region (no valid data)
```

so when copying data first, we use Tensor of `height / 2, width / 2, 2`, then reshape and enlarge it to the original size.

mthrok: Note that YUV420P occupies the same amount of memory, but the alignment is different. https://github.com/pytorch/audio/pull/2334

hwangjeff (approved):

PR #2333, opened 2022-04-13
Add first version of iOS build script

hwangjeff:
Adds first version of iOS build script that includes only csrc files with no third party dependencies.

Sample run command:
```
LIBTORCH_HEADER_ROOT="/path/LibTorch-Lite/install/include" IOS_ARCH="x86_64" ./build_ios.sh
```


atalman:
HI @hwangjeff  . Looks good. May I ask, please put in the description the list of artifacts that will be generated by this script and a small note how do we plan on distributing these binaries ?

PR #2335, opened 2022-04-13
Add feature_grad_mult argument to HuBERTPretrainModel

nateanl:
In Wav2Vec2 and HuBERT model training, the convolutional feature extraction layers use `group_norm` for normalization in `Base` model, while they use `layer_norm` in `Large` and `XLarge` models. For `Base` model, the gradients of feature extraction layers will be unstable in pre-training, thus we need to scale down the gradient by multiplying 0.1.

In this PR, we add such argument to `HuBERTPretrainModel` to control the gradient of feature extractor layers. We also put the argument in the factory functions (`hubert_pretrain_base`, `hubert_pretrain_large`, and `hubert_pretrain_xlarge`. The reason is in finetuning, the feature extractor's parameters are fixed, we can multiply the gradient with 0.0 to avoid back propagating gradients.

### DIFF
--- a/torchaudio/models/wav2vec2/model.py
+++ b/torchaudio/models/wav2vec2/model.py
@@ -138,7 +138,7 @@ class HuBERTPretrainModel(Module):
         logit_generator (torch.nn.Module):
             Logit generator that predicts the logits of the masked and unmasked inputs.
 
-        feature_grad_mult (float):
+        feature_grad_mult (float or None):
             The factor to scale the convolutional feature extraction layer gradients by.
             The scale factor will not affect the forward pass.
     """
@@ -155,8 +155,8 @@ class HuBERTPretrainModel(Module):
         self.mask_generator = mask_generator
         self.logit_generator = logit_generator
         assert (
-            0.0 <= feature_grad_mult <= 1.0
-        ), f"The value of `feature_grad_mult` must be between [0, 1]. Found {feature_grad_mult}"
+            feature_grad_mult is None or 0.0 <= feature_grad_mult < 1.0
+        ), f"The value of `feature_grad_mult` must be between [0, 1). Found {feature_grad_mult}"
         self.feature_grad_mult = feature_grad_mult
 
     def forward(
@@ -723,7 +723,7 @@ def hubert_pretrain_model(
     skip_nomask: bool,
     num_classes: int,
     final_dim: int,
-    feature_grad_mult: float,
+    feature_grad_mult: Optional[float],
 ) -> HuBERTPretrainModel:
     # Overriding the signature so that the return type is correct on Sphinx
     """hubert_pretrain_model(extractor_mode: str, extractor_conv_layer_config: Optional[List[Tuple[int, int, int]]], extractor_conv_bias: bool, encoder_embed_dim: int, encoder_projection_dropout: float, encoder_pos_conv_kernel: int, encoder_pos_conv_groups: int, encoder_num_layers: int, encoder_num_heads: int, encoder_attention_dropout: float, encoder_ff_interm_features: int, encoder_ff_interm_dropout: float, encoder_dropout: float, encoder_layer_norm_first: bool, encoder_layer_drop: float, mask_prob: float, mask_selection: str, mask_other: float, mask_length: int, no_mask_overlap: bool, mask_min_space: int, mask_channel_prob: float, mask_channel_selection: str, mask_channel_other: float, mask_channel_length: int, no_mask_channel_overlap: bool, mask_channel_min_space: int, skip_masked: bool, skip_nomask: bool, num_classes: int, final_dim: int) -> torchaudio.models.HuBERTPretrainModel
@@ -922,7 +922,7 @@ def hubert_pretrain_model(
 
             This option corresponds to ``final_dim`` from ``fairseq``.
 
-        feature_grad_mult (float):
+        feature_grad_mult (float or None):
             The factor to scale the convolutional feature extraction layer gradients by.
             The scale factor will not affect the forward pass.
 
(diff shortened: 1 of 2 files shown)


torchaudio/models/wav2vec2/model.py
```diff
@@ -910,6 +915,11 @@ def hubert_pretrain_model(
 
             This option corresponds to ``final_dim`` from ``fairseq``.
 
+        feature_grad_mult (float):
+            Whether to scale down the gradient of convolutional feature extraction layers.
```
hwangjeff, line 919: the language here implies that `feature_grad_mult` is of type `bool` when it's actually of type `float`. something like "factor to scale convolutional feature extraction layer gradients by" would be clearer

torchaudio/models/wav2vec2/model.py
```diff
@@ -144,11 +144,13 @@ def __init__(
         wav2vec2: Wav2Vec2Model,
         mask_generator: Module,
         logit_generator: Module,
+        feature_grad_mult: float,
```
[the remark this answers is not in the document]
hwangjeff: should add to the docstring above

torchaudio/models/wav2vec2/model.py
mthrok: Also probably worth mentioning that forward path is not affected.

torchaudio/models/wav2vec2/model.py
```diff
@@ -712,6 +723,7 @@ def hubert_pretrain_model(
     skip_nomask: bool,
     num_classes: int,
     final_dim: int,
+    feature_grad_mult: float,
```
mthrok, line 726: Should this be `float` or `Optional[float]`?

torchaudio/models/wav2vec2/model.py
```diff
@@ -137,18 +137,27 @@ class HuBERTPretrainModel(Module):
...
         logit_generator: Module,
+        feature_grad_mult: Optional[float],
     ):
         super().__init__()
         self.wav2vec2 = wav2vec2
         self.mask_generator = mask_generator
         self.logit_generator = logit_generator
+        assert (
+            0.0 <= feature_grad_mult <= 1.0
+        ), f"The value of `feature_grad_mult` must be between [0, 1]. Found {feature_grad_mult}"
```
mthrok, line 159: I think it should be exclusive in both end.
It does not make much sense if it is 0 or 1.

torchaudio/models/wav2vec2/model.py
```diff
@@ -1033,6 +1060,7 @@ def hubert_pretrain_large(
     encoder_ff_interm_dropout: float = 0.0,
     encoder_dropout: float = 0.0,
     encoder_layer_drop: float = 0.0,
+    feature_grad_mult=1.0,
```
mthrok, line 1063: If it's 1.0, gradient is not modified so why not default to `None`?

torchaudio/models/wav2vec2/model.py
nateanl: It makes sense when it is 0. In fine-tuning case, the parameters of feature extractor are fixed, which can be done by setting it to 0

torchaudio/models/wav2vec2/model.py
nateanl: Sure, will change it.

torchaudio/models/wav2vec2/model.py
mthrok: I think in such cases, freezing the entire model using PyTorch's API is more appropriate. That way, the backward path is stopped. Relying on `feature_grad_mult =0` means that backward path is still computed.

nateanl: There are two stages in the fine-tuning, in the first stage, only `aux` layer has gradient, the rest of parameters are fixed. In the second stage, the feature extractor's parameters are fixed. The train loop looks like:
```python
if self.global_step <= self.fix_encoder_updates:
    with torch.no_grad():
        x, out_len = self.model.wav2vec2.feature_extractor(waveforms, audio_lengths)
        padding_mask = components._get_padding_mask(x, out_len)
        x, attention_mask = self.model.wav2vec2.encoder._preprocess(x, out_len)
        x, mask = self.model.mask_generator(x, padding_mask)
        x = self.model.wav2vec2.encoder.transformer(x, attention_mask=attention_mask)
else:
    with torch.no_grad():
        x, out_len = self.model.wav2vec2.feature_extractor(waveforms, audio_lengths)
    padding_mask = components._get_padding_mask(x, out_len)
    x, attention_mask = self.model.wav2vec2.encoder._preprocess(x, out_len)
    x, mask = self.model.mask_generator(x, padding_mask)
    x = self.model.wav2vec2.encoder.transformer(x, attention_mask=attention_mask)
logits = self.aux(x)
```
Is there a better way to simplify the logic?

torchaudio/models/wav2vec2/model.py
```diff
@@ -137,18 +137,27 @@ class HuBERTPretrainModel(Module):
 
         logit_generator (torch.nn.Module):
             Logit generator that predicts the logits of the masked and unmasked inputs.
+
+        feature_grad_mult (float or None):
```
hwangjeff, line 141: can we explain the behavior when `feature_grad_mult` is `None`?

torchaudio/models/wav2vec2/model.py
mthrok: > Is there a better way to simplify the logic?

Since the data flow is same, I think one can set `requires_grad` outside of the loop when it hits the right condition.

The pseudo code would look like;
1. Initialize the `self.model.wav2vec2`
2. Disable gradient entirely
3. Run the training loop for `self.fix_encoder_updates`
4. Once the `self.fix_encoder_updates` step is done, enable gradient for `self.model.wav2vec2.feature_extractor`.

nateanl: I see. I need to search where is the best place to enable gradient. `train_step` is called every iteration, so it's not suitable to call gradient enabling here.

to calrify, during the whole fine-tuning, `self.model.wav2vec2.feature_extractor` is freezed, otherwise the training will be unstable. In the first `fix_encoder_updates` iterations, the wav2vec2 model is freezed, only `aux` layer has gradient. after that, the gradient of `wav2vec2.encoder` is enabled.

nateanl:
@mthrok do you want to merge https://github.com/pytorch/audio/pull/2153 first? then I adjust the annotations of hubert functions in this PR

mthrok (approved):

PR #2339, opened 2022-04-18
Change underlying implementation of RNN-T hypothesis to tuple

hwangjeff:
PyTorch Lite, which is becoming a standard for mobile PyTorch usage, does not support containers containing custom classes. Consequently, because TorchAudio's RNN-T decoder currently returns and accepts lists of `Hypothesis` namedtuples, it is not compatible with PyTorch Lite. This PR resolves said incompatibility by changing the underlying implementation of `Hypothesis` to tuple.


examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -67,15 +67,15 @@ def run_eval_streaming(args):
                 features, length = streaming_feature_extractor(segment)
                 hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
             hypothesis = hypos[0]
-            transcript = token_processor(hypothesis.tokens, lstrip=False)
```
nateanl, line 70: Does libtorch-light support `namedtuple`? `hypothesis.tokens` is more intuitive to users than `hypothesis[0]`

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -67,15 +67,15 @@ def run_eval_streaming(args):
...
+            transcript = token_processor(hypothesis[0], lstrip=False)
```
nateanl, line 70: Does libtorch-light support dictionary? `hypothesis["tokens"]` is more intuitive to users than `hypothesis[0]`

mthrok: Dictionary should be used when keys can vary at runtime.
The structure of `Hypothesis` does not change at runtime, so dictionary is not appropriate.

Defining `token_index = 0` would be the best it can do.

nateanl (approved):

PR #2342, opened 2022-04-20
Add lexicon free CTC decoder

carolineechen:
Add support for lexicon free decoding based on [fairseq's](https://github.com/pytorch/fairseq/blob/main/examples/speech_recognition/new/decoders/flashlight_decoder.py#L53) implementation. Reached numerical parity with fairseq's decoder in offline experimentation

Follow ups
- Add pretrained LM support for lex free decoding
- Add example in tutorial
- Replace flashlight C++ source code with flashlight text submodule


mthrok (commented):
Could you add alias for the previous API with deprecation warning, so that it is easy for those who use nightly to migrate to the new name? No need to keep the previous API in docs.

mthrok (approved):
LGTM

Follow-up: [optional] Add unit test to verify the numerical parity with fairseq.

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.cpp
```diff
@@ -0,0 +1,207 @@
...
+  hyp_[0].emplace_back(0.0, lm_->start(0), nullptr, sil_);
+  nDecodedFrames_ = 0;
+  nPrunedFrames_ = 0;
+}
+
+void LexiconFreeDecoder::decodeStep(const float* emissions, int T, int N) {
+  int startFrame = nDecodedFrames_ - nPrunedFrames_;
+  // Extend hyp_ buffer
+  if (hyp_.size() < startFrame + T + 2) {
+    for (int i = hyp_.size(); i < startFrame + T + 2; i++) {
```
nateanl, line 34: not required, but using the same increment operator in for loop looks better.  we can stick to one of `i++` and `++i` in this PR.

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.h
```diff
@@ -0,0 +1,160 @@
...
+
+#include "torchaudio/csrc/decoder/src/decoder/Decoder.h"
+#include "torchaudio/csrc/decoder/src/decoder/lm/LM.h"
+
+namespace torchaudio {
+namespace lib {
+namespace text {
+
+struct LexiconFreeDecoderOptions {
+  int beamSize; // Maximum number of hypothesis we hold after each step
```
nateanl, line 20:
```suggestion
  int beamSize; // Maximum number of hypotheses we hold after each step
```

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.h
```diff
@@ -0,0 +1,160 @@
...
+  int beamSizeToken; // Maximum number of tokens we consider at each step
+  double beamThreshold; // Threshold to prune hypothesis
```
nateanl, line 22:
```suggestion
  double beamThreshold; // Threshold to prune hypotheses
```

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.h
```diff
@@ -0,0 +1,160 @@
...
+  DecodeResult getBestHypothesis(int lookBack = 0) const override;
+
+  std::vector<DecodeResult> getAllFinalHypothesis() const override;
+
+ protected:
+  LexiconFreeDecoderOptions opt_;
+  LMPtr lm_;
+  std::vector<float> transitions_;
+
+  // All the hypothesis new candidates (can be larger than beamsize) proposed
```
nateanl, line 134: same above

nateanl (approved):

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.cpp
carolineechen: would prefer to leave the source implementation unchanged, as this is ported directly from flashlight repo and there may be performance implications for changing code. I can make the suggestions to the code comments though.

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.h
carolineechen, line 134:
```suggestion
  // All the hypotheses new candidates (can be larger than beamsize) proposed
```

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.h
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  bool isComplete() const {
+    return true;
+  }
+};
+
+/**
+ * Decoder implements a beam seach decoder that finds the word transcription
+ * W maximizing:
```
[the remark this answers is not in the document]
carolineechen: I think this is correct as is

torchaudio/csrc/decoder/src/decoder/LexiconFreeDecoder.cpp
nateanl: I see, if the source code is directly adopted, it makes sense to keep as what it is.

PR #2344, opened 2022-04-22
Fix bug with unsqueezing length tensor in RNNTBeamSearch

glynpu:
This PR amends `RNNTBeamSearch`'s streaming decoding method to correctly unsqueeze `length` when its dimension is 0.

Original comment: Is "input.dim() == 0" unreachable as it could only be 2 or 3 in assertion of Line 329?

### DIFF
--- a/torchaudio/models/rnnt_decoder.py
+++ b/torchaudio/models/rnnt_decoder.py
@@ -333,6 +333,8 @@ class RNNTBeamSearch(torch.nn.Module):
             input = input.unsqueeze(0)
 
         assert length.shape == () or length.shape == (1,), "length must be of shape () or (1,)"
+        if length.dim() == 0:
+            length = length.unsqueeze(0)
 
         enc_out, _, state = self.model.transcribe_streaming(input, length, state)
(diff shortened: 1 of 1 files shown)


glynpu:
According to ["transcribe_streaming" interface](https://github.com/pytorch/audio/blob/48facbd4499adefe622e4d618f9c5ef3f6110d0c/torchaudio/models/rnnt.py#L556), length should be with shape `(B,)`. 
So is the original `if input.dim() == 0` statement a typo of `if length.dim() == 0`?

hwangjeff:
> According to ["transcribe_streaming" interface](https://github.com/pytorch/audio/blob/48facbd4499adefe622e4d618f9c5ef3f6110d0c/torchaudio/models/rnnt.py#L556), length should be with shape `(B,)`. So is the original `if input.dim() == 0` statement a typo of `if length.dim() == 0`?

Hi @glynpu, yes, you're right — this is a typo. Thank you for the fix!

hwangjeff (approved):

PR #2349, opened 2022-04-25
Add VoxCeleb1 dataset

nateanl:
This PR adds two dataset classes of VoxCeleb1 corpus.
- `VoxCeleb1Identification`
Each data sample contains the waveform, sample rate, speaker id, and the file id.
- `VoxCeleb1Verification`
Each data sample contains a pair of waveforms, sample rate, the label indicating if they are from the same speaker, and the file ids.

### DIFF
--- a/torchaudio/datasets/voxceleb1.py
+++ b/torchaudio/datasets/voxceleb1.py
@@ -24,7 +24,7 @@ _RELEASE_CONFIGS = {
 
 
 class VoxCeleb1(Dataset):
-    """Create VoxCeleb1 Dataset.
+    """Create *VoxCeleb1* [:footcite:`nagrani2017voxceleb`] Dataset.
 
     Args:
         root (str or Path): Path to the directory where the dataset is found or downloaded.
--- a/torchaudio/datasets/voxceleb1.py
+++ b/torchaudio/datasets/voxceleb1.py
@@ -125,7 +125,7 @@ class VoxCeleb1Identification(VoxCeleb1):
     Args:
         root (str or Path): Path to the directory where the dataset is found or downloaded.
         subset (str, optional): Subset of the dataset to use. Options: ["train", "dev", "test"]. (Default: ``"train"``)
-        mate_url (str, optional): The url of meta file that contains the list of subset labels and file paths.
+        meta_url (str, optional): The url of meta file that contains the list of subset labels and file paths.
         download (bool, optional):
             Whether to download the dataset if it is not found at root path. (Default: ``False``).
     """
@@ -170,7 +170,7 @@ class VoxCeleb1Verification(VoxCeleb1):
     Args:
         root (str or Path): Path to the directory where the dataset is found or downloaded.
         subset (str, optional): Subset of the dataset to use. Options: ["train", "dev", "test"]. (Default: ``"train"``)
-        mate_url (str, optional): The url of meta file that contains
+        meta_url (str, optional): The url of meta file that contains
             a list of utterance pairs and the corresponding labels.
         download (bool, optional):
             Whether to download the dataset if it is not found at root path. (Default: ``False``).
@@ -199,7 +199,8 @@ class VoxCeleb1Verification(VoxCeleb1):
         file_id_spk1 = _get_file_id(file_path_spk1, self._ext_audio)
         file_id_spk2 = _get_file_id(file_path_spk2, self._ext_audio)
         waveform_spk1, sample_rate = torchaudio.load(os.path.join(self._path, file_path_spk1))
-        waveform_spk2, sample_rate = torchaudio.load(os.path.join(self._path, file_path_spk2))
+        waveform_spk2, sample_rate2 = torchaudio.load(os.path.join(self._path, file_path_spk2))
+        assert sample_rate == sample_rate2
         return (waveform_spk1, waveform_spk2, label, sample_rate, file_id_spk1, file_id_spk2)
 
     def __len__(self) -> int:
--- a/torchaudio/datasets/voxceleb1.py
+++ b/torchaudio/datasets/voxceleb1.py
@@ -1,0 +1,82 @@
+import os
+from pathlib import Path
+from typing import Tuple, Union, Optional
+
+import torchaudio
+from torch import Tensor
+from torch.hub import download_url_to_file
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import (
+    extract_archive,
+)
+
+
+_RELEASE_CONFIGS = {
+    "dev": {
+        "url": "https://thor.robots.ox.ac.uk/~vgg/data/voxceleb/vox1a/vox1_dev_wav.zip",
+        "checksum": "9c3b51e34038d1bdb2174dcc66543267",
+    },
+    "test": {
+        "url": "https://thor.robots.ox.ac.uk/~vgg/data/voxceleb/vox1a/vox1_test_wav.zip",
+        "checksum": "8e06592a5f604e23e8cd10f421b36cc3",
+    },
+}
+
+
+class Voxceleb1(Dataset):
+    """Create Voxceleb1 Dataset.
+
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str or None, optional): Subset of the dataset to use. Options: ["dev", "test"]. (Default: ``None``)
+        download (bool, optional):
+            Whether to download the dataset if it is not found at root path. (Default: ``False``).
+    """
+
+    _ext_audio = ".wav"
+
+    def __init__(self, root: Union[str, Path], subset: Optional[str] = None, download: bool = False) -> None:
+        assert subset is None or subset in ["dev", "test"], "`subset` must be one of ['dev', 'test']"
+
+        # Get string representation of 'root' in case Path object is passed
+        root = os.fspath(root)
+
+        basename = os.path.basename(_RELEASE_CONFIGS[subset]["url"])
+        archive = os.path.join(root, basename)
+        self._path = os.path.join(root, basename.replace(".zip", ""))
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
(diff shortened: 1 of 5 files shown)


torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,82 @@
...
+    },
+}
+
+
+class VoxCeleb1(Dataset):
+    """Create VoxCeleb1 Dataset.
+
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str or None, optional): Subset of the dataset to use. Options: ["dev", "test"]. (Default: ``None``)
```
carolineechen, line 31: Can you mention the default behavior if `None`? Also with the current implementation, `None` will throw an error because of `_RELEASE_CONFIGS[subset]["url"]` accessing an invalid key

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,82 @@
...
+        download (bool, optional):
+            Whether to download the dataset if it is not found at root path. (Default: ``False``).
+    """
+
+    _ext_audio = ".wav"
+
+    def __init__(self, root: Union[str, Path], subset: Optional[str] = None, download: bool = False) -> None:
+        assert subset is None or subset in ["dev", "test"], "`subset` must be one of ['dev', 'test']"
```
carolineechen, line 39: If `None` is supported:

```suggestion
        assert subset is None or subset in ["dev", "test"], "`subset` must be one of ['dev', 'test', None]"
```

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,82 @@
...
+
+        basename = os.path.basename(_RELEASE_CONFIGS[subset]["url"])
+        archive = os.path.join(root, basename)
+        self._path = os.path.join(root, basename.replace(".zip", ""))
+        if download:
+            if not os.path.isdir(self._path):
+                if not os.path.isfile(archive):
+                    checksum = _RELEASE_CONFIGS[subset]["checksum"]
+                    url = _RELEASE_CONFIGS[subset]["url"]
+                    # dev data is splited to
```
carolineechen, line 52: can you reword this comment?

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self, root: Union[str, Path], subset: str = "dev", download: bool = False) -> None:
+        assert subset in ["dev", "test"], "`subset` must be one of ['dev', 'test']"
+
+        # Get string representation of 'root' in case Path object is passed
+        root = os.fspath(root)
+
+        basename = os.path.basename(_RELEASE_CONFIGS[subset]["url"])
+        archive = os.path.join(root, basename)
+        self._path = os.path.join(root, basename.replace(".zip", ""))
+        if download:
```
carolineechen, line 47: I noticed that roughly half our dataset implementations will check and throw a Runtime error if the dataset does not exist and download is False, while the other half will not do anything if download is set to False, which is the default value. I think it makes sense to throw an error (or at least warning) in those cases otherwise the dataset intialization does not make sense, thoughts?

cc @mthrok

mthrok: Yeah, throwing runtime error makes sense and seems more intuitive UX.

nateanl: Cool, thanks. We can add such UX to other datasets in a separate pr.

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,83 @@
...
+
+        Args:
+            n (int): The index of the sample to be loaded
+
+        Returns:
+            (Tensor, int, int, str):
+            ``(waveform, sample_rate, speaker_id, youtube_id)``
+        """
+        filename = self._walker[n]
+        speaker, youtube_id, _ = filename.split("/")[-3:]
```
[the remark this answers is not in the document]
nateanl: I just realized that s3prl uses a pre-defined txt file to split train, dev, test data for speaker identification: https://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/iden_split.txt
id10270 - id 10309 files are in test zip file, but some of them are regarded as `train` split (i.e., has `1` at the beginning of txt file).
I need to re-think about the subset design for this dataset.

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,206 @@
...
+
+class VoxCeleb1Verification(VoxCeleb1):
+    """Create *VoxCeleb1* [:footcite:`nagrani2017voxceleb`] Dataset for speaker verification task.
+    Each data sample contains a pair of waveforms, the label indicating if they are from the same speaker,
+    sample rate, and the file names.
+
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str, optional): Subset of the dataset to use. Options: ["train", "dev", "test"]. (Default: ``"train"``)
+        mate_url (str, optional): The url of meta file that contains
```
carolineechen, line 173:
```suggestion
        meta_url (str, optional): The url of meta file that contains
```

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,206 @@
...
+
+        Returns:
+            (Tensor, Tensor, int, int, str, str):
+            ``(waveform_spk1, waveform_spk2, label, sample_rate, file_id_spk1, file_id_spk2)``
+        """
+        label, file_path_spk1, file_path_spk2 = self._flist[n]
+        label = int(label)
+        file_id_spk1 = _get_file_id(file_path_spk1, self._ext_audio)
+        file_id_spk2 = _get_file_id(file_path_spk2, self._ext_audio)
+        waveform_spk1, sample_rate = torchaudio.load(os.path.join(self._path, file_path_spk1))
```
carolineechen, line 201: is it ok to assume these two files have the same `sample_rate`?

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,206 @@
...
+
+
+class VoxCeleb1Verification(VoxCeleb1):
+    """Create *VoxCeleb1* [:footcite:`nagrani2017voxceleb`] Dataset for speaker verification task.
+    Each data sample contains a pair of waveforms, the label indicating if they are from the same speaker,
+    sample rate, and the file names.
+
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str, optional): Subset of the dataset to use. Options: ["train", "dev", "test"]. (Default: ``"train"``)
```
carolineechen, line 172: `subset` is unused in the current implementation here. In [s3prl],(https://github.com/s3prl/s3prl/blob/master/s3prl/downstream/sv_voxceleb1/dataset.py#L30) there is support for train and test splits, and they have different implementations for getitem

nateanl: Thanks. This line should be removed.
The `SpeakerVerifi_train` dataset in s3prl can be replaced by our `VoxCeleb1Identification` dataset. Given one utterance as the input, the label is the class of the speaker. `SpeakerVerifi_test` is for a different task, given two utterances, the label indicates whether they are from the same speaker or not, which can be replaced by our `VoxCeleb1Verification` dataset.

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,207 @@
...
+    def __getitem__(self, n: int):
+        raise NotImplementedError
+
+    def __len__(self) -> int:
+        raise NotImplementedError
+
+
+class VoxCeleb1Identification(VoxCeleb1):
+    """Create *VoxCeleb1* [:footcite:`nagrani2017voxceleb`] Dataset for speaker identification task.
+    Each data sample contains the waveform, correcponding speaker class label, sample rate, and the file name.
```
carolineechen, line 123:
```suggestion
    Each data sample contains the waveform, sample rate, speaker class label, and the file name.
```

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,207 @@
...
+        Returns:
+            (Tensor, int, int, str):
+            ``(waveform, speaker_id, sample_rate, file_id)``
+        """
+        file_path = self._flist[n]
+        file_id = _get_file_id(file_path, self._ext_audio)
+        speaker_id = file_id.split("-")[0]
+        speaker_id = int(speaker_id[3:])
+        waveform, sample_rate = torchaudio.load(os.path.join(self._path, file_path))
+        return (waveform, speaker_id, sample_rate, file_id)
```
carolineechen, line 159: can we switch order of sample_rate and speaker_id? since most torchaudio datasets first return wav, sr then other data. 

and maybe move up sample_rate in veri data as well, but that's a bit less clear.

torchaudio/datasets/voxceleb1.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class VoxCeleb1Identification(VoxCeleb1):
+    """Create *VoxCeleb1* [:footcite:`nagrani2017voxceleb`] Dataset for speaker identification task.
+    Each data sample contains the waveform, correcponding speaker class label, sample rate, and the file name.
+
+    Args:
+        root (str or Path): Path to the directory where the dataset is found or downloaded.
+        subset (str, optional): Subset of the dataset to use. Options: ["train", "dev", "test"]. (Default: ``"train"``)
+        meta_url (str, optional): The url of meta file that contains the list of subset labels and file paths.
```
carolineechen, line 128: add default value/description here, and same with veri dataset

torchaudio/datasets/voxceleb1.py
carolineechen: this line still needs to be removed

carolineechen (commented):
can you address this? otherwise looks good!

carolineechen (approved):

PR #2353, opened 2022-04-27
Fix audio win smoke test to use GPU hosts for CUDA builds

atalman:
Fix audio win smoke test to use GPU hosts for CUDA builds

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -403,7 +403,7 @@ jobs:
   smoke_test_windows_conda:
     <<: *binary_common
     executor:
-      name: windows-cpu
+        name: windows-cpu
     steps:
       - attach_workspace:
           at: ~/workspace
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -426,6 +426,32 @@ jobs:
             conda activate python${PYTHON_VERSION}
             python -c "import torchaudio"
 
+  smoke_test_windows_conda_gpu:
+    <<: *binary_common
+    executor:
+      name: windows-gpu
```
mthrok, line 432: nit: another way would be to parameterize the executor name, but I guess it's okay to duplicate the whole definition.

mthrok (approved):
Looks good. If you want to be extra sure, I think we can temporarily enable nightly workflow in the PR and see if this helps. I am okay with iterating on daily basis, if PR does not resolve it.

PR #2355, opened 2022-04-28
Use custom FFmpeg libraries for torchaudio binary distributions

mthrok:
This commit changes the way torchaudio binary distributions are built.

* For all the binary distributions (conda/pip on Linux/macOS/Windnows), build custom FFmpeg libraries.
* The custom FFmpeg libraries do not use `--use-gpl` nor `--use-nonfree`, so that they stay LGPL.
* The custom FFmpeg libraries employ rpath so that the torchaudio binary distributions look for the corresponding FFmpeg libraries installed in the runtime environment.
* The torchaudio binary build process will use them to bootstrap its build process.
* The custom FFmpeg libraries are NOT shipped.

This commit also add disclaimer about FFmpeg in README.


packaging/build_conda.sh
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -ex
 
+echo FFMPEG_ROOT=${FFMPEG_ROOT}
```
carolineechen, line 4: is this necessary or just used for debugging? (same for build_wheel.sh)

carolineechen (approved):

mthrok:
> Is the linux binary job related? https://app.circleci.com/pipelines/github/pytorch/audio/10603/workflows/d2da6a82-bd2f-4773-acea-c8e5fa2fb76c/jobs/641007

No, ROCm jobs are continuously failing these days.

PR #2360, opened 2022-05-04
Refactor mask_along_axis{,_iid} and add SpecAugment transform

xiaohui-zhang:
The previous way of doing SpecAugment via Frequency/TimeMasking transforms has the following problems:
- Only zero masking can be done; masking by mean value is not supported. 
- mask_along_axis is hard-coded to mask the 1st dimension and mask_along_axis_iid is hard-code to mask the 2nd or 3rd dimension of the input tensor. 
- For 3D spectrogram tensors where the first dimension is batch or channel, features from the same batch or different channels have to use the same mask, because mask_along_axis_iid only support 4D tensors, because of the above hard-coding
- For 2D spectrogram tensors w/o a batch or channel dimension, Time/Frequency masking can't be applied at all, since mask_along_axis only support 3D tensors, because of the above hard-coding. 
- It's not straightforward to apply multiple time/frequency masks by the current design.

To solve these issues, here we 
- Extend mask_along_axis_iid to support 3D tensors and mask_along_axis to support 2D tensors. Now both of them are able to mask one of the last two dimensions (where the time or frequency dimension lives) of the input tensor. 
- Add a SpecAugment transform which directly calls mask_along_axis{,_iid} with the flexibility of specifying number of time/frequency masks within a single transform, and zero/mean-value based masking. Note that if the input tensor has a dimension of 3 or 4, we always apply different masks along the first 1 or 2 dimensions.  

Additionally, TimeMasking/Frequency 
We may consider deprecating AxisMasking, TimeMasking and FrequencyMasking in future.

### DIFF
--- a/.github/workflows/build_wheels_windows.yml
+++ b/.github/workflows/build_wheels_windows.yml
@@ -1,0 +1,55 @@
+name: Build Windows Wheels
+
+on:
+  pull_request:
+  push:
+    branches:
+      - nightly
+      - main
+      - release/*
+    tags:
+        # NOTE: Binary build pipelines should only get triggered on release candidate builds
+        # Release candidate tags look like: v1.11.0-rc1
+        - v[0-9]+.[0-9]+.[0-9]+-rc[0-9]+
+  workflow_dispatch:
+
+jobs:
+  generate-matrix:
+    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
+    with:
+      package-type: wheel
+      os: windows
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+  build:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - repository: pytorch/audio
+            pre-script: packaging/ffmpeg/build.bat
+            env-script: packaging/vc_env_helper.bat
+            wheel-build-params: "--plat-name win_amd64"
+            post-script: ""
+            smoke-test-script: "test/smoke_test/smoke_test.py"
+            package-name: torchaudio
+    name: ${{ matrix.repository }}
+    uses: pytorch/test-infra/.github/workflows/build_wheels_windows.yml@main
+    with:
+      repository: ${{ matrix.repository }}
+      ref: ""
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: main
+      build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
+      pre-script: ${{ matrix.pre-script }}
+      env-script: ${{ matrix.env-script }}
+      post-script: ${{ matrix.post-script }}
+      package-name: ${{ matrix.package-name }}
+      smoke-test-script: ${{ matrix.smoke-test-script }}
+      # Using "development" as trigger event so these binaries are not uploaded
+      # to official channels yet
+      trigger-event: development
+    secrets:
+      AWS_PYTORCH_UPLOADER_ACCESS_KEY_ID: ${{ secrets.AWS_PYTORCH_UPLOADER_ACCESS_KEY_ID }}
+      AWS_PYTORCH_UPLOADER_SECRET_ACCESS_KEY: ${{ secrets.AWS_PYTORCH_UPLOADER_SECRET_ACCESS_KEY }}


torchaudio/functional/functional.py
```diff
@@ -864,25 +874,28 @@ def mask_along_axis(
...
         mask = mask.unsqueeze(-1)
 
     assert mask_end - mask_start < mask_param
 
     specgram = specgram.masked_fill(mask, mask_value)
 
+    # This line does nothing if specgram is 2D or 3D
     # unpack batch
-    specgram = specgram.reshape(shape[:-2] + specgram.shape[-2:])
+    # specgram = specgram.reshape(shape[:-2] + specgram.shape[-2:])
```
mthrok, line 898: With a quick skim, I cannot tell if the removal is intended or temporary. Is the intended?

torchaudio/transforms/_transforms.py
```diff
@@ -1242,6 +1241,61 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
...
+        self.zero_masking = zero_masking
+        self.n_time_masks = n_time_masks
+        self.time_mask_param = time_mask_param
+        self.n_freq_masks = n_freq_masks
+        self.freq_mask_param = freq_mask_param
+        self.iid_masks = iid_masks
+        self.p = p
+        self.zero_masking = zero_masking
+
+    def forward(self, specgram: Tensor) -> Tensor:
```
mthrok, line 1272: The pre-existing issue with the conventional making implementation is that it won't distinguish the padded frames and valid frames when batching samples. What do you think? We could extend (or add a new implementation) that performs masking in padding-aware manner.

mthrok:
> We may consider deprecating AxisMasking, TimeMasking and FrequencyMasking in future.

I am okay with that, but in that case, instead of extending them, I'd start a new implementation and keep them as-is, so that new implementations are tailored for proper implementations.

examples/asr/librispeech_conformer_rnnt/lightning.py
```diff
@@ -218,10 +218,7 @@ def __init__(
             FunctionalModule(lambda x: _piecewise_linear_log(x * _gain)),
             GlobalStatsNormalization(global_stats_path),
             FunctionalModule(lambda x: x.transpose(1, 2)),
-            torchaudio.transforms.FrequencyMasking(27),
-            torchaudio.transforms.FrequencyMasking(27),
-            torchaudio.transforms.TimeMasking(100, p=0.2),
-            torchaudio.transforms.TimeMasking(100, p=0.2),
+            torchaudio.transforms.SpecAugment(2,100,2,27,p=0.2),
```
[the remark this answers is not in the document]
xiaohui-zhang: will do thx

torchaudio/functional/functional.py
```diff
@@ -784,17 +784,22 @@ def mask_along_axis_iid(
     ``max_v = min(mask_param, floor(specgrams.size(axis) * p))`` otherwise.
 
     Args:
-        specgrams (Tensor): Real spectrograms `(batch, channel, freq, time)`
+        specgrams (Tensor): Real spectrograms `(batch, channel, freq, time)`, `(batch, freq, time)` or `(channel, freq, time)`.
         mask_param (int): Number of columns to be masked will be uniformly sampled from [0, mask_param]
         mask_value (float): Value to assign to the masked columns
-        axis (int): Axis to apply masking on (2 -> frequency, 3 -> time)
+        axis (int): Axis to apply masking on (3D specgrams:1 -> frequency, 2 -> time; 4D specgrams: 2 -> frequency, 3 -> time)
```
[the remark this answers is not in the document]
xiaohui-zhang: will do

torchaudio/functional/functional.py
```diff
@@ -864,25 +874,28 @@ def mask_along_axis(
     if mask_param < 1:
         return specgram
 
+    # This line does nothing if specgram is 2D or 3D
     # pack batch
-    shape = specgram.size()
-    specgram = specgram.reshape([-1] + list(shape[-2:]))
+    # shape = specgram.size()
+    # specgram = specgram.reshape([-1] + list(shape[-2:]))
```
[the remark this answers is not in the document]
xiaohui-zhang: sure

torchaudio/functional/functional.py
xiaohui-zhang: it's intended. I commented first just in order to get reviewers' confirmation before I finally remove it. @hwangjeff do you mind taking a look and confirming my judgement? i.e. line 898 and 880 are redundant if specgram is 2D or 3D.

torchaudio/transforms/_transforms.py
xiaohui-zhang: oh it's a good point. I thought about that and decided to leave it for later, since it'll make the function much more complicated and the effect should be minimal as we're already making sure utterances have similar lengths when batching data, so that padding areas are small

xiaohui-zhang:
> 

yeah agree. I'm exactly doing what you suggested: keeping them as they're. I extended mask_along_axis{,_iid} so that the new class SpecAugment can work properly.

xiaohui-zhang:
> > We may consider deprecating AxisMasking, TimeMasking and FrequencyMasking in future.
> 
> I am okay with that, but in that case, instead of extending them, I'd start a new implementation and keep them as-is, so that new implementations are tailored for proper implementations.

yeah agree. I'm exactly doing what you suggested: keeping them as they're. I extended mask_along_axis{,_iid} so that the new class SpecAugment can work properly.

mthrok:
> > > We may consider deprecating AxisMasking, TimeMasking and FrequencyMasking in future.
> > 
> > 
> > I am okay with that, but in that case, instead of extending them, I'd start a new implementation and keep them as-is, so that new implementations are tailored for proper implementations.
> 
> yeah agree. I'm exactly doing what you suggested: keeping them as they're. I extended mask_along_axis{,_iid} so that the new class SpecAugment can work properly.

What I mean was to even keep `mask_along_axis{,_iid}` as-is because it is public API. When I look at the implementation of them and related functions, I feel like the implementation can be more straightforward.

torchaudio/functional/functional.py
hwangjeff: they are indeed redundant if `specgram` has 2 or 3 dimensions, but maybe it's worth examining that premise. our spectrogram functions/transforms (which we expect users to use to produce inputs to this function) have the ability to produce outputs of shape `(..., freq, time)`, e.g. https://pytorch.org/audio/stable/transforms.html#torchaudio.transforms.Spectrogram.forward, https://pytorch.org/audio/stable/transforms.html#torchaudio.transforms.MelSpectrogram.forward. to achieve consistency with these, can we generalize the function to work on any tensor of shape `(..., freq, time)`? the line here seems to have been introduced with this in mind

torchaudio/functional/functional.py
```diff
@@ -784,17 +784,22 @@ def mask_along_axis_iid(
     ``max_v = min(mask_param, floor(specgrams.size(axis) * p))`` otherwise.
 
     Args:
-        specgrams (Tensor): Real spectrograms `(batch, channel, freq, time)`
+        specgrams (Tensor): Real spectrograms `(batch, channel, freq, time)`, `(batch, freq, time)` or `(channel, freq, time)`.
```
hwangjeff, line 787: similar comment — possible to generalize to `(..., freq, time)`?

torchaudio/transforms/_transforms.py
```diff
@@ -1242,6 +1241,61 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
...
+        r"""
+        Args:
+            specgram (Tensor): Tensor of dimension `(..., freq, time)`.
```
hwangjeff, line 1275: i like that we envision allowing for an arbitrary number of leading dimensions, but to make this a reality, we'd need to make the aforementioned changes to `mask_along_axis_iid` and `mask_along_axis`

torchaudio/transforms/_transforms.py
```diff
@@ -1242,6 +1241,61 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
...
+    Args:
+        n_time_masks (int): Number of time masks.
+        time_mask_param (int): Maximum possible length of the time mask.
+        n_freq_masks (int): Number of frequency masks.
+        freq_mask_param (int): Maximum possible length of the frequency mask.
+        iid_masks (bool): Applies iid masks to each of the examples in the batch dimension.
+            This option is applicable only when the input tensor is 4D.
+        p (float, optional): maximum proportion of time steps that can be masked.
+            Must be within range [0.0, 1.0]. (Default: 1.0)
+        zero_masking (bool): If True, use 0 as the mask value with 0, else use mean of the input tensor.
```
nateanl, line 1256:
```suggestion
        iid_masks (bool, optional): Applies iid masks to each of the examples in the batch dimension. (Default: ``True``)
            This option is applicable only when the input tensor is 4D.
        p (float, optional): maximum proportion of time steps that can be masked.
            Must be within range [0.0, 1.0]. (Default: 1.0)
        zero_masking (bool, optional): If True, use 0 as the mask value with 0, else use mean of the input tensor. (Default: ``False``)
```

torchaudio/transforms/_transforms.py
```diff
@@ -1242,6 +1241,61 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
...
+            Tensor: Masked spectrogram of dimensions `(..., freq, time)`.
+        """
+        if self.zero_masking:
+            mask_value = 0.0
+        else:
+            mask_value = specgram.mean()
+        time_dim = specgram.dim()-1
+        freq_dim = time_dim-1
+
+        if specgram.dim() > 2:
```
nateanl, line 1287: Is the method applicable to a 2D spectrogram?

xiaohui-zhang: no. for 2D transform we should use mask_along_axis instead

torchaudio/functional/functional.py
xiaohui-zhang: good idea thanks

torchaudio/functional/functional.py
```diff
@@ -824,7 +829,7 @@ def mask_along_axis_iid(
 
 
 def mask_along_axis(
-    specgram: Tensor,
+    specgrams: Tensor,
```
mthrok, line 832: Changing argument name is BC-breaking.

torchaudio/transforms/_transforms.py
```diff
@@ -1160,7 +1160,7 @@ def forward(self, specgram: Tensor, mask_value: float = 0.0) -> Tensor:
         Args:
             specgram (Tensor): Tensor of dimension `(..., freq, time)`.
             mask_value (float): Value to assign to the masked columns.
-
+
```
nateanl, line 1163:
```suggestion

```

torchaudio/transforms/_transforms.py
```diff
@@ -1243,6 +1243,72 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
         super(TimeMasking, self).__init__(time_mask_param, 2, iid_masks, p=p)
 
 
+class SpecAugment(torch.nn.Module):
```
nateanl, line 1246: does `SpecAug` sound better since Spec is short for `Spectrogram`?

torchaudio/transforms/_transforms.py
```diff
@@ -1243,6 +1243,72 @@ def __init__(self, time_mask_param: int, iid_masks: bool = False, p: float = 1.0
...
+
+    Args:
+        n_time_masks (int): Number of time masks.
+        time_mask_param (int): Maximum possible length of the time mask.
+        n_freq_masks (int): Number of frequency masks.
+        freq_mask_param (int): Maximum possible length of the frequency mask.
+        iid_masks (bool, optional): Applies iid masks to each of the examples in the batch dimension.
+            This option is applicable only when the input tensor is 4D. (Default: ``True``)
+        p (float, optional): maximum proportion of time steps that can be masked.
+            Must be within range [0.0, 1.0]. (Default: 1.0)
```
nateanl, line 1257: I feel those parameters can be put in `forward` method so that users have more flexibility to use the same module. Otherwise they have to create new `SpecAug` modules with different parameters, WDYT?

xiaohui-zhang: oh I think it's ok. given a specific certain training task, these parameters shouldn't change. so one instance of the class is enough

mthrok:
I think we need to add tests for the new capability. And perhaps can we split the PR into 1. changes to the existing functions and 2. addition of SpecAugment?

torchaudio/transforms/_transforms.py
xiaohui-zhang: how did you detect this lol

PR #2364, opened 2022-05-05
Run smoke tests on regular PRs

mthrok:
Currently smoke tests are only executed on nightly jobs.
This is inconvenient as PRs that changes build process do not get
the signal naturally.

This commit changes it by always executing smoke tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -775,6 +775,12 @@ workflows:
           python_version: '3.7'
           requires:
           - download_third_parties_nix
+      - smoke_test_linux_pip:
+          cuda_version: cpu
+          name: binary_linux_wheel_py3.7_cpu_smoke_test_pip
+          python_version: '3.7'
+          requires:
+          - binary_linux_wheel_py3.7_cpu
       - binary_linux_wheel:
           cuda_version: cu102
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -1428,7 +1818,7 @@ workflows:
           name: nightly_binary_linux_wheel_py3.7_cu113_smoke_test_pip
           python_version: '3.7'
           requires:
-          - nightly_binary_linux_wheel_py3.7_cu113_upload
```
atalman, line 1431: I think this is not correct. This change should not affect nightly tests.  The nightly_binary_linux_wheel_py3.7_cu113_smoke_test_pip should require nightly_binary_linux_wheel_py3.7_cu113_upload

.circleci/regenerate.py
```diff
@@ -186,13 +186,10 @@ def generate_upload_workflow(base_workflow_name, filter_branch, os_type, btype,
 
 def generate_smoketest_workflow(pydistro, base_workflow_name, filter_branch, python_version, cu_version, os_type):
 
-    required_build_suffix = "_upload"
```
atalman, line 189: I think we still need this step

PR #2365, opened 2022-05-05
Refactor smoke test executions

mthrok:
The smoke test jobs simply perform `import torchaudio` to check 
if the package artifacts are sane.
Originally, this was done in the  root directory of the CI env. 
This worked fine as long as they do not checkout the source code.
There are cases where source code has to be checked-out, and 
for these jobs, the smoke tests were not working as intended.

Since this error is very difficult to notice, this commit introduce a
common way to perform smoke test.


atalman (approved):
LGTM

PR #2368, opened 2022-05-06
Add RTFMVDR module

nateanl:
Add a new design of MVDR module.
The RTFMVDR module supports the method based on the relative transfer function (RTF) and power spectral density (PSD) matrix of noise.
The input arguments are:
- multi-channel spectrum.
- RTF vector of the target speech
- PSD matrix of noise.
- reference channel in the microphone array.
- diagonal_loading option to enable or disable diagonal loading in matrix inverse computation.
- diag_eps for computing the inverse of the matrix.
- eps for computing the beamforming weight.
The output of the module is the single-channel complex-valued spectrum for the enhanced speech.

PR #2369, opened 2022-05-06
Add diagonal_loading optional to rtf_power

nateanl:
When computing the MVDR beamforming weights using the power iteration method, the PSD matrix of noise can be applied with diagonal loading to improve the robustness. This is also applicable to computing the RTF matrix (See https://github.com/espnet/espnet/blob/master/espnet2/enh/layers/beamformer.py#L614 as an example). This also aligns with current `torchaudio.transforms.MVDR` module to keep the consistency.

This PR adds the `diagonal_loading` argument with `True` as default value to `torchaudio.functional.rtf_power`.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -1958,7 +1958,15 @@ def rtf_evd(psd_s: Tensor) -> Tensor:
     return rtf
 
 
-def rtf_power(psd_s: Tensor, psd_n: Tensor, reference_channel: Union[int, Tensor], n_iter: int = 3) -> Tensor:
+def rtf_power(
+    psd_s: Tensor,
+    psd_n: Tensor,
+    reference_channel: Union[int, Tensor],
+    n_iter: int = 3,
+    diagonal_loading: bool = True,
+    diag_eps: float = 1e-7,
+    eps: float = 1e-8,
+) -> Tensor:
     r"""Estimate the relative transfer function (RTF) or the steering vector by the power method.
 
     .. devices:: CPU CUDA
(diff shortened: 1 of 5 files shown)


carolineechen (commented):
was rtf_power included in the last release? the implementation looks good but will change the default behavior of the function to produce different values, so may need to be labeled as BC-breaking

nateanl:
> was rtf_power included in the last release? the implementation looks good but will change the default behavior of the function to produce different values, so may need to be labeled as BC-breaking

No, it was not in the previous release, I think we can regard the changes as BC. what do you think?

torchaudio/functional/functional.py
```diff
@@ -1975,12 +1983,20 @@ def rtf_power(psd_s: Tensor, psd_n: Tensor, reference_channel: Union[int, Tensor
             If the dtype is ``Tensor``, the dimension is `(..., channel)`, where the ``channel`` dimension
             is one-hot.
         n_iter (int): number of iterations in power method. (Default: ``3``)
+        diagonal_loading (bool, optional): whether to apply diagonal loading to psd_n
+            (Default: ``True``)
+        diag_eps (float, optional): The coefficient multiplied to the identity matrix for diagonal loading
+            (Default: ``1e-7``)
+        eps (float, optional): a value to avoid the correlation matrix is all-zero (Default: ``1e-8``)
```
carolineechen, line 1990: can this docstring be more helpful? does `a value added to the denominator in the beamforming weight computation.` from #2368 make sense here, and is it worth adding that this is only used for the case when `diagonal_loading=True`?

nateanl: Makes sense. I will align the eps docstring in the functions and modules.

carolineechen (approved):
oops think I was looking at the wrong branch and thought it was part of last release, that sounds good to me! just the docstring change then

torchaudio/functional/functional.py
nateanl: The `eps` here is for diagonal loading, which is confusing with `eps` in computing beamforming weight. I decided to exclude it from the API and use the default value in `_tik_reg`.

PR #2372, opened 2022-05-06
[testing] Enable nvfuser

davidberard98:
ahead of nvfuser enablement by default, we want to check for failures

### DIFF
--- a/.circleci/unittest/linux/scripts/run_test.sh
+++ b/.circleci/unittest/linux/scripts/run_test.sh
@@ -18,5 +18,5 @@ declare -a args=(
 )
 
 cd test
-pytest "${args[@]}" torchaudio_unittest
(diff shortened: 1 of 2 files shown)


mthrok (commented):
Hi @davidberard98 - Does Windows matter? Windows CI jobs use similar but separate script here. https://github.com/pytorch/audio/blob/main/.circleci/unittest/windows/scripts/run_test.sh#L15

davidberard98:
thanks! we don't expect windows to be much different, but I added it anyway just in case

davidberard98:
To summarize the failures:
- I didn't observe any NVFuser failures
- Many jobs are failing, but they are also failing on other PRs opened around the same time.
- Many CPU jobs began failing, but this is expected because the environment flag PYTORCH_JIT_ENABLE_NVFUSER=1 will attempt to turn on NVFuser, but NVFuser can't be turned on in cpu-only builds.

PR #2373, opened 2022-05-07
Refactor the constructors of pointer wrappers 

mthrok:
This commit refactor the constructor of wrapper classes so that
wrapper classes are only responsible for deallocation of underlying
FFmpeg custom structures.

The responsibility of custom initialization is moved to helper functions.

Context:

FFmpeg API uses bunch of raw pointers, which require dedicated allocater
and deallcoator. In torchaudio we wrap these pointers with
`std::unique_ptr<>` to adopt RAII semantics.

Currently all of the customization logics required for `Streamer` are
handled by the constructor of wrapper class. Like the following;

```
AVFormatContextPtr(
      const std::string& src,
      const std::string& device,
      const std::map<std::string, std::string>& option);
```

This constructor allocates the raw `AVFormatContext*` pointer,
while initializing it with the given option, then it parses the
input media.

As we consider the write/encode features, which require different way
of initializing the `AVFormatContext*`, making it the responsibility
of constructors of `AVFormatContextPtr` reduce the flexibility.

Thus this commit moves the customization to helper factory function.

- `AVFormatContextPtr(...)` -> `get_input_format_context(...)`
- `AVCodecContextPtr(...)` -> `get_decode_context(...)`

### DIFF
--- a/torchaudio/csrc/ffmpeg/decoder.cpp
+++ b/torchaudio/csrc/ffmpeg/decoder.cpp
@@ -11,7 +11,15 @@ Decoder::Decoder(
     const std::string& decoder_name,
     const std::map<std::string, std::string>& decoder_option,
     const torch::Device& device)
-    : pCodecContext(pParam, decoder_name, decoder_option, device) {}
+    : pCodecContext(get_decode_context(pParam->codec_id, decoder_name)) {
+  init_codec_context(
+      pCodecContext,
+      pParam,
+      decoder_name,
+      decoder_option,
+      device,
+      pHWBufferRef);
+}
 
 int Decoder::process_packet(AVPacket* pPacket) {
   return avcodec_send_packet(pCodecContext, pPacket);
--- a/torchaudio/csrc/ffmpeg/decoder.h
+++ b/torchaudio/csrc/ffmpeg/decoder.h
@@ -7,6 +7,7 @@ namespace ffmpeg {
 
 class Decoder {
   AVCodecContextPtr pCodecContext;
+  AVBufferRefPtr pHWBufferRef;
 
  public:
   // Default constructable
--- a/torchaudio/csrc/ffmpeg/ffmpeg.cpp
+++ b/torchaudio/csrc/ffmpeg/ffmpeg.cpp
@@ -62,7 +62,9 @@ std::string join(std::vector<std::string> vars) {
 #define AVINPUT_FORMAT_CONST
 #endif
 
-AVFormatContext* get_format_context(
+} // namespace
+
+AVFormatContextPtr get_input_format_context(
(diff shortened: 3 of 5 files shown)


torchaudio/csrc/ffmpeg/decoder.cpp
```diff
@@ -11,7 +11,15 @@ Decoder::Decoder(
...
-    : pCodecContext(pParam, decoder_name, decoder_option, device) {}
+    : pCodecContext(get_decode_context(pParam->codec_id, decoder_name)) {
+  init_codec_context(
+      pCodecContext,
+      pParam,
+      decoder_name,
+      decoder_option,
+      device,
+      pHWBufferRef);
+}
```
mthrok, line 22: Originally the codec context was initialized inside of `CodecContextPtr` class constructor, now it's done in call site.

torchaudio/csrc/ffmpeg/decoder.h
```diff
@@ -7,6 +7,7 @@ namespace ffmpeg {
 
 class Decoder {
   AVCodecContextPtr pCodecContext;
+  AVBufferRefPtr pHWBufferRef;
```
mthrok, line 10: `AVBufferRefPtr` has been moved out of `AVCodecContextPtr` to call site.

torchaudio/csrc/ffmpeg/ffmpeg.cpp
```diff
@@ -62,7 +62,9 @@ std::string join(std::vector<std::string> vars) {
 #define AVINPUT_FORMAT_CONST
 #endif
 
-AVFormatContext* get_format_context(
```
mthrok, line 65: `get_format_context` is now the factory function used by call site.

torchaudio/csrc/ffmpeg/ffmpeg.cpp
```diff
@@ -83,19 +85,11 @@ AVFormatContext* get_format_context(
...
-    const std::string& src,
-    const std::string& device,
-    const std::map<std::string, std::string>& option)
-    : Wrapper<AVFormatContext, AVFormatContextDeleter>(
-          get_format_context(src, device, option)) {
-  if (avformat_find_stream_info(ptr.get(), NULL) < 0)
-    throw std::runtime_error("Failed to find stream information.");
-}
+AVFormatContextPtr::AVFormatContextPtr(AVFormatContext* p)
+    : Wrapper<AVFormatContext, AVFormatContextDeleter>(p) {}
```
mthrok, line 92: `AVFormatContextPtr` is only responsible for pointer clean-up. The initialization logic has been moved to call site.

hwangjeff (approved):

Issue #2375, opened 2022-05-10
dtype Error on torchaudio.transforms.MVDR


CaA23187:
### 🐛 Describe the bug

the return of torchaudio.transforms.MVDR.forward() should be _return specgram_enhanced.to(dtype)_ rather than _return specgram_enhanced_, which mght cause dtype inconsistency between input and output when the dtype of input specgram is complex64

### Versions

PyTorch version: 1.11.0+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0
Clang version: Could not collect
CMake version: version 3.22.2
Libc version: glibc-2.31

Python version: 3.9.7 (default, Sep 16 2021, 13:09:58)  [GCC 7.5.0] (64-bit runtime)
Python platform: Linux-5.13.0-30-generic-x86_64-with-glibc2.31
Is CUDA available: True
CUDA runtime version: 10.1.243
GPU models and configuration: 
GPU 0: NVIDIA GeForce RTX 2080 Ti
GPU 1: NVIDIA GeForce RTX 2080 Ti
GPU 2: NVIDIA GeForce RTX 2080 Ti
GPU 3: NVIDIA GeForce RTX 2080 Ti

Nvidia driver version: 510.47.03
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.22.3
[pip3] numpydoc==1.1.0
[pip3] torch==1.11.0
[pip3] torch-tb-profiler==0.3.1
[pip3] torchaudio==0.11.0
[pip3] torchvision==0.12.0
[conda] blas                      1.0                         mkl  
[conda] mkl                       2021.4.0           h06a4308_640  
[conda] mkl-service               2.4.0            py39h7f8727e_0  
[conda] mkl_fft                   1.3.1            py39hd3c417c_0  
[conda] mkl_random                1.2.2            py39h51133e4_0  
[conda] mypy_extensions           0.4.3            py39h06a4308_0  
[conda] numpy                     1.22.3                   pypi_0    pypi
[conda] numpydoc                  1.1.0              pyhd3eb1b0_1  
[...]

PR #2376, opened 2022-05-10
Fix return dtype in MVDR module

nateanl:
Address #2375 
The MVDR module internally transforms the dtype of complex tensors to `torch.complex128` for computation and transforms it back to the original dtype before returning the Tensor. However, it didn't convert back successfully due to `specgram_enhanced.to(dtype)`, which should be `specgram_enhanced = specgram_enhanced.to(dtype)`. Fix it to make the output dtype consistent with original input.

### DIFF
--- a/test/torchaudio_unittest/transforms/transforms_test_impl.py
+++ b/test/torchaudio_unittest/transforms/transforms_test_impl.py
@@ -131,3 +131,28 @@ class TransformsTestBase(TestBaseMixin):
             psd_np = psd_numpy(spectrogram.detach().numpy(), mask, multi_mask)
         psd = transform(spectrogram, mask)
         self.assertEqual(psd, psd_np, atol=1e-5, rtol=1e-5)
+
+    @parameterized.expand(
+        [
+            param("ref_channel", True, torch.complex64),
+            param("ref_channel", False, torch.complex128),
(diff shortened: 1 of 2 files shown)


hwangjeff (approved):

mthrok (commented):
Can we have corresponding tests?

mthrok (approved):
Looks good. BTW can we move the implementation of MVDR into dedicated file, such as `torchaudio/transforms/_mvdr.py`?

mthrok:
> > Looks good. BTW can we move the implementation of MVDR into dedicated file, such as torchaudio/transforms/_mvdr.py?
> 
> Sure, I prefer using `_beamforming.py` or `_multi_channel.py` as there are other beamforming methods liks `GEV` or `WPD`, and other multi-channel methods like `WPE`, `IPD`, etc.
> 
> Or you mean for this specific `MVDR` module for deprecation purpose?

What I meant was to put all the multi channel transforms to dedicated script. 
`_transforms.py` is getting huge, but IIUC, multi-channel transforms are not logically connected to the rest of transforms.
Putting them in dedicated file should yield better [decoupling](https://en.wikipedia.org/wiki/Coupling_(computer_programming)).
 You can pick the name of the file, as long as it represents the module well.

PR #2382, opened 2022-05-11
Move multi-channel modules to a separate file

nateanl:
The modules include:
- PSD
- MVDR
- RTFMVDR
- SoudenMVDR


mthrok (approved):
The following method does not use `self`, so putting them outside of `MVDR` class would make the `MVDR` implementation lean and easier to follow.

- `MVDR._get_steering_vector_evd`
- `MVDR._get_steering_vector_power`
- `MVDR._apply_beamforming_vector`
- `MVDR._tik_reg`

nateanl:
> The following method does not use self, so putting them outside of MVDR class would make the MVDR implementation lean and easier to follow.
MVDR._get_steering_vector_evd
MVDR._get_steering_vector_power
MVDR._apply_beamforming_vector
MVDR._tik_reg

Yes, after this PR is merged, I will replace those methods with the ones in `functional`.

PR #2387, opened 2022-05-12
Refactor LibriSpeech dataset

hwangjeff:
Refactors `librispeech.py` to clarify its logic.

### DIFF
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -108,8 +108,13 @@ class LIBRISPEECH(Dataset):
         root = os.fspath(root)
         self._path = os.path.join(root, folder_in_archive, url)
 
-        if download and not os.path.isdir(self._path):
-            download_librispeech(root, url)
+        if not os.path.isdir(self._path):
+            if download:
+                download_librispeech(root, url)
+            else:
+                raise RuntimeError(
+                    f"Dataset not found at {self._path}. Please set `download=True` to download the dataset."
+                )
 
         self._walker = sorted(str(p.stem) for p in Path(self._path).glob("*/*/*" + self._ext_audio))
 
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/librispeech.py
```diff
@@ -23,22 +34,32 @@
...
+    base_url = "http://www.openslr.org/resources/12/"
+    download_url = os.path.join(base_url, filename)
+    if not os.path.isfile(archive):
+        checksum = _CHECKSUMS.get(download_url, None)
+        download_url_to_file(download_url, archive, hash_prefix=checksum)
+    extract_archive(archive)
+
+
 def load_librispeech_item(
-    fileid: str, path: str, ext_audio: str, ext_txt: str
```
mthrok, line 27: FYI: the reason why all the dataset load functions have this `ext_audio` argument is for testing. In unit test we generate mock data and we use WAV format, for the simplicity. (because there aren't many easy-to-use tool to generate non-WAV audio) To test the functionality of this function, we need overwrite of extension, so we have this extra argument.

see 
https://github.com/pytorch/audio/blob/595dc5d3a3ebf19a5a6d0f538bf2717f30f06b1a/test/torchaudio_unittest/datasets/librispeech_test.py#L88

hwangjeff: got it, thanks

torchaudio/datasets/librispeech.py
```diff
@@ -78,41 +96,20 @@ class LIBRISPEECH(Dataset):
...
-        self._path = os.path.join(root, folder_in_archive)
-
-        if download:
-            if not os.path.isdir(self._path):
-                if not os.path.isfile(archive):
-                    checksum = _CHECKSUMS.get(url, None)
-                    download_url_to_file(url, archive, hash_prefix=checksum)
-                extract_archive(archive)
+        if download and not os.path.isdir(self._path):
+            download_librispeech(root, url)
```
nateanl, line 112: shall we add an exception if `download` is `False` and there is no `_path` in the local disk? something like https://github.com/pytorch/audio/blob/main/torchaudio/datasets/cmudict.py#L144

nateanl (approved):
LGTM

PR #2400, opened 2022-05-18
Add file-like object support to Streaming API

mthrok:
This commit adds file-like object support to Streaming API.

## Features
- File-like objects are expected to implement `read(self, n)`.
- Additionally `seek(self, offset, whence)` is used if available.
- Without `seek` method, some formats cannot be decoded properly.
  - To work around this, one can use the existing `decoder` option to tell what decoder it should use.
  - The set of `decoder` and `decoder_option` arguments were added to `add_basic_[audio|video]_stream` method, similar to `add_[audio|video]_stream`.
  - So as to have the arguments common to both audio and video in front of the rest of the arguments, the order of the arguments are changed.
  - Also `dtype` and `format` arguments were changed to make them consistent across audio/video methods.

## Code structure

The approach is very similar to how file-like object is supported in sox-based I/O.
In Streaming API if the input src is string, it is passed to the implementation bound with TorchBind,
if the src has `read` attribute, it is passed to the same implementation bound via PyBind 11.

Untitled drawing

## Refactoring involved
- Extracted to #2402
  - Some implementation in the original TorchBind surface layer is converted to Wrapper class so that they can be re-used from PyBind11 bindings. The wrapper class serves to simplify the binding.
  - `add_basic_[audio|video]_stream` methods were removed from C++ layer as it was just constructing string and passing it to `add_[audio|video]_stream` method, which is simpler to do in Python.
  - The original core Streamer implementation kept the use of types in `c10` namespace minimum. All the `c10::optional` and `c10::Dict` were converted to the equivalents of `std` at binding layer. But since they work fine with PyBind11, Streamer core methods deal them directly.

## TODO:
- [x] Check if it is possible to stream MP4 (yuv420p) from S3 and directly decode (with/without HW decoding).

### DIFF
--- a/torchaudio/io/_stream_reader.py
+++ b/torchaudio/io/_stream_reader.py
@@ -29,21 +29,21 @@ class StreamReaderSourceStream:
        Still images, such as PNG and JPEG formats are reported as `video`.
     """
     codec: str
-    """Short name of the codec. Such as `pcm_s16le` and `h264`."""
+    """Short name of the codec. Such as ``"pcm_s16le"`` and ``"h264"``."""
     codec_long_name: str
     """Detailed name of the codec.
 
-    Such as `"PCM signed 16-bit little-endian"` and `"H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10"`.
+    Such as "`PCM signed 16-bit little-endian`" and "`H.264 / AVC / MPEG-4 AVC / MPEG-4 part 10`".
     """
     format: Optional[str]
-    """Media format. Such as `s16` and `yuv420p`.
+    """Media format. Such as ``"s16"`` and ``"yuv420p"``.
 
     Commonly found audio values are;
 
-    - `u8`, `u8p`: Unsigned 8-bit unsigned interger.
-    - `s16`, `s16p`: 16-bit signed integer.
-    - `s32`, `s32p`: 32-bit signed integer.
-    - `flt`, `fltp`: 32-bit floating-point.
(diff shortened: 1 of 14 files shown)


torchaudio/io/_stream_reader.py
```diff
@@ -232,51 +295,63 @@ class StreamReader:
             You can use this argument to change the input source before it is passed to decoder.
 
             Default: ``None``.
+
+        buffer_size (int):
+            The internal buffer size in byte. Unsed only when `src` is file-like object.
```
carolineechen, line 300:
```suggestion
            The internal buffer size in byte. Used only when `src` is file-like object.
```

torchaudio/io/_stream_reader.py
```diff
@@ -182,24 +174,95 @@ def _get_vfilter_desc(frame_rate: Optional[float], width: Optional[int], height:
...
+        return obj
+
+    return decorator
+
+
+_frames_per_chunk = """Number of frames returned as one chunk.
+                If the source stream is exhausted before enough frames are buffered,
+                then the chunk is returned as-is."""
+
+_buffer_chunk_size = """Internal buffer size.
```
carolineechen, line 194: could you add default args to the optional parameters as well?

mthrok: done!

torchaudio/io/_stream_reader.py
mthrok: done!

carolineechen (approved):

PR #2401, opened 2022-05-18
Add assertion checks to multi-channel functions

nateanl:
- The multi-channel functions only support complex-valued tensors for spectrogram and PSD matrices.
- The mask can be real-valued or complex-valued, hence there is no explicit assertion for mask.
- The shape of input Tensors need to be verified before the computation. For example, the shape of PSD matrix must be `(..., freq, channel, channel)`, the shape of the mask must be `(..., freq, time)`, etc.
- The autograd unittest of `apply_beamforming` has wrong dimensions for beamform_weights detected by the assertion check. FIx it in this PR.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -2099,7 +2099,7 @@ def apply_beamforming(beamform_weights: Tensor, specgram: Tensor) -> Tensor:
             Tensor of dimension `(..., freq, time)`
     """
     assert (
-        beamform_weights.shape[:-2] == specgram.shape[:-2]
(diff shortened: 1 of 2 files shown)


torchaudio/functional/functional.py
```diff
@@ -1861,6 +1863,21 @@ def mvdr_weights_souden(
     Returns:
         torch.Tensor: The complex-valued MVDR beamforming weight matrix with dimensions `(..., freq, channel)`.
     """
+    assert (
```
carolineechen, line 1866: since these assert statements are the same as for `rtf_power`, could you extract them into a common function to reuse?

carolineechen (approved):

PR #2402, opened 2022-05-18
Refactor Streamer implementation

mthrok:
* Move the helper wrapping code in TorchBind layer to proper wrapper class for so that it will be re-used in PyBind11.
* Move `add_basic_[audio|video]_stream` methods from C++ to Python, as they are just string manipulation. This will make PyBind11-based binding simpler as it needs not to deal with dtype.
* Move `add_[audio|video]_stream` wrapper signature to Streamer core, so that Streamer directly deals with `c10::optional` and `c10::Dic`. This reduces the code and gets rid of intermediate `std::map` structure.†

† Related to this, there is a slight change in how the empty filter expression is stored. Originally, if an empty filter expression was given to `add_[audio|video]_stream` method, the `StreamReaderOutputStream` was showing it as empty string `""`, even though internally it was using `"anull"` or `"null"`. Now `StreamReaderOutputStream` shows the corresponding filter expression that is actually being used.


torchaudio/csrc/ffmpeg/prototype.cpp
```diff
@@ -1,402 +1,129 @@
...
+    return ret;
+  }
+  for (const auto& it : dict.value()) {
+    ret.insert({it.key(), it.value()});
   }
-  return opts;
+  return ret;
 }
 
-struct StreamerHolder : torch::CustomClassHolder {
```
mthrok, line 23: The implementation of `StreamHolder` has been moved to StreamReaderBinding in `stream_reader_wrapper.[h|cpp]`.

nateanl (approved):

PR #2404, opened 2022-05-18
ci: Install libomp on macos

seemethere:


mthrok:
> > LGTM! Thanks! please take a look at the failing lint.
> 
> lint appears to have already been broken and is unrelated to this PR

Yeah, lint is broken due to the recent change on fbcode side, which I am having trouble resolving on GitHub. You can ignore it.

mthrok:
Thanks for finding the solution. Do you know what introduced this regression?

As far as I know;
 - We did not change the configuration
 - CircleCI Support told me that they did not change macOS image recently.
 - CMake is the same version.

seemethere:
> Thanks for finding the solution. Do you know what introduced this regression?
> 
> As far as I know;
> 
> * We did not change the configuration
> * CircleCI Support told me that they did not change macOS image recently.
> * CMake is the same version.

Not entirely sure, I did a search for if / when we used to install `libomp` and turns out we didn't. I also searched through the nightly commits for both `pytorch/pytorch` and `pytorch/audio` and found no related commits. Then I went to search through `pytorch/builder` for related commits and found nothing.

The only really plausible explanation to me is that something in the base image had changed that was no longer there but if CircleCI says that didn't happen then that's also not the explanation.

PR #2406, opened 2022-05-20
Install libomp from conda

atalman:
Install libomp from conda

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,7 +59,7 @@ commands:
       - run:
           name: Install ci dependencies
           command: |
-              brew install libomp
+              conda install -yq libomp
 
 binary_common: &binary_common
   parameters:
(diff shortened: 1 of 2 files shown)


mthrok:
Unit test, packaging (PyPI) and packaging (Conda) use different mechanism to initialize the environment, so unlike brew approach, `conda install` has to happen at the precise location.

- Conda packaging
   Note that conda packaging create a separate build environment transparent from us, which is described in [meta.yml](https://github.com/pytorch/audio/blob/6776299322ae0c8108b34af947e6e62ccd61b5ca/packaging/torchaudio/meta.yaml#L14), so I think adding the dependency to `host` section is the way.
- PyPI packaging
   Probably in [pkg_helpers.bash](https://github.com/pytorch/audio/blob/6776299322ae0c8108b34af947e6e62ccd61b5ca/packaging/pkg_helpers.bash#L140-L145)
- Unittest
   In [install.sh](https://github.com/pytorch/audio/blob/main/.circleci/unittest/linux/scripts/install.sh#L72)

PR #2407, opened 2022-05-21
Fix documentation initialization

mthrok:
The nbsphinx extension introduced in #2393 pulled a requirejs
which caused the initialization script to halt.

As a result, the right side bar was left uninitialized.
Also it turned out that PyTorch's theme expected the downstream projects
to define `collapsedSections` variable.

This commit fixes doc page initialization addressing the above.

Also this commit tweaks the right-side bar initialization behavior
so that expand-all only happens once, not at every resize.


carolineechen (approved):
changes 1-3 look good. 

I'm not too convinced about the GitHub override, since the tabs are all general PyTorch related links (Mobile/Tutorials/Resources etc are related to PyTorch and not torchaudio), so I think it does make sense of GitHub directed to PyTorch repo. I do see where you're coming from though and don't have a strong opinion so am ok with this change. Alternatively, thoughts on proposing for the pytorch website to have a dropdown for GitHub tab similar to Docs, for the main different repos?

mthrok:
> changes 1-3 look good.
> 
> I'm not too convinced about the GitHub override, since the tabs are all general PyTorch related links (Mobile/Tutorials/Resources etc are related to PyTorch and not torchaudio), so I think it does make sense of GitHub directed to PyTorch repo. I do see where you're coming from though and don't have a strong opinion so am ok with this change. Alternatively, thoughts on proposing for the pytorch website to have a dropdown for GitHub tab similar to Docs, for the main different repos?

That's a good idea, I was thinking to maybe making the href configurable for each project. Let me move on with this one.

PR #2410, opened 2022-05-25
Move CTC beam search decoder to beta

carolineechen:
Move CTC beam search decoder out of prototype to new `torchaudio.models.decoder` module.

@hwangjeff @mthrok any thoughts on the new module + naming, and if we should move rnnt beam search here as well??


mthrok:
The unit test issue was resolved in #2422. Can you rebase?

mthrok:
> @hwangjeff @mthrok any thoughts on the new module + naming, and if we should move rnnt beam search here as well??

If there were many decoders, then it would be good to have `torchaudio.decoder`, but since we do not and we are not sure how decoding effort continues, so I think `torchaudio.model.decoder` is a good middle ground.

Now sure about RNN-T beam search decoder. On the very surface, the file name is `torchaudio.models.rnnt_decoder.py`, so moving it to `torchaudio.models.decoder.rnnt.py` seems aligned, but unlike CTC decoder, which can work stand-alone, RNN-T requires underlying RNN models, so it's highly dependent on other NN models. We need to be confident that the interface of RNN and RNN-T beam search decoder are generic enough.

torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -293,44 +293,6 @@ def ctc_decoder(
     )
 
 
-def lexicon_decoder(
```
mthrok, line 296: Could you remind me what was the reason behind of the removal? Is there an alternative API?

mthrok:
Could you add deprecation warning for users who were importing it from `torchaudio.prototype.ctc_decoder` namespace?
Like [this]( https://github.com/pytorch/audio/blob/22a5d08448c9756f6a2ba536db9dca9592b68727/torchaudio/prototype/io/__init__.py)

torchaudio/models/decoder/_ctc_decoder.py
nateanl: I think this is included in `ctc_decoder`, by adding `lexicon` to it.

carolineechen:
@mthrok added the deprecation warning [here](https://github.com/pytorch/audio/pull/2410/commits/b66abbc04fae44d4ef545d0499aece127cc1651e), could you take a look if it looks fine? the functions are able to import and run successfully but this form of warning `DeprecationWarning` (also used for Streamer) is ignored by default, should we change it to ensure users are aware?

torchaudio/models/decoder/_ctc_decoder.py
carolineechen: `lexicon_decoder` functionality is now part of `ctc_decoder`. The function was left when we changed the naming/function to not break anyone's code who was previously using `lexicon_decoder`. In the beta stage we only need to keep `ctc_decoder` as `torchaudio.models.decoder.lexicon_decoder` is not something that people ever used. I added back the deprecation warning in prototype for both of these functions [here](https://github.com/pytorch/audio/pull/2410/commits/b66abbc04fae44d4ef545d0499aece127cc1651e) so that previous code using these functions don't break

torchaudio/prototype/ctc_decoder/__init__.py
```diff
@@ -1,42 +1,29 @@
...
+    elif name == "download_pretrained_files":
+        import warnings
 
-    global _INITIALIZED
-    _INITIALIZED = True
+        from torchaudio.models.decoder import download_pretrained_files
 
+        return download_pretrained_files
 
-def __getattr__(name: str):
```
carolineechen, line 20: @mthrok would we need to keep this code still?

torchaudio/models/decoder/_ctc_decoder.py
mthrok: sounds good, thanks for the clarification.

mthrok:
> @mthrok added the deprecation warning [here](https://github.com/pytorch/audio/pull/2410/commits/b66abbc04fae44d4ef545d0499aece127cc1651e), could you take a look if it looks fine? the functions are able to import and run successfully but this form of warning `DeprecationWarning` (also used for Streamer) is ignored by default, should we change it to ensure users are aware?

Oh, that's an oversight on my end. Yes, we should be using `UserWarning`. (though, we will be removing these prototype modules soon after the release)

mthrok (approved):

torchaudio/prototype/ctc_decoder/__init__.py
mthrok: Not 100% which part you are pointing but if you are referring to the lazy import mechanism, then yes.
Since we employ lazy initialization mechanism for CTC decoder in `torchaudio.models.decoder` module, we also need to do the same in `torchaudio.prototype.ctc_decoder`, otherwise the lazy initialization mechanism in `torchaudio.models.decoder` will be materialized as soon.

However, now that I think about it, the code and build for CTC decoder is 100% under our control (unlike the availability of CUDA or FFmpeg and such), so maybe we can get rid of the lazy initialization for CTC decoder. Let's follow up on it later. I think this PR is good to go.

PR #2411, opened 2022-05-25
Add conv_tasnet_base factory function to prototype

nateanl:


torchaudio/prototype/models/conv_tasnet.py
```diff
@@ -0,0 +1,26 @@
...
+
+    Args:
+        num_sources (int, optional): Number of sources in the output.
+            (Default: 2)
+    Returns:
+        ConvTasNet:
+            ConvTasNet model.
+    """
+    return ConvTasNet(
+        num_sources=num_sources,
```
carolineechen, line 17: where did you get these parameters from?

nateanl: Actually, there is one difference. In the official implementation (https://github.com/naplab/Conv-TasNet/blob/master/conv_tasnet.py), the mask activation function is sigmoid. Here we use "relu", which is the same as Asteroid (https://github.com/asteroid-team/asteroid/blob/master/egs/librimix/ConvTasNet/local/conf.yml) and ESPnet (https://github.com/espnet/espnet/blob/5d04018ce959bc17b7855067a9b088bfe0d450db/egs2/wsj0_2mix/enh1/conf/tuning/train_enh_conv_tasnet.yaml#L50).

carolineechen: got it, thanks. is there a reason why they use relu and why we should also use relu even though it deviates from the paper? since we are calling this the base implementation

nateanl: That's a good point. Jian Wu made comparisons of different activation settings (https://github.com/funcwj/conv-tasnet), with "relu" gets 0.3dB improvement than "sigmoid". Based on that, I think it's good to use the same settings as Asteroid and ESPnet, to keep the same performance number. 

That said, I think we need to change the description of the factory function, something like "the optimal setting from the paper, and change the mask activation function from ``sigmoid`` to  ``relu`` for performance boost".

carolineechen: yup that sounds good to me to update the description

carolineechen (approved):

PR #2425, opened 2022-05-31
Move Seed to Setup

skim0514:
Bringing in move seed commit from previous open commit #2267. Organizes seed to utils.


test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -337,8 +312,8 @@ def test_mvdr_weights_rtf_with_tensor(self):
             (3, False),
         ]
     )
-    def test_rtf_power(self, n_iter, diagonal_loading):
-        torch.random.manual_seed(2434)
+
+    def test_rtf_power(self, n_iter):
```
mthrok, line 316: The `diagonal_loading` argument should not be removed.

skim0514: Should it be fine to remove the manual_seed however?

carolineechen: yes manual seed should be removed here, as it's set in the setup function

nateanl:
We can remove all manual seed calls out of setUP method, if the class inherits from `TestBaseMixin`
For example, https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/functional/librosa_compatibility_test_impl.py#L134 
and https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/functional/torchscript_consistency_impl.py#L272

carolineechen:
> We can remove all manual seed calls that are outside the setUP method, if the class inherits from TestBaseMixin

Thinking we can focus this first PR on just pulling in and closing the original commit from #2267 (which handles functional and seemed to passing), and for familiarization with TorchAudio dev workflow. We can then follow up on removing `manual_seed` from the other files in a separate PR

nateanl:
Oh I see, didn't realize the commit is from a previous PR. Yeah we can merge this one and address the rest in separate PR.

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -337,8 +312,8 @@ def test_mvdr_weights_rtf_with_tensor(self):
             (3, False),
         ]
     )
+
```
nateanl, line 315:
```suggestion
```

test/torchaudio_unittest/functional/autograd_impl.py
```diff
@@ -351,8 +326,8 @@ def test_rtf_power(self, n_iter, diagonal_loading):
...
```
nateanl, line 329:
```suggestion
```

nateanl (approved):
LGTM. Just need to remove the two extra blank lines

carolineechen:
@mthrok @nateanl looks like one of the `T.TimeStretch` tests is failing now as a result of setup seed changing

mthrok:
> @mthrok @nateanl looks like one of the `T.TimeStretch` and hubert finetuning tests is failing now as a result of setup seed changing

Can we revert the changes for those individual tests? Ideally, tests should not depend on seed value, but for now we can revert and move on, then come back to it.

nateanl:
@mthrok  This PR actually didn't touch the code for TimeStretch, but since the transform's `AutogradTestMixin` class inherits from `TestBaseMixin`, it is affected.

mthrok:
> @mthrok This PR actually didn't touch the code for TimeStretch, but since the transform's `AutogradTestMixin` class inherits from `TestBaseMixin`, it is affected.

That's strange. The code for `get_whitenoise` has [this particular logic](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/data_utils.py#L68-L70) so that it is not affected by the global seed. And I do not see randomness in the implementation of the vocoder.

This implies that it is the randomness from `gradcheck` function. Okay, in such case, increasing the threshold seems like the proper approach.

test/torchaudio_unittest/models/wav2vec2/fairseq_integration_test.py
```diff
@@ -207,6 +207,7 @@ def test_import_finetuning_model(self, config, _):
         imported = import_fairseq_model(original).eval()
 
         # Without mask
+        torch.manual_seed(0)
```
skim0514, line 210: Added line to fix failing test for CircleCI

mthrok: For the context of this PR, it looks good.

carolineechen (approved):
looks good!

PR #2426, opened 2022-05-31
Dataset doc fixes

carolineechen:


torchaudio/datasets/quesst14.py
```diff
@@ -27,10 +27,10 @@ class QUESST14(Dataset):
 
     Args:
         root (str or Path): Root directory where the dataset's top level directory is found
+        subset (str): Subset of the dataset to use. Options: [``"docs"``, ``"dev"``, ``"eval"``].
```
nateanl, line 30: shall we follow this style (add double quotation marks) for the string options in ``language`` and other datasets?

nateanl (approved):

torchaudio/datasets/quesst14.py
carolineechen: hm this seems somewhat inconsistent across the datasets but I don't have a strong preference either way as I think both are readable. Let's address all datasets at once in a future PR if we decide to do this?

mthrok: I tend to add double quotation otherwise it can be misleading. i.e. ``1`` vs ``"1"``.

nateanl: sure, IMO following the same pattern can help future open-source contributors have a good reference if they want to add a new dataset to torchaudio. We can refactor all datasets in a separate PR.

mthrok: The order of arguments are different in docstring and constructor. If requiring subset, can we move the position as well? This dataset has not been released, so it should be fine.

mthrok (approved):

PR #2429, opened 2022-06-01
Tweak StreamReader error messages and tests

mthrok:
* Update error messages
* Update audio stream tests


test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -360,6 +360,20 @@ def test_seek_negative(self):
             s.seek(-1.0)
 
 
+def _to_fltp(original):
```
nateanl, line 363: can you add doc string here? This helps future maintainer understand what's the functionality of the method.

nateanl (approved):

test/torchaudio_unittest/io/stream_reader_test.py
mthrok: sure, done.

PR #2434, opened 2022-06-02
Retrieve version from version.txt 

atalman:
Retrieve version from version.txt
These improvement introduced in following PR: https://github.com/pytorch/vision/pull/6117
In addition to this we add version.txt file to help us manage torchaudio version


packaging/pkg_helpers.bash
```diff
@@ -97,10 +97,19 @@ setup_cuda() {
...
 setup_build_version() {
   if [[ -z "$BUILD_VERSION" ]]; then
-    export BUILD_VERSION="$1.dev$(date "+%Y%m%d")$VERSION_SUFFIX"
+    if [[ -z "$1" ]]; then
+      SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+      # version.txt for some reason has `a` character after major.minor.rev
+      # command below yields 0.10.0 from version.txt containing 0.10.0a0
+      _VERSION_BASE=$( cut -f 1 -d a "$SCRIPT_DIR/../version.txt" )
+    else
+      _VERSION_BASE="$1"
```
mthrok, line 110: Is there an existing use case for this `else` branch?
If not, it might introduce accidental/unexpected issue.
I think validating `$# -eq 0` is more error prof.

mthrok (commented):
Could we split the PR in two?. One for changes related to version.txt and the other for other M1 changes.

atalman:
> Could we split the PR in two?. One for changes related to version.txt and the other for other M1 changes.

Hm, I think in original M1 PR, the version was incorrect. But sure I can get the M1 changes out of this and then post another PR for M1

mthrok (approved):

PR #2435, opened 2022-06-02
Update QUESST14 getitem

carolineechen:
update QUESST14 getitem to include docstrings and additionally return sample rate


nateanl (approved):
LGTM

PR #2436, opened 2022-06-02
Remove possible manual seeds from test files. 

skim0514:
For test files where applicable, removed manual seeds where applicable. Refactoring #2267


test/torchaudio_unittest/functional/batch_consistency_test.py
```diff
@@ -32,12 +32,10 @@ def assert_batch_consistency(self, functional, inputs, atol=1e-8, rtol=1e-5, see
         for i in range(1, len(inputs)):
             self.assertEqual(inputs[i].size(0), n)
         # Compute items separately, then batch the result
-        torch.random.manual_seed(seed)
         items_input = [[ele[i].clone() for ele in inputs] for i in range(n)]
         items_result = torch.stack([functional(*items_input[i]) for i in range(n)])
 
         # Batch the input and run
-        torch.random.manual_seed(seed)
```
mthrok, line 40: These two seeds cannot be removed. This helper function is supposed to run functions on the same seed twice and compare the results.

test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -16,12 +16,10 @@ def assert_batch_consistency(self, transform, batch, *args, atol=1e-8, rtol=1e-5
         n = batch.size(0)
 
         # Compute items separately, then batch the result
-        torch.random.manual_seed(seed)
```
mthrok, line 19: Same here, the helper method must run the given function twice with the same seed. So these two seedings in this method have to stay.

mthrok (approved):
LG given tests pass.

PR #2439, opened 2022-06-03
Make FFmpeg log level configurable

mthrok:
and set the default level to 8 for less verbose logging.

Undesired logs are one of the loudest UX complains we get.
Yet, loading media files involves uncertainty which is
difficult to debug without debug log.

This commit introduces utility functions to configure logging level
so that we can ask users to enable it when they encounter an issue,
while defaulting to non-verbose option.

### DIFF
--- a/test/torchaudio_unittest/common_utils/case_utils.py
+++ b/test/torchaudio_unittest/common_utils/case_utils.py
@@ -8,6 +8,7 @@ import time
 import unittest
 
 import torch
+import torchaudio
 from torch.testing._internal.common_utils import TestCase as PytorchTestCase
 from torchaudio._internal.module_utils import (
     is_kaldi_available,
@@ -112,22 +113,8 @@ class TorchaudioTestCase(TestBaseMixin, PytorchTestCase):
     pass
 
 
-_IS_FFMPEG_AVAILABLE = None
-
-
 def is_ffmpeg_available():
-    if _eval_env("TORCHAUDIO_TEST_IN_FBCODE", default=False):
-        return True
-
-    global _IS_FFMPEG_AVAILABLE
-    if _IS_FFMPEG_AVAILABLE is None:
-        try:
-            from torchaudio.io import StreamReader  # noqa: F401
-
-            _IS_FFMPEG_AVAILABLE = True
-        except Exception:
-            _IS_FFMPEG_AVAILABLE = False
-    return _IS_FFMPEG_AVAILABLE
+    return torchaudio._extension._FFMPEG_INITIALIZED
 
 
 _IS_CTC_DECODER_AVAILABLE = None
--- a/test/torchaudio_unittest/utils/ffmpeg_utils_test.py
+++ b/test/torchaudio_unittest/utils/ffmpeg_utils_test.py
@@ -1,0 +1,22 @@
+from torchaudio.utils import ffmpeg_utils
+from torchaudio_unittest.common_utils import PytorchTestCase, skipIfNoFFmpeg
+
+
+@skipIfNoFFmpeg
+class TestFFmpegUtils(PytorchTestCase):
+    """Smoke test for ffmpeg_utils module"""
+
+    def tearDown(self):
+        ffmpeg_utils.set_log_level(8)
+        super().tearDown()
+
+    def test_get_log_level(self):
+        """`get_log_level` does not exhibit abnormal behavior"""
+        for _ in range(10):
+            ffmpeg_utils.get_log_level()
+
+    def test_set_log_level(self):
+        """`set_log_level` persists log level"""
+        for i in range(-100, 100):
+            ffmpeg_utils.set_log_level(i)
+            assert ffmpeg_utils.get_log_level() == i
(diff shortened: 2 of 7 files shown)


torchaudio/utils/ffmpeg_utils.py
```diff
@@ -0,0 +1,47 @@
...
+    return torch.ops.torchaudio.ffmpeg_get_log_level()
+
+
+def set_log_level(level: int):
+    """Set the log level of FFmpeg (libavformat etc)
+
+    Arguments:
+        level (int): Log level. The larger, the more verbose.
+
+            The following values are valid values, the corresponding ``ffmpeg``'s
```
carolineechen, line 18: what happens if they enter an invalid value? should we add a check?

carolineechen (commented):
utility function looks good, just want to double check default log value -- description mentions 0 but it is 8 in the code. generally enabling it for errors (16) by default makes more sense to me, but do you find this to be too noisy?

> we can ask users to enable it when they encounter an issue
is there a good place we could add this message upon failure so users are aware of this log setting utility

torchaudio/utils/ffmpeg_utils.py
mthrok: I think they will boil down to a simple level comparison, so it's safe to put other values.

https://ffmpeg.org/doxygen/4.1/log_8c_source.html#l00316

It's just normally, log-levels are set by `ffmpeg` CLI, which exposes the listed values. Let me rephrase, `common values`, instead of saying `valid values`.

mthrok:
> description mentions 0 but it is 8 in the code. 

Description also says 8. In test code it's reset to 0. 

> generally enabling it for errors (16) by default makes more sense to me, but do you find this to be too noisy?

16 can repeat messages across decoding units, so, decoding one audio file can still flood the log.

> is there a good place we could add this message upon failure so users are aware of this log setting utility

Decoder errors are not necessarily failure, that's why it's tricky. You would still get the tensor and keep going. But if you are training a model with many files, then the size of the training log could blow up with the decoding issues that is not an issue for DL practitioners.

carolineechen (approved):
>Description also says 8.

I meant it says 0 in the PR description, but yea code looks good if 8 is intended.

For the second point, was more wondering if there was a good spot in the ffmpeg documentation or logging output to let users know that this logging functionality exists (this is not blocking)

test/torchaudio_unittest/utils/ffmpeg_utils_test.py
```diff
@@ -0,0 +1,22 @@
+from torchaudio.utils import ffmpeg_utils
+from torchaudio_unittest.common_utils import skipIfNoFFmpeg, PytorchTestCase
+
+
+@skipIfNoFFmpeg
+class TestFFmpegUtils(PytorchTestCase):
+    """Smoke test for ffmpeg_utils moduls"""
```
hwangjeff, line 7:
```suggestion
    """Smoke test for ffmpeg_utils module"""
```

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -112,22 +113,8 @@ class TorchaudioTestCase(TestBaseMixin, PytorchTestCase):
...
-    global _IS_FFMPEG_AVAILABLE
-    if _IS_FFMPEG_AVAILABLE is None:
-        try:
-            from torchaudio.io import StreamReader  # noqa: F401
-
-            _IS_FFMPEG_AVAILABLE = True
-        except Exception:
-            _IS_FFMPEG_AVAILABLE = False
-    return _IS_FFMPEG_AVAILABLE
+    return torchaudio._extension._FFMPEG_INITIALIZED
```
hwangjeff, line 117: does this change relate to log level configurability?

hwangjeff (commented):
have you considered introducing constants for the different log levels, e.g. `FATAL` and `WARNING`, similar to `logging`?

test/torchaudio_unittest/common_utils/case_utils.py
mthrok: not technically. but as I write the test, I realized that the logic is duplicated in library and test util.

mthrok:
> I meant it says 0 in the PR description, but yea code looks good if 8 is intended.
> 

Oh my bad, fixed the description. 

> For the second point, was more wondering if there was a good spot in the ffmpeg documentation or logging output to let users know that this logging functionality exists (this is not blocking)

Good point, let me update the doc.

mthrok:
> have you considered introducing constants for the different log levels, e.g. `FATAL` and `WARNING`, similar to `logging`?

I thought about it, but unlike the `logging` module which is all about logging, this utility module is things more general about ffmpeg runtime configuration, so, I did not think putting alias at module level makes much sense, then thought adding some ENUM-like features for the `set_log_level` is kinda overkill.

mthrok:
@carolineechen let me revert the default log level modification, as it might be disruptive if other libraries / existing code base expects log level with already set.

hwangjeff (approved):

PR #2440, opened 2022-06-03
Add SourceSeparationBundle to prototype

nateanl:
- Add SourceSeparationBundle class for source separation pipeline
- Add `CONVTASNET_BASE_LIBRI2MIX` that is trained on Libri2Mix dataset.
- Add integration test with example mixture audio and expected scale-invariant signal-to-distortion ratio (Si-SDR) score. The test computes the Si-SDR score with permutation-invariant training (PIT) criterion for all permutations of sources and use the highest value as the final output. The test verifies if the score is equal to or larger than the expected value.

### DIFF
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -40,15 +40,11 @@ _FILES = {
     "fr": "20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac",
     "it": "20170516-0900-PLENARY-16-it_20170516-18_56_31_1.flac",
 }
-_MIXTURE_FILES = {
-    "speech_separation": "mixture_3729-6852-0037_8463-287645-0000.wav",
-}
-_CLEAN_FILES = {
-    "speech_separation": [
-        "s1_3729-6852-0037_8463-287645-0000.wav",
-        "s2_3729-6852-0037_8463-287645-0000.wav",
-    ],
-}
+_MIXTURE_FILE = "mixture_3729-6852-0037_8463-287645-0000.wav"
+_CLEAN_FILES = [
+    "s1_3729-6852-0037_8463-287645-0000.wav",
+    "s2_3729-6852-0037_8463-287645-0000.wav",
+]
 
 
 @pytest.fixture
@@ -63,24 +59,16 @@ def sample_speech(tmp_path, lang):
 
 
 @pytest.fixture
-def mixture_source(tmp_path, task):
-    if task not in _MIXTURE_FILES:
-        raise NotImplementedError(f"Unexpected task: {task}")
-    path = tmp_path.parent / _MIXTURE_FILES[task]
-    if not path.exists():
-        torchaudio.utils.download_asset(f"test-assets/{_MIXTURE_FILES[task]}", path=path)
+def mixture_source():
+    path = torchaudio.utils.download_asset(f"test-assets/{_MIXTURE_FILE}")
     return path
 
 
 @pytest.fixture
-def clean_sources(tmp_path, task):
-    if task not in _CLEAN_FILES:
-        raise NotImplementedError(f"Unexpected task: {task}")
+def clean_sources():
     paths = []
-    for file in _CLEAN_FILES[task]:
-        path = tmp_path.parent / file
-        if not path.exists():
-            torchaudio.utils.download_asset(f"test-assets/{file}", path=path)
+    for file in _CLEAN_FILES:
+        path = torchaudio.utils.download_asset(f"test-assets/{file}")
         paths.append(path)
     return paths
 
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -40,8 +40,12 @@ _FILES = {
     "fr": "20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac",
     "it": "20170516-0900-PLENARY-16-it_20170516-18_56_31_1.flac",
 }
-_MIXTURE_FILENAME = "mix_3729-6852-0037_8463-287645-0000.wav"
-_EXPECTED_FILENAME = "expected_3729-6852-0037_8463-287645-0000.pt"
+_MIXTURE_FILES = {
+    "speech_separation": "mix_3729-6852-0037_8463-287645-0000.wav",
+}
+_EXPECTED_TENSORS = {
+    "speech_separation": "expected_3729-6852-0037_8463-287645-0000.pt",
+}
 
 
 @pytest.fixture
@@ -56,18 +60,22 @@ def sample_speech(tmp_path, lang):
 
 
 @pytest.fixture
-def mixture_speech(tmp_path):
-    path = tmp_path.parent / _MIXTURE_FILENAME
+def mixture_speech(tmp_path, task):
+    if task not in _MIXTURE_FILES:
+        raise NotImplementedError(f"Unexpected task: {task}")
+    path = tmp_path.parent / _MIXTURE_FILES[task]
     if not path.exists():
-        torchaudio.utils.download_asset(f"test-assets/{_MIXTURE_FILENAME}", path=path)
+        torchaudio.utils.download_asset(f"test-assets/{_MIXTURE_FILES[task]}", path=path)
     return path
 
 
 @pytest.fixture
-def expected_tensor(tmp_path):
-    path = tmp_path.parent / _EXPECTED_FILENAME
+def expected_tensor(tmp_path, task):
+    if task not in _EXPECTED_TENSORS:
+        raise NotImplementedError(f"Unexpected task: {task}")
+    path = tmp_path.parent / _EXPECTED_TENSORS[task]
     if not path.exists():
-        torchaudio.utils.download_asset(f"test-assets/{_EXPECTED_FILENAME}", path=path)
+        torchaudio.utils.download_asset(f"test-assets/{_EXPECTED_TENSORS[task]}", path=path)
     return path
 
 
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -1,3 +1,5 @@
+import os
+
 import pytest
 import torch
 import torchaudio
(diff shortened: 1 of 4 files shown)


test/integration_tests/conftest.py
```diff
@@ -40,6 +40,8 @@ def ctc_decoder():
     "fr": "20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac",
     "it": "20170516-0900-PLENARY-16-it_20170516-18_56_31_1.flac",
 }
+_MIXTURE_FILENAME = "mix_3729-6852-0037_8463-287645-0000.wav"
+_EXPECTED_FILENAME = "expected_3729-6852-0037_8463-287645-0000.pt"
```
mthrok, line 44: This approach is problematic for couple of reasons;
1. The data is not easily recoverable.
2. The internal change of PyTorch can break this test anytime, and it is almost impossible to track down the exact reason, nor we want to spend time on it.

These are the reasons why TTS integration test does not have this kind of assertion.
I recommend to find alternative (not equivalent) way to gauge the correctness of the model.

torchaudio/prototype/pipelines/source_separation_pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+        >>> # Apply source separation on mixture audio.
+        >>> for i, data in enumerate(dataset):
+        >>>     sample_rate, mixture, clean_sources = data
+        >>>     # Make sure the shape of input suits the model requirement.
+        >>>     mixture = mixture.reshape(1, 1, -1)
+        >>>     estimated_sources = separator(mixture)
+        >>>     score = si_snr_pit(estimated_sources, clean_sources) # for demonstration
+        >>>     print(f"Si-SNR score is : {score}.)
+        >>>     break
+        >>> 16.24
```
skim0514, line 56:
```suggestion
        >>> Si-SNR score is : 16.24.
```

torchaudio/prototype/pipelines/source_separation_pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+class _Separator(torch.nn.Module):
+    def __init__(self, model: torch.nn.Module):
+        super().__init__()
+        self.model = model
+
+    def forward(self, feature: torch.Tensor) -> torch.Tensor:
+        """Separates the input mixture speech into different sources.
+
+        Args:
+            input (torch.Tensor): input tensor with dimensions `[batch, time]`.
```
skim0514, line 20: Is it possible to generalize shape for future inputs that could include things like num_channels

torchaudio/prototype/pipelines/source_separation_pipeline.py
```diff
@@ -0,0 +1,93 @@
...
+from torchaudio.prototype.models import conv_tasnet_base
+
+
+class _Separator(torch.nn.Module):
+    def __init__(self, model: torch.nn.Module):
+        super().__init__()
+        self.model = model
+
+    def forward(self, feature: torch.Tensor) -> torch.Tensor:
+        """Separates the input mixture speech into different sources.
```
skim0514, line 17:
```suggestion
        """Separates the input mixture audio into different sources.
```

test/integration_tests/source_separation_pipeline_test.py
```diff
@@ -0,0 +1,42 @@
...
+        [CONVTASNET_BASE_LIBRI2MIX, "speech_separation"],
+    ],
+)
+def test_source_separation_models(bundle, task, mixture_speech, expected_tensor):
+    """Smoke test of source separation pipeline"""
+    separator = bundle.get_separator()
+    mixture_speech, sample_rate = torchaudio.load(mixture_speech)
+    expected_tensor = torch.load(expected_tensor)
+    mixture_speech = mixture_speech.reshape(1, 1, -1)
+    estimated_sources = separator(mixture_speech)
```
carolineechen, line 19: would it be sufficient/better to test that `expected_tensor` and `estimated_sources` are equal?

test/integration_tests/source_separation_pipeline_test.py
```diff
@@ -0,0 +1,42 @@
...
+    "bundle,task",
+    [
+        [CONVTASNET_BASE_LIBRI2MIX, "speech_separation"],
+    ],
+)
+def test_source_separation_models(bundle, task, mixture_speech, expected_tensor):
+    """Smoke test of source separation pipeline"""
+    separator = bundle.get_separator()
+    mixture_speech, sample_rate = torchaudio.load(mixture_speech)
+    expected_tensor = torch.load(expected_tensor)
```
mthrok, line 17: Please do not use serialized tensor, rather use regular audio for this

test/integration_tests/conftest.py
```diff
@@ -40,6 +40,12 @@ def ctc_decoder():
     "fr": "20121212-0900-PLENARY-5-fr_20121212-11_37_04_10.flac",
     "it": "20170516-0900-PLENARY-16-it_20170516-18_56_31_1.flac",
 }
+_MIXTURE_FILES = {
+    "speech_separation": "mix_3729-6852-0037_8463-287645-0000.wav",
+}
+_EXPECTED_TENSORS = {
+    "speech_separation": "expected_3729-6852-0037_8463-287645-0000.pt",
```
mthrok, line 47: Please replace with this regular audio format.

test/integration_tests/source_separation_pipeline_test.py
mthrok: @carolineechen It's not the randomness that is an issue, but the fact that future minor changes in PyTorch would slightly change the result from time to time.

Using reference from a past run and comparing against it is not maintainable. It will suddenly break due to some change in core that happens in deep down of some BLAS routine change.

test/integration_tests/conftest.py
nateanl: The reason why I used tensor before is the output waveform has large values, after saving it to audio file it causes clipping.

Let me try to save the normalized waveform to see if the peak frequency test can pass.

mthrok: In that case, I think it makes more sense to compare against the ground truth audio, and apply threshold to the SNR between original reference and result.

nateanl: So rather than uploading the estimated waveforms, we can upload the ground truth waveforms and compute the SNR scores then compare it with a given values? Makes sense to me.
Can we use third-party library for metric computation, for example, `mir_eval`?

test/integration_tests/source_separation_pipeline_test.py
nateanl: hmm actually when I compared the waveforms on CI, the current test already failed, but I guess the error is tolerable if we set up a threshold the test can pass.

test/integration_tests/conftest.py
```diff
@@ -53,6 +62,29 @@ def sample_speech(tmp_path, lang):
     return path
 
 
+@pytest.fixture
+def mixture_source(tmp_path, task):
```
mthrok, line 66: Please get rid of `task` parameterization as there is only one at the moment. We can add it when it becomes absolutely necessary.

test/integration_tests/conftest.py
```diff
@@ -53,6 +62,29 @@ def sample_speech(tmp_path, lang):
...
+
+@pytest.fixture
+def clean_sources(tmp_path, task):
+    if task not in _CLEAN_FILES:
+        raise NotImplementedError(f"Unexpected task: {task}")
+    paths = []
+    for file in _CLEAN_FILES[task]:
+        path = tmp_path.parent / file
+        if not path.exists():
+            torchaudio.utils.download_asset(f"test-assets/{file}", path=path)
```
mthrok, line 83: Since this is using `torchaudio.utils.download_asset`, `tmp_path` is not necessary. `download_asset` function will resolve the path for you.

test/integration_tests/source_separation_pipeline_test.py
```diff
@@ -0,0 +1,38 @@
...
+import torch
+import torchaudio
+from torchaudio.prototype.pipelines import CONVTASNET_BASE_LIBRI2MIX
+
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "..", "examples"))
+from source_separation.utils.metrics import PIT, sdr
+
+BUNDLE = CONVTASNET_BASE_LIBRI2MIX
+EXPECTED_SCORE = 8.1373
```
mthrok, line 13: Please put these in test function. they should not be module-level global.

torchaudio/prototype/pipelines/source_separation_pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+    _sample_rate: int
+
+    @property
+    def sample_rate(self) -> int:
+        """Sample rate (in cycles per second) of input waveforms.
+        :type: int
+        """
+        return self._sample_rate
+
+    def get_separator(self) -> torch.nn.Module:
```
mthrok, line 53: Can we call this `get_model`? `separator` does not sound common, and there is no other structure related to this bundle.

test/integration_tests/conftest.py
```diff
@@ -53,6 +58,21 @@ def sample_speech(tmp_path, lang):
     return path
 
 
+@pytest.fixture
+def mixture_source():
+    path = torchaudio.utils.download_asset(f"test-assets/{_MIXTURE_FILE}")
```
mthrok, line 63: The use of slash prevents this test from running in Windows. Can you use os.path or pathlib?

torchaudio/prototype/pipelines/source_separation_pipeline.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+CONVTASNET_BASE_LIBRI2MIX = SourceSeparationBundle(
+    _model_path="models/conv_tasnet_base_libri2mix.pt",
+    _model_factory_func=partial(conv_tasnet_base, num_sources=2),
+    _sample_rate=8000,
+)
+CONVTASNET_BASE_LIBRI2MIX.__doc__ = """Pre-trained ConvTasNet pipeline for source separation.
+    The underlying model is constructed by :py:func:`torchaudio.prototyoe.models.conv_tasnet_base`
+    and utilizes weights trained on Libri2Mix using training script ``lightning_train.py``
```
mthrok, line 69: Can you add citation to Libri2Mix dataset?

Issue #2359, opened 2022-05-03
Cache Resample kernel inside transform.PitchShift() 


xvdp:
### 🚀 The feature

Cache PitchShift Resample kernel to improve speed of this transform on second usage.

### Motivation, pitch

In porting some augmentation code from `librosa` to `torchaudio`, I noticed that `PitchShift()` and `Resample()` are slower than librosa in CPU.

In the case of `transforms.Resample()`, this changes on second run as the kernel is cached. But PitchShift does no such caching, calling instead the functional.
As to the first run. I didnt dig into librosa `core.resample()` to see why under similar conditions on first run resample is faster in one of the directions. Perhaps this is another opportunity, but besides this suggestio.

`transforms.PitchShift()` calls the `resample()` functional, rebuilding the kernel every time.
 
Alt .1 declare `class PitchShift(Resample):` and hardcode whatever doesn't require the full set of options.

Alt.2 partial code replication in `PitchShift`, cache `kernel` buffer, and call  ` _get_sinc_resample_kernel()` and ` _apply_sinc_resample_kernel()` instead of `resample()` inside the PitchShift functional code.

mthrok:
Hi @xvdp 

Thanks for the pitch. This is a good point and make sense. I think Alt.2 is somewhat better from maintaining perspective.

I will put this as help-wanted issue, as I do not have bandwidth to look into this right now. We welcome your contribution as well.

xvdp:
Thanks, I fixed it on my end but I have some unrelated issue on the install from source; So I cannot test it properly in its context, even if with a simple overriding the classes on _transform and functional it works. I presume you want a proper set of tests before doing pull request.

PR #2441, opened 2022-06-03
Modifying Pitchshift for faster resampling

skim0514:
Split existing Pitchshift into multiple helper functions in order to cache kernel and speed up overall process addressing #2359. 
Existing unit tests pass. 

edit: functional and transforms unit test pass. Adopted lazy initialization to avoid BC-breaking.

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -4,7 +4,7 @@ import io
 import math
 import warnings
 from collections.abc import Sequence
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, List
 
 import torch
 import torchaudio
@@ -1389,10 +1389,10 @@ def _get_sinc_resample_kernel(
     orig_freq: int,
     new_freq: int,
     gcd: int,
-    lowpass_filter_width: int,
-    rolloff: float,
-    resampling_method: str,
-    beta: Optional[float],
+    lowpass_filter_width: int = 6,
+    rolloff: float = 0.99,
+    resampling_method: str = "sinc_interpolation",
+    beta: Optional[float] = None,
     device: torch.device = torch.device("cpu"),
     dtype: Optional[torch.dtype] = None,
 ):
@@ -1635,6 +1635,39 @@ def pitch_shift(
     Returns:
         Tensor: The pitch-shifted audio waveform of shape `(..., time)`.
     """
+    waveform_stretch = _stretch_waveform(
+        waveform,
+        n_steps,
+        bins_per_octave,
+        n_fft,
+        win_length,
+        hop_length,
+        window,
+    )
+    rate = 2.0 ** (-float(n_steps) / bins_per_octave)
+    waveform_shift = resample(waveform_stretch, int(sample_rate / rate), sample_rate)
+
+    return _fix_waveform_shape(waveform_shift, waveform.size())
+
+
+def _stretch_waveform(
+    waveform: Tensor,
+    n_steps: int,
+    bins_per_octave: int = 12,
+    n_fft: int = 512,
+    win_length: Optional[int] = None,
+    hop_length: Optional[int] = None,
+    window: Optional[Tensor] = None,
+) -> Tensor:
+    """
+    Pitch shift helper function to preprocess and stretch waveform before resampling step.
+
+    Args:
+        See pitch_shift arg descriptions.
+
+    Returns:
+        Tensor: The preprocessed waveform stretched prior to resampling.
+    """
     if hop_length is None:
         hop_length = n_fft // 4
     if win_length is None:
@@ -1666,7 +1699,24 @@ def pitch_shift(
     waveform_stretch = torch.istft(
         spec_stretch, n_fft=n_fft, hop_length=hop_length, win_length=win_length, window=window, length=len_stretch
     )
-    waveform_shift = resample(waveform_stretch, int(sample_rate / rate), sample_rate)
+    return waveform_stretch
+
+
+def _fix_waveform_shape(
+    waveform_shift: Tensor,
+    shape: List[int],
+) -> Tensor:
+    """
+    PitchShift helper function to process after resampling step to fix the shape back.
+
+    Args:
+        waveform_shift(Tensor): The waveform after stretch and resample
+        shape (List[int]): The shape of initial waveform
+
+    Returns:
+        Tensor: The pitch-shifted audio waveform of shape `(..., time)`.
+    """
+    ori_len = shape[-1]
     shift_len = waveform_shift.size()[-1]
     if shift_len > ori_len:
         waveform_shift = waveform_shift[..., :ori_len]
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -6,10 +6,15 @@ from typing import Callable, Optional
 
 import torch
 from torch import Tensor
+from torch.nn.modules.lazy import LazyModuleMixin
+from torch.nn.parameter import UninitializedParameter
+
 from torchaudio import functional as F
 from torchaudio.functional.functional import (
     _apply_sinc_resample_kernel,
     _get_sinc_resample_kernel,
+    _stretch_waveform,
+    _fix_waveform_shape,
 )
 
 __all__ = []
@@ -1511,7 +1516,7 @@ class SpectralCentroid(torch.nn.Module):
         )
 
 
-class PitchShift(torch.nn.Module):
+class PitchShift(LazyModuleMixin, torch.nn.Module):
     r"""Shift the pitch of a waveform by ``n_steps`` steps.
 
     .. devices:: CPU CUDA
@@ -1537,6 +1542,9 @@ class PitchShift(torch.nn.Module):
     """
     __constants__ = ["sample_rate", "n_steps", "bins_per_octave", "n_fft", "win_length", "hop_length"]
 
+    kernel: UninitializedParameter
+    width: int
+
     def __init__(
         self,
         sample_rate: int,
@@ -1548,7 +1556,7 @@ class PitchShift(torch.nn.Module):
         window_fn: Callable[..., Tensor] = torch.hann_window,
         wkwargs: Optional[dict] = None,
     ) -> None:
-        super(PitchShift, self).__init__()
+        super().__init__()
         self.n_steps = n_steps
         self.bins_per_octave = bins_per_octave
         self.sample_rate = sample_rate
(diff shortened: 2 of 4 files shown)


torchaudio/functional/__init__.py
```diff
@@ -43,7 +43,7 @@
     mvdr_weights_rtf,
     mvdr_weights_souden,
     phase_vocoder,
-    pitch_shift,
+    # pitch_shift,
```
nateanl, line 46:
```suggestion
   pitch_shift,
```

torchaudio/functional/__init__.py
```diff
@@ -98,7 +98,7 @@
     "apply_codec",
     "resample",
     "edit_distance",
-    "pitch_shift",
+    # "pitch_shift",
```
nateanl, line 101:
```suggestion
    "pitch_shift",
```

torchaudio/functional/functional.py
```diff
@@ -1678,6 +1678,92 @@ def pitch_shift(
     return waveform_shift
 
 
+def _pitch_shift_preprocess(
```
carolineechen, line 1681: to avoid code duplication, you should use these helper functions in `pitch_shift` as well. this way, if we make a change to the implementation in the future, we do not need to remember to do it in two places

carolineechen: also we generally unit test correctness for the functional version and do not have corresponding tests for transforms when the transform simply calls on the functional. Here you change the transforms to use these helper functions and but not for functional, so we don't actually know that the behavior of transforms wasn't changed. we can be more confident on that if functional and transforms have the same implementation, except one computes the kernel and one uses a cached one.

torchaudio/functional/functional.py
```diff
@@ -1678,6 +1678,92 @@ def pitch_shift(
...
+    bins_per_octave: int = 12,
+    n_fft: int = 512,
+    win_length: Optional[int] = None,
+    hop_length: Optional[int] = None,
+    window: Optional[Tensor] = None,
+) -> Tensor:
+    """
+    transforms PitchShift helper function to preprocess before resampling step.
+
+    .. devices:: CPU CUDA
```
carolineechen, line 1693: no need to include devices and properties -- these are displayed for users to know what is supported, but these functions are not user facing

torchaudio/transforms/_transforms.py
```diff
@@ -1557,6 +1559,18 @@ def __init__(
         self.hop_length = hop_length if hop_length is not None else self.win_length // 4
         window = window_fn(self.win_length) if wkwargs is None else window_fn(self.win_length, **wkwargs)
         self.register_buffer("window", window)
+        self.rate = 2.0 ** (-float(n_steps) / bins_per_octave)
+        self.orig_freq = int(sample_rate / self.rate)
```
carolineechen, line 1563:
```suggestion
        rate = 2.0 ** (-float(n_steps) / bins_per_octave)
        self.orig_freq = int(sample_rate / rate)
```

torchaudio/transforms/_transforms.py
```diff
@@ -1557,6 +1559,18 @@ def __init__(
...
+        self.new_freq = sample_rate
```
carolineechen, line 1564: `self.sample_rate` is already defined above, so this is unnecessary

torchaudio/transforms/_transforms.py
```diff
@@ -1557,6 +1559,18 @@ def __init__(
...
+        self.gcd = math.gcd(int(self.orig_freq), int(self.new_freq))
+
+        if self.orig_freq != self.new_freq:
```
carolineechen, line 1567: this is outside the scope of this PR, but for someone to follow up on later -- could be good to integrate the `orig_freq == new_freq` case into the resampling helper methods to avoid needing this if statement in various code places

torchaudio/functional/functional.py
```diff
@@ -1635,6 +1635,38 @@ def pitch_shift(
...
+    waveform: Tensor,
+    n_steps: int,
+    bins_per_octave: int = 12,
+    n_fft: int = 512,
+    win_length: Optional[int] = None,
+    hop_length: Optional[int] = None,
+    window: Optional[Tensor] = None,
+) -> Tensor:
+    """
+    PitchShift helper function to preprocess before resampling step.
```
carolineechen, line 1662: nit:
```suggestion
    Pitch shift helper function to preprocess before resampling step.
```

torchaudio/functional/functional.py
```diff
@@ -1635,6 +1635,38 @@ def pitch_shift(
...
+
+    Args:
+        see pitch_shift arg descriptions.
```
carolineechen, line 1665:
```suggestion
        See pitch_shift arg descriptions.
```

torchaudio/functional/functional.py
```diff
@@ -1666,7 +1698,24 @@ def pitch_shift(
...
-    waveform_shift = resample(waveform_stretch, int(sample_rate / rate), sample_rate)
+    return waveform_stretch
+
+
+def _pitch_shift_postprocess(
+    waveform_shift: Tensor,
+    shape: List[int],
+) -> Tensor:
+    """
+        PitchShift helper function to process after resampling step.
```
carolineechen, line 1709:
```suggestion
Pitch shift helper function to process after resampling step.
```

torchaudio/transforms/_transforms.py
```diff
@@ -1566,10 +1579,11 @@ def forward(self, waveform: Tensor) -> Tensor:
         Returns:
             Tensor: The pitch-shifted audio of shape `(..., time)`.
         """
+        # retrieving shape
```
carolineechen, line 1582: I would remove the comment and move `shape = waveform.size()` to above the post process function, right before it is used

carolineechen (commented):
some minor suggestions

torchaudio/functional/functional.py
```diff
@@ -1635,6 +1635,38 @@ def pitch_shift(
     Returns:
         Tensor: The pitch-shifted audio waveform of shape `(..., time)`.
     """
+    waveform_stretch = _pitch_shift_preprocess(
+        waveform,
+        n_steps,
+        bins_per_octave,
+        n_fft, win_length,
```
nateanl, line 1642:
```suggestion
        n_fft,
        win_length,
```

carolineechen, line 1642:
```suggestion
        n_fft,
        win_length,
```

torchaudio/functional/functional.py
```diff
@@ -1635,6 +1635,38 @@ def pitch_shift(
...
+        hop_length,
+        window
```
carolineechen, line 1644:
```suggestion
        window,
```

torchaudio/transforms/_transforms.py
```diff
@@ -1566,10 +1579,8 @@ def forward(self, waveform: Tensor) -> Tensor:
         Returns:
             Tensor: The pitch-shifted audio of shape `(..., time)`.
         """
-
-        return F.pitch_shift(
+        stretch = _pitch_shift_preprocess(
```
nateanl, line 1582:
```suggestion
        waveform_stretch = _pitch_shift_preprocess(
```

torchaudio/transforms/_transforms.py
```diff
@@ -1578,6 +1589,16 @@ def forward(self, waveform: Tensor) -> Tensor:
             self.window,
         )
 
+        if self.orig_freq != self.sample_rate:
+            stretch = _apply_sinc_resample_kernel(stretch, self.orig_freq, self.sample_rate, self.gcd, self.kernel,
```
nateanl, line 1593:
```suggestion
            waveform_shift = _apply_sinc_resample_kernel(stretch, self.orig_freq, self.sample_rate, self.gcd, self.kernel,
```

torchaudio/transforms/_transforms.py
```diff
@@ -1578,6 +1589,16 @@ def forward(self, waveform: Tensor) -> Tensor:
...
+                                                  self.width)
+
+        shape = waveform.size()
+        return _pitch_shift_postprocess(
+            stretch,
```
nateanl, line 1598:
```suggestion
            waveform_shift,
```

torchaudio/functional/functional.py
```diff
@@ -1635,6 +1635,38 @@ def pitch_shift(
...
+    )
+    rate = 2.0 ** (-float(n_steps) / bins_per_octave)
+    waveform_shift = resample(waveform_stretch, int(sample_rate / rate), sample_rate)
+
+    return _pitch_shift_postprocess(waveform_shift, waveform.size())
+
+
+def _pitch_shift_preprocess(
```
nateanl, line 1652: I would call it `_stretch_waveform`, which is more informative.

torchaudio/functional/functional.py
```diff
@@ -1666,7 +1698,24 @@ def pitch_shift(
     waveform_stretch = torch.istft(
         spec_stretch, n_fft=n_fft, hop_length=hop_length, win_length=win_length, window=window, length=len_stretch
     )
-    waveform_shift = resample(waveform_stretch, int(sample_rate / rate), sample_rate)
+    return waveform_stretch
+
+
+def _pitch_shift_postprocess(
```
nateanl, line 1704: maybe call it `_fix_waveform_shape`. The basic function here is fix the length of waveform if it's longer than original or pad zeros if it's longer.

torchaudio/functional/functional.py
```diff
@@ -1666,7 +1698,24 @@ def pitch_shift(
...
+    """
+    PitchShift helper function to process after resampling step.
+
+    Args:
+        waveform_shift (Tensor): The input shifted waveform of shape `(..., time)`.
+        shape (List[int]): shape of initial waveform
+
+    Returns:
+        Tensor: The pitch-shifted audio waveform of shape `(..., time)`.
+    """
```
nateanl, line 1717: The description needs to be changed. It returns the waveform with `shape` dimensions.

torchaudio/functional/functional.py
mthrok: +1. In general, names like pre-process / post-process should be avoided for implementation in library for the sake of clarity, unless the it's on the code is on application layer or it's really well-established, universally applicable and not change over time.

carolineechen:
As discussed offline, the current implementation ends up being BC-breaking because of the dtype used for the resampling kernel computation (the transform kernel here is computed w/ float64 during initialization, but previously, it is computed w/ the same dtype as the waveform input, which may or may not be float64). We previously ran into this issue for resampling in #1514 and had a workaround in #1556. A better approach to avoid BC-breaking could be using lazy initialization. Let's continue discussion offline to see which solution to take to close the PR.

In the meantime, there are still some unaddressed naming/style comments and to move the shape param up as discussed.

torchaudio/transforms/_transforms.py
carolineechen: edit: waveform shape is temporarily modified in the preprocess function, so we need to move it back

nateanl (approved):

carolineechen (approved):
per offline discussion w/ @mthrok, we can avoid BC-breaking by adopting lazy module for the kernel initialization on top of these changes, which he has implemented and tested in #2468. this PR looks good to go, and we can merge lazy initialization after this has landed.

i have additionally done some benchmarking and can see performance gains with this caching

PR #2448, opened 2022-06-06
[DOC/CI] Store doc as tar archive

### LINKED ISSUE #2340: Store documentation archive in CI
Currently, `build_docs` CI jobs uploads the built documentation as artifacts.
At the time of release, we need to update the documentation page manually from these artifacts.

CircleCI does not have a simple API to bulk download artifacts, so one needs to compose some API calls, which is tedious. We should simply add zipped documentation to CI artifact.

mthrok:
At the time of release, we need to download doc built by CI.
CircleCI does not have feature to download multiple files.

This commit add the archive of built documentations as
CI artifact so that the whole documentation can be downloaded
at once.

Resolves #2340

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -827,6 +827,9 @@ jobs:
             source /usr/local/etc/profile.d/conda.sh && conda activate python${PYTHON_VERSION}
             cd docs
             make 'SPHINXOPTS=-W' html
+            cd build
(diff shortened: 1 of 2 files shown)


nateanl (approved):
LGTM

hwangjeff (approved):

PR #2449, opened 2022-06-06
Add note about `normalize` argument

### LINKED ISSUE #2253: torchaudio.backend.sox_io_backend does not normalize
### 🐛 Describe the bug

Hi,

I am experiencing a bug when loading a WAV file using _sox_io_ backend where the loaded file is not normalized (i.e. fit in the range of ``[-1.0, 1.0]``).

Torchaudio version: 0.10.2

Below is the code and prints if I load the file without normalization (original file format int32):
image

And here's with normalization:
image

The audio is converted to float32 but it is NOT normalized. Unless it is a language problem and you used "normalize" instead of "convert", but I doubt since in this context _to normalize_ should mean `audio_normalized = audio / max(abs(audio))`.

Thanks

### Versions

PyTorch version: 1.10.2
Is debug build: False

OS: macOS 11.6 (x86_64)
Clang version: 13.0.0 (clang-1300.0.29.3)
CMake version: Could not collect
Libc version: N/A

Python version: 3.9.10 (main, Jan 15 2022, 11:48:00)  [Clang 13.0.0 (clang-1300.0.29.3)] (64-bit runtime)
Python platform: macOS-11.6-x86_64-i386-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.21.5
[pip3] torch==1.10.2
[pip3] torchaudio==0.10.2
[conda] Could not collect

mthrok:
`load` function has `normalize` argument, which converts the native
sample type to `torch.float32`.

This argument is confusing for audio practitioners as it seems
to perform [volume normalization](https://en.wikipedia.org/wiki/Audio_normalization).

See https://github.com/pytorch/audio/issues/2253

Due to the BC-breaking concern, we cannot easily change the argument name.
This commit adds warnings to documentations.

Fix #2253

### DIFF
--- a/torchaudio/backend/soundfile_backend.py
+++ b/torchaudio/backend/soundfile_backend.py
@@ -131,12 +131,6 @@ def load(
 ) -> Tuple[torch.Tensor, int]:
     """Load audio data from file.
 
-    .. warning::
-
-       `normalize` argument does not perform normalization.
-       It only converts the sample type to `torch.float32` from the native sample
-       type.
-
     Note:
         The formats this function can handle depend on the soundfile installation.
         This function is tested on the following formats;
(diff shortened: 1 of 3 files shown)


nateanl:
My understanding is that the `normalize` argument only affects `WAV` format, and is only used when dtype is `int`? We can also mention that in the warning (though it's already in the documentation), so that users who load `FLAC` or `OGG` audios know turning `normalize` on or off doesn't change the output.

torchaudio/backend/soundfile_backend.py
```diff
@@ -131,6 +131,12 @@ def load(
 ) -> Tuple[torch.Tensor, int]:
     """Load audio data from file.
 
+    .. warning::
+
+       `normalize` argument, despite the name, does not perform normalization.
```
carolineechen, line 136: nit: to stay consistent with the other warnings

```suggestion
       `normalize` argument does not perform normalization.
```

carolineechen (approved):

torchaudio/backend/sox_io_backend.py
```diff
@@ -181,7 +188,7 @@ def load(
 
     Returns:
         (torch.Tensor, int): Resulting Tensor and sample rate.
-            If the input file has integer wav format and normalization is off, then it has
+            If the input file has integer wav format and`` normalize=False``, then it has
```
carolineechen, line 191:
```suggestion
            If the input file has integer wav format and ``normalize=False``, then it has
```

carolineechen (approved):

nateanl (approved):

PR #2452, opened 2022-06-07
[release 0.12] Set release version and base PyTorch version

hwangjeff:
Sets release version to 0.12 and base PyTorch version to 1.12 in CircleCI config files in preparation for release.


mthrok (approved):
Looks good. The addition of "designate_upload_channel" should be back ported to main branch, so that we do not need to repeat it for future release.

PR #2456, opened 2022-06-07
Exclude prototype tests in release CI

hwangjeff:


mthrok:
Prototypes are not supposed to be part of release, so can you simply remove prototype source code `torchaudio.prototypes` and test code `test/torchaudio_unittest/prototypes`?

PR #2458, opened 2022-06-07
Quesst14 return type change

skim0514:
Fixing return types for quesst14


carolineechen (approved):
thanks for catching this, lg

PR #2467, opened 2022-06-09
Add utility function to fetch FFmpeg library versions

mthrok:
Follow-up of #2464. Add utility function to fetch the versions of FFmpeg.


carolineechen (approved):

mthrok:
I realized that version are defined as macros, meaning that they report the version numbers used for compiling, not the runtime versions. It might be more confusing for users of binary distributions because they can pick different versions. So I am going to close this.

PR #2477, opened 2022-06-13
Add special case for mp3 filelike object

mthrok:
Loading and querying file-like object is not possible to use the fallback
mechanism introduced in https://github.com/pytorch/audio/pull/2419 because file-like objects are not seekable.

This commit add special case handling to mp3.

For filelike object mp3 input, it was required to pass `format="mp3"`
because libsox did not auto detect the format.

With the transition of mp3 handling from libsox to ffmpeg, the logic
is to let the ffmpeg handle it without waiting for libsox to fail,
if the `format="mp3"`

### DIFF
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -84,6 +84,15 @@ def info(
     """
     if not torch.jit.is_scripting():
         if hasattr(filepath, "read"):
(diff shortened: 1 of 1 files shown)


carolineechen (approved):

mthrok:
Landing this directly on `release/0.12` as fbcode is having issue and won't let land on master. Will follow-up with main branch.

PR #2480, opened 2022-06-14
Add fluent speech commands

carolineechen:


docs/source/refs.bib
```diff
@@ -359,3 +359,16 @@ @misc{YesNo
   title="YesNo",
   url="http://www.openslr.org/1/"
 }
+@inproceedings{fluent,
+  author    = {Loren Lugosch and
+               Mirco Ravanelli and
+               Patrick Ignoto and
+               Vikrant Singh Tomar and
+               Yoshua Bengio},
```
nateanl, line 367: Making them into one line looks better?

docs/source/refs.bib
```diff
@@ -359,3 +359,16 @@ @misc{YesNo
...
+  editor    = {Gernot Kubin and
+               Zdravko Kacic},
```
nateanl, line 369: same here

torchaudio/datasets/fluentcommands.py
```diff
@@ -0,0 +1,56 @@
...
+
+FOLDER_IN_ARCHIVE = "fluent_speech_commands_dataset"
+
+
+class FluentSpeechCommands(Dataset):
+    """Create *Fluent Speech Commands* [:footcite:`fluent`] Dataset
+
+    Args:
+        root (str of Path): Path to the directory where the dataset is found.
+        subset (str): subset of the dataset to use. Options: ["train", "valid", "test"].
```
nateanl, line 17:
```suggestion
        subset (str, optional): subset of the dataset to use. Options: [`"train"`, `"valid"`, `"test"`]. (Default: `"train"`)
```

torchaudio/datasets/fluentcommands.py
```diff
@@ -0,0 +1,56 @@
...
+            (Default: ``"train"``)
+    """
+
+    def __init__(self, root: Union[str, Path], subset: str = "train"):
+        assert subset in ["train", "valid", "test"]
```
nateanl, line 22: Add an exception message after assert makes clearer to users.

test/torchaudio_unittest/datasets/fluentcommands_test.py
```diff
@@ -0,0 +1,145 @@
...
+
+from torchaudio.datasets import fluentcommands
+from torchaudio_unittest.common_utils import (
+    get_whitenoise,
+    save_wav,
+    TempDirMixin,
+    TorchaudioTestCase,
+)
+
+HEADER = ["", "path", "speakerId", "transcription", "action", "object", "location"]
```
nateanl, line 15: Is the `speakerId` from original csv? Or can we change it to
```suggestion
HEADER = ["", "path", "speaker_id", "transcription", "action", "object", "location"]
```

torchaudio/datasets/fluentcommands.py
```diff
@@ -0,0 +1,56 @@
+import csv
+import os
+from pathlib import Path
+from typing import Union
+
+import torchaudio
+from torch.utils.data import Dataset
+
+FOLDER_IN_ARCHIVE = "fluent_speech_commands_dataset"
```
nateanl, line 9: shall we add prefix underscore?
```suggestion
_FOLDER_IN_ARCHIVE = "fluent_speech_commands_dataset"
```

nateanl (approved):
LGTM. Just some nits to address.

test/torchaudio_unittest/datasets/fluentcommands_test.py
carolineechen: yea it is from the original csv

torchaudio/datasets/fluentcommands.py
```diff
@@ -0,0 +1,56 @@
+import csv
+import os
+from pathlib import Path
+from typing import Union
+
+import torchaudio
+from torch.utils.data import Dataset
+
+_FOLDER_IN_ARCHIVE = "fluent_speech_commands_dataset"
```
mthrok, line 9: Why should this be global variable? It is only used once, and it is not expected to be changed.

PR #2484, opened 2022-06-14
Create musdb handler and tests

skim0514:
Create dataset handler and tests for new dataset. Manually tested and unit tested to test validity. Pre-commit ran for style checks.

### DIFF
--- a/test/torchaudio_unittest/datasets/musdb_hq_test.py
+++ b/test/torchaudio_unittest/datasets/musdb_hq_test.py
@@ -1,0 +1,156 @@
+import os
+
+import torch
+from parameterized import parameterized
+from torchaudio.datasets import musdb_hq
+from torchaudio_unittest.common_utils import (
+    get_whitenoise,
+    save_wav,
+    TempDirMixin,
+    TorchaudioTestCase,
+)
+
+_SOURCE_SETS = [
+    (["bass", "drums", "other", "vocals"],),
+    (["bass", "drums", "other"],),
+    (["bass", "drums", "vocals"],),
+    (["bass", "vocals", "other"],),
+    (["vocals", "drums", "other"],),
+    (["mixture"],),
+]
+
+seed_dict = {
+    "bass": 0,
+    "drums": 1,
+    "other": 2,
+    "mixture": 3,
+    "vocals": 4,
+}
+
+EXT = ".wav"
+
+_VALIDATION_SET = ["0", "1", "2"]
+
+
+def _save_sample(dataset_dir, folder, song, source, sample_rate, seed):
+    # create and save audio samples to corresponding files
+    path = os.path.join(dataset_dir, folder)
+    os.makedirs(path, exist_ok=True)
+    song_path = os.path.join(path, str(song))
+    os.makedirs(song_path, exist_ok=True)
+    source_path = os.path.join(song_path, f"{source}{EXT}")
+
+    data = get_whitenoise(
+        sample_rate=sample_rate,
+        duration=5,
+        n_channels=2,
+        seed=seed,
+    )
+    save_wav(source_path, data, sample_rate)
+
+    sample = (data, sample_rate, song)
+
+    return sample
+
+
+def _get_mocked_samples(dataset_dir, sample_rate):
+    sample_count = 5
+
+    all_samples = {"train": {}, "test": {}}
+
+    folders = ["train", "test"]
+    sources = ["bass", "drums", "other", "mixture", "vocals"]
+
+    curr_idx = 0
+    for folder in folders:
+        for i in range(sample_count):
+            sample_list = []
+            for source in sources:
+                sample = _save_sample(dataset_dir, folder, str(curr_idx), source, sample_rate, seed_dict.get(source))
+                sample_list.append(sample)
+            all_samples[folder][str(curr_idx)] = sample_list
+            curr_idx += 1
+
+    return all_samples
+
+
+def get_mock_dataset(dataset_dir):
+    """
+    dataset_dir: directory to the mocked dataset
+    """
+    os.makedirs(dataset_dir, exist_ok=True)
+    sample_rate = 44100
+
+    return _get_mocked_samples(dataset_dir, sample_rate)
+
+
+class TestMusDB_HQ(TempDirMixin, TorchaudioTestCase):
+    root_dir = None
+    backend = "default"
+
+    train_all_samples = {}
+    train_only_samples = {}
+    validation_samples = {}
+    test_samples = {}
+
+    @classmethod
+    def setUpClass(cls):
+        cls.root_dir = cls.get_base_temp_dir()
+        dataset_dir = os.path.join(cls.root_dir, "musdb18hq")
+        full_dataset = get_mock_dataset(dataset_dir)
+        cls.train_all_samples = full_dataset["train"]
+        cls.test_samples = full_dataset["test"]
+        for key in cls.train_all_samples:
+            if key in _VALIDATION_SET:
+                cls.validation_samples[key] = cls.train_all_samples[key]
+            else:
+                cls.train_only_samples[key] = cls.train_all_samples[key]
+
+    def _test_musdb_hq(self, dataset, data_samples, sources):
+        num_samples = 0
+        for i, (data, sample_rate, name) in enumerate(dataset):
+            self.assertEqual(data, self.extractSources(data_samples[name], sources))
+            assert sample_rate == data_samples[name][0][1]
+            assert name == data_samples[name][0][2]
+            num_samples += 1
+
+        assert num_samples == len(data_samples)
+
+    @parameterized.expand(_SOURCE_SETS)
+    def testMusDBSources_train_all(self, sources):
+        dataset = musdb_hq.MUSDB_HQ(self.root_dir, sources=sources, subset="train")
+        self._test_musdb_hq(dataset, self.train_all_samples, sources)
+
+    @parameterized.expand(_SOURCE_SETS)
+    def testMusDBSources_train_with_validation(self, sources):
+        dataset = musdb_hq.MUSDB_HQ(
+            self.root_dir,
+            sources=sources,
+            subset="train",
+            split="train",
+            validation=_VALIDATION_SET,
+        )
+        self._test_musdb_hq(dataset, self.train_only_samples, sources)
+
+    @parameterized.expand(_SOURCE_SETS)
+    def testMusDBSources_validation(self, sources):
+        dataset = musdb_hq.MUSDB_HQ(
+            self.root_dir,
+            sources=sources,
+            subset="train",
+            split="validation",
+            validation=_VALIDATION_SET,
+        )
+        self._test_musdb_hq(dataset, self.validation_samples, sources)
+
+    @parameterized.expand(_SOURCE_SETS)
+    def testMusDBSources_test(self, sources):
+        dataset = musdb_hq.MUSDB_HQ(
+            self.root_dir,
+            sources=sources,
+            subset="test",
+        )
+        self._test_musdb_hq(dataset, self.test_samples, sources)
+
+    def extractSources(self, samples, sources):
+        return torch.stack([samples[seed_dict[source]][0] for source in sources])
--- a/test/torchaudio_unittest/datasets/musdb_hq_test.py
+++ b/test/torchaudio_unittest/datasets/musdb_hq_test.py
@@ -3,6 +3,7 @@ import os
 import torch
 from parameterized import parameterized
 from torchaudio.datasets import musdb_hq
+from torchaudio.datasets.musdb_hq import _VALIDATION_SET
 from torchaudio_unittest.common_utils import (
     get_whitenoise,
     save_wav,
@@ -11,6 +12,7 @@ from torchaudio_unittest.common_utils import (
 )
 
 _SOURCE_SETS = [
+    (None,),
     (["bass", "drums", "other", "vocals"],),
     (["bass", "drums", "other"],),
     (["bass", "drums", "vocals"],),
@@ -18,7 +20,6 @@ _SOURCE_SETS = [
     (["vocals", "drums", "other"],),
     (["mixture"],),
 ]
-
 seed_dict = {
     "bass": 0,
     "drums": 1,
@@ -26,10 +27,7 @@ seed_dict = {
     "mixture": 3,
     "vocals": 4,
 }
-
 EXT = ".wav"
-
-_VALIDATION_SET = ["0", "1", "2"]
 
 
 def _save_sample(dataset_dir, folder, song, source, sample_rate, seed):
@@ -48,7 +46,7 @@ def _save_sample(dataset_dir, folder, song, source, sample_rate, seed):
     )
     save_wav(source_path, data, sample_rate)
 
-    sample = (data, sample_rate, song)
+    sample = (data, sample_rate, 5 * sample_rate, song)
 
     return sample
 
@@ -63,13 +61,20 @@ def _get_mocked_samples(dataset_dir, sample_rate):
 
     curr_idx = 0
     for folder in folders:
-        for i in range(sample_count):
+        for _ in range(sample_count):
             sample_list = []
             for source in sources:
                 sample = _save_sample(dataset_dir, folder, str(curr_idx), source, sample_rate, seed_dict.get(source))
                 sample_list.append(sample)
             all_samples[folder][str(curr_idx)] = sample_list
             curr_idx += 1
+        if folder == "train":
+            for name in _VALIDATION_SET:
+                sample_list = []
+                for source in sources:
+                    sample = _save_sample(dataset_dir, folder, name, source, sample_rate, seed_dict.get(source))
+                    sample_list.append(sample)
+                all_samples[folder][name] = sample_list
 
     return all_samples
 
@@ -108,10 +113,11 @@ class TestMusDB_HQ(TempDirMixin, TorchaudioTestCase):
 
     def _test_musdb_hq(self, dataset, data_samples, sources):
         num_samples = 0
-        for i, (data, sample_rate, name) in enumerate(dataset):
+        for _, (data, sample_rate, num_frames, name) in enumerate(dataset):
             self.assertEqual(data, self.extractSources(data_samples[name], sources))
             assert sample_rate == data_samples[name][0][1]
-            assert name == data_samples[name][0][2]
+            assert num_frames == data_samples[name][0][2]
+            assert name == data_samples[name][0][3]
             num_samples += 1
 
         assert num_samples == len(data_samples)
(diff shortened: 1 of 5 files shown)


torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+from pathlib import Path
+from typing import Tuple, Union, List, Optional
+
+import torch
+import torchaudio
+from torch import Tensor
+from torch.utils.data import Dataset
+from torchaudio import transforms as T
+
+URL = "https://zenodo.org/record/3338373/files/musdb18hq.zip"
```
mthrok, line 11:
```suggestion
_URL = "https://zenodo.org/record/3338373/files/musdb18hq.zip"
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+EXT = ".wav"
```
mthrok, line 12:
```suggestion
_EXT = ".wav"
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+                )
+            wavs.append(wav)
+
+        stacked = torch.stack(wavs)
+        transform = T.Resample(sample_rate, self.user_samplerate)
+        waveform = transform(stacked)
+
+        return waveform, self.user_samplerate, name
+
+    def metadata_build(self):
```
mthrok, line 68: If this is not a public API, then please prefix the name with underscore. If it is public API, please add docstring.

Also the name of the method is not intuitively clear to me. Assuming it does not build metadata, is it about metadata of how dataset is built?

mthrok (approved):

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+            if sample_rate is None:
+                sample_rate = new_samplerate
+            elif sample_rate != new_samplerate:
+                raise ValueError(
+                    f"Invalid sample rate for file {file}: " f"expecting {sample_rate} but got {new_samplerate}."
+                )
+            wavs.append(wav)
+
+        stacked = torch.stack(wavs)
+        transform = T.Resample(sample_rate, self.user_samplerate)
```
mthrok, line 63: This constructs resampling kernel every time a sample is fetched from dataset.
Either construct the transform in constructor and cache, or let client code handle the resampling.

mthrok:
accidental approve. please ignore.

torchaudio/datasets/musdb_hq.py
skim0514: Sorry, I kept the old name from the other dataset class, will change

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+_EXT = ".wav"
+
+
+class MUSDB_HQ(Dataset):
+    def __init__(
+        self,
+        root: Union[str, Path],
+        sources: List[str],
+        subset: str,
+        samplerate: int = 44100,
```
nateanl, line 21:
```suggestion
        sample_rate: int = 44100,
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+
+_URL = "https://zenodo.org/record/3338373/files/musdb18hq.zip"
+_EXT = ".wav"
+
+
+class MUSDB_HQ(Dataset):
+    def __init__(
+        self,
+        root: Union[str, Path],
+        sources: List[str],
```
nateanl, line 19: It' not clear what kind of sources are required for this dataset. Could you give default values, for example, `_SOURCES = ["bass", "drums", "other", "mixture", "vocals"]`?

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+    def _get_file(self, name, source):
+        return Path(self._path) / name / f"{source}{_EXT}"
+
+    def _load_sample(self, n: int) -> Tuple[torch.Tensor, int, str]:
+        name = self.names[n]
+        wavs = []
+        sample_rate = None
+        for source in self.sources:
+            file = self._get_file(name, source)
+            wav, new_samplerate = torchaudio.load(str(file))
```
nateanl, line 53: Since `sample_rate` is in the argument of initialization, we can name it `sample_rate_file` here and assert it equal to `sellf.sample_rate`.

faroit:
@skim0514 great to have this in torchaudio! @nateanl @mthrok if you want me to review this, let me know!

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+                continue
+            names.append(name)
+        return names
+
+    def __getitem__(self, n: int) -> Tuple[torch.Tensor, int, str]:
+        """Load the n-th sample from the dataset.
+        Args:
+            n (int): The index of the sample to be loaded
+        Returns:
+            (Tensor, int, str): ``(waveform, sample_rate, file_name)``
```
carolineechen, line 88:
```suggestion
            (Tensor, int, str): ``(waveform, sample_rate, track_name)``
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+import torchaudio
+from torch import Tensor
+from torch.utils.data import Dataset
+from torchaudio import transforms as T
+
+_URL = "https://zenodo.org/record/3338373/files/musdb18hq.zip"
+_EXT = ".wav"
+
+
+class MUSDB_HQ(Dataset):
```
carolineechen, line 15: can you add docstrings here for the input args?

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+        assert subset in ["test", "train"], "`subset` must be one of ['test', 'train']"
+        assert self.split is None or self.split in [
+            "train",
+            "validation",
+        ], "`split` must be one of ['train', 'validation']"
+        assert (
+            self.split is None and self.validation is None or self.split and self.validation
+        ), "`split` and `validation` must both be None or not be None"
+        self._path = os.path.join(root, basename)
+        self._path = os.path.join(self._path, subset)
```
carolineechen, line 41:
```suggestion
        self._path = os.path.join(root, basename, subset)
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+        self.names = self._collect_songs()
+
+    def _get_file(self, name, source):
+        return Path(self._path) / name / f"{source}{_EXT}"
```
carolineechen, line 45: is `Path` here necessary?

torchaudio/datasets/musdb_hq.py
carolineechen: I think the point of this is to check that all sample rates across the sources within a track are the same, which may not be the same as the desired output sample rate (which is what I imagine the `sample_rate` in init args represents). In that case, I don't actually think we should keep `sample_rate` as an input arg (user can perform resampling on their own using transforms), and this function should just return the original wav, sample_rate, and track_name.

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _load_sample(self, n: int) -> Tuple[torch.Tensor, int, str]:
+        name = self.names[n]
+        wavs = []
+        sample_rate = None
+        for source in self.sources:
+            file = self._get_file(name, source)
+            wav, new_samplerate = torchaudio.load(str(file))
+            if sample_rate is None:
```
carolineechen, line 54: @nateanl the version in demucs also checked for track_length when loading metadata, do you think this is something we should include as well in the return?

nateanl: I thin so. And to make it generalized to all datasets, I have discussed with @leo19941227 and @xiaohui-zhang about supporting `skip_audio` argument in the dataset, which allows users to fetch the metadata (audio length, text transcription, etc) with faster speed.

@leo19941227 Can you open an issue about how you want to fetch the metadata, then we can share the discussion and code design there? Thanks!

mthrok:
> @skim0514 great to have this in torchaudio! @nateanl @mthrok if you want me to review this, let me know!

Sure, please do. We appreciate your perspective as a practitioner.

torchaudio/datasets/musdb_hq.py
nateanl: I see. I prefer leaving the resample function in users' code and return the original sample rate of the dataset. Considering DataLoader fetches data happens in CPU, users may prefer resampling waveforms in collator function which can be in GPU and much faster.
If we make sure all waveforms have the same sample_rate, we can remove the assertion here and return the sample rate directly.

torchaudio/datasets/musdb_hq.py
leo19941227: Oh sure! No problem, will do that soon.

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,103 @@
...
+        subset (str): Subset of the dataset to use. Options: [``"train"``, ``"test"``].
+        sources (List[str] or None, optional): Sources extract data from.
+            Options: [``bass``, ``drums``, ``other``, ``mixture``, ``vocals``].
+            If ``None``, dataset consists of tracks except mixture.
+            (default: ``"[``bass``, ``drums``, ``other``, ``vocals``]"``)
+        split (str, optional): Whether to split training set into train and validation set.
+            If None, no splitting occurs. If ``train`` or ``validation``, returns respective set.
+            (default: ``None``)
+        validation (List[str] or None, optional): Tracks which are included in validation set.
+            If ``None``, dataset consists of no tracks in validation set.
```
carolineechen, line 28: I think this is clearer:

```suggestion
        validation (List[str] or None, optional): Tracks included in the validation set, to be used for splitting.
            List must be provided if `split` is provided, or `None` if `split` is not provided.
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,103 @@
...
+        basename = os.path.basename(_URL)
+        basename = basename.rsplit(".", 2)[0]
+        assert subset in ["test", "train"], "`subset` must be one of ['test', 'train']"
+        assert self.split is None or self.split in [
+            "train",
+            "validation",
+        ], "`split` must be one of ['train', 'validation']"
+        assert (
+            self.split is None and self.validation is None or self.split and self.validation
+        ), "`split` and `validation` must both be None or not be None"
```
carolineechen, line 53:
```suggestion
        ), "`split` and `validation` must both be provided, or both be None"
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def _load_sample(self, n: int) -> Tuple[torch.Tensor, int, int, str]:
+        name = self.names[n]
+        wavs = []
+        sample_rate = None
+        num_frames = None
+        for source in self.sources:
+            track = self._get_track(name, source)
+            wav, _ = torchaudio.load(str(track))
+            if sample_rate is None:
```
carolineechen, line 68: my suggestion, since `sample_rate` should always match `_SAMPLE_RATE`, would be to do something along the lines of

```
    wav, sr = torchaudio.load(str(track))
    assert sr == _SAMPLE_RATE, f"expected sample rate {_SAMPLE_RATE}, but got {sr}"
    if num_frames is None:
         num_frames = info.num_frames
    else:
         assert info.num_frames == num_frames, "num_frames do not match across sources"
    ...
    return stacked, _SAMPLE_RATE, num_frames, name
```

carolineechen (commented):
I think the dataset implementation looks pretty good -- @nateanl @faroit any additional comments?

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,104 @@
...
+        path = Path(self._path)
+        names = []
+        for root, folders, files in os.walk(path, followlinks=True):
+            root = Path(root)
+            if root.name.startswith(".") or folders or root == path:
+                continue
+            name = str(root.relative_to(path))
+            if self.split and name in self.validation:
+                continue
+            names.append(name)
```
adefossez, line 91: adding sort() would make it deterministic

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,104 @@
...
+            (default: ``None``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        subset: str,
+        sources: List[str] = ["bass", "drums", "other", "vocals"],
+        split: Optional[str] = None,
+        validation: Optional[List[str]] = None,
```
adefossez, line 38: I dont think this should be a param, as this is hard to use without detailed knowledge of the dataset. 

One idea: have subset `train` `valid`, `full_train` or `test` and if it is `valid` or `train`, use a hardcoded list of the tracks that I sent. This would make the API mucher simpler.

If a someone still wants a custom split, they can find the indexes in `self.names` and build  a torch data `Subset` with those, but my take is that 99% of the users will use the standard splits.

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,104 @@
...
+        return Path(self._path) / name / f"{source}{_EXT}"
+
+    def _load_sample(self, n: int) -> Tuple[torch.Tensor, int, int, str]:
+        name = self.names[n]
+        wavs = []
+        num_frames = None
+        for source in self.sources:
+            track = self._get_track(name, source)
+            wav, sr = torchaudio.load(str(track))
+            info = torchaudio.info(str(track))
```
adefossez, line 67: is that required as `info.num_frames` is always `wav.shape[-1]` ?

torchaudio/datasets/musdb_hq.py
carolineechen: is there some reference that shows that the default valid set [here](https://github.com/sigsep/sigsep-mus-db/blob/master/musdb/configs/mus.yaml#L44) is the standard used for musdb-hq? since the paper only splits it into train/test, having `train` refer to a subset of the train can come up to confusing as me, which is why we created a further split param (which is also what is done in the repo linked above). I do see how it can be annoying to enter the validation list so maybe we could use a better default?

@faroit any thoughts on this?

carolineechen (commented):
some nits for docs

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class MUSDB_HQ(Dataset):
+    """Create *MUSDB_HQ Dataset* [:footcite:`MUSDB18HQ`] Dataset
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        subset (str): Subset of the dataset to use. Options: [``"train"``, ``"test"``].
+        sources (List[str] or None, optional): Sources extract data from.
+            Options: [``bass``, ``drums``, ``other``, ``mixture``, ``vocals``].
```
carolineechen, line 21: @nateanl wdyt of this wording, any additional suggestion?

```suggestion
        sources (List[str] or None, optional): Sources to extract data from.
            List can contain the following options: [``"bass"``, ``"drums"``, ``"other"``, ``"mixture"``, ``"vocals"``].
```

nateanl: ohh I understand. The sources here mean it can be any combination of these options? 
For example:
["bass", "drums"] is one option, ["mixture", "vocals"] is also an option?

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,119 @@
...
+            names.append(name)
+        names.sort()
+        return names
+
+    def __getitem__(self, n: int) -> Tuple[torch.Tensor, int, int, str]:
+        """Load the n-th sample from the dataset.
+        Args:
+            n (int): The index of the sample to be loaded
+        Returns:
+            (Tensor, int, str): ``(waveform, sample_rate, num_frames, track_name)``
```
carolineechen, line 114:
```suggestion
            (Tensor, int, int, str): ``(waveform, sample_rate, num_frames, track_name)``
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        subset: str,
+        sources: List[str] = ["bass", "drums", "other", "vocals"],
+        split: Optional[str] = None,
+        validation: Optional[List[str]] = None,
+        download: bool = False,
```
carolineechen, line 42: can you add docstring for this as well?

carolineechen (approved):
dataset impl looks good and we can merge this version after addressing the last two doc fixes I commented. let's continue discussion on handling train/valid set and follow up on that in a separate PR

torchaudio/datasets/musdb_hq.py
adefossez: The repo linked is the official musdb repository (see https://sigsep.github.io/datasets/musdb.html#tools). However I agree with your point that renaming the train set might be confusing. Then maybe the best is just to follow the musdb API with the extra parameter. The list of validation tracks given in the link should be automatically used though.

test/torchaudio_unittest/datasets/musdb_hq_test.py
```diff
@@ -0,0 +1,164 @@
...
+    for folder in folders:
+        for _ in range(sample_count):
+            sample_list = []
+            for source in sources:
+                sample = _save_sample(dataset_dir, folder, str(curr_idx), source, sample_rate, seed_dict.get(source))
+                sample_list.append(sample)
+            all_samples[folder][str(curr_idx)] = sample_list
+            curr_idx += 1
+        if folder == "train":
+            _VALIDATION_SET.sort()
```
carolineechen, line 74: global variables shouldn't be modified. also this line isn't necessary here since in the dataset, the names are only sorted if the split isn't the valid set

carolineechen (approved):

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,129 @@
...
+        for root, folders, _ in os.walk(path, followlinks=True):
+            root = Path(root)
+            if root.name.startswith(".") or folders or root == path:
+                continue
+            name = str(root.relative_to(path))
+            if self.split and name in _VALIDATION_SET:
+                continue
+            names.append(name)
+        names.sort()
+        return names
```
nateanl, line 117: The two lines can be simplified to
```suggestion
        return sorted(names)
```

torchaudio/datasets/musdb_hq.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __getitem__(self, n: int) -> Tuple[torch.Tensor, int, int, str]:
+        """Load the n-th sample from the dataset.
+        Args:
+            n (int): The index of the sample to be loaded
+        Returns:
+            (Tensor, int, int, str): ``(waveform, sample_rate, num_frames, track_name)``
```
nateanl, line 124:
```suggestion
            (Tensor, int, int, str): ``(waveforms, sample_rate, num_frames, track_name)``
```

PR #2485, opened 2022-06-14
Enable ROCm RNN-T Loss

jpvillam-amd:
Added HIPIFY code and small changes for ROCm. Targeting RNN-T loss.


mthrok:
Hi @jpvillam-amd 

Thanks for the contribution. Can you rebase on the latest main branch?
We had a bunch of CI breakage but it's been fixed on main branch.

CMakeLists.txt
```diff
@@ -85,6 +85,10 @@ if(USE_CUDA)
   enable_language(CUDA)
 endif()
 
+if(USE_ROCM)
+  enable_language(HIP)
```
malfet, line 89: HIP language is only available since CMake-3.21(see https://cmake.org/cmake/help/v3.21/command/enable_language.html), so please make `USE_ROCM` conditional on 3.21 CMake version

torchaudio/csrc/rnnt/options.h
```diff
@@ -18,6 +21,10 @@ typedef struct Options {
 #ifdef USE_CUDA
   // the stream to launch kernels in when using GPU.
   cudaStream_t stream_;
+#endif
+#ifdef USE_ROCM
+  // the stream to launch kernels in when using GPU.
+  hipStream_t stream_;
 #endif
```
malfet, line 28: Hmm, can we just add `#define cudaStream_t hipStream_t` under first `USE_ROCM` and then we would not need to duplicate the code here?

malfet (changes requested):
Please guard USE_ROCM with CMake version check (as TorchAudio should be builder on systems with CMake older than 3.21)

Also, can you please add an explanation why explicit `#ifdef __HIP_PLATFORM_AMD__`  must be explicitly added .cu files? I.e. why can't hipify module take care of those mechanistically changes?

nateanl:
cc @mthrok can you review this pr?

torchaudio/csrc/CMakeLists.txt
```diff
@@ -1,6 +1,29 @@
...
+  set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+  list( APPEND CMAKE_INSTALL_RPATH "/opt/rocm/llvm/lib" )
+  set(OPENMP_LIBRARIES "/opt/rocm/llvm/lib/")
+  set(OpenMP_CXX "${CMAKE_CXX_COMPILER}")
+  set(OpenMP_CXX_FLAGS "-fopenmp=libomp")
+  #set(OpenMP_CXX_LIB_NAMES "omp")
+  set(OpenMP_omp_LIBRARY omp)
+  find_package(OpenMP REQUIRED)
+
+endif()
```
mthrok, line 24: Could we move this to top-level CMakeLists.txt?
I would like to keep the logic in CMakeLists.txt files under `csrc` only related to source file selection.

torchaudio/csrc/CMakeLists.txt
```diff
@@ -39,6 +62,37 @@ if(BUILD_RNNT)
       rnnt/gpu/compute.cu
       )
   endif()
+
+  if (USE_ROCM)
+    hipify(CUDA_SOURCE_DIR ${PROJECT_SOURCE_DIR}/torchaudio/csrc/rnnt/gpu HIP_SOURCE_DIR ${PROJECT_SOURCE_DIR}/torchaudio/csrc/rnnt/hip)
+    if ( NOT HIP_ADD_LIBRARY_FOUND )
+      list(APPEND CMAKE_MODULE_PATH /opt/rocm/hip/cmake)
+      find_package(HIP REQUIRED)
+    endif()
```
mthrok, line 71: This looks like the duplicate from the above change, and better be handled in the root CMakeLists.txt?

mthrok:
> @malfet @mthrok It's not clear to me at all if these CI failures need any action from our end, since the failures are CUDA related. Can you please let us know what the next step should be here? Do we need another rebase?

Hi

torchaudio team does not have CI or expertise to validate the change, so I am thinking that we merge it given that it does not break existing CPU/CUDA. Could you address my comment? Thanks

torchaudio/csrc/CMakeLists.txt
jithunnair-amd: @pnunna93 I think the following snippet is the only one that's arguably related to source file update/selection. Do you agree?
```
  list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/third_party/hipify_torch/cmake")
  include(Hipify)
```

torchaudio/csrc/CMakeLists.txt
jithunnair-amd: @pnunna93 Here as well, I believe the hipify logic should stay here, since it's about source code modification (hipification), but the cmake stuff can probably move to root CMakeLists.txt?

jithunnair-amd:
@mthrok I'm taking a look at refactoring the CMake files to address the review comments

CMakeLists.txt
jeffdaily: The need for `enable_language(HIP)` was removed in the latest commit.

torchaudio/csrc/CMakeLists.txt
jeffdaily: Done.  Moved to top-level CMakeLists.txt.  New torchaudio/csrc/CMakeLists.txt contains a straightforward USE_ROCM section after each appearance of a USE_CUDA section.

jeffdaily:
@mthrok please re-review.  This PR is now up to date with latest main and the CMakeLists.txt has been improved.

jithunnair-amd:
@mthrok @malfet Can you please review?

mthrok (approved):

mthrok:
Looks good. Thanks. I will merge this.

jeffdaily:
> @jithunnair-amd
> 
> Should we be concerned about ROCm build job failures??
> 
> `libtorchaudio.so: undefined symbol: __kmpc_fork_call` https://github.com/pytorch/audio/actions/runs/5539782179/jobs/10111098244?pr=2485 https://github.com/pytorch/audio/actions/runs/5539782179/jobs/10111098349?pr=2485

Yes, we need to fix those before we can land this PR.

mthrok:
> > @jithunnair-amd
> > Should we be concerned about ROCm build job failures??
> > `libtorchaudio.so: undefined symbol: __kmpc_fork_call` https://github.com/pytorch/audio/actions/runs/5539782179/jobs/10111098244?pr=2485 https://github.com/pytorch/audio/actions/runs/5539782179/jobs/10111098349?pr=2485
> 
> Yes, we need to fix those before we can land this PR.

@jeffdaily @jithunnair-amd  Got it. Would you be able to look into them?

mthrok:
@jeffdaily @jithunnair-amd 

Seems like there was ROCm CI migration. Could you rebase or merge upstream? If possible, I would like to include this PR for the upcoming release.

torchaudio/csrc/rnnt/options.h
jeffdaily: typedef was added.  Please re-review.

jeffdaily:
> @jeffdaily @jithunnair-amd
> 
> Seems like there was ROCm CI migration. Could you rebase or merge upstream? If possible, I would like to include this PR for the upcoming release.

Done. @mthrok

jeffdaily:
Note that the CMakeLists.txt refactor was reverted.  Let's first focus on landing this PR that adds RNN-T for ROCm, then a follow-up PR that modernizes the CMake files.

mthrok:
> > Please guard USE_ROCM with CMake version check (as TorchAudio should be builder on systems with CMake older than 3.21)
> > Also, can you please add an explanation why explicit `#ifdef __HIP_PLATFORM_AMD__` must be explicitly added .cu files? I.e. why can't hipify module take care of those mechanistically changes?
> 
> Would you be okay with accepting this PR as-is and then a follow-up PR that revisits the hipify strategy here? It would be nice to get the feature in. The hipify and CMakeLists.txt refactor was causing more trouble than expected, so it was backed out of this PR.

@jeffdaily Yes, build CI jobs were green, so I will merge this.

jithunnair-amd:
@mthrok This PR introduces a dependency on `libomp.so` which is available with a standard ROCm installation in `/opt/rocm/llvm/lib/libomp.so`. However, since it's not packaged with the torchaudio wheel currently, and is also not packaged with the torch wheels, it results in an import-time failure for torchaudio:

```
>>> import torchaudio
  File "/usr/local/lib/python3.8/dist-packages/torchaudio/__init__.py", line 1, in <module>
    from . import (  # noqa: F401
  File "/usr/local/lib/python3.8/dist-packages/torchaudio/_extension/__init__.py", line 45, in <module>
    _load_lib("libtorchaudio")
  File "/usr/local/lib/python3.8/dist-packages/torchaudio/_extension/utils.py", line 64, in _load_lib
    torch.ops.load_library(path)
  File "/usr/local/lib/python3.8/dist-packages/torch/_ops.py", line 839, in load_library
    ctypes.CDLL(path)
  File "/usr/lib/python3.8/ctypes/__init__.py", line 373, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libomp.so: cannot open shared object file: No such file or directory
```

I'm investigating how to package the libomp.so with the torchaudio wheel for ROCm. If there are any existing instances where you do the same for any build of torchaudio (ROCm or otherwise), please point me to it, so I can re-use the same flow.

cc @jeffdaily @atalman

atalman:
@jeffdaily @jithunnair-amd how critical this PR is ? maybe we can just rever it ?

mthrok:
Hi @jeffdaily @jithunnair-amd - So this PR was reverted due to the packaging issue. I wanted to have this in the upcoming release and I do welcome the re-attempt. As to make the process faster for the future, I invited @jeffdaily for write permission. I hope we can continue and iterate on this feature. Regards

PR #2486, opened 2022-06-14
Pin MKL to 2020.04

mthrok:
The version of MKL that is installed alongside PyTorch has been bumped
to 2022.1 on Windows and it is causing installation issue in unit tests.

This commit pins the previous version

Backport of #2463

### DIFF
--- a/.circleci/unittest/windows/scripts/install.sh
+++ b/.circleci/unittest/windows/scripts/install.sh
@@ -29,7 +29,7 @@ else
     cudatoolkit="cudatoolkit=${version}"
 fi
 printf "Installing PyTorch with %s\n" "${cudatoolkit}"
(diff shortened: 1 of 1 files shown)


nateanl:
For unittest CI the mkl is fixed to [`2021.2.0`](https://github.com/pytorch/audio/blob/main/.circleci/unittest/linux/scripts/install.sh#L47), do we want to change that too?

mthrok:
> For unittest CI the mkl is fixed to [`2021.2.0`](https://github.com/pytorch/audio/blob/main/.circleci/unittest/linux/scripts/install.sh#L47), do we want to change that too?

It's okay to keep that way, as long as the job is not failing.

nateanl (approved):

PR #2493, opened 2022-06-15
Add docs for musdb and reference in bibliography

skim0514:
Add references and docs for musdb and citation taken from dataset.

### DIFF
--- a/docs/source/refs.bib
+++ b/docs/source/refs.bib
@@ -359,3 +359,15 @@
   title="YesNo",
   url="http://www.openslr.org/1/"
 }
+@misc{MUSDB18HQ,
+  author       = {Rafii, Zafar and
+                  Liutkus, Antoine and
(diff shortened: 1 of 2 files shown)


docs/source/refs.bib
```diff
@@ -359,3 +359,15 @@ @misc{YesNo
   title="YesNo",
   url="http://www.openslr.org/1/"
 }
+@misc{MUSDB18HQ,
+  author       = {Rafii, Zafar and
```
carolineechen, line 363: could you condense the authors into one line?

carolineechen (commented):
thanks for these additions! could you add this commit to the original musdb dataset PR (and close this one)? we should try to include docs/citation in the same PR so that we can confirm docs is building correctly, and the dataset description should also reference the citation, so it'll be needed in the original PR

PR #2496, opened 2022-06-15
Making sure channel flag is set correctly

atalman:
Making sure channel flag is set correctly for the test channel


mthrok (approved):
stamp

PR #2506, opened 2022-06-24
Hybrid Demucs model implementation

skim0514:
Draft PR with initial model implementation with minor changes from previous implementation


torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+        r"""Encoder layer. This used both by the time and the frequency branch.
+        Args:
+            chin (int): number of input channels.
+            chout (int): number of output channels.
+            kernel_size (int): Kernel size for encoder (Default: 8)
+            stride (int): Stride for encoder layer (Default: 4)
+            norm_groups (int): number of groups for group norm. (Default: 11)
+            empty: used to make a layer with just the first conv. this is used
+                before merging the time and freq. branches. (Default: ``False``)
+            freq (bool): boolean for whether conv layer is for frequency (Default: ``True``)
```
nateanl, line 76: can we make this argument to something like "encoder_type"? and make the default value to "frequency".

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+            norm (bool): use GroupNorm. (Default: ``True``)
```
nateanl, line 77: same here, we can change it to "norm_type", the options can be `["none", "group_norm"]`.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+    or complex as channels (CaC) [Choi et al. 2020]. Wiener filtering is based on
+    Open Unmix implementation [Stoter et al. 2019].
+    The loss is always on the temporal domain, by backpropagating through the above
+    output methods and iSTFT. This allows to define hybrid models nicely. However, this breaks
+    a bit Wiener filtering, as doing more iteration at test time will change the spectrogram
+    contribution, without changing the one from the waveform, which will lead to worse performance.
+    I tried using the residual option in OpenUnmix Wiener implementation, but it didn't improve.
+    CaC on the other hand provides similar performance for hybrid, and works naturally with
+    hybrid models.
+    This model also uses frequency embeddings are used to improve efficiency on convolutions
```
nateanl, line 281:
```suggestion
    Frequency embeddings are used to improve efficiency on convolutions
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+    over the freq. axis, following [Isik et al. 2020] (https://arxiv.org/pdf/2008.04470.pdf).
+    Unlike classic Demucs, there is no resampling here, and normalization is always applied.
+    """
+
+    def __init__(
+        self,
+        sources: List[str],
+        # Channels
```
nateanl, line 289:
```suggestion
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+        norm_starts: int = 4,
+        norm_groups: int = 4,
+        # DConv residual branch
+        dconv_depth: int = 2,
+        dconv_comp: int = 4,
+        dconv_attn: int = 4,
+        dconv_lstm: int = 4,
+        dconv_init: float = 1e-4,
+        # Metadata
+        samplerate: int = 44100,
```
nateanl, line 316:
```suggestion
        sample_rate: int = 44100,
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+        segment: int = 4 * 10,
+    ):
+        r"""
+        Args:
```
nateanl, line 320: put the docstring above the initialization method can make it visible in the documentation website, but I'm not sure. Could you add the model to `docs/source/models.rst` to see whether the docstring is visible?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+        # self.depth = depth
+        self.channels = channels
+        self.samplerate = samplerate
+        self.segment = segment
+
+        self.hop_length = self.nfft // 4
+        self.freq_emb = None
+
+        self.encoder = nn.ModuleList()
+        self.decoder = nn.ModuleList()
```
nateanl, line 374: I feel like renaming them to `frequency_encoder`, `frequency_decoder` could be more intuitive. what do you think?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+
+        self.tencoder = nn.ModuleList()
+        self.tdecoder = nn.ModuleList()
+
+        chin = audio_channels
+        chin_z = chin * 2  # number of channels for the freq branch
+        chout = channels_time or channels
```
nateanl, line 381: This is a little confusing. Which argument should user choose for `chout`?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+        conv.bias.data /= scale
+
+
+def rescale_module(module):
+    for sub in module.modules():
+        if isinstance(sub, (nn.Conv1d, nn.ConvTranspose1d, nn.Conv2d, nn.ConvTranspose2d)):
+            rescale_conv(sub)
+
+
+def spectro(x: torch.Tensor, n_fft: int = 512, hop_length: int = 0, pad: int = 0) -> torch.Tensor:
```
nateanl, line 899: The `spectro` method can be replaced by `torchaudio.functional.spectrogram`.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,939 @@
...
+def rescale_conv(conv):
+    """Rescale initial weight scale. It is unclear why it helps but it certainly does."""
+    std = conv.weight.std().detach()
+    scale = (std / 0.1) ** 0.5
+    conv.weight.data /= scale
+    if conv.bias is not None:
+        conv.bias.data /= scale
+
+
+def rescale_module(module):
```
nateanl, line 893: can you add docstring to this method?

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -0,0 +1,108 @@
...
+        out = model(x)
+
+        assert out.size() == (1, chout, 512, 739)
+
+
+class HDemucsDecoderTests(TorchscriptConsistencyMixin):
+    def test_hdemucs_decoder_torchscript_consistency(self):
+        r"""Validate the torchscript consistency of a Encoder."""
+        chin, chout = 48, 96
+        model = HEncLayer(chin, chout).to(self.device).eval()
```
carolineechen, line 60: Should this be HDecLayer?

carolineechen (commented):
overall code structure looks pretty good, will take a closer look at the model details later. some high level comments:
- docstring improvements
    - as Zhaoheng mentioned, moving out `__init__` params to above the function instead of under. you should add models.rst and be able to run `make html` or use CircleCI build_docs job to visualize the rendered output
    - for parameters with a default value, they should be annotated like `(bool, optional)` instead of just `(bool)`
- testing: rather than create a new test class for each major model component, you can put them together under larger test class (torchscript tests could be under the same class or a separate class, but it is likely we won't address torchscript in this PR). torch64 tests can be removed as well, as it is not supported in original hdemucs
- add underscore in front of helper functions or classes that aren't exposed to the user (basically everything except the main `HDemucs` class)
- bundling/options: if implementing the bundling here, could be good to add a brief comment/description on it (reason, how it works, corresponding values, or whatever you think relevant for the user to know that might not be obvious)
- would be nice to include more impl/feature details in the PR summary (ex/ removed xxxx parameters, improved docstring and param types, added xxx tests, ...), and maybe also a brief follow ups section (add pretrained factory functions, etc)

torchaudio/prototype/models/hdemucs.py
skim0514: resolved upon convo

nateanl: after offline discussion with @skim0514, we should keep the `freq` argument as it is. changing the argument will make it more challenging when importing the pre-trained weight from demucs repo.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,938 @@
...
+        if not self.last:
+            z = F.gelu(z)
+
+        return z, y
+
+
+class HDemucs(torch.nn.Module):
+    r"""
+    Hybrid Demucs model from *Hybrid Spectrogram and Waveform Source Separation* [:footcite:`defossez2021hybrid`].
+    The hybrid model has the same structure as Demucs, except having a frequency and time axis.
```
carolineechen, line 295: You can remove the explanation of the model (and extra footcite), keeping just the first line here is fine.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,938 @@
...
+    The opposite happens in the decoder. Models can either use naive iSTFT from masking,
+    The loss is always on the temporal domain, by backpropagating through the above
+    output methods and iSTFT. This allows to define hybrid models nicely.
+    Frequency embeddings are used to improve efficiency on convolutions
+    over the freq. axis, following [:footcite:`isik2020poconet`].
+    Unlike classic Demucs, there is no resampling here, and normalization is always applied.
+
+
+    Args:
+        sources (list[str]): list of source names.
```
carolineechen, line 307:
```suggestion
        sources (List[str]): list of source names.
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,938 @@
...
+    ):
+        super().__init__()
+        self.depth = depth
+        self.nfft = nfft
+        self.audio_channels = audio_channels
+        self.sources = sources
+        self.kernel_size = kernel_size
+        self.context = context
+        self.stride = stride
+        # self.depth = depth
```
carolineechen, line 370:
```suggestion
```

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -0,0 +1,105 @@
...
+    (["bass", "vocals"],),
+    (["vocals"],),
+]
+
+
+def _get_hdemucs_model(sources):
+    return HDemucs(sources)
+
+
+class TorchscriptConsistencyMixin(TestBaseMixin):
```
carolineechen, line 18: let's remove the torchscript tests for this PR. we can follow up on torchscriptability in the future

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,938 @@
...
+
+        Args:
+            x (torch.Tensor): tensor input of shape (1, chin, frequency_bins, num_frames/1024) for frequency and shape
+                (1, chin, num_frames/chin) for time
+            inject (torch.Tensor, optional): on last layer, combine frequency and time branches through inject param,
+                same shape as x (default: ``None``)
+
+        Returns:
+            Tensor
+                output tensor after encoder layer of shape (1, chout, frequency_bins, num_frames/1024) for frequency
```
carolineechen, line 152: can you wrap the tensor shapes in docstrings with `` backticks?

```suggestion
                output tensor after encoder layer of shape `(1, chout, frequency_bins, num_frames/1024)` for frequency
```

carolineechen (commented):
a few comments, and let's try to get the circle ci tests passing

torchaudio/prototype/models/hdemucs.py
skim0514: Leaving as is upon offline discussion, possible future adjustments to spectrogram normalization will be needed.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        scale (float, optional): amount to scale (Default: 10.0)
+        smooth (bool, optional): choose to apply smoothing (Default: ``False``)
+    """
+
+    def __init__(self, num_embeddings: int, embedding_dim: int, scale: float = 10.0, smooth: bool = False):
+        super().__init__()
+        self.embedding = nn.Embedding(num_embeddings, embedding_dim)
+        if smooth:
+            weight = torch.cumsum(self.embedding.weight.data, dim=0)
+            # when summing gaussian, overscale raises as sqrt(n), so we nornalize by that.
```
adefossez, line 50: nit: typo in the original code: overscale -> scale

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+from torch.nn import functional as F
+
+
+class _ScaledEmbedding(torch.nn.Module):
+    r"""Make continuous embeddings and boost learning rate
+
+    Args:
+        num_embeddings (int): number of embeddings
+        embedding_dim (int): embedding dimensions
+        scale (float, optional): amount to scale (Default: 10.0)
```
adefossez, line 41: nit: scale: amount to scale the learning rate by (actual embedding scale is not changed).

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+
+class _HEncLayer(torch.nn.Module):
+
+    r"""Encoder layer. This used both by the time and the frequency branch.
+    Args:
+        chin (int): number of input channels.
+        chout (int): number of output channels.
+        kernel_size (int, optional): Kernel size for encoder (Default: 8)
+        stride (int, optional): Stride for encoder layer (Default: 4)
+        norm_groups (int, optional): number of groups for group norm. (Default: 11)
```
adefossez, line 81: Should  be 1. Although in HDemucs we set the default to 4, so maybe we should use this here too.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        empty (bool, optional): used to make a layer with just the first conv. this is used
+            before merging the time and freq. branches. (Default: ``False``)
+        freq (bool, optional): boolean for whether conv layer is for frequency (Default: ``True``)
```
adefossez, line 84: nit: "is along the frequency axis", or "is for the frequency domain".

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        norm_type (bool, optional): Norm type, either ``group_norm `` or ``none`` (Default: ``group_norm``)
```
adefossez, line 85: nit: wrong type bool -> str

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+            self.norm2 = norm_fn(2 * chout)
+            self.dconv = _DConv(chout, **dconv_kw)
+
+    def forward(self, x: torch.Tensor, inject: Optional[torch.Tensor] = None) -> torch.Tensor:
+        r"""Forward pass for encoding layer.
+
+        Size depends on whether frequency or time
+
+        Args:
+            x (torch.Tensor): tensor input of shape `(1, chin, frequency_bins, num_frames/1024)` for frequency and shape
```
adefossez, line 144: not sure it is worth hardcoding the shape here, as this will depend on the hyper params of the model. Maybe just mention `B, C, T` for time domain and `B, C, F, T` for frequency domain.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        chin (int): number of input channels.
+        chout (int): number of output channels.
+        last (bool, optional): whether current layer is final layer (Default: ``False``)
+        kernel_size (int, optional): Kernel size for encoder (Default: 8)
+        stride (int): Stride for encoder layer (Default: 4)
+        norm_groups (int, optional): number of groups for group norm. (Default: 1)
+        empty (bool, optional): used to make a layer with just the first conv. this is used
+            before merging the time and freq. branches. (Default: ``False``)
+        freq (bool, optional): boolean for whether conv layer is for frequency (Default: ``True``)
+        norm_type (bool, optional): Norm type, either ``group_norm `` or ``none`` (Default: ``group_norm``)
```
adefossez, line 195: nit: bool -> str

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        self.rewrite = klass(chin, 2 * chin, 1 + 2 * context, 1, context)
+        self.norm1 = norm_fn(2 * chin)
+
+    def forward(self, x: torch.Tensor, skip: Optional[torch.Tensor], length):
+        r"""Forward pass for decoding layer.
+
+        Size depends on whether frequency or time
+
+        Args:
+            x (torch.Tensor): tensor input of shape `(1, chin, frequency_bins, num_frames/1024)` for frequency and shape
```
adefossez, line 254: same as for the encoder.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+                `(1, chin, num_frames/chin)` for time
+            inject (torch.Tensor, optional): on last layer, combine frequency and time branches through inject param,
+                same shape as x (default: ``None``)
+
+        Returns:
+            Tensor
+                output tensor after encoder layer of shape `(1, chout, frequency_bins, num_frames/1024)` for frequency
```
adefossez, line 151: Then shape would be 
`B, C, ceil(T / stride)` for time domain or `B, C, F / stride, T` for frequency domain.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+                    and shape `(1, chout, num_frames/chout)`
+        """
+        if not self.freq and x.dim() == 4:
+            B, C, Fr, T = x.shape
+            x = x.view(B, -1, T)
+
+        if not self.freq:
```
adefossez, line 158: we might need some assert here, that either `Fr == self.kernel_size or stride != 1 and Fr % stride = 0`. This should always be the case when using this through HDemucs, but better add the assert in case someone uses it outside of it.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        context: int = 1,
+        dconv_kw: dict = {},
+        pad: bool = True,
+    ):
+        super().__init__()
+        norm_fn = lambda d: nn.Identity()  # noqa
+        if norm_type == "group_norm":
+            norm_fn = lambda d: nn.GroupNorm(norm_groups, d)  # noqa
+        if pad:
+            pad = kernel_size // 4
```
adefossez, line 222: correct formula would be `(kernel_size - stride) // 2` + add an `assert kernel_size - stride % 2 == 0` for supporting generic params.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        Args:
+            x (torch.Tensor): tensor input of shape `(1, chin, frequency_bins, num_frames/1024)` for frequency and shape
+                `(1, chin, num_frames/chin)` for time
+            skip (torch.Tensor, optional): on first layer, separate frequency and time branches using param
+                (default: ``None``)
+            length (int): Size of tensor for output
+
+        Returns:
+            (Tensor, Tensor):
+                Tensor
```
adefossez, line 262: Here same thing, formula would either be `B, C, T` -> `B, C, stride * T` for time domain, `B, C, F, T` -> `B, C, F * stride, T` for freq domain, except if `pad=False` and `F=1`, in which case we get `B, C, kernel_size, T` (this is the case for the last frequency only layer, the one for which `last_freq=True` later in HDemucs.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+class HDemucs(torch.nn.Module):
+    r"""
+    Hybrid Demucs model from *Hybrid Spectrogram and Waveform Source Separation* [:footcite:`defossez2021hybrid`].
+
+    Args:
+        sources (List[str]): list of source names.
+        audio_channels (int, optional): input/output audio channels. (Default: 2)
+        channels (int, optional): initial number of hidden channels. (Default: 48)
+        growth (int, optional): increase the number of hidden channels by this factor at each layer. (Default: 2)
+        nfft (int, optional): number of fft bins. Note that changing this requires careful computation of
```
adefossez, line 301: is the change we discussed going to be in a second PR or you will update this one ?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,924 @@
...
+        pad = hl // 2 * 3
+        x = F.pad(x, [pad, pad + le * hl - x.shape[-1]], mode="reflect")
+
+        z = _spectro(x, nfft, hl)[..., :-1, :]
+        assert z.shape[-1] == le + 4, (z.shape, x.shape, le)
+        z = z[..., 2 : 2 + le]
+        return z
+
+    def _ispec(self, z, length=None, scale=0):
+        hl = self.hop_length // (4 ** scale)
```
adefossez, line 480: you can remove the scale stuff. At some point I wanted to do multi scale spectrogram, but that didn't add anything.

adefossez (commented):
Looking good ! Mostly minor nits and a few asserts to add.
This is probably the biggest chunk ! Let me know if this works with the `e51eebcc` model.

faroit:
Are you also adding training recipes later or will this just be for using the pretrained models?

skim0514:
> Are you also adding training recipes later or will this just be for using the pretrained models?

skim0514:
> Are you also adding training recipes later or will this just be for using the pretrained models?

The main goal of this is to use with pretrained models. Depending on the timeline of my internship adding further training information is a part of my goals, but is not the focus.

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -0,0 +1,76 @@
...
+        torch.random.manual_seed(40)
+        output = model(*tensors)
+
+        torch.random.manual_seed(40)
+        ts_output = ts_func(*tensors)
+
+        self.assertEqual(ts_output, output)
+
+
+class HDemucsTests(TorchscriptConsistencyMixin):
```
carolineechen, line 33: You should be able to just use `TestBaseMixin` here and remove `TorchscriptConsistencyMixin` above since we're not testing torchscript and using the function implemented above

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -0,0 +1,76 @@
...
+        assert split_sample.shape == (1, len(sources), 2, length * 44100)
+
+    def test_encoder_output_shape(self):
+        r"""Feed tensors with specific shape to HDemucs Decoder and validate
+        that it outputs with a tensor with expected shape.
+        """
+        chin, chout = 4, 48
+        model = _HEncLayer(chin, chout).to(self.device).eval()
+
+        x = torch.rand(1, chin, 2048, 739, device=self.device, dtype=self.dtype)
```
carolineechen, line 59: can you make variables similar to `chin`, `chout` to indicate what 1, 2048,  and 739 represent? and if you could reuse them in the result that makes it more intuitive to understand since these may look like arbitrary numbers currently

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -0,0 +1,76 @@
...
+        length = 10
+
+        model = _get_hdemucs_model(sources).to(self.device).eval()
+        inputs = self._get_inputs(length)
+
+        split_sample = model(inputs)
+
+        assert split_sample.shape == (1, len(sources), 2, length * 44100)
+
+    def test_encoder_output_shape(self):
```
carolineechen, line 52: looks like both encoder and decoder will accept different shapes depending on if it's working in the frequency or time domain and currently only the frequency version is being tested. can you include tests for time domain as well?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,927 @@
...
+
+    def __init__(self, channels: int, heads: int = 4, ndecay: int = 4):
+        r"""
+        Args:
+            channels (int): Size of Conv1d layers.
+            heads (int, optional):  (default: 4)
+            ndecay (int, optional): (default: 4)
+        """
+        super(_LocalState, self).__init__()
+        assert channels % heads == 0, (channels, heads)
```
nateanl, line 774: could you add more message when the assertion fails?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,927 @@
...
+            y = y.view(B, Fr, C, T).permute(0, 2, 1, 3)
+        else:
+            y = self.dconv(y)
+        z = self.norm2(self.rewrite(y))
+        z = F.glu(z, dim=1)
+        return z
+
+
+class _HDecLayer(torch.nn.Module):
+    r"""Decoder layer. This used both by the time and the frequency branch.
```
nateanl, line 184:
```suggestion
    r"""Decoder layer. This used both by the time and the frequency branches.
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -278,14 +278,14 @@ def forward(self, x: torch.Tensor, skip: Optional[torch.Tensor], length):
             y = F.glu(self.norm1(self.rewrite(x)), dim=1)
         else:
             y = x
-            assert skip is None
+            assert skip is None, "skip is none when empty is true."
```
carolineechen, line 281:
```suggestion
            assert skip is None, "skip must be None when empty is true."
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -469,13 +469,13 @@ def _spec(self, x):
         # which is not supported by torch.stft.
         # Having all convolution operations follow this convention allow to easily
         # align the time and frequency branches later on.
-        assert hl == nfft // 4
+        assert hl == nfft // 4, "hop length is nfft // 4"
```
carolineechen, line 472:
```suggestion
        assert hl == nfft // 4, "hop length must be be nfft // 4"
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -469,13 +469,13 @@ def _spec(self, x):
...
         le = int(math.ceil(x.shape[-1] / hl))
         pad = hl // 2 * 3
         x = F.pad(x, [pad, pad + le * hl - x.shape[-1]], mode="reflect")
 
         z = _spectro(x, nfft, hl)[..., :-1, :]
-        assert z.shape[-1] == le + 4, (z.shape, x.shape, le)
+        assert z.shape[-1] == le + 4, "z.shape's last index must be 4 + le"
```
carolineechen, line 478: can this be more descriptive, what are `z` and `le`? ex/ spectrogram's last dimension must be ...

carolineechen (commented):
some nits

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,931 @@
...
+        smooth (bool, optional): choose to apply smoothing (Default: ``False``)
+    """
+
+    def __init__(self, num_embeddings: int, embedding_dim: int, scale: float = 10.0, smooth: bool = False):
+        super().__init__()
+        self.embedding = nn.Embedding(num_embeddings, embedding_dim)
+        if smooth:
+            weight = torch.cumsum(self.embedding.weight.data, dim=0)
+            # when summing gaussian, scale raises as sqrt(n), so we normalize by that.
+            weight = weight / torch.arange(1, num_embeddings + 1).to(weight).sqrt()[:, None]
```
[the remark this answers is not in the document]
skim0514: Removing works fine, but isn't this just to align the new tensor's dtype and device with weight's

carolineechen (approved):
looks like `build_docs` CircleCI job is broken, could you build docs locally and comment or add to description the screenshot of what the docs looks like?

otherwise this looks good, thanks for addressing the feedback and working on this!

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
mthrok: Are we keeping this?

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,928 @@
...
+        norm_type: str = "group_norm",
+        context: int = 0,
+        dconv_kw: dict = {},
+        pad: bool = True,
+    ):
+        super().__init__()
+        norm_fn = lambda d: nn.Identity()  # noqa
+        if norm_type == "group_norm":
+            norm_fn = lambda d: nn.GroupNorm(norm_groups, d)  # noqa
+        pad_new = kernel_size // 4 if pad else 0
```
mthrok, line 110:
```suggestion
        pad_val = kernel_size // 4 if pad else 0
```

`new` is not intuitively clear what it semantically means

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,928 @@
...
+        return m
+
+    def _mask(self, m):
+        # `m` is a full spectrogram and `z` is ignored.
+        B, S, C, Fr, T = m.shape
+        out = m.view(B, S, -1, 2, Fr, T).permute(0, 1, 2, 4, 5, 3)
+        out = torch.view_as_complex(out.contiguous())
+        return out
+
+    def forward(self, input):
```
carolineechen, line 504:
```suggestion
    def forward(self, input: torch.Tensor):
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -0,0 +1,928 @@
...
+
+        r"""Demucs total forward call
```
carolineechen, line 506:
```suggestion
        r"""HDemucs forward call
```

carolineechen (approved):

PR #2528, opened 2022-07-07
Add NNLM support to CTC Decoder

carolineechen:
Expose flashlight's LM and LMState classes to support decoding with custom language models, including NN LMs.

The `ctc_decoder` API is as follows
- To decode with KenLM, pass in KenLM language model path to `lm` variable
- To decode with custom LM, create Python class with `CTCDecoderLM` subclass, and pass in the class to `lm` variable. Additionally create a file of LM words listed in order of the LM index, with a word per line, and pass in the file to `lm_path`.
- To decode without a language model, set `lm` to `None` (default)

Validated against fairseq w2l decoder on sample LibriSpeech dataset and LM. Code for validation can be found [here](https://github.com/facebookresearch/fairseq/compare/main...carolineechen:fairseq:ctc-decoder). Also added unit tests to validate custom implementations of ZeroLM and KenLM, and also using a biased LM.

Follow ups:
- Train simple LM on LibriSpeech and demonstrate usage in tutorial or examples directory

cc @jacobkahn


torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -218,7 +219,11 @@ def ctc_decoder(
             decoding.
         tokens (str or List[str]): file or list containing valid tokens. If using a file, the expected
             format is for tokens mapping to the same index to be on the same line
-        lm (str or None, optional): file containing language model, or `None` if not using a language model
+        lm (str, CTCDecoderLM, or None, optional): either a path containing KenLM language model,
```
carolineechen, line 222: would this need to be moved to the end for bc-breaking purposes?

mthrok: I think it's okay to introduce BC-breaking change. having two related arguments 10 arguments away seems not a great UX, and I bet many people use keyword argument here.

mthrok (commented):
Overall change looks good.

Let's think about what kind of unit test we can perform on this. 
We can use some biased toy NN LM as a placement in the test and verify the internal behavior from outside.

After this change, I think the branching around lexicon-free vs lexicon part can be refactored.

torchaudio/models/decoder/_ctc_decoder.py
carolineechen: should we add a warning inside the function to indicate this argument change?

mthrok: I think we can do type check on `lm_dict` value.
We want to avoid the situation where things keep working but in a wrong way. (though I doubt that happens)

Users using keywords arguments are not affected, but those pass `nbest` as positional argument are affected.
So ensuring that `lm_dict` is `None` or `str` should suffice.

torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -270,12 +282,22 @@ def ctc_decoder(
         >>> )
         >>> results = decoder(emissions) # List of shape (B, nbest) of Hypotheses
     """
+    if lm_dict is not None and type(lm_dict) is not str:
+        raise ValueError("lm_dict must be None or str type.")
+
     tokens_dict = _Dictionary(tokens)
+    if lm_dict:
+        word_dict = _Dictionary(lm_dict)
```
jacobkahn, line 290: super nit, but the lexicon is also a dictionary of words, so maybe call this `_lm_dict`

torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -270,12 +282,22 @@ def ctc_decoder(
         >>> )
         >>> results = decoder(emissions) # List of shape (B, nbest) of Hypotheses
     """
+    if lm_dict is not None and type(lm_dict) is not str:
+        raise ValueError("lm_dict must be None or str type.")
+
     tokens_dict = _Dictionary(tokens)
+    if lm_dict:
```
jacobkahn, line 289: it's a little expensive (esp in python so probably not worth doing), but a restriction in the decoder for obvious reasons is that LM dictionaries must only have words which are in lexicons [if both lexicons and NNLM dicts are provided]. Flashlight will throw an exception internally if this isn't the case iirc, but perhaps worth noting in docs.

jacobkahn (approved):
LGTM! This is great to see. Some quick notes.

torchaudio/models/decoder/_ctc_decoder.py
carolineechen: I have it as `word_dict` (lexicon dict) in the current implementation because if `lm_dict` is provided, it actually is used instead of the lexicon file to construct the dictionary of words. this way the custom LM index and scoring output aligns with the lexicon indexing, and we assume lm_dict and lexicon consist of the same words. let me know if this is reasonable

jacobkahn: Hmm — I see what you're saying. This could work if that's what you want the behavior to be, but I'm not sure if this behavior makes sense.

First — does this work as expected? The LM dict is typically a mapping of words to ID — this needs to be parsed out so it's only a list of words, right? There are a few other assumptions here: first, if you're constructing that dictionary manually, it assumes you're doing token-level decoding. You'd need to check consistency between the token set in the LM dict and the token output from the model.

Thing is, we've done lexicon free training with language models, so you still might not want a lexicon: https://github.com/flashlight/wav2letter/tree/main/recipes/lexicon_free

test/torchaudio_unittest/common_utils/ctc_decoder_utils.py
```diff
@@ -0,0 +1,32 @@
+from torchaudio.models.decoder import CTCDecoderLM, CTCDecoderLMState
```
mthrok, line 1: I don't think these will be reused outside of `ctc_decoder_test.py`. Can we put this there for now?

test/torchaudio_unittest/common_utils/ctc_decoder_utils.py
```diff
@@ -0,0 +1,32 @@
...
+    def start(self, start_with_nothing: bool):
+        return CTCDecoderLMState()
+
+    def score(self, state: CTCDecoderLMState, token_index: int):
+        return (state.child(token_index), 0.0)
+
+    def finish(self, state: CTCDecoderLMState):
+        return (state, 0.0)
+
+
```
mthrok, line 18: As a baseline, can we add some toy custom LM that's neither ZeroLM nor KenLM, that
- returns hardcoded scores for prefixed input pattern ?
- runs `torch.nn.Module` inside. (to be sure that such integration works without a software issue)

test/torchaudio_unittest/models/ctc_decoder/ctc_decoder_test.py
```diff
@@ -72,14 +80,53 @@ def test_timesteps_shape(self, use_lexicon):
...
+        emissions = self._get_emissions()
+        lm_dict = get_asset_path("decoder/nnlm_dict.txt")
+
+        custom_nnlm = self._get_custom_foolm(lm_dict)
+        nnlm_decoder = self._get_decoder(lm=custom_nnlm, lm_dict=lm_dict)
+        nnlm_results = nnlm_decoder(emissions)
+
+        for result in nnlm_results:
+            for word in result[0].words:
+                assert word == "foo"
```
mthrok, line 129: Can we unnest these two for-loops?
If either of `nnlm_resutls`  or `.words` are empty, then no assertion is performed and the tests pass.

something like
```
expected = ["foo" for _ in range(N) for _ in range(M)]  # N and M corresponds to expected length, but hard coded
output = [res[0].words for res in nnlm_results]
assert expected == output
```

mthrok (approved):
LGTM with some nits, given that . Can you reflect the library structure in test as well? `torchaudio_unittest/models/`**decoder**`/ctc_decodet_test.py`

PR #2543, opened 2022-07-13
Fix build_docs job

mthrok:
For some mysterious reason, `resampy=0.3.1` causes slowdown of unrelated code. https://github.com/bmcfee/resampy/issues/106

This commit fix build_docs job timeout.


carolineechen (approved):
nice!

Issue #2536, opened 2022-07-10
Are the nightly conda builds broken?


greaber:
```
13:35 $ conda install pytorch torchvision torchaudio -c pytorch-nightly
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.

PackagesNotFoundError: The following packages are not available from current channels:

  - torchaudio

Current channels:

  - https://conda.anaconda.org/pytorch-nightly/osx-arm64
  - https://conda.anaconda.org/pytorch-nightly/noarch
  - https://repo.anaconda.com/pkgs/main/osx-arm64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/osx-arm64
  - https://repo.anaconda.com/pkgs/r/noarch

To search for alternate channels that may provide the conda package you're
looking for, navigate to

    https://anaconda.org

and use the search bar at the top of the page.
```

Tested just now on OSX

hwangjeff:
Hi @greaber — are you still seeing this? I tried this just now on OS X and it worked fine.

greaber:
Yeah, I just tried with a fresh conda env and python 3.9, and I get the same error

greaber:
Are you using an M1 mac?  Possibly it could be related to this

mthrok:
It seems that it is related to macOS version. The recent necessary migration on CircleCI silently bumped the minimum supported version to macOS 11. We need to fix this.

greaber:
The minimum macOS version might be an issue for some people, but I don't think it is what I am seeing since I am on 12.4

mthrok:
@greaber Can you share the result of `python -m torch.utils.collect_env`?

greaber:
@mthrok Do you mean if I install `torch` and `torchvision` using `conda install pytorch torchvision -c pytorch-nightly`?

In that case,

```
$ python -m torch.utils.collect_env
Collecting environment information...
PyTorch version: 1.13.0.dev20220719
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 12.4 (arm64)
GCC version: Could not collect
Clang version: 13.1.6 (clang-1316.0.21.2.3)
CMake version: Could not collect
Libc version: N/A

[...]

PR #2546, opened 2022-07-13
Recent CircleCI migration #2529

mthrok:
silently bumped the minimum supported macOS version to 11.

PyTorch still supports 10.9 and the ecosystem still uses 10.9.
Issue: #2536

This commit sets MACOSX_DEPLOYMENT_TARGET=10.9, so that binary
distribution are compatible on macOS=10.9.


mthrok:
The package name now contains 10.9, so it should be good.

https://app.circleci.com/pipelines/github/pytorch/audio/11680/workflows/22411660-0e5f-4135-a1dc-1850e8b8bf5c/jobs/786683

`Copying torchaudio.egg-info to build/bdist.macosx-10.9-x86_64/wheel/torchaudio-0.13.0.dev20220713-py3.8.egg-info`

PR #2547, opened 2022-07-14
Adding pipeline changes, factory functions to HDemucs

skim0514:
Factory functions have been added to HDemucs class and test the implementation within the testing files.

### DIFF
--- a/test/torchaudio_unittest/prototype/hdemucs_cpu_test.py
+++ b/test/torchaudio_unittest/prototype/hdemucs_cpu_test.py
@@ -1,8 +1,8 @@
 import torch
 from torchaudio_unittest.common_utils import PytorchTestCase
-from torchaudio_unittest.prototype.hdemucs_test_impl import HDemucsTests
+from torchaudio_unittest.prototype.hdemucs_test_impl import HDemucsTests, TestDemucsIntegration
 
 
-class HDemucsFloat32CPUTest(HDemucsTests, PytorchTestCase):
+class HDemucsFloat32CPUTest(HDemucsTests, TestDemucsIntegration, PytorchTestCase):
     dtype = torch.float32
     device = torch.device("cpu")
--- a/test/torchaudio_unittest/prototype/hdemucs_test_impl.py
+++ b/test/torchaudio_unittest/prototype/hdemucs_test_impl.py
@@ -1,18 +1,28 @@
+from typing import List
+
 import torch
 from parameterized import parameterized
-from torchaudio.prototype.models.hdemucs import _HDecLayer, _HEncLayer, HDemucs
-from torchaudio_unittest.common_utils import TestBaseMixin
+from torchaudio.prototype.models.hdemucs import (
+    _HDecLayer,
+    _HEncLayer,
+    HDemucs,
+    hdemucs_high,
+    hdemucs_low,
+    hdemucs_medium,
+)
+from torchaudio_unittest.common_utils import skipIfNoModule, TestBaseMixin, TorchaudioTestCase
 
 
-def _get_hdemucs_model(sources):
-    return HDemucs(sources)
+def _get_hdemucs_model(sources: List[str], n_fft: int = 4096, depth: int = 6, sample_rate: int = 44100):
+    return HDemucs(sources, nfft=n_fft, depth=depth, sample_rate=sample_rate)
+
+
+def _get_inputs(duration: int, channels: int, batch_size: int, sample_rate: int):
+    sample = torch.rand(batch_size, channels, duration * sample_rate, dtype=torch.float32)
+    return sample
 
 
 class HDemucsTests(TestBaseMixin):
-    def _get_inputs(self, duration: int, channels: int, batch_size: int, sample_rate: int):
-        sample = torch.rand(batch_size, channels, duration * sample_rate, dtype=torch.float32, device=self.device)
-        return sample
-
     @parameterized.expand(
         [
             (["bass", "drums", "other", "vocals"],),
@@ -31,7 +41,7 @@ class HDemucsTests(TestBaseMixin):
         sample_rate = 44100
 
         model = _get_hdemucs_model(sources).to(self.device).eval()
-        inputs = self._get_inputs(duration, channels, batch_size, sample_rate)
+        inputs = _get_inputs(duration, channels, batch_size, sample_rate)
 
         split_sample = model(inputs)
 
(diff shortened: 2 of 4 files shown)


test/torchaudio_unittest/prototype/hdemucs_cpu_test.py
```diff
@@ -1,8 +1,8 @@
 import torch
 from torchaudio_unittest.common_utils import PytorchTestCase
-from torchaudio_unittest.prototype.hdemucs_test_impl import HDemucsTests
+from torchaudio_unittest.prototype.hdemucs_test_impl import HDemucsTests, TestDemucsIntegration
```
mthrok, line 3: we have integration test in `test/integration_test` directory, so can we call this something else?

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,115 @@ def test_decoder_output_shape_time(self):
...
         assert z.size() == (batch_size, chout, t * stride)
         assert y.size() == (batch_size, chin, t)
+
+
+@skipIfNoModule("demucs")
+class TestDemucsIntegration(TorchaudioTestCase):
+    """Test the process of importing the models from demucs.
+
+    Test methods in this test suite check the following things
+    1. Models loaded with demucs can be imported.
```
mthrok, line 126: I guess this description is from w2v2 test, but it does not apply here. w2v2 test also includes the function that converts original w2v2 to torchaudio's, which is not the case for Demucs.

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,115 @@ def test_decoder_output_shape_time(self):
...
+    2. The same model can be recreated without demucs.
+    """
+
+    def _get_original_model(self, sources: List[str], nfft: int, depth: int):
+        from demucs import hdemucs as original
+
+        old_model = original.HDemucs(sources, nfft=nfft, depth=depth)
```
mthrok, line 133: `original` sounds more appropriate.

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,115 @@ def test_decoder_output_shape_time(self):
...
+        nfft = 1024
+        depth = 5
+        duration = 10
+        channels = 2
+        batch_size = 1
+
+        torch.random.manual_seed(0)
+        factory_hdemucs = hdemucs_low(sources, sample_rate=sample_rate).eval()
+        torch.random.manual_seed(0)
+        model_hdemucs = HDemucs(sources, nfft=nfft, depth=depth, sample_rate=sample_rate).eval()
```
mthrok, line 155: I feel `HDemucs` is redundant, the factory function should cover this.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -929,3 +929,48 @@ def _ispectro(z: torch.Tensor, hop_length: int = 0, length: int = 0, pad: int =
     _, length = x.shape
     other.append(length)
     return x.view(other)
+
+
+def hdemucs_low(sources: List[str], sample_rate: int) -> HDemucs:
+    r"""Builds low nfft version of HDemucs model.
```
mthrok, line 935: `low nfft version of` is not quite helpful. Can we either have more detailed description or say something like "the version referred to X in the paper".

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -1,18 +1,28 @@
...
-    return HDemucs(sources)
+def _get_hdemucs_model(sources: List[str], n_fft: int = 4096, depth: int = 6, sample_rate: int = 44100):
+    return HDemucs(sources, nfft=n_fft, depth=depth, sample_rate=sample_rate)
 
 
-class HDemucsTests(TestBaseMixin):
-    def _get_inputs(self, duration: int, channels: int, batch_size: int, sample_rate: int):
-        sample = torch.rand(batch_size, channels, duration * sample_rate, dtype=torch.float32, device=self.device)
-        return sample
+def _get_inputs(duration: int, channels: int, batch_size: int, sample_rate: int):
```
carolineechen, line 20: we should also include device here

carolineechen (commented):
you will need to modify the `install.sh` scripts under `.circleci/unittest` to install demucs and be able to use it for circle tests

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,88 @@ def test_decoder_output_shape_time(self):
...
+
+    def _get_original_model(self, sources: List[str], nfft: int, depth: int):
+        from demucs import hdemucs as original
+
+        original = original.HDemucs(sources, nfft=nfft, depth=depth)
+        return original
+
+    @parameterized.expand(
+        [
+            (["bass", "drums", "other", "vocals"],),
```
carolineechen, line 137: this can also be moved into a single function/var (see [example](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/models/wav2vec2/fairseq_integration_test.py#L44)), since several tests reuse it

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,88 @@ def test_decoder_output_shape_time(self):
...
+        depth = 6
+        duration = 10
+        channels = 2
+        batch_size = 1
+
+        torch.random.manual_seed(0)
+        factory_hdemucs = hdemucs_high(sources, sample_rate=sample_rate).eval()
+        self._assert_equal_models(batch_size, channels, depth, duration, factory_hdemucs, nfft, sample_rate, sources)
+
+    def _assert_equal_models(self, batch_size, channels, depth, duration, factory_hdemucs, nfft, sample_rate, sources):
```
carolineechen, line 195: can you move helper functions to the top of the class, and move `factory_hdemucs` to the first arg rather than have it in the middle of model params

torchaudio/prototype/models/hdemucs.py
```diff
@@ -929,3 +929,52 @@ def _ispectro(z: torch.Tensor, hop_length: int = 0, length: int = 0, pad: int =
     _, length = x.shape
     other.append(length)
     return x.view(other)
+
+
+def hdemucs_low(sources: List[str], sample_rate: int) -> HDemucs:
```
carolineechen, line 934: remember to also declare these functions to `torchaudio/prototype/models/__init__.py`

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,88 @@ def test_decoder_output_shape_time(self):
...
+            (["bass", "drums", "other"],),
+            (["bass", "vocals"],),
+            (["vocals"],),
+        ]
+    )
+    def test_import_recreate_low_model_test(self, sources):
+        sample_rate = 8000
+        nfft = 1024
+        depth = 5
+        duration = 10
```
carolineechen, line 147: `duration`, `channels`, and `batch_size` are reused across tests -- what about setting these values as default params in `_get_inputs` and avoid redefining them in test cases

torchaudio/prototype/models/hdemucs.py
```diff
@@ -929,3 +929,55 @@ def _ispectro(z: torch.Tensor, hop_length: int = 0, length: int = 0, pad: int =
...
+    """
+
+    return HDemucs(sources=sources, nfft=1024, depth=5, sample_rate=sample_rate)
+
+
+def hdemucs_medium(sources: List[str], sample_rate: int) -> HDemucs:
+    r"""Builds medium nfft (2048) version of HDemucs model. This version is suitable for medium sample rates,and bundles
+    parameters together to call valid nfft and depth values for a model structured for sample rates around 16-32 kHZ
+
+    Medium HDemucs has not been tested against the original Hybrid Demucs as there have been slight adjustments in
```
carolineechen, line 954: Have this in "Note" format

```suggestion
    .. note::
        Medium HDemucs has not been tested against the original Hybrid Demucs as there have been slight adjustments in
```

carolineechen, line 954: has it not been tested or does it definitely produce different results? in the second case, might be better to explicitly state that. Also "slight adjustments" is vague

"Does not reach numerical parity with original HDemucs [link repo] because of  ..."

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,46 @@ def test_decoder_output_shape_time(self):
...
         assert z.size() == (batch_size, chout, t * stride)
         assert y.size() == (batch_size, chin, t)
+
+
+@skipIfNoModule("demucs")
+class CompareHDemucsOriginal(TorchaudioTestCase):
+    """Test the process of importing the models from demucs.
+
+    Test methods in this test suite will check to
+    1. Assure correctness in factory functions, comparing with original hybrid demucs
```
carolineechen, line 126:
```suggestion
    Test methods in this test suite will check to assure correctness in factory functions, comparing with original hybrid demucs
```

torchaudio/prototype/models/hdemucs.py
carolineechen: oh nvm I remember these parameters fail to run on original demucs

carolineechen: and as before, could you build docs and include a screenshot in your PR desc? as the build_docs job is still unfortunately broken

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
carolineechen: Since you still need to define and use these variables in the shape tests, I think the better approach would still be letting `_get_inputs` take in these parameters, but with a default value.

```
def _get_inputs(
    sample_rate: int,
    device: torch.device,
    batch: int = 1,
    duration: int = 10,
    channels: int = 2
)
```

Then, for the shape tests, since you declare the variables in the test, you can pass them into `_get_inputs` for consistency. But for the `import_recreate_xxx_tests`, since they are not explicitly defined and used in the tests, we can leave those tests as is to use the default values.

test/torchaudio_unittest/prototype/hdemucs_test_impl.py
```diff
@@ -106,3 +116,46 @@ def test_decoder_output_shape_time(self):
...
+    def _assert_equal_models(self, factory_hdemucs, depth, nfft, sample_rate, sources):
+        torch.random.manual_seed(0)
+        original_hdemucs = self._get_original_model(sources, nfft, depth).to(self.device).eval()
+        inputs = _get_inputs(sample_rate=sample_rate, device=self.device)
+        factory_output = factory_hdemucs(inputs)
+        original_output = original_hdemucs(inputs)
+        self.assertEqual(original_output, factory_output)
+
+    @SOURCES_OUTPUT_CONFIG
+    def test_import_recreate_low_model_test(self, sources):
```
carolineechen, line 144: nit (same with for high model), two tests in function name is a little redundant

```suggestion
    def test_import_recreate_low_model(self, sources):
```

carolineechen (approved):
looks good!

torchaudio/prototype/models/hdemucs.py
```diff
@@ -929,3 +929,57 @@ def _ispectro(z: torch.Tensor, hop_length: int = 0, length: int = 0, pad: int =
...
+
+    Returns:
+        HDemucs:
+            HDemucs model.
+    """
+
+    return HDemucs(sources=sources, nfft=2048, depth=6, sample_rate=sample_rate)
+
+
+def hdemucs_high(sources: List[str], sample_rate: int) -> HDemucs:
```
adefossez, line 971: how is this going to work with pretrained models ? is this going to be a separate function or an extra argument here ?

skim0514: Pretrained models will work with our pipeline! Here's a link to the initial bundle, my part is waiting rn. #2440

PR #2548, opened 2022-07-14
[Nova] Added draft calling GHA workflow for building linux wheels

osalpekar:
As part of Project Nova, we are consolidating CI/CD workflows and infra, making them reusable across PyTorch ecosystem libraries. https://github.com/pytorch/test-infra/pull/460 introduces a general-purpose reusable workflow to build linux wheels for python libraries. This PR introduces a caller workflow that triggers the reusable workflow. Details around modular env setup, passing input args across workflows, etc. are still being worked out.

Using reusable workflow defined in https://github.com/pytorch/test-infra/pull/506


.github/workflows/build_wheels_linux.yml
```diff
@@ -0,0 +1,8 @@
+name: Build TorchAudio Linux Wheels
+
+on:
+  workflow_dispatch
+
+jobs:
+  run-reusable-build-workflow:
+    users: pytorch/test-infra/.github/workflows/build_wheels_linux_reusable.yml@main
```
weiwangmeta, line 8: Is "users" a typo? Instead, it should be "uses"?

osalpekar:
Shall we merge what we have so far (after removing the on: PR trigger in the calling workflow)? That way we can unblock Wei to work on completing the build matrix without stepping on the other in-progress changes. Also so everybody owns their own PR's :)

weiwangmeta:
@mehtanirav is testing py37 and py38 wheels, once it is done, I will add the remaining ~22 test cases.

mehtanirav:
> Shall we merge what we have so far (after removing the on: PR trigger in the calling workflow)? That way we can unblock Wei to work on completing the build matrix without stepping on the other in-progress changes. Also so everybody owns their own PR's :)

Good point. I will try to close this PR today, and unblock Wei from his testing.

osalpekar (commented):
Changes look great to me!

mehtanirav (approved):

Issue #2104, opened 2021-12-26
`torch.audio.functional.spectrogram` does not use `normalized` parameter correctly


jojonki:
### 🐛 Describe the bug

I found `torchaudio.Spectrogram` does not use `normalized` argument correctly because the code uses hard code. Always it sets to `False`. Similar to #1007, but not griffin-lim in this case.

https://github.com/pytorch/audio/blob/main/torchaudio/functional/functional.py#L110
```python
def spectrogram(
    waveform: Tensor,
    pad: int,
    window: Tensor,
    n_fft: int,
    hop_length: int,
    win_length: int,
    power: Optional[float],
    normalized: bool,
    center: bool = True,
    pad_mode: str = "reflect",
    onesided: bool = True,
    return_complex: Optional[bool] = None,
) -> Tensor:
...
    spec_f = torch.stft(
        input=waveform,
        n_fft=n_fft,
        hop_length=hop_length,
        win_length=win_length,
        window=window,
        center=center,
        pad_mode=pad_mode,
        normalized=False, # HERE!!!
        onesided=onesided,
        return_complex=True,
    )
```

### Versions

```
Collecting environment information...
PyTorch version: 1.10.1
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 12.0.1 (arm64)
GCC version: Could not collect
Clang version: 13.0.0 (clang-1300.0.29.3)
CMake version: Could not collect
Libc version: N/A

Python version: 3.9.7 (default, Nov 14 2021, 14:30:16)  [Clang 13.0.0 (clang-1300.0.29.3)] (64-bit runtime)
Python platform: macOS-12.0.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.20.3
[pip3] torch==1.10.1
[pip3] torchaudio==0.10.1
[conda] Could not collect
```

mthrok:
Hi @jojonki 

`normalized` parameter is not passed to stft function, but used in couple of lines later.

https://github.com/pytorch/audio/blob/9397cd5b9c4389baa8b3891b3ba12aa2ae07100f/torchaudio/functional/functional.py#L117-L119

Is this not serving your purpose?

[...]

PR #2554, opened 2022-07-20
Add Union normalization parameter on spectrogram and inverse spectrogram

skim0514:
Add str to normalized parameter to enable frame_length based normalization to align with torch implementation of stft. Addresses issue #2104

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -245,6 +245,76 @@ class Functional(TestBaseMixin):
         spec.sum().backward()
         assert not x.grad.isnan().sum()
 
+    @parameterized.expand(
+        [
+            (1024,),
+            (2048,),
+            (4096,),
+        ]
+    )
+    def test_spectrogram_normalization_hann_window(self, nfft):
+        """This test assumes that currently, torch.stft and the existing math behind spectrogram has been correct.
+        The test is checking that in relation to one another, the normalization factors correctly align based on
+        mathematical prediction. Using spec_false as a base, which has no normalization factors, we check to see that
+        turning normalized as ``True`` or ``"window"`` will have a normalization factor of sqrt(3*nfft/8) given a
+        hann_window. The sum of squares of hann has been calculated to be 3/8 of frame_length, so we are assuming this.
+        Next, when normalized ``"frame_length"`` we are using the torch stft version of normalization, therefore we
+        assume that it correctly is normalized by a factor of sqrt(nfft). This test does not test the accuracy of
+        spectrogram, but is testing the relative factors of normalization and that they align upon the frame_length
+        and chosen normalize parameter.
+        """
+        x = torch.rand(1, 22050)
+        spec_false = F.spectrogram(
+            x,
+            pad=0,
+            window=torch.hann_window(nfft, device=x.device, dtype=x.dtype),
+            n_fft=nfft,
+            hop_length=4,
+            win_length=nfft,
+            power=None,
+            normalized=False,
+        )
+
+        spec_true = F.spectrogram(
+            x,
+            pad=0,
+            window=torch.hann_window(nfft, device=x.device, dtype=x.dtype),
+            n_fft=nfft,
+            hop_length=4,
+            win_length=nfft,
+            power=None,
+            normalized=True,
+        )
+
+        spec_window = F.spectrogram(
+            x,
+            pad=0,
+            window=torch.hann_window(nfft, device=x.device, dtype=x.dtype),
+            n_fft=nfft,
+            hop_length=4,
+            win_length=nfft,
+            power=None,
+            normalized="window",
+        )
+
+        spec_frame = F.spectrogram(
+            x,
+            pad=0,
+            window=torch.hann_window(nfft, device=x.device, dtype=x.dtype),
+            n_fft=nfft,
+            hop_length=4,
+            win_length=nfft,
+            power=None,
+            normalized="frame_length",
+        )
+
+        norm_factor = math.sqrt(3 * nfft / 8)
+        frame_norm_factor = math.sqrt(nfft)
+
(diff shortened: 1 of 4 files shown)


skim0514:
cc @adefossez

torchaudio/functional/functional.py
```diff
@@ -176,10 +188,20 @@ def inverse_spectrogram(
         Tensor: Dimension `(..., time)`. Least squares estimation of the original signal.
     """
 
+    if type(normalized) is str and normalized not in ["frame_length", "window"]:
```
mthrok, line 191: To continue supporting TorchScript, need to perform type refinement.
See example like;

https://github.com/pytorch/text/blob/ca2e5a4601dcaf08bcb7171a3feebb2ecd86d230/torchtext/functional.py#L26-L38

torchaudio/functional/functional.py
```diff
@@ -77,7 +77,8 @@ def spectrogram(
         power (float or None): Exponent for the magnitude spectrogram,
             (must be > 0) e.g., 1 for energy, 2 for power, etc.
             If None, then the complex spectrum is returned instead.
-        normalized (bool): Whether to normalize by magnitude after stft
+        normalized (bool or str): Whether to normalize by magnitude after stft. If input is str, choices are
```
mthrok, line 80: Can you update the docstrings of corresponding transforms?

nateanl: Also it's better mention if `normalized` is of `bool` type and is `True`, what's the mapped `str` option for it.

torchaudio/functional/functional.py
```diff
@@ -162,7 +173,8 @@ def inverse_spectrogram(
         n_fft (int): Size of FFT
         hop_length (int): Length of hop between STFT windows
         win_length (int): Window size
-        normalized (bool): Whether the stft output was normalized by magnitude
+        normalized (bool or str): Whether the stft output was normalized by magnitude. If input is str, choices are
+            ``"window"`` and ``"frame_length"``, dependent on normalization mode.
```
nateanl, line 177: same here, describe the mapping between `bool` and `str`.

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -245,6 +245,76 @@ def test_spectrogram_grad_at_zero(self, power):
...
+        """This test assumes that currently, torch.stft and the existing math behind spectrogram has been correct.
+        The test is checking that in relation to one another, the normalization factors correctly align based on
+        mathematical prediction. Using spec_false as a base, which has no normalization factors, we check to see that
+        turning normalized as ``True`` or ``"window"`` will have a normalization factor of sqrt(3*nfft/8) given a
+        hann_window. The sum of squares of hann has been calculated to be 3/8 of frame_length, so we are assuming this.
+        Next, when normalized ``"frame_length"`` we are using the torch stft version of normalization, therefore we
+        assume that it correctly is normalized by a factor of sqrt(nfft). This test does not test the accuracy of
+        spectrogram, but is testing the relative factors of normalization and that they align upon the frame_length
+        and chosen normalize parameter.
+        """
```
carolineechen, line 265: can you also add the link to the github issue that brought up this problem here, provides some context on why we add this test

carolineechen:
can you also rebase onto main for build_docs fix, or run it locally and include a screenshot?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -245,6 +245,77 @@ def test_spectrogram_grad_at_zero(self, power):
...
 
+    @parameterized.expand(
+        [
+            (1024,),
+            (2048,),
+            (4096,),
+        ]
+    )
+    def test_spectrogram_normalization_hann_window(self, nfft):
+        """This test assumes that currently, torch.stft and the existing math behind spectrogram has been correct.
```
nateanl, line 256:
```suggestion
        """This test assumes that currently, torch.stft and the existing math behind spectrogram are correct.
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -245,6 +245,77 @@ def test_spectrogram_grad_at_zero(self, power):
...
+        The test is checking that in relation to one another, the normalization factors correctly align based on
+        mathematical prediction. Using spec_false as a base, which has no normalization factors, we check to see that
+        turning normalized as ``True`` or ``"window"`` will have a normalization factor of sqrt(3*nfft/8) given a
+        hann_window. The sum of squares of hann has been calculated to be 3/8 of frame_length, so we are assuming this.
```
nateanl, line 260:
```suggestion
        turning normalized as ``True`` or ``"window"`` will have a normalization factor of the sum of squares of hann
        window, which is calculated to be sqrt(3 * nfft / 8).
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -245,6 +245,77 @@ def test_spectrogram_grad_at_zero(self, power):
...
+        Next, when normalized ``"frame_length"`` we are using the torch stft version of normalization, therefore we
```
nateanl, line 261:
```suggestion
        Next, when ``normalized`` is ``"frame_length"``, we are using the normalization in torch.stft, therefore we
```

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -245,6 +245,77 @@ def test_spectrogram_grad_at_zero(self, power):
...
+        assume that it correctly is normalized by a factor of sqrt(nfft). This test does not test the accuracy of
```
nateanl, line 262:
```suggestion
        assume that it is correctly normalized by a factor of sqrt(nfft). This test does not test the accuracy of
```

carolineechen (approved):
thanks for driving this! can you address zhaoheng's suggested change as well?

PR #2560, opened 2022-07-21
Fix fall back failure in sox_io backend

jumon:
Fix the fallback function of load fileobj function in sox_io backend.

The typo in the fallback function prevents showing the intended error message.


mthrok (approved):

mthrok:
Hi @jumon  - Thanks for fixing the bug.

PR #2563, opened 2022-07-21
Add dimension and shape check

skim0514:
Don't allow users to input incorrect dimensions

### DIFF
--- a/torchaudio/prototype/models/hdemucs.py
+++ b/torchaudio/prototype/models/hdemucs.py
@@ -516,8 +516,14 @@ class HDemucs(torch.nn.Module):
                 output tensor split into sources of shape `(batch_size, num_sources, channel, num_frames)`
         """
 
-        if input.ndim != 3 or input.shape[1] != self.audio_channels:
+        if input.ndim != 3:
             raise ValueError(f"Expected 3D tensor (batch, channel, frames). Found: {input.shape}")
+
+        if input.shape[1] != self.audio_channels:
+            raise ValueError(
+                f"The channel dimension of input Tensor must match `audio_channels` of HDemucs model. "
+                f"Found:{input.shape[1]}"
+            )
 
         x = input
         length = x.shape[-1]
--- a/torchaudio/prototype/models/hdemucs.py
+++ b/torchaudio/prototype/models/hdemucs.py
@@ -516,8 +516,8 @@ class HDemucs(torch.nn.Module):
                 output tensor split into sources of shape `(batch_size, num_sources, channel, num_frames)`
         """
 
-        if len(input.shape) == 2:
-            input = input.view(1, self.audio_channels, -1)
+        if input.ndim != 3 or input.shape[1] != self.audio_channels:
+            raise ValueError(f"Expected 3D tensor (batch, channel, frames). Found: {input.shape}")
 
         x = input
         length = x.shape[-1]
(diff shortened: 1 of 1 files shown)


torchaudio/prototype/models/hdemucs.py
```diff
@@ -516,8 +516,8 @@ def forward(self, input: torch.Tensor):
                 output tensor split into sources of shape `(batch_size, num_sources, channel, num_frames)`
         """
 
-        if len(input.shape) == 2:
-            input = input.view(1, self.audio_channels, -1)
+        if input.ndim != 3 or input.shape[1] != self.audio_channels:
+            raise ValueError(f"Expected 3D tensor (batch, channel, frames). Found: {input.shape}")
```
nateanl, line 520: The assertion can be more explicit on the value check of the channel dimension. In case where users do have 3D tensors but use a different channel number than required by the model, we can inform users what is the correct value for channel dimension.
```suggestion
        assert input.ndim == 3, f"Expected 3D tensor with dimensions (batch, channel, time). Found: {input.shape}."
        assert input.shape[1] == self.audio_channels, f"The channel dimension of input Tensor must match `audio_channels` of HDemucs model. Found {input.shape[1]}."
```

mthrok: Can we keep `raise`? `ValueError` seems more suited than `AssertionError`.

mthrok: > The [statement assert](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) can be used for checking conditions at runtime, but is removed if optimizations are requested from Python. The extended form is:

https://stackoverflow.com/a/40183030/3670924

nateanl: > The [statement assert](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) can be used for checking conditions at runtime, but is removed if optimizations are requested from Python. The extended form is:
https://stackoverflow.com/a/40183030/3670924

sure, LGTM

torchaudio/prototype/models/hdemucs.py
```diff
@@ -516,8 +516,14 @@ def forward(self, input: torch.Tensor):
                 output tensor split into sources of shape `(batch_size, num_sources, channel, num_frames)`
         """
 
-        if len(input.shape) == 2:
-            input = input.view(1, self.audio_channels, -1)
+        if input.ndim != 3:
+            raise ValueError(f"Expected 3D tensor (batch, channel, frames). Found: {input.shape}")
```
nateanl, line 520:
```suggestion
            raise ValueError(f"Expected 3D tensor with dimensions (batch, channel, frames). Found: {input.shape}.")
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -516,8 +516,14 @@ def forward(self, input: torch.Tensor):
...
+
+        if input.shape[1] != self.audio_channels:
+            raise ValueError(
+                f"The channel dimension of input Tensor must match `audio_channels` of HDemucs model. "
+                f"Found:{input.shape[1]}"
```
nateanl, line 525:
```suggestion
                f"Found:{input.shape[1]}."
```

nateanl (approved):
LGTM. just some nits.

carolineechen (approved):
looks good!

PR #2570, opened 2022-07-25
Change docstring for easier understanding

skim0514:
Edit factory function's docstrings.

### DIFF
--- a/torchaudio/prototype/models/hdemucs.py
+++ b/torchaudio/prototype/models/hdemucs.py
@@ -329,8 +329,6 @@ class HDemucs(torch.nn.Module):
         dconv_attn (int, optional): adds attention layers in DConv branch starting at this layer. (Default: 4)
         dconv_lstm (int, optional): adds a LSTM layer in DConv branch starting at this layer. (Default: 4)
         dconv_init (float, optional): initial scale for the DConv branch LayerScale. (Default: 1e-4)
-        sample_rate (int, optional): sample rate, serving as metadata not actually used (Default: 44100)
-        segment (int, optional): segment size (Default: 40)
     """
 
     def __init__(
@@ -356,8 +354,6 @@ class HDemucs(torch.nn.Module):
         dconv_attn: int = 4,
         dconv_lstm: int = 4,
         dconv_init: float = 1e-4,
-        sample_rate: int = 44100,
-        segment: int = 4 * 10,
     ):
         super().__init__()
         self.depth = depth
@@ -368,8 +364,6 @@ class HDemucs(torch.nn.Module):
         self.context = context
         self.stride = stride
         self.channels = channels
-        self.sample_rate = sample_rate
-        self.segment = segment
 
         self.hop_length = self.nfft // 4
         self.freq_emb = None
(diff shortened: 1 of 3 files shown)


torchaudio/prototype/models/hdemucs.py
```diff
@@ -942,8 +943,10 @@ def hdemucs_low(sources: List[str], sample_rate: int) -> HDemucs:
     parameters together to call valid nfft and depth values for a model structured for sample rates around 8 kHZ.
 
     Args:
-        sources (List[str]): Sources to use for audio split
-        sample_rate (int): Serves as metadata, recommend lower sample rates.
+        sources (List[str]): Sources to use for audio split. Input a list of strings out that can contain the following
+            options: [``"bass"``, ``"drums"``, ``"other"``, ``"mixture"``, ``"vocals"``].
+        sample_rate (int): Intended usage to serve as metadata. Saved alongside parameters and enables sample rate to be
```
mthrok, line 948: I think it is more intuitive to start with "what this is" and then continue to "intended usage". Also it should clarify that this has no effect on the model behavior.

```
Expected sample rate of model input. The value of this argument does not affect model behavior. It serves as metadata, which is saved alongside parameters, making it easy to retrieve later.
```

BTW Is this the behavior from the original Demucs repo? We use bundles for this specific purpose to split model from metadata.

torchaudio/prototype/models/hdemucs.py
```diff
@@ -298,7 +298,8 @@ class HDemucs(torch.nn.Module):
     Hybrid Demucs model from *Hybrid Spectrogram and Waveform Source Separation* [:footcite:`defossez2021hybrid`].
 
     Args:
-        sources (List[str]): list of source names.
+        sources (List[str]): list of source names. Input a list of strings out that can contain the following
```
carolineechen, line 301:
```suggestion
        sources (List[str]): list of source names. List can contain the following source
```

torchaudio/prototype/models/hdemucs.py
```diff
@@ -937,23 +942,23 @@ def _ispectro(z: torch.Tensor, hop_length: int = 0, length: int = 0, pad: int =
...
 
-def hdemucs_low(sources: List[str], sample_rate: int) -> HDemucs:
+def hdemucs_low(sources: List[str]) -> HDemucs:
     r"""Builds low nfft (1024) version of HDemucs model. This version is suitable for lower sample rates, and bundles
     parameters together to call valid nfft and depth values for a model structured for sample rates around 8 kHZ.
 
     Args:
-        sources (List[str]): Sources to use for audio split
-        sample_rate (int): Serves as metadata, recommend lower sample rates.
+        sources (List[str]): Sources to use for audio split. Input a list of strings out that can contain the following
```
carolineechen, line 950: To avoid redefining all the definitions, you can also refer users to `HDemucs`, like it is done [here](https://github.com/pytorch/audio/blob/main/torchaudio/models/wav2vec2/model.py#L402). This way there is less repetition and potential future docstring changes only need to be made in one place rather than 4

torchaudio/prototype/models/hdemucs.py
```diff
@@ -498,6 +493,16 @@ def _ispec(self, z, length=None):
         x = x[..., pad : pad + length]
         return x
 
+    def _pad1d(self, x: torch.Tensor, paddings: tp.Tuple[int, int], mode: str = "zero", value: float = 0.0):
```
carolineechen, line 496: suggest converting `paddings` into two separate variables (`pad_left`, `pad_right`)

torchaudio/prototype/models/hdemucs.py
```diff
@@ -498,6 +493,16 @@ def _ispec(self, z, length=None):
...
+        """Wrapper around F.pad, in order for reflect padding to not break. Add extra zero padding around"""
```
carolineechen, line 497: can you add a little more detail here? in which scenario does it break and how does this resolve it

carolineechen (approved):

PR #2580, opened 2022-07-27
Migrate CTC decoder code

mthrok:
This commit gets rid of our copy of CTC decoder code and
replace it with upstream Flashlight-Text repo.


torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -28,21 +28,23 @@
...
-        _LexiconDecoder,
-        _LexiconDecoderOptions,
-        _LexiconFreeDecoder,
-        _LexiconFreeDecoderOptions,
-        _LM,
-        _load_words,
-        _SmearingMode,
-        _Trie,
-        _ZeroLM,
+    torchaudio._extension._load_lib("libflashlight-text")
```
carolineechen, line 31: would fbcode be able to directly use this version as well or do we need to keep the separate branch above?

jacobkahn: +1

jacobkahn (approved):
LGTM! Glad we got to this point 🎉

torchaudio/models/decoder/_ctc_decoder.py
mthrok: We need to keep the branch.

In fbcode, flashlight-text and torchaudio are logically separated during the whole lifecycle, from source to deployment package. So it is not possible for torchaudio to have decoder code inside of torchaudio package. Instead, flashlight-text is always available, so the decoder code is imported from upstream flashlight packaget.

mthrok: BTW in OSS, the situation is different. I think there are main three audience we need to think;

1. users who only use torchaudio
2. users who use flashlight-text and torchaudio
3. users who develops flashlight-text and torchaudio.

I think the majority of torchaudio users are in category 1. and they do not import the native flashlight-text, they will use the one torchaudio ships. This is straightforward.
The users in category 2, if torchaudio attempts to import the version it ships, then importing both flashlight-text and torchaudio become impossible due to duplicated symbols in both binaries. This is why I put torchaudio's version in lower priority, and attempt to import flashlight first.
Those in category 3, if there is a sudden backward-incompatible change, (say LM attribute from flashlight binding becomes unavailable in favor of some new attribute) then using torchaudio with flashlight become impossible. This can be only prevented by flashlight-text provides soft BC-breaking change and migration period. But Jacob will provide guidance if this ever happens, so it should be fine.

third_party/flashlight-text/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
+# Custom CMakeLists for building flashlight-text decoder
+#
+# The main difference from upstream native CMakeLists from flashlight-text.
+#
+# 1. Build compression libraries statically and make KenLM self-contained
```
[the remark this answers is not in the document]
mthrok: I am open to upstream the features described here, but until at least 1, 2, 4 and 5 are upstreamed, we cannot get rid of this custom logic.

Regardless, I think properly upstreaming 1, 2 and 3 will make the build process of libflashlight-text simpler.
I am guessing that `FL_TEXT_BUILD_STANDALONE` is intended to make libflashlight-text self-contained, which is convenient for torchaudio as-well.

third_party/kenlm/CMakeLists.txt
```diff
@@ -1,45 +1,45 @@
...
-  submodule/lm/vocab.cc
+  kenlm/lm/bhiksha.cc
+  kenlm/lm/binary_format.cc
+  kenlm/lm/config.cc
+  kenlm/lm/lm_exception.cc
+  kenlm/lm/model.cc
+  kenlm/lm/quantize.cc
+  kenlm/lm/read_arpa.cc
+  kenlm/lm/search_hashed.cc
+  kenlm/lm/search_trie.cc
```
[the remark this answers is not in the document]
mthrok: Yeah, that will make the integration simpler, but once this PR is landed, it would not matter much until the next time we refactor the build process.

carolineechen (approved):

carolineechen:
thanks!

PR #2600, opened 2022-08-03
Added example for SlidingWindowCmn transform

ravimakhija:
Added example for `SlidingWindowCmn` transform as mentioned in issue #1564


carolineechen (approved):
thanks for the contribution!

PR #2602, opened 2022-08-04
Add convolution operator

hwangjeff:
Adds functions `convolve` and `fftconvolve`, which compute the convolution of two tensors along their trailing dimension. The former performs the convolution directly, whereas the latter performs it using FFT.


torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,42 @@
...
+        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(*, N + M - 1)`, where
+        the leading dimensions match those of ``x``.
+
+    .. _convolution:
+        https://en.wikipedia.org/wiki/Convolution
+    """
+    if use_fft:
+        n = x.size(-1) + y.size(-1) - 1
+        fresult = torch.fft.rfft(x, n=n) * torch.fft.rfft(y, n=n)
+        output = torch.fft.irfft(fresult, n=n)
```
nateanl, line 28: When `x` and `y` are both of dtype `int` or `long`, using FFT solution will change the dtype to `float` or `double`. Could you cast the output to be the same dtype as the input?
```suggestion
        output = torch.fft.irfft(fresult, n=n).to(x.dtype)
```

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,42 @@
...
+        use_fft (bool, optional): If ``True``, then convolve ``x`` and ``y`` using FFT.
+            Otherwise, convolve ``x`` and ``y`` directly via sums. (Default: ``False``)
+
+    Returns:
+        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(*, N + M - 1)`, where
+        the leading dimensions match those of ``x``.
+
+    .. _convolution:
+        https://en.wikipedia.org/wiki/Convolution
+    """
```
nateanl, line 24: the leading dimensions of `x` and `y` need to be checked before applying convolution.

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,42 @@
...
+    else:
+        num_signals = math.prod(dim for dim in x.shape[:-1])
+        reshaped_x = x.reshape((num_signals, x.size(-1)))
+        reshaped_y = y.reshape((num_signals, y.size(-1)))
+        output = torch.nn.functional.conv1d(
+            input=reshaped_x,
+            weight=reshaped_y.flip(-1).unsqueeze(1),
+            stride=1,
+            groups=reshaped_x.size(0),
+            padding=reshaped_y.size(-1) - 1,
```
nateanl, line 38: if the last dimension size of `x` is shorter than that of `y`, will there be performance difference if we swap `x` and `y`?

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,42 @@
+import math
+
+import torch
+
+
+def convolve(x: torch.Tensor, y: torch.Tensor, use_fft: bool = False) -> torch.Tensor:
```
mthrok, line 6: I feel like FFT-version can be a stand alone function. WDYT?

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,42 @@
...
+    """
+    Convolves inputs along last dimension. Note that, in contrast to ``torch.nn.functional.conv1d``, this
+    function applies the true `convolution`_ operator.
```
mthrok, line 9: Maybe clarifying that `torch.nn.functional.conv1d` is correlation?

torchaudio/prototype/functional/functional.py
hwangjeff: split it out into two functions

torchaudio/prototype/functional/functional.py
hwangjeff: i feel like it makes more sense to keep the output dtype as float and communicate that to users (scipy does this)

torchaudio/prototype/functional/functional.py
hwangjeff: asymptotically, no

hwangjeff: actually, turns out flipping in pytorch does a copy, so it can make a considerable difference — fixed

torchaudio/prototype/functional/functional.py
mthrok: I think automatically converting float into int is dangerous. It will lose many precision. It can be performed in client code if necessary.

nateanl: i see, sounds good

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,83 @@
+import torch
+
+
+def _check_convolve_inputs(x: torch.Tensor, y: torch.Tensor) -> None:
+    if x.shape[:-1] != y.shape[:-1]:
+        raise ValueError(f"Leading dimensions of x and y don't match (got {x.shape} and {y.shape}).")
+
+    if x.dim() < 2:
```
nateanl, line 8: wondering why the method doesn't support 1D Tensor?

hwangjeff: thanks — fixed

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+import torch
+import torchaudio.prototype.functional as F
+from scipy import signal
+from torchaudio_unittest.common_utils import nested_params, TestBaseMixin
+
+
+class FunctionalTestImpl(TestBaseMixin):
+    @nested_params(
+        [F.convolve, F.fftconvolve],
```
nateanl, line 10: shall we split the testing by comparing `F.convolve` with `scipy.signal.convolve(..., method="direct")`, and comparing `F.fftconvolve(..., method="fft")`?

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -0,0 +1,45 @@
...
+        x = torch.rand(*(leading_dims + (L_x,)), dtype=self.dtype, device=self.device)
+        y = torch.rand(*(leading_dims + (L_y,)), dtype=self.dtype, device=self.device)
+
+        actual = fn(x, y)
+
+        num_signals = torch.tensor(leading_dims).prod() if leading_dims else 1
+        x_reshaped = x.reshape((num_signals, L_x))
+        y_reshaped = y.reshape((num_signals, L_y))
+        expected = [
+            signal.convolve(x_reshaped[i].detach().cpu().numpy(), y_reshaped[i].detach().cpu().numpy())
```
nateanl, line 27: `scipy.signal.fftconvolve` can be applied on N-D arrays by setting `axes` to `[-1]`, so that we can optimize the testing speed.
To compare `F.convolve` with `scipy.signal.convolve` there is no `axes` argument, we have to use list iteration.

nateanl (approved):
some comments on the numerical unit testing, otherwise LGTM!

torchaudio/prototype/functional/functional.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def _check_convolve_inputs(x: torch.Tensor, y: torch.Tensor) -> None:
+    if x.shape[:-1] != y.shape[:-1]:
+        raise ValueError(f"Leading dimensions of x and y don't match (got {x.shape} and {y.shape}).")
+
+
+def fftconvolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
+    r"""
+    Convolves inputs along their last dimension using FFT.
```
mthrok, line 11: Should we give guidance when to use which?

mthrok (approved):

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -0,0 +1,65 @@
...
+        x = torch.rand(*(leading_dims + (L_x,)), dtype=self.dtype, device=self.device)
+        y = torch.rand(*(leading_dims + (L_y,)), dtype=self.dtype, device=self.device)
+
+        actual = F.fftconvolve(x, y)
+
+        num_signals = torch.tensor(leading_dims).prod() if leading_dims else 1
+        x_reshaped = x.reshape((num_signals, L_x))
+        y_reshaped = y.reshape((num_signals, L_y))
+        expected = signal.fftconvolve(x_reshaped.detach().cpu().numpy(), y_reshaped.detach().cpu().numpy(), axes=-1)
+        expected = torch.tensor(expected).reshape(leading_dims + (-1,))
```
nateanl, line 51:
```suggestion
        expected = signal.fftconvolve(x.detach().cpu().numpy(), y.detach().cpu().numpy(), axes=-1)
```

PR #2608, opened 2022-08-09
Add additive noise function

hwangjeff:
Adds function `add_noise`, which computes and returns the sum of a waveform and scaled noise.

### DIFF
--- a/torchaudio/prototype/functional/functional.py
+++ b/torchaudio/prototype/functional/functional.py
@@ -96,13 +96,14 @@ def add_noise(waveform: torch.Tensor, noise: torch.Tensor, lengths: torch.Tensor
     , with :math:`\text{SNR}_i` being the desired signal-to-noise ratio between :math:`x` and :math:`n_i`, in dB.
 
     Args:
-        waveform (torch.Tensor): (*, L)
-        noise (torch.Tensor): (*, N, L)
-        lengths (torch.Tensor): (*,); actual lengths of signals in `waveform`.
-        snr (torch.Tensor): (*, N); in dB.
+        waveform (torch.Tensor): Input waveform, with shape `(*, L)`.
+        noise (torch.Tensor): Noise, with shape `(*, N, L)` (leading dimensions must match those of ``waveform``).
+        lengths (torch.Tensor): Valid lengths of signals in `waveform`, with shape (*,).
+        snr (torch.Tensor): Signal-to-noise ratios in dB, with shape (*, N).
 
     Returns:
-        torch.Tensor: (*, length)
+        torch.Tensor: Result of scaling and adding ``noise`` to ``waveform``, with shape (*, L)
+        (same shape as ``waveform``).
     """
     # compute scale
     mask = torch.arange(0, waveform.size(-1)).expand(waveform.shape) < lengths.unsqueeze(-1)  # (*, L) < (*, 1) = (*, L)
--- a/torchaudio/prototype/functional/functional.py
+++ b/torchaudio/prototype/functional/functional.py
@@ -98,11 +98,11 @@ def add_noise(waveform: torch.Tensor, noise: torch.Tensor, lengths: torch.Tensor
     Args:
         waveform (torch.Tensor): Input waveform, with shape `(*, L)`.
         noise (torch.Tensor): Noise, with shape `(*, N, L)` (leading dimensions must match those of ``waveform``).
-        lengths (torch.Tensor): Valid lengths of signals in `waveform`, with shape (*,).
-        snr (torch.Tensor): Signal-to-noise ratios in dB, with shape (*, N).
+        lengths (torch.Tensor): Valid lengths of signals in `waveform` and `noise`, with shape `(*,)`.
+        snr (torch.Tensor): Signal-to-noise ratios in dB, with shape `(*, N)`.
 
     Returns:
-        torch.Tensor: Result of scaling and adding ``noise`` to ``waveform``, with shape (*, L)
+        torch.Tensor: Result of scaling and adding ``noise`` to ``waveform``, with shape `(*, L)`
         (same shape as ``waveform``).
     """
     # compute scale
--- a/torchaudio/prototype/functional/functional.py
+++ b/torchaudio/prototype/functional/functional.py
@@ -19,12 +19,12 @@ def fftconvolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
     .. properties:: Autograd TorchScript
 
     Args:
-        x (torch.Tensor): First convolution operand, with shape `(*, N)`.
-        y (torch.Tensor): Second convolution operand, with shape `(*, M)`
+        x (torch.Tensor): First convolution operand, with shape `(..., N)`.
+        y (torch.Tensor): Second convolution operand, with shape `(..., M)`
             (leading dimensions must match those of ``x``).
 
     Returns:
-        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(*, N + M - 1)`, where
+        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(..., N + M - 1)`, where
         the leading dimensions match those of ``x``.
 
     .. _convolution:
@@ -48,12 +48,12 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
     .. properties:: Autograd TorchScript
 
     Args:
-        x (torch.Tensor): First convolution operand, with shape `(*, N)`.
-        y (torch.Tensor): Second convolution operand, with shape `(*, M)`
+        x (torch.Tensor): First convolution operand, with shape `(..., N)`.
+        y (torch.Tensor): Second convolution operand, with shape `(..., M)`
             (leading dimensions must match those of ``x``).
 
     Returns:
-        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(*, N + M - 1)`, where
+        torch.Tensor: Result of convolving ``x`` and ``y``, with shape `(..., N + M - 1)`, where
         the leading dimensions match those of ``x``.
 
     .. _convolution:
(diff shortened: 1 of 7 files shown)


torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        noise (torch.Tensor): Noise, with shape `(*, N, L)` (leading dimensions must match those of ``waveform``).
+        lengths (torch.Tensor): Valid lengths of signals in `waveform` and `noise`, with shape `(*,)`.
+        snr (torch.Tensor): Signal-to-noise ratios in dB, with shape `(*, N)`.
+
+    Returns:
+        torch.Tensor: Result of scaling and adding ``noise`` to ``waveform``, with shape `(*, L)`
+        (same shape as ``waveform``).
+    """
+
+    input_leading_dims = (len(waveform.shape[:-1]), len(noise.shape[:-2]), len(lengths.shape), len(snr.shape[:-1]))
```
mthrok, line 112:
```suggestion
    input_leading_dims = (waveform.ndim-1, noise.ndim-2, lengths.ndim, snr.ndim-1)
```

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+    .. math::
+        a_i = \sqrt{ \frac{ ||x||_{2}^{2} }{ ||n_i||_{2}^{2} } \cdot 10^{-\frac{\text{SNR}_i}{10}} }
+
+    , with :math:`\text{SNR}_i` being the desired signal-to-noise ratio between :math:`x` and :math:`n_i`, in dB.
+
+    Note that this function broadcasts singleton leading dimensions in its inputs in a manner that is
+    consistent with the above formulae and PyTorch's broadcasting semantics.
+
+    Args:
+        waveform (torch.Tensor): Input waveform, with shape `(*, L)`.
```
mthrok, line 102: Ellipses seems more suitable than asterisk.

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        lengths (torch.Tensor): Valid lengths of signals in `waveform` and `noise`, with shape `(*,)`.
+        snr (torch.Tensor): Signal-to-noise ratios in dB, with shape `(*, N)`.
+
+    Returns:
+        torch.Tensor: Result of scaling and adding ``noise`` to ``waveform``, with shape `(*, L)`
+        (same shape as ``waveform``).
+    """
+
+    input_leading_dims = (len(waveform.shape[:-1]), len(noise.shape[:-2]), len(lengths.shape), len(snr.shape[:-1]))
+    if any([a != input_leading_dims[0] for a in input_leading_dims[1:]]):
```
mthrok, line 113: I think the following express the intention better. 
```
if not (waveform.ndim-1 == noise.ndim-2 == lengths.ndim == snr.ndim-1):
```

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+        raise ValueError(f"Input leading dimensions don't match (got {input_leading_dims}).")
+
+    if waveform.shape[-1] != noise.shape[-1]:
+        raise ValueError(f"Length dimensions of waveform and noise don't match (got {waveform[-1]} and {noise[-1]}).")
```
mthrok, line 117:
```suggestion
        raise ValueError(f"Length dimensions of waveform and noise don't match (got {waveform.shape[-1]} and {noise.shape[-1]}).")
```

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+
+    if noise.shape[-2] != snr.shape[-1]:
+        raise ValueError(f"Noise source dimensions of noise and snr don't match (got {noise[-2]} and {snr[-1]}).")
```
mthrok, line 120:
```suggestion
        raise ValueError(f"Noise source dimensions of noise and snr don't match (got {noise.shape[-2]} and {snr.shape[-1]}).")
```

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+
+    # compute scale
+    mask = torch.arange(0, waveform.size(-1)).expand(waveform.shape) < lengths.unsqueeze(-1)  # (*, L) < (*, 1) = (*, L)
```
mthrok, line 123: `waveform.size(-1)` is used many times, how about putting it in a constant?

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,61 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+    energy_noise = torch.linalg.vector_norm(noise * mask.unsqueeze(-2), ord=2, dim=-1) ** 2  # (*, N)
+    original_snr = energy_signal.unsqueeze(-1) / energy_noise  # (*, N)
+    snr_abs = 10 ** (snr / 10.0)  # (*, N)
+    scale = (original_snr / snr_abs).sqrt()  # (*, N)
+
+    # scale noise
+    scaled_noise = scale.unsqueeze(-1) * noise  # (*, N, 1) * (*, N, L) = (*, N, L)
+
+    # sum-reduce scaled noise
+    scaled_noise = scaled_noise.sum(-2)  # (*, L)
```
mthrok, line 134: I feel that letting client code do this will make this feature more flexible.
Not sure how common to add noises like this.

hwangjeff: a output of shape `(*, N, L)` that comprises individual sums of input waveform and scaled noise waveform would be difficult to reduce in client code — you'd need to do something like

`output.sum(-2) - (N - 1) * waveform`

to produce the current return value.

alternatively (or additionally), we could introduce a `scale_noise` function that simply returns noise scaled in accordance with the given waveform and snr

nateanl: computing `SNR` for each noise source then summing the noises is quite tricky, usually when we talk about `SNR`, it means the ratio of speech signal to all noises in the mixture. If users provide a `SNR` Tensor for 4 sources, for example. If the `SNR` values are [1, 2, -3, 0], after the mixing, none of the `SNR` values represent the real SNR of the mixture.
Shall we make a note in the method to let users be aware of it, or can we return the final `SNR` score for the mixture?

mthrok: > none of the SNR values represent the real SNR of the mixture.

This is a very a good point. I think at least we should have a function that returns the mixture with the specified SNR.
Other logics can be built on top of it.

hwangjeff: as it stands, the snr is defined per noise source, not on each waveform, so returning a "mixture with the specified snr" is not possible. and if it were instead defined on each waveform, a unique solution for the composition of noise sources to achieve the desired snr wouldn't exist

mthrok: How about making a core function take the form of `(waveform [N, L], noise [N, L], snr [N, ], length[N]) -> mixture [N, L]`, and build features with more elaborated logic using the core function?

hwangjeff: revised to eliminate hardcoded noise source dimension — users can account for multiple noise sources in one of the leading dimensions if they want

torchaudio/prototype/functional/functional.py
```diff
@@ -76,3 +76,56 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
+    # compute scale
+    mask = torch.arange(0, L).expand(waveform.shape) < lengths.unsqueeze(-1)  # (*, L) < (*, 1) = (*, L)
+    energy_signal = torch.linalg.vector_norm(waveform * mask, ord=2, dim=-1) ** 2  # (*,)
+    energy_noise = torch.linalg.vector_norm(noise * mask, ord=2, dim=-1) ** 2  # (*,)
+    original_snr = energy_signal / energy_noise  # (*,)
+    snr_abs = 10 ** (snr / 10.0)  # (*,)
+    scale = (original_snr / snr_abs).sqrt()  # (*,)
+
+    # scale noise
+    scaled_noise = scale.unsqueeze(-1) * noise  # (*, 1) * (*, L) = (*, L)
```
nateanl, line 129: in my experience, if `energy_noise` is very small, it may cause overflow. We can use log to avoid that,
```suggestion
    original_snr = 10 * torch.log10(energy_signal / energy_noise) # (*,)
    scale = 10 ** ((original_snr - snr) / 20)  # (*,)

    # scale noise
    scaled_noise = scale.unsqueeze(-1) * noise  # (*, 1) * (*, L) = (*, L)
```

torchaudio/prototype/functional/functional.py
```diff
@@ -19,12 +19,12 @@ def fftconvolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
     .. properties:: Autograd TorchScript
 
     Args:
-        x (torch.Tensor): First convolution operand, with shape `(*, N)`.
-        y (torch.Tensor): Second convolution operand, with shape `(*, M)`
+        x (torch.Tensor): First convolution operand, with shape `(..., N)`.
+        y (torch.Tensor): Second convolution operand, with shape `(..., M)`
```
nateanl, line 23: this seems related to `fftconvolve`, shall we make it in a separate PR?

torchaudio/prototype/functional/functional.py
```diff
@@ -48,12 +48,12 @@ def convolve(x: torch.Tensor, y: torch.Tensor) -> torch.Tensor:
...
```
nateanl, line 52: same here

torchaudio/prototype/functional/functional.py
hwangjeff: i revised to additionally compute `torch.log10(energy_signal) - torch.log10(energy_noise)` to avoid dividing by small values

torchaudio/prototype/functional/functional.py
hwangjeff: it is a very small change to a docstring in response to feedback on this pr — seems like overkill to create a separate pr for it

nateanl (approved):
LGTM!

Just a small nit, in the document page, there are leading commas after the equations. is it better to move it after the equation, or just remove it?
image

hwangjeff:
> LGTM!
> 
> Just a small nit, in the document page, there are leading commas after the equations. is it better to move it after the equation, or just remove it?

sure, moved the commas up to the equations in the latest commit

PR #2615, opened 2022-08-14
Added example for AmplitudeToDB transform

ravimakhija:
Added example for AmplitudeToDB transform as mentioned in issue #1564.

### DIFF
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -310,6 +310,11 @@ class AmplitudeToDB(torch.nn.Module):
             power being the elementwise square of the magnitude. (Default: ``'power'``)
         top_db (float or None, optional): minimum negative cut-off in decibels.  A reasonable
             number is 80. (Default: ``None``)
+
+    Example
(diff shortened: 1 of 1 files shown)


torchaudio/transforms/_transforms.py
```diff
@@ -310,6 +310,11 @@ class AmplitudeToDB(torch.nn.Module):
             power being the elementwise square of the magnitude. (Default: ``'power'``)
         top_db (float or None, optional): minimum negative cut-off in decibels.  A reasonable
             number is 80. (Default: ``None``)
+
+    Example
+        >>> waveform, sample_rate = torchaudio.load('test.wav', normalize=True)
```
nateanl, line 315: could you change the single quotes to double quotes, same as the melspectrogram PR?

nateanl (approved):
LGTM!

torchaudio/transforms/_transforms.py
ravimakhija: Fix is ready!

PR #2626, opened 2022-08-16
[Nova] Build Linux Conda Binaries using reusable workflow

osalpekar:
Calling the reusable workflow introduced in https://github.com/pytorch/test-infra/pull/546 to build conda binaries on linux.


osalpekar:
Build is now working. @mehtanirav @weiwangmeta can you take a look? I'll merge this as a checkpoint, then do a cleanup and use pkg-helpers in a follow-up

osalpekar:
Successful build of the conda binaries: https://github.com/pytorch/audio/runs/8004807210?check_suite_focus=true. I am rebasing and removing the on: PR trigger to make it importable so the binaries will not be built on the latest version.

packaging/pkg_helpers.bash
```diff
@@ -221,6 +222,11 @@ setup_conda_pytorch_constraint() {
   else
     export CONDA_CHANNEL_FLAGS="${CONDA_CHANNEL_FLAGS} -c pytorch -c pytorch-test -c pytorch-nightly"
   fi
+
+  export CONDA_CHANNEL_FLAGS="${CONDA_CHANNEL_FLAGS} -c pytorch-nightly"
```
mehtanirav, line 226: Why is this change required? It seems when `PYTORCH_VERSION` is not empty, we might end up adding `pytorch-nightly` channel twice?

mehtanirav:
LGTM. Just one minor comment about changes in `pkg_helpers.bash`

mehtanirav (approved):

packaging/pkg_helpers.bash
osalpekar: Good point, I'll remove the second one

PR #2635, opened 2022-08-18
Added example for InverseMelScale transform

ravimakhija:
Added example for InverseMelScale transform as mentioned in issue #1564.

### DIFF
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -433,6 +433,13 @@ class InverseMelScale(torch.nn.Module):
         norm (str or None, optional): If "slaney", divide the triangular mel weights by the width of the mel band
             (area normalization). (Default: ``None``)
         mel_scale (str, optional): Scale to use: ``htk`` or ``slaney``. (Default: ``htk``)
+
+    Example
+        >>> waveform, sample_rate = torchaudio.load("test.wav", normalize=True)
+        >>> mel_spectrogram_transform = transforms.MelSpectrogram(sample_rate, n_fft=1024)
+        >>> mel_spectrogram = mel_spectrogram_transform(waveform)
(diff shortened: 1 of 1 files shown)


torchaudio/transforms/_transforms.py
```diff
@@ -433,6 +433,13 @@ class InverseMelScale(torch.nn.Module):
         norm (str or None, optional): If "slaney", divide the triangular mel weights by the width of the mel band
             (area normalization). (Default: ``None``)
         mel_scale (str, optional): Scale to use: ``htk`` or ``slaney``. (Default: ``htk``)
+
+    Example
+        >>> waveform, sample_rate = torchaudio.load("test.wav", normalize=True)
+        >>> mel_spectrogram_transform = transforms.MelSpectrogram(sample_rate, n_fft=1024)
+        >>> mel_spectrogram = mel_spectrogram_transform(waveform)
+        >>> inverse_melscale_transform = transforms.InverseMelScale(n_stft=1024)
```
nateanl, line 441: The `n_stft` represents the number of STFT bins in spectrogram, which is `n_fft // 2 + 1`.
```suggestion
        >>> inverse_melscale_transform = transforms.InverseMelScale(n_stft=1024 // 2 + 1)
```

nateanl (commented):
Thanks for working on it. could you fix the value of `n_stft`? otherwise LGTM.

torchaudio/transforms/_transforms.py
ravimakhija: Ahh I think I understand now: "number of bins" versus "size of FFT" - so even though the inverse operation is an approximation, there's still value in being consistent. Committing the fix - thanks!

nateanl (approved):

PR #2644, opened 2022-08-23
Add simulate_rir_ism method for simulating RIR with Image Source Method

nateanl:


nateanl:
Doc page: https://output.circle-artifacts.com/output/job/a26503b3-98a5-4272-8add-cacb33932fef/artifacts/0/docs/prototype.functional.html#simulate-rir-ism

nateanl:
cc @fakufaku

torchaudio/prototype/functional/rir.py
```diff
@@ -0,0 +1,233 @@
...
+
+def simulate_rir_ism(
+    room: torch.Tensor,
+    source: torch.Tensor,
+    mic_array: torch.Tensor,
+    max_order: int,
+    e_absorption: Union[float, torch.Tensor],
+    output_length: Optional[int] = None,
+    delay_filter_length: int = 81,
+    center_frequency: torch.Tensor = _CENTER_FREQUENCY,
```
mthrok, line 121: Tensor is mutable, thus it should not be used for default value.
Please use `None` as a placeholder and create the tensor on-the-fly.

torchaudio/prototype/functional/rir.py
```diff
@@ -0,0 +1,233 @@
...
+    Returns:
+        (torch.Tensor): The simulated room impulse response waveform. Tensor with dimensions
+        `(channel, rir_length)`.
+
+    Note:
+        If ``e_absorption`` is a 2D Tensor, the center frequencies of octive bands are fixed to
+        ``[125.0, 250.0, 500.0, 1000.0, 2000.0, 4000.0, 8000.0]``. Users need to tune the values
+        to match the corresponding frequencies.
+    """
+    assert room.ndim == 1, f"room must be a 1D Tensor. Found {room.shape}."
```
mthrok, line 172: Please do not use assert in library code, as it does not invoked in optimized mode. Use `if ~ raise`.

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
...
+        make_filter_impl<scalar_t>(centers, sample_rate, n_fft, filters);
+      });
+  return filters;
+}
+
+TORCH_LIBRARY(rir, m) {
+  m.def(
+      "rir::build_rir(Tensor irs, Tensor delay_i, int rir_length) -> Tensor",
+      &torchaudio::rir::build_rir);
+  m.def("rir::make_filter", &torchaudio::rir::make_filter);
```
mthrok, line 130: Please use `torchaudio` as namescope.

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
...
+    int64_t n_fft) {
+  torch::Tensor filters;
+  AT_DISPATCH_FLOATING_TYPES_AND_HALF(
+      centers.scalar_type(), "make_filter", [&] {
+        make_filter_impl<scalar_t>(centers, sample_rate, n_fft, filters);
+      });
+  return filters;
+}
+
+TORCH_LIBRARY(rir, m) {
```
mthrok, line 126: Please use TORCH_LIBRARY_FRAGMENT and its variant for defining schema and registration.

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
...
+  });
+  auto n_freq = n_fft / 2 + 1;
+  torch::Tensor freq_resp = torch::zeros({n_freq, n}, centers.dtype());
+  torch::Tensor freq =
+      torch::arange(n_freq, centers.dtype()) / n_fft * sample_rate;
+  const scalar_t* freq_data = freq.data_ptr<scalar_t>();
+  scalar_t* freqreq_data = freq_resp.data_ptr<scalar_t>();
+
+  at::parallel_for(0, n, 0, [&](int64_t start, int64_t end) {
+    at::parallel_for(0, n_freq, 0, [&](int64_t start2, int64_t end2) {
```
mthrok, line 90: `at::parallel_for` can be used only when there is no call to `at::` in inner loop. https://fburl.com/tsg4ifjq
Thus it cannot be nested.

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
+#include <math.h>
+#include <torch/script.h>
+#include <torch/torch.h>
+using namespace torch::indexing;
+
+namespace torchaudio {
+namespace rir {
+
+template <typename scalar_t>
+void build_rir_impl(
```
mthrok, line 10: Can you add docstring / reference / description?
We do not have a way to build C++ doc yet, but we need make notes for future maintainers.

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -1,9 +1,10 @@
 import numpy as np
+import pyroomacoustics as pra
```
mthrok, line 2: please perform import guard

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -107,3 +108,77 @@ def test_add_noise_length_check(self):
...
+            max_order=max_order,
+            ray_tracing=False,
+            air_absorption=False,
+        )
+        mic_locs = np.asarray([mic_array[i].tolist() for i in range(channel)]).swapaxes(0, 1)
+        room.add_microphone_array(mic_locs)
+        room.add_source(source.tolist())
+        room.compute_rir()
+        max_len = max([room.rir[i][0].shape[0] for i in range(channel)])
+        actual = torch.zeros(channel, max_len, dtype=self.dtype, device=self.device)
```
mthrok, line 134: `expected` and `actual` are opposite.

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -107,3 +108,77 @@ def test_add_noise_length_check(self):
 
         with self.assertRaisesRegex(ValueError, "Length dimensions"):
             F.add_noise(waveform, noise, lengths, snr)
+
+    @skipIfNoModule("pyroomacoustics")
+    @parameterized.expand([(2, 1), (3, 4)])
+    def test_simulate_rir_ism_single_band(self, D, channel):
+        """Test simulate_rir_ism when absorption coefficients are identical for all walls."""
```
mthrok, line 115: I do not see the test logic checks the identity described in the docstring.
It is checking against pyroomacoustic, but the test logic does not check identity for walls.

Can you elaborate?

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
...
+    const torch::Tensor irs,
+    const torch::Tensor delay,
+    const int64_t rir_length) {
+  const int64_t num_band = irs.size(0);
+  const int64_t num_image = irs.size(1);
+  const int64_t num_mic = irs.size(2);
+  const int64_t ir_length = irs.size(3);
+  torch::Tensor rirs =
+      torch::zeros({num_band, num_mic, rir_length}, irs.dtype());
+  rirs.requires_grad_(true);
```
mthrok, line 49: Does this work with `torch.inference_mode()`?

mthrok: actually, what does it accomplish?
`build_rir_impl` manipulates the data of `rirs` by pointer, so autograd cannot track it.
In this case, the modification of `requires_grad=True` should be done by client code.

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+template <typename scalar_t>
+void make_filter_impl(
+    torch::Tensor& centers,
+    double sample_rate,
+    int64_t n_fft,
+    torch::Tensor& filters) {
+  int64_t n = centers.size(0);
+  torch::Tensor new_bands = torch::zeros({n, 2}, centers.dtype());
+  new_bands.requires_grad_(true);
```
mthrok, line 65: Similar to above, I double the effectiveness of `requires_grad_` here.

torchaudio/prototype/functional/rir.py
```diff
@@ -0,0 +1,233 @@
...
+    # attenuation
+    exp_lo = abs(torch.floor((XYZ / 2)))
+    exp_hi = abs(torch.floor((XYZ + 1) / 2))
+    t_lo = tr[:, ::2].unsqueeze(1).repeat(1, XYZ.shape[0], 1)  # (num_band, left walls)
+    t_hi = tr[:, 1::2].unsqueeze(1).repeat(1, XYZ.shape[0], 1)  # (num_band, right walls)
+    att = torch.prod((t_lo**exp_lo) * (t_hi**exp_hi), dim=-1)  # (num_band, num_image_source)
+    return img_loc, att
+
+
+def _hann(x: torch.Tensor, T: int):
```
mthrok, line 73: Can you add comment on why we cannot use `torch.hann_window`?

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
nateanl: Yes. if `e_absorption` is just a floating-point value, then it means it is identical to all walls. That's why the room in pyroonacoustics is constructed with `materials=pra.Material(e_absorption),`.
The difference of this test and the one below is if `e_absorption` is a 2D Tensor, the multi-band logic in https://github.com/pytorch/audio/pull/2644/files#diff-bfbe08000c5dc05901af3bad9503693a99068426351748925a322990daed0cb3R217-R222 is applied. We need a separate test for that.

torchaudio/csrc/build_rir.cpp
nateanl: I tried to remove this line and run autograd_test, the test fails but was working with this line.

torchaudio/prototype/functional/rir.py
fakufaku: Actually, `torch.hann_window` can only sample the window function at integer points, which is suitable to create the window functions for STFT and such. In the image source model, we need to sample the continuous window function at non-integer points.

torchaudio/csrc/build_rir.cpp
nateanl: It also passes the autograd test.

torchaudio/prototype/functional/rir.py
```diff
@@ -0,0 +1,241 @@
...
+        raise ValueError("The filter length must be odd")
+
+    pad = delay_filter_length // 2
+    n = torch.arange(-pad, pad + 1, device=delay.device) + delay_i[..., None]
+    delay = delay[..., None]
+
+    return torch.special.sinc(n - delay) * _hann(n - delay, 2 * pad)
+
+
+def simulate_rir_ism(
```
mthrok, line 111: Can you decorate this function so that if RIR is not compiled, it will fail immediately with easy-to-understand message?

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,187 @@
...
+      });
+  return filters;
+}
+
+TORCH_LIBRARY_FRAGMENT(torchaudio, m) {
+  m.def(
+      "torchaudio::build_rir(Tensor irs, Tensor delay_i, int rir_length) -> Tensor",
+      &torchaudio::rir::build_rir);
+  m.def(
+      "torchaudio::make_filter(Tensor centers, float sample_rate, int n_fft) -> Tensor",
```
mthrok, line 182: The function name is very generic to be in torchaudio namespace. Can you make it more descriptive / specific?

torchaudio/csrc/build_rir.cpp
```diff
@@ -0,0 +1,187 @@
+#include <math.h>
+#include <torch/script.h>
+#include <torch/torch.h>
+using namespace torch::indexing;
+
+namespace torchaudio {
+namespace rir {
```
mthrok, line 7: Since this file does not have corresponding header file, can you add anonymous namespace so that it is not referable from other C++ source?

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -1,9 +1,14 @@
 import numpy as np
+
+try:
+    import pyroomacoustics as pra
```
mthrok, line 4: Please follow https://github.com/pytorch/audio/blob/fda00bf7bfe0b6a3a67ae0274a892299fdbefc4f/test/torchaudio_unittest/backend/soundfile/info_test.py#L20

nateanl:
After discussing with @fakufaku and @NicolasHug, the gradients of converting a float tensor to int tensor and using it as indices for another tensor may require an analytic formula. Before getting the formula with proof, I will disable the autograd test and focus on the numerical correctness in the forward method.

nateanl:
will address the issue in a new PR #2880

Issue #2539, opened 2022-07-11
Metadata mode for torchaudio.dataset


leo19941227:
### 🚀 The feature

Hello,

Thanks for the handy tools for parsing the database!
I am wondering if it is possible to let the torchaudio.dataset classes have two modes:

1. Return the actual waveform
2. Return the waveform path

Others metadata like transcriptions and speaker labels can just remain the same.

### Motivation, pitch

Since usually we would like to organize a corpus into some standardized data format (like the kaldi's data directory), before loading the actual waveform. So we can have more flexibility on how many training data points we want to use, or some preprocessing steps which might be hard to do in __getitem__ and collate_fn. For example, In SUPERB ASV, if an utterance in VoxCeleb1 is not longer than 2 seconds after VAD, it is discarded and won't be used for training (There is no utterance in this case for the testing set). In the current dataset class implementation, I not yet come out a feasible way to realize this kind of logic since the __len__ is fixed.

Furthermore, by switching to waveform path mode, it becomes really fast to iterate over the entire dataset. Then, I can re-use the dataset classes as the corpus directory parsing helper, easily obtain all the transcription in LibriSpeech and train tokenizers, or easily obtain all the speakers in VoxCeleb1 to make a Categorical encoder. In the current implementation which returns the actual waveforms, it can become too slow to iterator over the entire dataset and is not tolerable for just preprocessing steps.

### Alternatives

No

mthrok:
I think the proposal makes sense. Decoupling dataset traversal and audio decoding makes it flexible.

For the second part, I think that the most flexible approach is to have explicit table representation of dataset internally, and apply operations like "drop rows that do not meet this criteria" and "split the dataset into batches with similar length" on it. I get the feeling that this will also give user code opportunity to improve the overall throughput. 

The existing TorchAudio datasets started as simple adaptation of classic `torch.utils.Dataset`, which was not designed for advanced usage like these. I think it's good time for us to move away from it.

leo19941227:
Cool!

nateanl:
Hi @leo19941227, we can add the metadata feature under prototype. One potential solution is adding `fetch_metadata` method for each dataset. The returned item is mostly identical, only the waveform is replaced by the absolute path of the audio. Does that suit your need?

leo19941227:
Hello @nateanl ,

Cool! I think this is nice.

leo19941227:
Hi @carolineechen,

Thanks, it looks good to me!

[...]

PR #2646, opened 2022-08-23
Add LibriSpeech Base class for metadata

carolineechen:
Add `LIBRISPEECHBase` class that returns the same information as `LIBRISPEECH`, but returns a file path instead of the loaded waveform. See #2539 for some context.

This design choice results in 2x API for the dataset -- one for metadata mode, and one for wav mode. We may want to think of a better way to arrange the docs on the website since most variables are repeated, so while this under discussion, I have simply left it out of the docs build for now.

TODO: testing

cc: @leo19941227

### DIFF
--- a/torchaudio/datasets/librispeech.py
+++ b/torchaudio/datasets/librispeech.py
@@ -43,16 +43,16 @@ def download_librispeech(root, url):
     extract_archive(archive)
 
 
-def load_librispeech_item(
+def get_librispeech_metadata(
     fileid: str, path: str, ext_audio: str, ext_txt: str
-) -> Tuple[Tensor, int, str, int, int, int]:
+) -> Tuple[str, int, str, int, int, int]:
     speaker_id, chapter_id, utterance_id = fileid.split("-")
 
-    # Load audio
+    # Get audio path and sample rate
     fileid_audio = f"{speaker_id}-{chapter_id}-{utterance_id}"
-    file_audio = fileid_audio + ext_audio
-    file_audio = os.path.join(path, speaker_id, chapter_id, file_audio)
-    waveform, sample_rate = torchaudio.load(file_audio)
+    filepath = os.path.join(speaker_id, chapter_id, f"{fileid_audio}{ext_audio}")
+    full_path = os.path.join(path, filepath)
+    sample_rate = torchaudio.info(full_path).sample_rate
 
     # Load text
     file_text = f"{speaker_id}-{chapter_id}{ext_txt}"
@@ -67,7 +67,7 @@ def load_librispeech_item(
             raise FileNotFoundError(f"Translation not found for {fileid_audio}")
 
     return (
-        waveform,
+        filepath,
         sample_rate,
         transcript,
         int(speaker_id),
@@ -76,7 +76,26 @@ def load_librispeech_item(
     )
 
 
-class LIBRISPEECH(Dataset):
+def load_librispeech_item(
+    fileid: str, path: str, ext_audio: str, ext_txt: str
+) -> Tuple[Tensor, int, str, int, int, int]:
+    filepath, sample_rate, transcript, speaker_id, chapter_id, utterance_id = get_librispeech_metadata(
+        fileid, path, ext_audio, ext_txt
+    )
+    full_path = os.path.join(path, filepath)
+    waveform, _ = torchaudio.load(full_path)
+
+    return (
+        waveform,
+        sample_rate,
+        transcript,
+        speaker_id,
+        chapter_id,
+        utterance_id,
+    )
+
+
+class LIBRISPEECHBase(Dataset):
     """Create a Dataset for *LibriSpeech* [:footcite:`7178964`].
 
     Args:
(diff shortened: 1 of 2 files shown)


torchaudio/datasets/librispeech.py
```diff
@@ -67,7 +67,7 @@ def load_librispeech_item(
             raise FileNotFoundError(f"Translation not found for {fileid_audio}")
 
     return (
-        waveform,
+        filepath,
```
carolineechen, line 70: should we return the relative file path to the provided path/root, or the full file path (including path/root) that can directly be passed into `torchaudio.load()`

torchaudio/datasets/librispeech.py
```diff
@@ -43,16 +43,16 @@ def download_librispeech(root, url):
     extract_archive(archive)
 
 
-def load_librispeech_item(
+def get_librispeech_metadata(
```
mthrok, line 46: Please move this to class method so that it's overridable from client code.

torchaudio/datasets/librispeech.py
mthrok: I vote for relative path, because that will allow to iterate on dataset exists in places other than local file system.

torchaudio/datasets/librispeech.py
mthrok: Same goes to `load_librispeech_item`.

torchaudio/datasets/librispeech.py
```diff
@@ -76,7 +76,26 @@ def load_librispeech_item(
...
+        waveform,
+        sample_rate,
+        transcript,
+        speaker_id,
+        chapter_id,
+        utterance_id,
+    )
+
+
+class LIBRISPEECHBase(Dataset):
```
mthrok, line 98: Let's not keep dragging the bad naming scheme.

```suggestion
class LibriSpeechBase(Dataset):
```

mthrok: Also, since we are introducing a new class, which gives rare opportunity to redefine the interface without BC, let's remove `folder_in_archive`, and make `root` optional.

mthrok: Similarly, we should split download from the constructor, downloading should not be the responsibility of this class (or at least constructor).

carolineechen: sure, we can redefine/simplify the API for the base class. curious what's the reasoning for removing downloading from this class?

mthrok: Downloading is a complex procedure. It involves network access, local file access and the permission.
There are so many ways to fail for downloading feature, for example, network unreachable, slow network, directory creation permission, file write permission, and file storage limit. The fact that the existing dataset interface ended up having `folder_in_archive` argument is one testament.

Also the data integrity is not resolved with the current approach. It is easy to verify the integrity of downloaded archive by using SHA-256 hash, but what about the individual data? Say, I delete one of the WAV file from the directory, should the download function refill it somehow? How to detect it? All these specifications and responsibility of flawless executions should be placed in a separate location. Dataset class should not be responsible. 

From the perspective of library aiming to be building block (or single responsibility principle), it's not something that should be performed inside of dataset constructor. It should be performed separately.
By introducing the Base dataset class, we can re-interpret the existing dataset classes as higher-level implementations, composed of multiple primitives.

torchaudio/datasets/librispeech.py
```diff
@@ -76,7 +76,74 @@ def load_librispeech_item(
...
+        sample_rate,
+        transcript,
+        speaker_id,
+        chapter_id,
+        utterance_id,
+    )
+
+
+class LibriSpeechBase(Dataset):
+    """Create a Dataset for *LibriSpeech* [:footcite:`7178964`].
```
nateanl, line 102: I feel the description can be different from `LIBRISPEECH`, for example, explicitly mention the class is only for fetching metadata, without loading the actual waveforms.

nateanl: shall we also add it to the documentation page?

carolineechen: I did not add it to docs yet since I wasn't sure the best way to do so with 2x API. Maybe something formatted like this?

```
+ LibriSpeech
   - LibriSpeechBase
   - LIBRISPEECH
```

torchaudio/datasets/librispeech.py
carolineechen: is this preferable to having users directly override the `__getitem__` method of the dataset (as opposed to overriding a class method that `__getitem__` calls which involves going a layer deeper)? also since `load_librispeech_item` is currently being used by another dataset/accessible to users

mthrok: `load_librispeech_item` should not be public. Can we follow up and make it private?

From the regular OOP perspective, the right approach is to let users overwrite `__getitem__`. However, this forces the users to re-write/copy-paste almost the whole logic of path-parsing and text processing. There are situations where only decoding should be overwritten but since these functions being module-level plain function, it is not plausible. These functions being class method allow to override the logic in the subclass. That's the pattern we should have been followed in the dataset class implementations, but existing datasets not doing that led to the situation where new dataset did not follow either. See https://github.com/pytorch/audio/issues/910#issuecomment-693564762 for the basic idea.

carolineechen:
switching to different approach for enabling easy access to metadata, similar to https://github.com/pytorch/audio/issues/910#issuecomment-693564762

carolineechen:
switching to a different approach for enabling easy access to metadata, similar to https://github.com/pytorch/audio/issues/910#issuecomment-693564762

PR #2647, opened 2022-08-24
Weiwangmeta/enumerate linux wheels

weiwangmeta:
Straightforward implementation of linux wheel caller workflow. Docker image is part of the parameters but is not suitable for matrix expression, therefore using this straightforward enumeration. The job names are consistent with circleci jobs and their corresponding docker image, difference is that github actions does not support job names with "." inside!


osalpekar:
Looks great - thanks for getting this to work! Let's rebase on latest master to resolve the merge conflicts with the already existing workflow, and then I think we should be good to check in.

.github/workflows/build_wheels_linux.yml
```diff
@@ -1,13 +1,180 @@
 name: Build TorchAudio Linux Wheels
 
 on:
+  push:
+    branches:
+      weiwangmeta/enumerate_linux_wheels
+  pull_request:
+    branches: [ main ]
+
```
osalpekar, line 9: Let's remove these changes so we don't start building on each PR

.github/workflows/build_wheels_linux.yml
```diff
@@ -1,13 +1,180 @@
...
+      docker_image: "pytorch/manylinux-cuda113"
+      python_version: "3.10"
+      gpu_arch_version: "cu113"
+  binary_linux_wheel_py310_cpu:
+    uses: pytorch/test-infra/.github/workflows/build_wheels_linux_reusable.yml@main
+    with:
+      repository: "audio"
+      docker_image: "pytorch/manylinux-cuda102"
+      python_version: "3.10"
+      gpu_arch_version: "cpu"
```
osalpekar, line 180: Just a general note: Apart from cpu builds using the cuda-102 image, there is a fairly straightforward mapping between gpu_arch_version and docker_image. So the matrix strategy can still work with simple if-else logic. Either way, let's merge this in first (and do the same for conda), and we can do the cleanup later.

osalpekar (approved):
LGTM. 2 minor notes, only 1 is actionable for this PR.

.github/workflows/build_wheels_linux.yml
weiwangmeta: thank you. Will explore the if-else logic afterwards. BTW, can you please merge this PR? I do not seem to have permission.

.github/workflows/build_wheels_linux.yml
weiwangmeta: Done. Removed pull_request on main.

PR #2648, opened 2022-08-24
Add file-like object support to StreamWriter

mthrok:


test/torchaudio_unittest/io/stream_writer_test.py
```diff
@@ -42,8 +42,42 @@ def get_video_chunk(fmt, frame_rate, *, width, height):
...
+_media_source = parameterized_class(
+    ("test_fileobj",),
+    [(False,), (True,)],
+    class_name_func=lambda cls, _, params: f'{cls.__name__}{"_fileobj" if params["test_fileobj"] else "_path"}',
+)
+
+
+class _MediaSourceMixin:
+    def setUp(self):
+        super().setUp()
```
nateanl, line 56: what's the base class of `_MediaSourceMixin`?

nateanl (approved):
some minor comments

test/torchaudio_unittest/io/stream_writer_test.py
mthrok: It's a [`Mixin`](https://en.wikipedia.org/wiki/Mixin) pattern. This class itself is not intended to be used directly, it supplements the functionality to other classes. The base class is defined in the inherited class.

Issue #2643, opened 2022-08-23
Use torch.linalg.lstsq optimizer in InverseMelScale


nateanl:
### 🚀 The feature

To increase the speed of `InverseMelScale` module, the SGD optimization can be replace with ` torch.linalg.lstsq`.

### Motivation, pitch

The current `InverseMelScale` module applies SGD optimizer and estimate the spectrogram in a for loop. The speed is much slower than librosa (See https://github.com/pytorch/audio/issues/2594) which uses non-negative least squares (nnls) algorithm. Another issue is the module can not be run in torch's inference mode (See https://github.com/pytorch/audio/issues/1902), because the gradient is disabled in inference mode, which blocks the SGD optimization.

The speed of `torch.linalg.lstsq` is much faster than SGD. Another benefit is it can be run without gradients, hence it enables the module to run in torch inference mode.

### Alternatives

Another solution is using the same `L-BFGS-B` optimizer as in librosa. However, there is no implementation in PyTorch yet.

mthrok:
One of the aspects brought up during the introduction of `InverseMelScale` was numerical stability. https://github.com/pytorch/audio/pull/366

What's the stability of `torch.linalg.lstsq`?

nateanl:
According to `torch.linalg.lstsq`'s doc page, it mentioned:
```
This function computes X = A.pinverse() @ B in a faster and more numerically stable way than performing the
computations separately.
```

It should be more stable than `SVD`, which is used in `pinverse`.

nateanl:
The issue of `torch.linalg.lstsq` is that there is no non-negative constraint, hence the generated spectrogram is not global optimal. The best solution should be implementing `L-BFGS-B` optimizer in PyTorch, which is faster than internal optimization, and it can be run in inference mode.

mthrok:
Seems like this is SciPy implementation.

https://github.com/scipy/scipy/blob/v1.9.3/scipy/optimize/_lbfgsb_py.py#L210-L399

Kinyugo:
I have been using an implementation where I compute the pseudo-inverse of the filterbanks. It seems to work nicely and it gives similar results as `lstsq`. An upside of the pseudo-inverse is that we can cache the inverse filterbanks and speed up the computation. 

A downside of this approach is that it introduces some noticable artifacts, though even my tests with librosa still show that the artifacts are present. 

Here is the algorithm: 
```python
from typing import Optional 

import torch
from einops import rearrange, repeat
[...]

PR #2649, opened 2022-08-24
Use torch.linalg.lstsq in InverseMelScale

nateanl:
To address https://github.com/pytorch/audio/issues/2643

### DIFF
--- a/test/torchaudio_unittest/transforms/transforms_test_impl.py
+++ b/test/torchaudio_unittest/transforms/transforms_test_impl.py
@@ -15,7 +15,7 @@ class TransformsTestBase(TestBaseMixin):
         """Gauge the quality of InverseMelScale transform.
 
         As InverseMelScale is currently implemented with
-        random initialization + iterative optimization,
+        torch.linalg.lstsq optimization.
         it is not practically possible to assert the difference between
         the estimated spectrogram and the original spectrogram as a whole.
         Estimated spectrogram has very huge descrepency locally.
--- a/test/torchaudio_unittest/transforms/autograd_test_impl.py
+++ b/test/torchaudio_unittest/transforms/autograd_test_impl.py
@@ -184,6 +184,13 @@ class AutogradTestMixin(TestBaseMixin):
         waveform = get_whitenoise(sample_rate=sample_rate, duration=0.05, n_channels=2)
         self.assert_grad(transform, [waveform])
 
+    def test_inverse_melscale(self):
+        n_mels = 13
+        n_stft = 129
+        mel_spec = torch.randn(3, 2, n_mels, 32) ** 2
+        transform = T.InverseMelScale(n_stft, n_mels)
+        self.assert_grad(transform, [mel_spec])
+
     def test_melscale(self):
         sample_rate = 8000
         n_fft = 400
--- a/test/torchaudio_unittest/transforms/batch_consistency_test.py
+++ b/test/torchaudio_unittest/transforms/batch_consistency_test.py
@@ -49,14 +49,12 @@ class TestTransforms(common_utils.TorchaudioTestCase):
         self.assert_batch_consistency(transform, specgram, atol=atol)
 
     def test_batch_InverseMelScale(self):
-        n_mels = 32
-        n_stft = 5
+        n_mels = 13
+        n_stft = 129
         mel_spec = torch.randn(3, 2, n_mels, 32) ** 2
         transform = T.InverseMelScale(n_stft, n_mels)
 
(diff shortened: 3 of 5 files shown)


test/torchaudio_unittest/transforms/transforms_test_impl.py
```diff
@@ -15,7 +15,7 @@ def test_InverseMelScale(self):
         """Gauge the quality of InverseMelScale transform.
 
         As InverseMelScale is currently implemented with
-        random initialization + iterative optimization,
+        torch.linalg.lstsq optimization.
         it is not practically possible to assert the difference between
         the estimated spectrogram and the original spectrogram as a whole.
         Estimated spectrogram has very huge descrepency locally.
```
carolineechen, line 21:
```suggestion
        Estimated spectrogram has very huge discrepancy locally.
```

carolineechen (commented):
This looks like a pretty significant BC-breaking change that would break existing implementations using any of the removed parameters. Do you think it makes sense to maintain BC, by adding an option to choose which underlying implementation to use, with the old version being the default to avoid BC-breaking? Also normally for BC-breaking changes we will first issue a warning in a release before formally modifying it in the following release -- we could consider something like that if we are certain this approach is better/correct. Curious what the numerical difference looks like?

nateanl:
> Curious what the numerical difference looks like?

The numerical difference is significant. So besides the API BC-breaking, it is also numerical BC-breaking.

mthrok:
> This looks like a pretty significant BC-breaking change (or would this be considered bug-fix?) that would break existing implementations using any of the removed parameters. Do you think it makes sense to maintain BC, by adding an option to choose which underlying implementation to use, with the old version being the default to avoid BC-breaking? Also normally for BC-breaking changes we will first issue a warning in a release before formally modifying it in the following release -- we could consider something like that if we are certain this approach is better/correct. Curious what the numerical difference looks like?
> 
> cc @mthrok

Since the change is motivated by the improvement of numerical result, first we need to validate the improvement. Passing existing tests are not sufficient for this. Then we can think of a better way to handle interface breakage. If the new implementation should yield perfect inversion, then interface breakage is minor issue.

Firstly, per the following comment, the current implementation runs the optimizer, which generates blur result, and we cannot gauge the element-level differences.

https://github.com/pytorch/audio/blob/72404de901cdb47049d8f0383a4dc51a1b739317/test/torchaudio_unittest/transforms/transforms_test_impl.py#L17-L26

So the current test looks at the distribution of relative difference. We need to re-assess the numerical property of the new implementation and revise this to make sure that the change is improvement.

https://github.com/pytorch/audio/blob/72404de901cdb47049d8f0383a4dc51a1b739317/test/torchaudio_unittest/transforms/transforms_test_impl.py#L49-L53

Second issue is the possibility of the numerical instability. I do not know enough of the underlying math, but we need to test the new implementation on variety of configuration on different real/synthetic audio. We should be able to answer questions like, "is it reliably stable on common audios?", "if the input mel-spectrogram meets some conditions that make `lstsq` unstable, can we do something about it?"

nateanl:
I doubt the numerical improvement would be marginal, stft -> mel is a lossy computation, hence both torchaudio and librosa use optimization method to estimate stft from mel spectrogram. The change in this PR mainly solves two issues: running speed and running in inference mode.

The [unit test in librosa](https://github.com/librosa/librosa/blob/f101492a5b6ee9ba09d12867b9718e58621003e3/tests/test_features.py#L814-L835) only compares the difference between `mel` and `mel_bank * stft`, where stft is the estimated one.
```python
def test_mel_to_stft(power, dtype, n_fft):
    srand()

    # Make a random mel spectrum, 4 frames
    mel_basis = librosa.filters.mel(sr=22050, n_fft=n_fft, n_mels=128, dtype=dtype)

    stft_orig = np.random.randn(n_fft // 2 + 1, 4) ** power
    mels = mel_basis.dot(stft_orig.astype(dtype))

    stft = librosa.feature.inverse.mel_to_stft(mels, power=power, n_fft=n_fft)

    # Check precision
    assert stft.dtype == dtype

    # Check for non-negative spectrum
    assert np.all(stft >= 0)

    # Check that the shape is good
    assert stft.shape[0] == 1 + n_fft // 2

    # Check that the approximation is good in RMSE terms
    assert np.sqrt(np.mean((mel_basis.dot(stft ** power) - mels) ** 2)) <= 5e-2
```

When comparing `stft` and `stft_orig` in librosa, the element-wise difference is also significant.

nateanl:
Close it as `torch.linalg.lstsq` could return negative values in the output. We can still make it non-negative by using `torch.clamp`, but the result is not globally optimal.

PR #2653, opened 2022-08-26
Add metadata function for LibriSpeech

carolineechen:
Adding support for metadata mode, requested in #2539, by adding a public `get_metadata()` function in the dataset. This function can be used directly by users to fetch metadata for individual dataset indices, or users can subclass the dataset and override `__getitem__` with `get_metadata` to create a dataset class that directly handles metadata mode.

### DIFF
--- a/torchaudio/datasets/librilight_limited.py
+++ b/torchaudio/datasets/librilight_limited.py
@@ -16,7 +16,7 @@ _CHECKSUM = "5d1efdc777b548194d7e09ba89126e2188026df9fd57aa57eb14408d2b2342af"
 _SUBSET_MAP = {"10min": ["1h/0"], "1h": ["1h/*"], "10h": ["1h/*", "9h"]}
 
 
-def _get_fileids_paths(path, subset, _ext_audio) -> List[Tuple[str, str]]:
+def _get_fileids_paths(path, folders, _ext_audio) -> List[Tuple[str, str]]:
     """Get the file names and the corresponding file paths without `speaker_id`
     and `chapter_id` directories.
     The format of path is like:
@@ -24,16 +24,12 @@ def _get_fileids_paths(path, subset, _ext_audio) -> List[Tuple[str, str]]:
         {root}/{_ARCHIVE_NAME}/9h/[clean, other]
     """
 
-    def _get_rel_folder(p, path):
-        rel_path = os.path.relpath(os.path.dirname(p), path)
-        return os.path.dirname(os.path.dirname(rel_path))  # move up 2 levels
-
+    path = Path(path)
     files_paths = []
-    for folder in _SUBSET_MAP[subset]:
-        files_paths += [
-            (_get_rel_folder(p, path), str(p.stem)) for p in Path(path).glob(f"{folder}/*/*/*/*" + _ext_audio)
-        ]
-    files_paths = sorted(files_paths, key=lambda x: x[0] + x[1])
+    for folder in folders:
+        paths = [p.relative_to(path) for p in path.glob(f"{folder}/*/*/*/*{_ext_audio}")]
+        files_paths += [(str(p.parent.parent.parent), str(p.stem)) for p in paths]  # get subset folder and file name
+    files_paths.sort(key=lambda x: x[0] + x[1])
     return files_paths
 
 
@@ -58,8 +54,9 @@ class LibriLightLimited(Dataset):
         subset: str = "10min",
         download: bool = False,
     ) -> None:
-        if subset not in ["10min", "1h", "10h"]:
-            raise ValueError("`subset` must be one of ['10min', '1h', '10h']")
+        if subset not in _SUBSET_MAP:
+            raise ValueError(f"`subset` must be one of {_SUBSET_MAP.keys()}. Found: {subset}")
+        folders = _SUBSET_MAP[subset]
 
         root = os.fspath(root)
         self._path = os.path.join(root, _ARCHIVE_NAME)
@@ -70,7 +67,7 @@ class LibriLightLimited(Dataset):
             if not os.path.isfile(archive):
                 download_url_to_file(_URL, archive, hash_prefix=_CHECKSUM)
             extract_archive(archive)
-        self._fileids_paths = _get_fileids_paths(self._path, subset, self._ext_audio)
+        self._fileids_paths = _get_fileids_paths(self._path, folders, self._ext_audio)
 
     def __getitem__(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
         """Load the n-th sample from the dataset.
@@ -84,7 +81,7 @@ class LibriLightLimited(Dataset):
         file_path, fileid = self._fileids_paths[n]
         metadata = _get_librispeech_metadata(fileid, self._path, file_path, self._ext_audio, self._ext_txt)
         waveform, _ = torchaudio.load(os.path.join(self._path, metadata[0]))
-        return (waveform, metadata[1:])
+        return (waveform,) + metadata[1:]
 
     def __len__(self) -> int:
         return len(self._fileids_paths)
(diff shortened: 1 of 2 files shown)


torchaudio/datasets/librispeech.py
```diff
@@ -43,20 +43,20 @@ def download_librispeech(root, url):
     extract_archive(archive)
 
 
-def load_librispeech_item(
-    fileid: str, path: str, ext_audio: str, ext_txt: str
-) -> Tuple[Tensor, int, str, int, int, int]:
+def _get_librispeech_metadata(
```
carolineechen, line 46: left this function outside of the dataset class as it is reused by librilight

torchaudio/datasets/librilight_limited.py
```diff
@@ -21,23 +23,16 @@ def _get_fileids_paths(path, subset, _ext_audio) -> List[Tuple[str, str]]:
...
-        files_paths = [
-            (os.path.join(os.path.dirname(p), "..", ".."), str(p.stem))
-            for p in Path(path).glob("1h/*/*/*/*/*" + _ext_audio)
+
+    def _get_rel_folder(p, path):
+        rel_path = os.path.relpath(os.path.dirname(p), path)
+        return os.path.dirname(os.path.dirname(rel_path))  # move up 2 levels
+
+    files_paths = []
+    for folder in _SUBSET_MAP[subset]:
```
mthrok, line 32: Looks like the argument check for `subset` value is removed. Can you add that back and improve the error message to include the valid choices?

torchaudio/datasets/librispeech.py
```diff
@@ -43,20 +43,20 @@ def download_librispeech(root, url):
...
 
-    # Load audio
+    # Get audio path and sample rate
     fileid_audio = f"{speaker_id}-{chapter_id}-{utterance_id}"
-    file_audio = fileid_audio + ext_audio
-    file_audio = os.path.join(path, speaker_id, chapter_id, file_audio)
-    waveform, sample_rate = torchaudio.load(file_audio)
+    filepath = os.path.join(url, speaker_id, chapter_id, f"{fileid_audio}{ext_audio}")
+    full_path = os.path.join(root, filepath)
+    sample_rate = torchaudio.info(full_path).sample_rate
```
mthrok, line 55: Instead of fetch sample_rate here, why not hard code `16000` and assert it at `torchaudio.load`? LibriSpeech *is* expected to be 16kHz.

torchaudio/datasets/librispeech.py
```diff
@@ -102,22 +102,43 @@ def __init__(
...
-                download_librispeech(root, url)
+                _download_librispeech(root, url)
             else:
                 raise RuntimeError(
                     f"Dataset not found at {self._path}. Please set `download=True` to download the dataset."
                 )
 
         self._walker = sorted(str(p.stem) for p in Path(self._path).glob("*/*/*" + self._ext_audio))
 
+    def get_metadata(self, n: int) -> Tuple[Tensor, int, str, int, int, int]:
```
mthrok, line 123: Add this method to doc?

torchaudio/datasets/librilight_limited.py
carolineechen: there's an argument check for subset in init in [line 61](https://github.com/pytorch/audio/pull/2653/files#diff-e87636061c799854eb56661a88dd412f58a39305253f26feb354f2168d0aff27R61) which includes the valid choices. since this is an internal function called only from the dataset, I don't think we need to check again.

torchaudio/datasets/librispeech.py
carolineechen: build_docs is broken on this PR, but it should already be included under the `:members:` in the .rst file. I tested locally as well.

torchaudio/datasets/librilight_limited.py
mthrok: In that case, can we use the `_SUBSET_MAP` dict at the argument check so that the change to `_SUBSET_MAP` is automatically reflected to the argument check?

Something like this;

```python
# In `__init__`
if subset not in _SUBSET_MAP:
    raise ValueError(f"`subset` must be one of {_SUBSET_MAP.keys()}. Found: {subset}")
folder = _SUBSET_MAP[subset]

_get_fileids_paths(path, folder, _ext_audio)
```

This way the use of `_SUBSET_MAP` is consolidated to only inside of `__init__`, so the question about validation won't come up again.

torchaudio/datasets/librilight_limited.py
```diff
@@ -2,42 +2,37 @@
...
 from torch.utils.data import Dataset
-from torchaudio.datasets.librispeech import load_librispeech_item
+from torchaudio.datasets.librispeech import _get_librispeech_metadata
 from torchaudio.datasets.utils import extract_archive
 
 
 _ARCHIVE_NAME = "librispeech_finetuning"
 _URL = "https://dl.fbaipublicfiles.com/librilight/data/librispeech_finetuning.tgz"
 _CHECKSUM = "5d1efdc777b548194d7e09ba89126e2188026df9fd57aa57eb14408d2b2342af"
+_SUBSET_MAP = {"10min": ["1h/0"], "1h": ["1h/*"], "10h": ["1h/*", "9h"]}
```
nateanl, line 16: this impl is very clean, nice 👍

nateanl (approved):

torchaudio/datasets/librilight_limited.py
```diff
@@ -2,42 +2,37 @@
...
-            for p in Path(path).glob("1h/*/*/*/*/*" + _ext_audio)
+
+    def _get_rel_folder(p, path):
+        rel_path = os.path.relpath(os.path.dirname(p), path)
+        return os.path.dirname(os.path.dirname(rel_path))  # move up 2 levels
+
+    files_paths = []
+    for folder in folders:
+        files_paths += [
+            (_get_rel_folder(p, path), str(p.stem)) for p in Path(path).glob(f"{folder}/*/*/*/*" + _ext_audio)
```
mthrok, line 34: Since `p` is already Path object, using the equivalent methods on `p` would make this code simpler. https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module

mthrok: I do not see much value on defining the `_get_rel_folder` function inside.
It does not capture a value from local scope, and it's only used once.

Either define the function on module level or do it in the `for` loop.

Also, it's not immediately and intuitively readable what it accomplishes. 
Perhaps change the function name to something more descriptive?

mthrok (approved):

torchaudio/datasets/librilight_limited.py
```diff
@@ -2,42 +2,33 @@
...
-                (os.path.join(os.path.dirname(p), "..", ".."), str(p.stem))
-                for p in Path(path).glob("9h/*/*/*/*" + _ext_audio)
-            ]
-    else:
-        raise ValueError(f"Unsupported subset value. Found {subset}.")
+
+    path = Path(path)
+    files_paths = []
+    for folder in folders:
+        paths = [p.relative_to(path) for p in path.glob(f"{folder}/*/*/*/*" + _ext_audio)]
```
mthrok, line 30:
```suggestion
        paths = [p.relative_to(path) for p in path.glob(f"{folder}/*/*/*/*{_ext_audio}")]
```

torchaudio/datasets/librilight_limited.py
```diff
@@ -2,42 +2,33 @@
...
+        files_paths += [(str(p.parent.parent.parent), str(p.stem)) for p in paths]  # get subset folder and file name
     files_paths = sorted(files_paths, key=lambda x: x[0] + x[1])
```
mthrok, line 32:
```suggestion
    files_paths.sort(key=lambda x: x[0] + x[1])
```

torchaudio/datasets/librispeech.py
```diff
@@ -102,22 +101,45 @@ def __init__(
...
```
kuraga, line 122: The type annotation is incorrect. The fix: #3757. @carolineechen @mthrok @nateanl

PR #2658, opened 2022-09-06
Add Decoder LM Docs

carolineechen:
modifications to ctc decoder LM docstrings on top of #2657


torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -113,12 +113,78 @@ class CTCHypothesis(NamedTuple):
...
+        """Returns child corresponding to usrIdx, or creates and returns a new state if input index
+        is not found.
+
+        Args:
+            usrIdx (int): index corresponding to child state
+
+        Returns:
+            CTCDecoderLMState: child state corresponding to usrIdx
+        """
+        pass
```
mthrok, line 132: Is the base class (and this class) intended to be an interface? If so, let's decorate it with [`abc.abstractmethod`](https://docs.python.org/3/library/abc.html#abc.abstractmethod).

torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -113,6 +114,83 @@ class CTCHypothesis(NamedTuple):
...
+        """Returns child corresponding to usr_index, or creates and returns a new state if input index
+        is not found.
+
+        Args:
+            usr_index (int): index corresponding to child state
+
+        Returns:
+            CTCDecoderLMState: child state corresponding to usr_index
+        """
+        return super().child(usr_index)
```
carolineechen, line 133: @mthrok do you think calling super here is fine? am running into an issue where nothing is returned if I just use pass

mthrok (commented):
Should we annotate the methods to include return types?

PR #2659, opened 2022-09-07
Add unit test for LibriMix dataset

nateanl:
Besides the unit test, the PR also addresses these issues:
- The original `LibriMix` dataset only supports "min" mode, which means the audio length is the minimum of all clean sources. It is default for source separation task. Users may also want to use "max" mode which allows for end-to-end separation and recognition. The PR adds ``mode`` argument to let users decide which dataset they want to use.
- If the task is ``"enh_both"``, the target is the audios in ``mix_clean`` instead of separate clean sources. The PR fixes it to use ``mix_clean`` as target.

### DIFF
--- a/test/torchaudio_unittest/datasets/librimix_test.py
+++ b/test/torchaudio_unittest/datasets/librimix_test.py
@@ -3,8 +3,9 @@ import os
 from parameterized import parameterized
 
 from torchaudio.datasets import LibriMix
-from torchaudio_unittest.common_utils import get_whitenoise, normalize_wav, save_wav, TempDirMixin, TorchaudioTestCase
+from torchaudio_unittest.common_utils import get_whitenoise, save_wav, TempDirMixin, TorchaudioTestCase
 
+_SAMPLE_RATE = 8000
 _TASKS_TO_MIXTURE = {
     "sep_clean": "mix_clean",
     "enh_single": "mix_single",
@@ -13,14 +14,24 @@ _TASKS_TO_MIXTURE = {
 }
 
 
+def _save_wav(filepath: str, seed: int):
+    wav = get_whitenoise(
+        sample_rate=_SAMPLE_RATE,
+        duration=0.01,
+        n_channels=1,
+        seed=seed,
+    )
+    save_wav(filepath, wav, _SAMPLE_RATE)
+    return wav
+
+
 def get_mock_dataset(root_dir: str, num_speaker: int):
     """
     root_dir: directory to the mocked dataset
     """
     mocked_data = []
-    sample_rate = 16000
     seed = 0
-    base_dir = os.path.join(root_dir, f"Libri{num_speaker}Mix", "wav16k", "min", "train-360")
+    base_dir = os.path.join(root_dir, f"Libri{num_speaker}Mix", "wav8k", "min", "train-360")
     os.makedirs(base_dir, exist_ok=True)
     for utterance_id in range(10):
         filename = f"{utterance_id}.wav"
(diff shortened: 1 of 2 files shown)


test/torchaudio_unittest/datasets/librimix_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def get_mock_dataset(root_dir: str, num_speaker: int):
+    """
+    root_dir: directory to the mocked dataset
+    """
+    mocked_data = []
+    sample_rate = 16000
+    seed = 0
+    base_dir = os.path.join(root_dir, f"Libri{num_speaker}Mix", "wav16k", "min", "train-360")
```
mthrok, line 23: Not sure why this does not fail on Windows. The resulting `base_dir` here should be `wav16k\min\train-360\`, while the corresponding part in Dataset implementation is `wav16k/min\train-360\`.

nateanl: I see what you mean. `wav16k/min` needs to be changed to `path.join("wav16k", "min")`

mthrok: However, that makes me wonder if the original generation script works that way on Windows. If they use slash in their script then a directory with slash in it is generated. We need to check that, but I don't have access to Windows at the moment.

test/torchaudio_unittest/datasets/librimix_test.py
```diff
@@ -0,0 +1,100 @@
...
+            self.assertEqual(mixture, samples[i][task][1], atol=5e-5, rtol=1e-8)
+            assert len(sources) == len(samples[i][task][2])
+            for j in range(len(sources)):
+                self.assertEqual(sources[j], samples[i][task][2][j], atol=5e-5, rtol=1e-8)
+            num_samples += 1
+
+        assert num_samples == len(samples)
+
+    @parameterized.expand([("sep_clean"), ("enh_single",), ("enh_both",), ("sep_noisy",)])
+    def test_librimix_2spaker(self, task):
```
carolineechen, line 93:
```suggestion
    def test_librimix_2speaker(self, task):
```

test/torchaudio_unittest/datasets/librimix_test.py
```diff
@@ -0,0 +1,100 @@
...
+        dataset = LibriMix(self.root_dir, num_speakers=2, sample_rate=_SAMPLE_RATE, task=task)
+        self._test_librimix(dataset, self.samples_2spk, task)
+
+    @parameterized.expand([("sep_clean"), ("enh_single",), ("enh_both",), ("sep_noisy",)])
+    def test_librimix_3spaker(self, task):
```
carolineechen, line 98:
```suggestion
    def test_librimix_3speaker(self, task):
```

test/torchaudio_unittest/datasets/librimix_test.py
```diff
@@ -0,0 +1,100 @@
...
+        cls.samples_3spk = get_mock_dataset(cls.root_dir, 3)
+
+    def _test_librimix(self, dataset, samples, task):
+        num_samples = 0
+        for i, (sample_rate, mixture, sources) in enumerate(dataset):
+            assert sample_rate == samples[i][task][0]
+            self.assertEqual(mixture, samples[i][task][1], atol=5e-5, rtol=1e-8)
+            assert len(sources) == len(samples[i][task][2])
+            for j in range(len(sources)):
+                self.assertEqual(sources[j], samples[i][task][2][j], atol=5e-5, rtol=1e-8)
```
carolineechen, line 87: is there a reason why default atol/rtol values dont work?

nateanl: not really, I was implementing it based on `librispeech_test`. 
https://github.com/pytorch/audio/blob/0b4b1fd4296fd7d54f0059a23cf9c9e168ad9376/test/torchaudio_unittest/datasets/librispeech_test_impl.py#L70
Removing it also passes the test. Will update the pr.

carolineechen (approved):

PR #2668, opened 2022-09-12
Move hybrid demucs model out of prototype

carolineechen:


torchaudio/prototype/models/hdemucs/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+        import warnings
+
+        warnings.warn(
+            f"{__name__}.{name} has been moved to torchaudio.models.hdemucs",
+            DeprecationWarning,
+        )
+
+        from torchaudio import models
+
+        demucs = eval(f"models.{name}")
```
mthrok, line 17: `eval` is frowned upon for the security.

Doesn't `return getattr(torchaudio.models, name)` work?

mthrok (approved):

nateanl (approved):

PR #2675, opened 2022-09-14
Move conv_tasnet_base doc out of prototype

nateanl:


nateanl:
@carolineechen  do you think the `ConvTasNet` and `HDemucs` can be put under the same `Source Separation` section?

carolineechen:
I think they could belong in the same section but currently we don't have any sections for models. Might want to address this in a future PR and see if any of the existing models could belong under the same section as well, or leave as is for now.

carolineechen (approved):

PR #2678, opened 2022-09-14
Update Sphinx to 5.1.1

mthrok:
Previous versions of Sphinx reported wrong path for return class. This issue is fixed on the latest Sphinx.

It allows to remove the patch we apply in `conf.py`. This is essential for the adoptation of `:autosummary:`, as it won't render correctly with the patch.


torchaudio/models/wav2vec2/model.py
```diff
@@ -8,9 +8,7 @@
 
 
 class Wav2Vec2Model(Module):
-    """torchaudio.models.Wav2Vec2Model(feature_extractor: torch.nn.Module, encoder: torch.nn.Module, aux: Optional[torch.nn.Module] = None)
```
carolineechen, line 11: There are also instances of overriding function def for RNNT, SourceSeparationBundle, RNNTBundle, Wav2Vec2Bundle and Wav2Vec2ASRBundle, can those also be removed?

mthrok: Those overrides are different. They are there so as not to show the constructor signature, because Bundles are not meant to be instantiated by users.

carolineechen (approved):

torchaudio/models/wav2vec2/model.py
mthrok: They will be handled (automated) using `:autosummary:`

PR #2687, opened 2022-09-16
Add Speech Commands metadata function

carolineechen:


torchaudio/datasets/speechcommands.py
```diff
@@ -27,9 +28,10 @@ def _load_list(root, *filenames):
     return output
 
 
-def load_speechcommands_item(filepath: str, path: str) -> Tuple[Tensor, int, str, str, int]:
+def _get_speechcommands_metadata(filepath: str, path: str) -> Tuple[Tensor, int, str, str, int]:
```
nateanl, line 31:
```suggestion
def _get_speechcommands_metadata(filepath: str, path: str) -> Tuple[Union[str, Path], int, str, str, int]:
```

torchaudio/datasets/speechcommands.py
```diff
@@ -131,6 +132,27 @@ def __init__(
             walker = sorted(str(p) for p in Path(self._path).glob("*/*.wav"))
             self._walker = [w for w in walker if HASH_DIVIDER in w and EXCEPT_FOLDER not in w]
 
+    def get_metadata(self, n: int) -> Tuple[str, int, str, str, int]:
```
nateanl, line 135:
```suggestion
    def get_metadata(self, n: int) -> Tuple[Union[str, Path], int, str, str, int]:
```

torchaudio/datasets/speechcommands.py
```diff
@@ -131,6 +132,27 @@ def __init__(
...
+            n (int): The index of the sample to be loaded
+
+        Returns:
+            (str, int, str, str, int):
+            ``(filepath, sample_rate, label, speaker_id, utterance_number)``
+        """
+        fileid = self._walker[n]
+        return _get_speechcommands_metadata(fileid, self._archive)
+
+    def _load_waveform(self, path: str):
```
nateanl, line 149:
```suggestion
    def _load_waveform(self, path: str) -> Tensor:
```

torchaudio/datasets/speechcommands.py
```diff
@@ -131,6 +132,27 @@ def __init__(
...
+        path = os.path.join(self._archive, path)
+        waveform, sample_rate = torchaudio.load(path)
+        if sample_rate != SAMPLE_RATE:
```
mthrok, line 152: The way `SAMPLE_RATE` global variable is used looks a bit strange to me. It should be used only once in `get_metadata`, and load function should be using the value returned from `get_metadata` function, instead of referring to the global value again.

torchaudio/datasets/speechcommands.py
mthrok: Should be `str` not `Union[str, Path]`

torchaudio/datasets/speechcommands.py
mthrok: It does not look like this function ever returns `Path`.

torchaudio/datasets/utils.py
```diff
@@ -189,3 +191,15 @@ def extract_archive(from_path: str, to_path: Optional[str] = None, overwrite: bo
         pass
 
     raise NotImplementedError("We currently only support tar.gz, tgz, and zip achives.")
+
+
+def load_waveform(
```
mthrok, line 196:
```suggestion
def _load_waveform(
```

I am personally, not sure if this will be very reusable, but let's see.

mthrok (approved):

PR #2688, opened 2022-09-16
Add QUESST14 metadata function

carolineechen:


torchaudio/datasets/quesst14.py
```diff
@@ -71,10 +72,27 @@ def __init__(
...
+            (str, int, str):
+            ``(filepath, sample_rate, file_name)``
+        """
         audio_path = self.data[n]
-        wav, sample_rate = torchaudio.load(audio_path)
-        return wav, sample_rate, audio_path.with_suffix("").name
+        relpath = os.path.relpath(audio_path, self._path)
+        return relpath, SAMPLE_RATE, audio_path.with_suffix("").name
+
+    def _load_waveform(self, path: str):
```
carolineechen, line 90: @nateanl @mthrok the `_load_waveform` and `__getitem__` functions are getting a bit redundant across datasets as we start adding the `get_metadata` functions to more datasets. thoughts on having a `TorchAudioDataset` base class with default implementations for these functions? where `get_metadata` and getitem docstring will be handled by individual dataset classes

mthrok: The reuse of `_load_waveform` method makes sense, but I am not sure how the default implementation of `__getitem__` would look like as all the datasets returns tuple instead of some class that allows attribute-based access. Is the default implementation of `__getitem__` going to make an assumption about which one is waveform?

carolineechen: I think in all of our datasets, waveform is the first element of the returned Tuple (I can double check), in which case the implementation of getitem can be written as the following, although docstring would need to be overridden since that is not consistent, what do you think? If we only want to reuse `_load_waveform` and not get item, we could just add it as a utils function.
```
        metadata = self.get_metadata(n)
        waveform = self._load_waveform(metadata[0])
        return (waveform,) + metadata[1:]
```

mthrok: > I think in all of our datasets, waveform is the first element of the returned Tuple (I can double check), in which case the implementation of getitem can be written as the following, although docstring would need to be overridden since that is not consistent, what do you think? If we only want to reuse `_load_waveform` and not get item, we could just add it as a utils function.

There are some datasets that do not return waveform as the first item in the return tuple type, however even without these datasets, I do not think the assumption of first item being waveform is general enough. It does not feel well-typed.

carolineechen:
superseeded by #2697

PR #2690, opened 2022-09-17
Adopt `:autosummary:` in `torchaudio.models` module doc

mthrok:
* Introduce the mini-index at `torchaudio.models` page.

https://output.circle-artifacts.com/output/job/25e59810-3866-4ece-b1b7-8a10c7a2286d/artifacts/0/docs/models.html

Screen Shot 2022-09-20 at 1 20 50 PM

Screen Shot 2022-09-20 at 1 20 58 PM


examples/tutorials/hybrid_demucs_tutorial.py
```diff
@@ -21,10 +21,12 @@
 # 3. Collect output chunks and combine according to the way they have been
 #    overlapped.
 #
-# The `Hybrid Demucs <https://arxiv.org/pdf/2111.03600.pdf>`__ model is a developed version of the
+# The Hybrid Demucs [`Défossez, 2021<https://arxiv.org/abs/2111.03600>`__]
```
carolineechen, line 24:
```suggestion
# The Hybrid Demucs [`Défossez, 2021 <https://arxiv.org/abs/2111.03600>`__]
```

torchaudio/models/conformer.py
```diff
@@ -213,7 +213,7 @@ def forward(self, input: torch.Tensor, key_padding_mask: Optional[torch.Tensor])
 
 
 class Conformer(torch.nn.Module):
-    r"""Implements the Conformer architecture introduced in
+    r"""The Conformer architecture introduced in
```
carolineechen, line 216: nit: prefer leaving out the "The", and just having "Conformer architecture introduced in xxx". this is also not standardized across models in this PR.

carolineechen (approved):

torchaudio/models/conformer.py
mthrok: Removed "The". I am not sure aobut the later half of the sentence. (the difference between `introduced`/`from`. I wonder if "proposed" is even better) so leaving it as-is.

PR #2702, opened 2022-09-22
[Nova] Move TorchAudio Linux Wheels Nightlies over to Nova

osalpekar:
This does 2 things:
1. Comments out Linux Wheels-related jobs in CircleCI so that they are not run on nightlies/releases.
2. Adds a GHA workflow that calls the build workflow in pytorch/test-infra.

Testing:
Verified that the builds are triggered by this workflow, and all builds are green: https://github.com/pytorch/audio/actions/runs/3109635749/jobs/5040029155


osalpekar:
All builds working: https://github.com/pytorch/audio/actions/runs/3109635749/jobs/5040029155

.github/workflows/build_wheels_linux.yml
```diff
@@ -0,0 +1,39 @@
+name: Trigger Linux Wheels Build for Nightlies/Release
+
+on:
+  pull_request:
+    paths: .github/workflows/build_wheels_linux.yml
+  push:
+    branches:
+      - nightly
```
DanilBaibak, line 8: As I understand, it should be triggered by the scheduler once a day, right? Or do we have some kind of automatic push to the night branch?

seemethere: We have an automatic push from an internal cron job

seemethere (approved):

mthrok:
@osalpekar 

The `.circleci/config.yml` is not re-generated and the `circleci_consistency` is now failing.

https://app.circleci.com/pipelines/github/pytorch/audio/12735/workflows/e251aa96-edf9-42df-bddf-c9d89663b1e9

osalpekar:
> @osalpekar
> 
> The `.circleci/config.yml` is not re-generated and the `circleci_consistency` is now failing.
> 
> https://app.circleci.com/pipelines/github/pytorch/audio/12735/workflows/e251aa96-edf9-42df-bddf-c9d89663b1e9

@mthrok We'll need a small fix to the regenerate.py script that @atalman is working on. Is it alright for that job to stay broken for a day or two so we can see the Nova binaries uploaded?

mthrok:
Fine but, wouldn't be as simple as skipping [this line](https://github.com/pytorch/audio/blob/main/.circleci/regenerate.py#L89) for linux wheel?

.circleci/config.yml
```diff
@@ -1829,11 +1829,11 @@ workflows:
           python_version: '3.8'
           requires:
           - build_docs
-      - docstring_parameters_sync:
-          name: docstring_parameters_sync
-          python_version: '3.8'
-          requires:
-          - binary_linux_wheel_py3.8_cpu
+      # - docstring_parameters_sync:
```
atalman, line 1832: Since this job is required should we revert this PR ?

PR #2707, opened 2022-09-24
Add CUDA version check

hwangjeff:
Adds check to ensure that TorchAudio and PyTorch versions use the same CUDA version.

### DIFF
--- a/torchaudio/csrc/utils.cpp
+++ b/torchaudio/csrc/utils.cpp
@@ -1,5 +1,9 @@
 #include <torch/script.h>
 
+#ifdef USE_CUDA
+#include <cuda.h>
+#endif
+
 namespace torchaudio {
 
 namespace {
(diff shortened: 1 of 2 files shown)


torchaudio/csrc/utils.cpp
```diff
@@ -30,12 +34,21 @@ bool is_ffmpeg_available() {
 #endif
 }
 
+int64_t cuda_version() {
+#ifdef USE_CUDA
+  return CUDA_VERSION;
+#else
+  return -1;
```
mthrok, line 41: PyTorch seems to return `None` in case it's not present. How about doing the same here?
I think that the following would do.

```c++
c10::optional<int64_t> cuda_version() {
#ifdef USE_CUDA
   return CUDA_VERSION;
 #else
   return {};
} 
```

PR #2710, opened 2022-09-24
Fix CUDA check

hwangjeff:
`torch.version.cuda` can return a string of form X.X or X.X.X. This PR modifies the CUDA version check to account for this.


mthrok:
@hwangjeff Does this look okay https://app.circleci.com/pipelines/github/pytorch/audio/12796/workflows/ae96794a-6df4-4a2a-84df-ada7a7250045/jobs/926603 ?

PR #2712, opened 2022-09-25
Update soundfile_backend.py for mp3 support

magicse:
Update soundfile_backend.py for mp3 support with libsndfile-1.1.0-win64.zip

### DIFF
--- a/torchaudio/backend/soundfile_backend.py
+++ b/torchaudio/backend/soundfile_backend.py
@@ -117,6 +117,7 @@ _SUBTYPE2DTYPE = {
     "PCM_32": "int32",
     "FLOAT": "float32",
     "DOUBLE": "float64",
+    "MPEG_LAYER_III": "float64",
 }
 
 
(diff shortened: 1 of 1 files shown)


magicse:
Enable support for mp3 because sndlibfile already support mp3 decoding. I also made pull request with mp3 support on soundfile repository https://github.com/bastibe/python-soundfile/pull/347

mthrok:
Hi @magicse 

Thanks for the PR. What is the current state? I see that python-soundfile has release a new version, even though there are some issues you reported which need to be fixed in libsndfile.
If the fix on libsndfile is transparent, then I think we want to merge this PR. We have a release soon, so it would be nice if we can merge it before that.

magicse:
Hi it's resolved. I think You could merge. https://github.com/libsndfile/libsndfile/issues/880#issuecomment-1264130930

vadimkantorov:
> Enable support for mp3 because sndlibfile already support mp3 decoding. I also made pull request with mp3 support on soundfile repository [bastibe/python-soundfile#347](https://github.com/bastibe/python-soundfile/pull/347)

It also seems that opus support in pysoundfile can be easily enabled: https://github.com/bastibe/python-soundfile/issues/252#issuecomment-574658559

carolineechen:
Hi @magicse, thanks for looking into this, a couple of questions/comments
- I seem to be unable to reproduce this on my local machine -- still getting the unimplemented format, and mp3 does not show up under soundfile.available_formats even with version 0.11.0 installed. do you have any insight on this?
- for the issue with libsndfile, does the default of float32 currently leads to issues? if so, we may need to wait for the fix in 1.2.0 before we merge it on our end
- it would be great if we could also add a test in `soundfile/load_test.py`. lmk if you need any help with that.

cc @mthrok

mthrok:
I tested this, and it seems that with the latest soundfile release, MP3 just works without this patch. 
the following existing part handles MP3.

```python
if file_.format != "WAV" or normalize:
            dtype = "float32"
```

So I am closing the PR.

PR #2716, opened 2022-09-26
Improve initialization of wav2vec2/hubert model for pre-training

nateanl:
This PR improves the Wav2Vec2/HuBERT model regarding model pre-training.

- The model initialization of positional embedding and transformer module is essential to model pre-training. The accuracy of unmasked frames should be higher than masked frames, as it is an easier task. but without the initialization, the accuracy of masked frames is higher than unmasked frames.
  Compared the performance after two epochs with 16 GPUs.
  - With model initialization, the accuracies of masked/unmasked frames are 0.08/0.11.
  - Without model initialization, the accuracies of masked/unmasked frames are 0.06/0.04.
- After adding the model initialization, the gradient is easy to overflow (aka `nan` gradient). In paper [Self-Supervised Learning for speech recognition with Intermediate layer supervision](https://arxiv.org/abs/2112.08778) the authors propose a simple but effective method to mitigate the overflow issue, by scaling down the multiplication of query and key and subtracting the maximum value from it (subtracting a constant value won't change the output of softmax). Then it guarantees the value won't be overflowed.
- In the original fairseq, the mask indices are generated by `numpy.random.choice`. Here replace `torch.multinomial` with `torch.randperm`. (cc @carolineechen).

Other improvements within training scripts will be included in a separate PR.

### DIFF
--- a/torchaudio/models/wav2vec2/components.py
+++ b/torchaudio/models/wav2vec2/components.py
@@ -302,9 +302,9 @@ class SelfAttention(Module):
         weights = (self.scaling * q) @ k  # B, nH, L, L
         if attention_mask is not None:
             weights += attention_mask
-        # subtract a constant value from the tensor won't change the output of softmax.
+        # subtracting a constant value from the tensor won't change the output of softmax.
         # apply the subtraction to avoid value overflow in torch.nn.functional.softmax.
-        # for more details, please find Equation 7 in https://arxiv.org/abs/2112.08778
+        # for more details, please see Equation 7 in https://arxiv.org/abs/2112.08778
         weights = weights - weights.max(dim=-1, keepdim=True)[0]
 
         weights = torch.nn.functional.softmax(weights, dim=-1)
--- a/torchaudio/models/wav2vec2/components.py
+++ b/torchaudio/models/wav2vec2/components.py
@@ -206,7 +206,7 @@ class ConvolutionalPositionalEmbedding(Module):
             padding=kernel_size // 2,
             groups=groups,
         )
-        # normalize the weight to normal distribution. It is essential to model training.
+        # normalize the weight to normal distribution. it is essential to model training.
         std = math.sqrt(4.0 / (embed_dim * kernel_size))
         nn.init.normal_(self.conv.weight, mean=0.0, std=std)
         nn.init.constant_(self.conv.bias, 0.0)
@@ -275,7 +275,7 @@ class SelfAttention(Module):
         self.q_proj = nn.Linear(embed_dim, embed_dim, bias=True)
         self.out_proj = nn.Linear(embed_dim, embed_dim, bias=True)
 
-        # normalize the parameters
+        # normalize the parameters.
         nn.init.xavier_uniform_(self.k_proj.weight, gain=1 / math.sqrt(2))
         nn.init.xavier_uniform_(self.v_proj.weight, gain=1 / math.sqrt(2))
         nn.init.xavier_uniform_(self.q_proj.weight, gain=1 / math.sqrt(2))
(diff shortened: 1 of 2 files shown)


hwangjeff:
should we gate the new parameter initialization scheme with some initializer arg to avoid disrupting existing dependent training flows?

nateanl:
Since the scope of code change is within Wav2Vec2/HuBERT, and without the model initialization there will be performance regression compared to the result in fairseq. I think it is fine to make the model initialization as default. what do you think?

xiaohui-zhang:
@hwangjeff This PR is more like a bug fix to address performance degradation over the fairseq baseline. Also fortunately this recipe hasn't been adopted by others yet as far as we know. But let's definitely merge this PR after @nateanl validates WER results with changes here

mthrok:
Those are good points and I wonder if the parameter initialization can be logically separated from the implementation  view point, like putting it inside of build function, instead of model constructor. To me, even though it is commonly done, parameter initialization feels different kind from building computation graph, because it is mutation of model instance. Besides, adding that logic to constructor means that when a pre-trained models are built using `torchaudio.pipelines` module, the parameter initialization is performed even though it will be replaced right after.

torchaudio/models/wav2vec2/model.py
```diff
@@ -967,12 +989,15 @@ def hubert_pretrain_model(
         skip_masked,
         skip_nomask,
     )
-    return HuBERTPretrainModel(
+    model = HuBERTPretrainModel(
```
carolineechen, line 992: nit: not necessarily in the scope of this PR, but I noticed the docs for the arguments of HuBERTPretrainModel don't match https://github.com/pytorch/audio/pull/2716/files#diff-41adbd739db9af82fe56cc5ad4883f8e89c2667c18b53d487e683d62667fa489R140

torchaudio/models/wav2vec2/components.py
```diff
@@ -267,9 +298,14 @@ def forward(
         k = self.k_proj(x).view(*shape).permute(0, 2, 3, 1)  # B, nH, Hd, L
         v = self.v_proj(x).view(*shape).transpose(2, 1)  # B, nH, L, Hd
 
-        weights = self.scaling * (q @ k)  # B, nH, L, L
+        # scale down q to avoid value overflow.
+        weights = (self.scaling * q) @ k  # B, nH, L, L
         if attention_mask is not None:
             weights += attention_mask
+        # subtract a constant value from the tensor won't change the output of softmax.
```
carolineechen, line 305:
```suggestion
        # subtracting a constant value from the tensor won't change the output of softmax.
```

torchaudio/models/wav2vec2/components.py
```diff
@@ -267,9 +298,14 @@ def forward(
...
+        # apply the subtraction to avoid value overflow in torch.nn.functional.softmax.
+        # for more details, please find Equation 7 in https://arxiv.org/abs/2112.08778
```
carolineechen, line 307:
```suggestion
        # for more details, please see Equation 7 in https://arxiv.org/abs/2112.08778
```

carolineechen (approved):

arlofaria:
> Also fortunately this recipe hasn't been adopted by others yet as far as we know. 

Actually, I adopted this recipe before changes were introduced in this PR (and #2744). I had made some minor changes (e.g. learn k-means on 10% of the data; use gradient accumulation to simulate 32 GPUs), and my results differed slightly:

|                   | WER% (Viterbi)|  WER% (KenLM) |
|-----------------|--------------|--------------|
| test-clean (mine pre-#2744)       |       11.8     |      4.6     |
| test-clean (yours pre-#2744)      |       10.8    |       4.4     |
| test-clean (yours post-#2744)    |       10.9    |       4.4     |
| test-clean ([as published](https://arxiv.org/pdf/2106.07447.pdf))    |      N/A    |       4.3     |
| | | |
| test-other (mine pre-#2744)      |       18.7    |       9.9     |
| test-other (yours pre-#2744)     |       18.5    |       10.1    |
| test-other (yours post-#2744)   |       17.8    |       9.5     |
| test-other ([as published](https://arxiv.org/pdf/2106.07447.pdf))    |      N/A    |       9.4     |

I look forward to trying the latest version of the recipe and hopefully can replicate your results (or at least improve mine)!

nateanl:
Hi @arlofaria thanks for trying the HuBERT training recipes. Please let me know if you meet any issue when training the model with the improved recipe.

mthrok:
Hi @arlofaria 

Thanks for trying out the HuBERT recipe. The torchaudio team is wondering if we can learn about your use cases of recipe and general torchaudio. If you are open to (light weight) chat, could you email me at moto@meta.com? We highly appreciate your feedback.

Thanks,

PR #2717, opened 2022-09-27
Fix docstring sync job

osalpekar:


.circleci/regenerate.py
```diff
@@ -260,6 +260,7 @@ def unittest_workflows(indentation=6):
                             }
                         }
                     )
+    jobs += docstring_parameters_sync_job(None)
```
mthrok, line 263: This will attach the job to all the unittest jobs, which is redundant. One job should do.

mthrok (changes requested):
The actual config file needs to be regenerated.

osalpekar:
Closing in favor of backing out CircleCI changes: https://github.com/pytorch/audio/pull/2718

PR #2720, opened 2022-09-27
[Nova] New GHA Workflow for Docstring Sync

osalpekar:
Create a standalone GitHub Actions workflow for Docstring Sync. This job (https://app.circleci.com/pipelines/github/pytorch/audio/12625/workflows/96223ad2-0fcd-4dae-a045-d530aaf9b55c/jobs/907466) currently depends on linux wheels builds, which creates a dependency that makes the migration to Nova trickier. This PR creates a fresh standalone workflow for this job that is triggered per-PR and before nightly/release cuts.

### DIFF
--- a/.github/workflows/docstring_parameters_sync.yml
+++ b/.github/workflows/docstring_parameters_sync.yml
@@ -1,8 +1,6 @@
 name: Docstring Parameters Sync
 on:
   pull_request:
-    paths:
-      - .github/workflows/docstring_parameters_sync.yml
   push:
     branches:
       - nightly
--- a/.github/workflows/docstring_parameters_sync.yml
+++ b/.github/workflows/docstring_parameters_sync.yml
@@ -34,4 +34,4 @@ jobs:
         shell: bash -l {0}
         run: |
           ${CONDA_RUN} pip install pydocstyle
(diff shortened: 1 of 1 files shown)


mthrok (approved):
LGTM. 

@carolineechen the issue with HuBERT you mentioned, should it be caught with this??

osalpekar:
> I think the default convention for pydocstyle (pep257) doesn't catch it (error D417), according to this http://www.pydocstyle.org/en/stable/error_codes.html. The google convention seems more comprehensive, and I think it's also possible to set up custom configuration to determine which errors we want to catch and which ones to allow.
> 
> Also, for some reason, running `pip install pydocstyle; pydocstyle torchaudio` locally never produces any error for me, even if I purposely mess up the docstrings. Not sure if I'm missing something, but could be good to double check and test on a faulty file to ensure errors are reported from this workflow.

@carolineechen My main purpose for this PR is to achieve parity with the corresponding job currently in CircleCI, so we can gradually deprecate CircleCI safely - I've kept the main scripts largely the same as that job. Happy to iteratively improve this workflow in future PR's

seemethere (approved):

PR #2722, opened 2022-09-28
Follow up removing cuda102

atalman:
Follow up removing cuda102. Need to regenerate to config

Original pr included removal of these config: https://github.com/pytorch/audio/pull/2715


nateanl:
The unit test still uses cuda 10.2 https://github.com/pytorch/audio/blob/main/.circleci/config.yml.in#L652

Could you address that?

PR #2728, opened 2022-09-30
Add citation for MuST-C dataset in Emformer RNNT pipeline.

nateanl:
The MuST-C reference is added in https://github.com/pytorch/audio/pull/2689. This PR adds the citation to the RNNT pipeline documentation.


nateanl:
I'm thinking of re-training the Emformer models including `LibriSpeech`, `TED-LIUM`, and `MuST-C` datasets using the updated conformer training recipe built by @hwangjeff. Then update the pre-trained weights correspondingly.
What would be the best strategy to make updates on the pre-trained weights for the pipelines? @hwangjeff @mthrok

carolineechen (approved):

PR #2732, opened 2022-10-04
Add IEMOCAP dataset

carolineechen:


torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,140 @@
...
+        relative_paths.append(wav_path)
+    return relative_paths
+
+
+class IEMOCAP(Dataset):
+    """*IEMOCAP* :cite:`iemocap` dataset.
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        sessions (Optional[List[int]], optional): List of sessions (1-5) to use. If ``None``, loads from all sessions.
```
mthrok, line 31: Didn't we standardize on the following format?

```suggestion
        sessions (List[int] or None, optional): List of sessions (1-5) to use. If ``None``, loads from all sessions.
```

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,140 @@
...
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        sessions: Optional[str] = None,
```
mthrok, line 37: The type annotation does not match with docstring.

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,140 @@
...
+    ):
+        root = os.fspath(root)
```
mthrok, line 39: I know many implementations do this, but can we rather make it `root = Path(root)`? IIO, that will make the code for path management easier to read. If we want to return string type from any method, the conversion can happen right before it's returned to client code.

torchaudio/datasets/iemocap.py
mthrok: nit: Changing the annotation to `sessions: Tuple[str] = (1, 2, 3, 4, 5)` would allow to get rid of `None` placeholder.

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,140 @@
...
+                all_data.append(wav_stem)
+
+            # add labels
+            label_dir = os.path.join(session_dir, "dialog/EmoEvaluation")
+            label_paths = Path(label_dir).glob("*.txt")
+
+            for label_path in label_paths:
+                with open(label_path, "r") as f:
+                    for line in f:
+                        if line[0] != "[":
```
mthrok, line 68: Could we use `startwith` just for peace of mind? If the file ends with a line without break, then this part could fail. (though I know that's not expected)

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,140 @@
...
+
+            for wav_path in wav_paths:
+                wav_stem = str(Path(wav_path).stem)
+                if wav_stem in self.mapping:
+                    self.data.append(wav_stem)
+                    self.mapping[wav_stem]["path"] = wav_path
+
+    def get_metadata(self, n: int) -> Tuple[str, int, str, str, str]:
+        """Get metadata for the n-th sample from the dataset. Returns filepath instead of waveform,
+        but otherwise returns the same fields as :py:func:`__getitem__`.
```
mthrok, line 90:
```suggestion
        but otherwise returns the same fields as :py:meth:`__getitem__`.
```

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,138 @@
...
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found
+        sessions (List[int]): List of sessions (1-5) to use. (Default: ``[1, 2, 3, 4, 5]``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        sessions: List[str] = [1, 2, 3, 4, 5],
```
mthrok, line 37: You cannot use List here. List is mutable. My suggestion was Tuple, which is immutable. These are different.

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,138 @@
...
+            session_dir = os.path.join(self._path, session_name)
+
+            # get wav paths
+            wav_paths = _get_wavs_paths(session_dir)
+            for wav_path in wav_paths:
+                wav_stem = str(Path(wav_path).stem)
+                all_data.append(wav_stem)
+
+            # add labels
+            label_dir = os.path.join(session_dir, "dialog/EmoEvaluation")
```
mthrok, line 60: > `"dialog/EmoEvaluation"`

How does this work in windows?

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,138 @@
...
+    ):
+        root = Path(root)
+        self._path = os.path.join(root, "IEMOCAP")
+
+        if not os.path.isdir(self._path):
+            raise RuntimeError("Dataset not found.")
+
+        all_data = []
+        self.data = []
+        self.mapping = defaultdict()
```
mthrok, line 47: Could you explain why the use of `defaultdict`?

It seems that each value of `self.mapping` is initialized to empty dict at L77 `self.mapping[wav_stem] = {}`, then the existence of the entry is checked as `if wav_stem in self.mapping:`, so it seems no advantage to use `defaultdict`.

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,138 @@
...
+from torch import Tensor
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import _load_waveform
+
+
+SAMPLE_RATE = 16000
+
+
+def _get_wavs_paths(data_dir):
+    wav_dir = os.path.join(data_dir, "sentences/wav")
```
mthrok, line 16: Does the slash here work on Windows?

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,138 @@
...
+        sessions (List[int]): List of sessions (1-5) to use. (Default: ``[1, 2, 3, 4, 5]``)
+    """
+
+    def __init__(
+        self,
+        root: Union[str, Path],
+        sessions: List[str] = [1, 2, 3, 4, 5],
+    ):
+        root = Path(root)
+        self._path = os.path.join(root, "IEMOCAP")
```
mthrok, line 40: Now that the `root` is Path object, can we get rid of `os.path.join` and use  `root / "IEMOCAP"` ?

torchaudio/datasets/iemocap.py
carolineechen: regular dict works, thanks

torchaudio/datasets/iemocap.py
carolineechen: good catch, fixed this

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,137 @@
...
+import re
+from pathlib import Path
+from typing import Tuple, Union
+
+from torch import Tensor
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import _load_waveform
+
+
+SAMPLE_RATE = 16000
```
mthrok, line 11: Does this need to be a global variable? Can we make it a private at least?

mthrok (approved):

nateanl (approved):

torchaudio/datasets/iemocap.py
```diff
@@ -0,0 +1,137 @@
...
+        return (wav_path, SAMPLE_RATE, wav_stem, label, speaker)
+
+    def __getitem__(self, n: int) -> Tuple[Tensor, int, str, str, str]:
+        """Load the n-th sample from the dataset.
+
+        Args:
+            n (int): The index of the sample to be loaded
+
+        Returns:
+            Tuple of the following items;
```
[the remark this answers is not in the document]
carolineechen: this is consistent with all other dataset docstrings, which use a semicolon. feel free to make a PR to change them to colons, I don't have a strong opinion on this

PR #2734, opened 2022-10-04
Increase CircleCi no_output_timeout for `install binaries` steps

izaitsevfb:
The goal is to to reduce the number of job failures due to timeouts, see https://app.circleci.com/pipelines/github/pytorch/audio/12882/workflows/f99da1a5-32e6-4bac-8ceb-fbf36d693e2d/jobs/936363?invite=true#step-105-105 for example.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -448,6 +448,7 @@ jobs:
       - load_conda_channel_flags
       - run:
           name: install binaries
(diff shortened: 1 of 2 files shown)


weiwangmeta (approved):
Thanks for the fix!

atalman (approved):

PR #2737, opened 2022-10-05
Add precise seek

jdsgomes:
cc @mthrok 

Implements precise seek and seek to any frame in torchaudio


torchaudio/csrc/ffmpeg/stream_reader/stream_reader.h
```diff
@@ -100,6 +100,12 @@ class StreamReader {
   // Retrieval
   //////////////////////////////////////////////////////////////////////////////
   std::vector<c10::optional<torch::Tensor>> pop_chunks();
+
+ private:
+  //////////////////////////////////////////////////////////////////////////////
+  // Instance variables
+  //////////////////////////////////////////////////////////////////////////////
+  int64_t seek_timestamp = -1;
```
mthrok, line 108: Could you move this to the section at the top of class definition where the rest of other variables are defined?

torchaudio/csrc/ffmpeg/stream_reader/stream_reader.cpp
```diff
@@ -163,12 +163,42 @@ bool StreamReader::is_buffer_ready() const {
...
 // Configure methods
 ////////////////////////////////////////////////////////////////////////////////
-void StreamReader::seek(double timestamp) {
+void StreamReader::seek(double timestamp, int64_t mode) {
   TORCH_CHECK(timestamp >= 0, "timestamp must be non-negative.");
+  TORCH_CHECK(
+      pFormatContext->nb_streams > 0,
+      "At least one stream must exist in this context");
+
+  AVRational time_base = pFormatContext->streams[0]->time_base;
```
mthrok, line 172: What happens if `stream[0]` is non-active (`AVDISCARD_ALL `) stream?

torchaudio/csrc/ffmpeg/stream_reader/stream_processor.cpp
```diff
@@ -82,7 +84,11 @@ int StreamProcessor::process_packet(AVPacket* packet) {
       return send_frame(NULL);
     if (ret < 0)
       return ret;
-    send_frame(pFrame1);
+
+    if (pFrame1->pts >= discard_before_pts)
+      send_frame(pFrame1);
```
mthrok, line 89: Please use parenthesis even if it is one-line. FBcode linter will complain later.

torchaudio/csrc/ffmpeg/stream_reader/stream_reader.cpp
jdsgomes: I tested it and still returns the correct time base for this stream so the seek will still work.

torchaudio/csrc/ffmpeg/stream_reader/stream_processor.cpp
```diff
@@ -82,7 +84,12 @@ int StreamProcessor::process_packet(AVPacket* packet) {
       return send_frame(NULL);
     if (ret < 0)
       return ret;
-    send_frame(pFrame1);
+
+    if (pFrame1->pts >= discard_before_pts) {
+      send_frame(pFrame1);
```
mthrok, line 89: Actually, I think that the discard must happen after the frame goes through FilterGraph here https://github.com/pytorch/audio/blob/main/torchaudio/csrc/ffmpeg/stream_reader/sink.cpp#L84

StreamReader pass the AVFrame to FilterGraph before converted to Tensor, and FilterGraph can be stateful depending on the type of filter. For example, when changing FPS with [`fps` filter](https://ffmpeg.org/ffmpeg-filters.html#fps-1), a frame before the target time might be duplicated to show later which exceeds the target time.

See the following for the process flow. 
https://github.com/pytorch/audio/tree/main/torchaudio/csrc/ffmpeg#streamreader-architecture

torchaudio/csrc/ffmpeg/stream_reader/stream_reader.cpp
```diff
@@ -163,12 +163,42 @@ bool StreamReader::is_buffer_ready() const {
...
+  TORCH_CHECK(timestamp_s >= 0, "timestamp must be non-negative.");
+  TORCH_CHECK(
+      pFormatContext->nb_streams > 0,
+      "At least one stream must exist in this context");
+
+  int64_t timestamp_av_tb = static_cast<int64_t>(timestamp_s * AV_TIME_BASE);
+  // rescale timestamp to the default stream timebase
+  AVRational default_tb = pFormatContext->streams[0]->time_base;
+  int64_t timestamp_default_tb =
+      av_rescale_q(timestamp_av_tb, AV_TIME_BASE_Q, default_tb);
```
mthrok, line 176: I am not sure why this is necessary to compute PTS on the first stream here. Isn't it more straightforward to pass discard target time stamp as second and let each StreamProcessor converts it to pts of the corresponding stream, then let Sink class discard the frame output from FilterGraph by comparing it?

jdsgomes: This is just for the initial seek (`av_seek_frame`).

mthrok: Why are we manually converting it to the time base of the **first stream** (which is not necessarily default stream)?

https://ffmpeg.org/doxygen/4.1/group__lavf__decoding.html#gaa23f7619d8d4ea0857065d9979c75ac8

According to the doc, `av_seek_frame` selects the default stream and do the conversion automatically with `stream_index=-1`. Is there a case that this is undesired?

mthrok (commented):
I think this PR is almost good. Can we have a test that checks the correct seek behavior?

torchaudio/csrc/ffmpeg/stream_reader/stream_reader.cpp
jdsgomes: you are right.

mthrok:
For testing correctness, using binary files as reference is not a sustainable approach for the following reasons;

1. Binary files are not browsable. Inspecting it requires writing a script, which is tedious and error-prune.
2. The binary files are not always possible to re-create in the future.
3. It looks like that decoded reference frames are checked-in. This means, it was decoded by some decoder implementation, which can be different from the decoder used during the test. Different decoders might produce the different result, even if the test logic is correct.
4.  It increases the size of source distribution.

Therefore, in torchaudio, we check-in binary files only when it's absolutely necessary. I think checking-in an AVI file might be necessary to test for ANY seek. (see below)

As an alternative, what we can do is to check whether the decoded frames after seek match frames decoded without seek.

The gist of the logic is;

1. decode the entire video and store the frames as reference.
2. for each seek test, decode frames after seeking and compare the resulting frames against reference frames by extracting the corresponding part.

For example, the video file at `test/torchaudio_unittest/assets/nasa_13013.mp4` has (only) two key frames; at the beginning and at 8.08 second (which is 203rd frame from the beginning). See the analysis [here](https://docs.google.com/spreadsheets/d/1sao95kVl4UihdfdbtBOGHl57oE3qm10RCkQio62ONsM/edit#gid=0). (this was obtained with `ffprobe test/torchaudio_unittest/assets/nasa_13013.mp4 -print_format 'csv' -select_streams v:0  -show_entries frame=pict_type,key_frame,pict_type,pkt_pts_time,pkt_dts_time,best_effort_timestamp_time,pkt_duration_time`)

Using this information, we can check whether the seek is working fine by slicing the reference frames at the index that corresponds to the seek timestamp.

The following code snippet implements such test logic.

```python
import torch
import torchaudio

src = 'test/torchaudio_unittest/assets/nasa_13013.mp4'

# Using the first video stream (which is not default video stream)
stream_index = 0


# Decode all frames for reference
s = StreamReader(src)
s.add_basic_video_stream(-1, stream_index=stream_index)
s.process_all_packets()
ref_frames, = s.pop_chunks()
print(ref_frames.shape)


def test(mode, seek_time, ref_indices):
    s = StreamReader(src)
    s.add_basic_video_stream(-1, stream_index=stream_index)
    s.seek(seek_time, mode=mode)
    s.process_all_packets()
    frame, = s.pop_chunks()
    print(frame.shape)

    hyp_index, ref_index = ref_indices
    torch.testing.assert_close(frame[hyp_index:], ref_frames[ref_index:])


configs = [
    # Test keyframe seek
    # The source video has two key frames the first frame and 203rd frame at 8.08 second.
    # If the seek time stamp is smaller than 8.08, it will seek into the first frame at 0.0 second.
    ("key", 0.2, (0, 0)),
    ("key", 8.04, (0, 0)),
    ("key", 8.08, (0, 202)),
    ("key", 8.12, (0, 202)),

    # Test precise seek
    ("precise", 0.0, (0, 0)),
    ("precise", 0.2, (0, 5)),
    ("precise", 8.04, (0, 201)),
    ("precise", 8.08, (0, 202)),
    ("precise", 8.12, (0, 203)),

    # MP4 decoder does not really support ANY.
    # we need an AVI file. Perhaps we check-in an AVI?

]

for conf in configs:
    print(conf)
    test(*conf)
```

mthrok:
We need to test, the combination of audio and filter graph, in addition to the script I shared above, but I think testing video-only seek behavior is enough for this PR.

jdsgomes:
@mthrok Thank you for the suggestion, I think it makes total sense. 
I added the tests and a avi video to test the "any" seek behaviour. My only issue is that with "any" seek I can only test equality for the i-frames. I will add some tests for delta frames to make sure they are different.

test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -418,22 +418,82 @@ def test_stream_smoke_test(self):
...
+        s.add_basic_video_stream(-1, stream_index=stream_index)
+        s.process_all_packets()
+        (ref_frames,) = s.pop_chunks()
+
+        s.seek(seek_time, mode=mode)
+        s.process_all_packets()
+        (frame,) = s.pop_chunks()
+
+        hyp_index, ref_index = ref_indices
+        if assert_different:
```
mthrok, line 492: I don't think this `assert_different` branch is necessary.
For "any" seek cases that result in starting with delta frames, setting `hyp_index` to a number of frames to hit the first key frame (that is a relative number of frames) should do.

mthrok:
> @mthrok Thank you for the suggestion, I think it makes total sense. I added the tests and a avi video to test the "any" seek behaviour.

Thanks. Regarding the new video, I am not comfortable to check-in this video for privacy reason as it contains a minor with face, even if the copyright allows to copy it. Could you replace it with something else? Perhaps convert the `nasa_13013.mp4` into AVI format?

> My only issue is that with "any" seek I can only test equality for the i-frames. I will add some tests for delta frames to make sure they are different.

For any seek that produces rubbish frames at the beginning, by spec, there is no reference we can have for these beginning frames. Those are like NaN values we cannot compare against anything. So  I don't think we can test the beginning frames, and we should not compare them against anything.

test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -418,22 +418,81 @@ def test_stream_smoke_test(self):
...
+            ("nasa_13013.avi", "precise", 0.0, (0, 0)),
+            ("nasa_13013.avi", "precise", 0.2, (0, 1)),
+            ("nasa_13013.avi", "precise", 8.1, (0, 238)),
+            ("nasa_13013.avi", "precise", 8.14, (0, 239)),
+            ("nasa_13013.avi", "precise", 8.17, (0, 240)),
+            # Test any seek
+            # The source avi video has one keyframe every twelve frames 0, 12, 24,.. or every 0.4004 seconds.
+            ("nasa_13013.avi", "any", 0.0, (0, 0)),
+            ("nasa_13013.avi", "any", 0.56, (0, 12)),
+            ("nasa_13013.avi", "any", 7.77, (0, 228)),
```
mthrok, line 474: Can you add cases like?

```
            ("nasa_13013.avi", "any", 0.2002, (11, 12)),
            ("nasa_13013.avi", "any", 0.233567, (10, 12)),
            ("nasa_13013.avi", "any", 0.266933, (9, 12)),
```

mthrok (approved):
Looks good. Couple of more test cases for any and we can merge this. We use ShipIt, so please land it via phabricator.

torchaudio/io/_stream_reader.py
```diff
@@ -425,13 +425,34 @@ def get_out_stream_info(self, i: int) -> torchaudio.io.StreamReaderOutputStream:
         """
         return _parse_oi(self._be.get_out_stream_info(i))
 
-    def seek(self, timestamp: float):
+    def seek(self, timestamp: float, mode: str = "key"):
```
mthrok, line 428: Actually, I think we should change the default mode to "precise". That's the behavior I intended in the original implementation and that's what most DL practitioners would expect in their use cases.

Then we can cherry-pick this into the release as bug fix/new feature.

test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -418,22 +418,84 @@ def test_stream_smoke_test(self):
...
+        s.process_all_packets()
+        (ref_frames,) = s.pop_chunks()
+
+        s.seek(seek_time, mode=mode)
+        s.process_all_packets()
+        (frame,) = s.pop_chunks()
+
+        hyp_index, ref_index = ref_indices
+
+        torch.testing.assert_close(frame[hyp_index:], ref_frames[ref_index:])
```
mthrok, line 497: Can we use `self.assertEqual` here?

mthrok:
Note: follow-ups

* test more cases where filters are involved, check the imprecation to other streams (such as audio)
* cache the timestamp in filtergraph/sink object so that tiemstamp computation is only done once instead of every time a frame is processed.

test/torchaudio_unittest/io/stream_reader_test.py
jdsgomes: sure will make the change now

jdsgomes: n00b question how to import the latest changes to Phabricator?

mthrok: You can either click `Import to fbsource` button again or click `update` button from [here](https://fburl.com/opensource/cpz5v10b).

test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -418,22 +418,84 @@ def test_stream_smoke_test(self):
...
+            ("nasa_13013.avi", "any", 0.0, (0, 0)),
+            ("nasa_13013.avi", "any", 0.56, (0, 12)),
+            ("nasa_13013.avi", "any", 7.77, (0, 228)),
+            ("nasa_13013.avi", "any", 0.2002, (11, 12)),
+            ("nasa_13013.avi", "any", 0.233567, (10, 12)),
+            ("nasa_13013.avi", "any", 0.266933, (9, 12)),
+        ]
+    )
+    def test_seek_modes(self, src, mode, seek_time, ref_indices):
+        """Testing expected behaviour for the different kinds of seek"""
```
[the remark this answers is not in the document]
jdsgomes: added a description here sine there are some other changes that need to go in this PR due to the failing tests internally.

One of them is regarding the gallery examples:

https://github.com/pytorch/audio/blob/main/examples/tutorials/streamreader_advanced_tutorial.py#L449 

I need to check what is going on tomorrow but my first intuition is that it is related to doing "precise" seek in a video that is streamed.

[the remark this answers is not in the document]
mthrok: Quickly looking into it, at `frame->pts >= discard_before_pts` in `Sink::process_frame`, the value of `discard_before_pts ` is `360000` while the value of `frame->pts` range is few thousand at most, thus the stream gets exhausted.

mthrok:
@jdsgomes I looked into the issue with the tutorial. The root cause is that the AVFrames returned by AVFilterGraph does not have the same `time_base` / `pts` as the input frames. (`pts` starts from 0 and `time_base` corresponds to the output frame rate.) I searched for a way to fetch the relevant information from AVFilterGraph struct, but it does not seem there is a way. Therefore it is difficult to apply discard mechanism on AVFilterGraph. So I think we can discard frames after decoder, which is what you were suggesting originally. I tested it and it passes tests and tutorial builds fine.

PR #2738, opened 2022-10-06
Add Snips Dataset

nateanl:


torchaudio/datasets/snips.py
```diff
@@ -0,0 +1,103 @@
...
+            int:
+                Sample rate
+            str:
+                Transcription of audio
+        """
+        audio_path = self.data[n]
+        relpath = os.path.relpath(audio_path, self._path)
+        file_name = audio_path.with_suffix("").name
+        transcripte = self.transcripts[file_name]
+        return relpath, SAMPLE_RATE, transcripte
```
carolineechen, line 80:
```suggestion
        transcripts = self.transcripts[file_name]
        return relpath, SAMPLE_RATE, transcripts
```

nateanl:
cc @leo19941227
The original format of Snips dataset is ``".mp3"``, hence it is set as default. But it looks like the [official s3 link](https://github.com/aws-samples/aws-lex-noisy-spoken-language-understanding) is expired. The copy of the dataset hosted by s3prl is processed and has ``".wav"`` format. If users want to use the copy hosted by s3prl, they need to manually set `_ext_audio` to ``".wav"`` before initializing the dataset.

nateanl:
It's not intuitive to initialize the dataset with `_ext_audio` class variable set as ".wav" . I decided to use ``ext_audio`` as an argument for initialization. When users want to initialize the dataset with ".wav" audios, they can write the code like
```python
from torchaudio.datasets import Snips
dataset = Snips(root, "train", ext_audio=".wav")
```

torchaudio/datasets/snips.py
```diff
@@ -0,0 +1,129 @@
...
+
+class Snips(Dataset):
+    """*Snips* :cite:`coucke2018snips` dataset.
+
+    Args:
+        root (str or Path): Root directory where the dataset's top level directory is found.
+        subset (str): Subset of the dataset to use. Options: [``"train"``, ``"valid"``, ``"test"``].
+        speakers (List[str] or None, optional): The speaker list to include in the dataset. If ``None``,
+            include all speakers in the subset. (Default: ``None``)
+        ext_audio (str, optional): The extention of the audios. (Default: ``".mp3"``)
```
carolineechen, line 51: can you include that the options are wav and mp3?

also slight pref if the "." was not included in the ext_audio variable, but rather done in the string processing afterwards. and maybe a name like `format`/`audio_format` would be better?

carolineechen (approved):

test/torchaudio_unittest/datasets/snips_test.py
```diff
@@ -0,0 +1,120 @@
...
+    def testSnipsTrain(self):
+        dataset = snips.Snips(self.root_dir, subset="train", audio_format="wav")
+        self._testSnips(dataset, self.train_samples)
+
+    def testSnipsValid(self):
+        dataset = snips.Snips(self.root_dir, subset="valid", audio_format="wav")
+        self._testSnips(dataset, self.valid_samples)
+
+    def testSnipsTest(self):
+        dataset = snips.Snips(self.root_dir, subset="test", audio_format="wav")
```
carolineechen, line 119: is it possible to add tests for mp3 format?

nateanl: The audio saving method under unit test uses `scipy.io.wavfile.write`, which doesn't support mp3 format I think.

leo19941227:
> @leo19941227 The transcript contains special tokens like `BOS`, `EOS`, `playlist_owner`, etc. For example:
> 
> ```
> BOS ESTELLE SHOULD BE ON MY SPRING PLAYLIST EOS\tO artist O O O playlist_owner playlist O AddToPlaylist
> ```
> 
> Is it desired for ctc training?

In this example, transcription and iob tagging are separated by the '\t', and they will be handled by the tokenizer to be encoded into suitable format for CTC training. Here is the tokenizer:

https://github.com/s3prl/s3prl/blob/6ad8ca7449befd6ef478756da011442a7552b2af/s3prl/downstream/ctc/text.py#L108

So, you can actually parse our corpus into some more intuitive format like for each dataset point there are:

transcription, iob labels (per word), intent label, wav_path

leo19941227:
Note that the final iob tagging (the tagging on the EOS) is the intent label for the sentence.

nateanl:
I see. So in the above example, the returned items are:
- transcript: `BOS ESTELLE SHOULD BE ON MY SPRING PLAYLIST EOS`
- iob_label: `O artist O O O playlist_owner playlist O AddToPlaylist`
- intent_label: `AddToPlaylist`
- wav_path

Is that correct?

leo19941227:
Yes, or you can remove the BOS and EOS if the intent label has a dedicated field. The BOS's tagging is always O, and the EOS's tagging is the intent label. Removing BOS and EOS might make the transcription looks more nature.

nateanl:
Sounds good. Since `BOS` and `EOS` are not used in the training. I will remove those from the transcript and iob_label.

PR #2741, opened 2022-10-07
[Nova] [WIP] Exclude test for StreamReader

DanilBaibak:
[Nova] [WIP] Exclude test for StreamReader. Can be an optional solution for MacOS only.


mthrok (changes requested):
This  test is super important to check the compatibility of separately installed FFmpeg.

So this is undesirable.

Can you explain the context?

malfet:
@DanilBaibak if this is just a work-in-progress PR, can you please prefix it with `WIP` or publish those as drafts?

osalpekar:
Done testing for this case, we can close this now.

PR #2744, opened 2022-10-09
Improve hubert recipe for pre-training and fine-tuning

nateanl:
following pr #2716
- For preprocessing
  - The HuBERT feature takes lots of memory which may not fit some machines. Enable to use a subset of feature for training a k-means model.

- For pre-training
  - Normalize the loss based on the total number of masked frames across all GPUs.
  - Use mixed precision training. fp16 is not well supported in pytorch_lightning.
  - Log accuracies of masked/unmasked frames during training.
  - Clip the gradients with norm `10.0`.

- For ASR fine-tuning
  - Normalize the loss based on the total number of batches across all GPUs, same as in the conformer recipe of TorchAudio.
  - Use mixed precision training.
  - Add "|" after the end of transcription to capture the silence/word termination, same as in fairseq recipe.

- Update the WER results on LibriSpeech dev and test sets.

|                   | WER% (Viterbi)|  WER% (KenLM) |
|:-----------------:|--------------:|--------------:|
| dev-clean         |       10.9    |       4.2     |
| dev-other         |       17.5    |       9.4     |
| test-clean        |       10.9    |       4.4     |
| test-other        |       17.8    |       9.5     |

### DIFF
--- a/examples/hubert/evaluate.py
+++ b/examples/hubert/evaluate.py
@@ -1,6 +1,6 @@
 import argparse
 import logging
-from typing import Dict, List, Optional
+from typing import Dict, List
 
 import torch
 import torch.nn.functional as F
@@ -36,10 +36,9 @@ def _viterbi_decode(emission: torch.Tensor, id2token: Dict, blank_idx: int = 0) 
     Returns:
         (List of str): The decoding result. List of string in lower case.
     """
-    hypothesis = F.log_softmax(emission, dim=-1)
-    hypothesis = hypothesis.argmax(-1).unique_consecutive()
+    hypothesis = emission.argmax(-1).unique_consecutive()
     hypothesis = hypothesis[hypothesis != blank_idx]
-    hypothesis = "".join(id2token[int(i)] for i in hypothesis).replace("|", " ")
+    hypothesis = "".join(id2token[int(i)] for i in hypothesis).replace("|", " ").strip()
     return hypothesis.split()
 
 
@@ -47,7 +46,7 @@ def _ctc_decode(emission, decoder: CTCDecoder) -> List[str]:
     """Run CTC decoding with a KenLM language model.
 
     Args:
-        emission (torch.Tensor): Output of CTC layer. Tensor with dimensions (..., time, num_tokens).
+        emission (torch.Tensor): Output of CTC layer. Tensor with dimensions `(..., time, num_tokens)`.
         decoder (CTCDecoder): The initialized CTCDecoder.
 
     Returns:
@@ -55,13 +54,19 @@ def _ctc_decode(emission, decoder: CTCDecoder) -> List[str]:
     """
     hypothesis = decoder(emission)
     hypothesis = hypothesis[0][0].words
+    hypothesis = [word for word in hypothesis if word != " "]
     return hypothesis
 
 
 def run_inference(args):
+    if args.use_gpu:
+        device = torch.device("cuda")
+    else:
+        device = torch.device("cpu")
+
     # Load the fine-tuned HuBERTPretrainModel from checkpoint.
     model = _load_checkpoint(args.checkpoint)
-    model.eval()
+    model.eval().to(device)
 
     if args.use_lm:
         # get decoder files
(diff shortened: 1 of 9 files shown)


examples/hubert/evaluate.py
```diff
@@ -138,8 +144,8 @@ def _parse_args():
     )
     parser.add_argument(
         "--beam-size-token",
-        type=Optional[int],
-        default=None,
+        type=int,
+        default=29,
         help="Number of tokens to consider at each beam search step. (Default: None)",
```
carolineechen, line 149:
```suggestion
        help="Number of tokens to consider at each beam search step. (Default: 29)",
```

examples/hubert/finetune.py
```diff
@@ -101,6 +105,18 @@ def _parse_args():
         required=True,
         help="Path to the LibriSpeech and LibriLightLimited datasets.",
     )
+    parser.add_argument(
+        "--checkpoint",
+        type=str,
+        required=True,
+        help="Path to the pre-trained HuBERTPretrainModel checpoint as the initialization.",
```
carolineechen, line 112:
```suggestion
        help="Path to the pre-trained HuBERTPretrainModel checkpoint as the initialization.",
```

examples/hubert/preprocess.py
```diff
@@ -60,6 +60,12 @@ def _parse_args():
         type=int,
         help="The number of clusters for KMeans clustering.",
     )
+    parser.add_argument(
+        "--percent",
+        default=-1,
+        type=float,
+        help="The number of clusters for KMeans clustering.",
```
carolineechen, line 67: should this be something like this:

```suggestion
        help="The percent of data for KMeans clustering. If negative, use all data. (Default: -1)",
```

examples/hubert/train.py
```diff
@@ -90,7 +92,7 @@ def _parse_args():
     )
     parser.add_argument(
         "--resume-checkpoint",
-        type=Optional[pathlib.Path],
+        type=pathlib.Path,
```
carolineechen, line 95: This should remain `Optional[pathlib.Path]`

nateanl: actually using `Optional` in argparse will throw an error, that's why I removed those from the parser.

carolineechen (approved):

arlofaria:
Looking at this now, I have a few thoughts...

> Normalize the loss based on the total number of masked frames across all GPUs.

Could you help me to understand the rationale for this change?  From my understanding, the prior code was doing a reasonable thing: a GPU that processes fewer masked frames should contribute less to the DDP-averaged gradient than a GPU that processes more masked frames.  Moreover, normalizing the loss now means that the predetermined learning rates from the original HuBERT paper might be on the wrong scale; from what I can tell of the paper and the Fairseq code, it doesn't seem like per-masked-frame loss normalization was used, right?

However, for logging purposes, the normalized loss can be very convenient as it allows experiments with different batch sizes to have their curves compared on the same scale.

> Use mixed precision training. fp16 is not well supported in pytorch_lightning.

Perhaps that's a bug that has since been fixed?  I tried today with `lightning==1.8.3` and setting `Trainer(precision=16)` seems to work correctly (same loss curves per step, but at faster convergence per walltime).

> Log accuracies of masked/unmasked frames during training.

Very nice, thanks for the improvement!  However, why are the stats reset only after every epoch (at the end of the validation step), rather than reset for every training step?  This leads to somewhat confusing curves that tend toward the cumulative epoch-level accuracy for steps that are late in an epoch, but are generally under-reporting the more instantaneous step-level accuracy.

> Clip the gradients with norm 10.0.

(nit) This is configurable for pre-training, but not for fine-tuning where it is hardcoded to 5.0.

nateanl:
> it doesn't seem like per-masked-frame loss normalization was used, right?

It actually does in fairsq, not in hubert loss, but in the general fairseq trainer. Here is the code point:
https://github.com/facebookresearch/fairseq/blob/c19aed8ef59acafc8c146b50aed0ff3275ffcf01/fairseq/trainer.py#L932-L946

> Perhaps that's a bug that has since been fixed?

Yes, after fixing the embedding dtype, pytorch_lightning's fp16 API can also be used.

> However, why are the stats reset only after every epoch (at the end of the validation step), rather than reset for every training step?

This is to estimate more reasonable accuracy for each epoch. It indeed introduces cumulative accuracy for each step which has momentum of previous steps. maybe a better way is to report the step accuracy based on the current batch, and use a separate variable to log the total number of count and accuracy sum for the epoch-level accuracies.

> This is configurable for pre-training, but not for fine-tuning where it is hardcoded to 5.0.

Yes, hyperparamters is always hard to tune. In my experiments 10 seems reasonable for pre-training, but 5.0 works better for fine-tuning.

arlofaria:
> It actually does in fairsq, not in hubert loss, but in the general fairseq trainer.

Oh, you're absolutely right -- thanks for looking into this and clarifying the reason for the loss normalization!

PR #2745, opened 2022-10-09
Add metadata mode for LibriMix

carolineechen:


torchaudio/datasets/librimix.py
```diff
@@ -55,45 +55,62 @@ def __init__(
...
         elif sample_rate == 16000:
-            self.root = self.root / "wav16k" / mode / subset
+            mix_dir = self.root / "wav16k" / mode / subset
         else:
             raise ValueError(f"Unsupported sample rate. Found {sample_rate}.")
         self.sample_rate = sample_rate
         self.task = task
-        self.mix_dir = (self.root / _TASKS_TO_MIXTURE[task]).resolve()
+
+        self.mix_dir = mix_dir / _TASKS_TO_MIXTURE[task]
```
carolineechen, line 66: @nateanl do you know why `.resolve()` was used here and if it's ok to remove? including it causes some tests to fail because it sometimes adds `\private\` to the filepath and messes up with the relative path for finding the metadata filepath.

nateanl: `.resolve()` is to convert relative path like `..` to absolute path. This is based on wsj0-2mix dataset under examples folder, so @mthrok may know the reason behind it. I feel it can be removed, as other datasets don't rely on `resolve` method.

torchaudio/datasets/librimix.py
```diff
@@ -55,45 +54,62 @@ def __init__(
...
+            str:
+                Path to mixed audio
+            List of str:
+                List of paths to source audios
+        """
+        filename = self.files[key]
+        mixed_path = os.path.relpath(self.mix_dir / filename, self.root)
+        srcs_paths = []
+        for dir_ in self.src_dirs:
+            src = os.path.relpath(dir_ / filename, self.root)
```
nateanl, line 105: `self.root` is removed in this pr. could you add it back?

nateanl: apparently it's not. my bad.

carolineechen: `self.root` is still there, but its value is changed. should I leave the original self.root value and rename this to something else to retain BC in case someone is using the self.root variable?

nateanl (approved):

torchaudio/datasets/librimix.py
nateanl: We didn't make `self.root` as a public variable, so it's not necessary to retain BC IMO.

PR #2756, opened 2022-10-11
Add 0.13 docs and version selector

carolineechen:
https://carolineechen.github.io/audio/0.13.0/


nateanl:
The prototype features need to be removed from the docs too.

nateanl (approved):
LGTM

PR #2762, opened 2022-10-12
Add custom lm example to decoder tutorial

carolineechen:


nateanl:
@atalman  should the timeout also be increased for `build_docs` job?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -28,21 +28,22 @@
...
 #
 # -  Acoustic Model: model predicting phonetics from audio waveforms
 # -  Tokens: the possible predicted tokens from the acoustic model
 # -  Lexicon: mapping between possible words and their corresponding
 #    tokens sequence
-# -  KenLM: n-gram language model trained with the `KenLM
-#    library <https://kheafield.com/code/kenlm/>`__
+# -  Language Model (LM): n-gram language model trained with the `KenLM
+#    library <https://kheafield.com/code/kenlm/>`__, or custom languge
+#    model wrapped using :py:class:`~torchaudio.models.decoder.CTCDecoderLM`
```
mthrok, line 40: Nit: “wrap” sounds like a class that has the said class as attribute is also compatible. Perhaps “inherit” is unambiguous?

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -159,6 +158,23 @@
...
+# factoring in a language model score that represents the likelihood of
+# the sequence into the beam search computation. Below, we outline the
+# different forms of language models that are supported for decoding.
+# 
+
+######################################################################
+# No Language Model
+# ^^^^^^^^^^^^^^^^^
+#
+# To train without a language model, set `lm=None` when initializing
```
mthrok, line 175: Without a language model, is decoder object still something trainable?
“To create a decoder instance without a language model” sounds more intuitive. (I’m reviewing from my phone and didn’t check the scrounging context that is not changed so I might be missing something’s here

mthrok (approved):

examples/tutorials/asr_inference_with_ctc_decoder_tutorial.py
```diff
@@ -172,6 +189,52 @@
...
+        with torch.no_grad():
+            score = self.language_model(self.sil)
+
+        self.states[state] = score
+        return state
+
+    def score(self, state: CTCDecoderLMState, token_index: int):
+        outstate = state.child(token_index)
+        if outstate not in self.states:
+            score = self.language_model(token_index)
```
nateanl, line 229: I'm not sure of the input argument is suitable for neural language model. Usually a neural language model takes `Tensor` as input which represents the token indices in the sentence. Should we verify the type of `language_model` and pass different types for `token_index`?

Issue #2743, opened 2022-10-09
GTZAN dataset cannot be downloaded properly


volkovwashere:
### 🐛 Describe the bug

# The issue
When trying to download the GTZAN dataset from torchaudio, the download fails every time due to connection timeout.

Snippet:
```
import torchaudio
ds = torchaudio.datasets.GTZAN(
    "", download=True
)
```

Error msg:
`URLError: <urlopen error [Errno 110] Connection timed out>`

Whereas every other torchaudio dataset downloads just fine.
My guess would be that probably the default url: http://opihi.cs.uvic.ca/sound/genres.tar.gz  is  outdated / not working anymore as I cant even access it with wget. 
Thanks in advance.

### Versions

PyTorch version: 1.12.1+cu102
Is debug build: False
CUDA used to build PyTorch: 10.2
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.3 LTS (x86_64)
GCC version: (Ubuntu 9.4.0-1ubuntu1~20.04.1) 9.4.0
Clang version: Could not collect
CMake version: Could not collect
Libc version: glibc-2.31

Python version: 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 06:04:18)  [GCC 10.3.0] (64-bit runtime)
Python platform: Linux-4.14.290-217.505.amzn2.x86_64-x86_64-with-glibc2.10
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

Versions of relevant libraries:
[pip3] numpy==1.23.3
[pip3] torch==1.12.1
[pip3] torchaudio==0.12.1
[pip3] torchvision==0.13.1
[conda] numpy                     1.23.3                   pypi_0    pypi
[conda] torch                     1.12.1                   pypi_0    pypi
[conda] torchaudio                0.12.1                   pypi_0    pypi
[conda] torchvision               0.13.1                   pypi_0    pypi

fszatkowski:
I have the same problem

carolineechen:
Hi, thanks for bringing this to our attention. The issue lies with the original dataset download link, which is outside our control. We'll keep an eye out on this to see if it is resolved anytime soon, or else we may need to find a replacement download link if there is one available, or remove the download support for this dataset.

[...]

PR #2763, opened 2022-10-12
Add gtzan download note

carolineechen:
GTZAN download link is no longer working, so the torchaudio download functionality for GTZAN does not work properly, per #2743. Add a note in the docs to reflect this discovery.

### DIFF
--- a/torchaudio/datasets/gtzan.py
+++ b/torchaudio/datasets/gtzan.py
@@ -1004,7 +1004,7 @@ class GTZAN(Dataset):
 
     Note:
         As of October 2022, the download link is not currently working. Setting ``download=True``
-        in this case will result in a URL connection error.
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/gtzan.py
```diff
@@ -1002,6 +1002,10 @@ class GTZAN(Dataset):
         Please see http://marsyas.info/downloads/datasets.html if you are planning to use
         this dataset to publish results.
 
+    Note:
+        As of October 2022, the download link is not currently working. Setting ``download=True``
+        in this case will result in a URL connection error.
```
nateanl, line 1007:
```suggestion
        in GTZAN dataset will result in a URL connection error.
```

nateanl (approved):

mthrok (approved):

PR #2764, opened 2022-10-13
Update tutorial author information

mthrok:
Adding and updating author information.


carolineechen (commented):
looks good, could you also update other @fb.com emails to @meta.com?

carolineechen (approved):

PR #2773, opened 2022-10-15
Update resampling tutorial

mthrok:
* Refactor benchmark script
* Rename `time` variable to avoid (potential) conflicting with time module
* Fix `beta` parameter in benchmark (it was not used previously)
* Use `timeit` module for benchmark
* Add plot
* Move the comment on result at the end
* Add link to an explanation of aliasing

https://output.circle-artifacts.com/output/job/20b57d2f-3614-4161-a18e-e0c1a537739c/artifacts/0/docs/tutorials/audio_resampling_tutorial.html

### DIFF
--- a/examples/tutorials/audio_resampling_tutorial.py
+++ b/examples/tutorials/audio_resampling_tutorial.py
@@ -22,19 +22,14 @@ print(torchaudio.__version__)
 #
 # First, we import the modules and define the helper functions.
 #
-# .. note::
-#    When running this tutorial in Google Colab, install the required packages
-#    with the following.
-#
-#    .. code::
-#
-#       !pip install librosa
 
 import math
-import time
+import timeit
 
 import librosa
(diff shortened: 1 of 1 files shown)


carolineechen (approved):
Screen Shot 2022-10-16 at 1 30 54 PM
This graph is a bit hard to read because of how small the green blocks are (I guess it's good transforms is that fast haha), and there's no blue bar for the left two either. Not sure if there's a good solution to this though so PR looks good

mthrok:
> Screen Shot 2022-10-16 at 1 30 54 PM
> 
> This graph is a bit hard to read because of how small the green blocks are (I guess it's good transforms is that fast haha), and there's no blue bar for the left two either. Not sure if there's a good solution to this though so PR looks good

Good point. Added labels.

Screen Shot 2022-10-17 at 10 34 06 AM

PR #2776, opened 2022-10-18
Add notes on file structure in Voxceleb1 based datasets

nateanl:
The file structure of VoxCeleb1 is as follows:
```
root/
└── wav/
    └── speaker_id folders
```
Users who use [Kaldi](https://github.com/kaldi-asr/kaldi/blob/f6f4ccaf213f0fe8b26e633a7dc0c802150626a0/egs/voxceleb/v1/local/make_voxceleb1_v2.pl) to get the VoxCeleb1 dataset have "dev" and "test" folders above "wav" folder. However, in the file lists like https://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/veri_test.txt or https://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/iden_split.txt there is not such differentiation. It's not necessary to put the extracted files into separate folders.

This PR adds notes in `VoxCeleb1Identification` and `VoxCeleb1Verification` datasets to inform the file structure to users.

### DIFF
--- a/torchaudio/datasets/voxceleb1.py
+++ b/torchaudio/datasets/voxceleb1.py
@@ -138,14 +138,15 @@ class VoxCeleb1Identification(VoxCeleb1):
 
     Note:
         The file structure of `VoxCeleb1Identification` dataset is as follows:
-            root/
 
-            └── wav/
+        └─ root/
 
-              └── speaker_id folders
+         └─ wav/
 
-            Users who pre-downloaded the ``"vox1_dev_wav.zip"`` and ``"vox1_test_wav.zip"`` files need to move
-            the extracted files into the same ``root`` directory.
+         └─ speaker_id folders
+
+        Users who pre-downloaded the ``"vox1_dev_wav.zip"`` and ``"vox1_test_wav.zip"`` files need to move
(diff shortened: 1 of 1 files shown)


torchaudio/datasets/voxceleb1.py
```diff
@@ -135,6 +135,17 @@ class VoxCeleb1Identification(VoxCeleb1):
             (Default: ``"https://www.robots.ox.ac.uk/~vgg/data/voxceleb/meta/iden_split.txt"``)
         download (bool, optional):
             Whether to download the dataset if it is not found at root path. (Default: ``False``).
+
+    Note:
+        The file structure of `VoxCeleb1Identification` dataset is as follows:
+            root/
+
```
carolineechen, line 142: nit: would it look better if the blank lines are removed? do you have the rendered output for this

carolineechen (approved):

torchaudio/datasets/voxceleb1.py
carolineechen: i see, what if you remove one of the `_`, to make it a bit narrower?

carolineechen (approved):

PR #2778, opened 2022-10-18
Add iemocap variants

carolineechen:
- Add `utterance_type` argument in the initialization method, which supports loading only improvised or only scripted utterances, or all utterances.
- In the previous version, the emotion label ``"exc"`` is converted to ``"hap"``, the utterances with "fru" label are filtered out. The PR changes the dataset to return utterances with raw labels that belong to [``"neu"``, ``"hap"``, ``"ang"``, ``"sad"``, ``"exc"``, ``"fru"``].


nateanl (approved):
LGTM. cc @leo19941227 Is there other usecase of Iemocap dataset?

leo19941227:
Cool! I replied in the feedback google doc.

PR #2782, opened 2022-10-21
[Nova] Added build wheels workflow for MacOS

DanilBaibak:
Added build wheels workflow for MacOS.

This does not change the existing builds/uploads in CircleCI, and should not break any existing jobs/workflows. This is just to add back workflows to build the MacOS Wheels with Nova.

We will create a workflow (most likely in test-infra) that does this comparison between the binaries to ensure there is parity between the binaries before we start uploading with Nova.


.github/workflows/build_wheels_macos.yml
```diff
@@ -0,0 +1,49 @@
+name: Trigger MacOS Wheels Build for Nightlies/Release
+
+on:
+  pull_request:
+    paths:
```
osalpekar, line 5: Let's run this on all PR's

osalpekar (approved):
LGTM!

hwangjeff (approved):

Issue #2783, opened 2022-10-21
Git repository is very big


ezyang:
### 🐛 Describe the bug

torchaudio repository is over 7GB large. What the heck?!

```
Cloning into 'audio'...                              
remote: Enumerating objects: 64834, done.            
remote: Counting objects: 100% (727/727), done.                                                           
remote: Compressing objects: 100% (218/218), done.                                                        
remote: Total 64834 (delta 498), reused 719 (delta 496), pack-reused 64107                                
Receiving objects: 100% (64834/64834), 7.63 GiB | 17.26 MiB/s, done.  
```

Running git-fatfiles, I see a tar.gz repeatedly committed in the repo:

https://stackoverflow.com/questions/1029969/why-is-my-git-repository-so-big

```
68480538 main/artifact.tar.gz                                                                             
68481054 main/artifact.tar.gz                        
68481265 main/artifact.tar.gz                                                                             
68481817 main/artifact.tar.gz                 
68482409 main/artifact.tar.gz                        
68483799 main/artifact.tar.gz                    
68484136 main/artifact.tar.gz              
68484923 main/artifact.tar.gz                       
68485008 main/artifact.tar.gz                        
68485772 main/artifact.tar.gz                                                                             
68486117 main/artifact.tar.gz                                                                             
68486657 main/artifact.tar.gz                                                                             
68487304 main/artifact.tar.gz  
```

Please don't commit tarballs into the repo, and please figure out if you can prune these objects from the repository (if they live in a branch, this should be relatively easy; if their on the mainline we will have to history rewrite.)

### Versions

master

nateanl:
@malfet is right. The tar file is committed into `gh-pages` branch unexpectedly. I will make a fix now.

malfet:
@nateanl mind if I'll just nuke part of gh-pages history?

nateanl:
No problem. Thanks for help.

malfet:
Wow, it has been going on for the last 5 month...

nateanl:
Awesome. Thanks @ezyang for sharing the issue and @malfet for the quick fix. I will close the issue when the tar file is removed in the nightly job.

PR #2786, opened 2022-10-21
Remove archive file in gh-pages branch

nateanl:
The motivation of generating `artifact.tar.gz` in the `build_docs` job is to easily use it for adding documentation in each stable release. But it is committed into `gh-pages` branch which causes the git repository very huge (see #2783). This PR removes the tar file from the commit.


malfet (approved):

carolineechen (approved):

mthrok:
@nateanl Thanks for handling this.

PR #2797, opened 2022-10-26
Deprecate 'onesided' init param for MelSpectrogram

hwangjeff:
Initializer parameter `onesided` isn't relevant to `MelSpectrogram` — it should always be `True`. In fact, the module already assumes `onesided == True` in the filterbank it generates and fails in its forward pass when `onesided == False`. Accordingly, this PR makes param `onesided` optional and adds a deprecation warning that's fired when the param is provided.

### DIFF
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -566,8 +566,6 @@ class MelSpectrogram(torch.nn.Module):
             (Default: ``True``)
         pad_mode (string, optional): controls the padding method used when
             :attr:`center` is ``True``. (Default: ``"reflect"``)
-        onesided (bool, optional): controls whether to return half of results to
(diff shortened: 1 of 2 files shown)


torchaudio/transforms/_transforms.py
```diff
@@ -599,11 +597,17 @@ def __init__(
         wkwargs: Optional[dict] = None,
         center: bool = True,
         pad_mode: str = "reflect",
-        onesided: bool = True,
```
mthrok, line 602: We need to keep the argument here, otherwise client code will break if it gives all the arguments as positional arguments.

Some strategy we employed in the past is to change the default value to None first, and issue a warning if the actual value is not None.
Then once we complete one release, we actually remove it.

carolineechen (approved):

PR #2798, opened 2022-10-26
Enable log probs input for rnnt loss

carolineechen:
Add `fused_log_softmax` argument (default/current behavior = True) to rnnt loss. 

If setting it to `False`, call `log_softmax` on the logits prior to passing it in to the rnnt loss function.

The following should produce the same output:
```
rnnt_loss(logits, targets, logit_lengths, target_lengths, fused_log_softmax=True)
```

```
log_probs = torch.nn.functional.log_softmax(logits, dim=-1)
rnnt_loss(log_probs, targets, logit_lengths, target_lengths, fused_log_softmax=False)
```

testing -- unit tests + get same results on the conformer rnnt recipe

### DIFF
--- a/test/torchaudio_unittest/common_utils/rnnt_utils.py
+++ b/test/torchaudio_unittest/common_utils/rnnt_utils.py
@@ -188,7 +188,7 @@ def compute_with_numpy_transducer(data):
     return costs, gradients
 
 
-def compute_with_pytorch_transducer(data):
+def compute_with_pytorch_transducer(data, fused_log_softmax=True):
     costs = rnnt_loss(
         logits=data["logits"],
         logit_lengths=data["logit_lengths"],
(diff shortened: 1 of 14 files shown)


xiaohui-zhang (approved):
LGTM. thanks @carolineechen for addressing this quickly to unblock @BriansIDP !

test/torchaudio_unittest/common_utils/rnnt_utils.py
```diff
@@ -189,13 +189,17 @@ def compute_with_numpy_transducer(data):
 
 
 def compute_with_pytorch_transducer(data):
+    fused_log_softmax = data["fused_log_softmax"] if "fused_log_softmax" in data else True
```
mthrok, line 192: If `data` is `dict`.

```suggestion
    fused_log_softmax = data.get("fused_log_softmax", True)
```

mthrok (approved):

PR #2799, opened 2022-10-26
Add back docstring for MelSpectrogram param onesided

hwangjeff:
Adds back docstring for `MelSpectrogram` initializer param `onesided`.


torchaudio/transforms/_transforms.py
```diff
@@ -566,6 +566,8 @@ class MelSpectrogram(torch.nn.Module):
             (Default: ``True``)
         pad_mode (string, optional): controls the padding method used when
             :attr:`center` is ``True``. (Default: ``"reflect"``)
+        onesided (bool or None, optional): controls whether to return half of results to
```
mthrok, line 569: Since this argument nothing but a placeholder for the sake of backward compatibility, let's remove the type annotation and just say "Deprecated and unused". No need to mention default value either because default value is there just to detect if user passed anything following the previous convention.

PR #2800, opened 2022-10-27
Add nova workflow for MacOS and Linux

DanilBaibak:
Added missed build workflows for MacOS and Linux:

- [x] Linux conda
- [x] MacOS conda

This does not change the existing builds/uploads in CircleCI, and should not break any existing jobs/workflows. This is just to add back workflows for the MacOS and Linux conda builds with Nova.

We will create a workflow (most likely in test-infra) that does this comparison between the binaries to ensure there is parity between the binaries before we start uploading with Nova.


.github/workflows/build_conda_linux.yml
```diff
@@ -0,0 +1,46 @@
+name: Trigger Linux Conda Build for Nightlies/Release
```
osalpekar, line 1: nit: can you rename this as "Build Linux Conda", and same below? See the job names we've used in text and vision as reference.

osalpekar (approved):
LGTM

.github/workflows/build_conda_linux.yml
osalpekar: This can be a follow-up PR, feel free to check this one in

osalpekar:
Build statuses look great - let's import this as a diff and land it!

PR #2809, opened 2022-10-31
Fix convolve mode docstring

hwangjeff:
Argument `mode` in `convolve` and `fftconvolve` is expected to be a string, but the docstrings incorrectly say bool. This PR fixes the docstrings accordingly.


mthrok (approved):

mthrok:
BTW regarding the docstring, isn't `abs(M-N)` easier to understand than `max(N, M) - min(N, M)`?

hwangjeff:
about the same to me — could go either way. got the current one from numpy docs

Issue #2686, opened 2022-09-16
torchaudio.io._compat.load_audio_fileobj returns shorter waveform than torchaudio.load for certain FLAC files


nateanl:
### 🐛 Describe the bug

When using the two loading methods on the same audio file, the lengths of the waveform tensors are different.

I can reproduce this issue with one audio file from LibriSpeech dataset, note that the issue only exists in some FLAC files. Here is the code to reproduce it:
```python
import torchaudio
waveform_load, _ = torchaudio.load("LibriSpeech/train-other-500/5350/205002/5350-205002-0014.flac")
print(waveform_load.shape)
with open("LibriSpeech/train-other-500/5350/205002/5350-205002-0014.flac", "rb") as f:
    waveform_file, _= torchaudio.io._compat.load_audio_fileobj(f)
prinyt(waveform_file.shape)
```
The output is:
```
torch.Size([1, 194160])
torch.Size([1, 184320])
```

### Versions

PyTorch version: 1.13.0.dev20220812
Is debug build: False
CUDA used to build PyTorch: None
ROCM used to build PyTorch: N/A

OS: macOS 12.5.1 (x86_64)
GCC version: Could not collect
Clang version: 13.1.6 (clang-1316.0.21.2.5)
CMake version: version 3.22.1
Libc version: N/A

Python version: 3.10.4 (main, Mar 31 2022, 03:38:35) [Clang 12.0.0 ] (64-bit runtime)
Python platform: macOS-10.16-x86_64-i386-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
GPU models and configuration: No CUDA
Nvidia driver version: No CUDA
cuDNN version: No CUDA
HIP runtime version: N/A
MIOpen runtime version: N/A

Versions of relevant libraries:
[pip3] numpy==1.22.4
[pip3] pytorch-sphinx-theme==0.0.24
[pip3] torch==1.13.0.dev20220812
[pip3] torchaudio==0.13.0a0+a42b266
[pip3] torchroomacoustics==0.0.1
[conda] blas                      1.0                         mkl  
[conda] mkl                       2021.2.0                 pypi_0    pypi
[conda] numpy                     1.22.4                   pypi_0    pypi
[conda] pytorch                   1.13.0.dev20220718        py3.10_0    pytorch-nightly
[conda] pytorch-sphinx-theme      0.0.24                    dev_0    <develop>
[conda] torch                     1.13.0.dev20220812          pypi_0    pypi
[...]

Issue #2356, opened 2022-04-28
Reading FLAC files from file-like objects is broken?


philgzl:
### 🐛 Describe the bug

Calling `torchaudio.load` with a file-like object as argument to load FLAC-encoded files sometimes returns incomplete tensors. It seems to happen with non-trivial tensors after a certain length.

# Minimal sample code:
Below I conduct the following test: I write a given input tensor on the filesystem using `torchaudio.save` in FLAC format, and then load it with `torchaudio.load` with a file-like object as argument. I repeat this with different tensors with different shape. The shape of the loaded tensor is written in comments; it does not always match the input tensor!

```python
import torch
import torchaudio


fs = 16000
filename = 'example.flac'


def test(input):
    # write data
    torchaudio.save(filename, input, fs)

    # load data that was just saved using file object
    f = open(filename, 'rb')
    output, _ = torchaudio.load(f)

    # check the output shape
    print(int(input.shape == output.shape), output.shape)


torch.manual_seed(0)

# non-random vectors as int or float work fine
test(torch.zeros(1, 1600, dtype=torch.int))          # correct
test(torch.ones(1, 1600, dtype=torch.float))         # correct
test(10*torch.ones(2, 16000, dtype=torch.int))       # correct
test(-100*torch.ones(2, 160000, dtype=torch.float))  # correct

# short random vectors also work fine (0.1s @ 16kHz)
test(torch.randn(1, 1600))  # correct
test(torch.randn(2, 1600))  # correct
test(torch.rand(1, 1600))   # correct
test(torch.rand(2, 1600))   # correct

# longer random vectors produce wrong shapes! (1s @ 16kHz)
test(torch.randn(1, 16000))  # wrong, torch.Size([1, 4096]), raises warning "flac: decoder MD5 checksum mismatch."
test(torch.randn(2, 16000))  # wrong, torch.Size([2, 0])
test(torch.rand(1, 16000))   # wrong, torch.Size([1, 8192]), raises warning "flac: decoder MD5 checksum mismatch."
test(torch.rand(2, 16000))   # wrong, torch.Size([2, 0])

# maybe staying in [-1.0, 1.0] fixes it?
test(0.01*torch.randn(1, 16000))  # correct
test(0.01*torch.randn(2, 16000))  # wrong, torch.Size([2, 0])
test(0.01*torch.rand(1, 16000))   # correct
test(0.01*torch.rand(2, 16000))   # wrong, torch.Size([2, 0])

[...]

PR #2810, opened 2022-10-31
Make buffer size configurable in ffmpeg file object operations and set size in backend

hwangjeff:
Partly addresses #2686 and #2356.

Currently, when the buffer used for file object decoding is insufficiently large, `torchaudio.load` returns a shorter waveform than expected. To deal with this, the user is expected to increase the buffer size via `torchaudio.utils.sox_utils.get_buffer_size`, but this does not influence the buffer used by the FFMpeg fallback. To fix this, this PR introduces changes that apply the buffer size set for the SoX backend to FFMpeg.

As a follow-up, we should see whether it's possible to programmatically detect that the buffer's too small and flag it to the user.

### DIFF
--- a/torchaudio/backend/sox_io_backend.py
+++ b/torchaudio/backend/sox_io_backend.py
@@ -4,6 +4,7 @@ from typing import Optional, Tuple
 import torch
 import torchaudio
 from torchaudio._internal import module_utils as _mod_utils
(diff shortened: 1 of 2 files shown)


mthrok (approved):
The proposed change looks good to me. Could you update the PR description to include how this change is addressing the linked issues?

PR #2812, opened 2022-11-01
Add links to training recipes

mthrok:
Screen Shot 2022-11-01 at 3 32 58 PM

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -54,6 +54,16 @@ model implementations and application components.
 
 .. toctree::
    :maxdepth: 1
+   :caption: Training Recipes
+   :hidden:
+
+   Conformer RNN-T ASR <https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_conformer_rnnt>
+   Emformer RNN-T ASR <https://github.com/pytorch/audio/tree/main/examples/asr/emformer_rnnt>
+   Conv-TasNet Source Separation <https://github.com/pytorch/audio/tree/main/examples/source_separation>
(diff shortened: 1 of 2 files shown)


carolineechen (approved):
I feel like direct linking can be a little bit confusing because it brings them to the github page while the rest of the docs stay within the torchaudio website, but I think it does the job

docs/source/index.rst
```diff
@@ -52,6 +52,16 @@ model implementations and application components.
...
    tutorials/hybrid_demucs_tutorial
 
+.. toctree::
+   :maxdepth: 1
+   :caption: Training Recipes
+   :hidden:
+
+   Conformer RNN-T ASR <https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_conformer_rnnt>
+   Emformer RNN-T ASR <https://github.com/pytorch/audio/tree/main/examples/asr/emformer_rnnt>
+   ConvTas-Net Source Separation <https://github.com/pytorch/audio/tree/main/examples/source_separation>
```
nateanl, line 62: nit: people usually call it "Conv-TasNet" or "ConvTasNet"
```suggestion
   Conv-TasNet Source Separation <https://github.com/pytorch/audio/tree/main/examples/source_separation>
```

nateanl (approved):

docs/source/index.rst
mthrok: oops, that's what I meant. Fixed

mthrok:
> I feel like direct linking can be a little bit confusing because it brings them to the github page while the rest of the docs stay within the torchaudio website, but I think it does the job

Yeah, I am feeling in a similar way. It would be nice to have a minimalistic introduction, but then we won't have the recipe listed in in the left side bar. I opted for listing recipes and making them more noticeable for now.

Issue #2103, opened 2021-12-26
Bark Filterbank for torchaudio


ahmed-fau:
### 🚀 The feature

Is there any plan/interest to enable Bark spectrogram calculation in torchaudio? 

### Motivation, pitch

More flexibility to users of torchaudio especially for ML-DSP purposes

mthrok:
Hi @ahmed-fau 

Thanks for the request. Extending torchaudio in DSP domain is generally our interest.
However I am new to Bark scale. Would you recommend any learning material?

After quick googling, and reading https://www.fon.hum.uva.nl/praat/manual/BarkSpectrogram.html,
it seems that the procedure looks like the following.

Waveform -> power spectrogram -> Bark scale conversion

So adding Bark Filterbank (+ optionally, BarkSpectorgram) will suffice.
Is that what you had in your mind?

ahmed-fau:
Hi @mthrok 

Exactly, it's the same interface of MelSpectrogram but with a different Psychoacoustic scale (Bark instead of Mel, so adding Bark filterbank is all that we need).

The Bark scale is recently used in efficient neural speech synthesis models such as [LPCNet](https://arxiv.org/pdf/1810.11846.pdf).

For the sake of completeness, you can also add another argument for the [ERB (equivalent rectangular bandwidth)](https://ccrma.stanford.edu/~jos/bbt/Equivalent_Rectangular_Bandwidth.html) scale, which is also used in recent neural speech enhancement systems such as [PercepNet](https://arxiv.org/pdf/2008.04259.pdf)

leonmck-dsp-dev:
hi 
i can implement this let me grab a look

mthrok:
Hi @ahmed-fau 

Sorry for the late response, but if you are still available, feel free to make a PR.

underdogliu:
So just come across tis post - so are we talking about RASTA-related spectrogram energies since we talk about bark scaling?

jdariasl:
Hi @mthrok, it's been a while since your comment about making a PR to include the Bark Spectrogram into torchaudio, and there has not been any response. I have implemented it, may I make a PR?

carolineechen:
Added as a prototype feature in #2823 and #2843

PR #2821, opened 2022-11-01
BarkSpectrogram

jdariasl:
I have added BarkScale transform, which can transform a regular Spectrogram into a BarkSpectrograms similar to MelScale. @ahmed-fau opened this requirement in December 2021 with the number (https://github.com/pytorch/audio/issues/2103). The new functionality includes three different well-known approximations of the Bark scale.


carolineechen:
Hi @jdariasl, thanks for the PR! please mark it as "Ready for review" when you think the feature is complete, and I'll try to take a look and review it within the week

PR #2822, opened 2022-11-02
Add WavLM model

sgrigory:
Closes T136364380

Added [WavLM Model](https://github.com/microsoft/UniSpeech/tree/main/WavLM):
- Added `WavLMSelfAttention` class (from [original implementation](https://github.com/microsoft/UniSpeech/blob/2e9dde8bf815a5f5fd958e3435e5641f59f96928/WavLM/modules.py)) and adjusted existing Encoder and Transformer classes to be compatible with it
- Added factory functions `wavlm_model`, `wavlm_base`, `wavlm_large` to `models/wav2vec2/model.py`
- ~~Added bundles for base and large models to pipelines.**TODO**: pre-trained model weights are not yet uploaded to `download.pytorch.org`, permissions not granted yet.~~ Split this to a follow-up PR #2833 

## Tests
- Expanded HuggingFace integration tests to cover WavLM. For there tests, added JSON configs for base and large models from HF ([base](https://huggingface.co/microsoft/wavlm-base/blob/main/config.json), [large](https://huggingface.co/microsoft/wavlm-large/blob/main/config.json)) into test assets
- Expanded TorchScript and quantization tests to cover WavLM

## Comments
There are a few workarounds I had to introduce:
- Quantization tests for WavLM were breaking down at [`torch.cat`](https://github.com/pytorch/audio/pull/2822/files#diff-6f1486901c94320ec0610a460dc674638fab9d104a61564ff7b59353a8b8547cR466) ~~until I excluded the arguments of `torch.cat` from quantization [here](https://github.com/pytorch/audio/pull/2822/files#diff-6f1486901c94320ec0610a460dc674638fab9d104a61564ff7b59353a8b8547cR368-R369). I haven't found a better way to fix it, let me know if there is one~~ The reason for this seems to be that quantization replaces `.bias` and `.weight` attributes of a `Linear` module with methods. Since we are using weights and biases directly, the code was break. The final solution suggested by @nateanl was to define attention weights and biases directly in `WavLMSelfAttention`, skipping the `Linear` layers
- ~~WavLM uses position embedding in the first layer of encoder, but not in the subsequent ones.  So [UniSpeech](https://github.com/microsoft/UniSpeech/blob/2e9dde8bf815a5f5fd958e3435e5641f59f96928/WavLM/modules.py#L342) and [HF](https://github.com/huggingface/transformers/blob/b047472650cba259621549ac27b18fd2066ce18e/src/transformers/models/wavlm/modeling_wavlm.py#L441-L442) implementations only create this embedding module in the layers where it's used. However, we can't do this here because it breaks TorchScript. So as a solution I add a dummy `Identity` module to `WavLMSelfAttention` when the actual embedding is not needed: [here](https://github.com/pytorch/audio/pull/2822/files#diff-6f1486901c94320ec0610a460dc674638fab9d104a61564ff7b59353a8b8547cR361-R368).~~ Thanks @nateanl for resolving this!
- I had to add dummy `position_bias` and `key_padding_mask` arguments to `SelfAttention.forward` to make TorchScript tests pass. Since both `SelfAttention` and `WavLMSelfAttention` are called from `EncoderLayer`, they need to have compatible signatures. Having a variable number of arguments with `**kwargs` or checking object class doesn't seem to work with TorchScript, so I instead made both types of attention accept `position_bias` and `key_padding_mask` arguments.

Nit: do we still need to specify `__all__` if there are no wildcard imports in `__init__.py`, e.g. in `torchaudio/models/__init__.py`?


torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
-        return output
+        return output, None  # Necessary for compatibility with WavLMSelAttention
+
+
+class WavLMSelfAttention(nn.Module):
+    """Multi-headed self-attention for WavLM model :cite:`wavlm2021`.
+    Source: https://github.com/microsoft/UniSpeech/blob/2e9dde8bf815a5f5fd958e3435e5641f59f96928/WavLM/modules.py
+
+    Args:
+        embed_dim: Total dimension of the model.
```
nateanl, line 329: could you add the type for each argument? Something like
`embed_dim (int): Total dimension of the model.`

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
+        num_heads: The number of heads.
+        dropout: Dropout probability on attn_output_weights.
+        bias: If ``True``, add bias to projections for queries and values.
```
nateanl, line 332: for argument with default value, the type will be `(type, optional)`. could you also add the default value at the end of docstring?
`bias (bool, optional): If ``True``, add bias to projections for queries and values. (Default: ``True``)`

torchaudio/models/wav2vec2/components.py
```diff
@@ -752,6 +946,85 @@ def _get_encoder(
     return Encoder(feature_projection, transformer)
 
 
+def _get_encoder_wavlm(
```
nateanl, line 949: nit: will name it `_get_wavlm_encoder` better?

torchaudio/models/wav2vec2/components.py
```diff
@@ -752,6 +946,85 @@ def _get_encoder(
...
+    layer_norm_first: bool,
+    layer_drop: float,
+) -> Encoder:
+    """
+    Construct encoder for WavLM model :cite:`wavlm2021`. The structure of the encoder and most of the argments are the
+    same as in :py:func:`_get_encoder` so refer there for documentation. The only difference from Wav2Vec2 encoder is
+    usage of `WavLMSelfAttention` instead of `SelfAttention` and two additional parameters: `num_buckets`
+    and `max_distance`.
+    Args:
+        in_features (int): See :py:func:`_get_encoder`.
```
nateanl, line 972: `_get_encoder` should be `_get_wavlm_encoder` for the following arguments.

torchaudio/models/wav2vec2/model.py
```diff
@@ -1216,3 +1216,219 @@ def hubert_pretrain_xlarge(
...
+    encoder_ff_interm_features: int,
+    encoder_ff_interm_dropout: float,
+    encoder_dropout: float,
+    encoder_layer_norm_first: bool,
+    encoder_layer_drop: float,
+    aux_num_out: Optional[int],
+) -> Wav2Vec2Model:
+    """Builds custom WaveLM model :cite:`wavlm2021`. The architecture is compatible
+    with Wav2Vec2 model :cite:`baevski2020wav2vec`, and so the output object is
+    :class:`~torchaudio.models.Wav2Vec2Model`. Most of the argument have the same meaning
```
nateanl, line 1243:
```suggestion
    :class:`~torchaudio.models.Wav2Vec2Model`. Most of the arguments have the same meaning
```

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -1177,3 +1180,90 @@ def _get_state_dict(self, dl_kwargs):
 
 Please refer to :py:class:`torchaudio.pipelines.Wav2Vec2ASRBundle` for the usage.
 """  # noqa: E501
+
+
+WAVLM_BASE = Wav2Vec2Bundle(
```
nateanl, line 1185: could you add another pipeline named `WAVLM_BASE_PLUS`? It has the same model architecture but is trained on more data, the checkpoint is in https://github.com/microsoft/UniSpeech/tree/main/WavLM

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
+        self.out_proj = nn.Linear(embed_dim, embed_dim, bias=bias)
+        self.out_proj.qconfig = None
+
+        self.gru_rel_pos = gru_rel_pos
+        if self.gru_rel_pos:
+            self.gru_rel_pos_linear = nn.Linear(self.head_dim, 8)
+            self.gru_rel_pos_const = nn.Parameter(torch.ones(1, num_heads, 1, 1))
+        self.has_position_bias = True
+
+    def compute_bias(self, query_length: int, key_length: int):
```
nateanl, line 389: although there is no docstring in the original repo, could you add descriptions for the input arguments so that it is easier for users to understand?

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
+                to the next one.
+        Returns:
+            attn_output: attention output of shape `(batch_size, src_len, embed_dim)`
+            position_bias: position bias of shape `(batch_size * num_heads, src_len, src_len)`.
+        """
+        bsz, seq_len, embed_dim = query.size()
+        assert embed_dim == self.embed_dim
+        assert attention_mask is None
+
+        if self.has_relative_attention_bias and position_bias is None:
```
nateanl, line 451: by changing the conditional logic it can pass the torchscript check.
```suggestion
        if self.rel_attn_embed is not None and position_bias is None:
```

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
+        self.max_distance = max_distance
+
+        # WavLM applies position embedding rel_attn_embed in the first encoder layer, but not in the subsequent ones.
+        # In order to keep our model TorchScript-able, this attribute needs to be present in all layers, so we create
+        # a dummy identity module when the actual embedding is not needed.
+        self.rel_attn_embed = torch.nn.Identity()
+        self.rel_attn_embed_device = "cpu"
+        if has_relative_attention_bias:
+            self.rel_attn_embed = nn.Embedding(num_buckets, num_heads)
+            self.rel_attn_embed_device = self.rel_attn_embed.weight.device
```
nateanl, line 368:
```suggestion
        if has_relative_attention_bias:
            self.rel_attn_embed = nn.Embedding(num_buckets, num_heads)
        else:
            self.rel_attn_embed = None
```

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,185 @@ def forward(
...
+            self.gru_rel_pos_linear = nn.Linear(self.head_dim, 8)
+            self.gru_rel_pos_const = nn.Parameter(torch.ones(1, num_heads, 1, 1))
+        self.has_position_bias = True
+
+    def compute_bias(self, query_length: int, key_length: int):
+        context_position = torch.arange(query_length, dtype=torch.long)[:, None]
+        memory_position = torch.arange(key_length, dtype=torch.long)[None, :]
+        relative_position = memory_position - context_position
+        relative_position_bucket = self._relative_positions_bucket(relative_position, bidirectional=True)
+        relative_position_bucket = relative_position_bucket.to(self.rel_attn_embed_device)
```
nateanl, line 394:
```suggestion
        relative_position_bucket = relative_position_bucket.to(self. rel_attn_embed.weight.device)
```

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
```diff
@@ -108,18 +118,52 @@ def _test_import_pretrain(self, original, imported, config):
...
         x = torch.randn(3, 10, config["hidden_size"])
         ref = original.lm_head(x)
         hyp = imported.aux(x)
         self.assertEqual(ref, hyp)
         # The whole model without mask
-        x = torch.randn(3, 1024)
-        ref = original(x).logits
-        hyp, _ = imported(x)
-        self.assertEqual(ref, hyp)
-        # The whole model without mask
```
mthrok, line 122: why is this part removed?

test/torchaudio_unittest/models/wav2vec2/model_test.py
```diff
@@ -33,6 +35,8 @@ def _name_func(testcase_func, i, param):
         (hubert_base,),
         (hubert_large,),
         (hubert_xlarge,),
+        (wavlm_base,),
+        (wavlm_large,),
```
mthrok, line 39: Looks like adding wavlm configs to existing one adds many skip/branch conditions to existing tests. (We should avoid introducing always-skip as much as possible.)

I think it's better to logically split them. Can you create a dedicated test suit for wavlm?

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +318,203 @@ def forward(
         output = output.transpose(2, 1).reshape(batch_size, length, embed_dim)
 
         output = self.out_proj(output)
-        return output
+        return output, None  # Necessary for compatibility with WavLMSelAttention
+
+
+class WavLMSelfAttention(nn.Module):
```
mthrok, line 324: So as to be extra careful of the license, can you move this to a separate file, then add license notice from the original unilm, instead of unispeech? We cannot port code from unispeech repo, even though they ported the code from unilm.

mthrok:
Can we split this PR into one for model architecture and one for bundle?

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
nateanl: I think it is duplicate of the following code.

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
```diff
@@ -213,6 +310,15 @@ def test_recreate_pretrain(self, config, factory_func):
         reloaded.eval()
         self._test_recreate(imported, reloaded, config)
 
+    @WAVLM_CONFIGS
+    def test_recreate_wavlm(self, config, factory_func):
```
nateanl, line 314: can this function be merged with `test_recreate_wavlm` above, as there is no finetune/aux case in wavlm model?

torchaudio/models/wav2vec2/wavlm_attention.py
```diff
@@ -0,0 +1,224 @@
...
+                gate_a_1 = gate_a * (gate_b * self.gru_rel_pos_const - 1.0) + 2.0
+                attn_mask_rel_pos = gate_a_1.view(bsz * self.num_heads, -1, 1) * position_bias
+
+            attn_mask_rel_pos = attn_mask_rel_pos.view((-1, seq_len, seq_len))
+
+        bias_k = bias_v = None
+        add_zero_attn = False
+        # multi_head_attention_forward expects query shape (seq_len, batch_size, embed_dim)
+        query = query.transpose(0, 1)
+        concat_bias = torch.cat((self.q_proj.bias, self.k_proj.bias, self.v_proj.bias))
```
nateanl, line 199: I figure out why `qproj` can't be quantized. After quantization, the Linear module is converted to `DynamicQuantizedLinear`, in which the `bias` parameter becomes a method: https://github.com/pytorch/pytorch/blob/d542aab5c1bc544f9dc0eb5632bfe4432223d890/torch/ao/nn/quantized/modules/linear.py#L236

With that, there seems no way to quantize the linear layers if the bias are used like this. We can leave the code as is now, could you add comment to explain the issue why it can't be quantized?

nateanl (approved):
LGTM overall. Thanks for working on it.
@mthrok could you review it again, also do you know how to address the quantization issue of concatenating bias parameters?

nateanl:
In PyTorch's `MultiheadAttention` module, the weights and biases are defined separately: https://pytorch.org/docs/stable/_modules/torch/nn/modules/activation.html#MultiheadAttention 
which can avoid the quantization issue. shall we follow the same solution here?

test/torchaudio_unittest/models/wav2vec2/huggingface_intergration_test.py
```diff
@@ -159,6 +164,50 @@ def test_import_finetune(self, config, _):
...
+        self.assertEqual(ref, hyp)
+        # Convolutional Positional Encoder
+        x = torch.randn(3, 256, config["hidden_size"])
+        ref = original.encoder.pos_conv_embed(x)
+        hyp = imported.encoder.transformer.pos_conv_embed(x)
+        self.assertEqual(ref, hyp)
+
+        position_bias = None
+        position_bias_imp = None
+        for original_, imported_ in zip(original.encoder.layers, imported.encoder.transformer.layers):
```
mthrok, line 191: Could we check that both `original.encoder.layers` and `imported.encoder.transformer.layers` have non-zero values and they are the same length?

This can be simply achieved with [`strict` option in Python 3.10](https://peps.python.org/pep-0618/), but until we drop the support for 3.9, this needs to be done manually.

mthrok: or you can introduce [`zip_common` like in torchtext](https://github.com/pytorch/text/blob/main/test/torchtext_unittest/common/case_utils.py#L45-L54) (and replace all the `zip` in test as a separate PR (different bootcamp task))

torchaudio/models/wav2vec2/components.py
```diff
@@ -273,15 +275,22 @@ def forward(
...
+    ) -> Tuple[Tensor, Optional[Tensor]]:
         """
         Args:
             x (Tensor): shape: ``[batch_size, sequence_length, embed_dim]``.
-            attention_mask (Tensor or None, optional):
+            attention_mask (Tensor or ``None``, optional):
                 shape: ``[batch_size, 1, sequence_length, sequence_length]``
-
+            position_bias (Tensor or ``None``): Has to be ``None``, necessary for compatibility
+                with :py:class:`WavLMSelfAttention`. (Default: ``None``)
```
mthrok, line 287: Since it is not used, there is no use of adding annotation in docstring.

```suggestion
            position_bias: Not used. Only for the compatibility with :py:class:`WavLMSelfAttention`.
```

torchaudio/models/wav2vec2/components.py
```diff
@@ -273,15 +275,22 @@ def forward(
...
+            key_padding_mask (Tensor or ``None``): Has to be ``None``, necessary for compatibility
+                with :py:class:`WavLMSelfAttention`. (Default: ``None``)
         Returns:
-            Tensor: The resulting tensor. shape: ``[batch, sequence_length, embed_dim]``
+            (Tensor, ``None``): The resulting attention output and ``None`` (necessary for compatibility
+                with :py:class:`WavLMSelAttention`).
+                Attention output shape: ``[batch, sequence_length, embed_dim]``.
```
mthrok, line 293: > necessary for compatibility

I understand why this change is necessary but this feels like code smell... but I do not know if there is a better way either... It's okay as this is not a public API 🧐🧐🧐

torchaudio/models/wav2vec2/components.py
```diff
@@ -314,7 +323,7 @@ def forward(
         output = output.transpose(2, 1).reshape(batch_size, length, embed_dim)
 
         output = self.out_proj(output)
-        return output
+        return output, None  # Necessary for compatibility with WavLMSelAttention
```
mthrok, line 326: code smell ...

torchaudio/models/wav2vec2/wavlm_attention.py
mthrok: I believe rolling out own MHA implementation can workaround this, and that's another motivation of own implementation in `Wav2Vec2Model`.

mthrok (approved):

sgrigory:
> In PyTorch's `MultiheadAttention` module, the weights and biases are defined separately: https://pytorch.org/docs/stable/_modules/torch/nn/modules/activation.html#MultiheadAttention which can avoid the quantization issue. shall we follow the same solution here?

Sure, that works, just pushed the changes. That also required renaming weights and biases when importing from HF, since in the HF model those parameters are nested inside `Linear`

Going forward, would it makes sense to migrate to `torch.nn.MultiheadAttention` instead of calling `F.multi_head_attention_forward` directly? The advantage could be the fast path implemented in native code that it can follow under certain conditions - and in other cases it falls back onto `F.multi_head_attention_forward`

mthrok:
@nateanl Follow-up: the documentation
- https://pytorch.org/audio/main/models.html#factory-functions
- Mention wavelm factory functions in note at https://pytorch.org/audio/main/generated/torchaudio.models.Wav2Vec2Model.html#wav2vec2model

PR #2823, opened 2022-11-02
BarkSpectrogram

jdariasl:
I have added BarkScale transform, which can transform a regular Spectrogram into a BarkSpectrograms similar to MelScale. @ahmed-fau opened this requirement in December 2021 with the number (https://github.com/pytorch/audio/issues/2103). The new functionality includes three different well-known approximations of the Bark scale.


jdariasl:
I'm unsure if the failing tests are related to the functionalities I'm trying to add.

test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -43,7 +43,7 @@ def test_batch_Resample(self):
     def test_batch_MelScale(self):
         specgram = torch.randn(3, 2, 201, 256)
 
-        atol = 1e-6 if os.name == "nt" else 1e-8
+        atol = 1e-4 if os.name == "nt" else 1e-6
```
carolineechen, line 46: can you revert back this change? these values should be fine for the original melscale tests

test/torchaudio_unittest/transforms/batch_consistency_test.py
```diff
@@ -58,6 +58,24 @@ def test_batch_InverseMelScale(self):
         # exactly same result. For this reason, tolerance is very relaxed here.
         self.assert_batch_consistency(transform, mel_spec, atol=1.0, rtol=1e-5)
 
+    def test_batch_BarkScale(self):
+        specgram = torch.randn(3, 2, 201, 256)
+
+        atol = 1e-4 if os.name == "nt" else 1e-6
```
carolineechen, line 64: do you know why these tolerances are increased from the working melscale ones above?

torchaudio/functional/functional.py
```diff
@@ -479,6 +480,127 @@ def _mel_to_hz(mels: Tensor, mel_scale: str = "htk") -> Tensor:
     return freqs
 
 
+def _hz_to_bark(freq: float, bark_scale: str = "traunmuller") -> float:
```
carolineechen, line 483: I think we can have `bark_scale` be required variables in `_hz_to_bark` and `_bark_to_hz`, as these are internal functions called only within the barkscale function. this way, we ensure that we do pass in the correct value every time.

torchaudio/transforms/_transforms.py
```diff
@@ -649,6 +649,298 @@ def forward(self, waveform: Tensor) -> Tensor:
         return mel_specgram
 
 
+class BarkScale(torch.nn.Module):
+    r"""Turn a normal STFT into a bark frequency STFT with triangular filter banks.
+
+    .. devices:: CPU CUDA
+
+    .. properties:: Autograd TorchScript
```
carolineechen, line 657: can you add an autograd test for `BarkScale` as well?

carolineechen (commented):
Hi @jdariasl, thanks for working on this! I left a couple of comments.

I also wanted to note that as this is a new feature, the TorchAudio team would like to first support it as a prototype feature ([feature classifications](https://pytorch.org/audio/main/feature_classifications.html)) available in our nightly builds, and will reevaluate moving it to beta in one of the future releases, pending use cases and API stability (it would be great if you could provide any additional context/existing implementations of this feature we could refer to!). If that sounds good to you, we can help move it to `torchaudio.prototype` after your PR is reviewed and merged, lmk if you have any questions. Thanks again!

test/torchaudio_unittest/transforms/batch_consistency_test.py
jdariasl: Yes of course. When a ran the test locally, MelScale didn't satisfy these values, so I played a little bit with the values and forgot to revert the changes. You can test it on your own.

torchaudio/functional/functional.py
jdariasl: To clarify, the functions `_hz_to_bark` and `_bark_to_hz` are called within the function `barkscale_fbanks` and not within `bark_scale`.

Regarding your request, `_hz_to_bark` and `_bark_to_hz` can be defined using only required variables, but that would not keep the coherence with the functions `_hz_to_mel` and `_mel_to_hz`. Is it recommendable?

jdariasl:
The bark scale is just another auditory scale used in speech and audio processing. Such scale is used to estimate RASTA-PLP coefficients (the acronym stands for Relative Spectral Transform - Perceptual Linear Prediction), which is also an alternative to MFCC used in different contexts, including biomedical applications. Most of the available implementations are developed in Matlab: https://www.ee.columbia.edu/~dpwe/resources/matlab/rastamat/, https://es.mathworks.com/help/audio/ref/designauditoryfilterbank.html, even though there is also an implementation in Praat. In this sense, I think a better integration of the Bark scale into torchaudio would be to develop an `auditory_fbanks` function that subsumes `melscale_fbanks` and `barkscale_fbanks`, but I didn't want to modify existing functions. 

On the other hand, I'm unsure whether it is possible to propose a comparison with Matlab implementations.

test/torchaudio_unittest/transforms/batch_consistency_test.py
jdariasl: Screenshot_20221103_103546

This is the output of batch_consistency_test with the original tolerance values.

jdariasl: Is there any feedback about this?

torchaudio/functional/functional.py
```diff
@@ -479,6 +480,127 @@ def _mel_to_hz(mels: Tensor, mel_scale: str = "htk") -> Tensor:
...
+    r"""Convert Hz to Barks.
+
+    Args:
+        freqs (float): Frequencies in Hz
```
carolineechen, line 487: `freqs` here is inconsistent to the `freq` variable used in the code/function call

torchaudio/functional/functional.py
```diff
@@ -479,6 +480,127 @@ def _mel_to_hz(mels: Tensor, mel_scale: str = "htk") -> Tensor:
...
+        bark_scale (str, optional): Scale to use: ``traunmuller``, ``schroeder`` or ``wang``. (Default: ``traunmuller``)
+    Returns:
```
carolineechen, line 489: nit: new line before "Returns"
```suggestion

    Returns:
```

torchaudio/functional/functional.py
```diff
@@ -479,6 +480,127 @@ def _mel_to_hz(mels: Tensor, mel_scale: str = "htk") -> Tensor:
...
+
+    .. properties:: TorchScript
+
+    Note:
+        For the sake of the numerical compatibility with librosa, not all the coefficients
+        in the resulting filter bank has magnitude of 1.
+
+        .. image:: https://download.pytorch.org/torchaudio/doc-assets/bark_fbanks.png
+           :alt: Visualization of generated filter bank
+
```
carolineechen, line 563: this note can be removed -- it's not supported in librosa, correct? and we also dont have tests or the asset image

test/torchaudio_unittest/transforms/batch_consistency_test.py
carolineechen: Hey, yea it is also erroring out for me with those values locally, but seems to run okay on our GitHub CircleCI tests -- I think we can leave as is for now, and I can also test out to see if bark scale is passing on CircleCI tests with smaller tolerances after this PR is merged.

carolineechen (commented):
Left a couple of small comments, and could you also add the new functions to `docs/source/functional.rst` and `docs/source/transforms.rst` to have it reflected in the pytorch documentation?

Otherwise, the implementation looks good, and should be ready to merge soon! Once merged, I'll create a PR to move this into prototype and investigate the bark spectrogram tolerance, thanks for the contribution :)

torchaudio/functional/functional.py
jdariasl: You are right, librosa does not support the bark scale. On the other hand, the asset image is easy to get. I modified the Feature Extractor tutorial to show them. If you want to take a look at it, the file can be downloaded from this [link](https://drive.google.com/file/d/12CZNTk7nPt7qgubASWoyh9ae8OZxOrDA/view?usp=share_link).

torchaudio/functional/functional.py
jdariasl: Done; I decided to change the variable to make it consistent with the plural meaning and with the definition in `_bark_to_hz`. By the way, I noticed the same inconsistency in `_hz_to_mel`. It could be fixed without much analysis because those functions are called only by `melscale_fbanks`.

torchaudio/functional/functional.py
carolineechen: I do not have access to the link, could you share it with the public or send it to me carolinechen@meta.com?

jdariasl: You should have access now.

carolineechen: cool thanks! I can help add that to our assets and docs in a future PR

carolineechen (approved):
thanks for the feature contribution, I think this implementation looks good!

follow ups on my end after merging
- move it to a prototype feature
- add assets image

torchaudio/functional/functional.py
jdariasl: Great!, thanks.

nateanl (approved):

PR #2827, opened 2022-11-02
Add conformer wav2vec2 pretrain model

carolineechen:
modeled after [paper](https://arxiv.org/pdf/2110.07313.pdf) and internal flow f288347302

internal comparison tests: D40080919


torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -143,6 +244,81 @@ def extract_features(
...
+
+        if lengths is not None:
+            padding_mask = components._get_padding_mask(x, lengths)
+        else:
+            padding_mask = None
+
+        x = self.wav2vec2.encoder.feature_projection.layer_norm(x)
+        x = self.wav2vec2.encoder.feature_projection.dropout(x)
+        x, _ = self.mask_generator(x, padding_mask)
+        x = self.wav2vec2.encoder.feature_projection.projection(x)
```
nateanl, line 310: wondering if the projection layer is the same linear layer between feature extractor and negative sampler?

torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -143,6 +244,81 @@ def extract_features(
         return self._get_intermediate_outputs(x, mask=masks, num_layers=num_layers)
 
 
+class ConformerWav2Vec2PretrainModel(Module):
```
nateanl, line 247: Is it possible to define a generic `Wav2Vec2PretrainModel` so that it is compatible with original transformer-based Wav2Vec2 model. In that case, we need to keep a location for quantization model between  feature extractor and negative sampler. And I think this conformer based model still has a linear layer even though it doesn't use quantization layer.

torchaudio/prototype/models/_conformer_wav2vec2.py
carolineechen: good point, the input/output dimensions should be the same, but the linear layer shouldnt be reused across these two purposes

torchaudio/prototype/models/_conformer_wav2vec2.py
carolineechen: Just to make sure I understand correctly, to make this follow the wav2vec2 pretraining, we would change the init to take in the following args: `wav2vec2`, `mask_generator`, `linear/quanitzer` (variable name suggestion?), `ngative_sampler`? I think that's fine and generic `Wav2Vec2PretrainModel` would follow the same 4 variable usages, but should probably be out of scope of this PR and something we could add separately if we want to support. (and in that case, after adding generic `Wav2Vec2PretrainModel`, it'd be easy to subclass)

torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -2,13 +2,114 @@
...
+    bsz, tsz, fsz = input.shape
+    input = input.view(-1, fsz)
+
+    cross_high = tsz * bsz
+    high = tsz
+
+    assert high > 1
+
+    if num_negatives > 0:
+        tszs = _buffered_arange(tsz).unsqueeze(-1).expand(-1, num_negatives).flatten()
```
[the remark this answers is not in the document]
carolineechen: `_buffered_arange` keeps a buffered tensor that is reused across calls and is supposed to result in performance improvements over `torch.arange` which creates a new tensor  https://stackoverflow.com/questions/68616053/why-is-buffered-arange-is-faster-than-torch-arange-in-pytorch

torchaudio/prototype/models/_conformer_wav2vec2.py
carolineechen: addressed this in the latest commit, lmk what you think of this format (`intermediate` (open to other variable names) can be replaced by `quantizer` variable if adding a generic w2v2 pretrained model). another possibility would be to group the intermediate and negative sampling layers into a class, that can be fetched similarly to feature_extractor or conformer_encoder modules.

nateanl: `preprocessor` sounds good. thanks for addressing it.

torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -2,13 +2,125 @@
...
+                The negative samples
+            Tensor
+                The indices of the negative samples
+            Tensor or None
+                If ``lengths`` argument was provided, a Tensor of shape `(batch, )` representing
+                valid length in time axis is returns.
+        """
+        input = self.preprocessor(input)
+        negs, neg_idxs = _sample_negatives(input, self.num_negatives, self.cross_sample_negatives)
+        return negs, neg_idxs, lengths
```
nateanl, line 121: I think `input` after `preprocessor` is also required in the loss function, could you verify it?

torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -143,6 +255,81 @@ def extract_features(
...
+        x, lengths = self.wav2vec2.feature_extractor(features, audio_lengths)
+
+        if lengths is not None:
+            padding_mask = components._get_padding_mask(x, lengths)
+        else:
+            padding_mask = None
+
+        x = self.wav2vec2.encoder.feature_projection.layer_norm(x)
+        x = self.wav2vec2.encoder.feature_projection.dropout(x)
+        x, _ = self.mask_generator(x, padding_mask)
```
nateanl, line 320: same for `mask_indices`, which may also be used in loss computation

nateanl (commented):
looks good overall. could you address which outputs are used in pre-training? after that we can merge this pr

torchaudio/prototype/models/_conformer_wav2vec2.py
```diff
@@ -318,3 +529,263 @@ def conformer_wav2vec2_base(
...
+    mask_prob: float = 0.3,
+    mask_length: int = 3,
+    num_negatives: int = 100,
+    cross_sample_negatives: int = 0,
+) -> ConformerWav2Vec2PretrainModel:
+    """Build Conformer Wav2Vec2 Model for pre-training with "small" architecture from
+    *Conformer-Based Self-Supervised Learning for Non-Speech Audio Tasks* :cite:`conformerssl`
+
+    Args:
+        extractor_input_dim (int): Input dimension of the features. (Default: 64)
```
nateanl, line 683: could you add `optional` to the following arguments and the one in `conformer_wav2vec2_pretrain_large`?
```suggestion
        extractor_input_dim (int, optional): Input dimension of the features. (Default: 64)
```

nateanl (approved):
LGTM. Could you rebase the pr and address the comments?

nateanl (approved):

mthrok:
@carolineechen Did you forget to update the bibtex? I am seeing the warnings.

```
/torchaudio/prototype/models/_conformer_wav2vec2.py:docstring of torchaudio.prototype.models._conformer_wav2vec2.conformer_wav2vec2_pretrain_base:1: WARNING: could not find bibtex key "conformerssl"
/torchaudio/prototype/models/_conformer_wav2vec2.py:docstring of torchaudio.prototype.models._conformer_wav2vec2.conformer_wav2vec2_pretrain_large:1: WARNING: could not find bibtex key "conformerssl"
```

mthrok:
Also these

```
/torchaudio/prototype/models/_conformer_wav2vec2.py:docstring of torchaudio.prototype.models._conformer_wav2vec2.conformer_wav2vec2_pretrain_base:8: WARNING: Inline literal start-string without end-string.
/torchaudio/prototype/models/_conformer_wav2vec2.py:docstring of torchaudio.prototype.models._conformer_wav2vec2.conformer_wav2vec2_pretrain_large:8: WARNING: Inline literal start-string without end-string.
```

carolineechen:
@mthrok thanks for catching them, addressing them in #2900. are these warnings from CI and should we throw errors for these on moving forward?

PR #2829, opened 2022-11-02
Add speed and speed perturbation functions and transforms

hwangjeff:
Adds functions and transforms for speed and speed perturbation (https://www.isca-speech.org/archive/interspeech_2015/ko15_interspeech.html).

### DIFF
--- a/torchaudio/prototype/functional/functional.py
+++ b/torchaudio/prototype/functional/functional.py
@@ -1,7 +1,10 @@
 import math
 import warnings
 
+from typing import Tuple
+
 import torch
+from torchaudio.functional import resample
 
 from torchaudio.functional.functional import _create_triangular_filterbank
 
@@ -306,3 +309,39 @@ def barkscale_fbanks(
         )
 
     return fb
+
+
+def speed(
+    waveform: torch.Tensor, lengths: torch.Tensor, orig_freq: int, factor: float
+) -> Tuple[torch.Tensor, torch.Tensor]:
+    r"""Adjusts waveform speed.
+
+    .. devices:: CPU CUDA
+
+    .. properties:: Autograd TorchScript
+
+    Args:
+        waveform (torch.Tensor): Input signals, with shape `(..., time)`.
+        lengths (torch.Tensor): Valid lengths of signals in ``waveform``, with shape `(...)`.
+        orig_freq (int): Original frequency of the signals in ``waveform``.
+        factor (float): Factor by which to adjust speed of input. Values greater than 1.0
+            compress ``waveform`` in time, whereas values less than 1.0 stretch ``waveform`` in time.
+
+    Returns:
+        (torch.Tensor, torch.Tensor):
+            torch.Tensor
+                Speed-adjusted waveform, with shape `(..., new_time).`
+            torch.Tensor
+                Valid lengths of signals in speed-adjusted waveform, with shape `(...)`.
+    """
+
+    source_sample_rate = int(factor * orig_freq)
+    target_sample_rate = int(orig_freq)
+
+    gcd = math.gcd(source_sample_rate, target_sample_rate)
+    source_sample_rate = source_sample_rate // gcd
+    target_sample_rate = target_sample_rate // gcd
+
+    return resample(waveform, source_sample_rate, target_sample_rate), torch.ceil(
+        lengths * target_sample_rate / source_sample_rate
+    ).to(lengths.dtype)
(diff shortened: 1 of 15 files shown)


torchaudio/prototype/transforms/_transforms.py
```diff
@@ -384,3 +385,56 @@ def forward(self, waveform: torch.Tensor) -> torch.Tensor:
...
+
+
+def _source_target_sample_rate(orig_freq: int, speed: float) -> Tuple[int, int]:
+    source_sample_rate = int(speed * orig_freq)
+    target_sample_rate = int(orig_freq)
+    gcd = math.gcd(source_sample_rate, target_sample_rate)
+    return source_sample_rate // gcd, target_sample_rate // gcd
+
+
+class TimeStretch(torch.nn.Module):
```
mthrok, line 397: what's your take on the name collision with the existing one?

https://pytorch.org/audio/stable/generated/torchaudio.transforms.TimeStretch.html#torchaudio.transforms.TimeStretch

hwangjeff: ah that's unfortunate. i feel like "time stretch" fits better with the operators introduced here (stretching waveforms in time domain vs stretching spectrograms along time axis). this is also more consistent with librosa (https://librosa.org/doc/main/generated/librosa.effects.time_stretch.html)

hwangjeff: actually, never mind — librosa's is neither: it accepts a waveform as input, generates spectrograms, and does the stretching on the spectrograms before converting back to time domain

hwangjeff: let's go with "speed" — consistent with sox and "speed perturbation"

torchaudio/prototype/functional/functional.py
```diff
@@ -295,3 +298,39 @@ def barkscale_fbanks(
...
+
+    Returns:
+        (torch.Tensor, torch.Tensor):
+            torch.Tensor
+                Speed-adjusted waveform, with shape `(..., new_time).`
+            torch.Tensor
+                Valid lengths of signals in speed-adjusted waveform, with shape `(...)`.
+    """
+
+    source_sample_rate = int(factor * orig_freq)
```
xiaohui-zhang, line 327: frequency and sample rate are the same thing, but maybe use terms more consistently. sample_rate -> freq as used in the implementation of "resample".  BTW are we flipping source and target here? I thought target_freq (or new_freq) should be int(factor * orig_freq) instead. Maybe I missed sth.

torchaudio/prototype/transforms/_transforms.py
```diff
@@ -384,3 +385,100 @@ def forward(self, waveform: torch.Tensor) -> torch.Tensor:
         specgram = self.spectrogram(waveform)
         bark_specgram = self.bark_scale(specgram)
         return bark_specgram
+
+
+def _source_target_sample_rate(orig_freq: int, speed: float) -> Tuple[int, int]:
+    source_sample_rate = int(speed * orig_freq)
```
xiaohui-zhang, line 391: same here

torchaudio/prototype/transforms/_transforms.py
```diff
@@ -384,3 +385,100 @@ def forward(self, waveform: torch.Tensor) -> torch.Tensor:
...
+
+        Returns:
+            (torch.Tensor, torch.Tensor):
+                torch.Tensor
+                    Speed-adjusted waveform, with shape `(..., new_time).`
+                torch.Tensor
+                    Valid lengths of signals in speed-adjusted waveform, with shape `(...)`.
+        """
+
+        if torch.bernoulli(self.perturb_prob):
```
xiaohui-zhang, line 475: it looks like all waveforms in a batch will be either all perturbed together or all kept original? can we apply the bernoulli factor on each waveform instead? that'll add more randomness in the data especially when batch size is large.

torchaudio/prototype/functional/functional.py
xiaohui-zhang: ok thanks for the explanation. you're right

torchaudio/prototype/transforms/_transforms.py
xiaohui-zhang: ok I agree. I don't it worth the performance degradation. let's keep it the current way. BTW I think it can also be achieved by inserting "1"s in the "factors" sequence, e.g. if we want perturb 40% data, then we just make sure 60% of the elements in "factors" are "1"s and use perturb_prob=1.

xiaohui-zhang (approved):

torchaudio/prototype/transforms/_transforms.py
hwangjeff: that's a good idea. it wasn't the api that was originally proposed, but this seems cleaner and clearer

Issue #2830, opened 2022-11-04
StreamWriter does not handle decimal point frame/sample rate properly


mthrok:
`torchaudio.io.StreamWriter` rounds decimal point frame rate.
The following code inputs FPS: 5000 / 167 (29.94), but produces 29.

```python
from io import BytesIO

import torch
from torchaudio.io import StreamReader, StreamWriter

width, height = 128, 96
frame_rate = 5000 / 167
print("Input frame rate:", frame_rate)

buffer = BytesIO()
writer = StreamWriter(buffer, format="mp4")
writer.add_video_stream(frame_rate=frame_rate, width=width, height=height)

video = torch.randint(256, (90, 3, height, width), dtype=torch.uint8)
with writer.open():
    writer.write_video_chunk(0, video)

buffer.seek(0)

reader = StreamReader(buffer, format="mp4")
print("Output frame rate:", reader.get_src_stream_info(0).frame_rate)
```

```
Input frame rate: 29.940119760479043
Output frame rate: 29.0
```

For audio it does not even handle decimal point.
Note: [`AVCodecContext.sample_rate`](https://ffmpeg.org/doxygen/4.1/structAVCodecContext.html#a8ff0b000c463361e234af48d03aadfc0) is integer, so it's not supported by FFmpeg??

```
Traceback (most recent call last):
  File "/Users/moto/Development/torchaudio/repro_fps.py", line 14, in <module>
    writer.add_audio_stream(sample_rate=sample_rate, num_channels=1)
  File "/Users/moto/Development/torchaudio/torchaudio/io/_stream_writer.py", line 150, in add_audio_stream
    self._s.add_audio_stream(sample_rate, num_channels, format, encoder, encoder_option, encoder_format)
TypeError: add_audio_stream(): incompatible function arguments. The following argument types are supported:
    1. (self: torchaudio._torchaudio_ffmpeg.StreamWriterFileObj, arg0: int, arg1: int, arg2: str, arg3: Optional[str], arg4: Optional[Dict[str, str]], arg5: Optional[str]) -> None

Invoked with: <torchaudio._torchaudio_ffmpeg.StreamWriterFileObj object at 0x101369cf0>, 5512.5, 1, 'flt', None, None, None
```

PR #2831, opened 2022-11-04
Fix decimal FPS handling StreamWriter

mthrok:
StreamWriter assumed that frame rate is always expressed as 1/something, which is a reasonable assumption.

This commit fixes it by properly computing time_base from frame rate.

Address #2830

### DIFF
--- a/test/torchaudio_unittest/io/stream_writer_test.py
+++ b/test/torchaudio_unittest/io/stream_writer_test.py
@@ -351,4 +351,4 @@ class StreamWriterInterfaceTest(_MediaSourceMixin, TempDirMixin, TorchaudioTestC
 
         # Load data
         reader = torchaudio.io.StreamReader(src=self.get_temp_path(filename))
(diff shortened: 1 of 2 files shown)


test/torchaudio_unittest/io/stream_writer_test.py
```diff
@@ -327,3 +327,28 @@ def test_audio_num_frames(self, ext, sample_rate, num_channels):
...
+
+        video = torch.randint(256, (90, 3, height, width), dtype=torch.uint8)
+        with writer.open():
+            writer.write_video_chunk(0, video)
+        if self.test_fileobj:
+            dst.flush()
+
+        # Load data
+        reader = torchaudio.io.StreamReader(src=self.get_temp_path(filename))
+        assert frame_rate == reader.get_src_stream_info(0).frame_rate
```
carolineechen, line 354: nit: assert actual == expected
```suggestion
        assert reader.get_src_stream_info(0).frame_rate == frame_rate
```

carolineechen (approved):

PR #2833, opened 2022-11-04
Add WavLM bundles

sgrigory:
Closes T136364380, follow-up to #2822 

- Added "base", "base+", and "large" bundles for WavLM
- Expanded `wav2vec2_pipeline_test.py` to include the new bundles
- Added the new bundles to docs in `pipelines.rst`

### DIFF
--- a/docs/source/pipelines.rst
+++ b/docs/source/pipelines.rst
@@ -55,7 +55,7 @@ Pretrained Models
    EMFORMER_RNNT_BASE_LIBRISPEECH
 
 
-wav2vec 2.0 / HuBERT - SSL
+wav2vec 2.0 / HuBERT / WavLM - SSL
 --------------------------
 
 Interface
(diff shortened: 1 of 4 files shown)


nateanl (approved):
LGTM!

docs/source/pipelines.rst
```diff
@@ -55,7 +55,7 @@ Pretrained Models
    EMFORMER_RNNT_BASE_LIBRISPEECH
 
 
-wav2vec 2.0 / HuBERT - SSL
+wav2vec 2.0 / HuBERT / WavLM - SSL
 --------------------------
```
mthrok, line 59: Need to update the undreline so that it matches the charlen of the title.

Screen Shot 2022-11-28 at 5 22 16 PM

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -69,7 +69,11 @@ def get_model(self, *, dl_kwargs=None) -> Wav2Vec2Model:
         Args:
             dl_kwargs (dictionary of keyword arguments): Passed to :func:`torch.hub.load_state_dict_from_url`.
         """
-        model = wav2vec2_model(**self._params)
+        model_type = self._params.pop("model_type", None)
```
mthrok, line 72: @sgrigory @nateanl Since model_type is not model configuration, can we move it outside of `_params`, and define it directly on Bundle class?

sgrigory:
@mthrok Thanks for the comments, opened a follow-up PR https://github.com/pytorch/audio/pull/2895

PR #2839, opened 2022-11-10
[Nova] Add M1 Wheels Build

osalpekar:
Adding Nova Reusable Workflow for M1 Wheels Build. Once this has been running well for a while, we can replace the old `build-m1-binaries.yml` workflow.

### DIFF
--- a/.github/workflows/build-wheels-m1.yml
+++ b/.github/workflows/build-wheels-m1.yml
@@ -30,7 +30,7 @@ jobs:
     uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@main
     with:
       repository: ${{ matrix.repository }}
-      ref: nightly
+      ref: ""
       test-infra-repository: pytorch/test-infra
       test-infra-ref: main
       build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-wheels-m1.yml
```diff
@@ -0,0 +1,47 @@
...
+          - repository: pytorch/audio
+            pre-script: packaging/pre_build_script.sh
+            post-script: packaging/post_build_script.sh
+            smoke-test-script: test/smoke_test/smoke_test.py
+            package-name: torchaudio
+    name: ${{ matrix.repository }}
+    uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@main
+    with:
+      repository: ${{ matrix.repository }}
+      ref: nightly
```
DanilBaibak, line 33:
```suggestion
      ref: ""
```

DanilBaibak (approved):
There is a tiny suggestion. Otherwise LGTM 👍

.github/workflows/build-wheels-m1.yml
osalpekar: Great catch - thanks!

PR #2840, opened 2022-11-10
[Nova] Add M1 Conda Builds

osalpekar:
Adding Nova Reusable Workflow for M1 Conda Build. Once this has been running well for a while, we can replace the old `build-m1-binaries.yml` workflow.


DanilBaibak (approved):
LGTM 👍

PR #2842, opened 2022-11-10
[WIP] Use BetterTransfomer in WavLM Self-Attention

sgrigory:
Closes T137506059

Replaces functional multi-head attention in `WavLMSelfAttention` with a module `torch.nn.MultiheadAttention`. The reason is that the latter uses native CPU/CUDA implementation (BetterTransfomer) under certain conditions, and can achieve significant speedup. It also simplifies the code in `WavLMSelfAttention`


nateanl (approved):
LGTM

PR #2860, opened 2022-11-17
Add HiFi GAN Generator to prototypes

sgrigory:
Part 1 of [T138011314](https://www.internalfb.com/intern/tasks/?t=138011314)

This PR ports the generator part of [HiFi GAN](https://arxiv.org/abs/2010.05646v2) from [the original implementation](https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/models.py#L75)

Adds tests:
- Smoke tests for architectures V1, V2, V3
- Check that output shapes are correct
- Check that the model is torchscriptable and scripting doesn't change the output
- Check that our code's output matches the original implementation. Here I clone the original repo inside `/tmp` and import necessary objects from inside the test function.  On test teardown I restore `PATH`, but don't remove the cloned code, so that it can be reused on subsequent runs - let me know if removing it would be a better practice

There are no quantization tests, because the model consists mainly of `Conv1d` and `ConvTransposed1d`, and they are [not supported by dynamic quantization](https://pytorch.org/docs/stable/quantization.html)


torchaudio/prototype/models/hifi_gan.py
```diff
@@ -0,0 +1,307 @@
...
+            x = xt + x
+        return x
+
+
+def get_padding(kernel_size, dilation=1):
+    """Find padding for which 1D convolution preserves the input shape."""
+    return int((kernel_size * dilation - dilation) / 2)
+
+
+def hifigan_model(
```
nateanl, line 220: Considering we may support training HiFiGAN in the future, we can call the factory function `hifigan_generator` in this PR. And we can add `hifigan_discriminator` and `hifigan_model` (which comprises `hifigan_generator` and `hifigan_discriminator`).
```suggestion
def hifigan_generator(
```

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -0,0 +1,307 @@
...
+        resblock_dilation_sizes=resblock_dilation_sizes,
+        resblock_type=resblock_type,
+        upsample_initial_channel=upsample_initial_channel,
+        upsample_kernel_sizes=upsample_kernel_sizes,
+        in_channels=in_channels,
+        lrelu_slope=lrelu_slope,
+    )
+
+
+def hifigan_v1() -> HiFiGANGenerator:
```
nateanl, line 256: same here, maybe call it `hifigan_generator_v1`?

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -0,0 +1,307 @@
...
+        To build the model, please use one of the factory functions: :py:func:`hifigan_model`, :py:func:`hifigan_v1`,
+        :py:func:`hifigan_v2`, :py:func:`hifigan_v3`.
+    Args:
+        in_channels: (int): Number of channels in the input features.
+        upsample_rates (tuple of int): Factors by which each upsampling layer increases the time dimension.
+        upsample_initial_channel (int): Number of channels in the input feature tensor.
+        upsample_kernel_sizes (tuple of int): Kernel size for each upsampling layer.
+        resblock_kernel_sizes (tuple of int): Kernel size for each residual block.
+        resblock_dilation_sizes (tuple of tuples of int): Dilation sizes for each 1D convolutional layer in each
+        residual block. For resblock type 1 inner tuples should have length ``3``, because there are ``3``
```
nateanl, line 22:
```suggestion
            residual block. For resblock type 1 inner tuples should have length ``3``, because there are ``3``
```

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -0,0 +1,307 @@
...
+
+        x = F.leaky_relu(x)
+        x = self.conv_post(x)
+        x = torch.tanh(x)
+
+        return x
+
+
+@torch.jit.interface
+class ResBlockInterface(torch.nn.Module):
```
nateanl, line 92: this is necessary to support torchscript, right?

sgrigory: Yes, since variable `res_block`  in `HiFiGANGenerator.forward` is obtained by indexing `nn.ModuleList`, TorchScript requires it to have a type annotation. But it only recognizes a custom class as a type if I define it as an interface. `ResBlockInterface` has a docstring with a short explanation, I can expand it if needed

torchaudio/prototype/models/hifi_gan.py
sgrigory: Done

torchaudio/prototype/models/hifi_gan.py
sgrigory: Done

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -0,0 +1,307 @@
...
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+from torch.nn import Conv1d, ConvTranspose1d
+
+
+class HiFiGANGenerator(torch.nn.Module):
+    """Generator part of *HiFi GAN* :cite:`NEURIPS2020_c5d73680`.
+    From https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/models.py#L75
```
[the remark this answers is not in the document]
sgrigory: Added the license

nateanl (approved):
LGTM. I notice the doc is missing for hifigan model and factory functions. Could you add it to the documents? 
Also, is test/torchaudio_unittest/prototype/hifi_gan/__init__.py necessary, can you run the tests with removing the file?
Thanks.

sgrigory:
> LGTM. I notice the doc is missing for hifigan model and factory functions. Could you add it to the documents? Also, is test/torchaudio_unittest/prototype/hifi_gan/**init**.py necessary, can you run the tests with removing the file? Thanks.

Thanks for catching - added the doc. 
Regarding `test/torchaudio_unittest/prototype/hifi_gan/__init__.py`, it seems necessary for importing from `torchaudio_unittest.common_utils`: if I remove it, I get
 ```
ModuleNotFoundError: No module named 'torchaudio_unittest'
```
 error when running `pytest test/torchaudio_unittest/prototype/hifi_gan/`. If I switch to relative import (`from ...common_utils import PytorchTestCase, skipIfNoCuda`), I get a different error
```
ImportError: attempted relative import with no known parent package
```
when `__init__.py` is absent

PR #2873, opened 2022-11-24
Add layer normalization to wav2vec2 large+ pretrained models

andreasfloros:
The original fairseq implementation had an extra layer normalization 
preprocessings for large/xlarge models.

https://github.com/facebookresearch/fairseq/blob/fcca32258c8e8bcc9f9890bf4714fa2f96b6b3e1/fairseq/data/audio/hubert_dataset.py#L355-L357

This commit modifies the pre-trained model bundle to include this 
preprocessing to the impacted pre-trained models listed bellow.
For the sake of keeping the interface identical to the other models, 
since the additional preprocessing is rather simple, the returned 
pre-trained model instance is modified to include the preprocess, 
instead of adding a method for preprocessing.

- WAV2VEC2_LARGE_LV60K
- WAV2VEC2_ASR_LARGE_LV60K_10M
- WAV2VEC2_ASR_LARGE_LV60K_100H
- WAV2VEC2_ASR_LARGE_LV60K_960H
- WAV2VEC2_XLSR53
- HUBERT_LARGE
- HUBERT_XLARGE
- HUBERT_ASR_LARGE
- HUBERT_ASR_XLARGE
- WAVLM_LARGE

### DIFF
--- a/torchaudio/pipelines/_wav2vec2/impl.py
+++ b/torchaudio/pipelines/_wav2vec2/impl.py
@@ -1,7 +1,9 @@
 from dataclasses import dataclass
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, List, Optional, Tuple
 
 import torch
+from torch import Tensor
+from torch.nn import functional as F, Module
 from torchaudio._internal import load_state_dict_from_url
 from torchaudio.models import wav2vec2_model, Wav2Vec2Model, wavlm_model
 
@@ -9,6 +11,31 @@ from . import utils
 
 
 __all__ = []
+
+
+class _Wav2Vec2Model(Module):
+    """Wrapper class for :py:class:`~torchaudio.models.Wav2Vec2Model`.
+
+    This is used for layer normalization at the input
+    """
+
+    def __init__(self, model: Wav2Vec2Model):
+        super().__init__()
+        self.model = model
+
+    def forward(self, waveforms: Tensor, lengths: Optional[Tensor] = None) -> Tuple[Tensor, Optional[Tensor]]:
+        waveforms = F.layer_norm(waveforms, waveforms.shape)
+        return self.model(waveforms, lengths)
+
+    @torch.jit.export
+    def extract_features(
+        self,
+        waveforms: Tensor,
+        lengths: Optional[Tensor] = None,
+        num_layers: Optional[int] = None,
+    ) -> Tuple[List[Tensor], Optional[Tensor]]:
+        waveforms = F.layer_norm(waveforms, waveforms.shape)
+        return self.model.extract_features(waveforms, lengths, num_layers)
 
 
 @dataclass
@@ -45,6 +72,7 @@ class Wav2Vec2Bundle:
     _path: str
     _params: Dict[str, Any]
     _sample_rate: float
+    _normalize_waveform: bool
 
     @property
     def sample_rate(self) -> float:
@@ -60,7 +88,7 @@ class Wav2Vec2Bundle:
         state_dict = load_state_dict_from_url(url, **dl_kwargs)
         return state_dict
 
-    def get_model(self, *, dl_kwargs=None) -> Wav2Vec2Model:
+    def get_model(self, *, dl_kwargs=None) -> Module:
         """Construct the model and load the pretrained weight.
 
         The weight file is downloaded from the internet and cached with
@@ -68,6 +96,24 @@ class Wav2Vec2Bundle:
 
         Args:
             dl_kwargs (dictionary of keyword arguments): Passed to :func:`torch.hub.load_state_dict_from_url`.
+
+        Returns:
+            Variation of :py:class:`~torchaudio.models.Wav2Vec2Model`.
+
+            For the models listed below, an additional layer normalization is performed on the input.
+
+            For all other models, a :py:class:`~torchaudio.models.Wav2Vec2Model` instance is returned.
+
+            - WAV2VEC2_LARGE_LV60K
+            - WAV2VEC2_ASR_LARGE_LV60K_10M
+            - WAV2VEC2_ASR_LARGE_LV60K_100H
+            - WAV2VEC2_ASR_LARGE_LV60K_960H
+            - WAV2VEC2_XLSR53
+            - HUBERT_LARGE
+            - HUBERT_XLARGE
+            - HUBERT_ASR_LARGE
+            - HUBERT_ASR_XLARGE
+            - WAVLM_LARGE
         """
         model_type = self._params.pop("model_type", None)
         if model_type == "WavLM":
@@ -75,6 +121,8 @@ class Wav2Vec2Bundle:
         else:
             model = wav2vec2_model(**self._params)
         model.load_state_dict(self._get_state_dict(dl_kwargs))
+        if self._normalize_waveform:
+            model = _Wav2Vec2Model(model)
         model.eval()
         return model
 
@@ -196,6 +244,7 @@ WAV2VEC2_BASE = Wav2Vec2Bundle(
         "aux_num_out": None,
     },
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 WAV2VEC2_BASE.__doc__ = """Wav2vec 2.0 model ("base" architecture),
 pre-trained on 960 hours of unlabeled audio from *LibriSpeech* dataset :cite:`7178964`
@@ -239,6 +288,7 @@ WAV2VEC2_ASR_BASE_10M = Wav2Vec2ASRBundle(
     },
     _labels=utils._get_en_labels(),
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 WAV2VEC2_ASR_BASE_10M.__doc__ = """Wav2vec 2.0 model ("base" architecture with an extra linear module),
 pre-trained on 960 hours of unlabeled audio from *LibriSpeech* dataset :cite:`7178964`
@@ -284,6 +334,7 @@ WAV2VEC2_ASR_BASE_100H = Wav2Vec2ASRBundle(
     },
     _labels=utils._get_en_labels(),
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 
 WAV2VEC2_ASR_BASE_100H.__doc__ = """Wav2vec 2.0 model ("base" architecture with an extra linear module),
@@ -329,6 +380,7 @@ WAV2VEC2_ASR_BASE_960H = Wav2Vec2ASRBundle(
     },
     _labels=utils._get_en_labels(),
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 WAV2VEC2_ASR_BASE_960H.__doc__ = """Wav2vec 2.0 model ("base" architecture with an extra linear module),
 pre-trained on 960 hours of unlabeled audio from *LibriSpeech* dataset :cite:`7178964`
@@ -372,6 +424,7 @@ WAV2VEC2_LARGE = Wav2Vec2Bundle(
         "aux_num_out": None,
     },
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 WAV2VEC2_LARGE.__doc__ = """Wav2vec 2.0 model ("large" architecture),
 pre-trained on 960 hours of unlabeled audio from *LibriSpeech* dataset :cite:`7178964`
@@ -415,6 +468,7 @@ WAV2VEC2_ASR_LARGE_10M = Wav2Vec2ASRBundle(
     },
     _labels=utils._get_en_labels(),
     _sample_rate=16000,
+    _normalize_waveform=False,
 )
 WAV2VEC2_ASR_LARGE_10M.__doc__ = """Wav2vec 2.0 model ("large" architecture with an extra linear module),
 pre-trained on 960 hours of unlabeled audio from *LibriSpeech* dataset :cite:`7178964`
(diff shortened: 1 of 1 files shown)


mthrok (changes requested):
This does not look like aligned to the design of pipeline modules. 

The whole point of pipeline is that users do not need to manually construct preprocessing pipeline.

So I suggest to update the Bundle `get_model` function to include this process

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -11,6 +13,22 @@
...
+    """Wrapper class for Wav2Vec2Model
+
+    This class allows for normalization at the input (see _params["normalize_waveform"])
+    """
+
+    def __init__(self, model: Wav2Vec2Model, normalize_waveform: Optional[bool]):
+        self.model = model
+        self.normalize_waveform = normalize_waveform
+
+    def forward(self, waveforms: Tensor) -> Tuple[Tensor, Optional[Tensor]]:
```
nateanl, line 26: lengths is also included in the Wav2Vec2Model's input arguments. could you add it?

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -11,6 +13,22 @@
...
+        if self.normalize_waveform:
+            waveforms = F.layer_norm(waveforms, waveforms.shape)
+        return self.model(waveforms)
+
```
nateanl, line 30: could you also add an `extract_features` method that applies normalization first and call `self.model.extract_features(...)`?

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -60,7 +78,7 @@ def _get_state_dict(self, dl_kwargs):
         state_dict = load_state_dict_from_url(url, **dl_kwargs)
         return state_dict
 
-    def get_model(self, *, dl_kwargs=None) -> Wav2Vec2Model:
+    def get_model(self, *, dl_kwargs=None) -> _Wav2Vec2Model:
```
mthrok, line 81: Can we make it to generic `torch.nn.Module`, then mention that this method returns a variation of `Wav2Vec2Model` in the docstring so that users browsing the documentation page can go to the model definition easily?

To link to the model definition with Sphinx syntax, please use ```:py:class:`~torchaudio.models.Wav2Vec2` ```.

---
Follow-up: We should also do the same for SourceSeparationBundle, which returns different models https://pytorch.org/audio/stable/generated/torchaudio.pipelines.SourceSeparationBundle.html#get-model

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -70,11 +88,13 @@ def get_model(self, *, dl_kwargs=None) -> Wav2Vec2Model:
             dl_kwargs (dictionary of keyword arguments): Passed to :func:`torch.hub.load_state_dict_from_url`.
         """
         model_type = self._params.pop("model_type", None)
+        normalize_waveform = self._params.pop("normalize_waveform", None)
```
mthrok, line 91: Since this operation is performed on the module-level instance, let's use non-destructive accessor like `get`.

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -70,11 +88,13 @@ def get_model(self, *, dl_kwargs=None) -> Wav2Vec2Model:
...
         if model_type == "WavLM":
             model = wavlm_model(**self._params)
         else:
             model = wav2vec2_model(**self._params)
         model.load_state_dict(self._get_state_dict(dl_kwargs))
+        model = _Wav2Vec2Model(model, normalize_waveform)
```
mthrok, line 97: Let's keep the Wav2Vec2Model intact for cases that does not require the normalization.

```
if self._params.get("normalize_waveform", False):
    model = _Wav2Vec2Model(model)
```

and make `_Wav2Vec2Model` always perform `layer_norm` for now.

torchaudio/pipelines/_wav2vec2/impl.py
andreasfloros: `return self.model(waveforms)` will make a call to `extract_features`. So I don't think we need to rewrite that logic, unless I'm misunderstanding something?

mthrok: we need to expose the same set of public methods from the Wav2Vec2Model class.

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -60,21 +77,29 @@ def _get_state_dict(self, dl_kwargs):
...
         The weight file is downloaded from the internet and cached with
         :func:`torch.hub.load_state_dict_from_url`
 
         Args:
             dl_kwargs (dictionary of keyword arguments): Passed to :func:`torch.hub.load_state_dict_from_url`.
+
+        Returns:
+            Variation of :py:class:`~torchaudio.models.Wav2Vec2`.
+            If self.params_["normalize_waveform"] is True, Wav2Vec2 is wrapped with _Wav2Vec2 to apply
+            an additional layer normalization at the input
```
mthrok, line 92:
```suggestion
```

`self.params_` is implementation detail so it should not mentioned in public-facing documentation. Rather, can you describe in plain English for what models the extra normalization is applied?

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -347,6 +372,7 @@ def _get_state_dict(self, dl_kwargs):
     "wav2vec2_fairseq_large_ls960.pth",
     {
         "extractor_mode": "group_norm",
+        "normalize_waveform": True,
```
mthrok, line 375: Could you move this parameter outside of `_params` and define a separate property directly on `Wav2Vec2Bundle`?

mthrok: I just realized that the current suggestion is following what was done for "model_type" parameter. Sorry that one should not be done like that. It should be outside of `_params` attribute @nateanl https://github.com/pytorch/audio/pull/2833/files#r1034111931

nateanl: sure, will make a fix for `model_type` in a separate PR.

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -60,21 +91,48 @@ def _get_state_dict(self, dl_kwargs):
...
+            WAV2VEC2_LARGE_LV60K
+            WAV2VEC2_ASR_LARGE_LV60K_10M
+            WAV2VEC2_ASR_LARGE_LV60K_100H
+            WAV2VEC2_ASR_LARGE_LV60K_960H
+            WAV2VEC2_XLSR53
+            HUBERT_LARGE
+            HUBERT_XLARGE
+            HUBERT_ASR_LARGE
+            HUBERT_ASR_XLARGE
+            WAVLM_LARGE
```
nateanl, line 120: The documents can be improved, could you add blank lines between the bundles?
image

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -60,21 +91,48 @@ def _get_state_dict(self, dl_kwargs):
...
+
+            the Wav2Vec2Model class is wrapped with _Wav2Vec2Model to apply
```
nateanl, line 122: Wav2Vec2Model can be updated to :py:class:`~torchaudio.models.Wav2Vec2Model`

nateanl (commented):
Some integration tests failed I think it's releated with the code change, could you look into it?
https://github.com/pytorch/audio/actions/runs/3569462051/jobs/6001202652
Otherwise LGTM.

torchaudio/pipelines/_wav2vec2/impl.py
mthrok: > The documents can be improved, could you add blank lines between the bundles?

I suggest
 - use the list syntax
 - move the list to the end of the docstring for improved readability.
 - Not mention internal structure `_Wav2Vec2Model`.

something like the following

```
For the following models, an additional layer normalization is performed on the input.

- WAV2VEC2_LARGE
- WAV2VEC2_ASR_LARGE_10M
...
```

mthrok:
We need to perform the evaluation of this change. Can we run the evaluation script https://github.com/pytorch/audio/tree/main/examples/asr/librispeech_ctc_decoder on some of the model being updated?

nateanl:
I'll run the CTC decoder test locally today.

torchaudio/pipelines/_wav2vec2/impl.py
```diff
@@ -11,6 +13,34 @@
...
+    This is used for layer normalization at the input
+    """
+
+    def __init__(self, model: Wav2Vec2Model):
+        super().__init__()
+        self.model = model
+
+    def forward(self, waveforms: Tensor, lengths: Optional[Tensor] = None) -> Tuple[Tensor, Optional[Tensor]]:
+        waveforms = F.layer_norm(waveforms, waveforms.shape)
+        return self.model(waveforms, lengths)
```
nateanl, line 28: okay, `self.model` here + `__getattr__` method will cause infinite recursion. can we rename the model argument to `wav2vec2model`? @mthrok @andreasfloros

nateanl: hmm, it doesn't matter what name it is, the infinite recursion is always there..

andreasfloros: `__getattr__` is only called if the attribute is not found (https://python-reference.readthedocs.io/en/latest/docs/dunderattr/getattr.html). Since `self.model` is defined in `__init__` I'm not sure why we have recursion

mthrok: I think @nateanl is talking about `__getattr__` in evaluation script. We can discuss that separately.

https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_ctc_decoder/inference.py#L15

nateanl: actually I'm talking about `__getattr__` in this pr. The integration tests failed also because of it https://github.com/pytorch/audio/actions/runs/3569462051/jobs/6001202652

mthrok: oh I see. Why do we need `getattr`? on the wrapper?

mthrok: It seems to due to some magic behind the `nn.Module`'s `__call__` implementation. We do not need generic `getattr` method on the wrapper, don't we?

andreasfloros: @mthrok to expose the same set of public methods as you mentioned https://github.com/pytorch/audio/pull/2873#discussion_r1034007855. Should I remove it and keep only `extract_features`?

mthrok: I see. Thanks for the explanation.

There are only two public methods `forward` and `extract_features`, so with these methods implemented manually, we do not need generic `getattr`.

https://pytorch.org/audio/main/generated/torchaudio.models.Wav2Vec2Model.html#torchaudio.models.Wav2Vec2Model

mthrok (approved):
LGTM. @nateanl Could you double-check that the new parameter `_normalize_waveform` is enabled for the correct set of models?

mthrok:
macOS wav2vec2 large model test failed, but I don't think this is related to the PR...

https://app.circleci.com/pipelines/github/pytorch/audio/13797/workflows/f0462e72-1bc0-49fd-b8ad-4f39b96d74f3/jobs/1039837/tests

nateanl (approved):
Thanks @andreasfloros for working on it. LGTM. 
After checking the paper and the evaluation results. I found the following bundles need `_normalize_waveform` to be False:
- `WAV2VEC2_LARGE`
- `WAV2VEC2_ASR_LARGE_10M`
- `WAV2VEC2_ASR_LARGE_100H`
- `WAV2VEC2_ASR_LARGE_960H`

The reason of using layer_norm on waveforms is make the training stable when using larger datasets like LibriLight or LibriVox. It's not related with the model architecture, instead it's related with the size of training data.

The argument is also consistent with `extractor_mode`, `F.layer_norm` is only used when `extractor_mode` is set to "layer_norm".

@andreasfloros could you make changes to the four bundles mentioned above? Then we can merge this PR. Thanks!

nateanl:
WER results on test-other subset of LibriSpeech via the evaluation script in https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_ctc_decoder/inference.py

Results show that `F.layer_norm` hurts the performance of `WAV2VEC2_ASR_LARGE_960H` a bit, which means it didn't apply normalization in the training. In `WAV2VEC2_ASR_LARGE_LV60K_960H` case, applying `F.layer_norm` can significantly improve the ASR performance.
| Bundle      | Normalization | WER |
| ----------------------------------- | ---------------- | -------------
| WAV2VEC2_ASR_LARGE_960H      | F.layer_norm       | 0.0501
| WAV2VEC2_ASR_LARGE_960H   | None        | 0.0498
| WAV2VEC2_ASR_LARGE_LV60K_960H | F.layer_norm | 0.0368
| WAV2VEC2_ASR_LARGE_LV60K_960H| None | 0.0495

mthrok:
> WER results on test-other subset of LibriSpeech via the evaluation script in https://github.com/pytorch/audio/blob/main/examples/asr/librispeech_ctc_decoder/inference.py
> 
> Results show that `F.layer_norm` hurts the performance of `WAV2VEC2_ASR_LARGE_960H` a bit, which means it didn't apply normalization in the training. In `WAV2VEC2_ASR_LARGE_LV60K_960H` case, applying `F.layer_norm` can significantly improve the ASR performance.
> 
> Bundle	Normalization	WER
> WAV2VEC2_ASR_LARGE_960H	F.layer_norm	0.0501
> WAV2VEC2_ASR_LARGE_960H	None	0.0498
> WAV2VEC2_ASR_LARGE_LV60K_960H	F.layer_norm	0.0368
> WAV2VEC2_ASR_LARGE_LV60K_960H	None	0.0495

Thanks @nateanl. Can you add the new numbers to the README of the script??

nateanl:
cc @Sanyuan-Chen @MarkWuNLP This PR addresses the waveform normalization issue of WavLM Large model in original repo.

PR #2874, opened 2022-11-26
Extend fftconvolve to support broadcast-able shapes

mthrok:
Currently, fftconvolve only accepts the tensors for the exact same leading dimensions.
This commit loosens the restriction to allow shapes that are broadcast-able.

This makes the fftconvolve operation more efficient for cases like signal filtering where one operand (waveform) is larger than the other (filter kernel) and the same filter kernels are applied across channels and batches.


test/torchaudio_unittest/prototype/functional/torchscript_consistency_test_impl.py
```diff
@@ -26,7 +26,7 @@ def _assert_consistency(self, func, inputs, shape_only=False):
         self.assertEqual(ts_output, output)
 
     @nested_params(
-        [F.convolve, F.fftconvolve],
+        ["convolve", "fftconvolve"],
```
mthrok, line 29: Changing this so that the memory address of the function is not leaked into the generated test names.

torchaudio/prototype/functional/functional.py
```diff
@@ -48,7 +57,7 @@ def fftconvolve(x: torch.Tensor, y: torch.Tensor, mode: str = "full") -> torch.T
     Args:
         x (torch.Tensor): First convolution operand, with shape `(..., N)`.
         y (torch.Tensor): Second convolution operand, with shape `(..., M)`
-            (leading dimensions must match those of ``x``).
+            (leading dimensions must be broadcast-able to those of ``x``).
```
hwangjeff, line 60: should we follow the broadcasting rules more closely, e.g. accept inputs of different lengths and check each dimension starting from second to last trailing dimension?

hwangjeff (commented):
would be nice to do this for `convolve` as well

torchaudio/prototype/functional/functional.py
hwangjeff: or, perhaps we could additionally clarify that the lengths of the leading dimensions need to be equal

mthrok: https://pytorch.org/docs/stable/notes/broadcasting.html ?

Let me look into that. I did not know about the case of different length.

mthrok: > or, perhaps we could additionally clarify that the lengths of the leading dimensions need to be equal

I'd go with firstly land this and followup for full broadcast support.

hwangjeff (approved):

torchaudio/prototype/functional/functional.py
hwangjeff: can you also make the same change to the docstring for `torchaudio.prototype.transforms.FFTConvolve`?

mthrok: done

carolineechen (approved):

vadimkantorov:
Btw could fftconvolve be upstreamed and merged with core conv1d (e.g. by introducing modes for extra mode for true convolution and a mode for fft)?

Fft convolutions are quite used in Hyena / S4 and similar state-space models, so making it more discoverable in core would be good

Issue #2885, opened 2022-12-03
Cannot create hubert_pretrain_large() due to missing attribute LayerNorm.conv


arlofaria:
### 🐛 Describe the bug

Here's a minimal example:

```python
import torchaudio
torchaudio.models.hubert_pretrain_large()
```

The error is `AttributeError: 'LayerNorm' object has no attribute 'conv'`

### Versions

From what I can tell, this is a regression between versions 0.12.1 and 0.13.0

nateanl:
oops, I made a mistake in `_init_hubert_pretrain_model` method. The line should be `if isinstance(module, components.ConvLayerBlock)` in 
https://github.com/pytorch/audio/blob/55e9978a68067f7f3b70ece667e25a12ed032a06/torchaudio/models/wav2vec2/model.py#L686

The error was not caught because there is no LayerNorm in the Base model, and I didn't test it on large and xlarge ones...

Let me make a fix and add tests for the pretrain models to make sure this won't happen again.

nateanl:
addressed in #2886, could you try with the latest codebase? let me know if you meet any other issue

arlofaria:
Seems to be working now, thanks!  I'll let you know if I run into any other issues.

PR #2886, opened 2022-12-03
Fix _init_hubert_pretrain_model

nateanl:
address #2885 

In `_init_hubert_pretrain_model ` method which initialize the hubert pretrain models, `kaiming_normal_` should be applied on `ConvLayerBlock` instead of `LayerNorm` layer. This PR fixes it and adds more unit tests.


hwangjeff (approved):
can you add a bit more context to the pr description?

PR #2895, opened 2022-12-07
Follow up on WavLM bundles

sgrigory:
Addressed @mthrok's comments in https://github.com/pytorch/audio/pull/2833:
- Moved model type from `_params` directly into the bundle definition. For now I defined model type as "WavLM" for WavLM bundles and "Wav2Vec2" for everything else. We can also distinguish between different Wav2Vec2 falvours - Hubert, VoxPopuli etc, but at the moment this won't imply any functional differences, so I didn't do it
- Expanded the title underline to match the title length

### DIFF
--- a/docs/source/models.rst
+++ b/docs/source/models.rst
@@ -59,6 +59,9 @@ Factory Functions
    hdemucs_low
    hdemucs_medium
    hdemucs_high
+   wavlm_model
+   wavlm_base
+   wavlm_large
 
 Utility Functions
 -----------------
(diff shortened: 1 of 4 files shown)


mthrok:
Hi @sgrigory

Thanks for the follow up. We also need to update the doc to mention wavelm.

- Update the docstring of Wav2Vec2Model to include factory functions for wavlm. https://github.com/pytorch/audio/blob/main/torchaudio/models/wav2vec2/model.py#L14-L18
- Add wavlm factory functions https://github.com/pytorch/audio/blob/main/docs/source/models.rst

mthrok (approved):
Looks good. Thanks!

nateanl (approved):

mthrok (approved):

PR #2896, opened 2022-12-07
Switch to nova Linux Wheel build

DanilBaibak:
Switch to Nova Linux Wheel build.

- [x] Disable the CircleCI builds for Linux Wheel.
- [x] Enable the Nova workflow for Linux Wheel.

~The Linux Wheel Python3.8 build has been kept because it is a dependency for the `docstring_parameters_sync` job.~ As Omkar pointed out, Docstring Parameters Sync also runs on GHA (https://github.com/pytorch/audio/actions/runs/3638187635/jobs/6140090209). So, we completely switched to the Nova Linux Wheel build.


.circleci/regenerate.py
```diff
@@ -49,6 +49,12 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     ):
                         # the fields must match the build_docs "requires" dependency
                         fb = "/.*/"
+
+                    # We'll stop building Linux Wheels from CircleCI.
+                    # Keeping the Python3.8 build just for docstring_parameters_sync until this job is migrated as well.
```
osalpekar, line 54: nit: Docstring Parameters Sync is also running on GHA so I think we should be good to remove the py3.8 cpu linux wheel job and the Dosctring Sync from CircleCI (if no opposition to this from the audio team :) ). If we do so, let's do it in a follow-up PR.

osalpekar (approved):
LGTM from my end, would love to know @hwangjeff 's thoughts

.github/workflows/build_wheels_linux.yml
```diff
@@ -24,21 +24,21 @@ jobs:
...
             pre-script: packaging/pre_build_script.sh
             post-script: packaging/post_build_script.sh
+            smoke-test-script: test/smoke_test/smoke_test.py
             package-name: torchaudio
     name: ${{ matrix.repository }}
     uses: pytorch/test-infra/.github/workflows/build_wheels_linux.yml@main
     with:
       repository: ${{ matrix.repository }}
-      ref: nightly
+      ref: ""
```
mthrok, line 33: What does this do? At a glance, it might as well be omitted.

.circleci/regenerate.py
```diff
@@ -49,13 +49,17 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     ):
                         # the fields must match the build_docs "requires" dependency
                         fb = "/.*/"
+
+                    # We'll stop building Linux Wheels from CircleCI.
+                    if os_type == "linux" and btype == "wheel":
+                        continue
```
mthrok, line 55: Can you move this to the top where other `continue` is performed?

.github/workflows/build_wheels_linux.yml
DanilBaibak: We want to use the main branch from the pytorch/test-infra repo.

PR #2899, opened 2022-12-08
Switch to Nova Linux Conda build

DanilBaibak:
Switch to Nova Linux Conda build.


.circleci/regenerate.py
```diff
@@ -49,6 +49,10 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     ):
                         # the fields must match the build_docs "requires" dependency
                         fb = "/.*/"
+
+                    if os_type == "linux" and btype == "conda":
```
osalpekar, line 53: We will need to keep around the py3.8 cu116 linux conda build because the docs_build job depends on that. See https://app.circleci.com/pipelines/github/pytorch/audio/13928 for reference.

.circleci/regenerate.py
```diff
@@ -49,6 +49,11 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     ):
                         # the fields must match the build_docs "requires" dependency
                         fb = "/.*/"
+
+                    # Keep the Python 3.8 cu116 build for docs builds until those are migrated as well.
```
mthrok, line 53: It would be nice if the build_doc job can also be migrated to GHA.
The thing is that upload_doc job requires credentials and I do not have them, so I could not do it. Perhaps release engineering team can configure the credential?

(Also currently the doc build is performed on top of binary build, but this incurs waiting time for binary build, which is not desirable. So it would be nice if we can make build_doc job separated from binary build job, by building torchaudio by itself).

.circleci/regenerate.py
```diff
@@ -87,18 +92,19 @@ def build_workflow_pair(btype, os_type, python_version, cu_version, filter_branc
...
-    if upload:
-
-        w.append(generate_upload_workflow(base_workflow_name, filter_branch, os_type, btype, cu_version))
-
-    if os_type != "macos":
-        pydistro = "pip" if btype == "wheel" else "conda"
-        w.append(
-            generate_smoketest_workflow(
-                pydistro, base_workflow_name, filter_branch, python_version, cu_version, os_type
+    skip_workflow = os_type == "linux" and btype == "conda" and python_version == "3.8" and cu_version == "cu116"
```
mthrok, line 95: It seems that moving this line to the call site and merge it into `upload` variable would make the logic simpler.

.circleci/regenerate.py
DanilBaibak: I created a follow up ticket [T141506623](https://www.internalfb.com/intern/tasks/?t=141506623).

.circleci/regenerate.py
```diff
@@ -45,6 +45,11 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     if (
                         (cu_version.startswith("rocm") and btype == "conda")
                         or (os_type == "linux" and btype == "wheel")
+                        or (
+                            os_type == "linux"
+                            and btype == "conda"
+                            and (python_version != "3.8" or cu_version != "cu116")
```
osalpekar, line 51: IIUC we want to remove all linux conda jobs except the py3.8 cu116 build? I think this should be 
`and (python_version != "3.8" and cu_version != "cu116")` in that case right? The regenerated `config.yml` seems to have removed/changed more jobs than intended.

DanilBaibak: Correct, we want to remove all linux conda jobs except `py3.8 cu116`. And the [config.yml for the current PR](https://github.com/pytorch/audio/blob/b2a6ee5a36afc6776de62daf3b105010a7b92b78/.circleci/config.yml#L1084-L1096) looks appropriate.

mthrok (approved):

seemethere (approved):

PR #2901, opened 2022-12-08
Toggle on/off ffmpeg test if needed

atalman:
Toggle on/off ffmpeg test if needed
By default it ON, hence should not affect any current tests.
To toggle ON no change required.
To toggle OFF use:
```
smoke_test.py --no-ffmpeg
```

To be used when calling from builder currently. Since we do not install ffmpeg currently.

### DIFF
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -17,8 +17,9 @@ def ffmpeg_test():
 
 def main() -> None:
     parser = argparse.ArgumentParser()
+
+    # Warning: Please note this option should not be widely used, only use it when absolutely necessary
     parser.add_argument("--no-ffmpeg", dest="ffmpeg", action="store_false")
-    parser.set_defaults(ffmpeg=True)
 
     options = parser.parse_args()
     if options.ffmpeg:
--- a/test/smoke_test/smoke_test.py
+++ b/test/smoke_test/smoke_test.py
@@ -1,6 +1,7 @@
 """Run smoke tests"""
 
 import argparse
+
 import torchaudio  # noqa: F401
 import torchaudio.compliance.kaldi  # noqa: F401
 import torchaudio.datasets  # noqa: F401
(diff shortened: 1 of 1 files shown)


test/smoke_test/smoke_test.py
```diff
@@ -9,4 +9,22 @@
...
+
+
+def streamreader_test():
+    from torchaudio.io import StreamReader  # noqa: F401
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--streamreader", action="store_true")
+    parser.add_argument("--no-streamreader", dest="streamreader", action="store_false")
```
carolineechen, line 21: I think this can be achieved w/ just one argument

atalman:
> This is very sensitive change, and I am not in favor of this, because the reason this test was added is to makes sure that FFmpeg installation (a version different from the one used for build) is properly integrated, which constitutes the most basic smoke test.
> 
> If this option is abused, the binary being tested will be in a bad shape. Can you elaborate on usecases?

This is to be used when calling from https://pytorch.org/get-started/locally/ and release validation workflows in https://github.com/pytorch/builder. Which do not install ffmpeg  currently.

test/smoke_test/smoke_test.py
mthrok: Also, can we have some warning and note on this option?
The use of this argument should restricted to minimal.

atalman: Done

test/smoke_test/smoke_test.py
```diff
@@ -9,4 +9,21 @@
...
+
+
+def ffmpeg_test():
+    from torchaudio.io import StreamReader  # noqa: F401
+
+
+def main() -> None:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--no-ffmpeg", dest="ffmpeg", action="store_false")
+    parser.set_defaults(ffmpeg=True)
```
mthrok, line 21: I think `action="store_false"` will set the default value to `True`, so the line `parser.set_defaults(ffmpeg=True)` is redundant.

mthrok (approved):

test/smoke_test/smoke_test.py
atalman: Done

test/smoke_test/smoke_test.py
atalman: Done!

PR #2907, opened 2022-12-09
Switch to Nova MacOS Wheel

DanilBaibak:
Switch to Nova MacOS and M1 Wheels. This PR is a step in migrating from CircleCI to the Nova workflow.

- [x] Disable the CircleCI builds for MacOS Wheel.
- [x] Disable the CircleCI builds for M1 Wheel.
- [x] Enable the Nova workflow for MacOS Wheel.
- [x] Enable the Nova workflow for M1 Wheel.


.circleci/regenerate.py
```diff
@@ -49,6 +49,10 @@ def build_workflows(prefix="", upload=False, filter_branch=None, indentation=6):
                     ):
                         # the fields must match the build_docs "requires" dependency
                         fb = "/.*/"
+
+                    if os_type == "macos" and btype == "wheel":
+                        continue
```
mthrok, line 54: Can we perform all the `continue` at the beginning at the same place as the existing one? (where "rocm" is checked)

.github/workflows/build-m1-binaries.yml
```diff
@@ -15,88 +15,6 @@ on:
 env:
   CHANNEL: "nightly"
 jobs:
-  build_wheels:
```
mthrok, line 18: why is build_wheels job removed from GHA?

DanilBaibak: It is old one. In this PR, we switch to the latest workflow for the M1 Wheel. https://github.com/pytorch/audio/pull/2907/files#diff-3dab8d7b6418e0047d1bdd73ae68c53d57e16fa2972a16060b86636140c3fb23L41-L44

mthrok (approved):

mthrok (approved):

PR #2908, opened 2022-12-09
Switch to Nova MacOS Conda

DanilBaibak:
Switch to Nova M1 Conda


.github/workflows/build-m1-binaries.yml
```diff
@@ -97,87 +97,6 @@ jobs:
           for pkg in dist/*; do
             aws s3 cp "$pkg" "s3://pytorch/whl/${CHANNEL}/cpu/" --acl public-read
           done
-  build_conda:
```
mthrok, line 100: Why is this definition removed?

mthrok (approved):

.github/workflows/build-m1-binaries.yml
osalpekar: `build-m1-binaries.yml` is actually the old (non-Nova) pipeline for publishing M1 conda/wheel binaries. Now that the Nova-based workflow will start publishing (in `build-conda-m1.yml`), this job in the old pipeline can be deleted. We should delete this file entirely once Nova-based M1 binaries are publishing nightlies regularly.

mthrok (approved):
stamp

seemethere (approved):

PR #2913, opened 2022-12-10
Fix type of arguments in torchaudio.io classes

nateanl:
The `src` or `dst` argument can be `str` or `file-like object`. Setting it to `str` in type annotation will confuse users that it only accepts `str` type.

### DIFF
--- a/torchaudio/io/_compat.py
+++ b/torchaudio/io/_compat.py
@@ -1,4 +1,4 @@
-from typing import Dict, Optional, Tuple
+from typing import BinaryIO, Dict, Optional, Tuple
 
 import torch
 import torchaudio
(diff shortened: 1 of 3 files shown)


mthrok (commented):
Rather than removing the annotation, why not `Union[str, BinaryIO]`?

https://docs.python.org/3/library/typing.html#typing.BinaryIO

nateanl:
yeah I was not sure which IO class it uses, will change to that

torchaudio/io/_compat.py
```diff
@@ -105,7 +105,7 @@ def load_audio(
 
 
 def load_audio_fileobj(
-    src: str,
+    src: Union[str, BinaryIO],
```
mthrok, line 108: This function does not accept `str`.

PR #2916, opened 2022-12-12
Fallback to best_effort_timestamp in case of invalid PTS

mthrok:
If the input video has invalid PTS, the current precise seek fails except when seeking into t=0.

This commit updates the discard mechanism to fallback to `best_effort_timestamp` in such cases.

`best_effort_timestamp` is just the number of frames went through decoder starting from the beginning of the file.

This means if the input file is very long, but seeking towards the end of the file, the StreamReader still decodes all the frames.

For videos with valid PTS, `best_effort_timestamp` should be same as `pts`. [[src](https://ffmpeg.org/doxygen/4.1/decode_8c.html#a8d86329cf58a4adbd24ac840d47730cf)]


jdsgomes:
Thanks for making the change. it looks good to me, happy to approve once its ready

mthrok:
@jdsgomes Do you know where the video "RATRACE_wave_f_nm_np1_fr_goo_37.avi" come from?
Looking at https://github.com/pytorch/vision/pull/1303, it seems from one of  HMDB-51, UCF-101 and Kinetics-400, but I cannot tell which and I need to check the license.

jdsgomes:
> @jdsgomes Do you know where the video "RATRACE_wave_f_nm_np1_fr_goo_37.avi" come from? Looking at [pytorch/vision#1303](https://github.com/pytorch/vision/pull/1303), it seems from one of HMDB-51, UCF-101 and Kinetics-400, but I cannot tell which and I need to check the license.

@bjuncek do you know?

YosuaMichael:
@mthrok You can see where it comes from in this readme: https://github.com/pytorch/vision/blob/main/test/assets/videos/README

and seems like the video RATRACE_wave_f_nm_np1_fr_goo_37.avi comes from hmdb51

PR #2921, opened 2022-12-14
Add HiFiGAN bundle

sgrigory:
Closes [T138011314](https://www.internalfb.com/intern/tasks/?t=138011314)
## Description
- Add  bundle `HIFIGAN_GENERATOR_V3_LJSPEECH` to prototypes. The bundle contains pre-trained HiFiGAN generator weights from the [original HiFiGAN publication](https://github.com/jik876/hifi-gan#pretrained-model), converted slightly to fit our model
- Add tests
  - unit tests checking that vocoder and mel-transform implementations in the bundle give the same results as the original ones. Part of the original HiFiGAN code is ported to this repo to enable these tests
  - integration test checking that waveform reconstructed from mel spectrogram by the bundle is close enough to the original
- Add docs


test/torchaudio_unittest/common_utils/gdrive_utils.py
```diff
@@ -0,0 +1,34 @@
+import requests
```
mthrok, line 1: We cannot use external resource in unit tests.
Also Google Drive throttle access. So it's unreliable.
What data is it?

sgrigory: @mthrok It's the pre-trained model from the original HiFiGAN publication (https://github.com/jik876/hifi-gan#pretrained-model), they put the weights on Google Drive: https://drive.google.com/drive/folders/1-eEYTB5Av9jNql0WGBlRoi-WH2J7bp5Y

In the test I'm downloading them to check that our bundle has the same weights. Or do you think this test is redundant? Alternatively, we can upload the original weights to `download.pytorch.org` too.

nateanl: The output of HiFiGAN is waveform, we can use similar testing method in https://github.com/pytorch/audio/blob/main/test/integration_tests/source_separation_pipeline_test.py which calculate the Si-SNR score of estimated waveform and the clean reference.

Here we can use a clean reference waveform, generate the melspectrogram, and feed it to the pre-trained HiFiGAN to get the estimated waveform. We can compute Si-SNR score locally as the threshold for unit testing.

mthrok: If the goal is to ensure the parity with the original model, then we do not need the trained weights. Just port the original implementation as reference implementation in test, and feed synthetic data to check if the outputs are same.

test/integration_tests/prototype/hifi_gan_pipeline_test.py
```diff
@@ -0,0 +1,65 @@
+import os
+import sys
+
+import pytest
+
+sys.path.append(os.path.join(os.path.dirname(__file__), "..", "..", "..", "examples"))
```
mthrok, line 6: @nateanl This makes me wonder if we should move sdr to library.

test/integration_tests/prototype/hifi_gan_pipeline_test.py
```diff
@@ -0,0 +1,65 @@
...
+        f_min=f_min,
+        f_max=f_max,
+        n_mels=bundle._params["in_channels"],
+        normalized=False,
+        mel_scale="slaney",
+        norm="slaney",
+        center=True,
+    )
+    mel_spectrogram = mel_transform(ref_waveform)
+    mel_spectrogram = torch.log(torch.clamp(mel_spectrogram, min=1e-5))
```
mthrok, line 55: The design principle of Bundle object is to free users from re-implementing the expected preprocessing like this. 

Please checkout the followings for the detail

- https://pytorch.org/audio/stable/pipelines.html#torchaudio-pipelines
- https://pytorch.org/audio/stable/tutorials/tacotron2_pipeline_tutorial.html#phoneme-based-encoding

Can you move the construction of preprocessing pipeline into bundle?

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,125 @@
...
+        >>>
+        >>> # Use Tactron2 to convert text to mel spectrogram
+        >>> text = "A quick brown fox jumped over a lazy dog"
+        >>> input, lengths = processor(text)
+        >>> specgram, lengths, _ = tacotron2.infer(input, lengths)
+        >>>
+        >>> # Load HiFiGAN bundle
+        >>> vocoder = bundle_hifigan.get_vocoder()
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 5.59M/5.59M [00:03<00:00, 1.55MB/s]
```
mthrok, line 62: Can you remove redundant `█`, so that it fits the 120 line limit? (I think 80 chars is enough)

test/integration_tests/prototype/hifi_gan_pipeline_test.py
nateanl: yeah we should include `Si-SDR` into core. In terms of `SDR`, the implementation in `bss_eval` can deal with signal pairs that are not aligned (for example, if there are some samples shifted, the SDR score still works fine). We might want to consider that when implementing `SDR` metric.

torchaudio/prototype/pipelines/hifigan_pipeline.py
sgrigory: Done

test/integration_tests/prototype/hifi_gan_pipeline_test.py
sgrigory: Sounds good, I've added:
 - a class `HiFiGANMelSpectrogram` performing waveform-to-mel transform in the same way as original HiFiGAN code
 - a factory function `hifigan_mel_spectrogram`
 - `HiFiGANGeneratorBundle.get_mel_transform` method, which returns `HiFiGANMelSpectrogram`
 - a unit test which checks that `HiFiGANMelSpectrogram` is indeed equivalent to the original pipeline

That said, the integration test for reconstructing waveform from a real audio is not properly working yet - even the shift-invariant metric @nateanl suggested gives negative SNR. The reconstructed audio has a noticeable noise, maybe that's the reason - I'll keep looking into this

test/integration_tests/prototype/hifi_gan_pipeline_test.py
sgrigory: In the integration test I've switched to the SDR metric computed by `mir_eval.separation.bss_eval_sources` - this metric is less negative, but still negative, I haven't found the reason yet

test/integration_tests/prototype/hifi_gan_pipeline_test.py
mthrok: > Sounds good, I've added:
> 
> * a class `HiFiGANMelSpectrogram` performing waveform-to-mel transform in the same way as original HiFiGAN code
> * a factory function `hifigan_mel_spectrogram`
> * `HiFiGANGeneratorBundle.get_mel_transform` method, which returns `HiFiGANMelSpectrogram`
> * a unit test which checks that `HiFiGANMelSpectrogram` is indeed equivalent to the original pipeline
> 

Another thing about the design around the Bundle is that we want to make it convenient but we want to limit the number of public interfaces we introduce.

In PyTorch, we make it rule to take 2 release cycles for BC-breaking changes to stable public interfaces. That's like 6+ month, which is a long time. (prototypes are okay to break backward compatibility anytime) So any public interface is basically maintenance cost since right when it is introduced.

For this reason, in Bundle and pipelines, the classes that implement pre/post-processing, (or even model structures themself) are not public interface, but their calling conventions (the signatures of `forward` method) are. The design is centered around the usage of pre-trained models, and we take advantage of the fact that Python is Duck Typing language here.

So, it is important to make pre-processings accessible for users, but pre-processings implementation themself are not important enough for us to declare them as public. Otherwise, we have to introduce ton of preprocessing classes as public interface, which is not maintainable.

An fine example is [source separation bundle](https://pytorch.org/audio/stable/pipelines.html#source-separation). [CONVTASNET_BASE_LIBRI2MIX](https://pytorch.org/audio/stable/generated/torchaudio.pipelines.CONVTASNET_BASE_LIBRI2MIX.html#torchaudio.pipelines.CONVTASNET_BASE_LIBRI2MIX) and [HDEMUCS_HIGH_MUSDB_PLUS](https://pytorch.org/audio/stable/generated/torchaudio.pipelines.HDEMUCS_HIGH_MUSDB_PLUS.html#torchaudio.pipelines.HDEMUCS_HIGH_MUSDB_PLUS) share the same usage and interface but the returned models are completely different as they process different kind of audios.

For this reason,

- We do not need a factory function `hifigan_mel_spectrogram`.
- I am inclined to call `HiFiGANMelSpectrogram` to something a bit more general, like preprocessing, but in the context of GAN training preprocessing is too ambiguous so I am open to other ideas.

mthrok: wait, actually, I might be misunderstanding an important thing. What's the expected usecase/user-flow for this bundle? For the actual vocoding, they are connected with some *other* mechanism that generates melspectrogram, right, and is such mechanism going to be added to this bundle??

docs/source/prototype.pipelines.rst
```diff
@@ -24,3 +24,29 @@ EMFORMER_RNNT_BASE_TEDLIUM3
...
+``HiFiGANGeneratorBundle`` defines HiFiGAN Generator pipeline capable of transforming mel spectrograms into waveforms.
+
+.. autosummary::
+   :toctree: generated
+   :nosignatures:
+   :template: autosummary/bundle_class.rst
+
+   HiFiGANGeneratorBundle
+
+.. minigallery:: torchaudio.prototype.pipelines.HiFiGANGeneratorBundle
```
mthrok, line 44: Since there is no tutorial using this bundle, this line can be removed.

test/integration_tests/prototype/hifi_gan_pipeline_test.py
sgrigory: For the complete flow we would need an external text-to-melspectrogram mechanism - for example, the one from Tacotron2. This allows a text -> melspectrogram -> waveform flow. Here is the example snippet: https://github.com/pytorch/audio/pull/2921/files#diff-12a2fb327badd19aa9ee81293aa2815ae0249fbb3cfada531e5552af6d02dd33R43-R66

In this case do we need to add Tacotron2 text-to-melspectrogram component to the HiFiGAN bundle? If we want to use HiFiGAN with different  text-to-melspectrogram components, would we define a bundle for each combination? I see that this is indeed how Tacotron2 bundles are organized https://github.com/pytorch/audio/blob/f0986b6044c43ac2cb17afb42352fadd43c5deb2/torchaudio/pipelines/_tts/impl.py#L187-L203

sgrigory:
Update:
- Copied parts of the original HiFiGAN implementation necessary for unit tests into `test/torchaudio_unittest/prototype/hifi_gan/original/`, instead of cloning the original repo during the test. Also modified pre-commit config to ignore that path. In principle, one could do it with `skip` directives in the files, but it looks like `usort` [doesn't support file-level skip directives](https://usort.readthedocs.io/en/stable/guide.html#directives), so that would mean placing a directive next to every import
- Modified the integration test for HiFiGAN: now it measures reconstruction quality by relative log-MSE of mel spectrograms of the reconstructed and reference audio. I use log because that's also what [HiFiGAN uses for creating mel spectrograms](https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/meldataset.py#L28) . In log scale the spectrograms look quite similar, so the test passes with relative threshold 2%. Without log the relative threshold is much higher
- Made `HiFiGANMelSpectrogram` non-public - removed from docs and from type annotation of public methods

docs/source/prototype.pipelines.rst
sgrigory: Done

sgrigory:
To illustrate the point why log transform of mel spectrograms in the test makes sense:
Screenshot 2022-12-20 at 13 20 42 Screenshot 2022-12-20 at 13 20 33
Screenshot 2022-12-20 at 13 22 11 Screenshot 2022-12-20 at 13 22 03

test/torchaudio_unittest/prototype/hifi_gan/hifi_gan_test_impl.py
```diff
@@ -134,3 +107,26 @@ def test_original_implementation_match(self):
         ref_output = model_ref(inputs)
         output = model(inputs)
         self.assertEqual(ref_output, output)
+
+    def test_mel_transform(self):
+        """Check that HiFiGANMelSpectrogram generates same mel spectrogram as the original HiFiGAN implementation
+        when applied on a synthetic waveform.
+        There seems to be no way to change dtype in the original implmentation, so we feed in the waveform with the
```
nateanl, line 114: I feel it's okay to change dtype in `ref_mel_spectrogram`, which part cannot change the dtype of tensor?

docs/source/prototype.pipelines.rst
```diff
@@ -24,3 +24,27 @@ EMFORMER_RNNT_BASE_TEDLIUM3
 
    .. autodata:: EMFORMER_RNNT_BASE_TEDLIUM3
       :no-value:
+
+
+HiFiGAN Generator
```
nateanl, line 29: Although the model and factory functions are named by Generator, I think we can make the bundle title to be "HiFiGAN Vocoder" so that users can better understand what is the model used for.

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -334,3 +335,85 @@ def hifigan_generator_v3() -> HiFiGANGenerator:
...
+        ref_waveform = ref_waveform.squeeze(1)
+
+        spectr = (self.mel_transform.spectrogram(ref_waveform) + 1e-9) ** 0.5
+        mel_spectrogram = self.mel_transform.mel_scale(spectr)
+        mel_spectrogram = torch.log(torch.clamp(mel_spectrogram, min=1e-5))
+        return mel_spectrogram
+
+
+def hifigan_mel_spectrogram() -> HiFiGANMelSpectrogram:
+    r"""Builds :py:class:`HiFiGANMelSpectrogram` with parameters used to prepare data for pre-training models
```
nateanl, line 410:
```suggestion
    r"""Builds :py:class:`HiFiGANMelSpectrogram` with parameters used to prepare data for training models
```

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -334,3 +335,85 @@ def hifigan_generator_v3() -> HiFiGANGenerator:
...
+    in the original HiFiGAN publication. The parameters are taken from
+    `here <https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/config_v3.json#L18-L27>`_.
+
+    Returns:
+        HiFiGANMelSpectrogram: Generated mel transformation object.
```
nateanl, line 415:
```suggestion
        HiFiGANMelSpectrogram: Generated MelSpectrogram transform object.
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+    to generate mel spectrogram from text. Please see below for the code example.
+
+    .. note::
+        Although this works with the existing Tactron2 bundles, for the best results one needs to retrain Tactron2
+        using the same data preprocessing pipeline which was used for training HiFiGAN. In particular, the original
+        HiFiGAN implemntation uses a custom method of generating mel spectrograms from waveforms, different from
+        :py:class:`torchaudio.transforms.MelSpectrogram`. We reimplemented this transform as
+        :py:meth:`get_mel_transform`.
+
+    Example: Synthetic spectrogram to audio.
```
nateanl, line 34:
```suggestion
    Example: Transform synthetic mel spectrogram to audio.
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        >>> vocoder = bundle.get_vocoder()
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|████████████| 5.59M/5.59M [00:00<00:00, 18.7MB/s]
+        >>>
+        >>> # Generate synthetic mel spectrogram
+        >>> specgram = torch.sin(0.5 * torch.arange(start=0, end=100)).expand(bundle._params["in_channels"], 100)
+        >>>
+        >>> # Trasform mel spectrogram into audio
+        >>> waveforms = vocoder(specgram)
+        >>> torchaudio.save('sample.wav', waveforms, bundle.sample_rate)
```
nateanl, line 50:
```suggestion
        >>> waveform = vocoder(specgram)
        >>> torchaudio.save('sample.wav', waveform, bundle.sample_rate)
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        >>> specgram, lengths, _ = tacotron2.infer(input, lengths)
+        >>>
+        >>> # Load HiFiGAN bundle
+        >>> vocoder = bundle_hifigan.get_vocoder()
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|██████████████████████████████████████████████████████ 5.59M/5.59M [00:03<00:00, 1.55MB/s]
+        >>>
+        >>> # Use HiFiGAN to convert mel spectrogram to audio
+        >>> waveforms = vocoder(specgram).squeeze(0)
+        >>> torchaudio.save('sample.wav', waveforms, bundle_hifigan.sample_rate)
```
nateanl, line 75:
```suggestion
        >>> waveform = vocoder(specgram).squeeze(0)
        >>> torchaudio.save('sample.wav', waveform, bundle_hifigan.sample_rate)
```

test/torchaudio_unittest/prototype/hifi_gan/hifi_gan_test_impl.py
sgrigory: If I feed `float64` into `ref_mel_spectrogram`, it throws an error 
```
RuntimeError: expected scalar type Double but found Float
```
on the line `spec = torch.matmul(mel_basis[str(fmax)+'_'+str(y.device)], spec)`. I think this is because `mel_basis` is created with explicit type `float` [here](https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/meldataset.py#L58). Now that we've copied the code into our repo we can of course modify it and enable passing `float64`. Do we want to do it or do we want to keep the copied code as close as possible to the original?

docs/source/prototype.pipelines.rst
sgrigory: Sure, renamed the bundle to "HiFiGAN Vocoder"

test/torchaudio_unittest/prototype/hifi_gan/original/meldataset.py
```diff
@@ -0,0 +1,45 @@
+# flake8: noqa
+# fmt: off
```
mthrok, line 2: Does fbcode linter recognize this?

mthrok: Alternatively, I think it's okay to format the code. It's simpler and no special treatment to carry on.

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        >>>
+        >>> # Load the HiFiGAN bundle
+        >>> vocoder = bundle.get_vocoder()
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|████████████| 5.59M/5.59M [00:00<00:00, 18.7MB/s]
+        >>>
+        >>> # Generate synthetic mel spectrogram
+        >>> specgram = torch.sin(0.5 * torch.arange(start=0, end=100)).expand(bundle._params["in_channels"], 100)
+        >>>
+        >>> # Trasform mel spectrogram into audio
```
nateanl, line 48:
```suggestion
        >>> # Transform mel spectrogram into audio
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        >>> waveform = vocoder(specgram)
+        >>> torchaudio.save('sample.wav', waveform, bundle.sample_rate)
+
+    Example: Usage together with Tactron2, text to audio.
```
nateanl, line 52: Could you update all `tactron2` with `tacotron2`? Thanks.
```suggestion
    Example: Usage together with Tacotron2, text to audio.
```

nateanl (approved):
I think the PR is good now. @mthrok could you also review and approve it?

docs/source/prototype.pipelines.rst
```diff
@@ -24,3 +24,27 @@ EMFORMER_RNNT_BASE_TEDLIUM3
 
    .. autodata:: EMFORMER_RNNT_BASE_TEDLIUM3
       :no-value:
+
+
+HiFiGAN Vocoder
+-----------------
```
mthrok, line 30:
```suggestion
---------------
```

test/integration_tests/prototype/hifi_gan_pipeline_test.py
```diff
@@ -0,0 +1,55 @@
...
+    bundle = HIFIGAN_VOCODER_V3_LJSPEECH
+
+    # Get HiFiGAN-compatible transformation from waveform to mel spectrogram
+    mel_transform = bundle.get_mel_transform()
+    # Get HiFiGAN vocoder
+    vocoder = bundle.get_vocoder()
+
+    ref_waveform, ref_sample_rate = torchaudio.load(sample_speech)
+    ref_waveform = F.resample(ref_waveform, orig_freq=ref_sample_rate, new_freq=bundle.sample_rate)
+    ref_waveform = ref_waveform[:, : -(ref_waveform.shape[1] % mel_transform.hop_size)]
```
mthrok, line 32: Looks like the test logic does not concern if the data used should be real speech or not. Can we replace this with some pseudo data? Since it's about (mel-)spectrogram,I think a sine sweep that covers all the frequency range bellow Nyquist frequency is well suited here.

something like [this](https://pytorch.org/audio/main/tutorials/oscillator_tutorial.html#changing-frequencies-across-time)

.pre-commit-config.yaml
```diff
@@ -5,7 +5,7 @@ repos:
       - id: check-docstring-first
       - id: check-toml
       - id: check-yaml
-        exclude: packaging/.*
+        exclude: (packaging/.*|test/torchaudio_unittest/prototype/hifi_gan/original/.*)
```
mthrok, line 8: Can we format the reference implementation and get rid of this?
It's better to have less special treatment from the perspective of library maintenance cost.

test/integration_tests/prototype/hifi_gan_pipeline_test.py
```diff
@@ -0,0 +1,55 @@
...
+
+import torch
+import torchaudio
+import torchaudio.functional as F
+from torchaudio.prototype.pipelines import HIFIGAN_VOCODER_V3_LJSPEECH
+
+
+def test_smoke_hifi_gan_bundle():
+    """Smoke test of downloading weights for pretraining model"""
+    HIFIGAN_VOCODER_V3_LJSPEECH.get_vocoder()
```
mthrok, line 11: This test looks redundant. Can we remove it?

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -334,3 +335,85 @@ def hifigan_generator_v3() -> HiFiGANGenerator:
         in_channels=80,
         lrelu_slope=0.1,
     )
+
+
+class HiFiGANMelSpectrogram(torch.nn.Module):
```
mthrok, line 340: Let's not have this as a public API yet.
Please move it to bundle implementation and prefix with underscore.

torchaudio/prototype/models/hifi_gan.py
```diff
@@ -334,3 +335,85 @@ def hifigan_generator_v3() -> HiFiGANGenerator:
...
+        ref_waveform = F.pad(waveform.unsqueeze(1), (self.pad_size, self.pad_size), mode="reflect")
+        ref_waveform = ref_waveform.squeeze(1)
+
+        spectr = (self.mel_transform.spectrogram(ref_waveform) + 1e-9) ** 0.5
+        mel_spectrogram = self.mel_transform.mel_scale(spectr)
+        mel_spectrogram = torch.log(torch.clamp(mel_spectrogram, min=1e-5))
+        return mel_spectrogram
+
+
+def hifigan_mel_spectrogram() -> HiFiGANMelSpectrogram:
```
mthrok, line 409: Please remove this.

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        Returns:
+            Variation of :py:class:`~torchaudio.prototype.models.HiFiGANGenerator`.
+        """
+        model = hifigan_generator(**self._params)
+        model.load_state_dict(self._get_state_dict(dl_kwargs))
+        model.eval()
+        return model
+
+    def get_mel_transform(self) -> Module:
+        """Construct an object which transforms waveforms into mel spectrograms.
```
mthrok, line 106: The implementation detail should not be mentioned in the interface docstring. 

Let's say, there is a possibility in the future we might add a HiFiGAN pre-trained model using torchaudio's melspectrogram. We want each bundle to use the appropriate melspectrogram pipeline.

Can you move the explanation of the melspectrogram to the docstring of `HIFIGAN_VOCODER_V3_LJSPEECH`?

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        100%|██████████████████████████████████████████████████████ 5.59M/5.59M [00:03<00:00, 1.55MB/s]
+        >>>
+        >>> # Use HiFiGAN to convert mel spectrogram to audio
+        >>> waveform = vocoder(specgram).squeeze(0)
+        >>> torchaudio.save('sample.wav', waveform, bundle_hifigan.sample_rate)
+    """  # noqa: E501
+
+    _path: str
+    _params: Dict[str, Any]  # Vocoder parameters
+    _mel_params: Dict[str, Any]  # Mel transformation parameters
```
mthrok, line 80: Can we have sample rate as a parameter independent from `_mel_params`?

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|██████████████████████████████████████████████████████ 5.59M/5.59M [00:03<00:00, 1.55MB/s]
+        >>>
+        >>> # Use HiFiGAN to convert mel spectrogram to audio
+        >>> waveform = vocoder(specgram).squeeze(0)
+        >>> torchaudio.save('sample.wav', waveform, bundle_hifigan.sample_rate)
+    """  # noqa: E501
+
+    _path: str
+    _params: Dict[str, Any]  # Vocoder parameters
```
mthrok, line 79: Why not call the attribute `_vocoder_params`?

```suggestion
    _vocoder_params: Dict[str, Any]
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        "in_channels": 80,
+        "lrelu_slope": 0.1,
+    },
+    _mel_params={
+        "hop_size": 256,
+        "n_fft": 1024,
+        "win_length": 1024,
+        "f_min": 0,
+        "f_max": 8000,
+        "sample_rate": 22050,
```
mthrok, line 145: Sample rate should be a high-level attribute, instead of `_mel_params`

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+        >>>
+        >>> # Use Tactron2 to convert text to mel spectrogram
+        >>> text = "A quick brown fox jumped over a lazy dog"
+        >>> input, lengths = processor(text)
+        >>> specgram, lengths, _ = tacotron2.infer(input, lengths)
+        >>>
+        >>> # Load HiFiGAN bundle
+        >>> vocoder = bundle_hifigan.get_vocoder()
+        Downloading: "https://download.pytorch.org/torchaudio/models/hifigan_generator_v3_ljspeech.pth"
+        100%|██████████████████████████████████████████████████████ 5.59M/5.59M [00:03<00:00, 1.55MB/s]
```
mthrok, line 71:
```suggestion
        100%|██████████████████████████ 5.59M/5.59M [00:03<00:00, 1.55MB/s]
```

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,155 @@
...
+    This bundle can convert mel spectrorgam to waveforms and vice versa. A typical use case would be a flow like
+    `text -> mel spectrogram -> waveform`, where one can use an external component, e.g. Tactron2,
+    to generate mel spectrogram from text. Please see below for the code example.
+
+    .. note::
+        Although this works with the existing Tactron2 bundles, for the best results one needs to retrain Tactron2
+        using the same data preprocessing pipeline which was used for training HiFiGAN. In particular, the original
+        HiFiGAN implemntation uses a custom method of generating mel spectrograms from waveforms, different from
+        :py:class:`torchaudio.transforms.MelSpectrogram`. We reimplemented this transform as
+        :py:meth:`get_mel_transform`.
```
mthrok, line 32: This note is specific to `HIFIGAN_VOCODER_V3_LJSPEECH` instance, so it is more appropriate to move there.

sgrigory: Moved it to the docstring of `HIFIGAN_VOCODER_V3_LJSPEECH`

torchaudio/prototype/pipelines/hifigan_pipeline.py
sgrigory: Done

torchaudio/prototype/pipelines/hifigan_pipeline.py
sgrigory: Done

torchaudio/prototype/pipelines/hifigan_pipeline.py
sgrigory: Moved it to the docstring of HIFIGAN_VOCODER_V3_LJSPEECH

torchaudio/prototype/models/hifi_gan.py
sgrigory: Done

.pre-commit-config.yaml
sgrigory: Formatted the ported code and removed the formatting exceptions here and in the code

test/integration_tests/prototype/hifi_gan_pipeline_test.py
sgrigory: Sure, I replaced the real audio with synthetic one which changes its frequency from 0 to Nyquist frequency.

The threshold on relative RME had to be raised from 2% to 4% though. This is probably because HiFiGAN was trained on mel spectrograms of real audio clips (English speech), and might work worse on artificial waveforms like the one I'm using.

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+
+    @property
+    def sample_rate(self):
+        """Sample rate of the audio that the model is trained on.
+
+        :type: float
+        """
+        return self._sample_rate
+
+    class _HiFiGANMelSpectrogram(torch.nn.Module):
```
mthrok, line 118: Please unnest this. `_HiFiGANMelSpectrogram` should not belong to `HiFiGANVocoderBundle`. 
The later defines an interface and it should not have a logical relationship with an implementation of pipeline.

docs/source/prototype.pipelines.rst
```diff
@@ -24,3 +24,29 @@ EMFORMER_RNNT_BASE_TEDLIUM3
...
+.. autosummary::
+   :toctree: generated
+   :nosignatures:
+   :template: autosummary/bundle_class.rst
+
+   HiFiGANVocoderBundle
+   HiFiGANVocoderBundle.get_mel_transform
+   HiFiGANVocoderBundle.get_vocoder
+
+Pretrained Models
```
mthrok, line 46: Please use `autosummary` with `bundle_data.rst` template.
ex) https://github.com/pytorch/audio/blob/main/docs/source/pipelines.rst?plain=1#L115-L123

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        "f_min": 0,
+        "f_max": 8000,
+    },
+    _sample_rate=22050,
+)
+HIFIGAN_VOCODER_V3_LJSPEECH.__doc__ = """Pre-trained HiFiGAN Vocoder pipeline, transforming mel spectrograms into
+    waveforms. The underlying model is constructed by :py:func:`torchaudio.prototype.models.hifigan_generator`
+    and utilizes weights trained on *The LJ Speech Dataset* :cite:`ljspeech17`. The weights are converted from the ones
+    published with the original paper :cite:`NEURIPS2020_c5d73680` (See links to pre-trained models on
+    `GitHub <https://github.com/jik876/hifi-gan#pretrained-model>`__).
```
mthrok, line 212: The first sentence of the docstring is used as a summary for `autosummary`. 
ex) https://pytorch.org/audio/stable/pipelines.html#id14

Can you rephrase the first sentence to mention datasets? Something like the following would do.

"Pre-trained HiFiGAN Vocoder pipeline, trained on *The LJ Speech Dataset* :cite:`ljspeech17`"

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,224 @@
...
+        "hop_size": 256,
+        "n_fft": 1024,
+        "win_length": 1024,
+        "f_min": 0,
+        "f_max": 8000,
+    },
+    _sample_rate=22050,
+)
+HIFIGAN_VOCODER_V3_LJSPEECH.__doc__ = """Pre-trained HiFiGAN Vocoder pipeline, transforming mel spectrograms into
+    waveforms. The underlying model is constructed by :py:func:`torchaudio.prototype.models.hifigan_generator`
```
mthrok, line 209: > The underlying model is constructed by :py:func:`torchaudio.prototype.models.hifigan_generator`

Perhaps `vocoder` instead of `model` is clearer.

torchaudio/prototype/pipelines/hifigan_pipeline.py
mthrok, line 212: > The weights are converted from the one published with the original paper

Can you also mention and link the license of the original model?

sgrigory: Linked MIT licence from the original repo

docs/source/prototype.pipelines.rst
sgrigory: Done

torchaudio/prototype/pipelines/hifigan_pipeline.py
sgrigory: Done

torchaudio/prototype/pipelines/hifigan_pipeline.py
mthrok: oh sorry about that, yeah, that's what I meant.

torchaudio/prototype/pipelines/hifigan_pipeline.py
```diff
@@ -0,0 +1,226 @@
...
+        "f_max": 8000,
+    },
+    _sample_rate=22050,
+)
+HIFIGAN_VOCODER_V3_LJSPEECH.__doc__ = """HiFiGAN Vocoder pipeline, trained on *The LJ Speech Dataset*
+    :cite:`ljspeech17`. The underlying vocoder is constructed by
+    :py:func:`torchaudio.prototype.models.hifigan_generator`. The weights are converted from the ones published
+    with the original paper :cite:`NEURIPS2020_c5d73680` under `MIT License
+    <https://github.com/jik876/hifi-gan/blob/4769534d45265d52a904b850da5a622601885777/LICENSE>`__. See links to
+    pre-trained models on `GitHub <https://github.com/jik876/hifi-gan#pretrained-model>`__.
```
mthrok, line 214: nit: Can we move the sentences, `The underlying vocoder is constructed by ...` and `The weights are converted from ...` to the end of the docstring?
They are important information, but they are meta.

mthrok (approved):

sgrigory:
@mthrok @nateanl Thanks for the review!
Before I ship: it seems that many GHA builds on CUDA fail at smoke test stage: https://github.com/pytorch/audio/actions/runs/3836270114/jobs/6530262893
Same for CircleCI smoke tests: https://app.circleci.com/pipelines/github/pytorch/audio/14250/workflows/bb1fac49-2f78-4409-9730-8a2c71b12230/jobs/1084370
It doesn't seem to be related to the changes in this PR, but just wanted to flag

nateanl (approved):

PR #2922, opened 2022-12-14
Rename resampling_method options

### LINKED ISSUE #2891: The way how the resampling method is set in Resample and F.resample is confusing
### 🐛 Describe the bug

The argument `resampling_method` can take the values `'sinc_interpolation'` or `'kaiser_window'`, as if they were different methods. [Actually both are the same sinc function-based method, just using different windows](https://github.com/pytorch/audio/blob/d8a5a11db64b1670fd0dd8f8e5e80327edbe466d/torchaudio/functional/functional.py#L1495-L1502).

Thus, the naming of the methods are confusing, especially in the case of `'kaiser_window'`, which as it doesn't contain the word **'sinc'**, but `'sinc_interpolation'` does, one could infer that `'kaiser_window'` doesn't use the sinc-function as a kernel (but it does). So, I propose to change the way this is set by:

1. Adding the argument `'window'`, which can take the values `'hann'` or `'kaiser'` to switch windows.

2. Removing the `'resampling_method'` argument, keeping it for some time for backward compatibility (throwing a warning).

3. Adding to the documentation that sinc function-based interpolation is the method used by Resample/F.resample always.

### Versions

torch==1.12.1
torchaudio==0.12.1  
(I think anything else is not relevant).

carolineechen:
resolves #2891 

Rename `resampling_method` options to more accurately describe what is happening. Previously the methods were set to `sinc_interpolation` and `kaiser_window`, which can be confusing as both options actually use sinc interpolation methodology, but differ in the window function used. As a result, rename `sinc_interpolation` to `sinc_interp_hann` and `kaiser_window` to `sinc_interp_kaiser`. Using an old option will throw a warning, and those options will be deprecated in 2 released. The numerical behavior is unchanged.

### DIFF
--- a/examples/tutorials/audio_resampling_tutorial.py
+++ b/examples/tutorials/audio_resampling_tutorial.py
@@ -240,13 +240,13 @@ plot_sweep(resampled_waveform, resample_rate, title="rolloff=0.8")
 sample_rate = 48000
 resample_rate = 32000
 
-resampled_waveform = F.resample(waveform, sample_rate, resample_rate, resampling_method="sinc_interpolation")
+resampled_waveform = F.resample(waveform, sample_rate, resample_rate, resampling_method="sinc_interp_hann")
 plot_sweep(resampled_waveform, resample_rate, title="Hann Window Default")
 
 ######################################################################
 #
 
-resampled_waveform = F.resample(waveform, sample_rate, resample_rate, resampling_method="kaiser_window")
+resampled_waveform = F.resample(waveform, sample_rate, resample_rate, resampling_method="sinc_interp_kaiser")
 plot_sweep(resampled_waveform, resample_rate, title="Kaiser Window Default")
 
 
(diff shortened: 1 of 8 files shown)


mthrok (approved):
The change looks good to me.

Regarding the deprecation process after the next release, how about keeping the backward-compatibility code and just remove the mention of old parameters from docstring?

carolineechen:
> Regarding the deprecation process after the next release, how about keeping the backward-compatibility code and just remove the mention of old parameters from docstring?

By this, do you mean keeping the warning in, but not throwing an explicit error?

mthrok:
> > Regarding the deprecation process after the next release, how about keeping the backward-compatibility code and just remove the mention of old parameters from docstring?
> 
> By this, do you mean keeping the warning in, but not throwing an explicit error?

Yes. The behavior is exactly the same, so I do not feel it necessary to break client code for changes like this.

carolineechen:
sounds good, the docstrings are already modified to only refer to the updated options, but we can update the warning in the following release.

PR #2928, opened 2022-12-17
Add filter_waveform

mthrok:
This commit adds "filter_waveform" prototype function.

This function can apply non-stationary filters across the time.
It also performs cropping at the end to compensate the delay introduced by filtering.
The figure bellow illustrates this.

See [subtractive_synthesis_tutorial](https://output.circle-artifacts.com/output/job/5233fda9-dadb-4710-9389-7e8ac20a062f/artifacts/0/docs/tutorials/subtractive_synthesis_tutorial.html) for example usages.

figure

### DIFF
--- a/test/torchaudio_unittest/prototype/functional/functional_test_impl.py
+++ b/test/torchaudio_unittest/prototype/functional/functional_test_impl.py
@@ -549,6 +549,46 @@ class FunctionalTestImpl(TestBaseMixin):
 
         assert filtered.shape == waveform.shape
 
+    @nested_params([1, 3, 5], [3, 5, 7, 4, 6, 8])
+    def test_filter_waveform_delta(self, num_filters, kernel_size):
+        """Applying delta kernel preserves the origianl waveform"""
+        waveform = torch.arange(-10, 10, dtype=self.dtype, device=self.device)
+        kernel = torch.zeros((num_filters, kernel_size), dtype=self.dtype, device=self.device)
+        kernel[:, kernel_size//2] = 1
+
+        result = F.filter_waveform(waveform, kernel)
+        self.assertEqual(waveform, result)
+
+    def test_filter_waveform_same(self, kernel_size=5):
+        """Applying the same filter returns the original waveform"""
+        waveform = torch.arange(-10, 10, dtype=self.dtype, device=self.device)
+        kernel = torch.randn((1, kernel_size), dtype=self.dtype, device=self.device)
+        kernels = torch.cat([kernel] * 3)
+
+        out1 = F.filter_waveform(waveform, kernel)
+        out2 = F.filter_waveform(waveform, kernels)
+        self.assertEqual(out1, out2)
+
+    def test_filter_waveform_diff(self):
+        """Filters are applied from the first to the last"""
+        kernel_size = 3
+        waveform = torch.arange(-10, 10, dtype=self.dtype, device=self.device)
+        kernels = torch.randn((2, kernel_size), dtype=self.dtype, device=self.device)
+
+        # use both filters.
+        mix = F.filter_waveform(waveform, kernels)
+        # use only one of them
+        ref1 = F.filter_waveform(waveform[:10], kernels[0:1])
+        ref2 = F.filter_waveform(waveform[10:], kernels[1:2])
+
+        print("mix:", mix)
+        print("ref1:", ref1)
+        print("ref2:", ref2)
+        # The first filter is effective in the first half
+        self.assertEqual(mix[:10], ref1[:10])
+        # The second filter is effective in the second half
+        self.assertEqual(mix[-9:], ref2[-9:])
+        # the middle portion is where the two filters affect
 
 class Functional64OnlyTestImpl(TestBaseMixin):
     @nested_params(
--- a/test/torchaudio_unittest/prototype/functional/functional_test_impl.py
+++ b/test/torchaudio_unittest/prototype/functional/functional_test_impl.py
@@ -486,6 +486,69 @@ class FunctionalTestImpl(TestBaseMixin):
 
         self.assertEqual(hyp, ref)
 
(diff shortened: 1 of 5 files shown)


torchaudio/prototype/functional/_dsp.py
```diff
@@ -306,3 +308,101 @@ def frequency_impulse_response(magnitudes):
...
+
+def filter_waveform(
+        waveform: torch.Tensor,
+        kernels: torch.Tensor,
+        delay_compensation: int = -1):
+    """Applies filters along time axis of the given waveform.
+
+    This function applies the given filters along time axis in the following manner:
+
+    1. Split the given waveform into chunks of the number of the given filters.
```
carolineechen, line 332: A little bit confusing here if chunk number or chunk length is equal to the number of filters.

```suggestion
    1. Split the given waveform into chunks. The number of chunks is equal to the number of given filters.
```

torchaudio/prototype/functional/_dsp.py
```diff
@@ -306,3 +308,101 @@ def frequency_impulse_response(magnitudes):
...
+       i.e. the same filtering is applied across the time axis.
+
+    Args:
+        waveform (Tensor): Shape `(..., time)`.
+        kernels (Tensor): Impulse responses.
+            Valid inputs are 2D tensor with shape `(num_filters, filter_length)` or
+            `N+1` D tensor with shape `(..., num_filters, filter_length)`, where N is
+            the dimension of waveform.
+
+            In case of 2D input, the same set of filters are used across channels and batches.
```
carolineechen, line 354:
```suggestion
            In case of 2D input, the same set of filters is used across channels and batches.
```

torchaudio/prototype/functional/_dsp.py
```diff
@@ -306,3 +308,101 @@ def frequency_impulse_response(magnitudes):
...
+            Otherwise, different set of filters are applied. In this case, the shape of
```
carolineechen, line 355:
```suggestion
            Otherwise, different sets of filters are applied. In this case, the shape of
```

torchaudio/prototype/functional/_dsp.py
```diff
@@ -306,3 +308,101 @@ def frequency_impulse_response(magnitudes):
...
+            the first `N-1` dimensions of filters must match (or broadcastable to) that of waveform.
```
carolineechen, line 356:
```suggestion
            the first `N-1` dimensions of filters must match (or be broadcastable to) that of waveform.
```

torchaudio/prototype/functional/_dsp.py
```diff
@@ -306,3 +308,101 @@ def frequency_impulse_response(magnitudes):
...
+    """
+    if kernels.ndim not in [2, waveform.ndim + 1]:
+        raise ValueError(
+            "`kernels` must be 2 or N+1 dimension where "
+            f"N is the dimension of waveform. Found: {kernels.ndim} (N={waveform.ndim})")
+
+    num_filters, filter_size = kernels.shape[-2:]
+    num_frames = waveform.size(-1)
+
+    if delay_compensation is not None and delay_compensation > filter_size:
```
carolineechen, line 376: With the current setup, `delay_compensation` is never `None`. I think it would be better to have it be `Optional[int] = None` default rather than `-1` (because other negative numbers don't really make sense) and then assert that `0 <= delay_compensation < filter_size` if not `None` (docstring will need to be modified a bit too).

carolineechen (commented):
left a couple comments, there are also a couple of lint errors

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -486,6 +486,109 @@ def test_freq_ir_reference(self, shape):
...
+
+        # use both filters.
+        mix = F.filter_waveform(waveform, kernels)
+        # use only one of them
+        ref1 = F.filter_waveform(waveform[:10], kernels[0:1])
+        ref2 = F.filter_waveform(waveform[10:], kernels[1:2])
+
+        print("mix:", mix)
+        print("ref1:", ref1)
+        print("ref2:", ref2)
```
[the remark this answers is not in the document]
mthrok: Why? It is useful to have the print statements when test fails.

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -486,6 +486,109 @@ def test_freq_ir_reference(self, shape):
...
+        # The first filter is effective in the first half
+        self.assertEqual(mix[:10], ref1[:10])
+        # The second filter is effective in the second half
+        self.assertEqual(mix[-9:], ref2[-9:])
+        # the middle portion is where the two filters affect
```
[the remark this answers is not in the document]
mthrok: no, the middle portion is mixture of two so we can't assert.

torchaudio/prototype/functional/_dsp.py
mthrok: oh no this is a leftover from the initial implementation. None is not a valid choice. Let me update it.

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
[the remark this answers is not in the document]
carolineechen: I see, I thought it was accidentally leftover from debugging. Can we print only if it's failing to avoid extra print clutter and make it more clear the function that is being tested in the print? Otherwise if we start adding generic print/debug statements in tests it can lead to extra clutter.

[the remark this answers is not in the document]
mthrok: We use `pytest` as runner, and it by default captures and hides the stdout/stderr. So it's shown only when told so or failed.

carolineechen (approved):

PR #2929, opened 2022-12-19
Extract libsox integration from libtorchaudio

mthrok:
This commit makes the following changes to the C++ library organization
- Move sox-related feature implementations from `libtorchaudio` to `libtorchaudio_sox`.
- Remove C++ implementation of `is_sox_available` and `is_ffmpeg_available` as it is now sufficient to check the existence of `libtorchaudio_sox` and `libtorchaudio_ffmpeg` to check the availability. This makes `libtorchaudio_sox` and `libtorchaudio_ffmpeg` independent from `libtorchaudio`.
- Move PyBind11-based bindings (`_torchaudio_sox`, `_torchaudio_ffmpeg`) into `torchaudio.lib` so that the built library structure is less cluttered. 

Background:
Originally, when the `libsox` was the only C++ extension and `libtorchaudio` was supposed to contain all the C++ code.
The things are different now. We have a bunch of C++ extensions and we need to make the code/build structure more modular.

The new `libtorchaudio_sox` contains the implementations and `_torchaudio_sox` contains the PyBin11-based bindings.

### DIFF
--- a/tools/setup_helpers/extension.py
+++ b/tools/setup_helpers/extension.py
@@ -47,21 +47,28 @@ _TORCH_CUDA_ARCH_LIST = os.environ.get("TORCH_CUDA_ARCH_LIST", None)
 def get_ext_modules():
     modules = [
         Extension(name="torchaudio.lib.libtorchaudio", sources=[]),
-        Extension(name="torchaudio._torchaudio", sources=[]),
+        Extension(name="torchaudio.lib._torchaudio", sources=[]),
     ]
+    if _BUILD_SOX:
+        modules.extend(
+            [
+                Extension(name="torchaudio.lib.libtorchaudio_sox", sources=[]),
+                Extension(name="torchaudio.lib._torchaudio_sox", sources=[]),
+            ]
+        )
     if _BUILD_CTC_DECODER:
         modules.extend(
             [
                 Extension(name="torchaudio.lib.libflashlight-text", sources=[]),
-                Extension(name="torchaudio.flashlight_lib_text_decoder", sources=[]),
-                Extension(name="torchaudio.flashlight_lib_text_dictionary", sources=[]),
+                Extension(name="torchaudio.lib.flashlight_lib_text_decoder", sources=[]),
+                Extension(name="torchaudio.lib.flashlight_lib_text_dictionary", sources=[]),
             ]
         )
     if _USE_FFMPEG:
         modules.extend(
             [
                 Extension(name="torchaudio.lib.libtorchaudio_ffmpeg", sources=[]),
-                Extension(name="torchaudio._torchaudio_ffmpeg", sources=[]),
+                Extension(name="torchaudio.lib._torchaudio_ffmpeg", sources=[]),
             ]
         )
     return modules
@@ -84,10 +91,16 @@ class CMakeBuild(build_ext):
         # However, the following `cmake` command will build all of them at the same time,
         # so, we do not need to perform `cmake` twice.
         # Therefore we call `cmake` only for `torchaudio._torchaudio`.
-        if ext.name != "torchaudio._torchaudio":
+        if ext.name != "torchaudio.lib.libtorchaudio":
             return
 
-        extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
+        # Note:
+        # the last part "lib" does not really matter. We want to get the full path of
+        # the root build directory. Passing "torchaudio" will be interpreted as
+        # `torchaudio.[so|dylib|pyd]`, so we need something `torchaudio.foo`, that is
+        # interpreted as `torchaudio/foo.so` then use dirname to get the `torchaudio`
+        # directory.
+        extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath("torchaudio.lib")))
 
         # required for auto-detection of auxiliary "native" libs
         if not extdir.endswith(os.path.sep):
--- a/torchaudio/_extension.py
+++ b/torchaudio/_extension.py
@@ -1,12 +1,15 @@
+import logging
 import os
 import sys
-import warnings
 from pathlib import Path
 
 import torch
-from torchaudio._internal import module_utils as _mod_utils  # noqa: F401
+
+from torchaudio._internal.module_utils import is_module_available
 
 _LIB_DIR = Path(__file__).parent / "lib"
+
+_LG = logging.getLogger(__name__)
 
 
 def _get_lib_path(lib: str):
@@ -62,7 +65,7 @@ def _init_ffmpeg():
     if _FFMPEG_INITIALIZED:
         return
 
-    if not torch.ops.torchaudio.is_ffmpeg_available():
+    if not is_module_available("torchaudio.lib._torchaudio_ffmpeg"):
         raise RuntimeError(
             "torchaudio is not compiled with FFmpeg integration. Please set USE_FFMPEG=1 when compiling torchaudio."
         )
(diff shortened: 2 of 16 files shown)


tools/setup_helpers/extension.py
```diff
@@ -84,10 +90,10 @@ def build_extension(self, ext):
         # However, the following `cmake` command will build all of them at the same time,
         # so, we do not need to perform `cmake` twice.
         # Therefore we call `cmake` only for `torchaudio._torchaudio`.
-        if ext.name != "torchaudio._torchaudio":
+        if ext.name != "torchaudio.lib.libtorchaudio":
             return
 
-        extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
+        extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath("torchaudio.foo")))
```
hwangjeff, line 96: could be helpful to add a comment here that explains why we use "torchaudio.foo"

torchaudio/_extension.py
```diff
@@ -103,12 +97,15 @@ def _init_extension():
                     pass
 
     _load_lib("libtorchaudio")
-    # This import is for initializing the methods registered via PyBind11
-    # This has to happen after the base library is loaded
-    from torchaudio import _torchaudio  # noqa
 
     # Because this part is executed as part of `import torchaudio`, we ignore the
     # initialization failure.
+    _load_lib("libtorchaudio_sox")
```
hwangjeff, line 103: since including sox is not compulsory, should this be moved to the try-catch block below so that failing to load `libtorchaudio_sox` doesn't prevent importing `torchaudio` altogether?

torchaudio/_internal/module_utils.py
```diff
@@ -67,7 +67,10 @@ def wrapped(*args, **kwargs):
 
 
 def is_kaldi_available():
-    return is_module_available("torchaudio._torchaudio") and torch.ops.torchaudio.is_kaldi_available()
+    try:
+        return torch.ops.torchaudio.is_kaldi_available()
+    except Exception:
+        return False
```
hwangjeff, line 73: why treat kaldi differently from sox below? doesn't the requirement of `torchaudio.lib.libtorchaudio` existing hold true here too?

tools/setup_helpers/extension.py
mthrok: Could be anything as the last path part is being removed by `dirname`.
However, can we put all the C++ extension in `lib`?
Currently extension modules like `torchaudio._torchaudio` is put directly under the top module.
We will have three of these, so I am thinking that putting all of them in `torchaudio.lib` make it cleaner.

torchaudio/_internal/module_utils.py
mthrok: The torchaudio codebase is guarded against the lack of C++ extension.
For example, you can clone the repo and do `import torchaudio` on the source directory without building.

So we make these helper function works even when libtorchaudio does not exist.

hwangjeff: makes sense. then, what's the reasoning for not using a try-catch in `is_sox_available`? seems like the logic in `is_sox_available` should apply here, e.g. we check that `libtorchaudio` exists via `is_module_available`, and if it does, we run `torch.ops.torchaudio.is_kaldi_available`

tools/setup_helpers/extension.py
hwangjeff: sounds reasonable to me

torchaudio/_internal/module_utils.py
mthrok: So the difference is the existence of PyBind-based wrapper. Kaldi-feature is implemented in libtorchaudio, which is a C++ library without a link to libPython. So Python does not see it as a Python module, and `is_module_available` does not work. We can check the existence of the `libtorchaudio` file as an added guard statement, but the existence of the file does not necessarily mean that `torch.ops.torchaudio.is_kaldi_available` is available. (also, the extension is different on OS, so it's a bit annoying to do the file existence check) So it's easier to just do `try ~ except` here.

For the case of sox, the implementation will be found in `libtorchaudio_sox` library, and there will be always `torchaudio._torchaudio_sox` wrapper, which is what Python can recognize as a valid Python module (`is_module_available` works). So When `is_module_available("torchaudio.lib._torchaudio_sox")` (changing the target extension name/location to the one I am actually planning to use) succeed, it means that sox-related features are available and `torch.ops.torchaudio.is_sox_available()` exist.

torchaudio/_extension.py
mthrok: When we move to dynamic linking of libsox, that will be absolutely necessary. Right now, it is not necessary but I can add it. Let me update the code.

torchaudio/sox_effects/sox_effects.py
```diff
@@ -274,7 +274,7 @@ def apply_effects_file(
     """
     if not torch.jit.is_scripting():
         if hasattr(path, "read"):
-            ret = torchaudio._torchaudio.apply_effects_fileobj(path, effects, normalize, channels_first, format)
+            ret = torchaudio._torchaudio_sox.apply_effects_fileobj(path, effects, normalize, channels_first, format)
```
hwangjeff, line 277:
```suggestion
            ret = torchaudio.lib._torchaudio_sox.apply_effects_fileobj(path, effects, normalize, channels_first, format)
```

torchaudio/csrc/pybind/pybind.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <torchaudio/csrc/utils.h>
+
+namespace torchaudio {
+namespace {
+
+
+// Note
+// These functions are not intended for a real usecase.
+// They are accessible via TorchBind.
+// It is benefitial to have _torchaudio that is linked to libtorchaudio,
```
hwangjeff, line 11:
```suggestion
// It is beneficial to have _torchaudio that is linked to libtorchaudio,
```

torchaudio/csrc/pybind/pybind.cpp
```diff
@@ -0,0 +1,23 @@
...
+// when torchaudio is deployed with PEX format, where the library location
+// is not in torchaudio/lib. But somewhere in LD_LIBRARY_PATH.
+// In this case, attempt to import _torchaudio will automatically resolves
+// libtorchauio, if _torchaudio is linked to libtorchaudio.
+
+PYBIND11_MODULE(_torchaudio, m) {
+  m.def("is_kaldi_available", &is_kaldi_available, "");
+  m.def("cuda_version", &cuda_version, "");
+}
+
```
hwangjeff, line 21: to clarify: this exists solely for there to be an extension `_torchaudio` that's linked to `libtorchaudio`?

torchaudio/csrc/pybind/pybind.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+// Note
+// These functions are not intended for a real usecase.
+// They are accessible via TorchBind.
+// It is benefitial to have _torchaudio that is linked to libtorchaudio,
+// when torchaudio is deployed with PEX format, where the library location
+// is not in torchaudio/lib. But somewhere in LD_LIBRARY_PATH.
+// In this case, attempt to import _torchaudio will automatically resolves
+// libtorchauio, if _torchaudio is linked to libtorchaudio.
```
hwangjeff, line 15:
```suggestion
// libtorchaudio, if _torchaudio is linked to libtorchaudio.
```

hwangjeff (approved):

torchaudio/csrc/pybind/pybind.cpp
mthrok: yes. we could move the FFI registeration from torchbind to here as well, if we want to have only one registration.

PR #2937, opened 2022-12-24
Fix confusing wording in the documentation

TruscaPetre:
I have spent 2 hours because I read the documentation wrongly and trying to figure out why the I couldn't read the data from the dataset. The initial phrase was very confusing.

### DIFF
--- a/torchaudio/datasets/librimix.py
+++ b/torchaudio/datasets/librimix.py
@@ -19,8 +19,8 @@ class LibriMix(Dataset):
     r"""*LibriMix* :cite:`cosentino2020librimix` dataset.
 
     Args:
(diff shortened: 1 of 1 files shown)


nateanl (approved):
Thanks for updating the documentation. The root path follows the pattern of [LIBRISPEECH](https://pytorch.org/audio/stable/generated/torchaudio.datasets.LIBRISPEECH.html#torchaudio.datasets.LIBRISPEECH) and other datasets where the real folders are under `root` directory. A better way could be throw a warning message if `Libri2Mix` or `Libri3Mix` is not found under `root`. I will add it in a separate PR.

PR #2941, opened 2022-12-27
Experimenting with Pytorch 2.0 compile() optimizations on wav2vec2 models

kalakris:
This pull request contains code that I added to the librispeech_ctc_decoder inference.py example to evaluate if pytorch 2.0 compile() optimizations can help this model. To summarize, my findings are:

- torch.inference_mode() had to be replaced by torch.no_grad()
- I couldn't use compile() with dynamic tensor sizes (yet).
- Passing the 'lengths' parameter to the wav2vec2 model also breaks the compilation.
- Overall, there's a ~10% performance improvement using CUDA, and up to 38% on CPU, but this is currently achievable only by padding all samples to make sure tensor sizes are equal, and this hurts performance.


mthrok:
Thanks @kalakris !

PR #2942, opened 2022-12-28
Use ring-buffer in StreamReader ChunkedVideoBuffer

mthrok:
Currently, all the Buffer implementation uses `std::deque<torch::Tensor>` for storing decoded frames.

In case of video stream, the decoded frames are converted to tensors newly allocated each time, then 
the resulting tensors are stored in the `std::dequeu`. They are concatenated at the time they are popped.

This commit replaces the buffer implementation for ChunkedVideoBuffer with ring buffer implementation.
In the new implementation the buffer is pre-allocated. Each time a decoded frame is given, it is written to
the buffer. No memory allocation happens. At the time of popping, the slice of the buffer is created and 
then cloned. Memory allocations still happen at the clone time, but this reduces the number of memory copies.

As a result, the wall time performance is 30~40% improved.


joecummings:
Is there any way to document some baseline numbers here? I assume you have them in another notebook or something.

mthrok:
> Is there any way to document some baseline numbers here? I assume you have them in another notebook or something.

Will do once the CL is finalized.

mthrok:
Closing due to conflict

Issue #2855, opened 2022-11-16
Limit StreamReader to use a specifed number of cores


DanTremonti:
### Impact of number of cores on StreamReader's performance:

Unlike `torchaudio.load`, the time taken by `StreamReader` to load an audio files (`.opus` files specifically) varies a lot from machine-to-machine. The inference is that, this variation (in loading time) is related to the number of cores in the machine as well as the length of audio samples.

Given the high expense of CPU usage by StreamReader (varies with machine) against a max CPU usage of 100% by `torchaudio.load`, it is expected to perform better than the latter. But this is not always the case.

Hence, the ability to limit/set the number of CPU cores to be used by `StreamReader` is required.

### How to reproduce

[Notebook to reproduce this behaviour](https://colab.research.google.com/drive/1IH13cVRkfeUnknrLzbVgxdP7y0c2VESd?usp=sharing)
Iterate the provided function `audio_loader_sr` for the given dataset and note the time taken.

### What was tested

A dataset of `.opus` files was loaded by iterating the below functions - `audio_loader` and `audio_loader_sr`. The average duration of audio samples in the provided `valentini` dataset is ~`2.95s`. For such short audio files, distributing the loading across all available cores seems to create a dip in the performance on machines with many cores.

### Results

The following benchmarking was done in different machines with different number of cores, to load 1000 `.opus` files

num_cores | Time taken
:--------------:|:--------------:
12   |  00:15s
24   |  00:12s
256 |  01:11s

### Some other benchmarking references
- [Ref1](https://streaminglearningcenter.com/blogs/ffmpeg-command-threads-how-it-affects-quality-and-performance.html)
- [Ref2](https://superuser.com/a/1196163/1747713)

### Solution

The underlying ffmpeg library has an option to set the number of cores to be used to load the audio, using `-threads` arg. The auto detection of the `-threads` value may not be preferable for all scenarios (say, we want to obtain a constant loading time in a certain machine at different times) and also does not always seem to do a good job in terms of loading times. If we could pass this argument value using `StreamReader`, we would be able to limit the number of cores used.

mthrok:
Hi

Thanks for the detailed report. Could you try passing `decoder_option={"threads": "NUM_THREADS"}` to `add_XXX_stream` and see if that makes difference?

[...]

PR #2949, opened 2023-01-03
Add multithread options to docstring and default to single thread

mthrok:
One can pass "threads" and "thread_type" to `decoder_option` of StreamReaader to change the multithreading configuration.

These affects the timing that decoder starts emitting the decoded frames. i.e. how many packets at minimum have to be processed before the first frame is decoded.

Overall, multithreading in decoder does not improve the performance.
(One possible reason is because the design of StreamReader, "decode few frames then fetch them", does not suited to saturate the decoder with incoming packets.)

num_threads=1 seems to exhibit overall good performance/resource balance.

Decoding time over threading
Decoding time over threading (1)
(Tested on 320x240 25 FPPS, YUV420P videos generated with `ffmpeg -f lavfi -t "${duration}" -i testsrc -pix_fmt "yuv420p"`)

For this reason, we default to single thread execution in StreamReader. 
closes https://github.com/pytorch/audio/issues/2855

Follow-up: Apply similar change to encoder option in StreamWriter.

### DIFF
--- a/torchaudio/io/_stream_reader.py
+++ b/torchaudio/io/_stream_reader.py
@@ -239,12 +239,26 @@ _decoder = """The name of the decoder to be used.
                 Default: ``None``."""
 
 _decoder_option = """Options passed to decoder.
-                Mapping from str to str.
+                Mapping from str to str. (Default: ``None``)
 
                 To list decoder options for a decoder, you can use
                 `ffmpeg -h decoder=<DECODER>` command.
 
-                Default: ``None``."""
+                In addition to decoder-specific options, you can also pass options related
+                to multithreading. They are effective only if the decoder support them.
+                If neither of them are provided, StreamReader defaults to single thread.
+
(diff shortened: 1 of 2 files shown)


torchaudio/io/_stream_reader.py
```diff
@@ -239,12 +239,26 @@ def decorator(obj):
...
 
                 To list decoder options for a decoder, you can use
                 `ffmpeg -h decoder=<DECODER>` command.
 
-                Default: ``None``."""
+                In addition to decoder-specific options, you can also pass options related
+                to multithreading. They are effective only if the decoder support them.
+                If neither of them are provided, StreamReader defaults to single thread.
+
+                 - ``"thread"``: The number of threads (in str) or the value ``"0"``
```
carolineechen, line 251: should this be `"threads"`?

carolineechen (approved):

carolineechen (approved):

PR #2952, opened 2023-01-04
Guard invocation of cuda_version

hwangjeff:
Currently, importing TorchAudio triggers a check of the CUDA version it was compiled with, which in turn calls `torch.ops.torchaudio.cuda_version()`. This function is available only if `libtorchaudio` is available; developers, however, may want to import TorchAudio regardless of its availability. To allow for such usage, this PR adds code that bypasses the check if `libtorchaudio` is not available.

### DIFF
--- a/torchaudio/_extension.py
+++ b/torchaudio/_extension.py
@@ -125,6 +125,8 @@ def _init_extension():
 
 
 def _check_cuda_version():
+    if not _get_lib_path("libtorchaudio").exists():
+        return None
     version = torch.ops.torchaudio.cuda_version()
     if version is not None and torch.version.cuda is not None:
         version_str = str(version)


torchaudio/_extension.py
```diff
@@ -125,7 +125,12 @@ def _init_extension():
 
 
 def _check_cuda_version():
-    version = torch.ops.torchaudio.cuda_version()
+    try:
+        version = torch.ops.torchaudio.cuda_version()
+    except Exception:
+        _LG.debug("Could not determine CUDA version of TorchAudio build; skipping CUDA version check.", exc_info=True)
+        return None
```
mthrok, line 132: Looking at [the failure like this](https://github.com/pytorch/audio/pull/2921#issuecomment-1370994739), it would be better to specify the condition we allow this to fail. Can we do

```python
if not _get_lib_path("libtorchaudio").exists():
    return None
version = torch.ops.torchaudio.cuda_version()
```

mthrok (changes requested):
Let's be more specific

PR #2959, opened 2023-01-05
Add XLS-R models

nateanl:
XLSR (cross-lingual speech representation) are a set of cross-lingual self-supervised learning models for generating cross-lingual speech representation. It was first proposed in https://arxiv.org/pdf/2006.13979.pdf which is trained on 53 languages (so-called XLSR-53). This PR supports more XLS-R models from https://arxiv.org/pdf/2111.09296.pdf that have more parameters (300M, 1B, 2B) and are trained on 128 languages.


nateanl:
fyi I removed ["checkpoint_activations"](https://github.com/facebookresearch/fairseq/blob/main/fairseq/models/wav2vec/wav2vec2.py#L256) from fairseq's configuration json files as they are added after the fairseq commit in the unit test, and it doesn't affect the model inference.

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -208,6 +208,13 @@ def skipIfNoModule(module, display_name=None):
     reason="CUDA is not available.",
     key="NO_CUDA",
 )
+# Skip test if model size is too large.
+# TODO: detect the real model size and compare with CUDA memory.
+skipIfCudaSmallMemory = _skipIf(
+    torch.cuda.is_available(),
+    reason="CUDA does not have enough memory.",
+    key="CUDA_SMALL_MEMORY",
+)
```
mthrok, line 217:
```suggestion
# Skip test if CUDA memory is not enough
# TODO: detect the real CUDA memory size and allow call site to configure how much the test needs
skipIfCudaSmallMemory = _skipIf(
    "CI" in os.environ and torch.cuda.is_available(),  # temporary
    reason="CUDA does not have enough memory.",
    key="CUDA_SMALL_MEMORY",
)
```

mthrok (approved):

PR #2965, opened 2023-01-06
Increase timeout

mthrok:
Windows GPU unit tests jobs have been time-out at installation. Increasing the timeout.


carolineechen (commented):
looks like it's still timing out, should it be increased even more?

PR #2975, opened 2023-01-12
Make StreamReader return PTS

mthrok:
This commit makes `StreamReader` report PTS (presentation time stamp) of the returned chunk as well.

Example

```python

s = StreamReader(...)
s.add_video_stream(...)
for (video_chunk, ) in s.stream():
    # video_chunk is Torch tensor type but has extra attribute of PTS
    print(video_chunk.pts)  # reports the PTS of the first frame of the video chunk.
```

For the backward compatibility, we introduce a `_ChunkTensor`, that is a composition
of Tensor and metadata, but works like a normal tensor in PyTorch operations.

The implementation of `_ChunkTensor` is based on [TrivialTensorViaComposition](https://github.com/albanD/subclass_zoo/blob/0eeb1d68fb59879029c610bc407f2997ae43ba0a/trivial_tensors.py#L83).

It was also suggested to attach metadata directly to Tensor object,
but the possibility to have the collision on torchaudio's metadata and new attributes introduced in
PyTorch cannot be ignored, so we use Tensor subclass implementation.

If any unexpected issue arise from metadata attribute name collision, client code can
fetch the bare Tensor and continue.

### DIFF
--- a/torchaudio/io/_stream_reader.py
+++ b/torchaudio/io/_stream_reader.py
@@ -5,16 +5,21 @@ from typing import BinaryIO, Dict, Iterator, Optional, Tuple, Union
 
 import torch
 import torchaudio
+from torch.utils._pytree import tree_map
+
+__all__ = [
+    "StreamReader",
+]
 
 
 @dataclass
-class StreamReaderSourceStream:
+class SourceStream:
     """The metadata of a source stream, returned by :meth:`~torchaudio.io.StreamReader.get_src_stream_info`.
 
     This class is used when representing streams of media type other than `audio` or `video`.
 
-    When source stream is `audio` or `video` type, :class:`StreamReaderSourceAudioStream` and
-    :class:`StreamReaderSourceVideoStream`, which reports additional media-specific attributes,
+    When source stream is `audio` or `video` type, :class:`SourceAudioStream` and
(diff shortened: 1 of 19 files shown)


torchaudio/io/_stream_reader.py
```diff
@@ -756,7 +778,13 @@ def pop_chunks(self) -> Tuple[Optional[torch.Tensor]]:
...
                 If a buffer does not contain any frame, then `None` is returned instead.
         """
-        return self._be.pop_chunks()
+        ret = []
+        for chunk in self._be.pop_chunks():
+            if chunk is None:
+                ret.append(None)
+            else:
+                ret.append(_ChunkTensor(chunk[0], chunk[1]))
+        return ret
```
hwangjeff, line 787: should update the return type annotation and docstring

hwangjeff (approved):

torchaudio/io/_stream_reader.py
mthrok: Since documenting the class means we need to make the class public. I made couple of changes. (making it inner class of StreamReader, decorated dataclass). Can you take a look again?

albanD (commented):
Sounds good! Sorry for the late review.

PR #2977, opened 2023-01-12
Make lengths optional for additive noise operators

hwangjeff:
For greater flexibility, this PR makes argument `lengths` optional for `add_noise` and `AddNoise`.

### DIFF
--- a/torchaudio/prototype/functional/functional.py
+++ b/torchaudio/prototype/functional/functional.py
@@ -1,7 +1,7 @@
 import math
 import warnings
 
-from typing import Tuple
+from typing import Optional, Tuple
 
 import torch
 from torchaudio.functional import lfilter, resample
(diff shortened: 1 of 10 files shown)


torchaudio/prototype/functional/functional.py
```diff
@@ -134,7 +134,9 @@ def convolve(x: torch.Tensor, y: torch.Tensor, mode: str = "full") -> torch.Tens
     return _apply_convolve_mode(result, x_size, y_size, mode)
 
 
-def add_noise(waveform: torch.Tensor, noise: torch.Tensor, lengths: torch.Tensor, snr: torch.Tensor) -> torch.Tensor:
+def add_noise(
+    waveform: torch.Tensor, noise: torch.Tensor, snr: torch.Tensor, lengths: Optional[torch.Tensor] = None
```
nateanl, line 138: shoud it be `length` as `waveform` and `noise` are both singular?

nateanl (approved):

torchaudio/prototype/functional/functional.py
hwangjeff: i'd stick with `lengths` because, whereas a waveform and noise can comprise multiple components, "length" represents only one quantity

this is also consistent with other library components, e.g. rnn-t loss, conformer, emformer

nateanl: makes sense. let's also use `lengths` in the future functions and modules then.

PR #2984, opened 2023-01-15
Refactor chunked buffer implementation

mthrok:
So that the number of Tensor frames stored in buffers is always a multiple of frames_per_chunk.

This makes it easy to store PTS values in aligned manner.


test/torchaudio_unittest/io/stream_reader_test.py
```diff
@@ -553,6 +553,28 @@ def test_unchunked_stream(self):
         assert video.shape == torch.Size([390, 3, 270, 480])
         assert audio.shape == torch.Size([208896, 2])
 
+    @parameterized.expand([(1,), (3,), (5,), (10,)])
+    def test_frames_per_chunk(self, fbc):
```
nateanl, line 557: should it be fpc?
```suggestion
    def test_frames_per_chunk(self, fpc):
```

nateanl (approved):

PR #2987, opened 2023-01-16
Fixes examples/source_separation for WSJ0_2mix dataset

fakufaku:
The `examples/source_separation` scripts use inconsistent keyword to indicate the WSJ0_2mix dataset. This PR does the following.

1. Use `wsj0mix` consistently as keyword indicating the WSJ0_2mix dataset
2. Corrects `args.data_dir` to `args.root_dir` in eval.py
3. Modify the parameters of `pytorch_lightning.Trainer` according to latest version (use `accelerator="gpu"` and `devices=args.num_devices`, instead of just `gpus=args.num_devices`)

### DIFF
--- a/examples/source_separation/eval.py
+++ b/examples/source_separation/eval.py
@@ -31,7 +31,7 @@ def _eval(model, data_loader, device):
 
 def cli_main():
     parser = ArgumentParser()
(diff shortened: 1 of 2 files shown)


nateanl (approved):
Thanks @fakufaku for contributing the code. The fix LGTM.

PR #2990, opened 2023-01-16
Fix buffer flushing mechanism

mthrok:
Summary:
When buffered data are cleared from ChunkedBuffer,
the `num_buffered_frames` variable was not updated.

This commit fixes that.

Reviewed By: xiaohui-zhang

Differential Revision: D42538519


mthrok:
Merged as [51731bf](https://github.com/pytorch/audio/commit/51731bf956a8c830e7f3c7a87e148ec264af90f2)

PR #2890, opened 2022-12-05
TCPGen in Conformer RNN-T

BriansIDP:
This pull request contains the implementation of the tree-constrained pointer generator (TCPGen) for contextual biasing. An example for Librispeech can be found in audio/examples/asr/librispeech_biasing.

PR #2991, opened 2023-01-17
Add TCPGen in Conformer RNN-T

mthrok:
This is a cleaned up version of https://github.com/pytorch/audio/pull/2890

> This pull request contains the implementation of
> the tree-constrained pointer generator (TCPGen) for contextual biasing.
> An example for Librispeech can be found in
> audio/examples/asr/librispeech_biasing.


torchaudio/datasets/librispeech_biasing.py
```diff
@@ -0,0 +1,134 @@
+import os
+from pathlib import Path
+from typing import Tuple, Union
+
+import torchaudio
+from torch import Tensor
+from torch.hub import download_url_to_file
+from torch.utils.data import Dataset
+from torchaudio.datasets.utils import extract_archive
```
nateanl, line 9: `extract_archive` is replaced with `_extract_tar` and `_extract_zip` in #2927. Here we can use `_extract_tar`

xiaohui-zhang (commented):
in general code structure and design look very good. We can land after another round of review by @hwangjeff  or @nateanl (who will run the recipe for validation as well). Also before merging, @BriansIDP plz update README to reflect the expected WERs (it's ok to be preliminary). 

Though I don't think we need to address now, I want to point out that, there're many recent Conformer RNN-T recipe improvements made by @hwangjeff which hasn't been synced into the Emformer RNN-T recipe. This may motivate us to make an even more "shared" modular design across ASR recipes in future. Otherwise as the Emformer RNN-T recipe with biasing support and the Conformer RNN-T recipe become more and more popular independently, they may branch even more and their consistency may become harder to maintain.

examples/asr/librispeech_biasing/train_spm.py
```diff
@@ -0,0 +1,87 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
BriansIDP: The key difference is that this file allows suffix-based wordpieces. There are also minor differences such as unigram wordpiece and only for clean-100, etc. I wonder if I can keep this.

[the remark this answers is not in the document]
xiaohui-zhang: oh sure. plz keep it and mention the difference in the header.

examples/asr/librispeech_biasing/train.sh
```diff
@@ -0,0 +1,13 @@
...
+    --exp-dir $expdir \
+    --librispeech-path /home/gs534/rds/hpc-work/data/Librispeech/ \
+    --global-stats-path ./global_stats_100.json \
+    --sp-model-path ./spm_unigram_600_100suffix.model \
+    --biasing true \
+    --biasinglist ./blists/rareword_f15.txt \
+    --droprate 0.0 \
+    --maxsize 500 \
+    --epochs 90 \
+    --resume experiments/librispeech_clean100_suffix600_tcpgen500_sche30_nodrop/checkpoints/epoch=45-step=89838.ckpt \
```
nateanl, line 13: This file can be removed IMO. As @xiaohui-zhang suggested, we can provide a template training script in README with slurm.

torchaudio/datasets/librispeech_biasing.py
```diff
@@ -0,0 +1,134 @@
...
+            if fileid_audio == fileid_text:
+                break
+        else:
+            # Translation not found
+            raise FileNotFoundError(f"Translation not found for {fileid_audio}")
+
+    return (waveform, sample_rate, transcript, int(speaker_id), int(chapter_id), int(utterance_id), wordlist)
+
+
+class LIBRISPEECH_BIASING(Dataset):
```
nateanl, line 76: It's better to use camel case for the naming, for example, `LibriSpeechBiasing`.

torchaudio/datasets/librispeech_biasing.py
```diff
@@ -0,0 +1,134 @@
...
+    """Create a Dataset for *LibriSpeech* [:footcite:`7178964`].
```
nateanl, line 77: Could you add short description on what's the difference between this dataset and original librispeech dataset?

examples/asr/librispeech_biasing/train.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    parser.add_argument(
+        "--epochs",
+        default=120,
+        type=int,
+        help="Number of epochs to train for. (Default: 120)",
+    )
+    parser.add_argument(
+        "--subset",
+        default='clean100',
```
nateanl, line 121: I would suggest following the subset name of librispeech dataset. In this case it will be "train-clean-100".

examples/asr/librispeech_biasing/train.py
```diff
@@ -0,0 +1,160 @@
...
+        help="Train on subset of librispeech.",
+    )
+    parser.add_argument(
+        "--biasing",
+        type=str,
+        help="Use biasing",
+        required=True,
+    )
+    parser.add_argument(
+        "--biasinglist",
```
nateanl, line 132:
```suggestion
        "--biasing-list",
```

examples/asr/librispeech_biasing/train.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    parser.add_argument(
+        "--subset",
+        default='clean100',
+        type=str,
+        help="Train on subset of librispeech.",
+    )
+    parser.add_argument(
+        "--biasing",
+        type=str,
```
nateanl, line 127: I recommend using `action="store_true"` for `--biasing` so that it is set to `False` if not given.

examples/asr/librispeech_biasing/global_stats_100.json
```diff
@@ -0,0 +1,166 @@
+{
```
nateanl, line 1: Is it computed based on only "train-clean-100" subset? The file name need to match the default one in `train.py`

examples/asr/librispeech_biasing/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Evaluation
+
+[`eval.py`](./eval.py) evaluates a trained Conformer RNN-T model with TCPGen on LibriSpeech test-clean.
+
+Sample decoding script: `eval.sh`
+
+Other decoding options:
+
+- `--biasinglist` should be [`all_rare_words.txt`](blists/all_rare_words.txt) for Librispeech experiments
```
nateanl, line 43:
```suggestion
- `--biasing-list` should be [`all_rare_words.txt`](blists/all_rare_words.txt) for Librispeech experiments
```

examples/asr/librispeech_biasing/train.sh
nateanl: same for `eval.sh` file.

torchaudio/datasets/librispeech_biasing.py
BriansIDP: Hi @nateanl. Thanks for the comments! For this one, I saw we use all capital for many other data sets including Librispeech in torchaudio/datasets/librispeech.py. Should I keep it in the same form?

examples/asr/librispeech_biasing/global_stats_100.json
BriansIDP: It is. Because I was only able to perform experiments on train-clean-100 at the moment, but with the help from others it should finally be on 960-hour. So should I change the default in train.py to train-clean-100 here for this PR?

torchaudio/datasets/librispeech_biasing.py
nateanl: Yes, some datasets use all capital case because they were first added, but that is not the design we want to follow.  Instead we can follow the ones like `LibriMix` or `Snips`.

examples/asr/librispeech_biasing/global_stats_100.json
nateanl: I see. In that case we can stick with `global_stats.json` for both file name and the default value in the train.py for simplicity. Just make sure the values are updated here when you update the WER scores for 960 hours of training data.

BriansIDP:
Hi @mthrok @xiaohui-zhang @nateanl . I have addressed all the comments and pushed them to [#2890](https://github.com/pytorch/audio/pull/2890). Please let me know if I should create a new clean PR, or could @mthrok move things here to continue the code review? Thank you!

torchaudio/prototype/models/rnnt.py
```diff
@@ -52,6 +53,412 @@ def infer(
...
+
+    Recurrent neural network transducer (RNN-T) model.
+
+    Note:
+        To build the model, please use one of the factory functions.
+
+    Args:
+        transcriber (torch.nn.Module): transcription network.
+        predictor (torch.nn.Module): prediction network.
+        joiner (torch.nn.Module): joint network.
```
xiaohui-zhang, line 154: plz document other args as well

BriansIDP: Thanks! I have made the change and pushed it to the original PR.

xiaohui-zhang:
congrats @BriansIDP! looking forward to the tutorial and/or follow-up improvements. Thanks @mthrok for facilitating the merging process!

Issue #2979, opened 2023-01-13
Nightly CUDA build does not have cuda code


mthrok:
I was testing HW decoding notebook https://pytorch.org/audio/stable/hw_acceleration_tutorial.html, and got error [here](https://github.com/pytorch/audio/blob/main/torchaudio/csrc/ffmpeg/stream_reader/stream_reader.cpp#L243-L247).

The following code on fresh Google colab env reports `None`, where it should print CUDA version.

```python
!pip uninstall -y -q torch torchaudio torchvision torchtext
!pip install --progress-bar off --pre torch torchaudio --extra-index-url https://download.pytorch.org/whl/nightly/cu116 2> /dev/null

import torchaudio
print(torchaudio._extension._check_cuda_version())
```

Looking at the nightly build log, USE_CUDA macro is not defined anywhere.

https://github.com/pytorch/audio/actions/runs/3911306154/jobs/6684555785

mthrok:
Perhaps https://github.com/pytorch/audio/pull/2930 caused this,

seemethere:
@DanilBaibak can you take a look at this on Monday?

mthrok:
From https://github.com/pytorch/audio/actions/runs/3914313934/jobs/6691238420, 
`torch.cuda.is_available()` reports `False` [here](https://github.com/pytorch/audio/blob/main/tools/setup_helpers/extension.py#L42).

So it could be either PyTorch installation or GPU availability.

malfet:
Well, `torch.cuda.is_available()` should return false as compilation is running on CPU nodes (presumably)
But `torch.backends.cuda.is_built()` should yield True.
For example, `python -c "import torch;print(torch.cuda.is_available(), torch.backends.cuda.is_built(), torch.backends.mps.is_built())"` returns: `True True False` on Linux box with GPU, `False True False` on Linux box without GPU if PyTorch+CUDA is installed and `False False True` on MacOS

mthrok:
> Well, `torch.cuda.is_available()` should return false as compilation is running on CPU nodes (presumably)
> But `torch.backends.cuda.is_built()` should yield True.
> For example, `python -c "import torch;print(torch.cuda.is_available(), torch.backends.cuda.is_built(), torch.backends.mps.is_built())"` returns: `True True False` on Linux box with GPU, `False True False` on Linux box without GPU if PyTorch+CUDA is installed and `False False True` on MacOS

[...]

PR #3008, opened 2023-01-24
Updated USE_ROCM detection

DanilBaibak:
We don't need the presence of physical HW to compile with CUDA.

This is a follow up PR regarding `USE_ROCM` for issue https://github.com/pytorch/audio/issues/2979.


malfet (approved):

malfet (approved):
Looks good to me, but please confirm that after the change CUDA wheels are still building and rocm wheels also include compilation of `compute.hip`/invocation of `hipcc`

jeffdaily (approved):

PR #3015, opened 2023-01-27
Introduce I/O backend dispatcher

hwangjeff:
Adds I/O backend dispatcher that routes I/O requests to FFmpeg, SoX, or Soundfile backend, per library availability. It allows users to specify a backend mapped to a media library, i.e. one of `["ffmpeg", "sox", "soundfile"]`, to use via keyword argument, with FFmpeg being the default. Environment variable `TORCHAUDIO_USE_BACKEND_DISPATCHER` gates enablement of the dispatcher; specifically, if `TORCHAUDIO_USE_BACKEND_DISPATCHER` is explicitly set to `1`, importing TorchAudio makes it accessible via `torchaudio.info`, `torchaudio.load`, and `torchaudio.save`.

### DIFF
--- a/torchaudio/__init__.py
+++ b/torchaudio/__init__.py
@@ -11,6 +11,7 @@ from torchaudio import (  # noqa: F401
     transforms,
     utils,
 )
+
 from torchaudio.backend import get_audio_backend, list_audio_backends, set_audio_backend
 
 try:
--- a/torchaudio/_backend/utils.py
+++ b/torchaudio/_backend/utils.py
@@ -1,0 +1,557 @@
+import os
+import re
+from abc import ABC, abstractmethod
+from functools import lru_cache
+from typing import BinaryIO, Dict, Optional, Tuple, Union
+
+import torch
+import torchaudio.backend.soundfile_backend as soundfile_backend
+from torchaudio._extension import _FFMPEG_INITIALIZED, _SOX_INITIALIZED
+from torchaudio.backend.common import AudioMetaData
+from torchaudio.io._compat import info_audio, info_audio_fileobj, load_audio, load_audio_fileobj, save_audio
+
+
+class Backend(ABC):
+    @staticmethod
+    @abstractmethod
+    def info(uri: Union[BinaryIO, str, os.PathLike], format: Optional[str], buffer_size: int = 4096) -> AudioMetaData:
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def load(
+        uri: Union[BinaryIO, str, os.PathLike],
+        frame_offset: int = 0,
+        num_frames: int = -1,
+        normalize: bool = True,
+        channels_first: bool = True,
+        format: Optional[str] = None,
+        buffer_size: int = 4096,
+    ) -> Tuple[torch.Tensor, int]:
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def save(
+        uri: Union[BinaryIO, str, os.PathLike],
+        src: torch.Tensor,
+        sample_rate: int,
+        channels_first: bool = True,
+        format: Optional[str] = None,
+        encoding: Optional[str] = None,
+        bits_per_sample: Optional[int] = None,
+        buffer_size: int = 4096,
+    ) -> None:
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def can_decode(uri: Union[BinaryIO, str, os.PathLike], format: Optional[str]) -> bool:
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def can_encode(uri: Union[BinaryIO, str, os.PathLike], format: Optional[str]) -> bool:
+        raise NotImplementedError
+
+
+def _map_encoding(encoding: str) -> str:
+    for dst in ["PCM_S", "PCM_U", "PCM_F"]:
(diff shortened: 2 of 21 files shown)


torchaudio/__init__.py
```diff
@@ -11,7 +11,12 @@
     transforms,
     utils,
 )
-from torchaudio.backend import get_audio_backend, list_audio_backends, set_audio_backend
+
+import os
+if os.getenv("TORCHAUDIO_USE_BACKEND_DISPATCHER") == "1":
+    import torchaudio._backend  # noqa: F401
+else:
+    from torchaudio.backend import get_audio_backend, list_audio_backends, set_audio_backend
```
mthrok, line 19: I think it's better to keep these attributes available and change only the behavior when `TORCHAUDIO_USE_BACKEND_DISPATCHER` is enabled, like warning message and no-op.

torchaudio/io/_compat.py
```diff
@@ -2,7 +2,61 @@
...
+            * ``FLAC``: Flac, Free Lossless Audio Codec
+            * ``ULAW``: Mu-law
+            * ``ALAW``: A-law
+            * ``MP3`` : MP3, MPEG-1 Audio Layer III
+            * ``VORBIS``: OGG Vorbis
+            * ``AMR_WB``: Adaptive Multi-Rate
+            * ``AMR_NB``: Adaptive Multi-Rate Wideband
+            * ``OPUS``: Opus
+            * ``HTK``: Single channel 16-bit PCM
+            * ``UNKNOWN`` : None of above
```
mthrok, line 33: Later we need to test that these formats are same regardless of the backend picked implicitly, at least for common ones.

torchaudio/_backend/utils.py
```diff
@@ -0,0 +1,292 @@
...
+    def info(uri: Union[BinaryIO, str], format: Optional[str]) -> AudioMetaData:
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def load(
+        uri: Union[BinaryIO, str],
+        frame_offset: int = 0,
+        num_frames: int = -1,
+        normalize: bool = True,
```
mthrok, line 32: The name `normalize` caused confusion to many users who thought it was about volume normalization.
let's think about some alternative name for this argument.

torchaudio/_backend/utils.py
```diff
@@ -0,0 +1,292 @@
...
+    @abstractmethod
+    def can_encode(uri, format) -> bool:
+        raise NotImplementedError
+
+
+class FFmpegBackend(Backend):
+    @staticmethod
+    def info(uri: Union[BinaryIO, str], format: Optional[str]) -> AudioMetaData:
+        if hasattr(uri, "read"):
+            buffer_size = get_buffer_size()
```
mthrok, line 64: How about making buffer_size a keyword argument?
I think it's confusing that this value is configurable via sox_utils though it's ffmpeg.

torchaudio/_backend/utils.py
```diff
@@ -0,0 +1,292 @@
...
+        backend_specs["ffmpeg"] = FFmpegBackend
+    if _SOX_INITIALIZED:
+        backend_specs["sox"] = SoXBackend
+    if soundfile_backend._IS_SOUNDFILE_AVAILABLE:
+        backend_specs["soundfile"] = SoundfileBackend
+    return backend_specs
+
+
+def get_backend(backend_name, backends) -> Backend:
+    if (backend := backends.get(backend_name)) :
```
mthrok, line 214: Nice. The first walrus operator in torchaudio!

torchaudio/_backend/utils.py
hwangjeff: `normalize` seems fairly apt for the behavior the arg gates: normalizing the input int data to range [-1.0, 1.0] after converting to float32. Perhaps we can simply update the docstring to clarify this?

torchaudio/__init__.py
```diff
@@ -11,8 +13,12 @@
     transforms,
     utils,
 )
+
 from torchaudio.backend import get_audio_backend, list_audio_backends, set_audio_backend
```
mthrok, line 17: Here, when `torchaudio.backend` is imported, the conventional backend initialization logic is executed in [`torchaudio.backend.__init__`](https://github.com/pytorch/audio/blob/main/torchaudio/backend/__init__.py).

I think it's easier to follow if the initialization logic is consolidated to one place. Either here (`torchaudio.__init__`) or in `torchaudio.backend.__init__.py`.

What do you think?

torchaudio/io/_compat.py
```diff
@@ -116,3 +117,151 @@ def load_audio_fileobj(
...
+
+def _native_endianness():
+    if sys.byteorder == "little":
+        return "le"
+    else:
+        return "be"
+
+
+def _get_encoder_for_wav(dtype, encoding, bits_per_sample):
+    assert bits_per_sample in {None, 8, 16, 24, 32, 64}
```
mthrok, line 145: `assert` does not run in optimized mode, (so in general `assert` should be used for validating logic internal to implementation) can we replace this with `if ~ raise` pattern?

mthrok (commented):
If it's too much work, it's fine but can we extract some implementations into separate PRs? 

1. get_demuxer implementation
2. _compat.save implementations (just implementation, no need for test)
3. typo

PR #3017, opened 2023-01-28
Fix hybrid demucs tutorial for CUDA

eyeplum:
Currently there will be a few errors when this tutorial is run with a CUDA device.

The reasons being:
- The source audio waveform is not properly moved to the GPU. The `to()` method is not in-place for Tensors, so we need to assign the return value of the method call to the variable (otherwise the Tensor would still be on the CPU).
- When performing further analysis and displaying of the output audio, we need to move them back from the GPU to the CPU. This is because some of the functions we call require the Tensor to be on the CPU (e.g. `stft()` and `bss_eval_sources()`).

Remaining question:
- Do we need to manually update [this page](https://pytorch.org/audio/stable/tutorials/hybrid_demucs_tutorial.html)?


nateanl:
Thanks for the fix @eyeplum. The document for nightly version is automatically updated. For 0.13 version I can update it once the pr is merged.

nateanl (approved):

PR #3018, opened 2023-01-28
feat: cuda kernel for lfilter

### LINKED ISSUE #1408: run lowpass_biquad in gpu is much slower that run in cpu
## 🐛 Bug

Hi everyone, I am doing some experiments via lowpass_biquad filter in torchaudio, but I was trying to use GPU to speed up by code. I found that running the lowpass_biquad filter in GPU is much slower than running in CPU (x 200 slower). I am not sure whether it is a bug, but I am curious about whether anyone has the same situation as me?

## To Reproduce

```
import time
import torch
from torchaudio.functional import lowpass_biquad

gpu_device = torch.device('cuda:0')
cpu_device = torch.device('cpu')

sample_rate = 44100
cutoff_freq = 1000.
Q           = .7

# Run in cpu
x     = torch.rand(sample_rate * 10, device=cpu_device)
begin = time.time()
y     = lowpass_biquad(x, sample_rate, cutoff_freq, Q)

print(f'Run in cpu: {time.time() - begin}')

# Run in gpu
x     = torch.rand(sample_rate * 10, device=gpu_device)
begin = time.time()
y     = lowpass_biquad(x, sample_rate, cutoff_freq, Q)

print(f'Run in gpu: {time.time() - begin}')
```

```
Run in cpu: 0.01553034782409668
Run in gpu: 21.124146461486816
```

## Expected behavior

Run in GPU should much faster that run in CPU

## Environment

 - What commands did you used to install torchaudio (conda/pip/build from source)? pip
 - If you are building from source, which commit is it? no
 - What does `torchaudio.__version__` print? (If applicable) 0.8.0

 - PyTorch Version (e.g., 1.0): 1.8.0
 - OS (e.g., Linux): Linux Ubuntu 18.04.3 LTS (Bionic Beaver)
 - How you installed PyTorch (`conda`, `pip`, source): pip
 - Build command you used (if compiling from source): no
 - Python version: 3.6.12
 - CUDA/cuDNN version: 10.0.130

```
[pip3] python==3.6.12
[pip3] conda==4.9.2
[pip3] torch==1.8.0
[pip3] torchaudio==0.8.0
```

yoyololicon:
close #1408 .


yoyololicon:
Hi @mthrok @nateanl ,

I think the code should be compilable on a machine with GPU, but I got the following error:
```
In file included from /home/ycy/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/include/torch/csrc/Device.h:4,
                 from /home/ycy/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/include/torch/csrc/api/include/torch/python.h:8,
                 from /home/ycy/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/include/torch/extension.h:6,
                 from /home/ycy/working/audio/torchaudio/csrc/iir_cuda.h:1,
                 from /home/ycy/working/audio/torchaudio/csrc/lfilter.cpp:3:
/home/ycy/miniconda3/envs/nightly/lib/python3.10/site-packages/torch/include/torch/csrc/python_headers.h:12:10: fatal error: Python.h: No such file or directory
   12 | #include <Python.h>
      |          ^~~~~~~~~~
compilation terminated.
```
This also happens on all the failed checks, so there must be something wrong in the compilation process.
Could you help me look into this or provide guidance on possible reasons?
I'm using conda and can build the latest torchaudio with cuda successfully.

torchaudio/csrc/iir_cuda.cu
```diff
@@ -0,0 +1,78 @@
+#include <torch/extension.h>
```
mthrok, line 1: `<torch/extension.h>` header for writing Python binding and it attempts to include Python header. Here `torch/torch.h` should do.

```suggestion
#include <torch/torch.h>
```

torchaudio/csrc/iir_cuda.h
```diff
@@ -0,0 +1,6 @@
+#include <torch/extension.h>
```
mthrok, line 1:
```suggestion
#include <torch/types.h>
```

mthrok (commented):
@yoyololicon Thanks for the PR! This is great.

yoyololicon:
@mthrok Thanks for helping me! 
Now the cuda builds are successful. I'm looking into customizing CPU builds.

mthrok:
Thanks @yoyololicon 

There are four tests related to lfilter implementation, which are run for CPU and CUDA.
Do you think the coverage is good enough? Do we need a new test?

https://github.com/pytorch/audio/blob/f663cb2874f918e7663c0daa1d3a3ae8636349ad/test/torchaudio_unittest/functional/functional_impl.py#L59-L144

torchaudio/csrc/lfilter.cpp
```diff
@@ -96,6 +98,12 @@ void lfilter_core_generic_loop(
   }
 }
 
+#ifndef USE_CUDA
+
+std::function<void(const torch::Tensor&, const torch::Tensor&, torch::Tensor&)>
+    cuda_lfilter_core_loop = lfilter_core_generic_loop;
```
mthrok, line 104: This is a cool trick, however, it feels like reducing the readability of code.
How about simply branching with USE_CUDA macro at L130 (inside of `else if (device.is_cuda())` branch?)


```c++
else if (device.is_cuda()) {
#ifdef USE_CUDA
    cuda_lfilter_core_loop(waveform, a_coeff_flipped, padded_output_waveform);
#else
    lfilter_core_generic_loop(waveform, a_coeff_flipped, padded_output_waveform);
#endif
}

```

torchaudio/csrc/iir_cuda.cu
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#ifdef USE_CUDA
```
mthrok, line 3: Since it's kind of premise that `iir_cuda.cu` and `iir_cuda.h` are used only when USE_CUDA is enabled, so I prefer to not use `#ifdef USE_CUDA` in these files, and rather use it in `lfilter.cpp`, where the final product is composed.
What do you think?

torchaudio/csrc/iir_cuda.cu
```diff
@@ -0,0 +1,84 @@
...
+                size_t>(),
+            padded_out.packed_accessor<
+                scalar_t,
+                3,
+                torch::RestrictPtrTraits,
+                size_t>());
+      }));
+}
+
+#endif
```
mthrok, line 84: Please add a new line at EOF. (same goes to the header file.)

mthrok (commented):
@yoyololicon Thanks for the PR!

As I have limited knowledge about lfilter, I assume that you did appropriate testing to ensure that the implementation is numerically correct.

torchaudio/csrc/lfilter.cpp
yoyololicon: Good, it's more readable.

torchaudio/csrc/iir_cuda.cu
yoyololicon: Agree.

torchaudio/csrc/iir_cuda.cu
yoyololicon: Totally forgot about this 😅

yoyololicon:
@mthrok The tests are enough.

mthrok (approved):
Looks good. Thanks!

torchaudio/csrc/lfilter.cpp
```diff
@@ -1,5 +1,6 @@
 #include <torch/script.h>
 #include <torch/torch.h>
+#include <torchaudio/csrc/iir_cuda.h>
```
mthrok, line 3: @yoyololicon Could you also apply `#ifdef USE_CUDA` here?

Some of the builds with more strict complier setting are failing with 

`Stderr: ld.lld: error: undefined symbol: cuda_lfilter_core_loop(at::Tensor const&, at::Tensor const&, at::Tensor&)`

yoyololicon: I see. I just commited it, did that work?

mthrok: let me try again.

mthrok: Actually it was something else. Things should be good. Will try to land this by tomorrow.

PR #3023, opened 2023-01-31
stream_reader: add rgb48le and CUDA p010 support (HDR/10bit)

d4l3k:
This adds 2 10 bit pix formats one for CPU and one for CUDA. This allows for training on HDR/10bit video datasets.

Test plan:

```py
r = StreamReader(
    reader, format='hevc',
)
stream = r.add_video_stream(
    frames_per_chunk=-1,
    decoder="hevc_cuvid",
    hw_accel="cuda",
)
frame = next(r.stream())
```

```py
r = StreamReader(
    reader, format='hevc',
)
stream = r.add_video_stream(
    frames_per_chunk=-1,
    filter_desc="format=rgb48le",
)
frame = next(r.stream())
```

audio-example


d4l3k:
@mthrok when you get a chance could you take a look at this? Thanks!

mthrok:
> @mthrok when you get a chance could you take a look at this? Thanks!

@d4l3k Thanks for the PR and quickly skimming, the code looks good. Is there a public video data of this format that I can test?

mthrok:
@d4l3k What is the order of color? RGB? YUV?

I created a synthetic data with the following command,
`ffmpeg -f lavfi -i mandelbrot -t 3 -c:v libx265 -pix_fmt yuv420p10le -vtag hvc1 -y test.hvec`.

and tested the PR code

```python
import torch

src = "test.hevc"

r = StreamReader(src=src)
r.add_video_stream(
    frames_per_chunk=-1,
    decoder="hevc_cuvid",
    hw_accel="cuda",
)
r.process_all_packets()
chunks, = r.pop_chunks()

print(chunks.device, chunks.shape, chunks.dtype)

plt.imshow(frames[0].permute(1, 2, 0) / 32768.)
plt.show()
```

This is what I got 
Figure_1

while what is expected is like https://pytorch.org/audio/stable/tutorials/streamreader_advanced_tutorial.html#mandelbrot

torchaudio/csrc/ffmpeg/stream_reader/buffer/common.cpp
```diff
@@ -307,6 +350,58 @@ void write_nv12_cuda(AVFrame* pFrame, torch::Tensor& yuv) {
...
+  }
+  uv = uv.permute({0, 3, 1, 2});
+  using namespace torch::indexing;
+  // Write to the UV plane
+  // very simplistic upscale using indexing since interpolate doesn't support
+  // shorts
+  yuv.index_put_({Slice(), Slice(1), Slice(None, None, 2), Slice(None, None, 2)}, uv);
+  yuv.index_put_({Slice(), Slice(1), Slice(1, None, 2), Slice(None, None, 2)}, uv);
+  yuv.index_put_({Slice(), Slice(1), Slice(None, None, 2), Slice(1, None, 2)}, uv);
+  yuv.index_put_({Slice(), Slice(1), Slice(1, None, 2), Slice(1, None, 2)}, uv);
```
mthrok, line 401: I thought `Slice(1)` translate to `:1`, like in Python's `slice(1)` but the code is working so it seems to translate to `1:`. (Perhaps this is where libtorch differs from Python?)
I think `Slice(1, 3)` is better as it's more explicit that we are pushing 2 channel UV tensor.

```suggestion
  yuv.index_put_({Slice(), Slice(1, 3), Slice(None, None, 2), Slice(None, None, 2)}, uv);
  yuv.index_put_({Slice(), Slice(1, 3), Slice(1, None, 2), Slice(None, None, 2)}, uv);
  yuv.index_put_({Slice(), Slice(1, 3), Slice(None, None, 2), Slice(1, None, 2)}, uv);
  yuv.index_put_({Slice(), Slice(1, 3), Slice(1, None, 2), Slice(1, None, 2)}, uv);
```

torchaudio/csrc/ffmpeg/stream_reader/buffer/common.cpp
```diff
@@ -96,6 +96,16 @@ torch::Tensor get_buffer(
   return torch::empty(shape, options);
 }
 
+torch::Tensor get_buffer16(
```
mthrok, line 99: I think this function can be merged to `get_buffer`, by adding `dtype` argument which defaults to `kUInt8`.

torchaudio/csrc/ffmpeg/stream_reader/buffer/common.cpp
```diff
@@ -128,6 +138,25 @@ std::tuple<torch::Tensor, bool> get_image_buffer(
...
+      AVPixelFormat sw_format = hwctx->sw_format;
+      switch (sw_format) {
+        case AV_PIX_FMT_P010: {
+          auto buffer = get_buffer16({num_frames, channels, height, width}, device);
+          return std::make_tuple(buffer, true);
+        }
+      }
+    }
+#endif
+  }
```
mthrok, line 159: Can you add `default:` label so that compiler won't issue warning on missing branches?

mthrok, line 159: follow-up: I think this part can be generalized using [`av_get_bits_per_pixel`](https://ffmpeg.org/doxygen/trunk/pixdesc_8c.html#a7b24a65f6bf07585396da3bd301799b2), and get rid of this switch statement.

mthrok:
> @mthrok P010 is cuvids name for yuv420p10le so it should be outputting yuv channels. rgb48le is rgb

ah okay, my visualization code was wrong. I mistook int16 and uint16. Now I get the proper output.

Figure_1

```python
def yuv_to_rgb(img):
    y = img[..., 0, :, :]
    u = img[..., 1, :, :]
    v = img[..., 2, :, :]

    u = u - 0.5
    v = v - 0.5

    r = y + 1.14 * v
    g = y + -0.396 * u - 0.581 * v
    b = y + 2.029 * u

    rgb = torch.stack([r, g, b], -1)
    rgb = (rgb * 255).clamp(0, 255).to(torch.uint8)
    return rgb


frames = frames.to(torch.float) / 65536 + 0.5
rgb = yuv_to_rgb(frames)[0]

plt.imshow(rgb)
plt.show()
```

mthrok:
Do you have a bandwidth to add test?
Currently we do not have testing for GPU decoder due to the complexity in CI, but it would be nice to have a tests for CPU.

d4l3k (commented):
added a unit test

mthrok (approved):

torchaudio/csrc/ffmpeg/stream_reader/buffer/common.cpp
```diff
@@ -148,6 +153,20 @@ void write_interlaced_image(AVFrame* pFrame, torch::Tensor& frame) {
...
+  uint8_t* buf = pFrame->data[0];
+  size_t height = frame.size(1);
+  size_t stride = frame.size(2) * frame.size(3);
+  for (int i = 0; i < height; ++i) {
+    memcpy(ptr, buf, stride * 2);
+    buf += pFrame->linesize[0];
+    ptr += stride;
+  }
+  // correct for int16
+  frame += 32768;
```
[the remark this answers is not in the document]
d4l3k: Ffmpeg scales the values to be between 0 and 2**16-1 as a uint16 even though they're only 10 bit. That offset corrects for the raw uint16 to int16 cast and makes it -2**15 to 2**15 range so it's linear. 

No strong preference here though

PR #3026, opened 2023-01-31
Add playback function

jazcarretao:
Summary: Allows user to play audio through the
device speaker.

Test plan:
Created a new test that mocks a call to the write audio chunk method from StreamWriter. To run the test:

`pytest test/torchaudio_unittest/io/_playback_test.py`


torchaudio/io/_playback.py
```diff
@@ -0,0 +1,55 @@
...
+            torch.float64: "dbl"
+        }[dtype]
+    except KeyError:
+        raise ValueError(f"Unsupported dtype: {dtype}")
+
+    if device is None:
+        device = torchaudio.utils.ffmpeg_utils.get_output_devices()
+        # need to get device key from dictionary
+    elif device == "audiotoolbox":
+        path = "-"
```
mthrok, line 40: The current implementation only supports macOS, so it is better to explicitly raise an error for other OS.

We will later need to pick an appropriate device based on OS and the availability of output device. Until then we do not need to use get_output_devices func. IIRC, `"audiotoolbox"` is the only supported output device on macOS.

```
if <OS is macOS>:
    device = device or "audiotoolbox"
    path = "-"
else:
    raise RuntimeError("This function is only supported on macOS.")
```

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,55 @@
...
+import torchaudio
+from torchaudio.io import StreamWriter
+from typing import Optional
+
+
+def play_audio(
+    waveform: torch.Tensor,
+    sample_rate: Optional[float],
+    device: Optional[str] = "audiotoolbox",
+    channels_first: Optional[bool] = True,
```
mthrok, line 11: Let's simplify the function and remove this `channels_first` argument, and make **channel LAST** a requirement. it's easy for users to transpose the input.

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,55 @@
...
+    if device is None:
+        device = torchaudio.utils.ffmpeg_utils.get_output_devices()
+        # need to get device key from dictionary
+    elif device == "audiotoolbox":
+        path = "-"
+
+    if channels_first:
+        waveform = waveform.transpose(0, 1)
+
+    num_frames, num_channels = waveform.size()
```
mthrok, line 45: Before doing this we need to check the `waveform.ndim == 2` and raise an error if not.

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,55 @@
...
+        format = {
+            torch.uint8: "u8",
+            torch.int16: "s16",
+            torch.int32: "s32",
+            torch.int64: "s64",
+            torch.float32: "flt",
+            torch.float64: "dbl"
+        }[dtype]
+    except KeyError:
+        raise ValueError(f"Unsupported dtype: {dtype}")
```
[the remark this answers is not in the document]
jazcarretao: I agree with the above. Do you have a suggestion on where to place the dictionary with the formats? Somewhere where it can be easily re-used by other methods or test if necessary. I can place it in the top of this file for now (below the imports).

torchaudio/io/_playback.py
mthrok: yeah agree. looks good.

torchaudio/io/_playback.py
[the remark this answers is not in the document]
mthrok: I am indifferent.

It is attempting to put it in module-level, but that means the dictionary is created for all the users who do not use this function. It's kinda waste of memory, though it's very small. This function does not have to be super fast, so in my opinion, we don't gain much from putting it in module-level.

I'lld let you decide.

test/torchaudio_unittest/io/playback_test.py
```diff
@@ -0,0 +1,23 @@
...
+from unittest.mock import patch
+
+
+@skipIfNoDevice
+class PlaybackInterfaceTest(TorchaudioTestCase):
+    @patch.object(StreamWriter, "write_audio_chunk")
+    def test_playaudio(self, writeaudio_mock):
+        """Test playaudio function"""
+        file = get_asset_path("steam-train-whistle-daniel_simon.mp3")
+        waveform, sample_rate = torchaudio.load(file)
```
mthrok, line 18: In the unit test, we want to limit the use of library function to what we are testing. (Otherwise, when this test fails, we cannot know immediately which function has the issue)

Can you use [`get_sinusoid`](https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/data_utils.py#LL82C7-L82C7)  helper function to generate a test data?

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,61 @@
...
+    available_devices = list(torchaudio.utils.ffmpeg_utils.get_output_devices().keys())
+    if device not in available_devices:
+        raise ValueError(f"Device {device} is not available. Available devices are: {available_devices}")
+
+
+    num_frames, num_channels = waveform.size()
+
+    # Write to speaker device
+    s = StreamWriter(dst=path, format=device)
+    s.add_audio_stream(sample_rate, num_channels, format=format)
```
mthrok, line 55: What happens if `num_channels >= 3`? Does it work?

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,68 @@
...
+        device = device or "audiotoolbox"
+        path = "-"
+    else:
+        raise ValueError(f"This function only supports MacOS, \
+            but current OS is {platform}")
+
+    available_devices = list(torchaudio.utils.ffmpeg_utils.get_output_devices().keys())
+    if device not in available_devices:
+        raise ValueError(f"Device {device} is not available. \
+            Available devices are: {available_devices}")
```
mthrok, line 55: The use of backslash is against the guideline.

Please check this and fix the style. https://app.circleci.com/pipelines/github/pytorch/audio/14667/workflows/45d93c58-ba1e-4f7f-976f-2cc45c748f5c/jobs/1120450

A simple way is to install pre-commit and run `pre-commit run -a`.

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,76 @@
...
+        raise ValueError(
+            f"This function only supports MacOS, \
+            but current OS is {platform}"
+        )
+
+    available_devices = list(torchaudio.utils.ffmpeg_utils.get_output_devices().keys())
+    if device not in available_devices:
+        raise ValueError(
+            f"Device {device} is not available. \
+            Available devices are: {available_devices}"
```
mthrok, line 63: Please do not use backslash at all.
The use of backslash like this make the indentation part of the message.

```python
>>> platform="foo"
>>> print(
...             f"This function only supports MacOS, \
...             but current OS is {platform}"
...         )
This function only supports MacOS,             but current OS is foo
```

```suggestion
            f"Device {device} is not available. Available devices are: {available_devices}"
```

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,76 @@
...
+    elif waveform.ndim == 1:
+        waveform = waveform.unsqueeze(-1)
+
+    if platform == "darwin":
+        device = device or "audiotoolbox"
+        path = "-"
+    else:
+        raise ValueError(
+            f"This function only supports MacOS, \
+            but current OS is {platform}"
```
mthrok, line 56:
```suggestion
            f"This function only supports MacOS, but current OS is {platform}"
```

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    if waveform.ndim > 2:
+        raise ValueError(
+            f"Expected 2D stereo tensor with shape \
+            (time, channels), or 1D mono tensor with shape (time), \
+                got {waveform.ndim}D tensor instead"
+        )
+    elif waveform.ndim == 1:
+        waveform = waveform.unsqueeze(-1)
```
mthrok, line 48: Please only accept 2D, and reject 1D.

torchaudio/io/_playback.py
```diff
@@ -0,0 +1,76 @@
...
+        device: Output device to use. If None, the default device is used.
+    """
+
+    if waveform.dtype in dict_format.keys():
+        format = dict_format[waveform.dtype]
+    else:
+        raise ValueError(
+            f"Unsupported dtype: {waveform.dtype}. \
+            The list of supported dtypes is: {dict_format.keys()}"
+        )
```
mthrok, line 39:
```suggestion
    if waveform.dtype not in dict_format:
        raise ValueError(
            f"Unsupported dtype: {waveform.dtype}. The list of supported dtypes is: {dict_format.keys()}"
        )
    format = dict_format[waveform.dtype]        
```

torchaudio/io/_playback.py
mthrok: Can you experiment with num_channels >=3 and report what happens?

mthrok (approved):

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -253,6 +253,16 @@ def skipIfNoModule(module, display_name=None):
     ),
     key="ON_PYTHON_310",
 )
+skipIfNoDevice = _skipIf(
```
mthrok, line 256: Can we rename this to `skipIfNoAudioDevice`?
In PyTorch context, normally device refers to compute device like CUDA

.circleci/config.yml
```diff
@@ -646,6 +646,8 @@ jobs:
           environment:
               TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_CUDA: true
               TORCHAUDIO_TEST_ALLOW_SKIP_IF_ON_PYTHON_310: true
+              TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_DEVICE: true
+              TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MACOS: true
```
mthrok, line 650: `config.yml` is auto-generated from `config.yml.in`. Can you apply the same change to `config.yml.in` and run `regenerate.py` then check-in both `config.yml` and `config.yml.in`?

test/torchaudio_unittest/io/playback_test.py
```diff
@@ -0,0 +1,65 @@
...
+from parameterized import parameterized
+from torchaudio.io import play_audio, StreamWriter
+from torchaudio_unittest.common_utils import get_sinusoid, skipIfNoDevice, skipIfNoMacOS, TorchaudioTestCase
+
+
+@skipIfNoDevice
+@skipIfNoMacOS
+class PlaybackInterfaceTest(TorchaudioTestCase):
+    @parameterized.expand(
+        [(torch.uint8,), (torch.int16,), (torch.int32,), (torch.int64,), (torch.float32,), (torch.float64,)]
```
mthrok, line 13: Please use str-type, as in some cases, this arguments become a part of test name and path, and cause an issue.

```suggestion
        [("uint8",), ("int16",), ("int32",), ("int64",), ("float32",), ("float64",)]
```

then inside of the test, do `dtype = getattr(torch, dtype)`

PR #3039, opened 2023-02-07
Add python 3.11 support for torchaudio and add workflow concurrency rule

weiwangmeta:
So far Linux and MacOS were tested to work fine out of the box. This PR is created to verify this -- disabled windows jobs and configs for now.


osalpekar (approved):
Once we add the if-exclusion for conda (which might be removed later), this is good to go!

PR #3055, opened 2023-02-13
Add deprecation warning to decoder

mthrok:
Flashlight Text decoder is now available on PyPI and KenLM support is being added at
https://github.com/flashlight/text/pull/43

Once this work is merged, we can rely on the official distribution of Flashlight Text package, so we are adding deprecation warning.

Once the decoder is fully available, one can install it with

```
pip install flashlight-text
pip install git+https://github.com/kpu/kenlm.git
```

### DIFF
--- a/torchaudio/models/decoder/_ctc_decoder.py
+++ b/torchaudio/models/decoder/_ctc_decoder.py
@@ -9,13 +9,13 @@ import torch
 import torchaudio
 from torchaudio.utils import download_asset
 
-try:
-    # We prioritize the version from upstream flashlight here.
-    # This will allow applications that use the upstream flashlight
-    # alongside torchaudio.
+
+# We prioritize the version from upstream flashlight here.
+# This will allow applications that use the upstream flashlight
+# alongside torchaudio.
+if torchaudio._internal.module_utils.is_module_available("flashlight"):
     from flashlight.lib.text.decoder import (
         CriterionType as _CriterionType,
-        KenLM as _KenLM,
         LexiconDecoder as _LexiconDecoder,
         LexiconDecoderOptions as _LexiconDecoderOptions,
         LexiconFreeDecoder as _LexiconFreeDecoder,
@@ -31,7 +31,14 @@ try:
         Dictionary as _Dictionary,
         load_words as _load_words,
     )
-except Exception:
+    try:
+        from flashlight.lib.text.decoder import KenLM as _KenLM
+    except Exception:
+        _KenLM = None
+else:
+    warnings.warn(
+        "The built-in flashlight integration is deprecated, and will be removed in future release."
+        "Please install flashlight-text. https://pypi.org/project/flashlight-text/")
     torchaudio._extension._load_lib("libflashlight-text")
     from torchaudio.lib.flashlight_lib_text_decoder import (
         CriterionType as _CriterionType,
@@ -427,6 +434,9 @@ def ctc_decoder(
     word_dict = _get_word_dict(lexicon, lm, lm_dict, tokens_dict, unk_word)
 
     if type(lm) == str:
+        if _KenLM is None:
+            raise RuntimeError(
+                "flashlight is installed, but KenLM is not installed. Please install KenLM.")
         lm = _KenLM(lm, word_dict)
     elif lm is None:
         lm = _ZeroLM()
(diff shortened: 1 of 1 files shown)


torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -9,13 +11,13 @@
...
 
-try:
-    # We prioritize the version from upstream flashlight here.
-    # This will allow applications that use the upstream flashlight
-    # alongside torchaudio.
+
+# We prioritize the version from upstream flashlight here.
+# This will allow applications that use the upstream flashlight
+# alongside torchaudio.
+if torchaudio._internal.module_utils.is_module_available("flashlight"):
```
jacobkahn, line 18: Flashlight is a namespace package FYI -- `flashlight.lib.sequence` is another package but doesn't have any of what you need. You might want to check if `flashlight.lib.text` is available.

torchaudio/models/decoder/_ctc_decoder.py
```diff
@@ -31,7 +33,16 @@
         Dictionary as _Dictionary,
         load_words as _load_words,
     )
-except Exception:
+
+    try:
+        from flashlight.lib.text.decoder import KenLM as _KenLM
```
jacobkahn, line 38: Any reason you wouldn't want to change this to `flashlight.lib.text.decoder.kenlm` once I merge things in if you're already changing this stuff as is? Given that these are the first PyPI releases we've done, I don't care about breaking things on our side this early.

torchaudio/models/decoder/_ctc_decoder.py
mthrok: The underlying implementation [`importlib.util.find_spec`](https://docs.python.org/3/library/importlib.html#importlib.util.find_spec) only works for top-level module, if the target module is not installed.
To inspect "flashlight.lib.text", it actually attempt to import the "flashlight.lib", so if flashlight is not available it fails with exception like the following.

The goal here is to inspect the module's existence without importing it (so that we won't have symbols from flashlight in case the initialization logic decides not to import from there), so we make assumption that (we ask users to) install when installing flashlight, decoder submodule is also installed, which I think is a reasonable assumption.

```
>>> torchaudio._internal.module_utils.is_module_available("ujktltvbhbvkrdnuvukjhejdhivetkeh")
False
>>> torchaudio._internal.module_utils.is_module_available("ujktltvbhbvkrdnuvukjhejdhivetkeh.foo")
  File "/Users/moto/Development/torchaudio/torchaudio/_internal/module_utils.py", line 14, in is_module_available
    return all(importlib.util.find_spec(m) is not None for m in modules)
  File "/Users/moto/Development/torchaudio/torchaudio/_internal/module_utils.py", line 14, in <genexpr>
    return all(importlib.util.find_spec(m) is not None for m in modules)
  File "/Users/moto/miniconda3/lib/python3.9/importlib/util.py", line 94, in find_spec
    parent = __import__(parent_name, fromlist=['__path__'])
ModuleNotFoundError: No module named 'ujktltvbhbvkrdnuvukjhejdhivetkeh'
>>>
```

torchaudio/models/decoder/_ctc_decoder.py
mthrok: From the strict backward compatibility perspective, this "if flashlight is available import from there" logic existed for users who installed flashlight from the source before PyPI release, and this initialization logic never worked for PyPI release, as it lacks KenLM, it is preferrable to keep the same naming sematics. I am fine with changing it to the new one, but that means, we ask users who installed FL Text from source to upgrade.

mthrok: BTW I just realized that this logic is used in fbcode, so to change the KenLM naming convention to `flashlight.lib.text.decoder.kenlm`, then I think the corresponding part of torchaudio has to be changed at the same time in fbcode.

jacobkahn: Right -- as an alternative, could we potentially try to import from both places and show a deprecation warning if the import with this method succeeds? That way, things can still work in fbcode once everything is changed.

torchaudio/models/decoder/_ctc_decoder.py
jacobkahn: Ah, I didn't know `is_module_available` used `find_spec` -- makes sense 👍

hwangjeff (approved):

PR #3061, opened 2023-02-14
Enable broadcasting for inputs to convolve

hwangjeff:
Relaxes input dimension matching constraint on `convolve` to enable broadcasting for inputs.

### DIFF
--- a/test/torchaudio_unittest/functional/functional_impl.py
+++ b/test/torchaudio_unittest/functional/functional_impl.py
@@ -937,26 +937,22 @@ class Functional(TestBaseMixin):
 
         self.assertEqual(expected, actual)
 
-    @parameterized.expand(
-        [
-            # fmt: off
-            ((5, 2, 3), (5, 1, 3)),
-            ((5, 2, 3), (1, 2, 3)),
-            ((5, 2, 3), (1, 1, 3)),
-            # fmt: on
-        ]
+    @nested_params(
+        ["convolve", "fftconvolve"],
+        [(5, 2, 3)],
+        [(5, 1, 3), (1, 2, 3), (1, 1, 3)],
     )
-    def test_fftconvolve_broadcast(self, x_shape, y_shape):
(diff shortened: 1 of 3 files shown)


torchaudio/functional/functional.py
```diff
@@ -2398,14 +2398,19 @@ def convolve(x: torch.Tensor, y: torch.Tensor, mode: str = "full") -> torch.Tens
     .. _convolution:
         https://en.wikipedia.org/wiki/Convolution
     """
-    _check_shape_compatible(x, y, allow_broadcast=False)
+    _check_shape_compatible(x, y, allow_broadcast=True)
```
mthrok, line 2401: After this change, `_check_shape_compatible` is always called with `allow_broadcast=True`, so can we remove the arg from `_check_shape_compatible`?

test/torchaudio_unittest/functional/functional_impl.py
```diff
@@ -972,17 +968,13 @@ def test_fftconvolve_broadcast(self, x_shape, y_shape):
...
-            (1, F.convolve, (5, 2, 3), (5, 3, 3)),
-            (1, F.convolve, (5, 2, 3), (5, 3, 4)),
-            (1, F.convolve, (5, 2, 3), (5, 3, 5)),
+            # non-broadcastable leading dimensions
+            (2, F.convolve, (5, 2, 3), (5, 3, 3)),
+            (2, F.convolve, (5, 2, 3), (5, 3, 4)),
+            (2, F.convolve, (5, 2, 3), (5, 3, 5)),
             (2, F.fftconvolve, (5, 2, 3), (5, 3, 3)),
             (2, F.fftconvolve, (5, 2, 3), (5, 3, 4)),
             (2, F.fftconvolve, (5, 2, 3), (5, 3, 5)),
```
mthrok, line 977: The current case `1` no longer exists, so can we remove the message for case 1 and migrate the what is case=2 now to case=1?

mthrok (approved):

PR #3064, opened 2023-02-15
Tweak docs around IO

mthrok:
* Mention context manager in StreamWriter
* Add FFmpeg as optional dependency

### DIFF
--- a/torchaudio/io/_stream_writer.py
+++ b/torchaudio/io/_stream_writer.py
@@ -226,7 +226,9 @@ class StreamWriter:
 
         :py:class:`StreamWriter` is also a context manager and therefore supports the
         ``with`` statement.
-        It is recommended use context manager, as the file is closed automatically
+        This method returns the instance on which the method is called (i.e. `self`),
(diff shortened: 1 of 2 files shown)


torchaudio/io/_stream_writer.py
```diff
@@ -224,6 +224,11 @@ def _print_output_stream(self, i: int):
     def open(self, option: Optional[Dict[str, str]] = None):
         """Open the output file / device and write the header.
 
+        :py:class:`StreamWriter` is also a context manager and therefore supports the
+        ``with`` statement.
+        It is recommended use context manager, as the file is closed automatically
+        when exiting from ``with`` clause.
+
```
hwangjeff, line 231: it isn't immediately clear how this relates to `open` — i suppose the connection is, since `open` returns `self`, which references `StreamWriter`, which is a context manager as described here, `open` can be used to create a context manager. perhaps this can be clarified

hwangjeff (approved):

PR #3067, opened 2023-02-15
Add CTC based forced alignment function

nateanl:


torchaudio/prototype/functional/functional.py
```diff
@@ -121,3 +122,131 @@ def barkscale_fbanks(
         )
 
     return fb
+
+
+def forced_align(
+    emission: torch.Tensor, label: torch.Tensor, blank_id: int = 0
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
```
mthrok, line 129: I think, instead of Tuple, we should define a micro protocol with dataclass.

PR #3073, opened 2023-02-16
Add guards to prevent ffmpeg failures during dispatcher import

hwangjeff:
With the introduction of the backend dispatcher, importing torchaudio fails when ffmpeg is not available. This PR adds guards to resolve these failures.


mthrok (approved):

NivekT (approved):
Thanks for the fix!

PR #3077, opened 2023-02-17
Release only changes

weiwangmeta:
To make sure we test with pytorch test channel instead of default nightly in non-release status.


version.txt
```diff
@@ -1 +1 @@
-2.0.0a0
+2.0.0
```
osalpekar, line 1: curious what this change is about?

osalpekar (approved):

PR #3082, opened 2023-02-21
Use dlopen for sox

mthrok:
This commit changes the way torchaudio integrates libsox.
Previously, libsox has been statically linked and shipped.
This commit changes it so that libsox will be dl-opened at runtime.

Users now need to install libsox separately. Conda users can do so by
`conda install -c conda-forge sox libopus`. Ubuntu users can use `apt install libsox-fmt-all`

The notable difference is that the supported formats are different from 
one distribution to another, thus some formats will not be available unless users build libsox 
with these formats enabled.

If this change causes an unrecoverable issue, one can use `TORCHAUDIO_USE_SOX=0` to disable
libsox integration at runtime.


mthrok:
superseded by #3497, which is simpler

PR #3127, opened 2023-03-01
Fix unit test failures because old cuda 10.2 is found

atalman:
Similar to : https://github.com/pytorch/vision/pull/7332

This resolves unit tests infra issues.
- cu116 images should not be used anymore since they are deprecated. We should use cu117 images


nateanl (approved):

mthrok:
closing as we have migrated to GHA.

Issue #3101, opened 2023-02-26
Documentation error about torchaudio.transforms.Spectrogram


cai525:
### 📚 The doc issue

In document of [torchaudio.transforms.Spectrogram](https://pytorch.org/audio/main/generated/torchaudio.transforms.Spectrogram.html#torchaudio.transforms.Spectrogram), The description about `power` is:
 >power ([float](https://docs.python.org/3/library/functions.html#float) or None, optional) – Exponent for the magnitude spectrogram, (must be > 0) e.g., 1 for energy, 2 for power, etc. If None, then the complex spectrum is returned instead. (Default: 2)

But if I have the right understanding of the description , i.e., , parameter power is the exponent for the magnitude spectrogram, then 1 is suppoed to be for the norm of the result of FFT, which is the magnitude spectrum  but not for energe. Both the energe and power spectrum have the relationship of square with magnitude spectrum.

ref : https://en.wikipedia.org/wiki/Spectral_density

### Suggest a potential alternative/fix

For the description of https://pytorch.org/audio/main/generated/torchaudio.transforms.Spectrogram.html#torchaudio.transforms.Spectrogram,parameter `power`, it's supposed to be " 1 for  magnitude  , 2 for  energe/power".

cai525:
Here is the code about torchaudio.transforms.Spectrogram
```python
def spectrogram(
        waveform: Tensor,
        pad: int,
        window: Tensor,
        n_fft: int,
        hop_length: int,
        win_length: int,
        power: Optional[float],
        normalized: bool,
        center: bool = True,
        pad_mode: str = "reflect",
        onesided: bool = True
) -> Tensor:
    r"""Create a spectrogram or a batch of spectrograms from a raw audio signal.
    The spectrogram can be either magnitude-only or complex.

    Args:
        waveform (Tensor): Tensor of audio of dimension (..., time)
        pad (int): Two sided padding of signal
        window (Tensor): Window tensor that is applied/multiplied to each frame/window
        n_fft (int): Size of FFT
        hop_length (int): Length of hop between STFT windows
        win_length (int): Window size
        power (float or None): Exponent for the magnitude spectrogram,
            (must be > 0) e.g., 1 for energy, 2 for power, etc.
            If None, then the complex spectrum is returned instead.
        normalized (bool): Whether to normalize by magnitude after stft
        center (bool, optional): whether to pad :attr:`waveform` on both sides so
[...]

PR #3134, opened 2023-03-02
Fix documentation of functional and transforms

cai525:
Address #3101. The documentation for `power=1` should represent magnitude instead of energy.


cai525:
Thank you for your advice @nateanl . I have changed the `power` documentation in functional and examined other documentation 
 to ensure that related comments were corrected.

nateanl (approved):
LGTM

PR #3145, opened 2023-03-04
Fix Adam and AdamW initializers in wav2letter example

maciek97x:
In wav2letter example there is passed `momentum` to `Adam` and `AdamW` initializer, which is not a correct parameter. To fix that we need to add `beta_1` and `beta_2` to arguments and replace `momentum` with them. I also added `eps` similar to `Adadelta` initializer.


nateanl (approved):
Hi @maciek97x, thanks for contributing the fix. LGTM!

PR #3148, opened 2023-03-06
Guidance on docs versioning

carljparker:
I've received feedback that the versioning UI is both not obvious enough and also that it is not clear what it means. Adding some copy to the page to clarify.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -7,6 +7,11 @@ Torchaudio is a library for audio and signal processing with PyTorch.
 It provides I/O, signal and data processing functions, datasets,
 model implementations and application components.
 
+This documentation is versioned. Use the version control in the
+upper-left of this page to select the version that matches your
+Torchaudio build. The *nightly* build is the most recent--and
+unstable--release.
+
 ..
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -7,6 +7,11 @@ Torchaudio is a library for audio and signal processing with PyTorch.
 It provides I/O, signal and data processing functions, datasets,
 model implementations and application components.
 
+This documentation is versioned. Use the version control in the
+upper-left of this page to select the version that matches your
+Torchaudio build. The *nightly* build is the most recent--and
+unstable--release.
```
mthrok, line 13: Can we move this to the bottom of the page?
We'd like to keep information more relevant to frequent visitors on top and one-time information like this lower.

mthrok:
I am not sure the effectiveness of this approach.
I think what we should do is to either warn readers on all the pages like done in boost https://www.boost.org/doc/libs/1_79_0/
or old Python https://docs.python.org/3.6/library/index.html, or/and make the version so noticeable like currently supported Python docs does.  https://docs.python.org/3/library/index.html

PR #3163, opened 2023-03-10
Add AudioEffector

### LINKED ISSUE #3161: Apply filter function
With file-like object support in StreamReader/Writer, we can add feature to apply arbitral FFmpeg effect to Tensor.

https://docs.google.com/document/d/1qrCWKqc8mjs1lKCvihloJihJhf3KRYCDD0JPCNeJV1s/edit#

Starting code

```python
import torch
from io import BytesIO
import torchaudio
from torchaudio.io import StreamReader, StreamWriter

torchaudio.utils.ffmpeg_utils.set_log_level(56)

class StreamingEncoder:
    def __init__(self, src: torch.Tensor, sample_rate):
        self.src = src

        self.buffer = BytesIO()
        self.writer = StreamWriter(self.buffer, format="f32le")
        self.writer.add_audio_stream(num_channels=src.size(1), sample_rate=sample_rate)
        self.writer.open()

        self.i_read = 0
        self.i_iter = 0
        self.step = 256  # samples

    def read(self, n):
        # refill
        while self.i_iter < self.src.size(0) and self.buffer.tell() - self.i_read < n:
            chunk = self.src[self.i_iter:self.i_iter+self.step]
            self.writer.write_audio_chunk(0, chunk)
            self.i_iter += self.step

        i_write = self.buffer.tell()
        self.buffer.seek(self.i_read)
        ret = self.buffer.read(n)
        self.i_read += len(ret)
        self.buffer.seek(i_write)
        return ret


waveform, sample_rate = torchaudio.load("test/torchaudio_unittest/assets/sinewave.wav")
waveform = waveform.T.contiguous()

fileobj = StreamingEncoder(waveform, sample_rate)

s = StreamReader(fileobj, format="f32le")
s.add_audio_stream(-1, filter_desc="afftfilt=real='hypot(re,im)*sin(0)':imag='hypot(re,im)*cos(0)':win_size=512:overlap=0.75")
s.process_all_packets()
applied, = s.pop_chunks()

f, axes = plt.subplots(2, 1)
axes[0].specgram(waveform[:, 0], Fs=sample_rate)
axes[1].specgram(applied[:, 0], Fs=sample_rate)
plt.show()
```

Figure_1

mthrok:
This commit adds a new feature AudioEffector, which can be used to
apply various effects and codecs to waveforms in Tensor.

Under the hood it uses StreamWriter and StreamReader to apply
filters and encode/decode.

This is going to replace the deprecated `apply_codec` and 
`apply_sox_effect_tensor` functions.

It can also perform online, chunk-by-chunk filtering.

Tutorial to follow.

closes #3161

### DIFF
--- a/torchaudio/io/_effector.py
+++ b/torchaudio/io/_effector.py
@@ -1,0 +1,335 @@
+import io
+from typing import Iterator, List, Optional
+
+import torch
+from torch import Tensor
+
+from ._stream_reader import _get_afilter_desc, StreamReader
+from ._stream_writer import CodecConfig, StreamWriter
+
+
+class _StreamingIOBuffer:
+    """Streaming Bytes IO buffer. Data are dropped when read."""
+
+    def __init__(self):
+        self._buffer: List(bytes) = []
+
+    def write(self, b: bytes):
+        if b:
+            self._buffer.append(b)
+        return len(b)
+
+    def pop(self, n):
+        """Pop the oldest byte string. It does not necessary return the requested amount"""
+        if not self._buffer:
+            return b""
+        if len(self._buffer[0]) <= n:
+            return self._buffer.pop(0)
+        ret = self._buffer[0][:n]
+        self._buffer[0] = self._buffer[0][n:]
+        return ret
+
+
+def _get_sample_fmt(dtype: torch.dtype):
+    types = {
+        torch.uint8: "u8",
+        torch.int16: "s16",
+        torch.int32: "s32",
+        torch.float32: "flt",
+        torch.float64: "dbl",
+    }
+    if dtype not in types:
+        raise ValueError(f"Unsupported dtype is provided {dtype}. Supported dtypes are: {types.keys()}")
+    return types[dtype]
+
+
+class _AudioStreamingEncoder:
+    """Given a waveform, encode on-demand and return bytes"""
+
+    def __init__(
+        self,
+        src: Tensor,
+        sample_rate: int,
+        effect: str,
+        muxer: str,
+        encoder: Optional[str],
+        codec_config: Optional[CodecConfig],
+        frames_per_chunk: int,
+    ):
+        self.src = src
+        self.buffer = _StreamingIOBuffer()
+        self.writer = StreamWriter(self.buffer, format=muxer)
+        self.writer.add_audio_stream(
+            num_channels=src.size(1),
+            sample_rate=sample_rate,
+            format=_get_sample_fmt(src.dtype),
+            encoder=encoder,
+            filter_desc=effect,
+            codec_config=codec_config,
+        )
+        self.writer.open()
+        self.fpc = frames_per_chunk
+
+        # index on the input tensor (along time-axis)
+        # we use -1 to indicate that we finished iterating the tensor and
+        # the writer is closed.
+        self.i_iter = 0
+
+    def read(self, n):
+        while not self.buffer._buffer and self.i_iter >= 0:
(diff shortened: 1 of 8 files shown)


sorgfresser:
You're using `process_all_packets` from time to time. While I get that collecting all packets is necessary for some of the filters, it poses a threat since one can not tell how large the underlying file will be. Do you plan on adding an option to limit the amount of packets processed at once - even though the filters would obviously become worse? I would be very glad.

mthrok:
> You're using `process_all_packets` from time to time. While I get that collecting all packets is necessary for some of the filters, it poses a threat since one can not tell how large the underlying file will be. Do you plan on adding an option to limit the amount of packets processed at once - even though the filters would obviously become worse? I would be very glad.

@sorgfresser 

My implementations are aware of the potentially long audios, and they don't force the one-go operation. In this particular feature, I am providing two options. one-go or chunk-by-chunk process. It's up to users to pick which fashion they want to process their data.

torchaudio/io/_effector.py
```diff
@@ -0,0 +1,333 @@
...
+        (applied,) = reader.pop_chunks()
+        return Tensor(applied)
+
+    def stream(self, waveform: Tensor, sample_rate: int, frames_per_chunk: int) -> Iterator[Tensor]:
+        """Apply the effect and/or codecs to the given tensor chunk by chunk.
+
+        Args:
+            waveform (Tensor): The input waveform. Shape: ``(time, channel)``
+            sample_rate (int): Sample rate of the waveform.
+            frames_per_chunk (int): The unit number of frames to return at a time.
```
hwangjeff, line 317:
```suggestion
            frames_per_chunk (int): The number of frames to return at a time.
```

torchaudio/io/_effector.py
```diff
@@ -0,0 +1,333 @@
...
+        effect: Optional[str] = None,
+        format: Optional[str] = None,
+        *,
+        encoder: Optional[str] = None,
+        codec_config: Optional[CodecConfig] = None,
+        pad_end: bool = True,
+    ):
+        self.effect = effect
+        self.format = format
+        self.encoder = encoder
```
hwangjeff, line 256: perhaps we can check that `encoder` is not `None` only when `format` is not `None`

torchaudio/io/_effector.py
```diff
@@ -0,0 +1,333 @@
...
+    return types[dtype]
+
+
+class AudioEffector:
+    """Apply various filters and/or codecs to waveforms.
+
+    .. versionadded:: 2.1
+
+    Args:
+        effect (str or None): Filter expressions or ``None`` to apply no filter.
```
hwangjeff, line 133:
```suggestion
        effect (str or None, optional): Filter expressions or ``None`` to apply no filter.
```

torchaudio/io/_effector.py
```diff
@@ -0,0 +1,333 @@
...
+        if waveform.numel() == 0:
+            return waveform
+
+        reader = self._get_reader(waveform, sample_rate)
+        i = reader.get_out_stream_info(0)
+        reader.process_all_packets()
+        (applied,) = reader.pop_chunks()
+        return Tensor(applied)
+
+    def stream(self, waveform: Tensor, sample_rate: int, frames_per_chunk: int) -> Iterator[Tensor]:
```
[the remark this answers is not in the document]
mthrok: Yes, without a user provided sample rate, we need to assume some default value, which is not universally applicable.

torchaudio/io/_effector.py
mthrok: Good catch. They will be used blindly. Addressed in the new commit.

PR #3185, opened 2023-03-20
Use keyword arguments for librosa.filters.mel in HiFiGAN unit test

nateanl:
In librosa 0.10 release, positional arguments are deprecated (see https://github.com/librosa/librosa/pull/1521 for details). The PR fixes the HiFiGAN unit test by using keyword arguments for `librosa.filters.mel` function.


mthrok:
Can we also pin the version of librosa in unittests??

nateanl:
addressed in https://github.com/pytorch/audio/pull/3185/commits/12137ca03df1bc6e77002c38c77dab960ec2b959

PR #3189, opened 2023-03-21
Add SquimSubjective Model

nateanl:
Add model architecture and factory functions for `SquimSubjective` which predicts subjective evaluation metric scores (e.g. MOS) for speech enhancement task.

### DIFF
--- a/torchaudio/prototype/models/squim/subjective.py
+++ b/torchaudio/prototype/models/squim/subjective.py
@@ -68,7 +68,8 @@ class Predictor(nn.Module):
 
 class SquimSubjective(nn.Module):
     """Speech Quality and Intelligibility Measures (SQUIM) model that predicts **subjective** metric scores
-    for speech enhancement (e.g., MOS).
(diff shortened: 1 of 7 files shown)


nateanl:
cc @JupiterEthan

torchaudio/prototype/models/squim/subjective.py
```diff
@@ -0,0 +1,149 @@
...
+        x = self.att_pool_layer(x)
+        x = nn.functional.softmax(x, dim=1)
+        B = torch.linspace(0, 4, steps=self.att_dim, device=x.device)
+        x = (x * B).sum(dim=1)
+        return x
+
+
+class SquimSubjective(nn.Module):
+    """Speech Quality and Intelligibility Measures (SQUIM) model that predicts **subjective** metric scores
+    for speech enhancement (e.g., MOS).
```
mthrok, line 71: Is there a reference we should add??

mthrok (approved):

torchaudio/prototype/models/squim/subjective.py
nateanl: There is. I forgot to add it here. Will make a quick fix.

PR #3220, opened 2023-03-30
Introduce packet passthrough feature to streaming api

hwangjeff:
Summary: Introduces methods to StreamReader and StreamWriter that allow for reading and writing AVPacket instances rather than tensors. Useful for efficiently remuxing data pulled as is from source.

Differential Revision: D44271536

### DIFF
--- a/torchaudio/csrc/ffmpeg/stream_reader/packet_buffer.h
+++ b/torchaudio/csrc/ffmpeg/stream_reader/packet_buffer.h
@@ -1,0 +1,16 @@
+#pragma once
+#include <torchaudio/csrc/ffmpeg/ffmpeg.h>
+
+namespace torchaudio {
+namespace io {
+class PacketBuffer {
+ public:
+  void push_packet(AVPacket* packet);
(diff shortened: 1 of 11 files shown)


mthrok (commented):
I think that the packet buffer can be placed on `StreamReader` and keep `StreamProcessor` intact, and it's simpler that way.

torchaudio/csrc/ffmpeg/stream_reader/packet_buffer.h
```diff
@@ -0,0 +1,16 @@
+#pragma once
+#include <torchaudio/csrc/ffmpeg/ffmpeg.h>
+
+namespace torchaudio {
+namespace io {
+class PacketBuffer {
+ public:
+  void push_packet(AVPacket* packet);
+  c10::optional<std::vector<AVPacketPtr>> pop_packets();
+  bool is_ready();
```
mthrok, line 10: Can we call this method something like `has_value`?
Because IMO, readiness is not well-defined for packet buffer, because packet buffer is more of a temp location where packets are just put.

PR #2850, opened 2022-11-14
[WIP] Add Ray tracing method for RIR

NicolasHug:
This PR adds the `ray_tracing()` helper to compute a RIR (part of https://github.com/pytorch/audio/issues/2624). The implementation is heavily based on `pyroomacoustics`.

Current status: support for 2D and 3D shoebox room is OK, results are very very close to those of `pyroomacoustics`.

My C++ is weak / rusty, so I welcome any suggestion on style / perf / best  practices are more than welcome.

Lots of TODOs left:

- [ ] write docs
- [ ] make autograd work properly
- [ ] support multiple bands
- [ ] support per-wall and per-band absorption and scatter
- [ ] integrate this properly with the Image Source Method.
- [ ] bunch of TODOs here and there in the code
- [ ] (maybe) parallelize code over `num_rays`

CC @nateanl

Issue #2624, opened 2022-08-16
Room Impulse Response Simulation Support in TorchAudio


nateanl:
For release 2.0, we plan to add support for multi-channel room impulse response simulation methods under `torchaudio.functional`. The implementation is based on [pyroomacoustics](https://github.com/LCAV/pyroomacoustics), that supports both "image source method", and "image source + ray tracing" (hybrid) method. We will support both modes in two separate methods.

## Diagram

Here is the diagram of how the code works:
<img width="756" alt="image" src="https://user-images.githubusercontent.com/8653221/184940958-c0b74798-baa4-4c34-9674-b82d276934e2.png">

Both methods compute image sources as the first step. The difference is, for pure image source method, only `absorption_coefficient` is used to estimate the attenuations for each order. while for the hybrid method, both `absorption_coefficient` and `scattering` are used, if `scattering` is provided by users. Then the image sources locations are applied to estimate impulse responses (IR) in `_build_rir` method. The hybrid method applies ray tracing to estimate IRs for late reverberations.

Besides the above two methods, we plan to support an Array Transfer Function (ATF) based simulation method, here is the diagram:
<img width="804" alt="image" src="https://user-images.githubusercontent.com/8653221/184942245-92c70293-15c4-4281-999d-ce02bd0f54e9.png">
The first few steps are the same as `simulate_rir_ism` or `simulate_rir_hybrid`, depending on the mode it selects. 

## API Design
The API of `simulate_rir_ism` will be like:
```python
simulate_rir_ism(
    room : Tensor,
    mic_array : Tensor,
    source : Tensor,
    sample_rate : int,
    max_order: int,
    wall_material : str = “”,
    ceiling_material : str = “”,
    floor_material : str = “”,
    air_absorption : bool = False,
    temperature: float = None,
    humidity: float = None,
) -> Tensor
```
where `room` is a 1D Tensor with `D` values that represents the room size, where D depend on whether the room is a 2D or 3D room.
`mic_array` is a 2D Tensor with dimensions `(channel, D)`, representing the coordinates of microphones in an array.
`source` is a 1D Tensor with `D` values which represents the coordinates of the sound source.
`sample_rate` is an integer to decide the sample rate of simulated RIRs.
`max_order` is the maximum order of wall reflections to save the computation in image source method.
`temperature` and `humidity` are parameters to compute the sound speed, by default the sound speed is 343 m/s.

[...]

PR #3234, opened 2023-04-03
Add Ray tracing method for RIR (#2850)

nateanl:
Summary:
This PR adds the `ray_tracing()` helper to compute a RIR (part of https://github.com/pytorch/audio/issues/2624). The implementation is heavily based on `pyroomacoustics`.

Pull Request resolved: https://github.com/pytorch/audio/pull/2850

Differential Revision: D41764237

Pulled By: nateanl


nateanl:
cc @fakufaku  @JupiterEthan ray_tracing function should be ready, would like to hear thoughts from you. Thanks!

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+  Wall(
+      const torch::Tensor& _absorption,
+      const torch::Tensor& _scattering,
+      const torch::Tensor& _normal,
+      const torch::Tensor& _origin)
+      : absorption(std::move(_absorption)),
+        reflection((scalar_t)1. - _absorption),
+        scattering(std::move(_scattering)),
+        normal(std::move(_normal)),
+        origin(std::move(_origin)) {}
```
mthrok, line 58: Applying `std::move` to `const` reference is strange.
You can probably remove `std::move` and copy construct them.

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+  torch::Tensor get_absorption() {
+    return absorption;
+  }
```
mthrok, line 62: seems not used.

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+    int64_t num_rays,
+    const torch::Tensor& absorption,
+    const torch::Tensor& scattering,
+    double mic_radius,
+    double sound_speed,
+    double energy_thres,
+    double time_thres,
+    double hist_bin_size) {
+  auto num_mics = mic_array.size(0);
+  auto num_bands = absorption.size(0);
```
mthrok, line 537: Cannot call `size(0)`, without first checking the `ndim >= 1`

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+        num_rays(_num_rays),
+        energy_0(2. / num_rays),
+        mic_radius(_mic_radius),
+        mic_radius_sq(mic_radius * mic_radius),
+        sound_speed(_sound_speed),
+        energy_thres(_energy_thres),
+        time_thres(_time_thres),
+        hist_bin_size(_hist_bin_size),
+        max_dist(VAL(room.norm()) + (scalar_t)1.),
+        D(room.size(0)),
```
mthrok, line 139: D is unbounded. Need validation

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -545,3 +546,288 @@ def test_simulate_rir_ism_multi_band(self, channel):
...
+            source=source,
+            mic_array=mic_array,
+            num_rays=num_rays,
+            absorption=absorption,
+            scattering=scattering,
+        )
+
+        assert hist.shape == expected_shape
+
+    def test_ray_tracing_input_errors(self):
```
mthrok, line 585: need more of these. currently the implementation does not reject the invalid shapes like empty tensors.

test/torchaudio_unittest/prototype/functional/functional_test_impl.py
```diff
@@ -545,3 +546,288 @@ def test_simulate_rir_ism_multi_band(self, channel):
...
+        hist = F.ray_tracing(
+            room=room_dim,
+            source=source,
+            mic_array=mic_array,
+            num_rays=num_rays,
+            absorption=absorption,
+            scattering=scattering,
+        )
+
+        assert hist.shape == expected_shape
```
mthrok, line 583: Please use `self.assertEqual`. (Buck does not report numbers with `assert`)

torchaudio/csrc/ray_tracing.cpp
mthrok: Rather, I'd make D a template parameter.

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+        scattering,
+        mic_radius,
+        sound_speed,
+        energy_thres,
+        time_thres,
+        hist_bin_size);
+    rt.compute_histograms(histograms);
+  });
+  histograms = histograms.transpose(
+      1, 2); // back into shape (num_mics, num_bands, num_bins)
```
mthrok, line 561: Can you move the comment to separate line? It's easier to read that way.

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+
+            log_hist(histograms, mic_idx, energy, travel_dist_at_mic);
+          }
+        }
+      }
+
+      travel_dist += hit_distance;
+      transmitted *= wall.get_reflection();
+
+      //   Let's shoot the scattered ray induced by the rebound on the wall
```
mthrok, line 376:
```suggestion
      // Let's shoot the scattered ray induced by the rebound on the wall
```

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+#include <torch/script.h>
+#include <torch/torch.h>
+#include <cmath>
+
+namespace torchaudio {
+namespace rir {
+namespace {
+
+const bool IS_HYBRID_SIM =
+    false; // TODO: remove this once hybrid method is supported
```
mthrok, line 36: please move the comment to separate line

torchaudio/csrc/ray_tracing.cpp
```diff
@@ -0,0 +1,577 @@
...
+        // when there is no intersection, we jump to the next plane
+        if ((hit_point_a[d[i]] <= -EPS) ||
+            (room_a[d[i]] + EPS <= hit_point_a[d[i]])) {
+          intersection_found = false;
+          break; // check next plane
+        }
+      }
+
+      if (intersection_found) {
+        next_wall_index = 2 * d[0] + ind_inc;
```
mthrok, line 265: @nateanl I feel like this should be `d[0] + 2 * ind_inc`. Do we have test to verify this is correct?

mthrok: @nateanl Actually, this is correct for 3D but not for 2D. Because for 2D the order of wall is SENW, while for 3D it's WESNCF.

PR #3236, opened 2023-04-04
Remove source for flashlight-text bundle

mthrok:
Following https://github.com/pytorch/audio/pull/3232, static build of flashlight-text has been disabled and removed from nightly build.

This commit removes the related source/build from torchaudio code base.


jacobkahn:
🥳 

Happy spring cleaning!

Issue #877, opened 2020-08-12
Add Stereo to Mono Convertions


RemonComputer:
## 🚀 Feature

Make mono to stereo or stereo to mono conversion

## Motivation

You Guys have made an amazing job, but stereo to mono and vice versa is simple, it seems that you missed it,
I think it might be done with simple mean in the channel dim

## Pitch

This should be a Simple Transform like ToMono(channel_first=True) which will be in Torchaudio.Transforms

RemonComputer:
Attached a small Implementation of Mine
[transformations.py.txt](https://github.com/pytorch/audio/files/5069233/transformations.py.txt)

mthrok:
Hi @RemonComputer 

Thanks for the suggestion.

I think this problem is underspecified. I am not sure if there is one and only one way to convert stereo to monaural. Taking average is the simplest way, but adjusting the power level of each signal before mixing them is also popular approach. So I imagine there will be a wide variety of opinion what "stereo to monaural conversion" should do.

Adding this will increase the maintenance cost, but I do not see much value especially when it is so easy for users to write one. Users can just pick their favorite conversion algorithm write them with only a few lines of code.

RemonComputer:
@mthrok Ok, If I can help in anything please let me know

bagustris:
I am supporting this feature request.

Although the conversion from stereo to mono can be done in just a few lines of code and or using other libraries, having the same pattern would make the overall audio processing workflow easier and more consistent (e.g., the type and format of output [tensor vs. array, int64 vs float32], methods of converting [mean or first channel or second channel], etc).

mthrok:
By the popular demand, we are reconsidering this.
There are couple of improvements we can make

1. In documentation, explain how monaural conversion can be achieved.
2. Add transforms.

I am not sure what kind of mixing methods exist, but I guess we can simply start from taking average waveform across channel dimension.

CLI tools like `sox` and `ffmpeg` allows to pick one channel. Perhaps that can be an extension to this.

https://www.nesono.com/node/275
https://trac.ffmpeg.org/wiki/AudioChannelManipulation
https://en.wikipedia.org/wiki/Out_of_Phase_Stereo

mthrok:
cc @rbracco @hwangjeff @xiaohui-zhang

rbracco:
[...]

PR #3242, opened 2023-04-05
Add stereo to mono transform

jjmmchema:
Implements a transformation for converting multi-channel audio to mono audio. The conversion is done by just taking the average of each channel and then dividing it by the number of channels.

- [x]  Added ToMono class to audio/torchaudio/transforms/ _transforms.py. 
- [x]  Added to_mono method in audio/torchaudio/functional/functional.py
- [ ] Indicate dimensions of output tensor.
- [ ] Create tests cases.

Solves Issue: #877 

cc @mthrok

### DIFF
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -1009,6 +1009,25 @@ class Resample(torch.nn.Module):
             return waveform
         return _apply_sinc_resample_kernel(waveform, self.orig_freq, self.new_freq, self.gcd, self.kernel, self.width)
 
+class ToMono(torch.nn.Module):
+    r""" Converts a multi-channel signal into a mono 
+
+        .. devices:: CPU CUDA
+    """
+
+    def __init__(self) -> None:
+        super().__init__()
+
+
+    def forward(self, waveform: Tensor) -> Tensor:
+        r"""
+        Args:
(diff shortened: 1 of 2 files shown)


torchaudio/transforms/_transforms.py
```diff
@@ -722,7 +722,7 @@ def __init__(
     def forward(self, waveform: Tensor) -> Tensor:
         r"""
         Args:
-            waveform (Tensor): Tensor of audio of dimension (..., time).
+            waveform (Tensor): Tensor of the audio signal of dimension (..., time).
```
jjmmchema, line 725: Nevermind this change, it seems that I got confused and edited the docstring of the class that was right above the one I created.

test/torchaudio_unittest/transforms/transforms_test.py
```diff
@@ -252,6 +252,13 @@ def test_resample_size(self):
         # we expect the downsampled signal to have half as many samples
         self.assertTrue(down_sampled.size(-1) == waveform.size(-1) // 2)
 
+    def test_to_mono(self):
+        channel = 3
+        wave_channels_first = common_utils.get_whitenoise(n_channels=3)
+        wave_not_channels_first = common_utils.get_whitenoise(n_channels=3, channels_first=False)
+        
+        pass
```
jjmmchema, line 260: Not implemented

PR #3247, opened 2023-04-06
Adding device argument in Resample class

fzalkow:
The `torchaudio.transforms.Resample` class specifies that it is CUDA compatible, but it is not. See the following small example:

    import torch
    from torchaudio.transforms import Resample

    Fs_source = 22050
    Fs_target = 16000
    device = 'cuda'

    x = 2 * torch.pi * 440 * torch.arange(Fs_source, device=device) / Fs_source

    resample = Resample(Fs_source, Fs_target)
    x_new = resample(x)
    # RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same

To fix this error, I introduced a `device` argument to the init function of the class.

### DIFF
--- a/torchaudio/transforms/_transforms.py
+++ b/torchaudio/transforms/_transforms.py
@@ -956,6 +956,7 @@ class Resample(torch.nn.Module):
             cached as ``torch.float64``. If you use resample with lower precision, then instead of providing this
             providing this argument, please use ``Resample.to(dtype)``, so that the kernel generation is still
             carried out on ``torch.float64``.
+        device (torch.device): The location to allocate for PyTorch tensors. (Default: ``cpu``)
 
     Example
         >>> waveform, sample_rate = torchaudio.load("test.wav", normalize=True)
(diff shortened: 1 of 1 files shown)


mthrok:
> The `torchaudio.transforms.Resample` class specifies that it is CUDA compatible, but it is not. See the following small example:
> 
> ```
> import torch
> from torchaudio.transforms import Resample
> 
> Fs_source = 22050
> Fs_target = 16000
> device = 'cuda'
> 
> x = 2 * torch.pi * 440 * torch.arange(Fs_source, device=device) / Fs_source
> 
> resample = Resample(Fs_source, Fs_target)
> x_new = resample(x)
> # RuntimeError: Input type (torch.cuda.FloatTensor) and weight type (torch.FloatTensor) should be the same
> ```
> 
> To fix this error, I introduced a `device` argument to the init function of the class.

You need to move the module to CUDA to run resampling in GPU.

```
resample.to(torch.device("cuda"))
```

fzalkow:
> 
> You need to move the module to CUDA to run resampling in GPU.
> 
> ```
> resample.to(torch.device("cuda"))
> ```

Excellent, that works for me as well. I wasn't aware of the option to not only move tenors but entire modules to a device. Thanks! No need to merge this PR from my side anymore.

Issue #3049, opened 2023-02-09
Add a torch implementation of "convolution reverb" 


gwenzek:
### 🚀 The feature

A pure Pytorch implementation of the "convolution reverb" like described in https://pytorch.org/audio/stable/tutorials/audio_data_augmentation_tutorial.html#simulating-room-reverberation

This should be implemented like "pitch shift" both in "functional" and as a module.

### Motivation, pitch

torchaudio has been recently ramping up the utilities for data augmentation but so far the convolution reverb hasn't been implemented. Having a pure pytorch implementation would allow to run it efficiently on GPU.

### Alternatives

note that sox_effects can be used to provide a "reverb" but it's a CPU only implementation.
Note that sox is **not** doing a convolution reverb but [another algorithm](https://web.archive.org/web/20070222101813/http://www.harmony-central.com/Effects/Articles/Reverb/)

### Additional context

I'm a Meta employee and @roa-beep and @gziz will be working on it throught the MLH fellowship.

hwangjeff:
@gwenzek thanks for the feature request. This sounds good. Could you elaborate on how such an operator would differ from our existing convolution operators, e.g. [fftconvolve](https://pytorch.org/audio/main/generated/torchaudio.functional.fftconvolve.html#torchaudio.functional.fftconvolve) and [convolve](https://pytorch.org/audio/main/generated/torchaudio.functional.convolve.html#torchaudio.functional.convolve)?

Re: RIR samples — our preference would be to have the user provide one for the sake of maintainability. If there exists some RIR dataset that we'd like users to use, we can facilitate fetching samples by adding a corresponding class to [torchaudio.datasets](https://pytorch.org/audio/main/datasets.html).

mthrok:
I think we can provide couple of APIs.

1. The low level function that convolve the given waveform with given RIR.
2. A high level transform that convolve the given waveform with an RIR from predefined set.

And one can configure the RIR set to be used in the transform.
To configure the RIR set, defining the interface expected by the transform should be straightforward.
In very simple view, this is iterable with infinite length, which keeps returning the new RIR.
If we add RIR dataset class, then an instance of such class can be extended to be used by the said RIR transform.

mthrok:
I am also open to the idea of default (set of) RIR.
Semantically, it's hard to make sense of default RIR, (does it represent room with average acoustic characteristic?)
but maybe this could be the kind of thing, users are happy/okay with whatever default.

gziz:
@mthrok
[...]

PR #3266, opened 2023-04-12
Implement Convolutional Reverb

gziz:
SUMMARY
Implement Convolutional Reverberation.
This is a pure pytorch implementation, compared to the current torchaudio reverb which is based on sox_effects (this restricts the augmentation to run on CPU only).

The implementation was inspired by the following torchaudio tutorial:
https://pytorch.org/audio/stable/tutorials/audio_data_augmentation_tutorial.html#simulating-room-reverberation

From issue #3049


gwenzek:
Note for reviewer: 

The main difficulty of this feature is which RIR sample to chose, where to store it and how to "center" the RIR sample on the actual "clap" (so that the convolution isn't too expensive).

We chose to use a RIR sample which is already in torchaudio torch hub, but ideally we could had more and chose one of them randomly. 

For "centering" the RIR sample we currently slice it between 1s and 1.3s. This works well for this particular RIR, but may be wrong on others. Gerardo also worked on a small algorithm to automatically detect the clap and its reverb. Let us know if you want to add this in this PR.

mthrok:
Notes from offline discussions:

1. The current design of having one default RIR signal could encourage users to use the same RIR, even though the purpose of applying RIR is to increase the variety of data by applying different RIR, therefore, we are against having one default RIR.
2. As pointed out by @hwangjeff, the value of this function does not seem to out weight the cost of maintenance, at the current form.

I think supporting a set of RIR samples could be a way to add a unique value of this function.

mthrok:
closing as this is unlikely to be merged in its current form.

PR #3278, opened 2023-04-25
Add LRS3 AV-ASR recipe

mpc001:
This PR adds AV-ASR recipe which contains sample implementations of training and evaluation pipelines for RNNT based automatic, visual, and audio-visual (ASR, VSR, AV-ASR) models on LRS3. This repository includes both streaming/non-streaming modes.

CC @stavros99 @xiaohui-zhang @YumengTao @mthrok @nateanl @hwangjeff

### DIFF
--- a/examples/asr/avsr_rnnt/models/resnet1d.py
+++ b/examples/asr/avsr_rnnt/models/resnet1d.py
@@ -1,14 +1,6 @@
 import torch.nn as nn
 
 
-class Swish(nn.Module):
-    """Construct an Swish object."""
-
-    def forward(self, x):
-        """Return Swich activation function."""
-        return x * torch.sigmoid(x)
-
-
 def conv3x3(in_planes, out_planes, stride=1):
     """conv3x3.
     :param in_planes: int, number of channels in the input sequence.
@@ -76,8 +68,8 @@ class BasicBlock1D(nn.Module):
             self.relu1 = nn.PReLU(num_parameters=planes)
             self.relu2 = nn.PReLU(num_parameters=planes)
         elif relu_type == "swish":
-            self.relu1 = Swish()
-            self.relu2 = Swish()
+            self.relu1 = nn.SiLU(inplace=True)
+            self.relu2 = nn.SiLU(inplace=True)
         else:
             raise NotImplementedError
         # --------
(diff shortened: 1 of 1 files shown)


examples/asr/avsr_rnnt/models/resnet.py
```diff
@@ -0,0 +1,251 @@
+import math
+import torch.nn as nn
+import pdb
+import torch
+
+class Swish(nn.Module):
```
hwangjeff, line 6: pytorch already has this ([torch.nn.SiLU](https://pytorch.org/docs/stable/generated/torch.nn.SiLU.html))

examples/asr/avsr_rnnt/models/resnet.py
```diff
@@ -0,0 +1,251 @@
+import math
+import torch.nn as nn
+import pdb
```
hwangjeff, line 3: can remove unused imports

torchaudio/datasets/lrs3.py
```diff
@@ -0,0 +1,76 @@
+import os
```
xiaohui-zhang, line 1: @mpc001  can you open a separate PR for this file, with thorough documentation and units tests (refer to other examples in the dataset folder), or keep it in the current PR, move it to the example folder?

hwangjeff: @mpc001 yes, if you'd like to introduce LRS3 to the torchaudio library, we'd request excluding the file from this PR and opening a separate PR for it. We'd also request some additional changes, e.g. code mods, documentation, tests, since the API will be public and expected to be longer-lived. This might take some time to do though, so if you'd prefer, we can first keep the dataset in this PR but move it to the examples folder, and then in a future PR add the dataset to torchaudio and migrate this recipe to it.

hwangjeff (approved):

examples/asr/avsr_rnnt/models/resnet1d.py
```diff
@@ -0,0 +1,248 @@
+import torch.nn as nn
+
+
+class Swish(nn.Module):
+    """Construct an Swish object."""
+
+    def forward(self, x):
+        """Return Swich activation function."""
+        return x * torch.sigmoid(x)
```
hwangjeff, line 9: note that `torch` is undefined — need to add the corresponding import above

mthrok:
@mpc001 Could you update documentation index so that this tutorial is listed?

https://github.com/pytorch/audio/blob/main/docs/source/index.rst
(Search "Training Recipes")

PR #3288, opened 2023-05-01
[Nova] Add windows conda workflows

atalman:
[Nova] Add windows conda workflows
Same as: https://github.com/pytorch/vision/pull/7547


osalpekar:
cc @mthrok @hwangjeff

PR #3291, opened 2023-05-02
Remove top-level _static

mthrok:
Top level _static directory is not referenced by docs


mthrok:
will manually delete it once #3239 is landed

PR #3295, opened 2023-05-02
Improve RNN-T streaming decoding

lakshmi-speak:
This commit fixes the following issues affecting streaming decoding quality
1. The `init_b` hypothesis is only regenerated from blank token if no initial hypotheses are provided.
2. Allows the decoder to receive top-K hypothesis to continue decoding from, instead of using just the top hypothesis at each decoding step.  This dramatically affects decoding quality especially for speech with long pauses and disfluencies.
3. Some minor errors regarding shape checking for length.

This also means that the resulting output is the entire transcript up until that time step, instead of just the incremental change in transcript.

### DIFF
--- a/examples/tutorials/online_asr_tutorial.py
+++ b/examples/tutorials/online_asr_tutorial.py
@@ -39,6 +39,7 @@ to perform online speech recognition.
 # --------------
 #
 
+import os
 import torch
 import torchaudio
 
@@ -222,9 +223,9 @@ def run_inference(num_iter=100):
         segment = cacher(chunk[:, 0])
         features, length = feature_extractor(segment)
         hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
-        hypothesis = hypos[0]
-        transcript = token_processor(hypothesis[0], lstrip=False)
-        print(transcript, end="", flush=True)
+        hypothesis = hypos
+        transcript = token_processor(hypos[0][0], lstrip=False)
+        print(transcript, end="\r", flush=True)
 
         chunks.append(chunk)
         feats.append(features)
--- a/examples/asr/emformer_rnnt/pipeline_demo.py
+++ b/examples/asr/emformer_rnnt/pipeline_demo.py
@@ -65,9 +65,9 @@ def run_eval_streaming(args):
             with torch.no_grad():
                 features, length = streaming_feature_extractor(segment)
                 hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
-            hypothesis = hypos[0]
-            transcript = token_processor(hypothesis[0], lstrip=False)
-            print(transcript, end="", flush=True)
+            hypothesis = hypos
+            transcript = token_processor(hypos[0][0], lstrip=True)
+            print(transcript, end="\r", flush=True)
         print()
 
         # Non-streaming decode.
--- a/test/torchaudio_unittest/models/rnnt_decoder/rnnt_decoder_test_impl.py
+++ b/test/torchaudio_unittest/models/rnnt_decoder/rnnt_decoder_test_impl.py
@@ -99,7 +99,7 @@ class RNNTBeamSearchTestImpl(TestBaseMixin):
             self.assertEqual(res, scripted_res)
 
             state = res[1]
-            hypo = res[0][0]
+            hypo = res[0]
 
             scripted_state = scripted_res[1]
-            scripted_hypo = scripted_res[0][0]
+            scripted_hypo = scripted_res[0]
--- a/torchaudio/models/rnnt_decoder.py
+++ b/torchaudio/models/rnnt_decoder.py
@@ -109,13 +109,9 @@ class RNNTBeamSearch(torch.nn.Module):
 
         self.step_max_tokens = step_max_tokens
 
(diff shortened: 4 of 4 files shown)


mthrok:
Hi @lakshmi-speak 

Thanks for the PR. Do you have some information about the improvement before and after?
For example running this script gives better result?
https://github.com/pytorch/audio/blob/main/examples/asr/emformer_rnnt/pipeline_demo.py

(note: Perhaps we should update the demo script to compute WER so that this kind of comparison is easy)

hwangjeff:
Hi @lakshmi-speak, thanks for the PR. One observation here is that, although we're still processing the input waveform segment by segment, we're now storing token sequences in the hypotheses that grow unboundedly, and rather than outputting text in a streaming manner as before, we now output text only after all input segments have been processed. This behavior runs counter to what we want with streaming inference. Do you have any proposals around how to address these?

lakshmi-speak:
Actually, every time we process an input segment we do have the
transcription up until that time step. I only made the changes to the demo
script to print in the end so as to not have the output repeat itself a
bunch of times. So intermediate results are available.
Regarding the hypotheses growing, this is correct. I am not sure how big of
an issue that is - perhaps there can be a way to reset the hypotheses as a
user defined flag?
The hypotheses will grow to reflect the input audio, similar to
non-streaming use case. But it’s also just text tokens so perhaps not that
big of an issue?
Happy to hear your feedback on this.

hwangjeff:
@lakshmi-speak got it, thanks. I suppose the situations in which the hypotheses would grow beyond reason are fairly niche. For now, then, I think we can go with what you have here and make it the caller's responsibility to reset the hypotheses. As for printing the transcripts at the very end in the demo and tutorials, perhaps we can go back to printing after each iteration but with a carriage return appended, e.g. replace `print(transcript, end="", flush=True)` with `print(transcript, end="\r", flush=True)`, so that each subsequent transcript overwrites the previous transcript. This way, users can see the updates in real time.

lakshmi-speak:
Made the changes to the pipeline_demo.py script (will only refresh the current line, but I think that's ok for librispeech?)

mthrok:
> Regarding the hypotheses growing, this is correct. I am not sure how big of an issue that is - perhaps there can be a way to reset the hypotheses as a user defined flag? The hypotheses will grow to reflect the input audio, similar to non-streaming use case. But it’s also just text tokens so perhaps not that big of an issue?

Is the hypothesis the result of inference or is it something fed to models? Or alternatively, does the growing hypothesis have performance penalty on inference on later stage?

examples/tutorials/online_asr_tutorial.py
```diff
@@ -222,9 +223,10 @@ def run_inference(num_iter=100):
         segment = cacher(chunk[:, 0])
         features, length = feature_extractor(segment)
         hypos, state = decoder.infer(features, length, 10, state=state, hypothesis=hypothesis)
-        hypothesis = hypos[0]
-        transcript = token_processor(hypothesis[0], lstrip=False)
-        print(transcript, end="", flush=True)
+        hypothesis = hypos
+        transcript = token_processor(hypos[0][0], lstrip=False)
+        os.system('cls' if os.name == 'nt' else 'clear')
```
hwangjeff, line 228: is this necessary?

lakshmi-speak: This is because the transcript is long and single line refresh won't work.

lakshmi-speak:
> > Regarding the hypotheses growing, this is correct. I am not sure how big of an issue that is - perhaps there can be a way to reset the hypotheses as a user defined flag? The hypotheses will grow to reflect the input audio, similar to non-streaming use case. But it’s also just text tokens so perhaps not that big of an issue?
> 
> Is the hypothesis the result of inference or is it something fed to models? Or alternatively, does the growing hypothesis have performance penalty on inference on later stage?

There is not additional performance penalty because of the hypotheses growing in length, since the prediction and `hypo_predictor_out` are one-step. However, since we continue decoding at each step from top-k hypothesis, we do process k- batches of predictions in parallel at the first time-step per input frame.
This is to be expected and is the reason why we see superior performance in beam search decoding. Previously, we would use the just top-1 hypothesis at every step, even though top-k was being predicted.

examples/tutorials/online_asr_tutorial.py
mthrok: Can you show a screen shot of what issue happens due to long line?
I don't think system call belongs here.

lakshmi-speak: This example is a transcription of a very long audio, which spans multiple lines on the screen. Since the transcript now updates every step but keeps all the prior history, if we didn't clear the screen, we end up with repetitions of the same line. Using carriage return on print works, if we are refreshing the same line. Here we want to refresh several previous print lines.

hwangjeff: @lakshmi-speak it seems to work fine without the system call. can you remove it? the rest of the pr looks good — we can merge it after you make the change.

lakshmi-speak: @hwangjeff ready to merge!

hwangjeff:
@lakshmi-speak note that we've merged your PR — thanks for contributing to the library!

PR #3096, opened 2023-02-24
Add cuctc decoder

yuekaizhang:
This PR related to #2957. It implements a CUDA based ctc prefix beam search decoder. 

Attach serveral benchmark results using V100 below:
|decoder type| model |datasets       | decoding time (secs)| beam size | batch size | model unit | subsampling times | vocab size |
|--------------|---------|------|-----------------|------------|-------------|------------|-----------------------|------------|
| cuctc |  conformer nemo    |dev clean        |7.68s | 8           |  32       | bpe         |    4  | 1000|
| cuctc |  conformer nemo   |dev clean  (sort by length)      |1.6s | 8           |  32       | bpe         |    4  | 1000|
| cuctc |  wav2vec2.0 torchaudio |dev clean                                |22s | 10           |  1       | char         |    2  | 29|
| cuctc |   conformer espnet   |aishell1 test                             | 5s | 10           |  24       | char         |    4  | 4233|

Note: 
1.  The design is to parallel computation through batch and vocab axis, for loop the frames axis. So it's more friendly with smaller sequence lengths, larger vocab size comparing with CPU implementations.
2. WER is the same as CPU implementations. However, it can't decode with LM now.

TODO:
- [x] Add docs
- [x] Add unit test (I am not sure github actions work for GPU codes)
- [x] Fix lint error
- [x] Anything else?

What's your thought about this draft for now? @mthrok

PR #3297, opened 2023-05-03
add cuctc tutorial, change blank skip threshould into prob

yuekaizhang:
Add a separate tutorial for cuctc.
Reslove #3096


Adel-Moumen:
Hello @yuekaizhang, 

Many thanks for your CUDA Implementation. When it will be release in the pip install branch I will definitely make sure that we support it in SpeechBrain. 

In the meantime, can I ask you if you have any documentation that explain how you did your implementation ? This is something that could be relevant to the community to have a PyTorch version which does not involve custom CUDA kernels. 

Indeed, it is easier for people to directly modify PyTorch code than CUDA code when one wants to modify/implement research papers. Hence, I'm asking you if you could elaborate a little bit more on how we could re-implement what you did but with PyTorch directly. Many thanks!

Adel

yuekaizhang:
> 

Thanks. @mthrok Do you know when it will be available in pip install ?

We would have a detailed documentation about how the cuda kernels were implemented in one or two months later.

Unfortunately, it can't be reimplemented with pure pytorch. (Or you would get worse or similar performace with current cpu version decoder in torchaudio.) Maybe you could try openai triton for python programmers.

Adel-Moumen:
Ok good! 

Tbh I don't think Triton will be a good alternative to your implementation. The CPU kernel overheads hits really hard with Triton especially in this case where we have a for loop on the time steps. Btw did you tried CUDA Graph with your implementation? I guess it could improve the results by a lot since there is a for loop. 

And do you know if you are planning to integrate support for kenLM ? 

Thanks again.

Adel-Moumen:
Thanks for the answers! Do you have a timeline in mind of when your code will be available in the pip install branch please?

mthrok:
Hi folks

Sorry for the belated response. The next PyTorch release is scheduled at Oct 2023. https://dev-discuss.pytorch.org/t/pytorch-release-2-1-0/1271

The CUCTC code is accessible through nightly channel, so if you would like to try out, please refer to https://pytorch.org and install nightly build.

@yuekaizhang Sorry for not responding these days, I will review this soon.

PR #3307, opened 2023-05-04
Fix MKL issue on Intel mac build

mthrok:
* Remove MKL and NumPy from Conda build env
* Remove `caffe2::mkl` dependency from `torch_cpu`, which introduced unnecessary and undesided dependency on Intel mac.

TorchAudio does not use BLAS libraries directly, thus all the mentions to MKL should be removed from the codebase.
However, this was causing an issue on Intel mac. It turned out that `torch_cpu` target is pulling `caffe2::mkl` dependency, and the linker on macOS keeps library dependency even if no symbol from that library is used. This stray mkl dependency should be fixed on core side, but also we can modify the target temporarily and remove them.

Also we don't need NumPy on build/run time, so that is removed as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,6 +136,8 @@ if (MSVC)
 else()
   set(warning_flags -Wall)
 endif()
+message(STATUS CMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS}")
+message(STATUS TORCH_CXX_FLAGS="${TORCH_CXX_FLAGS}")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${warning_flags} ${TORCH_CXX_FLAGS}")
 
 find_program(CCACHE_PROGRAM ccache)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,8 +136,6 @@ if (MSVC)
 else()
   set(warning_flags -Wall)
 endif()
-message(STATUS CMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS}")
-message(STATUS TORCH_CXX_FLAGS="${TORCH_CXX_FLAGS}")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${warning_flags} ${TORCH_CXX_FLAGS}")
 
 find_program(CCACHE_PROGRAM ccache)
--- a/cmake/TorchAudioHelper.cmake
+++ b/cmake/TorchAudioHelper.cmake
@@ -1,7 +1,12 @@
 find_package(Torch REQUIRED)
 
-# Remove stray mkl dependency, which is embed in the resulting binary,
-# even though torchaudio does not use MKL.
+# Remove stray mkl dependency found in Intel mac.
+#
+# For Intel mac, torch_cpu has caffe2::mkl, which adds link flags like
+# -lmkl_intel_ilp64, -lmkl_core and -lmkl_intel_thread.
+# Even though TorchAudio does not call any of MKL functions directly,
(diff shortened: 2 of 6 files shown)


atalman:
Looks like it fixes MKL issue. However I see numpy warning, should we keep numpy ?:
```
+ conda run -p /Users/runner/work/_temp/pytorch_pkg_helpers_4889580093_smoke python3 pytorch/audio/test/smoke_test/smoke_test.py
/Users/runner/work/_temp/pytorch_pkg_helpers_4889580093_smoke/lib/python3.8/site-packages/torchaudio/compliance/kaldi.py:22: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /Users/runner/work/_temp/anaconda/conda-bld/pytorch_1683184149390/work/torch/csrc/utils/tensor_numpy.cpp:84.)
```

atalman (approved):

mthrok:
> Looks like it fixes MKL issue. However I see numpy warning, should we keep numpy ?:
> 
> ```
> + conda run -p /Users/runner/work/_temp/pytorch_pkg_helpers_4889580093_smoke python3 pytorch/audio/test/smoke_test/smoke_test.py
> /Users/runner/work/_temp/pytorch_pkg_helpers_4889580093_smoke/lib/python3.8/site-packages/torchaudio/compliance/kaldi.py:22: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /Users/runner/work/_temp/anaconda/conda-bld/pytorch_1683184149390/work/torch/csrc/utils/tensor_numpy.cpp:84.)
> ```

NumPy has been an optional dependency of PyTorch from very early time, 
but I think PyTorch's meta.yml has NumPy (at least [in nightly](https://github.com/pytorch/builder/blob/e0399f017be348625d2df621e627f64e2d956e4a/conda/pytorch-nightly/meta.yaml)), so it should be covered for regular users.

For torchaudio, it's also an optional dependency.
So it is more useful for us to test this way, so that if we do bare `import numpy` accidentally, it will be caught.
Plus, we don't need to worry about maintaining NumPy compatible versions in torchaudio's `meta.yml`.
We can rely on PyTorch's NumPy version constraint.

For a peace of mind I can move the removal of NumPy to a different PR if you prefer that way.

malfet (approved):
LGTM, but are you sure there are no workflow where TorchAudio would depend on NumPy? I see it's used conditionally in https://github.com/pytorch/audio/blob/1e48af06a95fe839110a24d53dd57216d7986906/torchaudio/kaldi_io.py#L40
and also pybind_numpy is included here, even though it looks unused:
https://github.com/pytorch/audio/blob/1e48af06a95fe839110a24d53dd57216d7986906/torchaudio/csrc/cuctc/src/python_binding.cpp#L26

If one can remove if from `python_binding.cpp`, then feel free to remove it from host, but keep it in  `run` section, though remove all version constraints.

mthrok:
> LGTM, but are you sure there are no workflow where TorchAudio would depend on NumPy? I see it's used conditionally in
> 
> https://github.com/pytorch/audio/blob/1e48af06a95fe839110a24d53dd57216d7986906/torchaudio/kaldi_io.py#L40
> 
> 
> and also pybind_numpy is included here, even though it looks unused:
> https://github.com/pytorch/audio/blob/1e48af06a95fe839110a24d53dd57216d7986906/torchaudio/csrc/cuctc/src/python_binding.cpp#L26
> 
> If one can remove if from `python_binding.cpp`, then feel free to remove it from host, but keep it in `run` section, though remove all version constraints.

Good point. pybind_numpy is a oversight. It is not used. Let me defer the removal of NumPy and follow up then.

Issue #3311, opened 2023-05-05
Failed to write packet (Invalid argument) caused by missing PTS


whoyao:
### 🐛 Describe the bug

Compared with version 2.0.1, the active branch made major changes to `StreamWriter` in [torchaudio/csrc/ffmpeg/stream_writer](https://github.com/pytorch/audio/tree/main/torchaudio/csrc/ffmpeg/stream_writer). When I used NVIDIA hardware encoders (`h264_nvenc`) to encode flv video streams, I encountered unexpected errors, causing the process to exit.

Snippet to reproduce the error is provided below. 
```python
import os
import shutil
import av
import numpy as np
from fractions import Fraction
import torch
import torchaudio
import torch.nn.functional as F
from torchaudio.io import StreamWriter, StreamReader

VIDEO_FPATH = 'rtmp://xxx.xxx.xxx.xxx/live/test'
VIDEO_URL = 'test.mp4'

video_config = dict(
    frame_rate=25,
    width=256,
    height=256,
    hw_accel='cuda:0',
    encoder='h264_nvenc',
    encoder_format='rgb0',
    encoder_option={'gpu': '0'},
)
stream = StreamWriter(VIDEO_FPATH, format='flv')
stream.add_video_stream(**video_config)
stream.add_audio_stream(sample_rate=44100, num_channels=2, encoder="aac")
stream_handler = stream.open()
streamer = StreamReader(VIDEO_URL)
streamer.add_basic_audio_stream(frames_per_chunk=44100,sample_rate=44100)
streamer.add_basic_video_stream(frames_per_chunk=25,frame_rate=25,width=256,height=256,format="rgb24")

for i, (audio_chunk, frame) in enumerate(streamer.stream()):
    if audio_chunk is not None:
        stream_handler.write_audio_chunk(1, audio_chunk)
    if len(frame.shape) == 3:
        frame = frame.unsqueeze(0)
    frame = F.pad(frame, (0, 0, 0, 0, 0, 1), 'constant', 0)
    stream_handler.write_video_chunk(0, frame.to('cuda:0'))
```
Output:
```
[h264_nvenc @ 0x6802840] Calling dl_fn->cuda_dl->cuCtxPopCurrent(&dummy)
[flv @ 0x68021c0] Timestamps are unset in a packet for stream 0. This is deprecated and will stop working in the future. Fix your code to set the timestamps properly
Traceback (most recent call last):
  File "/workspace/sources/audio-fix/main.py", line 75, in <module>
[...]

PR #3312, opened 2023-05-05
Fix missing PTS initialization with NVIDIA encoder

whoyao:
Fix **Failed to write packet (Invalid argument)** error when encoding FLV video streams using NVIDIA hardware encoders.

Resolve https://github.com/pytorch/audio/issues/3311


mthrok (approved):
Hi @whoyao 

Thanks for the the fix. LGTM.

Issue #3317, opened 2023-05-06
Changing video frame properties on the fly is not supported by all filters


whoyao:
### 🐛 Describe the bug

I used NVIDIA hardware encoders (`h264_nvenc`) in the project. Compared with v2.0.1, the latest code will perform the following validation on the input tensor:
```c
InitFunc init_func = [](const torch::Tensor& t, AVFrame* f) {
    validate_video_input(t, f, 4);
    return init_interlaced(t);
};
```
Compared with the old version, the accepted channel has changed from 3 to 4, which means that developers cannot input rgb24 format data and need to complete padding operations externally. 
However, after the conversion is completed, our input has become rgb0, and rgb0 is not a legal input. (`get_src_pix_fmt` only accepts `AV_PIX_FMT_GRAY8`, `AV_PIX_FMT_RGB24`, `AV_PIX_FMT_BGR24`, and `AV_PIX_FMT_YUV444P` as legal inputs) 
In order to pass the validation, I pretended that the input video format was rgb24, and the filter seemed to work. I dumped the structure of the filter. The filter was initialized with rgb24, but the actual input data was rgb0. 
The dump result of the filter is as follows:
```
+----------+
|    in    |default--[1024x1024 1:1 rgb24]--Parsed_null_0:default
| (buffer) |
+----------+

                                                     +--------------+
Parsed_null_0:default--[1024x1024 1:1 rgb24]--default|     out      |
                                                     | (buffersink) |
                                                     +--------------+

                                          +---------------+
in:default--[1024x1024 1:1 rgb24]--default| Parsed_null_0 |default--[1024x1024 1:1 rgb24]--out:default
                                          |    (null)     |
                                          +---------------+
```
It seems that the output video is fine, but if you open the ffmpeg log, you can see the following warnings:
```
[in @ 0x755f1600] filter context - w: 1024 h: 1024 fmt: 2, incoming frame - w: 1024 h: 1024 fmt: 119 pts_time: NOPTS
[in @ 0x755f1600] Changing video frame properties on the fly is not supported by all filters.
```
`fmt: 2` represents AV_PIX_FMT_RGB24, while `fmt: 119` represents AV_PIX_FMT_CUDA.  The true pix_fmt is AV_PIX_FMT_CUDA and was set in `configure_hw_accel`.

Therefore, I believe that this usage still has hidden problems. So I submit a PR to add AV_PIX_FMT_CUDA as a valid format.

Snippet to reproduce the error is provided below.
```python
import os
[...]

PR #3318, opened 2023-05-06
Fix warning of changing video frame properties on the fly 

whoyao:
Fix the following warnings when using the NVIDIA hardware encoder:
```
[in @ 0x755f1600] filter context - w: 1024 h: 1024 fmt: 2, incoming frame - w: 1024 h: 1024 fmt: 119 pts_time: NOPTS
[in @ 0x755f1600] Changing video frame properties on the fly is not supported by all filters.
```

Resolve issue https://github.com/pytorch/audio/issues/3317

### DIFF
--- a/torchaudio/csrc/ffmpeg/stream_writer/encode_process.cpp
+++ b/torchaudio/csrc/ffmpeg/stream_writer/encode_process.cpp
@@ -96,7 +96,9 @@ enum AVSampleFormat get_src_sample_fmt(const std::string& src) {
       ".");
 }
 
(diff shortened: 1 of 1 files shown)


whoyao:
Now, the filter can be set correctly by the following code:

```python
video_config = dict(
    format='cuda',
    frame_rate=25,
    width=256,
    height=256,
    hw_accel='cuda:0',
    encoder='h264_nvenc',
    encoder_format='rgb0',
    encoder_option={'gpu': '0'},
)
stream.add_video_stream(**video_config)
```

mthrok:
Fixed by #3426

PR #3323, opened 2023-05-09
[Nova] Linux CPU Unittests to Nova

osalpekar:
As discussed in the [Torchaudio Migration Proposal](https://docs.google.com/document/d/1PF8biwiGzsjzfEBM78mlLiRrkcsGsvuYkeqkI66Ym8A/edit), this PR moves the Linux CPU unittest job to Nova tooling. Note that this does not disable the existing CircleCI job at the moment.

Passing Job: https://github.com/pytorch/audio/actions/runs/4986115298/jobs/8926499354?pr=3323


osalpekar:
@mthrok Looks like installing kaldi in the testing environment is causing some build breakages (link above has a pointer to the cmake error). Would love your feedback on the rest of this PR - happy to disable/skip kaldi tests here or in a follow-up.

mthrok (approved):

PR #3333, opened 2023-05-15
Fix for breadcrumbs displaying "Old version (stable)" on Nightly build

carljparker:
Previously, `breadcrumbs.html` identified a nightly build version by the prefix "Nightly" which would normally be prepended to the version in `conf.py`. However, the version string is coming through without the "Nightly" prefix, so this change causes `breadcrumbs.html` to key on the substring "dev" instead.

The reason we aren't getting "Nightly" is apparently because the environment variable BUILD_VERSION is available, so `conf.py` is using the value of that env var instead of the version string imported from the `torchaudio` module itself, which actually appears to be incorrect; see below.

If I install torchaudio using

    conda install torchaudio -c pytorch-nightly

then `torchaudio.__version__` returns the incorrect version string:

    2.0.0.dev20230309

### DIFF
--- a/docs/source/_templates/breadcrumbs.html
+++ b/docs/source/_templates/breadcrumbs.html
@@ -42,7 +42,7 @@
           <li><a href="{{ doc.link|e }}">{{ doc.title }}</a> &gt;</li>
         {% endfor %}
       <li>{{ title }} &gt;</li>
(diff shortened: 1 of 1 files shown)


mthrok (approved):
Looks good. Thanks!

PR #3356, opened 2023-05-22
[audio] add CTC forced alignment API tutorial to torchaudio

xiaohui-zhang:
Summary: move the forced aligner tutorial to torchaudio, with some formatting changes

Differential Revision: D46060238

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -67,6 +67,7 @@ model implementations and application components.
    tutorials/asr_inference_with_ctc_decoder_tutorial
    tutorials/online_asr_tutorial
    tutorials/device_asr
+   tutorials/ctc_forced_alignment_api_tutorial
    tutorials/forced_alignment_tutorial
    tutorials/tacotron2_pipeline_tutorial
    tutorials/mvdr_tutorial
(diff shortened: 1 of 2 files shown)


docs/source/index.rst
```diff
@@ -40,6 +40,7 @@ model implementations and application components.
    :hidden:
 
    tutorials/audio_io_tutorial
+   tutorials/ctc_forced_alignment_api_tutorial
```
mthrok, line 43: Can you move it to Pipeline tutorial next to the existing FA tutorial, so that they are close together? Technically this tutorial might not be about pipeline, but the topic is more about FA task, so I think it's more appropriate there.

examples/tutorials/ctc_forced_aligment_api_tutorial.py
```diff
@@ -0,0 +1,683 @@
+"""
+CTC forced alignment API tutorial
+=================================
+
+**Author**: `Xiaohui Zhang <xiaohuizhang@meta.com>`__
+
+(Thanks to `Vineel Pratap <vineelkpratap@meta.com>`__ and `Zhaoheng
+Ni <zni@meta.com>`__ for working on the forced aligner API, and `Moto
+Hira <moto@meta.com>`__ for providing alignment visualization
```
mthrok, line 9: Can we put the acknowledgment at the bottom? Or can we just mention that this tutorial was brought with the help of those people?

Tutorials are for readers and for the readers, this is not really information that they want to see first.

examples/tutorials/ctc_forced_aligment_api_tutorial.py
```diff
@@ -0,0 +1,683 @@
...
+def normalize_uroman(text):
+    text = text.lower()
+    text = text.replace("’", "'")
+    text = re.sub("([^a-z' ])", " ", text)
+    text = re.sub(' +', ' ', text)
+    return text.strip()
+
+os.system("echo 'aber seit ich bei ihnen das brot hole brauch ich viel weniger schulze wandte sich ab die kinder taten ihm leid' > test.txt")
+os.system("git clone https://github.com/isi-nlp/uroman")
+os.system("uroman/bin/uroman.pl < test.txt > test_romanized.txt")
```
mthrok, line 617: We can't expect Perl interpreter is available always, so this needs to be replaced with something different.

mthrok (commented):
I think that multi-lingual parts are better decoupled with the API usage parts. 
The FA API usage and multi-lingual are tangential, and extracting multi-lingual will allow users to easier to experiment with multi-lingual FA.
(the API usage tutorial will be more for reading and multi-lingual tutorial will be more fore playing around through Google Colab.)

PR #3365, opened 2023-05-23
Update forced_align method to only support batch Tensors

nateanl:
Summary: Current design of forced_align accepts 2D Tensor for `log_probs` and 1D Tensor for `targets`. To make the API simple, the PR make changes to only support batch Tensors (3D Tensor for `log_probs` and 2D Tensor for `targets`).

Differential Revision: D46126226

### DIFF
--- a/torchaudio/csrc/forced_align/cpu/compute.cpp
+++ b/torchaudio/csrc/forced_align/cpu/compute.cpp
@@ -17,8 +17,10 @@ void forced_align_impl(
   const scalar_t kNegInfinity = -std::numeric_limits<scalar_t>::infinity();
   using target_t = typename std::
       conditional<target_scalar_type == torch::kInt, int, int64_t>::type;
-  const auto T = logProbs.size(0);
-  const auto L = targets.size(0);
+  const auto batchIndex =
+      0; // TODO: support batch version and use the real batch index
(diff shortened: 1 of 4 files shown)


torchaudio/csrc/forced_align/cpu/compute.cpp
```diff
@@ -142,30 +147,35 @@ std::tuple<torch::Tensor, torch::Tensor> compute(
...
-  TORCH_CHECK(inputLengths.dim() == 0, "input_lengths must be 0-D");
-  TORCH_CHECK(targetLengths.dim() == 0, "target_lengths must be 0-D");
+      logProbs.dim() == 3,
+      "log_probs must be 3-D (batch_size, input length, num classes)");
+  TORCH_CHECK(
+      targets.dim() == 2, "targets must be 2-D (batch_size, target length,)");
+  TORCH_CHECK(
+      inputLengths.dim() == 1, "input_lengths must be 1-D (batch_size,)");
+  TORCH_CHECK(
+      targetLengths.dim() == 1, "target_lengths must be 1-D (batch_size,)");
```
mthrok, line 163: `ndim` check has to happen before calling `size` method, otherwise it could fail with generic error message.

PR #3368, opened 2023-05-24
[BC-Breaking] Remove compute_kaldi_pitch

mthrok:
This commit removes compute_kaldi_pitch function and the underlying Kaldi integration from torchaudio.

Kaldi pitch function was added in a short period of time by integrating the original Kaldi implementation, instead of reimplementing it in PyTorch.

The Kaldi integration employed a hack which replaces the base vector/matrix implementation of Kaldi with PyTorch Tensor so that there is only one blas library within torchaudio.

Recently, we are making torchaudio more lean, and we don't see a wide adoption of kaldi_pitch feature, so we decided to remove them.

See some of the discussion #1269


vadimkantorov:
Also, could make sense to just create a GitHub feature request issue for some pitch extraction algos (which could be useful for TTS?), and later on add a new function reimplemented from the ground up without any kaldi-compat as a goal

mthrok:
> Also, could make sense to just create a GitHub feature request issue in torchaudio for some pitch extraction algos (which could be useful for TTS?), and later on add a new function reimplemented from the ground up without any kaldi-compat as a goal

Yeah, that makes sense, however I don't know much about what kind of pitch extraction ppl are using these days, so I don't know what are the good candidates. I also hear that people use NN-based pitch extraction. In that case, it requires more work to add such.

roedoejet:
I understand why this was done but I wish there was a deprecation warning. We were using this and upgrading to 2.1.0 broke our system as a result. I echo @vadimkantorov - it would be nice to have a replacement for this feature. Thanks for all the work on this great library.

Issue #3376, opened 2023-05-26
unit tests Windows GPU failing


atalman:
### 🐛 Describe the bug

Following test are failing:
https://app.circleci.com/pipelines/github/pytorch/audio/16491/workflows/e9d2d0de-56ba-42f8-804b-77bf26fa291f/jobs/1213093

unittest_windows_gpu_py3.8:
```
FAILED torchaudio_unittest\io\stream_reader_test.py::FilterGraphWithCudaAccel::test_scale_cuda_format - RuntimeError: Failed to create the filter from "scale_cuda=format=yuv444p" (Invalid argument.)
FAILED torchaudio_unittest\io\stream_reader_test.py::FilterGraphWithCudaAccel::test_sclae_cuda_change_size - RuntimeError: Failed to create the filter from "scale_cuda=iw/2:ih/2" (Invalid argument.)
FAILED torchaudio_unittest\models\wav2vec2\huggingface_intergration_test.py::TestHFIntegration::test_import_finetune_2_wav2vec2_large_lv60k - AssertionError: Tensor-likes are not close!

Mismatched elements: 49150 / 49152 (100.0%)
Greatest absolute difference: 8.202713966369629 at index (10, 0, 812) (up to 0.0001 allowed)
Greatest relative difference: 970.2109375 at index (12, 2, 562) (up to 0.001 allowed)
FAILED torchaudio_unittest\models\wav2vec2\huggingface_intergration_test.py::TestHFIntegration::test_import_finetune_3_wav2vec2_large_lv60k - AssertionError: Tensor-likes are not close!

Mismatched elements: 49150 / 49152 (100.0%)
Greatest absolute difference: 8.202713966369629 at index (10, 0, 812) (up to 0.0001 allowed)
Greatest relative difference: 970.2109375 at index (12, 2, 562) (up to 0.001 allowed)
FAILED torchaudio_unittest\models\wav2vec2\huggingface_intergration_test.py::TestHFIntegration::test_import_finetune_4_wav2vec2_large_lv60k - AssertionError: Tensor-likes are not close!

Mismatched elements: 49151 / 49152 (100.0%)
Greatest absolute difference: 8.277942657470703 at index (3, 1, 336) (up to 0.0001 allowed)
Greatest relative difference: 590.3825073242188 at index (12, 2, 377) (up to 0.001 allowed)
[...]

PR #3384, opened 2023-05-29
Disable failing GPU unit test

atalman:
Disable failing GPU unit test.
See associated issue: https://github.com/pytorch/audio/issues/3376


mthrok:
Thanks for looking into this.

Instead of disabling each of them, can we introduce a common decorator and environment variable so that when we want to check which one is temporary disabled, we can just flip the env var in CI and see which one fails?

https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/common_utils/case_utils.py#L280-L284

```
disabledInCI = _skipIf(
    "CI" in os.environ,
    reason="Tests are failing on CI consistently. Disabled while investigating.",
    key="TEMPORARY_DISABLED",
)
```

and 

`export TORCHAUDIO_TEST_ALLOW_SKIP_IF_ TEMPORARY_DISABLED=1`

This will skip those decorated with `disabledInCI` in CI, but will fail in local env.

PR #3391, opened 2023-05-30
Delete CCI Linux and MacOS Unittest Jobs

osalpekar:
Deprecates the Linux and MacOS Unittest jobs now that they've been running on Nova for over a week. 

Aside: There was also a stylecheck job that was dependent on the Linux Unittest job. I also put up https://github.com/pytorch/audio/pull/3390 to move that stylecheck job to Nova. I'm happy to reintroduce the CCI stylecheck job standalone in CCI if we want the Nova version to run on main for a week.


mthrok (approved):
LGTM. Though technically I think deletion is a better description than deprecation.

PR #3397, opened 2023-06-01
Fix apply_codec to use named file

mthrok:
Follow-up #3386 The intended change was to use path of temporary file, instead of file-like object

### DIFF
--- a/torchaudio/functional/functional.py
+++ b/torchaudio/functional/functional.py
@@ -1330,12 +1330,11 @@ def apply_codec(
         Tensor: Resulting Tensor.
         If ``channels_first=True``, it has `(channel, time)` else `(time, channel)`.
     """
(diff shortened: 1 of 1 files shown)


hwangjeff (approved):

vadimkantorov:
Would it be possible to introduce some sort of `backend=` argument for this function? If I understand, SoX is going to be gradually replaced by ffmpeg in torchaudio, so letting a way for being explicit is a good thing (same as specifying backend for torchaudio.load)

mthrok:
> If still SoX has useful effects/codecs, adding explicit `backend=` option to AudioEffector would be great

The syntax and semantics of those sox effect chain and FFmpeg's filter graph are different, so unifying them is not straightforward. I think users can create such wrapper that suit their requirement and constraints if necessary.

PR #3399, opened 2023-06-01
[Nova] Deleting Remaining CircleCI jobs

osalpekar:
This job completely deletes the CircleCI `config.yml`. Here is what was remaining in the config at the point of deletion:

Used Jobs:
* **Lint** - Now running on Nova - see https://github.com/pytorch/audio/actions/runs/5144082942 for an example run on the latest PR in trunk
* **CircleCI Consistency** - Not needed anymore now if there is no CCI config.

Unused Jobs:
* **build-ffmpeg-$OS** - For the build jobs, we are already building FFMPEG from source as part of the Nova workflows. 
* **download-third-parties** - This is caching. We currently do not have caching in Nova jobs, but @atalman is working on adding support for this as a future optimization.


mthrok (approved):
Yay!

PR #3400, opened 2023-06-02
[Test] Sox windows

mthrok:


mthrok:
The code should work if there is libsox dll, but it seems only static libraries are available for libsox. closing

PR #3409, opened 2023-06-06
[Nova] Remove unused files

atalman:
We are using Project Nova workflows now. These are not required.

Same as: https://github.com/pytorch/vision/pull/7656


mthrok (approved):

.github/workflows/build_docs.yml
```diff
@@ -109,10 +109,9 @@ jobs:
         # TODO: Enable NVDec/NVEnc
         conda install --quiet -y 'ffmpeg>=4.1' pkg-config
         pip --quiet install cmake>=3.18.0 ninja
-        cd packaging
-        . ./pkg_helpers.bash
-        setup_build_version
-        cd ../
```
mthrok, line 115: Can we append a date if BUILD_VERSION is not set?

Without it, it's difficult when the nightly doc was deployed last.

PR #3436, opened 2023-06-12
[SoX/Flac] disable xmms_plugin dependency

wrongbad:
This plugin pulls glib and gtk which breaks the build on some headless systems

Since the plugin is not actually used, it seems right to disable it

This change fixed the build on my system


mthrok (approved):
Looks good. Thanks!

PR #3465, opened 2023-07-10
Update package smoke test

mthrok:
1. Update smoke test script to change directory so that there is no `torchaudio` directory in CWD when smoke test is being executed.
2. Disable the part of smoke test which requires FFmpeg for wheel. The preparation for https://github.com/pytorch/test-infra/pull/4358


.github/workflows/build_wheels_macos.yml
```diff
@@ -36,7 +36,7 @@ jobs:
             cache-key: macos-ffmpeg
             pre-script: packaging/pre_build_script.sh
             post-script: packaging/post_build_script.sh
-            smoke-test-script: test/smoke_test/smoke_test.py
+            smoke-test-script: test/smoke_test/smoke_test_no_ffmpeg.py
```
atalman, line 39: Should this be ``test/smoke_test/smoke_test.py --no-ffmpeg`` instead?

atalman (changes requested):
Please see comment

.github/workflows/build_wheels_macos.yml
mthrok: It seems that you cannot pass an argument

https://github.com/pytorch/test-infra/blob/main/.github/workflows/build_conda_linux.yml#L176

atalman (approved):

PR #3473, opened 2023-07-12
Fix resampling to support dynamic input lengths for onnx exports.

bogdanteleaga:
This is a port of https://github.com/adefossez/julius/pull/17 for torchaudio.

Not sure if it's possible/desirable to add tests to test the functionality of ONNX exports, but I did a quick test on my machine to ensure this works. The logic is a bit simpler compared to the other PR because the torchaudio version does not support the additional flags available in julius.


mthrok:
Hi @bogdanteleaga 

Thanks for the PR. The change looks good to me.

Regard the testing and ONNX support guarantee, in general, we do not make any promise about ONNX compatibility in torchaudio.

I do not anticipate to change resample code anytime soon,  but if you can add a test, then we will try out best to not break it.

mthrok (approved):

bogdanteleaga:
@mthrok 
I don't mind leaving it as is. Adding a test means onnxruntime would have to become a test-time dependency as well so it might break some additional workflows. Let me know if you'd like a test though and I can look into it.

PR #3477, opened 2023-07-12
Linux CPU job should respect set Python version

osalpekar:
Reintroduce a conda environment within which we will do all deps installation, audio builds, and tests runs. This conda environment will use the python version set by the GHA job - previously this just defaulted to using the system 3.10 python which was default inside the container.

### DIFF
--- a/.github/scripts/unittest-linux/install.sh
+++ b/.github/scripts/unittest-linux/install.sh
@@ -19,6 +19,9 @@ case "$(uname -s)" in
         eval "$("/opt/conda/bin/conda" shell.bash hook)"
 esac
 
+conda create -n ci python=${PYTHON_VERSION}
+conda activate ci
+
 # 1. Install PyTorch
 if [ -z "${CUDA_VERSION:-}" ] ; then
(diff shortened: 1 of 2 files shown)


mthrok:
Looks good on Linux side 

https://github.com/pytorch/audio/actions/runs/5536182084/jobs/10103496032?pr=3477#step:10:6890
https://github.com/pytorch/audio/actions/runs/5536182084/jobs/10103496282?pr=3477#step:10:6831
https://github.com/pytorch/audio/actions/runs/5536182084/jobs/10103495845?pr=3477#step:10:6445

However, now the macOS job seems to be failing due to this change...

https://github.com/pytorch/audio/actions/runs/5536182078/jobs/10103495640

Do you have an idea why? I wonder if macOS image has conda installed in a different location than this

```eval "$(/opt/conda/bin/conda shell.bash hook)"```
https://github.com/pytorch/audio/blob/main/.github/scripts/unittest-linux/run_test.sh#L5C1-L6C1

mthrok (approved):
Beautiful. Thank you!

PR #3491, opened 2023-07-24
Add pretrained VGGish inference pipeline

hwangjeff:
Adds pre-trained VGGish inference pipeline ported from https://github.com/harritaylor/torchvggish and https://github.com/tensorflow/models/tree/master/research/audioset.

### DIFF
--- a/torchaudio/prototype/pipelines/_vggish/_vggish_impl.py
+++ b/torchaudio/prototype/pipelines/_vggish/_vggish_impl.py
@@ -1,0 +1,233 @@
+# Derived from torchvggish (https://github.com/harritaylor/torchvggish).
+# Copyright 2017 The TensorFlow Authors All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+
+import math
+
+import torch
+
+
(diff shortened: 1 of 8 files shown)


torchaudio/pipelines/_vggish/_vggish_pipeline.py
```diff
@@ -0,0 +1,25 @@
...
+        state_dict = torch.load(path)
+        model.load_state_dict(state_dict)
+        model.eval()
+        return model
+
+    def get_input_processor(self):
+        return VGGishInputProcessor()
+
+
+VGGISH = VGGishBundle("models/vggish.pt")
```
mthrok, line 25: Could you add the docstring?

torchaudio/prototype/pipelines/_vggish/_vggish_impl.py
```diff
@@ -0,0 +1,262 @@
...
+        """
+        x = self.features_network(input)
+
+        x = x.permute(0, 2, 3, 1)
+        x = x.reshape(x.size(0), -1)
+
+        return self.embedding_network(x)
+
+
+class VGGishInputProcessor:
```
mthrok, line 108: This class is completely stateless. None of the method really require `self`. So I think it is better to extract them as plain functions.

torchaudio/prototype/pipelines/_vggish/_vggish_impl.py
```diff
@@ -0,0 +1,262 @@
...
+                **kwargs,
+            ).to(spectrogram),
+        )
+        return torch.log(mel_spectrogram + log_offset)
+
+    def _waveform_to_examples(self, data, sample_rate):
+        # Input: (T,) or (C, T)
+        # Convert to mono.
+        if len(data.shape) > 1:
+            data = torch.mean(data, axis=0)  # (T,)
```
mthrok, line 218: Stereo to mono conversion must be done carefully and taking mean is anti-pattern. See https://github.com/pytorch/audio/issues/877 for some discussion. We should land https://github.com/pytorch/audio/pull/3242, yet I am not sure if it is usable.

I suggest to remove stereo to mono conversion here and require user codes to convert their waveforms before passing them to the model.

PR #3531, opened 2023-08-04
Revise VGGish pipeline to accept arbitrary state dict function

hwangjeff:
Summary: Revises VGGish pipeline to accept arbitrary state dict function to accommodate loading weights from any source.

Differential Revision: D48056390


mthrok:
Just to double check, the purpose is to allow bundle instances to fetch data from places other than download.pytorch.org, right?

mthrok (approved):

hwangjeff:
> Just to double check, the purpose is to allow bundle instances to fetch data from places other than download.pytorch.org, right?

right

Issue #3539, opened 2023-08-08
nightly torchaudio.load cannot find sox 


subramen:
### 🐛 Describe the bug

The nightly version (2.1.0) appears to not find sox. 

Repro on colab:

```python

## STABLE
print(torchaudio.__version__)
# 2.0.2+cu118

print(torchaudio.utils.sox_utils.list_read_formats())
# ['aifc', 'aiffc', 'aiff', 'aif', 'al', 'au', 'snd', 'avr', 'cdda', 'cdr', 'cvsd', 'cvs', 'cvu', 'dat', 'dvms', 'vms', 'f4', 'f32', 'f8', 'f64', 'gsrt', 'hcom', 'htk', 'ima', 'la', 'lu', 'maud', 'null', 'prc', 'raw', 's1', 's8', 'sb', 's2', 's16', 'sw', 's3', 's24', 's4', 's32', 'sl', 'sf', 'ircam', 'sln', 'smp', 'sndr', 'sndt', 'sox', 'sph', 'nist', '8svx', 'txw', 'u1', 'u8', 'ub', 'sou', 'fssd', 'u2', 'u16', 'uw', 'u3', 'u24', 'u4', 'u32', 'ul', 'voc', 'vox', 'wav', 'wavpcm', 'amb', 'wve', 'xa', 'amr-nb', 'anb', 'amr-wb', 'awb', 'flac', 'gsm', 'lpc10', 'lpc', 'opus', 'vorbis', 'ogg']

print(torchaudio.backend.get_audio_backend())
# 'sox_io'

-----

## NIGHTLY
print(torchaudio.__version__)
# 2.1.0.dev20230807+cu118

print(torchaudio.utils.sox_utils.list_read_formats())
# RuntimeError: list_read_formats requires sox extension which is not available. Please refer to the stacktrace above for how to resolve this

print(torchaudio.backend.get_audio_backend())
# UserWarning: get_audio_backend's return value is irrelevant when the I/O backend dispatcher is enabled.
# ValueError: Unknown backend
```

Loading a wav file also errors in the nightly:
```
>>> torchaudio.load('audio.wav', format='wav')
RuntimeError: Couldn't find appropriate backend to handle uri .assets/bird_audio.wav and format wav
```

### Versions

```
PyTorch version: 2.1.0.dev20230807+cu118
Is debug build: False
CUDA used to build PyTorch: 11.8
ROCM used to build PyTorch: N/A

[...]

PR #3543, opened 2023-08-08
Update backend.rst for sox availability

subramen:
#3539

### DIFF
--- a/docs/source/backend.rst
+++ b/docs/source/backend.rst
@@ -26,7 +26,7 @@ There are currently two implementations available.
 Availability
 ------------
 
-``"sox_io"`` backend requires C++ extension module, which is included in Linux/macOS binary distributions. This backend is not available on Windows.
+``"sox_io"`` backend requires C++ extension module. torchaudio<2.1.0 will include libsox in the Linux/macOS binary distributions, for later versions please install `libsox`. This backend is not available on Windows.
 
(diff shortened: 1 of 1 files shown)


docs/source/backend.rst
```diff
@@ -26,7 +26,7 @@ There are currently two implementations available.
 Availability
 ------------
 
-``"sox_io"`` backend requires C++ extension module, which is included in Linux/macOS binary distributions. This backend is not available on Windows.
+``"sox_io"`` backend requires C++ extension module. torchaudio<2.1.0 will include libsox in the Linux/macOS binary distributions, for later versions please install `libsox`. This backend is not available on Windows.
```
mthrok, line 29: backend module here refers to the part of our code that use libsox, so this statement still holds.
Can we keep this statement as-is and use `note` notation and redirect to installation ?

```suggestion
``"sox_io"`` backend requires C++ extension module, which is included in Linux/macOS binary distributions. This backend is not available on Windows.

.. note::
   Starting 2.1.0, torchaudio searches and links libsox at runtime. Please install libsox. See also `Optional Dependencies <./installation.html#optional-dependencies>`__.
```

mthrok (approved):

mthrok:
@subramen let me handle this in #3555. I was planning to update the doc around backend and related things before release, and I had some time to do so during the weekend.

PR #3576, opened 2023-08-25
Refactoring

tosemml:
This minor update aims to enhance the code's Pythonic nature and conciseness by employing set intersection.


mthrok (commented):
Hi 

Thanks for proposing the change, however, I don't think this is improving the readability, while it can potentially consume a way more memory at `set(transcript.split())`.

mthrok:
> Thank you for your comment. This change is intended to encourage the use of Python idioms, which is a commonly recommended practice.

I am not convinced that just using `set` makes the code more Pythonic. I'd reorganize the logic of the entire helper function, rather than focusing on one line.

PR #3590, opened 2023-09-01
Automate Release-Only Changes

osalpekar:
First Pass for automating release-only changes.

Tested this in audio to ensure it works.

### DIFF
--- a/packaging/cut_release.sh
+++ b/packaging/cut_release.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
-# Usage:
-# TEST_INFRA_BRANCH=release/2.1 RELEASE_BRANCH=release/2.1 RELEASE_VERSION=2.1.0 cut_release.sh
+# Usage (run from root of project):
+# TEST_INFRA_BRANCH=release/2.1 RELEASE_BRANCH=release/2.1 RELEASE_VERSION=2.1.0 packaging/cut_release.sh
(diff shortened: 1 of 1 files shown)


huydhn (approved):
LGTM!  Although I think this script, as it stands, doesn't need to be on a per-repo basis because replacing branch name is done everywhere.  But there might be more audio-specific steps later on I assume

osalpekar:
@huydhn Yeah I was wondering something similar. There's nothing inherently audio-specific here, and with how the nova workflows are structured, I don't think there should be anything in the future. We can add such a script in test-infra, but might need to make some assumptions about the domain library setup. On the other hand, we could just make it a GHA, but might be overkill. Decided to just have this in each repo for now, but open to consolidation ideas.

mthrok:
@osalpekar @huydhn We ususally have release-specific changes, which is removing prototypes https://github.com/pytorch/audio/pull/3592. It would be nice that it can be incorporated.

osalpekar:
@mthrok Sure, let me look into doing that. I'm also wondering if there's something safer we could for for handling the prototypes in the release branch - we could make the prototype directory hidden (`_prototypes`) signaling the underlying components are not to be imported, and we could set a flag on the prototype unittests so that they are not run on the release-branch (or just have a separate CI job for them that doesn't run on the release branch). Let me know what you think!

PR #3599, opened 2023-09-07
Fix type casting issue in mask length calculation

mpc001:
Reference in fairseq: https://github.com/facebookresearch/fairseq/blob/main/fairseq/data/data_utils.py#L489


nateanl (approved):
lgtm, thanks for the fix

PR #3637, opened 2023-10-05
Fix breadcrumbs for v2.1

carljparker:
Change the stable version specified in `docs/source/conf.py` so that breadcrumbs display correctly on TorchAudio documentation pages. I tested on the nightly build using the docs build preview, and tested on the release by building locally on my MBP. See attached screenshots.
Screenshot 2023-10-05 at 4 47 25 PM
Screenshot 2023-10-05 at 4 45 13 PM

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -201,7 +201,7 @@ else:
 #
 # https://stackoverflow.com/a/33845358/1106930
 #
-html_context = {"version_stable": "2.1.0"}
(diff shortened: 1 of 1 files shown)


mthrok:
@carljparker Thanks for looking into this. I think this need to be documented as release process. To update the 2.1 stable doc do we need to cherry-pick this and rebuild the doc on release branch??

mthrok (approved):

PR #3662, opened 2023-10-17
Support compression level in i/o dispatcher backend

hwangjeff:
Differential Revision: D50367721


src/torchaudio/_backend/ffmpeg.py
```diff
@@ -304,7 +306,13 @@ def save(
         encoding: Optional[str] = None,
         bits_per_sample: Optional[int] = None,
         buffer_size: int = 4096,
+        compression: Optional[Union[torchaudio.io.CodecConfig, int]] = None,
```
mot0, line 309: I think that the compression type for sox should be float.

https://sox.sourceforge.net/soxformat.html

> In order to squeeze the selection of VBR into the the −C value float we use negative numbers to select VRR. -4.2 would select default VBR encoding (size) with high quality (speed). One special case is 0, which is a valid VBR encoding parameter but not a valid bitrate. Compression value of 0 is always treated as a high quality vbr, as a result both -0.2 and 0.2 are treated as highest quality VBR (size) and high quality (speed).

mthrok (approved):

PR #3702, opened 2023-11-15
pytorch/audio/ import cleanup

azad-meta:
Differential Revision: D51316163


test/smoke_test/smoke_test.py
```diff
@@ -5,7 +5,6 @@
 
 
 def base_smoke_test():
-    import torchaudio  # noqa: F401
```
mthrok, line 8: Please do not change this. This script is for OSS testing, and this line is checking if the extension is properly configured.

mthrok (changes requested):

PR #3774, opened 2024-04-17
Update stable symlink to 2.3.0.

ahmadsharif1:
This is in preparation for release 2.3.0.


stable
```diff
@@ -1 +1 @@
-2.2
\ No newline at end of file
+2.3.0
```
NicolasHug, line 1: For consistency with previous releases we should create a `2.3` symlink to `2.3.0`, and let this `stable` symlink point to `2.3`.

NicolasHug (approved):
Thanks @ahmadsharif1 , LGTM

PR #3787, opened 2024-05-11
Reduce memory usage of forced alignment on CPU

MahmoudAshraf97:
In the forced alignment c++ code, `backPtr` is an `int8` tensor while only storing the values 0,1, and 2 which can be effectively stored using only 2 bits instead of 8, and since the `backPtr` tensor size is `log_probs_len * (targets_length * 2 + 1)`, it can grow to unmanageable sizes in audio files that exceed 2 hours.
By using two `std::vector<bool>` to represent the two bits needed for `backPtr` we guarantee that the results are exactly the same while lowering memory usage since `std::vector<bool>` should use 1 bit to represent a boolean.
Best case scenario is memory usage drops to 25%, worst case scenario memory usage stays the same if a boolean is represented using 1 byte.

### DIFF
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -117,7 +117,8 @@ void forced_align_impl(
   auto ltrIdx = alphas_a[idx1][S - 1]; 
   // path stores the token index for each time step after force alignment.
   for (auto t = T - 1; t > -1; t--) {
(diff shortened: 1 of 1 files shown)


vadimkantorov:
I used this compression trick in my Python CTC-based forced alignment: https://github.com/vadimkantorov/ctc/blob/master/ctc.py

And this compression indeed works

MahmoudAshraf97:
since `torchaudio` is no longer maintained, I've created a library that builds upon this code and improves it [here](https://github.com/MahmoudAshraf97/ctc-forced-aligner)

PR #3803, opened 2024-06-26
Fix model downloading in bento

nateanl:
Summary: The model checkpoint path can not be created for Squim models. Use the latest download_asset method to fix it.

Differential Revision: D59061348


mthrok (approved):
Can you add a description that this is BC-breaking change (`dl_kwargs` removed), so that it information can be found easily when crafting the release note?

Issue #588, opened 2020-04-26
Constant Q-Transform 


vincentqb:
We would like to have in torchaudio 
- [ ] constant q-transform, as [librosa](https://librosa.github.io/librosa/generated/librosa.core.cqt.html)
- [ ] inverse constant q-transform, as [librosa](https://librosa.github.io/librosa/generated/librosa.core.griffinlim_cqt.html)

dhgrs:
Hi, I'm interested in implementing CQT and I have questions about it.

- After LibROSA 0.8, the links in your post are outdated. Maybe [cqt](https://librosa.org/doc/latest/generated/librosa.cqt.html) and [griffinlim_cqt](https://librosa.org/doc/latest/generated/librosa.griffinlim_cqt.html), right?
- LibROSA has some variants of CQT. If we test the code by comparing the results with LibROSA, these is a difficulty. Because `librosa.cqt` and `librosa.griffinlim` include kaiser_* resampling inside of it but torchaudio doesn't have them.
- How about focusing on `librosa.pseudo_cqt`? It doesn't include resampling but doesn't support inverse conversion.

KinWaiCheuk:
I have implemented the librosa CQT in my project, I hope it would be useful for you guys.
[https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/Spectrogram.py#L990](https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/Spectrogram.py#L990)

Several improvements can be made:
1. The `for` loop at line [1223](https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/Spectrogram.py#L1223) is looping over different octaves, however, is currently the bottleneck.
1.  The lowpass filter at [1032](https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/Spectrogram.py#L1032) is not as good as the librosa version
1. In librosa, they use a sparse matrix to store the frequency-domain CQT kernels. I did not use any sparse matrix in my implementation. Instead, I realized that it is not necessary to obtain the frequency-domain CQT kernels, the time-domain CQT kernels also work as well and even faster. This improved version has been implemented as another PyTorch class called [CQT2010v2](https://github.com/KinWaiCheuk/nnAudio/blob/master/Installation/nnAudio/Spectrogram.py#L1429) in my project.

[...]

PR #3804, opened 2024-06-27
CQT, iCQT, and VQT implementations and testing

d-dawg78:
...as requested in [issue 588](https://github.com/pytorch/audio/issues/588).

(More detailed explanations coming bery soon)


zaptrem:
Awesome contribution! A bunch of `torch.ones` tensors are initiated on CPU regardless of the input tensor's device. Also, it would be nice if there was an inverse VQT function as well. Also also, do you know a set of parameters that would result in a perfect or nearly perfect reconstruction? I had to fiddle with the filter lengths code to get something that was even close, but there's still an upper frequency buzzing sound and increased loudness at the start/end.

d-dawg78:
Hey, here's to addressing the feedback ☝️ 

1. Good catch on the `torch.ones` front - the most recent commit should address this issue.
2. We are following the `librosa` VQT, CQT, and iCQT algorithms, and [they opted not to implement the inverse VQT for good reason](https://github.com/librosa/librosa/issues/1161). I think we should do the same, at least for now.
3. Here are parameters that led to decent waveform reconstruction on my end:
```python
SAMPLE_RATE = 16000
HOP_LENGTH = 256
F_MIN = 32.703
N_BINS = 672
BINS_PER_OCTAVE = 96
```
Increasing the `N_BINS` and `BINS_PER_OCTAVE` accordingly increases CQT resolution, and by extension the reconstruction is much better 🙂 
4. I don't really have a good answer for this. Probably the result of the set of parameters you're using..?

zaptrem:
Why does this CQT's output have this surprising spiraly pattern whereas a normal STFT and CQT from [this other library](https://github.com/eloimoliner/CQT_pytorch) do not?
(Your CQT vs STFT)
image

(Other library's CQT)
image


My model requires a less/non periodic input a-la the normal STFT.

d-dawg78:
Hey,

Thank you for your patience - busy week! I managed to get decent reconstruction, without any audible artefacts, using the following parameters, and without any transformations to the original signal:
```python
SAMPLE_RATE = 44100
HOP_LENGTH = 256
F_MIN = 32.703
N_BINS = 1728
BINS_PER_OCTAVE = 192
```
There are two issues with the parameters you are using:
1. Your `f_min` should be mapped to [a note frequency](https://newt.phys.unsw.edu.au/jw/notes.html), so that your CQT bins are properly aligned with [the equal temperament tuning system](https://en.wikipedia.org/wiki/Equal_temperament).
2. Your `bins_per_octave` parameter should be a multiple of 12, so that your frequency bins align with [tones, semitones etc.](https://en.wikipedia.org/wiki/Interval_(music)).

Of course, feel free to play around with lower resolutions!

https://github.com/user-attachments/assets/1932c6c9-8c61-4661-8eb2-0f4a2e353e61

zaptrem:
> Hey,
> 
> Thank you for your patience - busy week! I managed to get decent reconstruction, without too many audible artefacts, using the following parameters, and without any transformations to the original signal:
> 
> ```python
> SAMPLE_RATE = 44100
> HOP_LENGTH = 256
> F_MIN = 32.703
> N_BINS = 1728
> BINS_PER_OCTAVE = 192
> ```
> 
> There are two issues with the parameters you are using:
> 
> 1. Your `f_min` should be mapped to [a note frequency](https://newt.phys.unsw.edu.au/jw/notes.html), so that your CQT bins are properly aligned with [the equal temperament tuning system](https://en.wikipedia.org/wiki/Equal_temperament).
> 2. Your `bins_per_octave` parameter should be a multiple of 12, so that your frequency bins align with [tones, semitones etc.](https://en.wikipedia.org/wiki/Interval_(music)).
> 
> Of course, feel free to play around with lower resolutions!
> 
>  icqt.mp4

Thanks! However, this has 1728 * 256 size, whereas a normal STFT can store a perfect strong COLA reconstruction in only 512 * 256 numbers (or 1024 * 512, etc). Shouldn't a CQT be able to do the same without significant artifacting?

PR #3811, opened 2024-07-15
Fix CUDA 12.5 build

HennerM:
CUDA 12.5 removed the FLT_MAX symbol.
This was previously used without being explicitly imported. FLT_MAX is defined in <float.h>, including this header fixes the issue. 

A similar issue has been reported in Kaldi before: https://github.com/kaldi-asr/kaldi/issues/4914


HennerM:
@atalman do you mind having a look at this? I can't build Torchaudio on CUDA 12.5 without this fix in, thanks

adamjstewart:
Looks like this was fixed in #3843 so this PR is no longer needed.

PR #3849, opened 2024-10-27
Make some buffers non-persistent (e.g. window, fbank...)

### LINKED ISSUE #3059: Use non-persistent buffers
### 🚀 The feature

Suggest using `register_buffer()` with `persistent=False`, so the buffer (e.g. window of spectrogram) will not be included in module's state dict.

### Motivation, pitch

When I add new transforms to my model and load a pre-existing checkpoint, a `missing key` error will be raised, e.g. missing resample kernel in `transforms.Resample`. It can be solved by specifying `strict=False` at load time, however, I don't see any reason to save the buffers. They can be recomputed on construction time and that won't affect model's behavior.

### Alternatives

_No response_

### Additional context

Same motivation as https://github.com/pytorch/pytorch/issues/18056.

gau-nernst:
Fixes #3059


stonepd:
These code changes look good to me, can someone from @team-audio-core please review and/or approve

PR #3855, opened 2024-11-21
fix: fixing batching in MMS_FA

ex3ndr:
Fixing batching in MMS_FA bundle


ytcheng:
Hi, I see that a pull request has already been submitted to fix this issue, but it seems to be stuck due to the Facebook CLA check. Is there any update on this? I’m waiting for this fix and would appreciate any information on the progress. Thanks!

PR #3874, opened 2025-01-29
Update Supported Python version 3.9-3.13

atalman:
Fix wheel metadata


kit1980 (approved):

kit1980:
Maybe just remove 3.8, but don't add 3.12 and 3.13, I don't think it's tested.

PR #3883, opened 2025-02-19
[ROCm][Windows] Enable build with ROCm on Windows

tvukovic-amd:
Added changes for Windows in LoadHIP.cmake:

- set default ROCM_PATH and HIP_PATH for Windows case
- update CMAKE_MODULE_PATH with appropriate HIP path - because the find_package call to this module uses the Module mode search
- added code for finding HIP version
- skipped find hsa-runtime64, rccl packages on Windows case
- skipped find rccl, roctx and roctracer libraries on Windows case

CMAKE_C_COMPILER and CMAKE_CXX_COMPILER redefinition on Windows case (if we are building with ROCm on Windows we need to use compiler specified with CXX and CC flags instead of cl)

### DIFF
--- a/cmake/LoadHIP.cmake
+++ b/cmake/LoadHIP.cmake
@@ -139,96 +139,213 @@ macro(find_package_and_print_version PACKAGE_NAME)
   message("${PACKAGE_NAME} VERSION: ${${PACKAGE_NAME}_VERSION}")
 endmacro()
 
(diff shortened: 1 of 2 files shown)


tvukovic-amd:
@mthrok This PR is for torchaudio build enablement with ROCm on Windows. Can you please review changes made here and inform me if they sound good to you?

PR #3884, opened 2025-02-19
Avoid crash when __doc__ is optimized away to None (-OO)

FremyCompany:
When the `python -OO` flag is used, doc strings are no longer stored, and __doc__ always returns None.

This makes the library fail at import time for a rather dubious reason.

It would be good to have this fixed.

### DIFF
--- a/src/torio/io/_streaming_media_decoder.py
+++ b/src/torio/io/_streaming_media_decoder.py
@@ -338,7 +338,8 @@ class ChunkTensor(ChunkTensorBase):
 
 def _format_doc(**kwargs):
     def decorator(obj):
-        obj.__doc__ = obj.__doc__.format(**kwargs)
+        if obj.__doc__ is not None:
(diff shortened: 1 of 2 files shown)


FremyCompany:
Also mentioned here:
https://discuss.pytorch.org/t/pr-torchaudio-incompatible-with-python-flag-oo-due-to-doc-being-none/216897/1

Note that the error message before the fix looks like this:
```
  File ".../.venv/lib/python3.12/site-packages/torio/io/_streaming_media_decoder.py", line 634, in StreamingMediaDecoder
    @_format_audio_args
     ^^^^^^^^^^^^^^^^^^
  File ".../.venv/lib/python3.12/site-packages/torio/io/_streaming_media_decoder.py", line 341, in decorator
    obj.__doc__ = obj.__doc__.format(**kwargs)
                  ^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'format'
```

PR #3913, opened 2025-05-20
Fix typos in backend dispatch UserWarning message

Tindbergg:
This PR corrects typos in the UserWarning message found in 3 backend modules:

This one requires review, I assume it is supposed to be per-call:
par-call → per-call

These are obvious:
bakcend → backend
udnerlying → underlying


DomHudson:
Thank you @Tindbergg !

PR #3923, opened 2025-06-07
Fix tests that failed in Python 3.13 when building the conda-forge package

aywi:
The following tests failed in Python 3.13 when building the conda-forge package (conda-forge/torchaudio-feedstock#30):

```
FAILED test/torchaudio_unittest/backend/dispatcher/soundfile/info_test.py::TestInfo::test_unknown_subtype_warning - assert 2 == 1
 +  where 2 = len([<warnings.WarningMessage object at 0x1712d0cb0>, <warnings.WarningMessage object at 0x1712d0d40>])
```
This was caused by `FutureWarning: functools.partial will be a method descriptor in future Python versions; wrap it in staticmethod() if you want to preserve the old behavior` in Python 3.13.

```
FAILED test/torchaudio_unittest/datasets/cmuarctic_test.py::TestCMUARCTIC::test_cmuarctic_path - ValueError: bad delimiter value
FAILED test/torchaudio_unittest/datasets/cmuarctic_test.py::TestCMUARCTIC::test_cmuarctic_str - ValueError: bad delimiter value
```
These were caused by `csv.reader(text, delimiter="\n")` in `torchaudio/datasets/cmuarctic.py`, which is not allowed in Python 3.13.

### DIFF
--- a/examples/asr/emformer_rnnt/librispeech/lightning.py
+++ b/examples/asr/emformer_rnnt/librispeech/lightning.py
@@ -79,20 +79,20 @@ class LibriSpeechRNNTModule(LightningModule):
         self.train_data_pipeline = torch.nn.Sequential(
             FunctionalModule(piecewise_linear_log),
             GlobalStatsNormalization(global_stats_path),
-            FunctionalModule(partial(torch.transpose, dim0=1, dim1=2)),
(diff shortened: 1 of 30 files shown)


aywi:
@NicolasHug @samanklesaria Please help take a look since there has been no response from the original code owners. Thanks!

Issue #3902, opened 2025-04-24
Update on TorchAudio’s future


scotts:
Dear TorchAudio users,

TorchAudio is the most popular audio library for PyTorch. It has critical transforms, models and datasets that we know the community relies on. That is why we wanted to let the community know that we have started a refactoring effort to transition TorchAudio into a **maintenance phase**. This process will involve **removal of user-facing features**. We have three goals we want to achieve with this effort:

1. **Make TorchAudio easier to maintain to ensure long-term reliability.** We plan to eliminate all C++ code so that TorchAudio is a Python-only library. We also plan to reduce external dependencies as much as possible. Both efforts will simplify testing and release.
2. **Reduce redundancies with the rest of the PyTorch ecosystem.** Some of the functionality in TorchAudio is also available in TorchVision and TorchCodec. We are working across all three libraries to ensure a given capability lives in one library.
3. **Focus on TorchAudio’s strengths.** Those strengths are the audio transforms, models and datasets that are integral to users training and inference pipelines. As a result, we will deprecate and eventually remove some functionality that is outside of these strengths.  

The diagram below depicts the various components of TorchAudio. We have highlighted it according to the user-facing API changes that we are making:

<img width="735" alt="Image" src="https://github.com/user-attachments/assets/ecf01218-77fb-49c7-96bf-55e582111954" />

Starting with TorchAudio 2.8 (expected around August 2025), APIs slated for removal will trigger a deprecation warning. These APIs will be fully removed in TorchAudio 2.9 (anticipated by the end of 2025).

Most of the APIs in `transforms`, `functional`, `compliance.kaldi`, `models` and `pipelines` modules will remain. These are the APIs that we identified as the most popular and valuable ones.

* A few APIs, specifically those relying on C++ implementations like RNNT loss and forced-alignment, may be dropped. Some, like `lfilter` and `overdrive`, will switch to pure-Python implementations, which might affect performance. We are exploring options to retain C++-backed APIs, but this is unlikely.
* Remaining APIs will be compatible with the latest stable PyTorch version. No new features will be added.

[...]

PR #3924, opened 2025-06-09
Add deprecation warnings

samanklesaria:
This PR adds deprecation warnings to all IO functions, prototypes, and functionality requiring C++ extensions as discussed in #3902.

### DIFF
--- a/src/torchaudio/_backend/utils.py
+++ b/src/torchaudio/_backend/utils.py
@@ -16,6 +16,7 @@ from .ffmpeg import FFmpegBackend
 from .soundfile import SoundfileBackend
 from .sox import SoXBackend
 
+from torchaudio._internal.module_utils import dropping_support
 
 @lru_cache(None)
 def get_available_backends() -> Dict[str, Type[Backend]]:
@@ -115,6 +116,7 @@ def get_load_func():
                 return backend
         raise RuntimeError(f"Couldn't find appropriate backend to handle uri {uri} and format {format}.")
 
+    @dropping_support
     def load(
         uri: Union[BinaryIO, str, os.PathLike],
         frame_offset: int = 0,
@@ -221,6 +223,8 @@ def get_save_func():
                 return backend
         raise RuntimeError(f"Couldn't find appropriate backend to handle uri {uri} and format {format}.")
 
+
+    @dropping_support
     def save(
         uri: Union[BinaryIO, str, os.PathLike],
         src: torch.Tensor,
--- a/src/torchaudio/_internal/module_utils.py
+++ b/src/torchaudio/_internal/module_utils.py
@@ -94,6 +94,9 @@ def deprecated(direction: str, version: Optional[str] = None, remove: bool = Fal
 
     return decorator
 
+dropping_support = deprecated(
+    "As TorchAudio is no longer being actively developed, this function can no longer be supported."
+    "See https://github.com/pytorch/audio/issues/3902 for more details.", version="2.9", remove=True)
 
 def fail_with_message(message):
     """Generate decorator to give users message about missing TorchAudio extension."""
--- a/src/torchaudio/functional/_alignment.py
+++ b/src/torchaudio/functional/_alignment.py
@@ -4,10 +4,13 @@ from typing import List, Optional, Tuple
 import torch
 from torch import Tensor
 from torchaudio._extension import fail_if_no_align
+from torchaudio._internal.module_utils import dropping_support
+
 
 __all__ = []
 
 
+@dropping_support
 @fail_if_no_align
 def forced_align(
     log_probs: Tensor,
--- a/src/torchaudio/functional/functional.py
+++ b/src/torchaudio/functional/functional.py
@@ -9,7 +9,7 @@ from typing import List, Optional, Tuple, Union
 import torch
 import torchaudio
 from torch import Tensor
-from torchaudio._internal.module_utils import deprecated
+from torchaudio._internal.module_utils import dropping_support, deprecated
 
 from .filtering import highpass_biquad, treble_biquad
 
@@ -1760,6 +1760,7 @@ def _fix_waveform_shape(
     return waveform_shift
 
 
+@dropping_support
 def rnnt_loss(
     logits: Tensor,
     targets: Tensor,
--- a/src/torchaudio/io/__init__.py
+++ b/src/torchaudio/io/__init__.py
@@ -2,7 +2,11 @@ from torio.io import CodecConfig, StreamingMediaDecoder as StreamReader, Streami
 
 from ._effector import AudioEffector
 from ._playback import play_audio
+from torchaudio._internal.module_utils import dropping_support
 
+CodecConfig.__init__ = dropping_support(CodecConfig.__init__)
+StreamReader.__init__ = dropping_support(StreamReader.__init__)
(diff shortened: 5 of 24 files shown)


src/torchaudio/_backend/utils.py
```diff
@@ -115,6 +116,7 @@ def dispatcher(
                 return backend
         raise RuntimeError(f"Couldn't find appropriate backend to handle uri {uri} and format {format}.")
 
+    @dropping_support
```
NicolasHug: I think it'll be easier for us to keep track of what is deprecated at a glance if we add the decorator in the `__init__.py` files, on the public functions. This way, we can just look at the `__init__.py` and be sure we didn't miss anything.
Do you mind moving the `@dropping_support` there ?

src/torchaudio/_internal/module_utils.py
```diff
@@ -94,6 +94,9 @@ def wrapped(*args, **kwargs):
 
     return decorator
 
+dropping_support = deprecated(
+    "As TorchAudio is no longer being actively developed, this function can no longer be supported."
+    "See https://github.com/pytorch/audio/issues/3902 for more details.", version="2.9", remove=True)
```
NicolasHug, line 101: Nice to see there are already some depreaction utils, and good job on re-using those!

We can work towards consolidating the exact message in future PRs. For now, I think we should at least have 2 decorators:

- one for the io functionality like `load`, `save` etc. above, inviting users to migrate to TorchCodec
- one for general functions that are getting removed

src/torchaudio/functional/_alignment.py
```diff
@@ -4,10 +4,13 @@
 import torch
 from torch import Tensor
 from torchaudio._extension import fail_if_no_align
+from torchaudio._internal.module_utils import dropping_support
+
 
 __all__ = []
 
 
+@dropping_support
```
NicolasHug: Similar to the other comment above, we should try to add the deprecations in the `__init__.py` files when possible.

src/torchaudio/functional/functional.py
```diff
@@ -1760,6 +1760,7 @@ def _fix_waveform_shape(
     return waveform_shift
 
 
+@dropping_support
```
NicolasHug: In the case of the functional, we should make sure their corresponding transforms classes in `transforms/` are deprecated too. Here there is the `RNNTLoss` class.

I was going to comment that we should move the deprecation to `__init__`, but unfortunately this file is technically public, because it has no leading underscore anywhere: `src/torchaudio/functional/functional.py`. So it's possible that some users have been importing things as `from torchaudio.functional.functional improt XYZ` instead of the intended `from torchaudio.functional improt XYZ`. As a result, I think we'll have to keep the deprecations in this file, instead of in ` torchaudio/functional/__init__.py`, which would have been more convenient

src/torchaudio/io/__init__.py
```diff
@@ -2,7 +2,11 @@
 
 from ._effector import AudioEffector
 from ._playback import play_audio
+from torchaudio._internal.module_utils import dropping_support
 
+CodecConfig.__init__ = dropping_support(CodecConfig.__init__)
+StreamReader.__init__ = dropping_support(StreamReader.__init__)
+StreamWriter.__init__ = dropping_support(StreamWriter.__init__)
```
NicolasHug, line 9: I'm interested to see if it works! Let's discuss tests during our chat

NicolasHug (commented):
Thanks a lot for working on this @samanklesaria ! I made a few comments below, let's chat more during our sync

src/torchaudio/_backend/utils.py
samanklesaria: Just making sure I understand here: in the `__init__.py` files, we usually export a function `foo` with `from submodule import foo` (and then including `"foo"` in `__all__`). You're proposing we change each of these imports to something like the following:
```python
from submodule import foo as foo_deprecated
foo = drop_support(foo_deprecated)
```
Is this correct?

src/torchaudio/functional/functional.py
samanklesaria: As the version of `RRNTLoss` in `transforms` itself calls the version in `functional`, putting a deprecation warning in both places would show duplicate warnings. I guess the way to get around this would be to have the version in `transforms` call an un-deprecated version of the on in `functional` instead?

src/torchaudio/io/__init__.py
samanklesaria: It seems to work when testing interactively.

src/torchaudio/_backend/utils.py
NicolasHug: That's what I had in mind yes. I haven't checked but I hope it should work? To avoid `foo_deprecated` to be public, I think we'd just need to import it as:

```py
from submodule import foo as _foo_deprecated
# or shorter
from submodule import foo as _foo
```

samanklesaria:
The codebase already uses a `@deprecated` decorator to deprecate certain functions. These are currently decorated inline rather than separate in the `__init__.py` files as you proposed. Should I move these deprecations to the init files as well?

NicolasHug:
> Should I move these deprecations to the init files as well?

I think it'll be easier for us if we don't touch anything that has already been deprecated, at least for the time being. It helps to keep the new deprecations we're introducing separate from the rest, which is why I like that you created a new `dropping_support` decorator, on top of the `deprecated` decorator.

src/torchaudio/_backend/utils.py
samanklesaria: I've changed the code to use these pattern wherever `__init__.py` files exist. Not all submodules are structured this way though (for example `kaldi_io`), so when there isn't an init file, I keep the deprecations inline. Also: as every single function in the `prototype` directory is being deprecated, I've kept those deprecations inline as well.

test/torchaudio_unittest/deprecation_test.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class TestDeprecations(PytorchTestCase):
+    def test_deprecations(self):
+        for func in UNSUPPORTED:
+            with self.assertWarnsRegex(UserWarning, r"deprecated"):
+                try:
+                    func()
+                except:
+                    pass
```
NicolasHug: Nice, that's a lot simpler than what I anticipated.

For my own ref and for @scotts, I was surprised this would even work because I thought the call to `func()` would raise a loud `TypeError` error before we even get a chance to raise the deprecation warning. But the `@deprecated` decorator that we use actually raises the warning **before** entering the underlying function, so this is correct.

I would just suggest to make it slightly more robust with the following changes:

- Use `pytest` and `pytest.mark.parametrize()`. Turns out torchaudio already invokes pytest, they just don't use it in the test, probably for historical reasons. But there's no reason we can't do that anymore.
- assert the type of the raised exception

```py
import pytest

from torchaudio._internal.module_utils import UNSUPPORTED


@pytest.mark.parametrize("func", UNSUPPORTED)
def test_deprecations(func):
    with pytest.warns(UserWarning, match="deprecated"):
        try:
            func()
        except Exception as e:
	     # Type or Runtime error because we call func() without proper parameters.
             # The deprecation warning is still properly raised, since it is emitted before
             # the underlying (deprecated) function is called. 
            assert isinstance(e, (TypeError, RuntimeError))
```

PR #3926, opened 2025-06-10
Fix CPU Linux unittest job - part 1

NicolasHug:
See below for summary of changes


samanklesaria:
Is this what we want to do for the CI going forward: install with pip rather than conda?

NicolasHug:
For installing `torch`, yes, we should use `pip` instead of `conda`. PyTorch stopped published conda packages earlier this year.
But we can still use conda as the env manager, and we can still install non-pytorch packages with conda.

Right now, the changes in this PR have fixed one part of the unit-test linux CPU jobs: on `main` we can't even compile torchaudio, while now we can compile it, but I think the next step is timing out when installing the rest of the dependencies through conda.

NicolasHug:
OK, we're not fully done but I'll merge this PR as-is as it's already good progress. I finally got the test to run. They [fail](https://github.com/pytorch/audio/actions/runs/15609671867/job/43967331170?pr=3926) during collection with:

```
  E   ImportError: cannot import name 'NP_SUPPORTED_MODULES' from 'torch._dynamo.utils' (/opt/conda/envs/ci/lib/python3.11/site-packages/torch/_dynamo/utils.py)
```

but at least they run.


Summary of changes:

- install torch nightly with `pip` instead of with `conda`, since torch doesn't push conda packages anymore. I left the previous code as commented-out instead of removing it, so we have the previous reference when trying to get the GPU tests to run.
- unpin librosa which was causing `conda install` to time-out
- unpin scipy which was causing a dependency resolution conflict
- Only test Python 3.11 for simplicity. Later we'll want to put back more Python versions.


I too shortcuts and hard-coded
```
export GPU_ARCH_TYPE="cpu" # TODO change this
```

so we'll have to update this when we try to enable back GPU testing.

PR #3929, opened 2025-06-12
Add skip condition to vggish_pipeline_test test 

amd-sriram:
Add skip condition to vggish_pipeline_test test similar to https://github.com/pytorch/audio/blob/main/test/torchaudio_unittest/backend/dispatcher/ffmpeg/save_test.py#L166 if ffmpeg executable is not found

Tested on both rocm and nvidia

### DIFF
--- a/test/integration_tests/prototype/vggish_pipeline_test.py
+++ b/test/integration_tests/prototype/vggish_pipeline_test.py
@@ -1,7 +1,8 @@
 import torchaudio
 from torchaudio.prototype.pipelines import VGGISH
+from torchaudio_unittest.common_utils import skipIfNoExec
 
-
+@skipIfNoExec("ffmpeg")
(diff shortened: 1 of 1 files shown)


amd-sriram:
cc @NicolasHug Please review

NicolasHug:
Thanks for the PR @amd-sriram . 
Unfortunately I don't think we'll be able to merge most incoming PRs. We're in the process of a large re-factoring of torchaudio (https://github.com/pytorch/audio/issues/3902) and a lot of functionality will be deprecated. During that time, we won't be able to prioritize fixes or improvements, sorry.

amd-sriram:
@NicolasHug : When would the refactoring of torch audio be completed ? is there a tentative date by when it will be completed and we can revisit these PRs again ?

PR #3930, opened 2025-06-12
For rocm, atol is reduced for gradcheck of oscillator_test to pass wh…

amd-sriram:
For rocm, atol is reduced for gradcheck of oscillator_test to pass when using slow mode. 

Tested on rocm and nvidia


amd-sriram:
cc @NicolasHug Please review

PR #3942, opened 2025-06-17
Disable tests for deprecated functionality

samanklesaria:

This is a draft PR for testing the CI failures.

### DIFF
--- a/.github/scripts/unittest-linux/run_test.sh
+++ b/.github/scripts/unittest-linux/run_test.sh
@@ -30,5 +30,5 @@ fi
 
 (
     cd test
-    pytest torchaudio_unittest -k "not backend and not /io/ and not prototype and not sox and not ffmpeg and not fairseq"
+    pytest torchaudio_unittest -k "not backend and not /io/ and not prototype and not sox and not ffmpeg and not fairseq and not hdemucs"
 )


samanklesaria:
Most of the errors are of the form 'RuntimeError: Error saving audio file: failed to open file /tmp/...'. These don't occur when I run the tests locally. I assume this means that the script doesn't have permission to write to the tmp directory. But tmp should have been given 777 permissions in the github actions yaml.

samanklesaria:
It seems like the CI hasn't been working for a while: the last release didn't seem to work either (https://github.com/pytorch/audio/actions/runs/14930335102).

.github/workflows/unittest-linux-cpu.yml
```diff
@@ -1,69 +1,72 @@
...
 
 on:
-  pull_request:
-  push:
-    branches:
-      - nightly
-      - main
-      - release/*
-  workflow_dispatch:
+    pull_request:
```
NicolasHug: If possible, revert to the previous indentation, because it makes it hard to know what load-bearing changes were made in the PR and thus complicates the review

src/torchaudio/transforms/_transforms.py
```diff
@@ -1846,7 +1847,7 @@ def forward(
             Tensor: Loss with the reduction option applied. If ``reduction`` is  ``"none"``, then size (batch),
             otherwise scalar.
         """
-        return F.rnnt_loss(
+        return rnnt_loss(
```
NicolasHug, line 1850: Can you share why this change was needed?

.github/scripts/unittest-linux/install.sh
```diff
@@ -95,8 +95,8 @@ fi
 # Note: installing librosa via pip fail because it will try to compile numba.
 (
     set -x
-    conda install -y -c conda-forge ${NUMBA_DEV_CHANNEL} sox libvorbis librosa parameterized 'requests>=2.20' 'ffmpeg>=6,<7'
-    pip install kaldi-io SoundFile coverage pytest pytest-cov scipy expecttest unidecode inflect Pillow sentencepiece pytorch-lightning 'protobuf<4.21.0' demucs tinytag pyroomacoustics flashlight-text git+https://github.com/kpu/kenlm
+    conda install -y -c conda-forge ${NUMBA_DEV_CHANNEL} sox libvorbis parameterized 'requests>=2.20' 'ffmpeg>=6,<7'
+    pip install kaldi-io SoundFile librosa coverage pytest pytest-cov scipy expecttest unidecode inflect Pillow sentencepiece pytorch-lightning 'protobuf<4.21.0' demucs tinytag pyroomacoustics flashlight-text git+https://github.com/kpu/kenlm
```
NicolasHug, line 98: There's a comment above about installing librosa via pip being problematic. Is it still relevant?

.github/scripts/unittest-linux/run_test.sh
```diff
@@ -30,6 +30,7 @@ fi
 
 (
     cd test
-    pytest "${args[@]}" torchaudio_unittest
+    export CUBLAS_WORKSPACE_CONFIG=:16:8
```
NicolasHug: I suspect this may not be needed - at least not for now, since we're not testing on GPU?

src/torchaudio/transforms/_transforms.py
samanklesaria: Functions that get wrapped with a decorator confuse torchscript. There's a bunch of tests for torchscript functionality in the repository, and torchscript isn't officially deprecated yet. Probably the best solution would be to fix this bug in torchscript. But the easier solution is to avoid using the wrapped versions when possible. In this case, because we give a deprecation warning on initialization of the transform, we don't need to give one again when the transform gets called. And this way, the initialization (which is wrapped) takes place outside of torchscript's tracing, while the running part (which torchscript *does* trace) doesn't have any wrapping.

.github/scripts/unittest-linux/install.sh
samanklesaria: Good point- it seems no longer relevant. The one installed from conda was failing (some version incompatibility), but the pip one seems to work just fine.

.github/workflows/unittest-linux-cpu.yml
samanklesaria: Done. Took a minute to figure out how to disable auto-formatting on save. This is what I get for trying a new editor.

.github/scripts/unittest-linux/run_test.sh
samanklesaria: True- this was to get tests working locally on my own machine.

samanklesaria:
It's odd that `torchaudio_unittest/io/stream_writer_test.py` is still running: I thought pytest's `-k 'not /io/'` should exclude that. It seems to on my machine.

PR #3945, opened 2025-06-26
Rework LoadHIP.cmake to be based purely on CMAKE_PREFIX_PATH.

jammm:
* Eliminates dependence on `/opt/rocm` and path based heuristics.
* Normalizes package finding for Rocm 6.5+ layout.
* Workaround cmake >= 4.0 for hiprtc

This will help with windows support. I understand that audio is in a maintenance phase, but I highly recommend this to be merged in order to allow torchaudio to be built on for AMD systems on Windows.

Technically, this still doesn't really compile the CUDA kernels because USE_ROCM doesn't include some of those .cu files. But it still does allow building the wheels and getting things functionally running.

### DIFF
--- a/cmake/LoadHIP.cmake
+++ b/cmake/LoadHIP.cmake
@@ -1,270 +1,157 @@
-set(PYTORCH_FOUND_HIP FALSE)
-
-if(NOT DEFINED ENV{ROCM_PATH})
-  if(UNIX)
-    set(ROCM_PATH /opt/rocm)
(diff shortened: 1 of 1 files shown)


ScottTodd:
This would probably be nice to have, though I don't think it is strictly required right now, at least when building 'pytorch' and 'audio' together. I minimally just needed https://github.com/pytorch/audio/pull/3991 to build with ROCm on Windows.

PR #3950, opened 2025-06-27
Add IO warning for functionality moved to TorchCodec

samanklesaria:

### DIFF
--- a/src/torchaudio/__init__.py
+++ b/src/torchaudio/__init__.py
@@ -1,4 +1,4 @@
-from torchaudio._internal.module_utils import dropping_support
+from torchaudio._internal.module_utils import dropping_io_support
 
 # Initialize extension and backend first
 from . import _extension  # noqa  # usort: skip
@@ -12,8 +12,8 @@ from ._backend import (  # noqa  # usort: skip
     set_audio_backend,
 )
 
-load = dropping_support(_load)
-save = dropping_support(_save)
+load = dropping_io_support(_load)
+save = dropping_io_support(_save)
 
 from . import (  # noqa: F401
     compliance,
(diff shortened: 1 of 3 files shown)


src/torchaudio/io/__init__.py
```diff
@@ -1,14 +1,14 @@
...
 
-CodecConfig.__init__ = dropping_support(CodecConfig.__init__)
-StreamReader.__init__ = dropping_support(StreamReader.__init__)
-StreamWriter.__init__ = dropping_support(StreamWriter.__init__)
-AudioEffector.__init__ = dropping_support(AudioEffector.__init__)
-play_audio = dropping_support(_play_audio)
+CodecConfig.__init__ = dropping_io_support(CodecConfig.__init__)
+StreamReader.__init__ = dropping_io_support(StreamReader.__init__)
+StreamWriter.__init__ = dropping_io_support(StreamWriter.__init__)
+AudioEffector.__init__ = dropping_io_support(AudioEffector.__init__)
```
NicolasHug: there won't actually be a replacement for audioeffector, so we should use the normal decorator for this. same for play_audio.

src/torchaudio/__init__.py
```diff
@@ -16,8 +16,8 @@
 get_audio_backend = dropping_support(_get_audio_backend)
 info = dropping_support(_info)
 list_audio_backends = dropping_support(_list_audio_backends)
-load = dropping_support(_load)
-save = dropping_support(_save)
+load = dropping_io_support(_load)
+save = dropping_io_support(_save)
```
NicolasHug, line 20: Can you also add that for `get_audio_backend()`, `info()` etc. ? thank you!

samanklesaria: Done.

src/torchaudio/io/__init__.py
samanklesaria: Done.

PR #3954, opened 2025-07-01
Port autograd parts of lfilter to python

samanklesaria:


src/libtorchaudio/lfilter.cpp
```diff
@@ -100,194 +100,19 @@ void lfilter_core_generic_loop(
   }
 }
```
NicolasHug, line 102: We've got `lfilter_core_generic_loop` left-over above in this file. Is it still needed?

samanklesaria: Good point. That path was used before if you had tensors on a cuda device but you hadn't compiled torchaudio with support for CUDA. We should probably still keep it around. I think we can register it with the dispatcher as a  `CompositeExplicitAutograd` path, which, as I understand it, will work as a catch-all if other dispatcher keys don't kick in.

PR #3964, opened 2025-07-09
Use torchcodec for loading

samanklesaria:
As a first step to removing deprecated functionality, this PR gets the datasets to use `torchcodec` to load their files.


NicolasHug:
Thanks for the PR and sorry for the extra work. I think we can close this now, as we'll be relying on something like in https://github.com/pytorch/audio/pull/3978 and / or https://github.com/pytorch/audio/pull/3986

PR #3965, opened 2025-07-09
Remove io deprecated

samanklesaria:
This PR removes all use of audio loading functionality within the codebase, as well as deprecated directories like prototype and sox.


NicolasHug:
Thanks for the PR  - as discussed offline we'll be submitting multiple PRs to delete / migrate individual parts. This will be more tractable and more easily reversible

PR #3971, opened 2025-07-12
Fix argument comment in pytorch/audio/src/libtorchaudio/rnnt/cpu/compute.cpp +6

r-barnes:
Summary:
Argument comments ensure that the correct values are assigned to the correct arguments.

This diff is a manual fix (assisted by the script in D78190897) of one or more argument comments in pytorch/audio/src/libtorchaudio/rnnt/cpu/compute.cpp. Once existing violations are drawn down we will make mismatches between argument comments and argument names an error.

Reviewed By: dtolnay

Differential Revision: D78191311


pearu:
Closing as the PR is included in https://github.com/pytorch/audio/pull/3987

PR #3978, opened 2025-07-16
Let load and save rely on *_with_torchcodec

NicolasHug:
Not meant to be merged as-is, just illustrating what we'll eventually do, morally speaking.


samanklesaria:
What more needs to happen before we can merge this?

PR #3982, opened 2025-07-18
Remove test_[tedlium3|mustc]_lightning.py

samanklesaria:
The EMFORMER_RNNT_BASE_MUSTC and EMFORMER_RNNT_BASE_TEDLIUM3 pipelines are prototypes. Only the the EMFORMER_RNNT_BASE_LIBRISPEECH pipeline is in the main `models` directory.


NicolasHug:
I think these tests are relying on the dataset, I'm not sure if it's related to the pipeline. We may need to remove them anyway since they're using external dependencies (ligntning, sentencepiece), but let's clarify offline.

samanklesaria:
You're right, this doesn't use the prototype pipeline stuff. My mistake.

PR #3986, opened 2025-07-18
Add torchcodec mock with wav loading and saving

samanklesaria:
This mock module will allow tests that load and save audio to still work as long as they only load and save wavs (which they all do).


NicolasHug:
This sounds like a good solution, thank you. It doesn't seem that these mocks are being used in the current state of this PR. Can you show how we'd be using this mock by modifying one or two existing tests?

samanklesaria:
This PR depends on the functionality of https://github.com/pytorch/audio/pull/3978.
I will change it back from draft status once 3978 is merged.

pearu:
Closing as irrelevant after https://github.com/pytorch/audio/pull/4092.

PR #4006, opened 2025-07-31
Disable asr cuda ctc tutorial

samanklesaria:


NicolasHug:
This isn't disable the build of the tutorial, this is only removing links to the tutorial (and to other APIS  that shouldn't be removed)

We can see the tutorial is still built in the logs https://github.com/pytorch/audio/actions/runs/16659498677/job/47152927025?pr=4006

Let's revert and discuss how to do this.

PR #4009, opened 2025-08-01
Replace direct calls to setup.py with pip equivalents

zklaus:


NicolasHug (approved):
Thanks for the PR @zklaus 

I think we have deactivated most of those in the CI at the moment, but the linux CPU tests are happy, so that's what matters.

PR #4010, opened 2025-08-01
Mocking Librosa in tests

samanklesaria:


samanklesaria:
Strangely, `torchaudio_unittest/transforms/librosa_compatibility_cpu_test.py` passes just fine on my machine. But a difference of `0.007` between the cached librosa value and the torchaudio computed value is quite large.

Perhaps this is a versioning thing? I have librosa 0.11.0 and numpy 2.1.2. 

I think the next step is to ensure that my changing the numpy version during testing didn't break anything by always comparing with librosa computed values rather than the cache.

samanklesaria:
The next logical thing to try is to get the exact versions of all the packages used in CI and ensure that my local setup (from which I'm generating the comparison files) matches them.

samanklesaria:
The CI system seems to use numpy 2.2.6 and librosa 0.11.0. I'll pin this and see what happens.

PR #4011, opened 2025-08-01
Remove sox test

samanklesaria:


NicolasHug:
I think these tests were meant to be test both for `sox_effect`, but also for the `torchaudio.functional` module that `sox_effects` were compared against. So I'm not sure we should be deleting those tests. I think we should be porting them in the same what that we're porting the librosa and kaldi tests

PR #4013, opened 2025-08-01
Remove deprecated functionality from smoke_test

samanklesaria:

### DIFF
--- a/test/smoke_test/smoke_test_no_ffmpeg.py
+++ b/test/smoke_test/smoke_test_no_ffmpeg.py
@@ -1,0 +1,3 @@
+from smoke_test import main
+
+main(["--no-ffmpeg"])


test/smoke_test/smoke_test.py
```diff
@@ -6,28 +6,14 @@
 
 def base_smoke_test():
     import torchaudio  # noqa: F401
-    import torchaudio.compliance.kaldi  # noqa: F401
     import torchaudio.datasets  # noqa: F401
     import torchaudio.functional  # noqa: F401
     import torchaudio.models  # noqa: F401
-    import torchaudio.pipelines  # noqa: F401
```
NicolasHug, line 13: we're keeping this one I think?

test/smoke_test/smoke_test.py
```diff
@@ -38,14 +24,13 @@ def main(args=None) -> None:
...
     _chdir()
-    _run_smoke_test(options.ffmpeg)
+    _run_smoke_test()
 
 
 def _parse_args(args):
     parser = argparse.ArgumentParser()
 
     # Warning: Please note this option should not be widely used, only use it when absolutely necessary
-    parser.add_argument("--no-ffmpeg", dest="ffmpeg", action="store_false")
```
NicolasHug, line 48: let's keep this line because I think some of the scripts (from other repos) are potentially relying on it still. But we should edit the comment above and say "this does nothing, only exists for BC"

NicolasHug (approved):
Thanks for the PR, approving to unblock but please see comments below before merging

test/smoke_test/smoke_test.py
samanklesaria: Oops, yes, quite right.

NicolasHug:
Looks like the Windows wheel building jobs are failing with `pytorch/audio/test/smoke_test/smoke_test_no_ffmpeg.py not found
`

I think it's because the scripts in `pytorch/test-infra` are still relying on it. Let's just put the file back (and forget about it forever)

PR #4017, opened 2025-08-04
Replace tests comparing functionality to sox with hard coded results.

samanklesaria:
To eliminate the sox dependency, this PR pre-computes the result of calls in the test suite that compare functionality to sox.


test/torchaudio_unittest/common_utils/sox_utils.py
```diff
@@ -114,3 +121,5 @@ def run_sox_effect(input_file, output_file, effect, *, output_sample_rate=None,
         command += ["rate", str(output_sample_rate)]
     print(" ".join(command))
     subprocess.run(command, check=True)
+    path.parent.mkdir(parents=True, exist_ok=True)
+    shutil.copyfile(output_file, path)
```
NicolasHug: I'm not sure I understand this logic. We want to remove the call to `subprocess.run(command, check=True)`, don't we?

samanklesaria: I forgot to put a return statement in above. But we don't want to remove the subprocess call- that's still something to leave so that the reference files can get generated if they don't exist.

test/torchaudio_unittest/common_utils/sox_utils.py
```diff
@@ -103,8 +105,14 @@ def _flattern(effects):
...
-def run_sox_effect(input_file, output_file, effect, *, output_sample_rate=None, output_bitdepth=None):
-    """Run sox effects"""
+def run_sox_effect(request, input_file, output_file, effect, *, output_sample_rate=None, output_bitdepth=None):
+    """Save or load the result of running sox effects for the test-id `request`."""
+
+    path = Path(f"torchaudio_unittest/assets/sox_expected_results/{request}.wav")
+    if os.path.exists(path):
+        shutil.copyfile(path, output_file)
+        return
+
```
NicolasHug, line 118: IIUC the end state we want is for `if os.path.exists(path):` to be True, right? If it's false it means we forgot to check-in a file? If that's the case, let's not have the `if` part and always do 

```
        shutil.copyfile(path, output_file)
        return
```
unconditionally.

I agree with your previous point that the file generation should still exist for reference, but for safety it might be best to comment it out, so that in never gets executed.

In the current state, as I understand it, if `sox` is available and the file doesn't exist for whatever reason, then we are effectively checking against sox's output instead of checking against a file which *should* exist - so we're not testing what we think we're testing.

samanklesaria: Your understanding is correct! I was planning to address your point by just removing the sox dependency from the dependencies installed on CI, so that if the file was missing we'd see an error about importing the library. But I can certainly comment the generation part out if you think that's clearer.

NicolasHug: It'll be safer to do all of those:

- remove dep from CI
- remove if condition
- comment-out file-generation logic

PR #4018, opened 2025-08-05
Use pre-computed values in librosa compat test

samanklesaria:
This PR adds functionality for both saving files containing pre-computed values and using these precomputed values for testing that torchaudio's implementations of functionality is the same as that offered by librosa.


test/librosa_mock.py
```diff
@@ -0,0 +1,37 @@
+import re
+import os
+from pathlib import Path
+import torch
+
+def mock_function(f):
```
NicolasHug, line 6: please add a short description of what this function does, and why

samanklesaria:
GPU tests are failing because CI is using an old version of numpy still.

test/librosa_mock.py
samanklesaria: Done.

test/torchaudio_unittest/deprecation_test.py
```diff
@@ -1,34 +0,0 @@
-import pytest
```
NicolasHug: Why was this removed? We haven't removed all of the deprecated APIs yet (e.g. the prototypes), so I'm not sure we should remove this in this PR?

samanklesaria: Agreed! I had removed it temporarily as it was complicating generation, but forgot to add it back. Thanks for pointin this out.

NicolasHug (approved):
Thanks @samanklesaria

PR #4019, opened 2025-08-05
[COMPARISON TESTS] Precompute values for kaldi comparison tests 

samanklesaria:
To remove the kaldi dependency during testing, this PR has code to both compute files we can use for static comparisons between kaldi functionality and torchaudio functionality as well as code to use these static files to test for similar functionality when no kaldi dependency is available.

### DIFF
--- a/examples/tutorials/squim_tutorial.py
+++ b/examples/tutorials/squim_tutorial.py
@@ -73,13 +73,11 @@ print(torchaudio.__version__)
 #
 # First import the modules and define the helper functions.
 #
-# We will need torch, torchaudio to use Torchaudio-squim, Matplotlib to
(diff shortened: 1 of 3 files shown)


samanklesaria:
Windows is freaking out about the filename
"test/torchaudio_unittest/assets/kaldi_expected_results/test/torchaudio_unittest/compliance/kaldi/kaldi_compatibility_cpu_test.py::TestKaldiFloat32::test_fbank_00.pt". What about this is not allowed in windows? The colons in filenames?

samanklesaria:
I guess the natural thing to do here is to replace the colons with something else. Perhaps an underscore.

NicolasHug (approved):

PR #4020, opened 2025-08-05
Make alphas_a standard C array in forced_align

samanklesaria:
As a first step in porting forced_align to the stable ABI, this turns a torch::Tensor used internally into a standard multidimensional C array for the cpu implementation.

### DIFF
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -1,5 +1,9 @@
 #include <torch/script.h>
 #include <torch/torch.h>
+#include <torch/csrc/stable/library.h>
+#include <torch/csrc/stable/tensor.h>
+#include <torch/csrc/stable/ops.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
 
 using namespace std;
 
@@ -22,17 +26,17 @@ void forced_align_impl(
   const auto T = logProbs.size(1);
   const auto L = targets.size(1);
   const auto S = 2 * L + 1;
-  torch::Tensor alphas = torch::empty(
-                             {2, S},
-                             torch::TensorOptions()
-                                 .device(logProbs.device())
-                                 .dtype(logProbs.dtype()))
-                             .fill_(kNegInfinity);
+
+  auto alphas_a = new scalar_t[S][2];
+  for (int i = 0; i < S; i++) {
+    alphas_a[i][0] = kNegInfinity;
+    alphas_a[i][1] = kNegInfinity;
+  }
+
   torch::Tensor backPtr = torch::empty({T, S}, torch::kInt8).fill_(-1);
   auto logProbs_a = logProbs.accessor<scalar_t, 3>();
   auto targets_a = targets.accessor<target_t, 2>();
   auto paths_a = paths.accessor<target_t, 2>();
-  auto alphas_a = alphas.accessor<scalar_t, 2>();
   auto backPtr_a = backPtr.accessor<int8_t, 2>();
   auto R = 0;
   for (auto i = 1; i < L; i++) {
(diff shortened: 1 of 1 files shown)


src/libtorchaudio/forced_align/cpu/compute.cpp
```diff
@@ -22,17 +26,17 @@ void forced_align_impl(
...
-                             torch::TensorOptions()
-                                 .device(logProbs.device())
-                                 .dtype(logProbs.dtype()))
-                             .fill_(kNegInfinity);
+
+  auto alphas_a = new scalar_t[S][2];
+  for (int i = 0; i < S; i++) {
+    alphas_a[i][0] = kNegInfinity;
+    alphas_a[i][1] = kNegInfinity;
+  }
```
NicolasHug, line 33: QQ was there a reason to specify the array as (S, 2) while the tensor was declared as (2, S)?
Some of the loops now become:

```
    for (auto j = 0; j < S; ++j) {
      alphas_a[j][curIdxOffset] = -std::numeric_limits<scalar_t>::infinity();
    }
```

which may not be optimizing for cache hit anymore?

src/libtorchaudio/forced_align/cpu/compute.cpp
```diff
@@ -22,17 +26,17 @@ void forced_align_impl(
...
   const auto L = targets.size(1);
   const auto S = 2 * L + 1;
-  torch::Tensor alphas = torch::empty(
-                             {2, S},
-                             torch::TensorOptions()
-                                 .device(logProbs.device())
-                                 .dtype(logProbs.dtype()))
-                             .fill_(kNegInfinity);
+
+  auto alphas_a = new scalar_t[S][2];
```
NicolasHug: If my understanding is correct, we expect `scalar_t` to be of logProbs' dtype - let's add a comment in the code to make that explicit.

NicolasHug (approved):

NicolasHug:
approving, please also add comments as suggested in  https://github.com/pytorch/audio/pull/4021#discussion_r2257101452 before merging

src/libtorchaudio/forced_align/cpu/compute.cpp
samanklesaria: The reason is that I can still use multidimensional indexing syntax if the inner dimension is 2, whereas if the inner dimension is the dynamic S, I would need to flatten the array and calculate the appropriate 1-d offset.

samanklesaria:
As far as adding comments about indexing goes, this PR doesn't remove multidimensional indexing syntax. So what exactly do you want the comments at each indexing location to say?

src/libtorchaudio/forced_align/cpu/compute.cpp
samanklesaria: Done.

src/libtorchaudio/forced_align/cpu/compute.cpp
NicolasHug: Thanks for clarifying.
I still wonder whether we should preserve contiguity over `S` rather than over the `2` dimension, for the cache hit properties I mentioned above. Do you know how large we can expect `S` to be in practice? If it's large, then maybe that's something we need to pay attention to.

If we decide to preserve the original contiguity, maybe a vec of vec could help?

NicolasHug:
Sorry, I got mixed up - no need to add the indexing comments.

I left a follow-up comment in https://github.com/pytorch/audio/pull/4020/files#r2262472393

src/libtorchaudio/forced_align/cpu/compute.cpp
samanklesaria: I don't know how large we can expect $S$ to be in practice. I'll close this PR and open a different one in which alphas_a is treated as a single dimensional array in the original ordering, just like backPtr in the other PR.

PR #4021, opened 2025-08-05
Convert backptr to standard array

samanklesaria:
Working towards removing the dependence on the `torch` library, this PR changes the `backPtr` tensor in `forced_align` to a standard array. As the dimensions are not known at compile time and we're not using gnu extensions (which would allow for dynamic multidimensional arrays), multidimensional array access is converted to linear indexing.

### DIFF
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -33,11 +33,14 @@ void forced_align_impl(
     alphas_a[i][1] = kNegInfinity;
   }
 
-  torch::Tensor backPtr = torch::empty({T, S}, torch::kInt8).fill_(-1);
+  auto backPtr_a = new int8_t[T * S];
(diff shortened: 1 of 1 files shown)


src/libtorchaudio/forced_align/cpu/compute.cpp
```diff
@@ -84,7 +87,7 @@ void forced_align_impl(
     if (start == 0) {
       alphas_a[0][curIdxOffset] =
           alphas_a[0][prevIdxOffset] + logProbs_a[batchIndex][t][blank];
-      backPtr_a[t][0] = 0;
+      backPtr_a[S * t] = 0;
```
NicolasHug: For the sake of clarity, here and in other places you modified, let's add the previous version as a comment:

```suggestion
      backPtr_a[S * t] = 0;  // backPtr_a[t][0] = 0;
```

This should help the reader understand what's going on

NicolasHug (approved):

PR #4022, opened 2025-08-05
[STABLE ABI] Stable forced_align on cpu

samanklesaria:
This PR ports the `forced_align` code to the stable ABI. It uses the temporary Accessor class defined in the `stable_accessor` PR.

### DIFF
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -131,10 +131,10 @@ void forced_align_impl(
 }
 
 std::tuple<Tensor, Tensor> compute(
-    const Tensor& logProbs,
-    const Tensor& targets,
-    const Tensor& inputLengths,
-    const Tensor& targetLengths,
+    const Tensor logProbs,
+    const Tensor targets,
+    Tensor inputLengths,
+    Tensor targetLengths,
     const int64_t blank) {
   AOTI_TORCH_CHECK(logProbs.is_cpu(), "log_probs must be a CPU tensor");
   AOTI_TORCH_CHECK(targets.is_cpu(), "targets must be a CPU tensor");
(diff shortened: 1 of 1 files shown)


samanklesaria:
Note: this commit now depends on the `dtype` and `is_cpu` methods of stable::Tensor in `pytorch`, which have yet to be merged in. I will keep this PR in draft state until they make their way to nightly.

samanklesaria:
This PR passes all tests for `forced_align` on CPU except expected failures. This is because the correct handling of the expected failures requires porting the `amax` and `item` ops to the stable ABI. I will port these ops next.

MahmoudAshraf97:
Hello, is there a chance to get any features or improvements from #3787 merged? on the other side, I managed to remove the torch dependency completely from the FA api, so I guess it's doable, only the input `log_probs` array needs to be a torch tensor, all other arrays can be vectors or regular arrays, check https://github.com/MahmoudAshraf97/ctc-forced-aligner/blob/main/ctc_forced_aligner/forced_align_impl.cpp for reference

NicolasHug:
Hi @MahmoudAshraf97 , we're currently trying to migrate these ops to be ABI-stable, which is our current top priority. Unfortunately, we won't have the bandwidth to consider improvements like in https://github.com/pytorch/audio/pull/3787 at the moment, sorry. It's possible that we'll be able to re-consider in the future, but the near-term goal is for torchaudio to be stable while minimizing maintenance requirements.

samanklesaria:
The CI used to run fine. A recent change to the stable ABI now leads to `stable::Tensor::scalar_type()` being defined twice, which gives all the errors. Perhaps there's a new preferred way of importing the stable API headers?

pearu:
> The CI used to run fine. A recent change to the stable ABI now leads to `stable::Tensor::scalar_type()` being defined twice, which gives all the errors. Perhaps there's a new preferred way of importing the stable API headers?

The cause of `scalar_type()` being defined multiple times is in including `torch/csrc/stable/tensor.h` (read: including `torch/csrc/stable/tensor_inl.h` that defines `scalar_type()` as a C++ function) from different files. A workaround (that is used in https://github.com/pytorch/audio/pull/4073 and https://github.com/pytorch/audio/pull/4079) is to include `torch/csrc/stable/tensor.h` only from `utils.c` while all other files include `torch/csrc/stable/tensor_struct.h` to get the declaration of `Tensor`.

pearu:
Closing via https://github.com/pytorch/audio/pull/4079

PR #4023, opened 2025-08-05
Remove generic lfilter loop

samanklesaria:
The code for lfilter already has explicit cpu and CUDA. Do we really need a third fallback option? I guess there's an mps device type too. But shouldn't everything else get handled by cpu and cuda? If we removed the fallback, we wouldn't have to worry about porting the transpose, squeeze, and index_put operations to the stable ABI.  

If it's important to keep support for other devices, close this PR.

### DIFF
--- a/src/libtorchaudio/lfilter.cpp
+++ b/src/libtorchaudio/lfilter.cpp
@@ -73,33 +73,6 @@ void cpu_lfilter_core_loop(
       });
 }
 
-void lfilter_core_generic_loop(
-    const torch::Tensor& input_signal_windows,
-    const torch::Tensor& a_coeff_flipped,
-    torch::Tensor& padded_output_waveform) {
-  int64_t n_samples_input = input_signal_windows.size(2);
-  int64_t n_order = a_coeff_flipped.size(1);
-  auto coeff = a_coeff_flipped.unsqueeze(2);
-  for (int64_t i_sample = 0; i_sample < n_samples_input; i_sample++) {
-    auto windowed_output_signal =
-        padded_output_waveform
-            .index(
-                {torch::indexing::Slice(),
-                 torch::indexing::Slice(),
-                 torch::indexing::Slice(i_sample, i_sample + n_order)})
-            .transpose(0, 1);
-    auto o0 =
-        input_signal_windows.index(
-            {torch::indexing::Slice(), torch::indexing::Slice(), i_sample}) -
-        at::matmul(windowed_output_signal, coeff).squeeze(2).transpose(0, 1);
-    padded_output_waveform.index_put_(
-        {torch::indexing::Slice(),
-         torch::indexing::Slice(),
-         i_sample + n_order - 1},
-        o0);
-  }
-}
-
 } // namespace
 
 TORCH_LIBRARY(torchaudio, m) {
@@ -116,7 +89,3 @@ TORCH_LIBRARY_IMPL(torchaudio, CUDA, m) {
   m.impl("torchaudio::_lfilter_core_loop", &cuda_lfilter_core_loop);
 }
 #endif
-
-TORCH_LIBRARY_IMPL(torchaudio, CompositeExplicitAutograd, m) {
-  m.impl("torchaudio::_lfilter_core_loop", &lfilter_core_generic_loop);
-}


samanklesaria:
This PR will be converted back from draft once tests for MPS devices are turned on.

pearu:
Heads up: https://github.com/pytorch/audio/pull/4091 enables M1 tests.

pearu:
If we plan to land this, then apply also
```
diff --git a/src/torchaudio/functional/filtering.py b/src/torchaudio/functional/filtering.py
index 76deb04a..6dc488cf 100644
--- a/src/torchaudio/functional/filtering.py
+++ b/src/torchaudio/functional/filtering.py
@@ -923,7 +923,7 @@ def highpass_biquad(waveform: Tensor, sample_rate: int, cutoff_freq: float, Q: f
     return biquad(waveform, b0, b1, b2, a0, a1, a2)
 
 
-def _lfilter_core_generic_loop(input_signal_windows: Tensor, a_coeffs_flipped: Tensor, padded_output_waveform: Tensor):
+def _lfilter_core_loop(input_signal_windows: Tensor, a_coeffs_flipped: Tensor, padded_output_waveform: Tensor):
     n_order = a_coeffs_flipped.size(1)
     a_coeffs_flipped = a_coeffs_flipped.unsqueeze(2)
     for i_sample, o0 in enumerate(input_signal_windows.permute(2, 0, 1)):
@@ -932,12 +932,6 @@ def _lfilter_core_generic_loop(input_signal_windows: Tensor, a_coeffs_flipped: T
         padded_output_waveform[:, :, i_sample + n_order - 1] = o0
 
 
-if _IS_TORCHAUDIO_EXT_AVAILABLE:
-    _lfilter_core_loop = torch.ops.torchaudio._lfilter_core_loop
-else:
-    _lfilter_core_loop = _lfilter_core_generic_loop
-
-
 class DifferentiableFIR(torch.autograd.Function):
     @staticmethod
     def forward(ctx, waveform, b_coeffs):
```
to prevent breaking tests.

samanklesaria:
@pearu I had originally thought that we'd remove this generic loop from the C++ code and add it back as python code for the special case of mps (or anything that's both not cpu and not cuda). This way, select, matmul, unsqueeze, index_put, etc are no longer necessary in the ABI, and the only thing holding us back is parallel_for. I don't think I understand your previous comment; now that M1 tests exist, to make this work, I'd need to add the python port of the generic code. Does that make sense?

samanklesaria:
Just realized that there's already a python implementation of the generic loop.

samanklesaria:
> If we plan to land this, then apply also
> 
> ```
> diff --git a/src/torchaudio/functional/filtering.py b/src/torchaudio/functional/filtering.py
> index 76deb04a..6dc488cf 100644
> --- a/src/torchaudio/functional/filtering.py
> +++ b/src/torchaudio/functional/filtering.py
> @@ -923,7 +923,7 @@ def highpass_biquad(waveform: Tensor, sample_rate: int, cutoff_freq: float, Q: f
>      return biquad(waveform, b0, b1, b2, a0, a1, a2)
>  
>  
> -def _lfilter_core_generic_loop(input_signal_windows: Tensor, a_coeffs_flipped: Tensor, padded_output_waveform: Tensor):
> +def _lfilter_core_loop(input_signal_windows: Tensor, a_coeffs_flipped: Tensor, padded_output_waveform: Tensor):
>      n_order = a_coeffs_flipped.size(1)
>      a_coeffs_flipped = a_coeffs_flipped.unsqueeze(2)
>      for i_sample, o0 in enumerate(input_signal_windows.permute(2, 0, 1)):
> @@ -932,12 +932,6 @@ def _lfilter_core_generic_loop(input_signal_windows: Tensor, a_coeffs_flipped: T
>          padded_output_waveform[:, :, i_sample + n_order - 1] = o0
>  
>  
> -if _IS_TORCHAUDIO_EXT_AVAILABLE:
> -    _lfilter_core_loop = torch.ops.torchaudio._lfilter_core_loop
> -else:
> -    _lfilter_core_loop = _lfilter_core_generic_loop
> -
> -
>  class DifferentiableFIR(torch.autograd.Function):
>      @staticmethod
>      def forward(ctx, waveform, b_coeffs):
> ```
> 
> to prevent breaking tests.

We don't want to *always* use the generic loop. When it's applicable (cpu or cuda), we *do* want to use the extension version of `lfilter_core_loop`. It's just for macs (or when the extension doesn't exist) that we want to swap in the python one.

pearu:
> @pearu I had originally thought that we'd remove this generic loop from the C++ code and add it back as python code for the special case of mps (or anything that's both not cpu and not cuda). This way, select, matmul, unsqueeze, index_put, etc are no longer necessary in the ABI, and the only thing holding us back is parallel_for. I don't think I understand your previous comment;

The comment originates from reading the existing code: when this PR lands as it is, the following statement
```python
_lfilter_core_loop = torch.ops.torchaudio._lfilter_core_loop
``` 
will fail.

> Just realized that there's already a python implementation of the generic loop.

:+1:

samanklesaria:
> > @pearu I had originally thought that we'd remove this generic loop from the C++ code and add it back as python code for the special case of mps (or anything that's both not cpu and not cuda). This way, select, matmul, unsqueeze, index_put, etc are no longer necessary in the ABI, and the only thing holding us back is parallel_for. I don't think I understand your previous comment;
> 
> The comment originates from reading the existing code: when this PR lands as it is, the following statement
> 
> ```python
> _lfilter_core_loop = torch.ops.torchaudio._lfilter_core_loop
> ```
> 
> will fail.
> 
> > Just realized that there's already a python implementation of the generic loop.
> 
> 👍

Why? `_lfilter_core_loop` is still defined on line 80 of `lfilter.pp`.

pearu:
> We don't want to _always_ use the generic loop. When it's applicable (cpu or cuda), we _do_ want to use the extension version of `lfilter_core_loop`. It's just for macs (or when the extension doesn't exist) that we want to swap in the python one.

OK, I did not realize that. :crossed_fingers: M1 CI will pass.

pearu:
> Why? `_lfilter_core_loop` is still defined on line 80 of `lfilter.pp`.

I was misreading:
```
+ TORCH_LIBRARY_IMPL(torchaudio, CompositeExplicitAutograd, m) {
+   m.impl("torchaudio::_lfilter_core_loop", &lfilter_core_generic_loop);
+ }
```
Sorry for the noise.

pearu:
There is a bug in generic lfilter loop fixed in https://github.com/pytorch/audio/pull/4149.

The bug was revealed when disabling the CUDA kernel.

pearu:
Everything in generic lfilter loop has been ported to stable ABI and the code is now tested for correctness. So, closing this PR as unneeded.

PR #4028, opened 2025-08-07
[TUTORIAL DEPENDENCY REMOVAL] Use char-based pipeline instead of phoneme one for tacotron tutorial

samanklesaria:
This PR allows the the tacotron tutorial to lose its deep-phonemizer dependency.

### DIFF
--- a/examples/tutorials/tacotron2_pipeline_tutorial.py
+++ b/examples/tutorials/tacotron2_pipeline_tutorial.py
@@ -19,7 +19,7 @@ Text-to-Speech with Tacotron2
 # 1. Text preprocessing
 #
 #    First, the input text is encoded into a list of symbols. In this
-#    tutorial, we will use English characters and phonemes as the symbols.
+#    tutorial, we will use English characters as the symbols.
 #
 # 2. Spectrogram generation
 #
@@ -47,16 +47,6 @@ Text-to-Speech with Tacotron2
 # Preparation
 # -----------
 #
-# First, we install the necessary dependencies. In addition to
-# ``torchaudio``, ``DeepPhonemizer`` is required to perform phoneme-based
-# encoding.
-#
-
-# %%
-#  .. code-block:: bash
-#
-#      %%bash
-#      pip3 install deep_phonemizer
 
 import torch
 import torchaudio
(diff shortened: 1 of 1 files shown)


examples/tutorials/tacotron2_pipeline_tutorial.py
```diff
@@ -256,7 +203,7 @@ def plot():
 # WaveRNN model from the same bundle.
 #
 
-bundle = torchaudio.pipelines.TACOTRON2_WAVERNN_PHONE_LJSPEECH
+bundle = torchaudio.pipelines.TACOTRON2_WAVERNN_CHAR_LJSPEECH
```
NicolasHug, line 206: Something is off in this tutorial as a result of the change. The section above about spectrogram generation is still using `TACOTRON2_WAVERNN_PHONE_LJSPEECH`. And just above, the tutorial says

> Continuing from the previous section, we can instantiate the matching WaveRNN model from the same bundle


but the bundle has changed from `TACOTRON2_WAVERNN_PHONE_LJSPEECH` to `TACOTRON2_WAVERNN_CHAR_LJSPEECH`.

Can you please advise on how to resolve the mismatch?

samanklesaria: We should never be using `TACOTRON2_WAVERNN_PHONE_LJSPEECH`, as this will require the Deep-Phonemizer library. I must have missed some references to it if you're still seeing some.

samanklesaria: Good point- I indeed missed the previous reference. I'll change that

samanklesaria: I should have clarified at the start of the PR: the way I propose to deal with the Deep-Phonemizer dependency here is to swap out the `TACOTRON2_WAVERNN_CHAR_LJSPEECH` pipeline (which us a character-level model) for the TACOTRON2_WAVERNN_PHONE_LJSPEECH pipeline (which is a phoneme level model and requires the dependency). Most of the tutorial can continue unchanged (except for the section comparing character-level to phoneme level model results).

PR #4029, opened 2025-08-07
[TUTORIAL DEPENDENCY REMOVAL] Remove last trace of librosa in tutorials

samanklesaria:
This PR ports the use of librosa for amplitude-to-db calculations to the internal torchaudio functionality, eliminating the last dependency that the tutorials have on librosa.

### DIFF
--- a/examples/tutorials/audio_feature_extractions_tutorial.py
+++ b/examples/tutorials/audio_feature_extractions_tutorial.py
@@ -75,7 +75,8 @@ def plot_spectrogram(specgram, title=None, ylabel="freq_bin", ax=None):
     if title is not None:
         ax.set_title(title)
     ax.set_ylabel(ylabel)
-    ax.imshow(librosa.power_to_db(specgram), origin="lower", aspect="auto", interpolation="nearest")
+    power_to_db = T.AmplitudeToDB("power", 80.0)
+    ax.imshow(power_to_db(specgram), origin="lower", aspect="auto", interpolation="nearest")
 
 
 def plot_fbank(fbank, title=None):


examples/tutorials/audio_feature_extractions_tutorial.py
```diff
@@ -75,7 +75,8 @@ def plot_spectrogram(specgram, title=None, ylabel="freq_bin", ax=None):
     if title is not None:
         ax.set_title(title)
     ax.set_ylabel(ylabel)
-    ax.imshow(librosa.power_to_db(specgram), origin="lower", aspect="auto", interpolation="nearest")
+    power_to_db = T.AmplitudeToDB("power", 80.0)
+    ax.imshow(power_to_db(specgram), origin="lower", aspect="auto", interpolation="nearest")
```
NicolasHug, line 79: Before merging, can you please double-check that this is correct? Specifically about 

```
T.AmplitudeToDB("power", 80.0)
```

my understanding is that setting `stype="power"` means that `T.AmplitudeToDB` expects a power, not an amplitude (although it's not very clear to me). If that's the case, then does that mean we should be passing `specgram**2` instead of `specgram`?


I checked the plotted output and it seems visually very close to the current docs, so that's a good sign, but just want to make sure

samanklesaria: As the old code read `librosa.power_to_db`, we want to find a `torchaudio` native way to convert a *power* to a db. The `T.z("power", 80.0)` transform is supposed to do that. This correspondence is already tested in the `test_amplitude_to_db` test in `transforms/autograd_test_impl.py`. So we should be good to go!

PR #4031, opened 2025-08-11
Remove comparison dependencies and skipIfs

samanklesaria:
As previous PRs have precomputed the results to various calls to external libraries, we no longer need to depend on those libraries when running the test suite. This PR removes instructions to install `sox`, `librosa` and `kaldi` in the linux test suite runner and removes the `skipIf` declarations on the associated external library comparison tests.


samanklesaria:
We're now failing because we lack `sklearn`. But `sklearn` wasn't even in our dependencies before! All I can think of is that something else (like librosa) depended on it, and we were able to import it because pip installing librosa gave us access to sklearn too.

NicolasHug:
On the sklearn issue, I hope this is the only once, but if that's the case then let's just ignore the corresponding test. We can probably just skip it and comment-out the offending import line in `test/torchaudio_unittest/example/hubert/__init__.py` ?

samanklesaria:
We can't just remove the change to the python path, as that will break `test/example/hubert/test_crop_audio_label.py`. But the test itself doesn't use the kmeans functionality. The `learn_kmeans` function that uses sklearn is never tested. So as long as we hide the import inside the call to `learn_kmeans`, we're okay.

samanklesaria:
It seems like we need to remove the backends. If they're in the repository, they'll get added to the deprecations tests. But because the dependencies are there, we'll get errors. But that's a much larger change than we probably want to include in this PR. If we want to make the CI green for this PR without removing the sox backend from the codebase, the easiest fix I see is a hack to make the deprecation tests ignore anything from a sox module.

samanklesaria:
It also seems like the kaldi sliding window test expected results we're properly added when we merged. This is probably because in the upstream commit (the librosa tests) I changed the naming convention to remove cpu/cuda naming differences. And the kaldi files were't regenerated after the change.

src/torchaudio/_internal/module_utils.py
```diff
@@ -97,7 +97,8 @@ def decorator(func):
     {func.__doc__}
     """
 
-        UNSUPPORTED.append(wrapped)
+        if 'sox' not in func.__module__:
+            UNSUPPORTED.append(wrapped)
```
NicolasHug, line 103: Can you please add a comment saying this is a very temporary hack that we'll have to remove as soon as we remove the sox-related functionalities

samanklesaria: Done.

samanklesaria:
Verbally approved by Nicholas on slack. Merging.

PR #4035, opened 2025-08-11
Remove libtorio

samanklesaria:
This removes the deprecated libtorio components of torchaudio


samanklesaria:
This is a bigger chunk than I realized. The right thing to do is probably to remove the `AudioEffector` first, followed by other smaller pieces.

samanklesaria:
Whoops! I forgot that the base of this branch wasn't main. I'll do the ffmpeg one first.

PR #4036, opened 2025-08-11
Remove io directory

samanklesaria:
This PR removes the `io` directory. It depends on the `remove_backends` branch, contained in a separate PR.


samanklesaria:
It seems like I won't be able to do this until I remove the `backend` directory and port loading and saving to torchcodec as discussed.

src/torchaudio/__init__.py
```diff
@@ -12,11 +12,9 @@
     compliance,
     datasets,
     functional,
-    io,
     kaldi_io,
     models,
     pipelines,
-    sox_effects,
```
NicolasHug: Let's remove this in a separate PR dedicated to `sox_effects`. From what I can tell, `sox_effects.py` still exists.

samanklesaria: Good catch. I'll move it to the sox_effects PR.

PR #4037, opened 2025-08-12
Remove unnecessary installations for linux CPU test

samanklesaria:
Tests should no longer have external dependencies beyond scipy, numpy, pytorch, and testing utilities like parameterized. This PR removes the installation of dependencies beyond these when running CI.


samanklesaria:
We have to merge https://github.com/pytorch/audio/pull/4031 first.

samanklesaria:
It seems that CTC Decoder requires flashlight-text package, which it depends on pretty heavily. We could disable the tests for it. 

test_librispeech_lightning and test_mustc_lightning and test_tedlium3_lightning require lightning. It seems these just test lightning integration (which we're deprecating anyway), so we should be able to delete these tests. 

tacotron2/test_text_preprocessing.py requires unidecode. But we're only testing on ascii text. Would could give `text_to_sequence` a flag for whether or not to use `unidecode`? Or we could just add a warning to it saying that only ascii input is allowed? 

I'm still not sure how much we're testing the use of `inflect`, which is also used in `tacotron2/test_text_preprocessing.py`. Perhaps we can restrict our testing to examples that don't require it, and not load the module until it is needed.

NicolasHug:
OK, thanks. Let's just skip all of those by setting the appropriate env variable e.g. `TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_CTC_DECODER=1`.

.github/scripts/unittest-linux/install.sh
```diff
@@ -98,14 +98,9 @@ if [[ "$(python --version)" = *3.9* || "$(python --version)" = *3.10* ]]; then
 fi
 (
     set -x
-    conda install -y -c conda-forge ${NUMBA_DEV_CHANNEL} libvorbis parameterized 'requests>=2.20'
-    pip install SoundFile coverage pytest pytest-cov scipy expecttest unidecode inflect Pillow sentencepiece pytorch-lightning 'protobuf<4.21.0' demucs tinytag pyroomacoustics flashlight-text git+https://github.com/kpu/kenlm
+    conda install -y -c conda-forge ${NUMBA_DEV_CHANNEL} parameterized 'requests>=2.20'
+    pip install coverage pytest pytest-cov scipy expecttest Pillow
```
NicolasHug: Approving to unblock, feel free to merge when green after skipping the currently failing tests.

We shouldn't need Pillow, and we should remove the `pip install` line below as well with the TODO

samanklesaria: Done.

PR #4038, opened 2025-08-12
Remove gpu test installs

samanklesaria:
Removes gpu test installs.

### DIFF
--- a/.github/workflows/unittest-linux-gpu.yml
+++ b/.github/workflows/unittest-linux-gpu.yml
@@ -75,9 +75,7 @@ jobs:
         echo "::endgroup::"
         echo "::group::Install other Dependencies"
         
-        pip3 install parameterized requests
-        pip3 install coverage pytest pytest-cov scipy expecttest Pillow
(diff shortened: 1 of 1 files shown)


.github/workflows/unittest-linux-gpu.yml
```diff
@@ -73,33 +73,9 @@ jobs:
...
-        #   'protobuf<4.21.0' \
-        #   demucs \
-        #   tinytag \
-        #   flashlight-text \
-        #   git+https://github.com/kpu/kenlm/ \
-        #   git+https://github.com/pytorch/fairseq.git@e47a4c8
         
         pip3 install parameterized requests
-        pip3 install SoundFile coverage pytest pytest-cov scipy expecttest unidecode inflect Pillow sentencepiece pytorch-lightning 'protobuf<4.21.0' demucs tinytag 
+        pip3 install coverage pytest pytest-cov scipy expecttest Pillow
```
NicolasHug: we shouldn't need pillow, can you double check? it's also installed below so we should try to remove it from both statements

EDIT: and probably just remove the line below altogether

PR #4039, opened 2025-08-12
Torchcodec loading

samanklesaria:
This PR wraps the `load_with_torchcodec` and `save_with_torchcodec` functions with functions of the name `load` and `save` so that code that depends on the old `load` and `save` functions can continue to work in the future once we remove backend-specific code.

PR #4043, opened 2025-08-14
Remove dependencies from doc job

NicolasHug:

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -59,12 +59,6 @@ jobs:
 
         echo "::endgroup::"
         echo "::group::Install PyTorch"
-        # conda install \
-        #   --yes \
-        #   --quiet \
-        #   -c "pytorch-${CHANNEL}" \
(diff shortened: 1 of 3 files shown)


samanklesaria:
I guess we need to disable building `asr_inference_with_ctc_decoder_tutorial`.

samanklesaria:
I believe the remaining test errors have to do with the use of torchcodec. It might be worth waiting until https://github.com/pytorch/audio/pull/4039 is merged in to see if the problem goes away when torchcodec is no longer in the picture.

.github/scripts/unittest-linux/install.sh
```diff
@@ -40,6 +40,7 @@ case $GPU_ARCH_TYPE in
     ;;
 esac
 PYTORCH_WHEEL_INDEX="https://download.pytorch.org/whl/${UPLOAD_CHANNEL}/${GPU_ARCH_ID}"
+pip install "numpy>=1.26"
```
NicolasHug: LGTM but before merging can you double-check that changes to this file are necessary?

samanklesaria: Looks like they weren't necessary after all!

PR #4046, opened 2025-08-18
Remove sox

samanklesaria:
This PR removes the sox extensions built as part of libtorchaudio as well as the sox_effects subdirectory within the python source. It depends on the `remove_io` branch, which is in PR #4036 
Strangely, setting that branch as the target causes github not to run the CI currently, so I have set the target branch to main, duplicating the changes in 4036.


examples/libtorchaudio/augmentation/README.md
samanklesaria, line 1: This example (augmentation) heavily depends on sox and torchscript. I think the only way forward is to remove the example.

NicolasHug: seems to be about the C API anyway, which we're not going to support anymore

test/torchaudio_unittest/common_utils/case_utils.py
```diff
@@ -109,7 +109,7 @@ class TorchaudioTestCase(TestBaseMixin, PytorchTestCase):
 
 
 _IS_FFMPEG_AVAILABLE = torio._extension.lazy_import_ffmpeg_ext().is_available()
-_IS_SOX_AVAILABLE = torchaudio._extension.lazy_import_sox_ext().is_available()
+_IS_SOX_AVAILABLE = False
```
NicolasHug: We should/can proabbly remove all usages of this variable now?

samanklesaria: It doesn't seem to be used anywhere now. Sounds good!

PR #4048, opened 2025-08-19
Remove prototype files

samanklesaria:
This PR removes the prototype directory from `src` along with prototype references in `docs`. It currently depends on, and should be merged *after*, the CUTC wrapper removal PR, because sphinx now breaks on our deprecation-wrapped classes. Specifically, we get the docs build error:

```
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This suggests that if we just merged the CUTC PR first, this error would go away. Note that this error does not show up in the CI: the CI logs for sphinx only show the exit code and suppress all other output.


docs/source/_templates/autosummary/model_class.rst
```diff
@@ -147,25 +147,6 @@ Factory Functions
 
 {%- if prototype_factory[fullname] %}
 
-Prototype Factory Functions
-===========================
```
NicolasHug, line 151: please also remove mentions of prototype in `index.rst`

examples/asr/emformer_rnnt/pipeline_demo.py
```diff
@@ -1,107 +0,0 @@
-#!/usr/bin/env python3
```
NicolasHug: let's not touch the examples folder for now. It would require more work. For example this pipeline_demo.py file is mentioned in the README, so removing it would mean updating the readme. We'll do that as follow-ups

samanklesaria: This example used functionality from the prototype files, which is why I included it here. If you like, I can do a separate PR for the examples first, before merging this one?

docs/source/_templates/autosummary/model_class.rst
samanklesaria: Done.

examples/asr/emformer_rnnt/pipeline_demo.py
samanklesaria: Or should I just update the readme for this single demo first?

samanklesaria:
As written elsewhere in the PR, I changed the target to be the `no_wrap_ctc_decoder` branch because sphinx recently became unable to handle the wrapped CUTC decoder. Unfortunately, this isn't quite sufficient for docs building to work yet. Now we're getting a large number of errors like `libtorio.stream_writer.rst:40: ERROR: Unknown directive type "doxygenfunction"`. Of course, we're eventually going to delete these anyway. But my PR for removing libtorio is currently set up to depend on this one! What a mess! So here's an intermediate solution: I'll make a separate PR based on the `no_wrap_cyc_decoder` branch just to delete the *documentation* for libtorio rather than *all* of libtorio. This shouldn't break anything. This PR will depend on *that* one. And then the final libtorio removal PR will depend on this one.

examples/asr/emformer_rnnt/pipeline_demo.py
NicolasHug: I was hoping we could just merge this PR while leaving the example folder as-is. It'd be in a broken state but that only temporary. 

As long as the unit-test CI job is green this should be fine (or are the unit-tests broken if we keep the file?!)

samanklesaria: Hm.. I don't remember whether it was just docs breaking or tests too. I don't want to reset this PR back to the commit before I added the example file in case I need to undo it. Maybe I'll just try running tests locally on that commit and see what happens.

samanklesaria: Yes, it's just the docs build that fails when we don't remove the example. No unit tests fail. I'll reset the PR back to before I removed the example.

PR #4051, opened 2025-08-19
Port overdrive to stable ABI

samanklesaria:
This PR uses the Accessor class defined in the PR for `forced_align`, so it is a child of that branch. Otherwise, this PR ports all accessor use to the Accessor class and all at::Tensor use to stable tensors.


pearu:
Closing as https://github.com/pytorch/audio/pull/4131 has landed.

PR #4052, opened 2025-08-19
[STABLE ABI] Stable Accessor shim

samanklesaria:
Although Accessors will eventually be ported to the stable ABI, this may take some time (as Array Refs need to have a representation in the stable ABI first). This PR offers an independent Accessor functionality we can use for the time being. This allows us to make progress on the stable ABI porting (which requires accessor functionality). Especially when it comes to 3 dimensional tensors and higher, writing the stride multiplication logic manually and indexing into the flat array produces code that is not particularly readable and makes it easy to make mistakes. I think using a simple Accessor abstraction is a better way forward. This PR includes tests for the abstraction. The tests are only defined for integer tensors, but as the accessor is templated, the index calculation logic does not depend on the underlying datatype, and we should not need to test other datatypes.

### DIFF
--- a/src/libtorchaudio/accessor.h
+++ b/src/libtorchaudio/accessor.h
@@ -1,6 +1,6 @@
 #pragma once
 
-#include <torch/csrc/stable/tensor.h>
+#include <torch/csrc/stable/tensor_struct.h>
 #include <type_traits>
 #include <cstdarg>
 


pearu:
Closing as torch stable ABI provides tensor accessors.

PR #4055, opened 2025-08-20
Remove deprecation of CUTC decoder

samanklesaria:
We deprecated the CUTC decoder for the 2.8 release as we thought we might need to remove all C++ extensions from the package. However, realizing now that this does not need to be the case, we should un-deprecate the functionality.

### DIFF
--- a/src/torchaudio/models/decoder/__init__.py
+++ b/src/torchaudio/models/decoder/__init__.py
@@ -1,4 +1,3 @@
-from torchaudio._internal.module_utils import dropping_support, dropping_class_support
 import inspect
 _CTC_DECODERS = [
     "CTCHypothesis",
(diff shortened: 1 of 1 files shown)


NicolasHug:
thanks, let's merge it at the end once we finished the migrations

samanklesaria:
What other deprecation logic needs to be removed? A git grep (after merging the current PRs in the queue - see `psuedo_main` for what this looks like) gives no results.

samanklesaria:
I actually believe this PR needs to be merged in next, due to a change in how sphinx is working nowadays. In all other PRs now, we are getting an error like this when building documentation locally:

```
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This suggests that if we just merged *this* PR first, which eliminating the CUTC warnings, this error would go away. Note that this error does not show up in the CI: the CI logs for sphinx only show the exit code and suppress all other output.

pearu:
This PR requires:
- resolve conflicts
- be merged before https://github.com/pytorch/audio/pull/4056 or the changes here need to moved to https://github.com/pytorch/audio/pull/4056

samanklesaria:
I'm not actually sure this should be added to the 2.9 milestone. I believe Nicholas only wanted to remove the deprecations once the stable ABI ports are complete. I'll resolve the new conflicts, but I think this PR should be set aside for now.

pearu:
This is included in https://github.com/pytorch/audio/pull/4147, closing.

PR #4057, opened 2025-08-20
Misc documentation cleanup

samanklesaria:
This PR mostly changes references to that were phrased as "will be" (in the context of deprecating IO functionality) to "have been". A variety of issues noticed along the way that were too small for their own PRs have been included as well.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -4,16 +4,16 @@ Torchaudio Documentation
 .. image:: _static/img/logo.png
 
 Torchaudio is a library for audio and signal processing with PyTorch.
-It provides I/O, signal and data processing functions, datasets,
+It provides signal and data processing functions, datasets,
 model implementations and application components.
 
 .. note::
     Starting with version 2.8, we are refactoring TorchAudio to transition it
     into a maintenance phase. As a result:
 
-    - Some APIs are deprecated in 2.8 and will be removed in 2.9.
+    - Some APIs were deprecated in 2.8 and removed as of 2.9.
     - The decoding and encoding capabilities of PyTorch for both audio and video
-      are being consolidated into TorchCodec.
+      have been consolidated into TorchCodec.
 
     Please see https://github.com/pytorch/audio/issues/3902 for more information.
 


docs/source/index.rst
```diff
@@ -8,12 +8,12 @@ It provides I/O, signal and data processing functions, datasets,
...
 .. note::
-    Starting with version 2.8, we are refactoring TorchAudio to transition it
+    Starting with version 2.8, TorchAudio has transitioned
     into a maintenance phase. As a result:
 
-    - Some APIs are deprecated in 2.8 and will be removed in 2.9.
+    - Some APIs were deprecated in 2.8 and removed in 2.9.
     - The decoding and encoding capabilities of PyTorch for both audio and video
-      are being consolidated into TorchCodec.
+      have been consolidated into TorchCodec.
```
NicolasHug, line 16: Below we are including the `build.*` stuff with lots of references to ffmpeg. We should remove those in a follow-up PR.

docs/source/torchaudio.rst
```diff
@@ -34,75 +33,3 @@ it easy to handle audio data.
    save
    save_with_torchcodec
    list_audio_backends
-
```
NicolasHug, line 36: `info` is still listed above but it was removed. Same with `list_audio_backends`. Let's follow-up in another PR

examples/libtorchaudio/speech_recognition/README.md
```diff
@@ -2,10 +2,6 @@
 
 This example demonstarates how you can use torchaudio's I/O features and models to run speech recognition in C++ application.
 
-**NOTE**
-This example uses `"sox_io"` backend for loading audio, which does not work on Windows. To make it work on
-Windows, you need to replace the part of loading audio and converting it to Tensor object.
-
```
NicolasHug: Let's put back this file as it originally was, and create a follow-up PR for all the `example/` at once. Let's add a notice the README of all the `examples/` that aren't tutorials and indicate that a bunch of those are probably relying on removed APIs and that these example only exist for reference but may not work as-is.

docs/source/index.rst
NicolasHug: Above there's a mention that torchaudio supports `IO` stuff which isn't reallt the case anymore. We should edit that, and edit the same statement in the main `README.md` in a follow-up PR.

NicolasHug (approved):
Thanks for the PR, approving with requests for follow-up in other PR. Since this PR is made against `remove_dropping_support` and not against `main`, please be extra careful with the merge.

examples/libtorchaudio/speech_recognition/README.md
samanklesaria: Done. I will add a notice in the readme in a separate PR.

docs/source/torchaudio.rst
samanklesaria: Removed.

docs/source/index.rst
samanklesaria: I removed the reference to IO. By "removing these in a follow up PR", do you mean removing the special ffmpeg-specific install instructions for each platform?

NicolasHug: yes

PR #4058, opened 2025-08-20
Enable Mac OS tests

samanklesaria:
Mac tests were disabled previously, as dealing with dependencies were causing problems. Now that dependencies have been removed, we should be able to re-enable them. Note that this PR depends on #4057.

### DIFF
--- a/.github/workflows/unittest-macos-cpu.yml
+++ b/.github/workflows/unittest-macos-cpu.yml
@@ -1,64 +1,65 @@
-# name: Unit-tests on Macos CPU
+name: Unit-tests on Macos CPU
 
-# on:
-#   pull_request:
-#   push:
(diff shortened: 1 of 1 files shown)


pearu:
There are lots of changes in this PR that does not match the title. Is this PR (as it is) still relevant?

samanklesaria:
This PR was originally based on a different branch than main, as there were a long chain of PRs that took a while to finally get merged. I had started to use `pseudo-main` rather than `main`, as the base. Now, the base has been changed back to `main`. The way things eventually got merged in was a little different, which is why so many extra changes show up here. Ideally, this PR would only contain commit [862905a]

samanklesaria:
Handled by #4091

PR #4059, opened 2025-08-20
Enable windows tests

samanklesaria:
Windows tests were disabled as dependency installation was failing. Now that most dependencies are removed, we should try again.


pearu:
There are lots of changes in this PR that does not match the title. Is this PR (as it is) still relevant?

samanklesaria:
This PR is really just for commit [0607381].  The rest are from merges with a pseudo-main branch that was used while PRs to the main branch were bottlenecked.

samanklesaria:
We're getting `fatal error C1083: Cannot open include file: 'torch/csrc/stable/library.h': No such file or directory
2025-09-02T21:40:26.6551037Z `. Perhaps the windows build isn't using pytorch nightly?

samanklesaria:
Now we're seeing "Cannot open include file: 'cuda_fp16.h': No such file or directory" for cuda builds.

samanklesaria:
The CTC decoder isn't being built even for the cuda jobs. Maybe that's okay for now? I think just getting the windows tests working, even if they don't test the ctc stuff, is a good place to have this PR.

samanklesaria:
CI is passing now! But we use `conda run` rather than `activate`. As #4085 uses the old "activate" style and also passes the tests, I will close this PR in favor of that one.

PR #4060, opened 2025-08-20
[STABLE ABI] Port lfilter to stable ABI

samanklesaria:
This PR ports lfilter to the stable ABI. The c++ generic interface (for non-cpu and non-cuda devices) is temporarily removed, an will be added back in python for now. The code still uses parallel_for for now. Although the code currently works off of the `stable_accesor` branch (incorporating those commits), it can be rebased to depend only on `main` if necessary.


pearu:
Closing as https://github.com/pytorch/audio/pull/4135 has landed.

PR #4061, opened 2025-08-20
Enable all tests

samanklesaria:
Now that deprecated functionality has been removed, we can remove the restrictions given when running the test suite during CI. This PR depends on the functionality in the branch `pseudo_main`.

### DIFF
--- a/.github/scripts/unittest-linux/run_test.sh
+++ b/.github/scripts/unittest-linux/run_test.sh
@@ -34,5 +34,5 @@ fi
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_inflect=true
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_pytorch_lightning=true
     cd test
-    pytest torchaudio_unittest -k "not backend and not /io/ and not prototype and not ffmpeg and not fairseq and not hdemucs and not (torchscript and rnnt) and not torchscript_consistency"
+    pytest torchaudio_unittest
 )


samanklesaria:
The torchscript consistency tests don't seem to work with manual autograd in python (torch.autograd.Function overloads). Pretty much all the torchscript consistency checks seem to have this issue. The only way forward is to either:
A) Investigate more whether it's possible to get torch.autograd.Function working with torchscript
B) Delete the torchscript consistency checks. 
As torchscript is essentially deprecated anyway, I propose option (B)

pearu:
It is possible to fix torchscript consistency tests by avoiding autograd functions under jit-scripting. For instance, `rnnt_loss` tests pass when applying the following patch:
```diff
diff --git a/src/torchaudio/functional/functional.py b/src/torchaudio/functional/functional.py
index 5d4284d5..a72c6091 100644
--- a/src/torchaudio/functional/functional.py
+++ b/src/torchaudio/functional/functional.py
@@ -1821,7 +1821,13 @@ def _rnnt_loss(
     if blank < 0:  # reinterpret blank index if blank < 0.
         blank = logits.shape[-1] + blank
 
-    costs = RnntLoss.apply(
+    if torch.jit.is_scripting():
+        # TorchScript does not support autograd functions
+        func = torch.ops.torchaudio.rnnt_loss_forward
+    else:
+        func = RnntLoss.apply
+
+    costs = func(
         logits,
         targets,
         logit_lengths,
@@ -1831,6 +1837,9 @@ def _rnnt_loss(
         fused_log_softmax
     )
 
+    if torch.jit.is_scripting():
+        costs = costs[0]
+
     if reduction == "mean":
         return costs.mean()
     elif reduction == "sum":
```

.github/scripts/unittest-linux/run_test.sh
```diff
@@ -34,5 +34,5 @@ fi
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_inflect=true
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_pytorch_lightning=true
     cd test
-    pytest torchaudio_unittest -k "not backend and not /io/ and not prototype and not ffmpeg and not fairseq and not hdemucs and not (torchscript and rnnt) and not torchscript_consistency"
+    pytest torchaudio_unittest
```
NicolasHug: we can still skip the torchscript-related tests for now until they're fixed.
Also please do the same for the GPU tests.

samanklesaria: Done. GPU tests are already unrestricted besides torchscript.

.github/scripts/unittest-linux/run_test.sh
```diff
@@ -34,5 +34,5 @@ fi
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_inflect=true
     export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_MOD_pytorch_lightning=true
     cd test
-    pytest torchaudio_unittest -k "not backend and not /io/ and not prototype and not ffmpeg and not fairseq and not hdemucs and not (torchscript and rnnt) and not torchscript_consistency"
+    pytest torchaudio_unittest -k "not torchscript"
```
pearu: Locally, I can confirm that the following pytest call
```
pytest test/torchaudio_unittest -k "(not torchscript) and (not (lfilter and cuda)) and not (no_lm_decoder or construct_basic_decoder_2 or construct_basic_decoder_3 or construct_basic_decoder_6 or construct_basic_decoder_7 or custom_kenlm_decoder)"
```
is 100% successful.

The decoder errors are due to missing kenlm support in flashlight-text (building kenlm is broken when using Python 3.13, although, there is https://github.com/kpu/kenlm/pull/473).

So, I suspect that the following
```sh
pytest test/torchaudio_unittest -k "(not torchscript) and (not (lfilter and cuda))
```
is a minimal pytest call provided that all dependencies are installed. Apparently, not all optional dependencies (fairseq, ...) are not available in CI, hence the CI failures.

For now, I suggest to use
```suggestion
    pytest torchaudio_unittest -k "not torchscript and not fairseq and not demucs" 
```
(I did not suggest `(not (lfilter and cuda))` because apparently CI has no issues with the corresponding test.)

pearu (changes requested):

pearu (approved):
LGTM! Thanks, @samanklesaria!

PR #4063, opened 2025-08-21
Remove libtorio

samanklesaria:
This PR removes the deprecated libtorio components of torchaudio. It is a replacement for 4035 which I accidentally merged into the ffmpeg_util removal PR rather than main, as I had forgotten that it depended on the former's functionality.


samanklesaria:
To eliminate the doc-build error, I am making this PR depend on the PR for removing the prototype directory.

samanklesaria:
I have no idea why the linux gpu tests are failing. The "create conda env" step of the gpu build gives "CondaHTTPError: HTTP 000 CONNECTION FAILED for url <https://repo.anaconda.com/pkgs/r/noarch/current_repodata.json>". But nothing about the conda env was changed in this PR. Perhaps this is an upstream problem?

samanklesaria:
The doc errors are because something about the docs CI build changed. When I try it locally (because there's no error messages given in the CI) it seems like it can't deal with the deprecation wrappers. We get the error:
```
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
This suggests that if we just merged the PR for eliminating the CUTC warnings first, this error would go away. I'll try that next.

PR #4064, opened 2025-08-21
Remove torio docs

samanklesaria:
The docs build CI job recently stopped working. Running sphinx locally, it seems like one of the things going wrong is that sphinx might be unable to handle the wrapped CUTC decoder. So to get CI to be green, we need to merge in the removal of the deprecation of the CUTC decoder (4055). The CI for that PR ran before docs CI stopped working, so it's green. Unfortunately, that change by itself isn't quite sufficient for docs building to work yet. We're also getting large number of errors like `libtorio.stream_writer.rst:40: ERROR: Unknown directive type "doxygenfunction"`. Of course, we're eventually going to delete these anyway in 4063. But to get there, we need to merge in 4048. And 4048 was updated recently enough that its docs are failing. So here's the proposed solution: we remove the torio documentation *before* we actually remove libtorio. That way, (hopefully) the documentation errors go away and we can merge 4048, which allows us to merge in 4063.


samanklesaria:
Hm- still not working. Now we get `UndefinedError("'style' is undefined")`.

samanklesaria:
We've decided not to worry about docs builds for now. Closing this PR, as the functionality will be addressed in the later torio removal PR.

PR #3977, opened 2025-07-15
Use stable ABI for RNNT

samanklesaria:

PR #4073, opened 2025-08-30
[STABLE ABI] Port rnnt.

pearu:
As in the title.

Same as https://github.com/pytorch/audio/pull/3977 but using C++ `torch::stable::Tensor` API instead of C shim Tensor API.


samanklesaria:
Looks good to me!

samanklesaria (approved):

src/libtorchaudio/rnnt/cpu/compute.cpp
```diff
@@ -1,77 +1,89 @@
...
+  STD_TORCH_CHECK(
+      logits.size(2) == max_ivalue(target_lengths) + 1,
       "output length mismatch");
-  TORCH_CHECK(
-      targets.size(1) == at::max(target_lengths).item().toInt(),
-      "target length mismatch");
+  STD_TORCH_CHECK(
+      targets.size(1) + 1 == logits.size(2), "target length mismatch");
 
   Options options;
```
NicolasHug, line 89: Nit, we can leave it as a TODO as this isn't something introduced in this PR anyway, but worth knowing for later if we ever clean that up:

```suggestion
  // TODO: Use static_cast and check bounds when down-casting from double to float (clamp_)
  // and from int64 to int (blank_)
  Options options;
```

pearu: Applied, thanks!

amd-sriram:
@NicolasHug @samanklesaria : Is this the final version of RNNT code in torch audio or are you intending to moving out of torch audio ?

pearu:
> @NicolasHug @samanklesaria : Is this the final version of RNNT code in torch audio or are you intending to moving out of torch audio ?

@amd-sriram yes, it is the final version (modulo maintenance modifications). AFAIK, this is no intention to moving RNNT out of torchaudio.

PR #4081, opened 2025-09-02
Modify installation instructions

samanklesaria:
This PR updates the installation instructions for each platform to remove references to ffmpeg.


docs/source/build.rst
```diff
@@ -22,15 +22,12 @@ TorchAudio integrates PyTorch for numerical computation and third party librarie
 Customizing the build
 ---------------------
 
-TorchAudio's integration with third party libraries can be enabled/disabled via
+TorchAudio functionality can be enabled/disabled via
```
pearu: The original text was clear to me.
"TorchAudio functionality" sounds very general to me. Perhaps use plural?

docs/source/build.rst
```diff
@@ -22,15 +22,12 @@ TorchAudio integrates PyTorch for numerical computation and third party librarie
...
 environment variables.
 
 They can be enabled by passing ``1`` and disabled by ``0``.
 
-- ``BUILD_SOX``: Enable/disable I/O features based on libsox.
```
pearu, line 30: There are many references to `sox`:
```sh
$ $ grep -r -I --exclude "*~" sox . | wc
     93     637   12170
```

Would it make sense to create a PR that removes all `sox` related stuff from the entire source tree at once?

docs/source/build.rst
samanklesaria: What remains are CUDA and AMD ROCM features. These are optimizations of the cpu-specific functionality. Technically yes, libcuda is a third party library, but should we really phrase it as such? What about "optimizations" instead of "functionality"?

samanklesaria:
As far as occurances of 'ffmpeg' remaining, there's a couple in `.github/workflows`, there's some lines in CMakelists, there's a smoke test (that I remember Nicholas advocating we keep for backward compatability), there's the torchcodec loading tests (which will be kept), and there's some flags for including parts of common_utils. I was planning on keeping this PR *just* about installation instructions. How about I make a second one for the CMakelist removals, common_utils flag removals, and github workflow removals?

docs/source/build.rst
pearu: Is rnnt an optimization? What about "TorchAudio optional features"?

samanklesaria:
Most of the occurances of 'sox' you're seeing are for sox compatibility tests, which we want to preserve. What remains is:
- A bunch of references to `http://sox.sourceforge.net/sox.html` in `filtering.py` which are just saying that the torchaudio function replications the functionality of a sox function.
- The `apply_codec` function, which is deprecated and needs to be removed. I'll address this function in a separate PR.

docs/source/build.rst
samanklesaria: See comment below.

docs/source/build.rst
pearu: Or "TorchAudio customizations/custom features .."?

samanklesaria: How about "Optional features of torchaudio can be enabled ..."

PR #4083, opened 2025-09-02
Remove ffmpeg use in github workflows

samanklesaria:
Remove the ffmpeg workflow now that we do not directly depend on ffmpeg. Currently depends on `alt-gpu-test-workflow` (#4103)

### DIFF
--- a/.github/scripts/unittest-linux/install.sh
+++ b/.github/scripts/unittest-linux/install.sh
@@ -52,6 +52,7 @@ pip install . -v --no-build-isolation
 
 # 3. Install Test tools
 printf "* Installing test tools\n"
-
+# On this CI, for whatever reason, we're only able to install ffmpeg 4.
+conda install -y "ffmpeg<5"
 
 pip3 install parameterized requests coverage pytest pytest-cov scipy numpy expecttest


.github/workflows/build_docs.yml
```diff
@@ -67,14 +67,11 @@ jobs:
         echo "::endgroup::"
         echo "::group::Install TorchAudio"
         conda install --quiet --yes cmake>=3.18.0 ninja
+        conda install "ffmpeg<5" -c conda-forge
```
pearu: Why is this required?

pearu (changes requested):

.github/workflows/build_docs.yml
pearu: Per [examples/README.md](https://github.com/pytorch/audio/pull/4086/files#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9), we cannot expect that codes under examples work. Apparently these work when ffmpeg is installed properly for torchcodec. 

Do we want to depend on ffmpeg for building torchaudio docs?
^ @NicolasHug

samanklesaria: @pearu we do want the *tutorials* to build. They are the one subdirectory under "examples" that we still expect to work. And for them to work, we need audio loading functionality, which comes from torchcodec, which requires ffmpeg.

samanklesaria: Trying to install ffmpeg from conda rather than from the ffmpeg workflow in the github CI seems like it didn't work (in that conda took forever to resolve the environment). I'm realizing there's no particular need to remove this fmpeg workflow after all. I think this PR should be closed.

pearu: > in that conda took forever to resolve the environment

The conda used here appears to be 2 years old meaning that it does not use the fast solver that all recent conda versions (<= 1 y old) use.

Btw, this is also the reason why setup scripts use the latest miniconda as these come with fast solvers.

So, a fix would be to use miniforge as in #4085 to get rid of the ffmpeg dependency as in this PR.

pearu: We can updated conda for docs build in a separate PR. What do you think?

samanklesaria: Sounds good. I'll copy the miniforge approach into a new PR.

PR #4085, opened 2025-09-03
Use miniforge for conda and enable windows unit-testing in CI

pearu:
As in title.

Previously, conda was installed from miniconda but that appears to be broken for Windows (I think `start /wait ..` is not effective to the miniconda installation processes and the scripts try to use conda while its installation process is still in progress).

Installing conda from miniforge works correctly, although, one must use `curl -L ..` to retrieve the installer, otherwise, its download will be broken.

Same as https://github.com/pytorch/audio/pull/4059 but this PR, as it was the case originally, does not use `conda run ...` commands.


pearu:
Closing as the fix should land in https://github.com/pytorch/audio/pull/4059 .

.github/scripts/unittest-windows/run_test.sh
```diff
@@ -12,5 +12,5 @@ python -m torch.utils.collect_env
 env | grep TORCHAUDIO || true
 
 cd test
-pytest --continue-on-collection-errors --cov=torchaudio --junitxml=${RUNNER_TEST_RESULTS_DIR}/junit.xml -v --durations 20 torchaudio_unittest
+pytest --continue-on-collection-errors --cov=torchaudio --junitxml=${RUNNER_TEST_RESULTS_DIR}/junit.xml -v --durations 20 torchaudio_unittest/functional -k "not torchscript and not fairseq and not demucs and not librosa"
```
samanklesaria: We don't just want to test functional! Can you replace "torchaudio_unittest/functional" with just "torchaudio_unittest" here?

pearu: Absolutely. Fixed.

samanklesaria (approved):
Looks good to me!

PR #4088, opened 2025-09-04
Remove image_utils and PIL reference

samanklesaria:
As the original use of PIL was to test ffmpeg related functionality, but that functionality was removed, we no longer need the PIL dependency or the utilities that use it. This PR removes these parts.


pearu (approved):
LGTM! Thanks, @samanklesaria!

NicolasHug (approved):

PR #4089, opened 2025-09-04
Clarify load and save behavior

samanklesaria:
Now that `load` and `save` are effectively aliases to `load_with_torchcodec` and `save_with_torchcodec`, that should be clarified in the docs.


pearu (approved):
LGTM! Thanks, @samanklesaria!

PR #4090, opened 2025-09-04
Install conda from miniforge rather than using old conda installation with a slow solver

samanklesaria:
This PR is in response to #4083. The old version of conda used by pytorch-infra can't resolve the environment well enough to install ffmpeg, so we have to resort to the manual building step given by the ffmpeg github workflow. But @pearu thinks this problem could go away if we just used a more modern conda. This PR is a work in progress- it installs a fresh miniconda and attempts to activate it before building the rest of the docs.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -62,6 +62,7 @@ jobs:
         git config --global --add safe.directory /__w/audio/audio
         conda create --quiet -y --prefix ci_env python="${PYTHON_VERSION}"
         conda activate ./ci_env
+        conda install -q -y pip
 
         echo "::endgroup::"
         echo "::group::Install PyTorch"
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -62,7 +62,7 @@ jobs:
         git config --global --add safe.directory /__w/audio/audio
         conda create --quiet -y --prefix ci_env python="${PYTHON_VERSION}"
         conda activate ./ci_env
-        conda install -q -y pip
+        conda install -q -y pip numpy
 
         echo "::endgroup::"
         echo "::group::Install PyTorch"
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -50,6 +50,13 @@ jobs:
           export CHANNEL=nightly
           export BUILD_VERSION="$( cut -f 1 -d a version.txt )".dev"$(date "+%Y%m%d")"
         fi
+        
+        # Install minoconda
+        mkdir -p ~/miniconda3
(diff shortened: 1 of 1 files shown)


pearu (commented):
Replace
```
        echo "::group::Build FFmpeg"
        .github/scripts/ffmpeg/build_gpu.sh
```
with
```
conda install -q -y ffmpeg
```
It the doc build time will stay around 15 minutes (as it is now), building ffmpeg within torchaudio can be eliminated.

samanklesaria:
Not working yet- it seems the system pip is still being used, not one associated with miniconda. In general, I find working with all the different kinds of conda spinoffs quite confusing; if I continue to encounter problems with the miniconda approach, I might just switch doing the build with pixi instead.

.github/workflows/build_docs.yml
```diff
@@ -67,12 +75,12 @@ jobs:
...
         echo "::group::Install TorchAudio"
         conda install --quiet --yes cmake>=3.18.0 ninja
-        pip3 install --progress-bar off -v -e . --no-use-pep517
+        pip install --progress-bar off -v . --no-build-isolation
 
         # TODO: Need to rely on torchcodec instead of building ffmpeg from source.
         echo "::endgroup::"
         echo "::group::Build FFmpeg"
-        .github/scripts/ffmpeg/build_gpu.sh
+        conda install -q -y ffmpeg
```
pearu: Notice:
```
  RuntimeError: Could not load libtorchcodec. Likely causes:
            1. FFmpeg is not properly installed in your environment. We support
               versions 4, 5, 6 and 7.
```
but conda installed version is 8. So:
```suggestion
        conda install -q -y ffmpeg<=7
```

.github/workflows/build_docs.yml
```diff
@@ -50,12 +51,19 @@ jobs:
           export CHANNEL=nightly
           export BUILD_VERSION="$( cut -f 1 -d a version.txt )".dev"$(date "+%Y%m%d")"
         fi
+        
+        # Install miniforge
+        mkdir -p ~/miniconda3
```
pearu:
```suggestion
```
as unused.

.github/workflows/build_docs.yml
pearu: Sorry, it should read:
```
conda install -q -y "ffmpeg<=7"
```

pearu:
If waiting for green CI is blocking on https://github.com/pytorch/audio/pull/4083, I suggest merge this PR as the changes in this PR affects only docs build and it was successful (modulo a known test-infra issue).

pearu:
Retire this PR when https://github.com/pytorch/audio/pull/4095 lands.

pearu:
Closing via https://github.com/pytorch/audio/pull/4095

PR #4091, opened 2025-09-05
Enable MacOS M1 CI unittests

pearu:
As in the title.

Functional alternative to https://github.com/pytorch/audio/pull/4058.

### DIFF
--- a/.github/workflows/unittest-windows-cpu.yml
+++ b/.github/workflows/unittest-windows-cpu.yml
@@ -1,4 +1,4 @@
-# name: Unittests on Windows CPU
+name: Unittests on Windows CPU
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/unittest-windows-cpu.yml
samanklesaria, line 1: This shouldn't be part of this commit, no?

pearu: Strictly speaking no, but it is a minor bug fix (fixes the label in the CI list) and I think it should be ok to keep it here.

samanklesaria: Got it. Then looks good to me!

PR #4092, opened 2025-09-05
Enable integration tests

samanklesaria:
Enable CI integration tests.


samanklesaria:
Even though CUDA builds should be turned off, we're still getting compiler errors about missing CUDA. I think I should copy the installation strategy of the other CI scripts and use conda rather than the system packages.

samanklesaria:
The wav2vec2_pipeline_test and rnnt_pipeline_test both read flac assets that get downloaded from "https://download.pytorch.org/torchaudio/". But scipy doesn't allow loading flac files: We get the error " File format b'fLaC' not understood". We really need to use torchcodec for this, not torchaudio. The original reason we had used the scipy hack was because we couldn't get a recent enough ffmpeg version installed on the CI machines. But now that we're completely ignoring the pytorch-infra conda environment and building our own, we might not need to do that. I'll make a separate PR that tries to get the linux unit tests to run without the scipy hack.

PR #4093, opened 2025-09-05
Remove scipy torchcodec hack

samanklesaria:
Previously, we found that the pytorch CI infrastructure made installing ffmpeg beyond version 4 impossible. And this version was buggy, which was making our tests fail. We hacked around the issue by making scipy pretend it was torchcodec. But the integration tests require reading flac files, which aren't supported by scipy. Fortunately, @pearu's miniconda approach to CI seems like it's been able to install more recent ffmpeg versions. So maybe we can do the same thing with the linux test CI!


samanklesaria:
Now part of #4092

PR #4097, opened 2025-09-08
Fix disabled pr-labels CI workflow

pearu:
As in the title.

We'll enable only the name of the workflow to let the CI workflow to succeed.

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -1,4 +1,6 @@
-# name: pr-labels
+name: pr-labels
+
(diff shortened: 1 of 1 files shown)


NicolasHug:
Honestly we probably don't need that, I remember it being noisy in other repos (like torchvision) where we eventually removed it. I'd suggest not re-enabling it.

pearu:
After landing this PR, the pr-labels can be disabled under github Actions.

pearu:
The pr-labels is now manually disabled from Actions.

PR #4095, opened 2025-09-08
Revise docs building CI workflow

pearu:
As in title.

Alternative to https://github.com/pytorch/audio/pull/4090

<strong>PLEASE NOTE THAT THE TORCHAUDIO REPOSITORY IS NO LONGER ACTIVELY MONITORED.</strong> You may not get a response. For open discussions, visit https://discuss.pytorch.org/.

PR #4098, opened 2025-09-09
Revise Linux CPU unittests CI workflow

samanklesaria:
As we've decided against the approach to solving the CI ffmpeg>4 incompatibility issue by installing our own miniconda and ignoring the pytorch-infra environment (begun in #4093), this draft PR tries a different approach: using a standard github runner, as is used to build the docs for torchcodec. ~Note that this PR depends on files included in #4095 , so it has that PR as a target.~

### DIFF
--- a/.github/workflows/unittest-linux-cpu.yml
+++ b/.github/workflows/unittest-linux-cpu.yml
@@ -49,11 +49,11 @@ jobs:
         conda create -c conda-forge --strict-channel-priority -y -n ci_env python="${PYTHON_VERSION}" ffmpeg="${FFMPEG_VERSION}" cmake ninja
         conda activate ci_env
         conda info
-        conda list
         ffmpeg -version
         python -m pip install --upgrade pip
+        # We add conda library path as otherwise torchcodec is not
+        # able to load ffmpeg shared libraries:
         export LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH}
-        echo "LD_LIBRARY_PATH=${LD_LIBRARY_PATH}"
         echo "::endgroup::"
 
         echo "::group::Install TorchAudio test and PyTorch dependencies"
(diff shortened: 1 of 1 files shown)


samanklesaria:
Strange- why isn't CI running any more? Perhaps the force pushing confused github. I had to trigger it manually. The run is at https://github.com/pytorch/audio/actions/runs/17588954128

samanklesaria:
I'm getting the deadly "Could not load libtorchcodec" error from an incorrectly installed ffmpeg. But conda should be installing ffmpeg correctly! Very confusing.

pearu:
> I'm getting the deadly "Could not load libtorchcodec" error from an incorrectly installed ffmpeg. But conda should be installing ffmpeg correctly! Very confusing.

I saw this as well. Adding
```
export LD_LIBRARY_PATH=${CONDA_PREFIX}/lib:${LD_LIBRARY_PATH}
```
fixed that.

.github/workflows/unittest-linux-cpu.yml
```diff
@@ -23,27 +23,64 @@ jobs:
...
           export UPLOAD_CHANNEL=test
         else
           export UPLOAD_CHANNEL=nightly
         fi
 
-        export PATH="${PWD}/third_party/install/bin/:${PATH}"
+        echo "::group::Create conda env"
+        # Mark Build Directory Safe
+        git config --global --add safe.directory /__w/audio/audio
+        conda create -c conda-forge --strict-channel-priority -y -n ci_env python="${PYTHON_VERSION}" ffmpeg="${FFMPEG_VERSION}" cmake ninja
```
NicolasHug, line 49: There should be a conda env already created by `    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main` ?

I also don't undertsand the description of this PR which says we are now relying on native GH runners but we're still using `linux_job_v2` from test-infra - can you clarify the current status and what this adds? If this allows to get rid of our scipy mocking then let's have it all in the same PR

NicolasHug: I see thie conda env is already created on doc job so that's probably OK

PR #4100, opened 2025-09-09
Remove BUILD_SOX and USE_FFMPEG flags

samanklesaria:
Now that sox and ffmpeg are not part of torchaudio, we can remove these build flags from CMakeLists.


CMakeLists.txt
```diff
@@ -166,14 +164,6 @@ else()
 endif()
 
 add_subdirectory(src/libtorchaudio)
-if (USE_FFMPEG)
-  if (DEFINED ENV{FFMPEG_ROOT})
-    add_subdirectory(third_party/ffmpeg/single)
-  else()
-    message(STATUS "Building FFmpeg integration with multi version support")
-    add_subdirectory(third_party/ffmpeg/multi)
```
NicolasHug, line 174: We can probably remove both of these now?

```
~/dev/audio (main*) » tree third_party/ffmpeg/                                   nicolashug@nicolashug-fedora-PW0H326Y
third_party/ffmpeg/
├── multi
│   └── CMakeLists.txt
└── single
    └── CMakeLists.txt
```

pearu: Yes, this is addressed in https://github.com/pytorch/audio/pull/4102

PR #4102, opened 2025-09-09
Remove third_party directory

samanklesaria:
As we no longer depend on sox and ffmpeg, we can remove the `third_party` directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,12 +13,12 @@ cmake_policy(SET CMP0025 NEW)
 # mandatory that we do this (and we don't if cmake is old), but it's
 # nice when it's possible, and it's possible on our Windows configs.
 if(NOT CMAKE_VERSION VERSION_LESS 3.15.0)
-    cmake_policy(SET CMP0092 NEW)
+  cmake_policy(SET CMP0092 NEW)
 endif()
 
 # Suppress warning about ExternalProject_Add timestamp
 if(NOT CMAKE_VERSION VERSION_LESS 3.24.0)
-    cmake_policy(SET CMP0135 OLD)
+  cmake_policy(SET CMP0135 OLD)
 endif()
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -13,12 +13,12 @@ cmake_policy(SET CMP0025 NEW)
 # mandatory that we do this (and we don't if cmake is old), but it's
 # nice when it's possible, and it's possible on our Windows configs.
 if(NOT CMAKE_VERSION VERSION_LESS 3.15.0)
-  cmake_policy(SET CMP0092 NEW)
+    cmake_policy(SET CMP0092 NEW)
```
NicolasHug: avoid making cosmetic changes if possible

third_party/LICENSES_BUNDLED.txt
```diff
@@ -1,11 +0,0 @@
...
-compatibly licensed.  We list some here.
-
-Name: cuctc
-License: BSD-2-Clause (Files without specific notes)
-BSD-3-Clause File:
-     torchaudio/csrc/cuctc/src/ctc_fast_divmod.cuh,
-Apache 2.0 Files:
-     torchaudio/csrc/cuctc/src/bitonic_topk
-  For details, see: cuctc/LICENSE,
-     torchaudio/csrc/cuctc/src/bitonic_topk/LICENSE
```
NicolasHug: out of precaution, we should probably keep this file? We're still including the corresponding sources

CMakeLists.txt
samanklesaria: Forgot to disable autoformatting for CMakeLists. I'll revert these.

third_party/LICENSES_BUNDLED.txt
samanklesaria: Done.

PR #4106, opened 2025-09-11
rely on release/2.9 from test-infra and set version

NicolasHug:

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -57,7 +57,8 @@ jobs:
         echo "::group::Install PyTorch and TorchCodec"
         PYTORCH_WHEEL_INDEX="https://download.pytorch.org/whl/${UPLOAD_CHANNEL}/cu${CU_VERSION_WITHOUT_PERIODS}"
         python -m pip install numpy
-        python -m pip install --pre torch torchcodec --index-url="${PYTORCH_WHEEL_INDEX}"
+        python -m pip install --pre torch --index-url="${PYTORCH_WHEEL_INDEX}"
+        python -m pip install --pre torchcodec --index-url="https://download.pytorch.org/whl/nightly/cu${CU_VERSION_WITHOUT_PERIODS}"
         python -c 'import torch; print(f"{torch.__version__}"); print(f"{torch.__file__}"); print(f"{torch.cuda.is_available()=}")'
         echo "::endgroup::"
 


NicolasHug:
Something seems wrong with the [wheel build jobs](https://github.com/pytorch/audio/actions/runs/17644801797/job/50140055417?pr=4106) as they're installing torch 2.8 instead of 2.9RC

```
++ export BUILD_VERSION=2.9.0+cpu
++ BUILD_VERSION=2.9.0+cpu
++ export PATH=/opt/python/cp39-cp39/bin:/opt/rh/gcc-toolset-13/root/usr/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
++ PATH=/opt/python/cp39-cp39/bin:/opt/rh/gcc-toolset-13/root/usr/bin:/opt/conda/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
++ export 'PIP_INSTALL_TORCH=pip install torch==2.8.0 --index-url https://download.pytorch.org/whl/test/cpu'
++ PIP_INSTALL_TORCH='pip install torch==2.8.0 --index-url https://download.pytorch.org/whl/test/cpu'
```

pearu:
> Something seems wrong with the [wheel build jobs](https://github.com/pytorch/audio/actions/runs/17644801797/job/50140055417?pr=4106) as they're installing torch 2.8 instead of 2.9RC

See

https://github.com/pytorch/test-infra/blob/e1f9a88d4563a71ddbbf3fc1b6948872a7831e2e/.github/workflows/build_wheels_linux.yml#L184

IIUC, we'll need to add `stable_version: "2.9"` to the matrix sections of wheel jobs.

.github/workflows/unittest-linux-gpu.yml
```diff
@@ -63,7 +63,8 @@ jobs:
         echo "::group::Install PyTorch"
         export GPU_ARCH_ID="cu126"  # TODO this is currently hardcoded, should depend on matrix's value.
         PYTORCH_WHEEL_INDEX="https://download.pytorch.org/whl/${UPLOAD_CHANNEL}/${GPU_ARCH_ID}"
-        pip install --progress-bar=off --pre torch torchcodec --index-url="${PYTORCH_WHEEL_INDEX}"
+        pip install --progress-bar=off --pre torch --index-url="${PYTORCH_WHEEL_INDEX}"
+        pip install --progress-bar=off --pre torchcodec --index-url="https://download.pytorch.org/whl/nightly/${GPU_ARCH_ID}"
```
pearu, line 67: Does it mean that `torchcodec` must be released before `torchaudio`?

pearu: According to https://github.com/pytorch/torchcodec/actions/runs/17490234874/job/49679289728, torchcodec 0.7.0 wheels are built against torch 2.8.

NicolasHug: Sort of, but to be more complete on what the issue is:

The previous

```
        pip install --progress-bar=off --pre torch torchcodec --index-url="${PYTORCH_WHEEL_INDEX}"
```

would be trying to install both `torch` and `torchcodec` from the `test` channel, where the RCs are. Torch publishes RCs in there, but we haven't set that up in TorchCodec yet. So it's getting a very very old torchcodec version that we had pushed to `test` channel in the past, and it's not compatible with torch.

The workaround you see here just installs torchcodec from the nightly channel instead, hoping that it's ABI and API-compatible with torch. It's hacky a brittle, but it's the best we have until torchcodec pushes its RC wheel in the `test` channel

NicolasHug:
Waiting on https://github.com/pytorch/test-infra/pull/7150/files to land which should hopefully fix the torch version mismatch

PR #4126, opened 2025-11-04
Retain init args as attributes in MelScale and InverseMelScale

### LINKED ISSUE #4122: MelScale should retain n_stft as an attribute
### 🐛 Describe the bug

The issue with not retaining this argument as an attribute is that it makes an identical object difficult to recreate from the object's attributes. Setting `self.n_stft=n_stft` during `__init__` will make it so that the same object could be re-initialized from the object's attributes (for instance, if we seralize the object to json)

### Versions

PyTorch version: 2.8.0
Is debug build: False

OS: macOS 15.1 (arm64)
Clang version: 16.0.0 (clang-1600.0.26.4)
CMake version: Could not collect
Libc version: N/A

Python version: 3.13.5 | packaged by Anaconda, Inc. | (main, Jun 12 2025, 11:23:37) [Clang 14.0.6 ] (64-bit runtime)
Python platform: macOS-15.1-arm64-arm-64bit-Mach-O
Is CUDA available: False
CUDA runtime version: No CUDA
CUDA_MODULE_LOADING set to: N/A
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M1 Pro

[pip3] mypy_extensions==1.1.0
[pip3] numpy==2.2.0
[pip3] onnx==1.19.1
[pip3] onnx-ir==0.1.11
[pip3] onnxruntime==1.23.2
[pip3] onnxscript==0.5.4
[pip3] pytorch-lightning==2.5.2
[pip3] torch==2.8.0
[pip3] torchaudio==2.8.0
[pip3] torchmetrics==1.8.0
[pip3] torchvision==0.23.0
[conda] numpy                            2.2.0            pypi_0              pypi
[conda] pytorch-lightning                2.5.2            pypi_0              pypi
[conda] torch                            2.8.0            pypi_0              pypi
[conda] torchaudio                       2.8.0            pypi_0              pypi
[conda] torchmetrics                     1.8.0            pypi_0              pypi
[conda] torchvision                      0.23.0           pypi_0              pypi

sammlapp:
Resolves #4122 
MelScale should retain n_stft as an attribute

I observed inconsistency between the MelScale and InverseMelScale in which arguments were retained as attributes. Perhaps there was some reason for this, but I think it probably makes sense to save all arguments as attributes for both classes. This allows serialization of the object for re-initialization with the same parameters.

### DIFF
--- a/src/torchaudio/transforms/_transforms.py
+++ b/src/torchaudio/transforms/_transforms.py
@@ -59,7 +59,6 @@ class Spectrogram(torch.nn.Module):
         >>> spectrogram = transform(waveform)
 
     """
-
     __constants__ = ["n_fft", "win_length", "hop_length", "pad", "power", "normalized"]
 
     def __init__(
@@ -157,7 +156,6 @@ class InverseSpectrogram(torch.nn.Module):
         >>> transform = transforms.InverseSpectrogram(n_fft=512)
         >>> waveform = transform(spectrogram, length)
     """
-
     __constants__ = ["n_fft", "win_length", "hop_length", "pad", "power", "normalized"]
 
     def __init__(
@@ -244,7 +242,6 @@ class GriffinLim(torch.nn.Module):
         >>> transform = transforms.GriffinLim(n_fft=512)
         >>> waveform = transform(spectrogram)
     """
-
     __constants__ = ["n_fft", "n_iter", "win_length", "hop_length", "power", "length", "momentum", "rand_init"]
 
     def __init__(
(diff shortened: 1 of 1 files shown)


pearu (approved):
LGTM! Thanks, @sammlapp!

sammlapp:
Hi, is this likely to be merged soon? I ask because I currently need to keep a patched implementation in my codebase, and would prefer to use the official torchaudio implementation directly

Thanks!

Issue #4075, opened 2025-08-30
[STABLE ABI] Porting overdrive.cpp


pearu:
This issue collects tasks that block porting [overdrive.cpp](https://github.com/pytorch/audio/blob/main/src/libtorchaudio/overdrive.cpp) to use torch stable ABI.

- [x] implement `accessor` template as a `torch::stable::Tensor` template method
      Fix available: https://github.com/pytorch/pytorch/pull/161967
- [x] can we use `at::parallel_for` in torch stable ABI code?
- [x] expose `AT_DISPATCH_FLOATING_TYPES` to stable ABI, currently one need to implement the dispatch logic using `switch` block. Not a blocker but would nice to have.
      For a workaround, see https://github.com/pytorch/audio/issues/4078

^ @NicolasHug @scotts @janeyx99

PR #4131, opened 2025-11-16
[STABLE ABI] Port overdrive

pearu:
As in the title.

Fixes https://github.com/pytorch/audio/issues/4075

### DIFF
--- a/src/libtorchaudio/rnnt/gpu/compute.cu
+++ b/src/libtorchaudio/rnnt/gpu/compute.cu
@@ -14,10 +14,10 @@ using torch::headeronly::ScalarType;
 
 // Entry point into RNNT Loss
 std::tuple<Tensor, Tensor> compute(
-    const Tensor& logits,
-    const Tensor& targets,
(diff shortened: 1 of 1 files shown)


pearu:
Possibly blocked by  https://github.com/pytorch/pytorch/pull/167923

pearu:
`pytorch/audio / build-manywheel-py3_10-rocm7_1` fails because the latest cp310 in nightly is from Nov 13.

pearu:
`Unit-tests on Windows GPU / build / windows-job (pull_request)` fails in forced_align tests and the failure is not related to this PR.

src/libtorchaudio/rnnt/gpu/compute.cu
NicolasHug, line 4: Do we need to remove these headers?

pearu: At some point yes, that is, after revising their usage and having the equivalent tools available in stable ABI/headeronly.

PR #4147, opened 2025-12-07
Undo dropping support for forced_align and rnnt_loss

pearu:
As of https://github.com/pytorch/audio/pull/4142, torchaudio is fully torch ABI stable that includes the extension operations forced_align and rnnt_loss_forward. So, we can continue supporting these in the future.


pearu:
> We can probably remove the other usage of dropping_support and dropping_class_support, as well as their implementation now?

Yes, I have removed `dropping*` tools as well as `deprecated*` ones as these are not used anymore either.

PR #4149, opened 2025-12-09
[STABLE ABI] Eliminate src/libtorchaudio/stable/ops.h

pearu:
This PR requires https://github.com/pytorch/pytorch/pull/169880 to land in nightly.

Also, this PR fixes a bug in `lfilter_core_generic_loop` where `t[:, :, i:i+n]` was incorrectly translated to `narrow(t, 2, i, i + n)`, the correct translation is `narrow(t, 2, i, n)`.

### DIFF
--- a/src/libtorchaudio/utils.h
+++ b/src/libtorchaudio/utils.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <torch/csrc/stable/device.h>
 #include <torch/csrc/stable/ops.h>
 #include <torch/csrc/stable/tensor.h>
 #include <torch/headeronly/core/TensorAccessor.h>


src/libtorchaudio/forced_align/gpu/compute.cu
```diff
@@ -135,7 +135,8 @@ void forced_align_impl(
   const int L = targets.size(1); // label length
   const int S = 2 * L + 1;
 
-  auto targetsCpu = torchaudio::stable::cpu(targets);
+  auto cpu_device = torch::stable::Device("cpu");
+  auto targetsCpu = torch::stable::to(targets, std::nullopt, std::nullopt, cpu_device);
```
mikaylagawarecki: nit: if it helps btw there's a convenience overload for `to` I added that should allow just passing device

https://github.com/pytorch/pytorch/blob/main/torch/csrc/stable/ops.h#L523-L539

same nit below

pearu: Thanks! Nit applied.

PR #4157, opened 2025-12-13
[Python STABLE ABI] Port pybind11 extension modules to torch stable libraries

pearu:
As in the title.

In addition, the PR removes cmake dependency as unused.


src/libtorchaudio/forced_align/cpu/compute.cpp
```diff
@@ -202,9 +193,14 @@ std::tuple<Tensor, Tensor> compute(
       "forced_align_impl",
       AT_WRAP([&] {
         if (targets.scalar_type() == ScalarType::Long) {
-          forced_align_long_impl<scalar_t>(logProbs, targets, blank, paths);
+          (forced_align_impl<scalar_t, int64_t>(
+              logProbs, targets, blank, paths));
+        } else if (targets.scalar_type() == ScalarType::Int) {
+          (forced_align_impl<scalar_t, int32_t>(
```
pearu, line 199: This change fixes access violation in one of the autograd tests on Windows CPU.
It is likely related to MSVC compiler handling of templates that sometimes have mysterious (d)effects.

src/torchaudio/_extension/utils.py
```diff
@@ -52,10 +47,13 @@ def _load_lib(lib: str) -> bool:
             This behavior was chosen because the expected failure case is not recoverable.
             If a dependency is missing, then users have to install it.
     """
-    path = _get_lib_path(lib)
-    if not path.exists():
+    suffix = ".pyd" if os.name == "nt" else ".so"
+    paths = list(_LIB_DIR.glob(f"{lib}*{suffix}"))
```
pearu, line 51: To support ABI components in library paths.

tools/setup_helpers/extension.py
```diff
@@ -43,119 +39,102 @@ def _get_build(var, default=False):
...
+        ],
+    }
+
+    if platform.system() != "Windows":
+        extra_compile_args["cxx"].append("-fdiagnostics-color=always")
+
+    extension = CppExtension
+
+    if _USE_CUDA:
+        extension = CUDAExtension
```
pearu, line 71: The libtorch ABI libraries  are really not Python extension modules, so it's weird to use Cpp/CudaExtension here. Using Cpp/CudaLibrary is more appropriate but torch/tools/cpp_extension does not have these. Although, Cpp/CudaExtension and Cpp/CudaLibrary are fairly similar in their functionalities.

.github/workflows/unittest-linux-cpu.yml
```diff
@@ -35,6 +35,7 @@ jobs:
         export FFMPEG_VERSION="${{ matrix.ffmpeg-version }}"
         export PIP_PROGRESS_BAR=off
         export CONDA_QUIET=1
+        export NO_COLOR=1
```
pearu, line 38: To avoid garbage from pytest output in logs.

PR #4171, opened 2026-01-21
Remove pinning against torch version

NicolasHug:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -86,8 +86,12 @@ def _main():
     print("-- Git branch:", branch)
     print("-- Git SHA:", sha)
     print("-- Git tag:", tag)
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm. Lets target this for 2.10.1 release

adamjstewart:
Isn't pytorch still required? We should keep `torch` without the version requirement.

NicolasHug:
it's required, but installing torchaudio on its own without also installing `torch` is always incorrect anyway (and unfortunately).

adamjstewart:
Yes, but if you list `torch` as a dep (even without the version), pip will install torch when it install torchaudio. That is no longer true after this PR.

PR #4175, opened 2026-02-09
Include file path in audio loading error messages

### LINKED ISSUE #3810: Loading failure errors should indicate what was being loaded when error occured
I was recently working with a project where some of the audio files wouldn't load properly at the end. Since the loading order is shuffled, it was difficult to tell which file triggered the issue. All I knew is I was getting an error here:

https://github.com/pytorch/audio/blob/69b2a0adc2ec03ab99990d7e8be3d4510438c148/src/torchaudio/_backend/ffmpeg.py#L72

Obviously I put a breakpoint on that line, re-ran the project, and worked my way back up the stack to find the problematic file, but it would be nice if the error message gave some details about what it was trying to load. If this is an intentional omission, I understand, but I do think it would be helpful to include that information.

Mr-Neutr0n:
## Summary
- When `torchaudio.load()` fails on a file, the error message now includes the file path/URI that caused the failure
- Updated error messages in `load_with_torchcodec()` for sample rate detection and audio decoding failures to include the URI
- Added a catch-all try/except in `torchaudio.load()` that ensures any unexpected exception also surfaces the file path

Fixes #3810

## Test plan
- [ ] Call `torchaudio.load()` with a non-existent file path and verify the error message includes the file path
- [ ] Call `torchaudio.load()` with a corrupted/invalid audio file and verify the error message includes the file path
- [ ] Call `torchaudio.load()` with a valid audio file and verify it still works correctly
- [ ] Verify that error messages from `load_with_torchcodec()` that already contain the URI are not duplicated

### DIFF
--- a/src/torchaudio/__init__.py
+++ b/src/torchaudio/__init__.py
@@ -83,16 +83,21 @@ def load(
         - Not all audio formats supported by torchaudio backends may be supported
         by TorchCodec.
     """
-    return load_with_torchcodec(
-        uri,
-        frame_offset=frame_offset,
-        num_frames=num_frames,
-        normalize=normalize,
-        channels_first=channels_first,
-        format=format,
-        buffer_size=buffer_size,
-        backend=backend,
-    )
+    try:
+        return load_with_torchcodec(
+            uri,
+            frame_offset=frame_offset,
+            num_frames=num_frames,
+            normalize=normalize,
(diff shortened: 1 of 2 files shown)


Mr-Neutr0n:
Closing — no maintainer activity for 195 days and the repo has moved on. Happy to reopen if the fix is still wanted.

PR #4198, opened 2026-06-01
Fix CUCTC top-k shared memory offset

### LINKED ISSUE #4181: Pointer arithmetic bug in `first_matrix__bitonic_topk_kernel` causes shared memory out-of-bounds write
## Bug

There is a pointer arithmetic bug in `src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu` ([line 298](https://github.com/pytorch/audio/blob/main/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu#L298)) that causes out-of-bounds shared memory writes. 
The bug is latent on most GPU architectures but produces incorrect results or crashes on NVIDIA H100 due to stricter shared memory bounds checking.

## Location

```cpp
float* block_topk_key =
    reinterpret_cast<float*>(smem_buf_bytes + smem_result_byte_offset);
int* block_topk_value =
    reinterpret_cast<int*>(block_topk_key + sizeof(float) * beam);  // ← BUG
```

## impact
- A10 / A100: The out-of-bounds write silently lands on physically present but unallocated shared memory regions. Tests pass, but results may be subtly incorrect.
- H100: Stricter shared memory bounds checking causes data corruption, leading to incorrect CTC decoding results or kernel failures.

## Root Cause
block_topk_key is a float*. Pointer arithmetic operates in units of the pointed-to type, not bytes. Adding sizeof(float) * beam to a float* advances by sizeof(float) × sizeof(float) × beam = 16 × beam bytes, instead of the intended sizeof(float) × beam = 4 × beam bytes.

## Solution
```cpp
int* block_topk_value =
    reinterpret_cast<int*>(block_topk_key + beam);
```

### Versions

main

lanarkite99:
Fixes #4181

This fixes the shared-memory layout used by `first_matrix__bitonic_topk_kernel`.

`block_topk_key` is a `float*`, so pointer arithmetic is in units of `float`, not bytes. The previous code computed `block_topk_key + sizeof(float) * beam`, which advanced the value buffer by `sizeof(float) * beam` float elements instead of by `beam` float elements.

The launch allocates the result region as:

```cpp
beam * sizeof(float) + beam * sizeof(int)

### DIFF
--- a/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu
+++ b/src/libtorchaudio/cuctc/src/ctc_prefix_decoder_kernel_v2.cu
@@ -294,8 +294,7 @@ __global__ __launch_bounds__(BLOCK_SIZE) void first_matrix__bitonic_topk_kernel(
   queue.done();
   float* block_topk_key =
       reinterpret_cast<float*>(smem_buf_bytes + smem_result_byte_offset);
-  int* block_topk_value =
-      reinterpret_cast<int*>(block_topk_key + sizeof(float) * beam);
+  int* block_topk_value = reinterpret_cast<int*>(block_topk_key + beam);
 
   queue.store(block_topk_key, block_topk_value);
   for (int idx = tx; idx < beam; idx += BLOCK_SIZE) {
--- a/test/torchaudio_unittest/models/decoder/cuda_ctc_decoder_test.py
+++ b/test/torchaudio_unittest/models/decoder/cuda_ctc_decoder_test.py
@@ -13,7 +13,7 @@ NUM_TOKENS = 7
 @skipIfNoCuda
 @skipIfNoCuCtcDecoder
 class CUCTCDecoderTest(TempDirMixin, TorchaudioTestCase):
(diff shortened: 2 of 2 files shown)


gilbsgilbs:
Also fixes the older https://github.com/pytorch/audio/issues/3754 . I created https://github.com/pytorch/audio/pull/4199 today, before seeing this PR (and issue https://github.com/pytorch/audio/issues/4181).

lanarkite99:
@gilbsgilbs I dont think this will ever get merged since this library is no longer maintained.  but hopefully this fix helps if anyone else runs into this error.

gilbsgilbs:
I share your pessimism about the fate of this PR, but "maintenance mode" may not mean "unmaintained". The fix is simple, the bug is easily reproduced and affects multiple users, and there's still a tiny possibility that the project is not entirely dead yet. Latest release and commits weren't so long ago, and https://github.com/pytorch/audio/issues/3902 wasn't so pessimistic of a message, so we'll see. Otherwise we can always fork.

PR #4209, opened 2026-07-22
Fix 32-bit index overflow and error-path leak in forced_align (CPU)

### LINKED ISSUE #4208: forced_align (CPU) segfaults when frames * (2*targets+1) exceeds 2^31: 32-bit loop index overflow
## 🐛 Describe the bug

`torchaudio.functional.forced_align` (CPU) crashes the whole process with SIGSEGV when `frames * (2 * target_length + 1)` exceeds 2^31. This is reachable with realistic inputs: character-level CTC alignment (WhisperX-style) of a ~20-minute chunk of dense speech at 50 fps crosses the limit (we hit it in production aligning a German podcast).

### Repro

```python
# needs ~2.2 GB RAM; crashes the interpreter (run in a subprocess)
import torch, torchaudio

T, L, C = 70_000, 15_500, 30            # T*(2L+1) = 2.17e9 > 2^31
g = torch.Generator().manual_seed(0)
emission = torch.log_softmax(torch.rand((T, C), generator=g), dim=-1)
targets = torch.randint(1, C, (1, L), generator=g, dtype=torch.int32)

torchaudio.functional.forced_align(emission.unsqueeze(0), targets, blank=0)
# -> SIGSEGV / SIGBUS
```

Observed behaviour (torchaudio 2.11.0, macOS arm64):

| `T * (2L+1)` | targets dtype | result |
|---|---|---|
| 2.17e9 (> 2^31) | int32 | 💥 SIGSEGV |
| 2.17e9 (> 2^31) | **int64** | 💥 SIGSEGV (no dtype workaround) |
| 2.10e9 (< 2^31) | int32 | ✅ correct result, ~8 s |

### Root cause

`src/libtorchaudio/forced_align/cpu/compute.cpp`, the `backPtr_a` init loop uses a 32-bit loop index against a 64-bit trip count:

```cpp
auto backPtr_a = new int8_t[T * S];     // T, S are int64_t -- allocation is fine
for (int i = 0; i < T * S; i++) {       // int i overflows at 2^31 (UB)
  backPtr_a[i] = -1;
}
```

Once `i` wraps to `INT_MIN`, the store lands at `backPtr_a - 2^31` bytes. The macOS crash report confirms this exactly: fault address = buffer base − `0x80000000`, as a byte write inside `forced_align_impl<float, int>`:

```
Exception Type:    EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0000000892e94000
  x0/x19 (buffer): 0x0000000912e94000       <- base - fault = 0x80000000
0  libtorchaudio.abi3.so  forced_align_impl<float, int>(...) + 316
1  libtorchaudio.abi3.so  torchaudio::alignment::cpu::compute(...) + 1384
```

This also explains why int64 targets crash identically — the dispatch only changes `target_t`, not the init loop. The `alphas_a` init loop two lines above uses the same `int i < 2 * S` idiom (not practically reachable today, but the same latent pattern).

The bug is platform-independent (plain C++ signed-overflow UB); it is present in current `main` (c0cbdb9).

**Correction (2026-08-04):** an earlier version of this line said the bug went back to 2.8. It does not. #4021 (`ea7855b`), which introduced the raw array and the `int` fill loop, merged 2025-08-11 — five days after v2.8.0 shipped on 2025-08-06. v2.8.0 still carries the safe `torch::Tensor backPtr = torch::empty({T, S}, torch::kInt8).fill_(-1)`. The affected range is **2.9.0 through 2.11.0**.

### Fix

One-liner: make the loop indices `int64_t`. PR incoming.

## Versions

- torchaudio 2.11.0 (also inspected current main, c0cbdb9)
- torch 2.13.0
- Python 3.12.8, macOS 26.5 (arm64, M1 Max) — but the overflow is platform-independent

happyarts:
Fixes #4208

## Summary

`forced_align` (CPU) segfaults when `frames * (2 * target_length + 1)` exceeds 2^31, and leaks its whole DP buffer whenever the `targets length is too long for CTC` check fires. Both come from the same handful of lines: `forced_align_impl` allocates two raw buffers, fills them through 32-bit loop indices, and frees them only at the very end of the function.

```cpp
auto backPtr_a = new int8_t[T * S];   // T, S are int64_t -- allocation is fine
for (int i = 0; i < T * S; i++) {     // int i overflows at 2^31 (UB)
  backPtr_a[i] = -1;
}
```

Once `i` wraps, the store lands at `backPtr_a - 2^31` bytes → SIGSEGV. This is reachable with realistic inputs: the crash case below is 23 minutes of ordinary speech at 50 fps and ~11 characters per second. int64 targets crash identically, since the dispatch only changes `target_t`. Full analysis, including a crash report whose fault address is exactly `buffer_base - 0x80000000`, in #4208.

This is a **regression**. #4021 (`ea7855b`, Aug 2025) replaced `backPtr`'s `torch::Tensor ... .fill_(-1)`, which handles 64-bit sizes correctly, with the raw array and the `int` fill loop above. (`alphas` had been converted three days earlier by #4020.) #4021 merged five days after v2.8.0 shipped, so **2.8 is not affected** — the regression reaches users in 2.9 and runs through 2.11.

## The change

Let `std::vector` size and fill both buffers:

```cpp
std::vector<scalar_t> alphas(2 * S, kNegInfinity);
auto alphas_a = alphas.data();

std::vector<int8_t> backPtr(T * S, -1);
auto backPtr_a = backPtr.data();
```

The overflowing loops are gone rather than widened, and the buffers cannot outlive a throw. The accessors, the repeat count and the `targets length is too long for CTC` check also move above the allocations — none of them depend on the buffers, and an unalignable input is now rejected without allocating anything, where before it built the whole `backPtr` first. RAII is then a safety net rather than the mechanism.

Per review feedback from @pearu. The first commit widened the indices to `int64_t`, the second replaces the buffers outright; both are kept so the review history reads in order. **Please squash on merge and keep the second commit's message** — concatenating them would leave `main` claiming the indices were widened, which the final diff does not do.

The pointer aliases stay so the ~90 lines of untouched loop body keep indexing exactly as before, and so the innermost DP loop still stores through a raw pointer. `std::vector` rather than `new_empty` + `fill_` as on the GPU side: there the tensor is load-bearing (it is the destination of a `cudaMemcpyAsync`), here the buffers are function-local scratch that never crosses the ABI boundary. It matches what the sibling CPU kernel already does for host scratch (`rnnt/cpu/cpu_kernels.h`) and preserves the flat 1-D indexing #4020/#4021 introduced. After this change there is no raw `new[]` left anywhere under `src/libtorchaudio/`.

One behaviour change worth naming: `alphas` was freed before the final backtrace loop and now lives to end of scope. That holds `2 * S * sizeof(scalar_t)` longer — 242 KiB against `backPtr`'s 2.02 GiB in the case above — and it is untouched during that loop, so it costs neither cache nor TLB pressure. `shrink_to_fit` would reallocate and copy, which is strictly worse.

## Verification

macOS 27 arm64 (M1 Max), torch 2.13.0, against a **locally built stock `main` with identical flags** — the released wheel is not a fair baseline, its CI toolchain differs.

**The overflow**

| Case | stock `main` | this branch |
|---|---|---|
| T=70 000, L=15 500 (2.17e9 cells), int32 targets | 💥 crash | ✅ correct, 7.2 s |
| same, int64 targets | 💥 crash | ✅ 7.4 s, paths and scores identical to int32 |

**The leak** — all-repeat targets one frame below the threshold (L=15 500, T=2L−2, so `backPtr` is 961 MB per call), RSS after each raise:

| | stock `main` | this branch |
|---|---|---|
| RSS before | 201.8 MB | 199.2 MB |
| RSS after 1 / 2 / 3 raises | 1120.0 / 2036.7 / 2953.4 MB | 200.8 / 200.8 / 200.9 MB |
| peak RSS (`ru_maxrss`) over 3 raises | +2751.9 MB | +1.7 MB |

917 MiB leaked per raise — the whole `backPtr` allocation (`T * S` = 960,968,998 B), plus `alphas` riding along.

The peak row is why the check moved rather than only the ownership: with `std::vector` but the allocations still above the check, the leak is gone but peak RSS is +918.4 MiB, because every rejected input still builds and discards the whole buffer. Below the check it is +1.7 MB.

**No change in results or speed**

- 43 recorded small/medium alignment cases (random emissions, heavy repeats, tight `T == L+R` fits, single-token targets) plus `merge_tokens`: bit-identical
- all six dispatch instantiations — float32/float64/float16 × int32/int64 targets — bit-identical to stock across paths and scores
- `functional_cpu_test.py -k forced_align`, including `test_forced_align_fail`: 12 passed — note this is pre-existing coverage over small inputs, so it shows nothing broke rather than that the fix is covered; a repro for the overflow needs ~2.2 GB of RAM and I did not judge it CI-appropriate, happy to add one if you disagree
- 2.10e9 cells, the largest case stock survives: 6.93 s best / 6.94 s median on **both** builds — the value-init costs nothing measurable

Repro script for the crash in #4208 (needs ~2.2 GB RAM; kills the interpreter without this patch).

### DIFF
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -29,12 +29,12 @@ void forced_align_impl(
   const auto S = 2 * L + 1;
 
   auto alphas_a = new scalar_t[2 * S]; // scalar_t is just logProbs.dtype()
-  for (int i = 0; i < 2 * S; i++) {
+  for (int64_t i = 0; i < 2 * S; i++) {
     alphas_a[i] = kNegInfinity;
   }
 
   auto backPtr_a = new int8_t[T * S];
-  for (int i = 0; i < T * S; i++) {
+  for (int64_t i = 0; i < T * S; i++) {
     backPtr_a[i] = -1;
   }
   auto logProbs_a = torchaudio::accessor<scalar_t, 3>(logProbs);
--- a/src/libtorchaudio/forced_align/cpu/compute.cpp
+++ b/src/libtorchaudio/forced_align/cpu/compute.cpp
@@ -6,6 +6,8 @@
 #include <torch/headeronly/core/Dispatch_v2.h>
 #include <torch/headeronly/core/ScalarType.h>
 
+#include <vector>
+
 namespace torchaudio {
 namespace alignment {
 namespace cpu {
@@ -28,15 +30,6 @@ void forced_align_impl(
   const auto L = targets.size(1);
   const auto S = 2 * L + 1;
 
-  auto alphas_a = new scalar_t[2 * S]; // scalar_t is just logProbs.dtype()
-  for (int64_t i = 0; i < 2 * S; i++) {
-    alphas_a[i] = kNegInfinity;
-  }
-
-  auto backPtr_a = new int8_t[T * S];
-  for (int64_t i = 0; i < T * S; i++) {
-    backPtr_a[i] = -1;
-  }
   auto logProbs_a = torchaudio::accessor<scalar_t, 3>(logProbs);
   auto targets_a = torchaudio::accessor<target_t, 2>(targets);
   auto paths_a = torchaudio::accessor<target_t, 2>(paths);
@@ -54,6 +47,15 @@ void forced_align_impl(
       L,
       ", and number of repeats: ",
       R);
+
+  // scalar_t is just logProbs.dtype(). Both buffers are allocated after the
+  // length check: backPtr is T * S bytes, gigabytes for long audio, and the
+  // vectors own their memory so that a throw cannot leak them.
+  std::vector<scalar_t> alphas(2 * S, kNegInfinity);
+  auto alphas_a = alphas.data();
+
+  std::vector<int8_t> backPtr(T * S, -1);
+  auto backPtr_a = backPtr.data();
   auto start = T - (L + R) > 0 ? 0 : 1;
   auto end = (S == 1) ? 1 : 2;
   for (auto i = start; i < end; i++) {
@@ -128,14 +130,12 @@ void forced_align_impl(
   auto ltrIdx = alphas_a[S * idx1 + S - 1] > alphas_a[S * idx1 + S - 2]
       ? S - 1
       : S - 2; // alphas_a[idx1][S - 1], alphas_a[idx1][S - 2]
-  delete[] alphas_a;
   // path stores the token index for each time step after force alignment.
   for (auto t = T - 1; t > -1; t--) {
     auto lbl_idx = ltrIdx % 2 == 0 ? blank : targets_a[batchIndex][ltrIdx / 2];
     paths_a[batchIndex][t] = lbl_idx;
     ltrIdx -= backPtr_a[t * S + ltrIdx]; // backPtr_a[t][ltrIdx]
   }
-  delete[] backPtr_a;
 }
 
 std::tuple<Tensor, Tensor> compute(


yoyolicoris:
The scope is small. LGTM!

happyarts:
Gentle ping — this one looks ready and just needs a manual merge.

@yoyolicoris LGTM'd it on Jul 25 and ran `@pytorchbot merge`, but mergebot isn't configured for this repo, so the bot asked for the GitHub merge button instead (comments above). CLA is signed, checks are green, nothing outstanding on my side.

One piece of context that isn't in the description: this is a regression, not a long-standing bug. #4021 (Aug 2025) replaced

```cpp
torch::Tensor backPtr = torch::empty({T, S}, torch::kInt8).fill_(-1);
```

with the raw `new int8_t[T * S]` plus the init loop that this PR fixes. The `fill_` it replaced handled 64-bit sizes correctly, so the segfault first shipped in 2.8 and is still present in 2.11.

@pearu (most recent owner of this file) / @NicolasHug — could one of you merge this, or say what else it needs?

src/libtorchaudio/forced_align/cpu/compute.cpp
```diff
@@ -29,12 +29,12 @@ void forced_align_impl(
   const auto S = 2 * L + 1;
 
   auto alphas_a = new scalar_t[2 * S]; // scalar_t is just logProbs.dtype()
-  for (int i = 0; i < 2 * S; i++) {
+  for (int64_t i = 0; i < 2 * S; i++) {
     alphas_a[i] = kNegInfinity;
   }
```
pearu: Consider eliminating the for-loop, say:
```C++
  std::fill_n(alphas_a, 2 * S, kNegInfinity);
```
or better yet, use vector:
```suggestion
  std::vector<scalar_t> alphas(2 * S, kNegInfinity);
  auto alphas_a = alphas.data();
```
that would not leak memory when TORCH_CHECK raises contrary to never reaching `delete[] alphas_a;`.

Same comment applies to `backPtr_a`.

happyarts:
Done in 71c7950 — `std::vector` for both buffers, so the init loops are gone rather than widened.

I also moved the accessors, the repeat count and the `T >= L + R` check above the allocations, since none of them depend on the buffers. Owning the memory stops the leak, but the buffer was still being built and discarded on every rejected input; with the check first, an unalignable input allocates nothing at all. Peak RSS over three raises, same case as below: +918.4 MB with the vectors above the check, +1.6 MB below it. Say the word if you would rather keep that reorder out of this PR.

You were right that the leak matters, and it is bigger than it looks from the code. `targets length is too long for CTC` is a routine, catchable error that callers retry around, and every raise leaked the whole `backPtr` allocation. Measured with all-repeat targets one frame below the threshold (L=15500, T=2L-2, so 0.96 GB per call):

| RSS | stock `main` | this branch |
|---|---|---|
| before | 201.8 MB | 200.2 MB |
| after 1 raise | 1120.0 MB | 201.8 MB |
| after 2 raises | 2036.7 MB | 201.8 MB |
| after 3 raises | 2953.4 MB | 201.8 MB |

917 MB per raise, exactly the `T * S` allocation.

One behaviour change worth naming: `alphas` used to be freed before the backtrace loop and now lives to end of scope. That holds `2 * S * sizeof(scalar_t)` longer — 242 KiB against `backPtr`'s 2.02 GiB in the #4208 case, and untouched during that loop — so I left it rather than working around it with `shrink_to_fit`, which would reallocate and copy.

Verification on macOS 27 arm64 (M1 Max), torch 2.13.0, against a locally built stock `main` with identical flags (the released wheel is not a fair baseline — different CI toolchain, which cost me a bogus 1.6% regression before I rebuilt):

- 2.17e9 cells: crashes before, correct after, with int32 and int64 targets producing identical paths and scores
- 43 recorded alignment cases plus `merge_tokens`: bit-identical
- all six dispatch instantiations (float32/float64/float16 × int32/int64 targets): bit-identical to stock, paths and scores
- `functional_cpu_test.py -k forced_align`, including `test_forced_align_fail`: 12 passed
- 2.10e9 cells, the largest case stock survives: 6.93 s stock vs 6.92 s patched (best of 3), so the value-init costs nothing measurable

Happy to squash the two commits into one if you would rather have a single change.

happyarts:
Two corrections to what I wrote above, plus a comment tweak. Force-pushed as `362f1a8`; the code is unchanged apart from the comment.

**The affected range is 2.9–2.11, not 2.8.** I had this wrong in the PR description and in #4208, and both are now fixed. #4021 (`ea7855b`) merged 2025-08-11, five days after v2.8.0 shipped on 2025-08-06, so v2.8.0 still has the safe `torch::Tensor backPtr = torch::empty({T, S}, torch::kInt8).fill_(-1)`; the raw array first appears in v2.9.0. `git compare` agrees: `ea7855b` is not an ancestor of `v2.8.0` but is one of `v2.9.0`, `v2.10.0` and `v2.11.0`. Also, only `backPtr` was still a Tensor immediately before #4021 — `alphas` had been converted by #4020 three days earlier.

**"917 MB, exactly the T * S allocation" was sloppy on both counts.** It is 917 MiB (`T * S` = 960,968,998 B = 961 MB decimal), and the throw leaks `alphas` too, so "exactly" was wrong. The table now uses MiB where the numbers are MiB.

**The comment over the allocations no longer describes deleted code.** It read "an int fill index overflowed once that passed 2^31", which is a thing this commit removes; and "known to be alignable" is not true for `L == 0`, where `T >= L + R` passes trivially. It now says:

```cpp
// scalar_t is just logProbs.dtype(). Both buffers are allocated after the
// length check: backPtr is T * S bytes, gigabytes for long audio, and the
// vectors own their memory so that a throw cannot leak them.
```

Everything was re-verified against a freshly built stock `main` on the new head: 44 recorded cases and `merge_tokens` bit-identical, all six dispatch instantiations bit-identical, `functional_cpu_test.py -k forced_align` 12 passed, the 2.17e9-cell case correct for int32 and int64, error path allocating nothing (peak RSS +1.7 MB), and 6.93 s best / 6.94 s median on both builds at 2.10e9 cells.

One thing I should flag rather than leave implicit: **neither fix ships a regression test.** The 12 passing tests are pre-existing and all small, so they show nothing broke, not that the fix is covered — which is roughly how #4021 shipped this in the first place. A repro needs ~2.2 GB of RAM, which I did not think belonged in CI unguarded. If you would like one behind a memory guard, say so and I will add it.

pearu (approved):
LGTM! Thanks, @happyarts!

The CI failure is not related to the changes in this PR.

happyarts:
Another gentle ping — the review side is now fully settled, only the merge is outstanding.

Since my last note the situation changed: @pearu approved on Aug 5 after the `std::vector` rework, on top of @yoyolicoris's LGTM from Jul 25. GitHub still reports the branch as mergeable with no conflicts, and CI was green.

The blocker is purely mechanical: `@pytorchbot merge` isn't configured for this repository, so the bot asked for the GitHub merge button instead — and that step is the one still missing.

Happy to rebase if anything has drifted in the meantime, or to split it further if that would help it land.

*Written with AI assistance, reviewed and approved by me before posting.*

PR #4210, opened 2026-07-23
[release/2.11] Pass extension sources as str, not pathlib.Path

ethanwee1:
## Why

`get_ext_modules()` builds the extension `sources` lists from `_CSRC_DIR / "..."`, i.e. `pathlib.Path` objects. Older setuptools accepted `os.PathLike`, but newer setuptools/distutils now asserts that `sources` is a list of `str`:

```
File ".../tools/setup_helpers/extension.py", line 106, in get_ext_modules
    extension(
  ...
  File ".../setuptools/_distutils/extension.py", line 110, in __init__
    raise AssertionError("'sources' must be a list of strings")
AssertionError: 'sources' must be a list of strings
```

This breaks `python setup.py bdist_wheel` with a sufficiently new setuptools.

## Fix

Wrap each source path in `str()` in the three `extension(sources=...)` lists. `include_dirs` are left unchanged (distutils does not assert on them).

### DIFF
--- a/tools/setup_helpers/extension.py
+++ b/tools/setup_helpers/extension.py
@@ -106,8 +106,8 @@ def get_ext_modules():
         extension(
             name="torchaudio.lib._torchaudio",
             sources=[
-                _CSRC_DIR / "_torchaudio.cpp",
-                _CSRC_DIR / "utils.cpp",
+                str(_CSRC_DIR / "_torchaudio.cpp"),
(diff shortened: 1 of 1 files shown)


ethanwee1:
Closing — for ROCm release/2.11 the fix will land in ROCm/audio (see ROCm/audio#19) and the wheel pipeline's related_commits will point at ROCm/audio, not pytorch/audio.

PR #4217, opened 2026-08-12
Raise in vad on non-finite input instead of reporting silence (closes #4216)

### LINKED ISSUE #4216: transforms.Vad silently returns an empty tensor when the waveform contains one non-finite sample
### 🐛 Describe the bug

A single non-finite sample in the leading region of a waveform makes
`torchaudio.transforms.Vad` return an empty tensor. The whole signal is discarded, with no
warning and no exception.

The failure is position-dependent: a NaN inside the speech is harmless, a NaN in the
leading silence that `Vad` scans destroys everything.

```python
import numpy as np, torch, torchaudio

sr = 16000
t = torch.arange(sr * 3) / sr
sig = torch.zeros(sr * 3)
sig[sr:2 * sr] = 0.3 * torch.sin(2 * np.pi * 220 * t[sr:2 * sr]) + 0.05 * torch.randn(sr)

for label, idx in [("clean", None),
                   ("NaN in leading silence (idx 100)", 100),
                   ("NaN inside speech (idx 24000)", 24000)]:
    x = sig.clone()
    if idx is not None:
        x[idx] = float("nan")
    out = torchaudio.transforms.Vad(sr)(x)
    print(f"{label:<34} kept {out.numel()/sr:.2f}s of 3.00s  shape={tuple(out.shape)}")
```

Output:

```
clean                              kept 2.10s of 3.00s  shape=(33600,)
NaN in leading silence (idx 100)   kept 0.00s of 3.00s  shape=(0,)
NaN inside speech (idx 24000)      kept 2.10s of 3.00s  shape=(33600,)
```

`Vad` locates the first frame whose measure crosses `trigger_level`. Every comparison
against NaN is False, so once the running measure is poisoned no frame ever triggers and
the function reports that the entire input was silence.

Nothing in `torchaudio` rejects the input at any stage:

```python
x = torch.full((16000,), float("nan"))
torchaudio.transforms.Vad(16000)(x)             # no error
torchaudio.transforms.MelSpectrogram(16000)(x)  # no error
torchaudio.transforms.Resample(16000, 8000)(x)  # no error
```

For reference, the amplification through the other transforms on the same one-NaN input
(input non-finite fraction 0.0000625):

```
MelSpectrogram    NaN fraction out 0.025
Spectrogram       NaN fraction out 0.025
Resample          NaN fraction out 0.002
MFCC              NaN fraction out 1.000     <- DCT spreads it to every coefficient
Loudness          NaN fraction out 1.000
```

### Why this is worth an error rather than documentation

`Vad` is a preprocessing step, so this fails open in the worst direction: a batch job
receives an empty tensor and silently drops the file instead of raising on it. Non-finite
samples arise from truncated decodes, failed resamples, and division by a zero-energy
normalisation window, so the input is not exotic.

For comparison, `librosa` raises `ParameterError: Audio buffer is not finite everywhere`
from `stft`, `melspectrogram`, `resample` and `zero_crossing_rate` on this input.

A `torch.isfinite(waveform).all()` check in `functional.vad`, or more broadly at the
transform boundary, would turn a silent data-loss bug into a clear failure. Happy to open
a PR with a regression test if that is wanted.

### Versions

torchaudio 2.11.0, torch 2.11.0, numpy 1.26.4, Python 3.12, macOS, CPU.

Kayvan-Zahiri:
Closes #4216.

### The bug

`vad` builds a running trigger measure:

```python
mean_meas[i] = mean_meas[i] * trigger_meas_time_mult + meas * (1.0 - trigger_meas_time_mult)
has_triggered = has_triggered or (mean_meas[i] >= trigger_level)
```

One non-finite sample makes `meas` non-finite, and the recurrence then keeps
`mean_meas[i]` non-finite for the rest of the waveform. Every later comparison
against `trigger_level` is false, so nothing triggers and the whole input is
reported as silence. The caller gets an empty Tensor, with no warning and no
exception.

It is position dependent, which is what makes it easy to miss:

```
clean                              kept 2.10s of 3.00s  shape=(33600,)
NaN in leading silence (idx 100)   kept 0.00s of 3.00s  shape=(0,)
NaN inside speech (idx 24000)      kept 2.10s of 3.00s  shape=(33600,)
```

The third row is not a working case. That output still contains the NaN, so
`vad` hands back a waveform that is just as broken, only without the obvious
symptom.

### The fix

Check the input once at the top of `functional.vad` and raise. This file already
raises `ValueError` for out-of-range arguments in several places, so the shape is
consistent with the surrounding code.

### Behaviour change

Input containing NaN or infinity now raises `ValueError` where it previously
returned a Tensor. That includes the "NaN inside speech" case above, which used to
return trimmed audio. Given that output still carried the non-finite sample, an
error seems better than a result the caller is likely to trust.

Clean input is unaffected.

### Verification

- The reproduction in the issue: clean input still returns 2.10s of 3.00s;
  NaN and infinity now raise instead of returning an empty Tensor.
- `torch.isfinite` is safe for every dtype `vad` accepts, including integer
  tensors, where it is true everywhere.
- **TorchScript**: `Vad` is covered by `torchscript_consistency_impl.py`, so I
  checked the guard compiles under `torch.jit.script` and raises correctly when
  scripted. `bool()` on the 0-dim result is required for that.
- No existing test feeds non-finite audio to `vad`, so nothing already in the
  suite changes meaning.
- `transforms_cpu_test.py -k vad`: 24 passed. The 6 new cases (3 values x 2
  dtypes) fail with `AssertionError: ValueError not raised` without the guard.
- flake8 and black clean at the repo's 120 column setting.

I could not build torchaudio from source on this machine, so the behavioural runs
above were done by applying the same guard to an installed 2.11.0 and restoring it
afterwards. CI is authoritative over my numbers.

### Note

@adityaanikam confirmed the root cause on the issue and stood aside for me to open
this, which I appreciated.

I also saw the notice that this repository is no longer actively monitored, so no
expectation of a quick review. The fix is small and self-contained if it is ever
useful.

### DIFF
--- a/src/torchaudio/functional/filtering.py
+++ b/src/torchaudio/functional/filtering.py
@@ -1567,18 +1567,6 @@ def vad(
     """
     device = waveform.device
 
-    if not bool(torch.isfinite(waveform).all()):
-        # The trigger measure is a running mean, so one non-finite sample poisons it
-        # permanently, and every later comparison against `trigger_level` is False.
-        # Nothing ever triggers and the whole waveform is reported as silence, which
-        # returns an empty tensor with no warning. Fail here instead.
-        raise ValueError(
-            "waveform must be finite everywhere, but it contains NaN or infinite values. "
-            "Vad tracks a running measure of the signal, and a single non-finite sample "
-            "makes every later comparison against trigger_level false, so the whole input "
-            "would be discarded as silence."
-        )
-
     if waveform.ndim > 2:
         warnings.warn(
             "Expected input tensor dimension of 1 for single channel"
@@ -1706,6 +1694,19 @@ def vad(
             flushedLen_ns = (measures_len - num_measures_to_flush) * measure_period_ns
             break
     # end for window
+    if not has_triggered and not bool(torch.isfinite(waveform).all()):
+        # Nothing triggered. Either the signal really is silent, or the running trigger
+        # measure was poisoned: one non-finite sample makes it non-finite for the rest of
+        # the waveform, so every later comparison against trigger_level is false and the
+        # whole input is discarded as silence. Only the second case is an error, and the
+        # scan sits in this branch so a waveform that triggers never pays for it.
+        raise ValueError(
+            "waveform must be finite everywhere, but it contains NaN or infinite values. "
+            "Vad tracks a running measure of the signal, and a single non-finite sample "
+            "makes every later comparison against trigger_level false, so the whole input "
+            "would be discarded as silence."
+        )
+
     if not has_triggered and shape[-1] >= fixed_pre_trigger_len_ns:
         return waveform[..., :fixed_pre_trigger_len_ns].view(shape[:-1] + torch.Size([fixed_pre_trigger_len_ns]))
 
--- a/src/torchaudio/functional/filtering.py
+++ b/src/torchaudio/functional/filtering.py
@@ -1567,6 +1567,18 @@ def vad(
     """
     device = waveform.device
 
+    if not bool(torch.isfinite(waveform).all()):
+        # The trigger measure is a running mean, so one non-finite sample poisons it
+        # permanently, and every later comparison against `trigger_level` is False.
+        # Nothing ever triggers and the whole waveform is reported as silence, which
+        # returns an empty tensor with no warning. Fail here instead.
+        raise ValueError(
+            "waveform must be finite everywhere, but it contains NaN or infinite values. "
+            "Vad tracks a running measure of the signal, and a single non-finite sample "
+            "makes every later comparison against trigger_level false, so the whole input "
+            "would be discarded as silence."
+        )
+
     if waveform.ndim > 2:
         warnings.warn(
             "Expected input tensor dimension of 1 for single channel"
--- a/test/torchaudio_unittest/transforms/transforms_test_impl.py
+++ b/test/torchaudio_unittest/transforms/transforms_test_impl.py
@@ -498,3 +498,18 @@ class TransformsTestBase(TestBaseMixin):
         expected_output = torch.zeros(output_shape, dtype=self.dtype, device=self.device)
         result = T.Vad(sample_rate, pre_trigger_time=pre_trigger_time)(inpt)
         self.assertEqual(result, expected_output)
+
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
This adds a full extra pass over every waveform and, on CUDA, `bool(torch.isfinite(...).all())` also forces a device-to-host synchronization before VAD starts. Since this transform is exercised on CUDA too, that can serialize every call. Could the finite check avoid the host sync or be folded into the existing VAD computation?

Kayvan-Zahiri:
Good catch, both halves are fair. Moved the scan off the hot path in 7132688.

The failure this guards against can only present one way: the running measure gets
poisoned, every comparison against `trigger_level` goes false, and nothing ever triggers.
So the scan only needs to run when `has_triggered` is false, right before the branch that
was about to report the whole waveform as silence:

```python
if not has_triggered and not bool(torch.isfinite(waveform).all()):
    raise ValueError(...)
```

A waveform that triggers now pays nothing, no extra pass and no sync. A waveform that does
not trigger pays one scan on a path that was already terminal.

On the sync specifically, worth noting the loop already syncs. Line 1669 is

```python
has_triggered = has_triggered or (mean_meas[i] >= trigger_level)
```

and that `or` calls `bool()` on a 0-dim tensor once per measure period until it triggers.
So the eager check was not introducing a new class of stall, though it did move one to
before any work started, which is the worse place for it. After this change the scan
cannot run at all on a triggering call.

Cost of the scan itself, CPU, median of 20:

```
  10s audio (  160,000 samples):  0.22 ms
  60s audio (  960,000 samples):  0.56 ms
 600s audio (9,600,000 samples):  2.27 ms
```

That is what a triggering waveform no longer pays. I did not benchmark end-to-end VAD
before and after, because on CPU the difference sat inside run-to-run noise at 4.2 to
4.5 ms and I would rather not quote a number that does not separate. I also have no CUDA
device here, so the sync behavior is reasoned from the code rather than measured, and I
would welcome a correction if it does not match what you see.

Behavior is unchanged: NaN, inf and -inf still raise, a clean waveform still trims to
2.10s of 3.00s, and genuine silence still returns an empty tensor without an error, since
silence is not an error.

